PR #49, opened 2025-02-18
Added the first version of Bing Ads connector.

kuievda:
Expected credentials for connector:
1. DeveloperToken
2. ClientID
3. ClientSecret
4. RefreshToken

### DIFF
--- a/src/Integrations/BingAds/Connector.js
+++ b/src/Integrations/BingAds/Connector.js
@@ -1,0 +1,205 @@
+/**
+ * Copyright (c) OWOX, Inc.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+// API Documentation: https://learn.microsoft.com/en-us/advertising/reporting-service/reporting-service-reference
+
+var BingConnector = class BingConnector extends AbstractConnector {
+  //---- constructor -------------------------------------------------
+    constructor(config) {
+      //fields parsing
+        let fields = config.Fields.value.split(",");
+        let storageKeys = [];
+        let storageFields = {};
+        for (var i = 0; i < fields.length; i++) {
+          fields[i] = fields[i].trim();
+          if (fields[i].indexOf("Id") != -1) storageKeys.push(fields[i]);
+          storageFields[fields[i]] = {
+            type: "string",
+            description: ""
+          };
+        }
+        if (fields.includes("TimePeriod")) storageKeys.push("TimePeriod");
+
+      super(config.mergeParameters({
+        AccountID: {
+          isRequired: true,
+          requiredType: "string"
+        },
+        CustomerID: {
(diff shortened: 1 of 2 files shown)


sol-en:
> Expected credentials for connector:
> 
> 1. DeveloperToken
> 2. ClientID
> 3. ClientSecret
> 4. RefreshToken

Great job, thank you! 
This is extremely helpful but I found that there is an issue with data duplication when the connector for the time period overlaps. The `GoogleSheetsStorage` object does not correctly handle the specified keys, leading to unexpected behavior.

I have identified the root cause, but the issue lies within the Core objects, I suppose.

**Steps to Reproduce**

**Report Configuration:**
1. Fields: CampaignId, Clicks, Impressions, Spend, TimePeriod
2. Unique Key: TimePeriod, CampaignId

**Steps:**
1. Ensure that the Data sheet does not exist in the document and that the fields Current Status, Last Import Date, and Last Requested Date are empty.
2. Run the connector and confirm that the Data sheet is created with populated values.
3. Clear a metric field (e.g., Impressions). If the merge logic works correctly, the values should be restored/rewrited.
4. Clear the Current Status, Last Import Date, and Last Requested Date fields to force the logic to work with the full date range instead of relying on the last import.
5. Re-run the connector.

**The Result:**
After rerunning the connector, duplicate rows appear for the same TimePeriod and CampaignId combinations.

**Root Cause Analysis**
The issue stems from how TimePeriod is handled between API responses and Google Sheets.
Thanks!

The API returns TimePeriod in the yyyy-mm-dd format (as a string).
When inserting this value into Google Sheets, it is automatically recognized as a Date object.
When reading back from Google Sheets, the value is interpreted as a Date object rather than a string and is formatted differently than the API response.
As a result, keys from the sheet do not match those from the API, leading to duplicate entries. 🫠


**Potential Solutions**

**1. Set Text Format for Cells Before Inserting Values**
If a field is defined as string in the input schema, enforce text formatting for the corresponding cells in Google Sheets.
This ensures that inserted values remain in their original format.
Expected fix in: `src/Core/GoogleSheetsStorage`

2. Convert Read Values Back to yyyy-mm-dd Format
When reading data from Google Sheets, check if a value is a Date object and convert it to yyyy-mm-dd.
However, this approach risks misinterpreting formats or applying the wrong timezone.
Expected fix in: `src/Core/GoogleSheetsStorage`

3. Transform API Values to Match Google Sheets Format
Convert TimePeriod values from the API to match the format used by Google Sheets.
This is a workaround rather than a real fix, as it modifies original data, which is bad practice—especially if different storage solutions are used in the future.
Expected fix in: `src/Integrations/BingAds/Connector`


**Next Steps**
Would love feedback on the best approach. I’m leaning towards Solution 1, but open to discussion.

zapolsky (approved):

PR #79, opened 2025-04-18
Standardize API limits for Facebook endpoints

AnastasiaKhukhra:
- Set limit to Facebook Marketing API endpoints where it is missed


vladflaks (changes requested):
@AnastasiaKhukhra  I think it would be better to define the limits in `FieldsSchema.js`, as part of the endpoint configuration. This makes it easier to adjust the limits for a specific project without changing the Connectors code.

AnastasiaKhukhra:
> @AnastasiaKhukhra I think it would be better to define the limits in `FieldsSchema.js`, as part of the endpoint configuration. This makes it easier to adjust the limits for a specific project without changing the Connectors code.

Good catch! Fixed.

vladflaks (approved):

PR #149, opened 2025-06-13
Implement DataStorage management

amarchenk0:


apps/backend/src/data-marts/dto/schemas/big-query-config.schema.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from 'zod';
+
+export const BigQueryConfigSchema = z.object({
+  projectId: z.string().min(1, 'projectId is required'),
+  datasetId: z.string().min(1, 'datasetId is required'),
```
max-voloshyn, line 5: `datasetId` is not part of BigQuery Storage config

apps/backend/src/data-marts/services/data-storage-access.validator.ts
```diff
@@ -0,0 +1,75 @@
...
+      return new ValidationResult(true);
+    } catch (e) {
+      this.logger.warn('Invalid config or credentials', e);
+      return new ValidationResult(false, 'Invalid config or credentials');
+    }
+  }
+}
+
+@Injectable()
+export class AthenaAccessValidator implements DataStorageAccessValidator {
```
max-voloshyn, line 54: Is there a way to organize classes so that adding new Storage would result in adding a new directory rather than modifying a bunch of files?

max-voloshyn (changes requested):

PR #198, opened 2025-06-25
Add ShortLinks to FB

AnastasiaKhukhra:


AnastasiaKhukhra:
> @AnastasiaKhukhra is there a way to decouple shortlink parsing from Facebook source? I believe we need such a helper in other sources too.

Agree, done

max-voloshyn (approved):

PR #207, opened 2025-06-27
feat(backend): added cross-platform sqlite db path

laskevych:
This PR introduces enhanced logic for determining the SQLite database file path, improving flexibility and cross-platform compatibility.

**Key Changes:**

*   **Dynamic SQLite Path Resolution:**
    *   A new function `getSqliteDatabasePath` has been implemented in `apps/backend/src/config/data-source-options.config.ts`.
    *   This function allows specifying the SQLite database path via the `SQLITE_DB_PATH` environment variable.
    *   If `SQLITE_DB_PATH` is not set, the database will automatically be stored in the system's application data directory (e.g., `~/Library/Application Support/owox/sqlite/app.db` on macOS) using the `env-paths` library.
    *   Automatically creates necessary directories if they do not exist.

*   **Configuration Updates:**
    *   The `apps/backend/.env.example` file has been updated to use `SQLITE_DB_PATH` instead of `DB_NAME`.
    *   A new `env-paths` dependency has been added to `apps/backend/package.json` for cross-platform path resolution.

**Why was this done?**

These changes provide greater flexibility in deploying and using the backend across different operating systems, as the SQLite database path can now be easily configured or automatically determined according to OS standards.

### DIFF
--- a/apps/backend/.env.example
+++ b/apps/backend/.env.example
@@ -2,7 +2,7 @@ TYPEORM_LOGGING=true
 RUN_MIGRATIONS=false
 DB_TYPE=sqlite
 
-DB_NAME=/path/to/db/backend.db
+SQLITE_DB_PATH=./path/to/database/backend.db
 
 
 DB_HOST=localhost
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -56,6 +56,7 @@
     "@nestjs/typeorm": "^11.0.0",
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.2",
+    "env-paths": "^3.0.0",
     "mysql2": "^3.14.1",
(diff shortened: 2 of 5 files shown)


apps/backend/.env.example
```diff
@@ -2,7 +2,7 @@ TYPEORM_LOGGING=true
 RUN_MIGRATIONS=false
 DB_TYPE=sqlite
 
-DB_NAME=/path/to/db/backend.db
+SQLITE_DB_PATH=./path/to/database/backend.db
```
max-voloshyn, line 5: @laskevych, how would MySQL work?

laskevych: You need to specify the `DB_TYPE` variable in `.env`

amarchenk0 (approved):

PR #214, opened 2025-06-30
feat(org): add markdown linting 

AntonyIshchenko:


max-voloshyn:
@AntonyIshchenko should we place `*.md` linter scripts only in the root package.json script?

AntonyIshchenko:
> @AntonyIshchenko should we place `*.md` linter scripts only in the root package.json script?

@max-voloshyn , I use the same approach as with ESLint and Prettier. We can put all scripts in the root package.json, create commands for each workspace, but first, it wouldn't be consistent with other lint scripts, and second, the CLI configuration would seem to be more complex.

max-voloshyn (approved):

PR #307, opened 2025-07-15
feat(docs): added gtm id from env

laskevych:
Ability to specify GTM ID via environment variables.


apps/docs/src/components/starlight/PageFrame.astro
```diff
@@ -1,11 +1,13 @@
 ---
 import MobileMenuToggle from 'virtual:starlight/components/MobileMenuToggle';
 
+const GTM_ID = import.meta.env.PUBLIC_GTM_ID || '';
```
AntonyIshchenko, line 4: Why don't you use the env-config script to get the PUBLIC_GTM_ID?

apps/docs/src/components/starlight/PageFrame.astro
```diff
@@ -1,11 +1,13 @@
...
+
 const { hasSidebar } = Astro.locals.starlightRoute;
 ---
 
 <!-- Google Tag Manager (noscript) -->
-<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-TVQPWGV6"
+<noscript><iframe src={`https://www.googletagmanager.com/ns.html?id=${GTM_ID}`}
```
AntonyIshchenko, line 10: What do you think about using conditional rendering to render the iframe tag? If we don't have a GTM_ID value, there is no sense in connecting Google Tag Manager.

apps/docs/astro.config.mjs
```diff
@@ -31,7 +31,7 @@ export default defineConfig({
 new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
 j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-})(window,document,'script','dataLayer','GTM-TVQPWGV6');`,
```
AntonyIshchenko, line 34: What do you think about using a conditional adding head tag? If we don't have a gtmId value, there is no sense in connecting Google Tag Manager.

AntonyIshchenko (changes requested):

apps/docs/astro.config.mjs
laskevych: Great idea. Added the condition.

AntonyIshchenko (approved):
👍

PR #312, opened 2025-07-16
Org/script fix

AntonyIshchenko:
### Changes:
- Fixed deploy docs.owox.com
- Fixed build warnings in docs.owox.com related to using sh-session in md files
- Fixed `package-lock.json` file
- Fixed `.prettierignore` file that disables code formatting during development 
- Add `.prettierignore.root` file for `format:root` scripts.

### News:
- Added useful scripts for cleaning and reinstalling dependencies for fixing the `package-lock.json` file.
Code_yxXy09i6HZ


package.json
```diff
@@ -22,6 +22,8 @@
   "scripts": {
     "build": "npm run build --workspaces --if-present",
     "build:docs": "npm run build -w @owox/docs",
+    "clean": "echo '☢️  Starting total cleanup...' && rm -f package-lock.json && find . -name 'node_modules' -type d -prune -exec rm -rf '{}' + && find . -name 'dist' -type d -prune -exec rm -rf '{}' + && echo '✅ Cleanup complete!'",
+    "clean:reinstall": "npm run clean && echo '⚙️  Installing dependencies...' && npm install && echo '✅ Install complete!'",
```
max-voloshyn, line 26: @AntonyIshchenko doesn’t npm ci solve this task? It seems odd to delete the lock file and run npm install, because that would create a new project state rather than simply cleaning the environment

laskevych: I don't understand the value of these scripts. Personally, I have never encountered a need to delete everything. Do we really need this?

laskevych (dismissed):

package.json
```diff
@@ -22,6 +22,8 @@
   "scripts": {
     "build": "npm run build --workspaces --if-present",
     "build:docs": "npm run build -w @owox/docs",
+    "clean": "npx -y shx echo '☢️  Starting total cleanup...' && npx -y shx rm -f package-lock.json && npx -y shx rm -rf node_modules apps/*/node_modules packages/*/node_modules apps/*/dist packages/*/dist + && npx -y shx echo '✅ Cleanup complete!'",
+    "clean:reinstall": "npm run clean && npx -y shx echo '⚙️  Installing dependencies...' && npm install && npx -y shx echo '✅ Install complete!'",
```
max-voloshyn, line 26: Ok, I've got an idea behind these commands 👌 Let's move `npx -y shx rm -f package-lock.json` from `clean` to `clean:reinstall` since removing this file is not part of cleaning, but rather part of reinstalling

package.json
max-voloshyn: @AntonyIshchenko thank you for explanation, I've created new conversation

max-voloshyn (approved):

PR #323, opened 2025-07-21
Add files via upload

ikrasovytskyi:
Cover for the video used on the Connector Building Tutorial


AntonyIshchenko:
@ikrasovytskyi , please, use lower kebab case for asset naming.

PR #326, opened 2025-07-21
fix(connectors): adapt running connectors on Windows in development

AntonyIshchenko:
### Connectors

- Added minimal ordering file dependencies in the **Connector Sources** build process. In Windows, this file sorting by glob pattern is not the same as in Unix-like OS

### Connector-runner

- Fixed running NpmDependencyManager on Windows development environment
- Updated NpmDependencyManager test for Windows environment
- Refactored NpmDependencyManager class: run i/o operation in the constructor and cache results.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,7 @@
     },
     "apps/backend": {
       "name": "@owox/backend",
-      "version": "0.3.0",
+      "version": "0.4.0",
       "license": "ELv2",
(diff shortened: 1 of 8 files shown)


vladflaks:
It’s so great to have a Windows user on the team. 😊

zapolsky (changes requested):

zapolsky (approved):

PR #327, opened 2025-07-21
feat(org): change audit process

AntonyIshchenko:
### What's new

- Add overall audit GitHub workflow

### Changes

- Fixed [@eslint/plugin-kit packege vulnerability](https://github.com/advisories/GHSA-xffm-g5w8-qvg7)
- Updated prepublish scripts - removed audit step

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,6 @@
     "typescript-eslint": "^8.30.1"
   },
   "overrides": {
-    "multer": "2.0.2"
+    "@eslint/plugin-kit": "0.3.3"
   }
 }
--- a/.github/workflows/audit-all.yml
+++ b/.github/workflows/audit-all.yml
@@ -1,0 +1,32 @@
+name: Audit all
+
+on:
+  push:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  audit:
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -20,6 +20,7 @@
     "node": ">=22.16.0"
   },
   "scripts": {
+    "audit": "npm run audit --workspaces --if-present",
```
max-voloshyn, line 23: @AntonyIshchenko does `npm audit` support workspaces out of the box? If so, why do we need audit script?

.github/workflows/test-all.yml
```diff
@@ -99,3 +99,23 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'npm'
+          cache-dependency-path: '**/package-lock.json'
```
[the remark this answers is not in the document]
AntonyIshchenko: Yes, only `cache` property. `cache-dependency-path` is redundant. For details, see https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data

package.json
AntonyIshchenko: Yo're right - it's not working. I will cancel these changes.

PR #332, opened 2025-07-22
introduce to idp-protocol

zapolsky:


packages/idp-better-auth/README.md
```diff
@@ -0,0 +1,285 @@
...
+const provider = new BetterAuthProvider(
+  {
+    issuer: 'your-app',
+    audience: 'your-audience',
+    defaultProjectId: 'default',
+  },
+  {
+    database: db,
+    secret: process.env.AUTH_SECRET!,
+    baseURL: process.env.BASE_URL || 'http://localhost:3000',
```
vladflaks, line 50: I suspect OWOX might be running locally on a port other than 3000, so it’s better to avoid hardcoding the port number

PR #338, opened 2025-07-23
Add README for OWOX Data Marts

max-voloshyn:
### Description
This pull request adds a README file for the OWOX Data Marts project


Obolrus (approved):
good
this description should also be visible on npmjs.com/package/owox

PR #384, opened 2025-08-06
feat: enhance data mart run functionality with Manual backfill [f-5805]

zapolsky:
- Added optional payload parameter to the runDataMart method in the DataMartService.
- Updated RunDataMartCommand to include payload for additional configuration.
- Modified ConnectorExecutionService to handle payload during connector execution.
- Enhanced UI components to support manual run with payload input.
- Updated related DTOs and mappers to accommodate new payload structure.
- Improved connector runner to utilize run configuration for different execution types.


packages/connectors/src/Core/AbstractRunConfig.js
```diff
@@ -61,4 +62,18 @@ var AbstractRunConfig = class AbstractRunConfig {
     
     return true;
   }
+
+  /**
+   * Transforms manual backfill data to the correct type if it's a string
+   * @param {AbstractConfig} config - The connector configuration
+   * @private
+   */
+  _transformManualBackfillData(config) {
```
AnastasiaKhukhra, line 71: @zapolsky please check, if condition `if (this.config[item.configField].requiredType === 'date' && typeof value === 'string')` in AbstractConnector is needed after adding _transformManualBackfillData method.

artemfenoga (approved):

packages/connectors/src/Core/AbstractRunConfig.js
zapolsky: Good catch! 

 I'll review the `_transformManualBackfillData` method to check if the date transformation condition in `_processRunConfig` ( in AbstractConnector) is still needed.
 
With `_transformManualBackfillData` removed:

  1. No date transformation happens during validation - the `validate()` method - only checks field existence and manual backfill support
  2. Date conversion now only happens in `_processRunConfig` - this is the correct place where date strings converted to Date
  3. The condition is necessary - without it, date strings would remain as strings instead of being converted to Date objects
  
Flow 

  1. `AbstractRunConfig.validate()` - validates config fields exist and support manual backfill
  2. `AbstractConnector._processRunConfig()` - handles all data transformation including date conversion

AnastasiaKhukhra (approved):

PR #415, opened 2025-08-14
Update README.md

Obolrus:
added short video with demo


Obolrus (commented):
in this case video is not displayed properly in github, but works for docs.owox.com

AntonyIshchenko (approved):

PR #434, opened 2025-08-20
org: fix preparing release

AntonyIshchenko:
This pull request makes a small update to the release workflow by ensuring that the changelog commit step fails if there are no changes to commit, rather than silently continuing. This improves the reliability of the workflow.

### DIFF
--- a/.changeset/connector-target-setup-editable-defaults.md
+++ b/.changeset/connector-target-setup-editable-defaults.md
@@ -6,5 +6,5 @@
 
 - Added editable dataset/database and table fields with sensible defaults
 - Defaults come from sanitized destination name: dataset/database `${sanitizedDestinationName}_owox`, table `${sanitizedDestinationName}`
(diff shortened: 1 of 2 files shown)


.changeset/connector-target-setup-editable-defaults.md
```diff
@@ -6,5 +6,5 @@
 
 - Added editable dataset/database and table fields with sensible defaults
 - Defaults come from sanitized destination name: dataset/database `${sanitizedDestinationName}_owox`, table `${sanitizedDestinationName}`
-- Inline validation: required, `^[A-Za-z][A-Za-z0-9_]*$`, accessible error state
+- Inline validation: required, `^[A-Za-z][A-Za-z0-9_]*$ `, accessible error state
```
max-voloshyn, line 9: Let's state here only `[A-Za-z][A-Za-z0-9_]` if there is problem with changelog publication 🤪

laskevych (approved):

PR #437, opened 2025-08-20
fix: update package.json for backend

AntonyIshchenko:
This pull request makes a small update to the formatting scripts in the `apps/backend/package.json` file. The change removes the use of a shared `.prettierignore` file by deleting the `--ignore-path ../../.prettierignore` option from both the `format` and `format:check` scripts.


AntonyIshchenko:
Not necessary

PR #469, opened 2025-08-29
Feature/add project id in url

zapolsky:
feat: integrate useProjectRoute hook for dynamic routing in Data Mart components

- Updated multiple components to utilize the useProjectRoute hook for consistent routing.
- Replaced hardcoded paths with scoped paths in AppSidebar, DataMartDetails, DataMartTable, and various empty state components.
- Enhanced navigation handling in DataMartDetailsPage and NotFound page to ensure project context is maintained.

### DIFF
--- a/apps/web/src/shared/hooks/useProjectRoute.ts
+++ b/apps/web/src/shared/hooks/useProjectRoute.ts
@@ -1,0 +1,51 @@
+import { useCallback } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { useProjectId } from './useProjectId';
+
+/**
+ * Hook for navigation with automatic project ID inclusion
+ */
+export function useProjectRoute() {
+
+  const reactRouterNavigate = useNavigate();
+  const projectId = useProjectId();
+
+  /**
+   * Navigate to a route with project ID automatically prepended
+   * @param path - The path relative to the project (e.g., '/data-marts', '/data-marts/123/overview')
+   * @param options - Navigation options
+   */
+  const navigate = useCallback(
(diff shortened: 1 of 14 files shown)


apps/web/src/shared/hooks/useProjectRoute.ts
```diff
@@ -0,0 +1,69 @@
...
+   * @returns Full path with project ID or fallback path
+   */
+  const scope = useCallback(
+    (path: string): string => {
+      if (!projectId) {
+        return path;
+      }
+
+      const normalizedPath = path.startsWith('/') ? path : `/${path}`;
+      const scopedPath = `/ui/${projectId}${normalizedPath}`;
```
romandubovyi, line 52: **Suggestion: Extract hardcoded path prefix to environment variables**

Consider moving the hardcoded `/ui` path prefix to environment variables for better configuration flexibility. This would allow different deployments to use different base paths without code changes.

**Proposed approach:**
- Add `VITE_APP_PATH_PREFIX=/ui` to `.env` file
- Create a utility function to retrieve the prefix: `getPathPrefix()`
```typescript
// apps/web/src/utils/path.ts
export const getPathPrefix = (): string => {
  return import.meta.env.VITE_APP_PATH_PREFIX || '/ui';
};

export const buildProjectPath = (projectId: string, path: string): string => {
  const prefix = getPathPrefix();
  const normalizedPath = path.startsWith('/') ? path : `/${path}`;
  return `${prefix}/${projectId}${normalizedPath}`;
};

``` 
- Update `ProjectRedirect` and `useProjectRoute` to use the configurable prefix

PR #473, opened 2025-09-02
fix: change exit from publish github workflow after publish idp-better-auth

zapolsky:


max-voloshyn:
We should fix "next" part also, @AntonyIshchenko help please

PR #484, opened 2025-09-05
feat(idp-better-auth): enhance MySQL configuration with SSL support

zapolsky:


laskevych:
@zapolsky, I created #524 . They added MySQL SSL support for `@owox/idb-better-auth` and `@owox/backend`.

PR #522, opened 2025-09-19
fix(backend): update logic for retrieving configuration values

AntonyIshchenko:


apps/backend/src/common/scheduler/facades/scheduler-facade.impl.ts
```diff
@@ -31,31 +36,46 @@ export class SchedulerFacadeImpl implements SchedulerFacade {
...
+    const isExecutionEnabled = this.configService.get<boolean>('SCHEDULER_EXECUTION_ENABLED');
+
+    // Registration-only mode: just log and return early
+    if (!isExecutionEnabled) {
+      this.logger.log(`Handler '${handlerName}' registered but execution disabled.`);
+      return;
+    }
+
+    // Worker mode: create cron jobs and full functionality
+    const timezone = this.configService.get<string>('SCHEDULER_TIMEZONE');
```
kalnik-a-a, line 62: Is it safe to use **get** without default fallback value here?

kalnik-a-a (approved):

PR #529, opened 2025-09-23
Feature/create table even without data

AnastasiaKhukhra:
# Enhanced connectors to support CreateEmptyTables configuration option

- Now tables will be created even when no data is fetched, if the CreateEmptyTables parameter is set to "true".


Obolrus:
cool feature.
Please, do not forget to add a .changeset file for changelog

zapolsky (approved):

PR #540, opened 2025-09-24
feat(migration): rename Bing Ads connector to Microsoft Ads

AntonyIshchenko:
This pull request introduces a database migration to update the naming of a connector in the `data_mart` table. The migration ensures consistency by renaming "BingAds" to "MicrosoftAds" and provides a rollback option.


max-voloshyn:
@AntonyIshchenko does it work for both MySQL and SQLite?

max-voloshyn:
@zapolsky is there an idea about a less dangerous way to fix it?

AntonyIshchenko:
> @AntonyIshchenko does it work for both MySQL and SQLite?

Yes

PR #541, opened 2025-09-24
backend/remove-userId

amarchenk0:


max-voloshyn:
I doubt we should remove userId from commands. Couldn’t it be useful for future audit logs or system log context along with projectId?

amarchenk0:
> I doubt we should remove userId from commands. Couldn’t it be useful for future audit logs or system log context along with projectId?
It’s better to add it only where it’s really needed, since right now not all commands have userId anyway.

laskevych (approved):
Good work 👍

PR #581, opened 2025-10-02
feat(web): add Intercom chat widget integration 

romandubovyi:


artemfenoga (approved):

apps/backend/src/idp/controllers/intercom.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+  @Post('jwt')
+  @HttpCode(HttpStatus.OK)
+  @ApiOperation({ summary: 'Returns Intercom user verification JWT' })
+  @ApiOkResponse({
+    description: 'JWT token',
+    schema: { example: { token: 'eyJhbGciOiJIUzI1NiIsInR5...' } },
+  })
+  @ApiBadRequestResponse({ description: 'Missing INTERCOM_SECRET_KEY' })
+  @ApiUnauthorizedResponse({ description: 'Unauthorized' })
+  async issueJwt(@AuthContext() ctx: AuthorizationContext): Promise<{ token: string }> {
```
amarchenk0, line 31: Let’s move the logic from the controller into a separate use-case service and introduce a type for { token: string }.

romandubovyi: Done ✅ Moved business logic to IssueIntercomJwtService use-case. Introduced IntercomTokenResponseApiDto type for the token response. Controller now only handles HTTP concerns and delegates to the use-case

PR #596, opened 2025-10-06
Demo

BadisLaffet1:


vladflaks:
@BadisLaffet1 Could you please unpack the motivation and use case behind these ‘minor’ changes?

PR #666, opened 2025-10-22
Add new Facebook Marketing insights endpoints and improve field schema filtering

prokopetsv:
Introduced multiple insights endpoints with specific breakdowns (age/gender, device platform, product ID, publisher platform/position, region). 

- Annotated fields that do **not** support the `fields` parameter (`supportsFieldsParam: false`)  
- Improved schema definitions for breakdown fields

### DIFF
--- a/.changeset/thirty-years-serve.md
+++ b/.changeset/thirty-years-serve.md
@@ -2,7 +2,7 @@
 'owox': minor
 ---
 
-# Add new acebook Marketing insights endpoints and improve Facebook field schema filtering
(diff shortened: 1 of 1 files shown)


packages/connectors/src/Sources/FacebookMarketing/Source.js
prokopetsv: @AnastasiaKhukhra Please check carefully, as additional attention is needed. While testing, the connector is working properly, but I had to make some changes in the code to prevent errors, and I am not sure if the code is clean. Thank you!

PR #697, opened 2025-10-30
(feat): Rid of npm install connecor runner

zapolsky:
# Remove npm install dependency from connector runner execution

## Summary

This PR refactors the connector execution architecture to eliminate the need for per-run `npm install` operations. The entire `@owox/connector-runner` package has been removed and its functionality has been integrated directly into the `@owox/connectors` package as a runtime script.

## Problem Statement

Previously, each connector run required:
- Creating an isolated Node.js environment
- Installing dependencies via `npm install` in a temporary directory
- Generating execution templates dynamically
- Cleaning up temporary files and dependencies after execution

This approach had several drawbacks:
- **Performance**: Each run incurred significant overhead from npm install operations
- **Complexity**: Complex environment isolation and dependency management logic
- **Resource usage**: Temporary directories and node_modules for each run

## Solution

The connector runner is now a simple Node.js script bundled with the `@owox/connectors` package that:
- Pre-bundles all required dependencies (BigQuery, AWS SDK, etc.)
- Can be executed directly via `node` without any setup
- Eliminates the need for dynamic dependency installation
- Simplifies the execution flow significantly

### DIFF
--- a/.changeset/remove-connector-runner-package.md
+++ b/.changeset/remove-connector-runner-package.md
@@ -1,0 +1,18 @@
+---
+owox: minor
+---
+
+# Refactor connector execution architecture by removing the standalone `@owox/connector-runner` package and integrating its functionality directly into `@owox/connectors` package
+
+**Breaking changes:**
+
+- Removed `@owox/connector-runner` package entirely
+- Moved connector execution logic to `@owox/connectors/src/connector-runner.js`
+- Migrated DTOs to `@owox/connectors/src/Core/Dto/`
+
+**Improvements:**
+
+- Simplified dependency management by consolidating connector-related packages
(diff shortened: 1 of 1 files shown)


apps/backend/src/data-marts/services/connector-execution.service.ts
```diff
@@ -588,8 +679,16 @@ export class ConnectorExecutionService implements OnApplicationBootstrap {
           };
         })
       : [];
-
-    return new RunConfig({
+    this.logger.log(`Creating run config`, {
+      payload,
+      state,
+    });
+    this.logger.log(`Returning run config`, {
```
laskevych: I suggest changing it from `log` to `debug`, because the information is not very useful for the user.

PR #699, opened 2025-10-31
Refactor exception filter to improve type safety and update `IdpError` to implement `HttpError` interface.

amarchenk0:


laskevych (approved):
Good work 👍

PR #707, opened 2025-11-04
feat(web): show connector state

laskevych:
# Show Connector State in Manual Run

- In Manual Run → Advanced for incremental runs, you can now view the Connector State.
- For connectors with multiple configurations, the state is shown for each configuration with a "Created at" tooltip.
- If no state is available, we show "No state available".
- The state is displayed as read-only JSON with a copy option.

# Result
> Chante title Advanced -> State Info
CleanShot 2025-11-04 at 15 39 29
CleanShot 2025-11-04 at 15 40 02


laskevych:
> Maybe we shouldn't repeat "What is the Connector State" info for each config and make it like a paragraph at the start of the Advanced section 🤔

@artemfenoga and I couldn't find such a ready-made component. Usually, users have one configuration.

@AntonyIshchenko FYI

artemfenoga (approved):

apps/backend/src/data-marts/entities/data-mart.entity.ts
```diff
@@ -27,6 +29,9 @@ export class DataMart {
   @JoinColumn()
   storage: DataStorage;
 
+  @OneToOne(() => ConnectorState, cs => cs.dataMart, { eager: true })
```
zapolsky: May be remove `{ eager: true }` for more perfomance and less database query? 

Eager loading means every query for DataMart will automatically join and load ConnectorState, even when its not need. We need state in two place (when DataMart run and in this case), but wuth `{ eager: true }` we will query every tyme it.

Can load with state like this 
```
await this.dataMartRepository.findOne({
  where: { id },
  relations: ['connectorState']
});
```

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
```diff
@@ -125,6 +132,86 @@ export function ConnectorRunForm({ configuration, onClose, onSubmit }: Connector
                 </FormItem>
               )}
             />
+
+            {form.watch('runType') === RunType.INCREMENTAL && (
```
zapolsky: Мaybe its possible to extract this to a separate component?

apps/backend/src/data-marts/entities/data-mart.entity.ts
laskevych: @zapolsky, thank you for pointing that out.

1. I changed 
```
@OneToOne(() => ConnectorState, cs => cs.dataMart, { eager: true })
connectorState?: ConnectorState;
```
to 
```
@OneToOne(() => ConnectorState, cs => cs.dataMart)
connectorState?: ConnectorState;
```

2. I added `relations: ['connectorState'],` in:

```ts
async getByIdAndProjectId(id: string, projectId: string): Promise<DataMart> {
  const entity = await this.dataMartRepository.findOne({
    where: { id, projectId },
    relations: ['connectorState'],
  });

  if (!entity) {
    throw new NotFoundException(`DataMart with id ${id} and projectId ${projectId} not found`);
  }

  return entity;
  }
```

PR #801, opened 2025-12-12
feat(backend): added email-based report run events

laskevych:


apps/backend/src/data-marts/data-destination-types/ee/email/services/email-report-writer.ts
```diff
@@ -175,6 +182,37 @@ abstract class BaseEmailReportWriter implements DataDestinationReportWriter {
       emailHtml
     );
     this.executionLogger?.log({ type: 'email_sent', to: this.emailCredentials.to });
+
+    await this.produceSuccessEvent();
+  }
+
+  private async produceSuccessEvent(): Promise<void> {
```
amarchenk0: Maybe it would be better to extract this logic into a separate component?

PR #802, opened 2025-12-12
feat(facebook): enhance Facebook Marketing Connector with improved data handling and stability

prokopetsv:
Resolves an issue where nested fields were not correctly requested from the API and prevents crashes when parsing invalid date values.


AnastasiaKhukhra (approved):
Well done!

PR #1136, opened 2026-04-09
Stage 3: Add Sharing And Activate Owner Vs Non-Owner Access

zapolsky:


apps/backend/src/migrations/1774800000000-add-availability-columns.ts
```diff
@@ -0,0 +1,64 @@
...
+        name: 'availableForUse',
+        type: 'boolean',
+        isNullable: false,
+        default: true,
+      }),
+      new TableColumn({
+        name: 'availableForMaintenance',
+        type: 'boolean',
+        isNullable: false,
+        default: true,
```
laskevych: By default, we set the value to `true`, but in services we set it to `false`.

It appears that we will set the value to `true` for all existing Data Marts/Storages/Destinations, but to `false` for new ones. If this is what needs to be done, then we should set the default value to `false` and use a separate migration to set it to `true` for all existing ones. In services, set it to `false`.

apps/backend/src/data-marts/services/access-decision/access-decision.service.ts
```diff
@@ -0,0 +1,308 @@
...
+      }
+      default:
+        return OwnerStatus.NON_OWNER;
+    }
+  }
+
+  async getSharingState(
+    entityType: EntityType,
+    entityId: string,
+    _projectId: string
```
laskevych: Should delete `_projectId`? Or you can use this setting for better protection.

apps/backend/src/migrations/1774800000001-add-ownership-user-id-indexes.ts
```diff
@@ -0,0 +1,33 @@
+import { MigrationInterface, QueryRunner, TableIndex } from 'typeorm';
+
+export class AddOwnershipUserIdIndexes1774800000001 implements MigrationInterface {
+  public readonly name = 'AddOwnershipUserIdIndexes1774800000001';
+
+  private readonly indexes = [
+    { table: 'data_mart_technical_owners', column: 'user_id' },
+    { table: 'data_mart_business_owners', column: 'user_id' },
```
laskevych: I suggest creating a composite index, because in the `getOwnerStatus()` method we perform a search based on `entityId + userId`.

apps/backend/src/data-marts/services/access-decision/access-decision.service.ts
```diff
@@ -0,0 +1,308 @@
...
+        entityType,
+        action,
+        Role.ADMIN,
+        OwnerStatus.ADMIN,
+        SharingState.NOT_SHARED
+      );
+    }
+
+    const ownerStatus = await this.getOwnerStatus(userId, entityType, entityId);
+    const sharingState = await this.getSharingState(entityType, entityId, projectId);
```
laskevych: I suggest using `Promise.all()` because there will be a lot of requests in the report lists.

laskevych (approved):
Great job 🏆 Here are a few corrections

PR #1138, opened 2026-04-13
fix(google-sheets): sheet access denial error

prokopetsv:
# Problem

The previous iteration of this message (merged in #1131) already replaced Google's raw "The caller does not have permission" string, but stopped short of explaining what was wrong before telling the user what to do.

# Changes

- Updated the Google Sheets access-denied error in `google-sheets-access-validator.ts` to start with "The account used for authentication doesn't have access to this Google Sheet."
- Updated `.changeset/google-sheets-permission-error.md` to describe the new two-part wording (cause + action)

### DIFF
--- a/.changeset/google-sheets-permission-error.md
+++ b/.changeset/google-sheets-permission-error.md
@@ -4,4 +4,4 @@
 
 # Clearer error when Google Sheet access is denied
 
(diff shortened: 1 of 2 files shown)


kalnik-a-a (approved):
LGTM 👍🏻

PR #1142, opened 2026-04-15
Actualize OpenAPI specs for data mart APIs

max-voloshyn:
## Summary
- Actualize Swagger/OpenAPI specs for data mart, data storage, data destination, connector, report, scheduled trigger, and SQL dry run trigger APIs.
- Add coverage for availability endpoints, member ownership warnings, owner filters, pagination parameters, OAuth error responses, manual run responses, and scheduled trigger behavior.
- Explicitly expose and document SQL dry run trigger status, result, and cancel endpoints.

## Context
- This work was triggered by an incorrect `ListDataMartsSpec`, then expanded to actualize related specs against the current implementation.
- The changes were generated by GPT-5.4 by applying the prompt template `"Actualize X spec considering current implementation"` to the corresponding spec files.

## Testing
- Tested `ListDataMartsSpec`.
- I understand the lack of broader testing here; suggestions for how to properly test these OpenAPI spec-only changes are welcome.

### DIFF
--- a/apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
+++ b/apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
@@ -18,11 +18,11 @@ import { UpdateDataMartTitleApiDto } from '../../dto/presentation/update-data-ma
 import { UpdateDataMartDefinitionApiDto } from '../../dto/presentation/update-data-mart-definition-api.dto';
 import { UpdateDataMartSchemaApiDto } from '../../dto/presentation/update-data-mart-schema-api.dto';
 import { DataMartValidationResponseApiDto } from '../../dto/presentation/data-mart-validation-response-api.dto';
-import { SqlDryRunRequestApiDto } from '../../dto/presentation/sql-dry-run-request-api.dto';
-import { SqlDryRunResponseApiDto } from '../../dto/presentation/sql-dry-run-response-api.dto';
 import { DataMartRunsResponseApiDto } from '../../dto/presentation/data-mart-runs-response-api.dto';
 import { DataMartRunResponseApiDto } from '../../dto/presentation/data-mart-run-response-api.dto';
 import { UpdateDataMartOwnersApiDto } from '../../dto/presentation/update-data-mart-owners-api.dto';
+import { PaginatedDataMartsResponseApiDto } from '../../dto/presentation/paginated-data-marts-response-api.dto';
+import { OwnerFilter } from '../../enums/owner-filter.enum';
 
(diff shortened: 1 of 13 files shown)


apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
```diff
@@ -102,23 +116,51 @@ export function UpdateDataMartOwnersSpec() {
...
   );
 }
 
 export function RunDataMartSpec() {
   return applyDecorators(
     ApiOperation({ summary: 'Manual run DataMart' }),
-    ApiParam({ name: 'id', type: String }),
-    ApiNoContentResponse({ description: 'DataMart run' })
+    ApiParam({ name: 'id', description: 'DataMart ID' }),
+    ApiBody({
```
zapolsky: We have RunDataMartRequestApiDto in https://github.com/OWOX/owox-data-marts/blob/main/apps/backend/src/data-marts/dto/presentation/run-data-mart-request-api.dto.ts

And after changes Spec and real body will not consisteny. 

  Propose:
  1. Use ApiBody({ type: RunDataMartRequestApiDto }) instead of inline schema in RunDataMartSpec — the DTO already has @ApiProperty with description and example, plus
  @MaxJsonSize(1MB) validation that the inline schema doesn't reflect.
  2. ~~Change RunDataMartRequestApiDto schema~~ — Record<string, unknown> is intentional here since payload structure varies by data mart definition type (connector fields,
  etc.). No changes needed in the DTO itself.

zapolsky (approved):

apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
max-voloshyn: @zapolsky good point, fixed 🤝

I replaced the inline `ApiBody` schema in `RunDataMartSpec` with `ApiBody({ type: RunDataMartRequestApiDto, required: false })` so the spec now follows the actual controller DTO.

I also changed `RunDataMartRequestApiDto.payload` from `@ApiProperty` to `@ApiPropertyOptional`. No runtime behavior or payload shape changes there; it only preserves the existing optional semantics in generated OpenAPI after switching the spec to use the DTO.

max-voloshyn: P.S. I also applied the same DTO-based approach to the availability request bodies, replacing inline schemas with the existing availability DTOs and adding Swagger metadata to those DTO fields.

zapolsky:
LGFM

PR #1143, opened 2026-04-15
feat: PME: Stage3: fix availability api access check

zapolsky:
## Summary

- Fixed critical security gap: `AccessDecisionService` was implemented with correct access matrix but not called on single-entity endpoints (GET by ID, PUT, DELETE, RUN). Non-owner users could access "Not available" entities via direct URL even though list filtering worked correctly.
- Enforced ownership + availability checks on **all** entity endpoints: DataMart (8 services), Storage (4 services), Destination (5 services + OAuth), Report (GET + create), and all sub-entity controllers (Scheduled Triggers, Insights, Insight Artifacts, Insight Templates, AI Assistant — 35 endpoints total).
- Fixed report creation on unavailable DataMart: now returns 403 instead of creating the report and showing confusing dual notifications.
- Fixed missing `COPY_CREDENTIALS` check on destination credential copy in `UpdateDataDestinationService`.
- Fixed `canBeOwner()` — now verifies Destination USE access in addition to DataMart visibility when adding Report owners.
- Fixed frontend: mutation 403 errors (e.g. editing title without maintenance access) no longer redirect to "No Access" page — toast is shown instead, user stays on the DataMart page.
- Renamed e2e test files and comments: removed internal "Stage X" references, replaced with "Permissions Model".

## Test plan

- [ ] Run all permissions e2e tests: `npx jest --config apps/backend/test/jest-e2e.json "permissions-|ownership"` → 149 tests pass
- [ ] Verify: non-owner TU cannot access "Not available" DataMart by direct URL → 403
- [ ] Verify: non-owner TU cannot edit DataMart with "Available for reporting" only → 403 + toast (no redirect)
- [ ] Verify: creating report on unavailable DataMart → 403 (no "Report created" toast)
- [ ] Verify: admin always has full access regardless of availability settings
- [ ] Verify: owner always has access to own entities regardless of availability
- [ ] Verify: existing entities (migration defaults: available=true) — no behavior change after deploy


apps/backend/src/data-marts/controllers/ai-assistant-run-trigger.controller.ts
```diff
@@ -25,6 +27,17 @@ export class AiAssistantRunTriggerController extends UiTriggerController<AiRunTr
     @Param('dataMartId') dataMartId: string,
     @Query() query: ListAiAssistantRunTriggersQueryApiDto
   ): Promise<{ data: AiAssistantRunTriggerListItemResponseApiDto[] }> {
+    if (context.userId) {
+      const canSee = await this.accessDecisionService.canAccess(
+        context.userId,
+        context.roles ?? [],
```
laskevych, line 33: Could this be a potentially risky decision `context.roles ?? []`? If there are no roles, we process such a request as if the role were VIEWER. The permissions are minimal, but on the other hand, a user without permissions cannot exist 🤔

apps/backend/src/data-marts/controllers/data-destination.controller.ts
```diff
@@ -195,6 +218,7 @@ export class DataDestinationController {
     @AuthContext() context: AuthorizationContext,
     @Param('id') id: string
   ): Promise<void> {
+    await this.checkDestinationAccess(id, context, Action.DELETE);
     const command = this.mapper.toDeleteCommand(id, context);
     await this.deleteService.run(command);
```
laskevych, line 223: Is it expected that we call two identical access checks:
- 1 - in the controller
- 2 - in the use case
?

For example, in the following endpoints:
- `GET /data-destinations/:id`
- `DELETE /data-storages/:id`
- `GET /data-storages/:id`
- `DELETE /data-marts/:id`
- `GET /data-marts/:id`
- `rotate-secret-key`
- and others

apps/backend/src/data-marts/dto/domain/actualize-data-mart-schema.command.ts
```diff
@@ -1,6 +1,8 @@
 export class ActualizeDataMartSchemaCommand {
   constructor(
     public readonly id: string,
-    public readonly projectId: string
+    public readonly projectId: string,
+    public readonly userId: string = '',
```
laskevych, line 5: It feels like this might be a risky decision. It seems we could end up with both system commands and user commands.

It turns out that if we don’t pass `userId`, we’ll skip the `if (command.userId)` check. I’m not sure if this is what the code expects? Maybe we should somehow split the commands into those that require a userId and those that don’t 🤔

apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
```diff
@@ -89,6 +89,21 @@ export class UpdateDataDestinationService {
...
+        command.roles,
+        EntityType.DESTINATION,
+        command.id,
+        Action.EDIT,
+        command.projectId
+      );
+      if (!canEdit) {
+        throw new ForbiddenException('You do not have permission to edit this Destination');
+      }
+    }
```
laskevych, line 105: Why is the validation performed **inline** in some controllers (`insight-run-trigger`, `insight-template-run-trigger`, `ai-assistant-run-trigger`, `insight-artifact-sql-preview-trigger`), while in others the check has been moved to a **separate private method** `private async checkDataMartAccess()`  (`scheduled-trigger`, `insight`, `insight-artifact`, `ai-assistant`)?

laskevych (approved):
Good job! Please take a look at my comments.

apps/backend/src/data-marts/controllers/ai-assistant-run-trigger.controller.ts
zapolsky: No its safe. 

`@Auth()` decorator guarantees authentication - a request without roles won't pass the guard and won't reach the controller. `?? []` is a defensive fallback for the TypeScript type `roles?: RoleType[]`. 

In practice, roles are always populated after `@Auth()`. If roles were hypothetically undefined, `resolveRole([]) `returns VIEWER (minimal permissions) - this is fail-closed behavior, not fail-open.

apps/backend/src/data-marts/controllers/data-destination.controller.ts
zapolsky: Yes, for Destination endpoints this is duplication (2x DB roundtrip). 

For DataMart and Storage - the check is only in the service layer, no controller-level check. This is an inconsistency, but both approaches are safe - no endpoint is left unprotected. 

Unification (pick one layer) is in the TechDebt task for next stage 4 (6481).

apps/backend/src/data-marts/dto/domain/actualize-data-mart-schema.command.ts
zapolsky: Intentional design choice. 

System callers (trigger handlers, scheduled jobs) don't have user context - they pass `userId: '', roles: []`. 

Access is checked at trigger creation time (controller-level), not at execution time. 
`if (command.userId)` - empty string is falsy → skip. Controller-initiated requests always have a real context.userId from the `@Auth()` decorator. 

The idea to split into UserCommand / SystemCommand is valid and already tracked in  TechDebt task for next stage 4 (6481).

apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
zapolsky: The split is based on endpoint count: trigger controllers have 1-2 endpoints → inline is simpler and more readable. 

CRUD controllers have 5-10 endpoints → private checkDataMartAccess() helper avoids copy-paste. This is effectively a rule, but not documented. 

Will add a comment or extract into a shared Guard in the TechDebt task.

PR #1155, opened 2026-04-21
feat(security): opt-in HTTP security headers for SPA and IDP HTML

laskevych:
## Summary

Apply the five DoD-required HTTP security headers unconditionally on every HTML response
served by OWOX — SPA root, SPA client-side routes, and IDP-rendered auth pages.

- `Strict-Transport-Security: max-age=31536000`
- `Content-Security-Policy: frame-ancestors 'none'`
- `X-Content-Type-Options: nosniff`
- `X-XSS-Protection: 1; mode=block`
- `Referrer-Policy: no-referrer-when-downgrade`

### Scope of coverage

HTML responses only — adjacent JSON APIs stay untouched (CSP on JSON has no purpose and
can break clients).

- **SPA HTML (root + client-side fallback)**: `apps/owox/src/web/static-config.ts` attaches
  headers via `express.static({ setHeaders })` for any `.html` file and via the SPA
  fallback middleware before `res.sendFile('index.html', ...)`.
- **IDP HTML (sign-in, sign-up, magic-link confirm, forgot-password, password setup,
  password success, onboarding, auth-error)**: every controller that renders HTML goes
  through a single `sendSecureHtml(res, html)` helper instead of a blanket `/auth/*`
  middleware. Point-of-callsite application keeps JSON endpoints under `/auth/*` clean.

### Implementation

New shared primitives in `@owox/internal-helpers/security`:

- `SECURITY_HEADERS` — frozen `[name, value][]` list (single source of truth).
- `applySecurityHeaders(res)` — writes all five headers via a minimal `ResponseLike`
  interface (framework-agnostic; works with anything that implements `setHeader`).
- `sendSecureHtml(res, html)` — applies headers then `res.send(html)`. Used by every IDP
  controller HTML callsite.

### 2 Main HTML response (`/`) — all 5 headers must be present

```bash
curl -sI http://localhost:3000/ \
  | grep -iE 'strict-transport-security|content-security-policy|x-content-type-options|x-xss-protection|referrer-policy'
  
 
 ```
 
 Expected:

```
 Strict-Transport-Security: max-age=31536000
Content-Security-Policy: frame-ancestors 'none'
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Referrer-Policy: no-referrer-when-downgrade
```

2.2. SPA route (fallback to index.html) — same 5 headers
```
curl -sI http://localhost:3000/data-marts \
  | grep -iE 'strict-transport-security|content-security-policy|x-content-type-options|x-xss-protection|referrer-policy'
```

Expected: same 5 lines as in 2.1.

2.3. API — headers must NOT be set (scope limited to HTML)

```
curl -sI http://localhost:3000/api/flags \
  | grep -iE 'strict-transport-security|content-security-policy|x-content-type-options|x-xss-protection|referrer-policy'
```
Expected: empty output.

2.4. Static asset (.js / .css) — headers must NOT be set either
Pick any real file from apps/web/dist/assets/:
```
ls /Users/andrewlaskevych/Projects/owox-data-marts-main/apps/web/dist/assets | head -3
# pick e.g. index-XXXX.js
curl -sI http://localhost:3000/assets/index-XXXX.js \
  | grep -iE 'strict-transport-security|content-security-policy|x-content-type-options|x-xss-protection|referrer-policy'
```
Expected: empty output.

2.5. IDP HTML (sign-in) — all 5 headers must be present
```
curl -sI http://localhost:3000/auth/sign-in \
  | grep -iE 'strict-transport-security|content-security-policy|x-content-type-options|x-xss-protection|referrer-policy'
```
Expected: same 5 lines as in 2.1.

### DIFF
--- a/apps/owox/package.json
+++ b/apps/owox/package.json
@@ -34,11 +34,13 @@
     "@types/express": "^5",
     "@types/mocha": "^10",
     "@types/node": "^18",
+    "@types/supertest": "^6.0.2",
     "chai": "^4",
     "eslint-config-oclif": "6.0.146",
     "eslint-config-prettier": "~10.1.0",
     "mocha": "^11",
     "oclif": "^4",
+    "supertest": "^7.0.0",
     "ts-node": "^10"
   },
   "engines": {
--- a/apps/owox/src/commands/serve.ts
+++ b/apps/owox/src/commands/serve.ts
@@ -236,7 +236,6 @@ export default class Serve extends BaseCommand {
 
     const idpProvider = await IdpFactory.createFromEnvironment(this);
     await idpProvider.initialize();
-
     const idpProtocolMiddleware = new IdpProtocolMiddleware(idpProvider);
     idpProtocolMiddleware.register(expressApp);
     expressApp.set('idp', idpProvider);
--- a/apps/owox/src/commands/serve.ts
+++ b/apps/owox/src/commands/serve.ts
@@ -236,6 +236,7 @@ export default class Serve extends BaseCommand {
 
     const idpProvider = await IdpFactory.createFromEnvironment(this);
     await idpProvider.initialize();
+
     const idpProtocolMiddleware = new IdpProtocolMiddleware(idpProvider);
     idpProtocolMiddleware.register(expressApp);
     expressApp.set('idp', idpProvider);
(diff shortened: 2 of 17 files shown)


max-voloshyn:
@laskevych I don’t see a reason for `SECURITY_HEADERS_ENABLED` to exist.

The PR describes these headers as required by the OWOX security DoD, so they should be part of the default application behavior. Making them configurable adds deployment risk without a clear product or operational requirement: any environment that forgets to set the flag silently ships without the required protections.

Unless we have a concrete documented case where these headers break a supported deployment, I’d remove the option and always apply the required security baseline.

amarchenk0 (approved):

PR #1159, opened 2026-04-21
feat(backend): update gbq storage validation

laskevych:
**Preview**
CleanShot 2026-04-21 at 18 13 44

**Summary**

BigQuery storages authenticated via OAuth bypassed `verifyAccess` on update, so invalid `projectId` values (e.g. `GTM-NC2077`) were persisted silently and only surfaced at data-mart runtime.

**Changes**

- `BigQueryConfigSchema.projectId` now enforces the official GCP format via regex (6–30 lowercase letters/numbers/hyphens, must start with a letter, not end with a hyphen, optional `domain:id` prefix).
- `UpdateDataStorageService` runs `BigQueryConfigSchema.safeParse` early for BigQuery/Legacy-BigQuery types — before the OAuth skip — and throws `BadRequestException` with a detailed message that survives `GlobalExceptionFilter` (uses a concatenated string instead of an object payload, which was dropped to `"Invalid config"`).
- Added unit tests: schema positive/negative cases (incl. `GTM-NC2077`) and update-flow tests for OAuth invalid/valid `projectId`.

**Out of scope**

- Live `checkAccess` API call for OAuth on save.
- Regex validation for other storage types (Athena, Snowflake, Databricks, Redshift).

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/schemas/bigquery-config.schema.spec.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/schemas/bigquery-config.schema.spec.ts
@@ -1,0 +1,31 @@
+import { BigQueryConfigSchema } from './bigquery-config.schema';
+
+describe('BigQueryConfigSchema.projectId', () => {
+  describe('valid project IDs', () => {
+    it.each([
+      'my-project',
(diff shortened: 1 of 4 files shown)


prokopetsv (approved):
Awesome!

PR #1177, opened 2026-04-30
feat(backend): correct Looker Studio types and aggregations for blended fields

kalnik-a-a:
## Summary

When a Data Mart pulls fields from joined Data Marts (blended fields), the Looker Studio connector now reports correct column types and respects the user-selected aggregation function for each blended field.

Before this change:
- `COUNT(string_field)` arrived as **STRING / DIMENSION** (should be **NUMBER / METRIC**).
- `COUNT_DISTINCT(...)` was also **STRING / DIMENSION** and was marked re-aggregatable — Looker would sum per-group distincts and inflate totals.
- `MAX(integer)` always produced `defaultAggregationType=SUM` instead of `MAX`.
- `STRING_AGG(...)` / `ANY_VALUE(...)` tried to behave as metrics.
- `getSchema` and `getData` duplicated the type-mapping logic and could drift apart (two independent sources of truth).

After this change: a single source of truth + storage-aware effective type + explicit mapping of aggregation to Looker concept.

## Approach

**Hybrid: effective type on the produce side + aggregateFunction propagated separately.**

1. **`computeEffectiveType(rawType, aggFunc, storageType): StorageFieldType`** — pure function returning the storage-specific value type after aggregation:

   | `aggFunc` | Effective type |
   |---|---|
   | `undefined` | `rawType` |
   | `COUNT` / `COUNT_DISTINCT` | INTEGER (or `INT` for Databricks) |
   | `STRING_AGG` | STRING (or `VARCHAR` for Redshift) |
   | `SUM` / `MIN` / `MAX` / `ANY_VALUE` | `rawType` |

2. **`ReportDataHeader.aggregateFunction`** — new field, propagated alongside the type. Type alone is not enough: `MAX(int)`, `COUNT_DISTINCT`, `ANY_VALUE` all yield INTEGER but must behave differently in Looker.

3. **`LookerStudioAggregationMapperService.mapAggregateFunctionToLookerType(aggFunc, dataType)`** — exhaustive switch:

   | backend `AggregateFunction` | `defaultAggregationType` | `isReaggregatable` | conceptType |
   |---|---|---|---|
   | `SUM` | `SUM` | `true` | METRIC if NUMBER, otherwise DIMENSION |
   | `COUNT` | `SUM` | `true` | METRIC |
   | `COUNT_DISTINCT` | — | `false` | METRIC |
   | `MIN` / `MAX` | `MIN` / `MAX` | `true` | METRIC if NUMBER, otherwise DIMENSION |
   | `STRING_AGG` / `ANY_VALUE` | — | — | DIMENSION |

   ⃰ `COUNT → SUM` because summing counts across non-overlapping groups is correct. `COUNT_DISTINCT → reagg=false` because `SUM(COUNT_DISTINCT_per_group) ≠ COUNT_DISTINCT_overall`.

4. **`LookerStudioTypeMapperService.buildSchemaField(header, storageType)`** — single source of truth shared by `getSchemaFields` and `buildResponseSchema`. Guarantees `getSchema.dataType === getData.schema.dataType` (locked in by an e2e contract test).

5. **`StorageFieldType` union alias** — deduplicates the 5-member union previously copy-pasted in `report-data-header.dto.ts`, `sheet-values-formatter.ts`, and the `mapToLookerStudioDataType` signature.

Native (no `aggregateFunction`) behaviour is **unchanged**: `NUMBER → METRIC + SUM + reagg=true`, otherwise DIMENSION.

### DIFF
--- a/.changeset/looker-studio-blended-types-aggregations.md
+++ b/.changeset/looker-studio-blended-types-aggregations.md
@@ -1,0 +1,15 @@
+---
+'owox': minor
+---
+
+# Correct column types and aggregations for blended fields in Looker Studio
+
+When a Data Mart pulls in fields from joined Data Marts (blended fields), the Looker Studio connector now reports the right column types and respects the aggregation function you selected for each field.
+
+- `COUNT` and `COUNT_DISTINCT` blended fields appear as numeric metrics — no more text dimensions for counts.
+- `STRING_AGG` and `ANY_VALUE` blended fields appear as dimensions, so Looker Studio doesn't try to sum already-aggregated values.
+- `MIN` / `MAX` / `SUM` blended fields use the matching default aggregation in Looker Studio (`MIN`/`MAX`/`SUM`), instead of always defaulting to `SUM`.
+- `COUNT_DISTINCT` is no longer marked as re-aggregatable, preventing inflated totals when the field is used across filters or breakdowns.
+- The fix works consistently across BigQuery, Snowflake, Redshift, Athena, and Databricks.
+
+Native (non-blended) numeric fields continue to behave as before — they default to `SUM` and remain re-aggregatable.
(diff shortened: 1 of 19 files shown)


max-voloshyn:
**GPT 5.5**: No blocking issues found.

max-voloshyn (approved):

PR #1189, opened 2026-05-06
feat: output controls per report (filters, sort, limit)

zapolsky:
Reports now carry their own filter / sort / limit configuration, applied
on the final SELECT before writing to the destination. The underlying
data mart definition is not touched — output controls live on the Report
entity and re-apply on every run.

Backend (apps/backend)

- New `Report` columns: `filterConfig` (JSON, FilterRule[]),
  `sortConfig` (JSON, SortRule[]), `limitConfig` (int, nullable).
  Migration is idempotent.
- Zod schemas: `FilterConfigSchema` is a discriminated union over
  operator, covering 19 operators (scalar comparators, no-value ops,
  `between`, `relative_date` presets like Last N days). `SortConfigSchema`
  enforces `direction: 'asc' | 'desc'`.
- `OutputControlsCapabilityService` gates support per-storage —
  Google BigQuery is enabled, Athena/Redshift/Snowflake/Databricks
  reject output-control configurations with a clear capability error
  at the use-case layer.
- `OutputControlsValidatorService.validateForReport()` orchestrates
  capability + filter (column known, operator allowed for type) +
  sort (column selected) checks; surfaces structured errors through
  `BadRequestException.details.errors`.
- New `SqlClauseRenderer` abstract base + `BigQueryClauseRenderer`
  render WHERE / ORDER BY / LIMIT. Filter values flow through the
  BigQuery client as named parameters (`@p0`, `@p1`, …) — no string
  interpolation, no SQL injection surface.
- `BigQueryQueryBuilder` and the blended query builder both append
  output controls when present; non-BigQuery adapters short-circuit
  with `NotImplementedException` only when output controls are
  actually requested (legacy `{limit: 0}` schema-fetch path is
  preserved).
- `ReportSqlComposerService` composes the SQL+params bundle for
  non-blended reports with output controls and resolves
  `mainTableReference` for SQL-defined data marts.
- `BlendedReportDataService` now treats filter-only columns as part
  of the chain reference set (`columnConfig ∪ filterColumns`) and
  marks them `isHidden` so they flow through CTEs without leaking
  into the final SELECT — supports filtering by a column you didn't
  select for output.
- `RunReportService` threads `sqlOverride` + `sqlOverrideParams`
  through the reader interface; `BigQueryReportReaderService`
  forwards the params to `adapter.executeQuery(sql, params)`.
- API surface: `Update/CreateReportRequestApiDto` accept
  `filterConfig?`, `sortConfig?`, `limitConfig?` with class-validator
  bounds (`ArrayMaxSize 50/10`, `IsInt @ispositive @max 10_000_000`).
  `ReportMapper` and `ReportResponseApiDto` round-trip the new fields.
  Cache invalidation triggers when any of the four configs change.
- E2E coverage in `output-controls.e2e-spec.ts`: CRUD with output
  controls, validator rejections, class-validator boundaries.

Web (apps/web)

- New types in `features/data-marts/shared/types/output-config.ts`
  (FilterRule discriminated union, SortRule, OutputConfig, helpers).
- `output-controls-support.ts` — frontend capability check mirroring
  backend.
- New components under `edit/components/ReportColumnPicker/`:
  `OutputSettingsButton`, `OutputSettingsDropdown`,
  `FilterEditorPopover`, `FilterRow`, `SortRow`, `LimitInput`,
  `RowFilterIcon`, `output-controls-operators` (operator catalog
  per field type with human labels). Native HTML5 drag-and-drop
  reorders sort rules; multi-filter on the same column allowed
  (rendered with per-rule remove); row-level filter icon opens a
  popup that lists all rules for that column with quick add /
  remove.
- `ReportColumnPicker` integrates output controls behind a
  capability fallback (a clear "not supported on this storage"
  banner with a Clear shortcut for non-BigQuery storages).
- `useGoogleSheetsReportForm`, `useLookerStudioReportForm`,
  `useEmailReportForm` extended with `filterConfig`/`sortConfig`/
  `limitConfig`; reset() in the corresponding edit forms now
  re-hydrates the new fields from `initialReport` so saved values
  show up on edit.
- API client request types updated to carry the new fields end
  to end.

Visual polish (web)

- Output controls panel shares the column list's scroll container
  so heavy filter/sort/limit configurations don't push the columns
  out of view.
- `LimitInput`: native number-spinner buttons removed; placeholder
  reflects "no limit" semantics; row controls aligned vertically
  with the rest of the panel.
- `FilterRow`: two-line layout (column · condition + value);
  condition rendered with semantic typography rather than a color
  tag; pencil edit icon visible only on row hover or when its
  popup is open.
- `SortRow`: drag handle, lucide arrow icons, ghost direction
  toggle aligned with the rest of the iconography; fixed-width
  direction label so arrows line up across rows.
- All action icons unified to `h-4 w-4` in `h-6 w-6` containers,
  matching the icon size in DATA MART / REPORT header sections.
- Add filter / Add sort by triggers expose the chevron on the
  right (instead of next to the placeholder).
- Numeric operators rendered as words (`equals`, `greater than`,
  …) instead of `=`/`>`/`≥`.

Notes

- Web tests: 101/101 passing. Backend e2e green.


laskevych (approved):
Great solution 👍

PR #1193, opened 2026-05-07
feat: allow report runs and trigger CRUD for users who see the Data Mart

zapolsky:
Manual report runs and REPORT_RUN trigger CRUD now require canOperate(report) = canSee(DataMart) AND canUse(Destination) instead of Report Ownership. Project members who see the DM (Business Owner, Tech Owner, or non-owner via shared-for-reporting / for-maintenance / for-both) can run any report and manage its triggers regardless of report ownership. Cross-owner trigger management is an explicit trade-off per the DoD.

Editing report config (columns, filters, owners, destination) remains on the unchanged canMutate path: Owner with effective Destination, DM maintenance bypass for Tech Users, or Admin.

Backend
- ReportAccessService gains canOperate / checkOperateAccess; isEffective is now per-user (canAccess USE on Destination), aligning the code with Stage 2 §76 of the permissions spec — owners who lose Destination USE become ineffective.
- run-report and create / update / delete-scheduled-trigger services route REPORT_RUN through checkOperateAccess and surface reason-keyed errors from OPERATE_DENIED_MESSAGES (not-found / dm-invisible / destination- unusable) instead of one-size-fits-all 403.
- scheduled-trigger.controller drops DM MANAGE_TRIGGERS for REPORT_RUN create; update / delete keep coarse SEE and let the service apply the type-specific gate. CONNECTOR_RUN is unchanged.
- ReportResponseApiDto exposes canRun / canManageTriggers / canEditConfig, computed in get / list / create / update services so frontends can gate UI without recomputing the access matrix.
- ListReportsByInsightTemplateCommand extended with userId / roles to enable capability computation there too.
- UpdateReportService now validates caller's USE on a newly-supplied dataDestinationId, fixing a parity gap with CreateReportService.

Web
- DataMartReport and the FE response DTO carry the same three flags. UI gating is intentionally deferred to a follow-up.

Tests
- 27 unit cases for the canOperate / checkOperateAccess / per-user isEffective matrix in report-access.service.spec.
- 57 e2e cases in report-operate-access.e2e-spec covering DoD end-to-end: Manual Run (14, full role x sharing matrix + ineffective owner), REPORT_RUN trigger CRUD (16, including cross-owner ops, cross-DM attacks on create and update, CONNECTOR_RUN regression sanity), config protection (15, including DM maintenance bypass and admin bypass), and capability flags on GET / list / POST / PUT (12).

### DIFF
--- a/apps/backend/src/data-marts/controllers/report.controller.ts
+++ b/apps/backend/src/data-marts/controllers/report.controller.ts
@@ -38,7 +38,6 @@ import {
   GetReportGeneratedSqlSpec,
   CopyReportAsDataMartSpec,
 } from './spec/report.api';
-import { RunType } from '../../common/scheduler/shared/types';
 import { OwnerFilter } from '../enums/owner-filter.enum';
 
 @Controller('reports')
@@ -142,7 +141,7 @@ export class ReportController {
     @AuthContext() context: AuthorizationContext,
     @Param('id') id: string
   ): Promise<void> {
-    const command = this.mapper.toRunReportCommand(id, context, RunType.manual);
+    const command = this.mapper.toManualRunReportCommand(id, context);
     await this.runReportService.run(command);
   }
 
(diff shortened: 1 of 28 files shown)


laskevych (approved):
Good job 👏

PR #1196, opened 2026-05-08
feat(web): promo bulk data mart creation

artemfenoga:
# Discover BigQuery Bulk Import When You Need It

## Contextual BigQuery Promo

When you have exactly one BigQuery Data Mart, you'll now see a helpful prompt suggesting bulk import from your existing BigQuery tables and views, plus a direct manual setup path.

Frame 23137585

## Clearer Storage Health UX in Resource Pickers

If a selected storage is not fully configured, resource browsers now show a consistent health-state block instead of raw loading errors, so it's obvious why tables or views cannot be listed yet.

Frame 23137584

## Unified Behavior Across Bulk and Single Flows

- The same unhealthy-storage UI is now reused in both Bulk Create and single Data Mart table selection flows, giving users a predictable experience.
- The table picker dialog now uses a cleaner, card-based content area that improves readability and keeps the error/health state visually structured.

Frame 23137586

### DIFF
--- a/.changeset/promo-block-bigquery-bulk-import.md
+++ b/.changeset/promo-block-bigquery-bulk-import.md
@@ -1,0 +1,13 @@
+---
+'owox': minor
+---
+
+# Discover BigQuery Bulk Import When You Need It
+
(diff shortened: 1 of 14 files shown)


laskevych (approved):
Amazing 🤩

PR #1214, opened 2026-05-15
fix(backend): verify credential project ownership on destination create

laskevych:
## Summary

`POST /api/data-destinations` accepted a caller-supplied `credentialId` and linked
it to a new destination after only verifying that Google returned an access token.
Two checks that the rest of the credential surface already enforced were missing
on this code path, so a caller could link a credential they should not be able to
use.

This PR closes both gaps and mirrors the existing checks from
`UpdateDataDestinationService` and the `sourceDestinationId` copy flow:

- **Project ownership.** Reject the request if the credential does not belong to
  the authenticated project (`403 Forbidden — Credential does not belong to this
  project`). Same wording and behavior as the destination update path.
- **Copy permission.** If the supplied credential is already linked to another
  destination in the project, require `Action.COPY_CREDENTIALS` on that existing
  destination via `AccessDecisionService.canAccess(...)`. Same check the
  `sourceDestinationId` branch already performs.

The COPY_CREDENTIALS gate fires only when the credential is already in use, so
the standard "complete OAuth flow → create destination with the returned
`credentialId`" UX path is unaffected (the credential is fresh, no existing
destination references it).

## What changed

- `apps/backend/src/data-marts/use-cases/create-data-destination.service.ts`:
  add ownership check + conditional `COPY_CREDENTIALS` check, before the OAuth
  token verification so cross-project credential UUIDs are not even probed
  against Google.
- `apps/backend/src/data-marts/use-cases/create-data-destination.service.spec.ts`:
  5 new Jest cases (cross-project reject, missing credential reject, same-project
  not-yet-linked accept, reused-without-permission reject, reused-with-permission
  accept).
- `.changeset/improved-security-for-destination-copying.md`: user-facing note
  describing both layers.

## Test plan

- [x] `npx jest src/data-marts/use-cases/create-data-destination.service.spec.ts`
      — 8/8 pass (3 existing + 5 new).
- [x] `npx jest src/data-marts/use-cases/update-data-destination.service.spec.ts`
      — 9/9 pass, no regression in the symmetric path.
- [ ] Manual smoke against a dev instance:
  - `POST /api/data-destinations` with a cross-project `credentialId` → expect
    `403 Forbidden — Credential does not belong to this project`.
  - Same-project `credentialId` from a fresh OAuth flow (not yet linked) →
    expect `201`.
  - Same-project `credentialId` already linked to another destination, caller
    lacks copy access → expect `403 Forbidden — You do not have permission to
    copy credentials from this destination`.
  - Same scenario, caller has copy access (e.g., destination owner or
    shared-for-maintenance) → expect `201`.

### DIFF
--- a/.changeset/improved-security-for-destination-copying.md
+++ b/.changeset/improved-security-for-destination-copying.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Stricter project ownership check when linking an existing OAuth credential to a new destination
+
+Creating a Google Sheets destination with a pre-existing `credentialId` now requires that credential to belong to the same project as the destination being created. Requests that reference a credential from a different project are rejected with `403 Forbidden — Credential does not belong to this project`.
--- a/.changeset/improved-security-for-destination-copying.md
+++ b/.changeset/improved-security-for-destination-copying.md
@@ -2,6 +2,11 @@
 'owox': minor
 ---
 
-# Stricter project ownership check when linking an existing OAuth credential to a new destination
+# Stricter project ownership and copy-permission checks when linking an existing OAuth credential to a new destination
 
-Creating a Google Sheets destination with a pre-existing `credentialId` now requires that credential to belong to the same project as the destination being created. Requests that reference a credential from a different project are rejected with `403 Forbidden — Credential does not belong to this project`.
+Creating a Google Sheets destination with a pre-existing `credentialId` is now protected by the same two checks that already guarded credential copying from a source destination.
+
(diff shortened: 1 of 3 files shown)


max-voloshyn:
Reviewed with Codex using GPT-5.5.

Looks good from the source-code review. The `credentialId` create path now validates project ownership and enforces `COPY_CREDENTIALS` when the credential is already linked to an existing destination, with regression tests covering the report-scope cases.

max-voloshyn (approved):

PR #1224, opened 2026-05-19
Emoji picker to data mart title

ikrasovytskyi:
Adds a quick-access emoji picker button next to the data mart title, both on the edit page and the create page. Selecting an emoji prepends it with a space to the current title, removing the need to copy-paste emojis manually.

## Changed files

- **New:** `EmojiPickerButton.tsx` — popover component with emoji grid
- `InlineEditTitle.tsx` — extended context to expose `value` alongside `setValue`
- `DataMartDetails.tsx` — wired emoji picker into the title area
- `DataMartCreateForm.tsx` — added emoji picker next to the title input

emoji1
emoji2

## Test plan
- [ ] Open a data mart, focus the title — emoji picker icon appears
- [ ] Click an emoji — it prepends to the title
- [ ] Create a new data mart — emoji picker appears next to the title input
- [ ] AI title button still works alongside the emoji picker

@ikrasovytskyi

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/DataMartCreateForm.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartCreateForm.tsx
@@ -15,6 +15,7 @@ import { Input } from '@owox/ui/components/input';
 import { Plus } from 'lucide-react';
 import { useEffect, useMemo, useRef, useState } from 'react';
 import { EmojiPickerButton } from './EmojiPickerButton';
+import { prependEmoji } from './emoji-picker.constants';
 import { useForm } from 'react-hook-form';
 import { Combobox } from '../../../../shared/components/Combobox/combobox';
 import { DataStorageHealthIndicator, DataStorageType } from '../../../data-storage';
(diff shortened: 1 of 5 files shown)


Obolrus:
Hi there,

Thank you so much for putting this PR together and for your proactive contribution to making OWOX Data Marts better 💪😎 

I completely agree with your reasoning—utilizing **visual anchoring** through emojis is a fantastic way to significantly reduce **cognitive load** and help users scan and identify entities much faster. 

However, we believe there is a more systemic and scalable way to implement this feature. Instead of allowing users to manually append emojis directly into the title text (as proposed in this PR), we want to introduce a dedicated emoji/icon picker field for all entities across ODM. This will keep the data structured and open up better UI/UX possibilities in the future. 

We really value this insight, so we have officially moved this feature idea into our product backlog to design and implement it properly on our end. 

As a side note/workaround for right now: users can already manually insert any emoji into entity titles simply by copying them from the clipboard or using the native system OS emoji picker (for instance, using the `Control + Command + Space` shortcut on macOS).

Thank you once again for your effort, time, and care for the product's usability. We truly appreciate your support!

Happy analyzing!

PR #1227, opened 2026-05-20
feat: slices — pre-join filters for joined data marts (#6537)

zapolsky:
Add a new slicesConfig on Report alongside the existing filterConfig/sortConfig/limitConfig. A slice is a filter applied INSIDE the joined data mart's *_raw CTE (before the JOIN), so large subsidiary marts (e.g. users with 10y of rows joined into orders) are narrowed upstream instead of dragging every row through the join.

## Backend (apps/backend)

- New `SliceRuleSchema` / `SlicesConfigSchema` (Zod) with `aliasPathToCteName` helper. Nested `{ aliasPath, rule: FilterRule }` shape so FilterRule narrowing stays intact for consumers.
- New `slicesConfig` JSON column on `report` (migration + entity + DTOs + mapper + use-case wiring + cache invalidation).
- SQL codegen:
  - `SqlClauseRenderer.renderWhere` gains a `paramPrefix` arg so per-CTE slice WHEREs don't collide on `@p0` with the final post-join WHERE.
  - `AbstractBlendedQueryBuilder.buildRawCte` injects WHERE inside each subsidiary `*_raw` CTE; `collectSubsidiaryReferences` projects slice columns alongside join keys.
  - `BlendedReportDataService.buildRelationshipChains` pulls chains referenced ONLY by slices (chain build was previously columnConfig-only — slices would silently no-op without this).
- Capability + validation: `SLICES_REQUIRE_JOINED_DATA_MART`, `SLICE_ALIAS_PATH_UNKNOWN`, `SLICE_ALIAS_PATH_NOT_ALLOWED_ON_HOME`, `SLICE_COLUMN_UNKNOWN`, `SLICE_OPERATOR_UNSUPPORTED_FOR_TYPE`, `SLICE_INVALID_REGEX_PATTERN`, plus existing `OUTPUT_CONTROLS_NOT_SUPPORTED` for non-BigQuery storages.

## Web UI (apps/web)

- Mirror `SliceRule` / `SlicesConfig` types; form schema + round-trip.
- Extract reusable `FilterValueEditor` from `FilterEditorPopover`.
- New `SliceEditorPopover` (one-column edit) and `FilterOrSliceEditorPopover` (tabbed wrapper with parallel drafts and aliasPath injection).
- New SLICES section in `OutputSettingsDropdown` between FILTERS and SORT, with `AddSlicePicker` grouping joined columns by `aliasPath`.
- Column-row UX:
  - Single funnel icon highlights for both filters and slices (no separate Layers indicator); counter sums both; `defaultTab='slice'` when only slices are active.
  - Click on a column with EXACTLY one active rule opens the popover pre-filled in edit mode (Apply replaces instead of appending). New `handleReplaceFilterAt` / `handleReplaceSliceAt` threaded through NativeFieldRow / BlendedFieldRow / BlendedGroupItem.
- Info tooltips on FILTERS / SLICES / SORT / LIMIT section headers explaining each section's apply order.

## Trade-offs (agreed at portfolio level)

- BigQuery only for v1. Other storages light up automatically once they grow a `SqlClauseRenderer` subclass (same capability gate as today).
- Slices operate on subsidiary dimension columns only — home-mart aggregations are NOT rewritten to account for slice semantics. Slices on the home mart (aliasPath='main') are rejected explicitly.
- `relative_date` relative to another field is out for v1 (operator variant + renderer impl when product needs it).

## Tests (~320 new tests, all green)

Backend:
- Zod schema + helper unit tests.
- BigQueryBlendedQueryBuilder: full-flow combo (filters + slices + sort + limit), edge cases (SQL safety, tree shapes), slice operator matrix (all 20+ FilterRule operators inside raw CTE WHERE), paramPrefix matrix (multi-slice numbering, between/relative_date/no-value param counts, cross-CTE uniqueness), determinism (same context → byte-identical SQL, dedup, undefined === empty, stable column order), simple-BQ edges.
- Validator service: validateSlices matrix.
- Composer service: SLICES_REQUIRE_JOINED_DATA_MART + capability gate.
- E2E `output-controls.e2e-spec.ts`: persistence + validation matrix.
- New E2E `blended-output-controls.e2e-spec.ts` (47 tests) via real REST → composer → /generated-sql pipeline. Uses real `PUT /api/data-marts/:id/schema` to seed schemas (no `BlendableSchemaService` mock).

Web:
- vitest+RTL: RowFilterIcon (edit-on-single + tab routing), FilterOr- SliceEditorPopover (parallel tab drafts + aliasPath injection), OutputSettingsDropdown SLICES section (round-trip + collision-safe `\0` separator), useGoogleSheetsReportForm round-trip, FilterValueEditor onChange semantics, SliceRow + SliceEditorPopover render/edit, SectionHeader info tooltips.

### DIFF
--- a/.changeset/slices-pre-join-filters.md
+++ b/.changeset/slices-pre-join-filters.md
@@ -4,28 +4,7 @@
 
 # Slices: filter joined data marts before they are joined
 
-Reports on joined data marts gain a new `slicesConfig` field. A **slice** is a filter
-applied **before** the JOIN (inside the subsidiary's raw CTE) — so large subsidiary
-tables (e.g. `users` with 10 years of data joined into `orders`, `reportRuns` over a
-long window) can be narrowed to the rows you actually need before the join, instead
-of filtering the full join product after the fact.
-
-**UI**: on a joined column, the filter popover now offers a `Filter | Slice` tab
-toggle, and a new SLICES section appears in the Report editor alongside Filters /
-Sort / Limit. Clicking the filter icon on a column with exactly one active rule
-opens the popover pre-filled in edit mode (Apply replaces instead of appending).
-Each section header also gained an info tooltip describing what the section does
-and when filters vs. slices apply.
-
-**Backend**: SQL codegen extended in `AbstractBlendedQueryBuilder` to inject a
-per-CTE WHERE inside each joined data mart's `*_raw` CTE using the existing
-`SqlClauseRenderer` (now parameter-prefix-aware to avoid `@p0` collisions). Two
-upstream tweaks: `BlendedReportDataService.buildRelationshipChains` pulls chains
-referenced only by slices, and `collectSubsidiaryReferences` projects slice
-columns into the raw CTE.
-
-**Scope**: BigQuery only for v1 (same capability gate as filter/sort/limit; other
-storages light up automatically once they grow a `SqlClauseRenderer` subclass).
-Slices on the home data mart, slices on simple (non-joined) data marts, and
-`relative_date_to_field` (date relative to another field) are out for v1 and
-return clear `SLICES_REQUIRE_JOINED_DATA_MART` / `SLICE_*` errors when attempted.
+Reports on joined data marts gain pre-join filters ("slices") — a filter applied INSIDE the joined data mart's `*_raw` CTE before the JOIN,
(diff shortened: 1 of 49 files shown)


kalnik-a-a (approved):
Great work 👍🏻

PR #1232, opened 2026-05-22
fix(google-sheets): prevent insertDimension grid-size error on column refresh

prokopetsv:
# Problem

When a report refresh both removed columns and added new ones, with the number of removals greater than or equal to additions, the refresh failed with:

```
Invalid requests[N].insertDimension: range.startIndex must be less than the grid size
if inheritFromBefore is false
```

The root cause: the structural ops batch applies deletes before inserts. When the sheet grid has no user columns to the right of the imported range, those deletes shrink the grid to exactly `survivors.length`. The subsequent `insertDimension` at `startIndex = survivors.length` then violates the Sheets API rule that `startIndex` must be strictly less than grid size when `inheritFromBefore` is false.

# Solution

Pre-allocate `deleteCount` extra columns before the batch so the grid remains wide enough for inserts after the deletes complete. The extra columns land past `finalCount` once the batch runs; they are removed by appending matching `deleteDimension` requests to the end of the same batch, keeping the grid at exactly `finalCount` and leaving no trailing empty columns.

- `applyDeferredSheetMutations` records the column count before pre-allocation and computes `extraAllocated` — the number of columns added beyond what `finalCount` requires — then passes it to `applyStructuralColumnOps`
- `applyStructuralColumnOps` appends `extraAllocated` cleanup delete requests (descending) after all inserts in the batch, and accounts for them in `availableColumnsCount` tracking

# Changes

- In `applyDeferredSheetMutations`: pre-allocate `finalCount + deleteCount` columns when inserts are present; compute `extraAllocated` from the difference between the expanded grid and `max(prevCount, finalCount)`
- In `applyStructuralColumnOps`: added `extraAllocated` parameter; appends cleanup `deleteDimension` requests at `[finalCount .. finalCount + extraAllocated - 1]` descending; adjusts `availableColumnsCount` by `inserts - deletes - extraAllocated`
- Updated comment to precisely state the failure condition: only triggers when shrinkage lands at exactly `survivors.length`, not in all delete+insert cases

### DIFF
--- a/apps/backend/src/data-marts/data-destination-types/google-sheets/services/google-sheets-report-writer.ts
+++ b/apps/backend/src/data-marts/data-destination-types/google-sheets/services/google-sheets-report-writer.ts
@@ -208,7 +208,16 @@ export class GoogleSheetsReportWriter implements DataDestinationReportWriter {
     if (this.structuralOpsApplied) {
       return;
     }
-    await this.prepareSheetColumns(this.columnPlan.finalImportedNames.length);
+    // When the batch contains both deletes and inserts, the deletes run first
+    // and temporarily shrink the grid to `survivors.length`. The subsequent
+    // insertDimension (inheritFromBefore: false) requires startIndex < gridSize,
+    // so we must pre-allocate `deleteCount` extra columns to keep room for the
+    // inserts after the deletes have been applied.
+    const deleteCount = this.columnPlan.ops.filter(op => op.kind === 'delete').length;
+    const insertCount = this.columnPlan.ops.filter(op => op.kind === 'insert').length;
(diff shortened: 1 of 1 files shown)


laskevych:
Closing in favor of a different approach. The current implementation has a data-loss edge case: when the user has content past the imported range and insertCount > userColCount, the cleanup deletes target user columns instead of pre-allocated empties (concrete trace in the linked task).

Re-implementing with a minimal sentinel approach in #XXXX — pre-allocate exactly 1 column only when availableColumnsCount === prevWidth, cleanup from the rightmost column. Preserves the ColumnPlan "user content right of imported range is left untouched" invariant.

PR #1233, opened 2026-05-22
refactor(backend): pre-clear Google Sheets imported rectangle before writes

laskevych:
## Summary

Replaces the post-write truncation + null-normalization pair from PR #1229
with a single up-front `values.clear` of the imported rectangle. The user-
visible contract is unchanged — manual edits inside the imported rectangle
never survive a refresh, user content right of it does, failed refresh
leaves the sheet untouched — but the mechanism is unified into one
intent-driven step instead of two coupled ones.

## Motivation

PR #1229 fixed a customer-blocking regression where manual edits in cells
that SQL produced NULL for would survive across refreshes. The fix
normalized `null` / `undefined` / sparse-array holes to `""` in the
payload sent to `spreadsheets.values.update`, because Sheets treats those
three as "skip this cell" — silently preserving stale or manually-edited
content.

The normalize fix worked, but had two architectural costs:

1. **Footgun.** Future contributors needed to know that `values.update`
   has asymmetric null-handling, and that the formatter quietly
   normalizes nullish to keep the contract. The bug that triggered #1229
   was missed in the original column-preservation PR (#1191) precisely
   because this Sheets API quirk is undocumented in the way the writer
   uses it.
2. **Two coupled mechanisms** for "ODM owns the imported rectangle":
   nullish-normalization for cells SQL overwrites, plus
   `truncateTrailingImportedRows` for cells past the new data. A single
   primitive could do both jobs.

## Approach

`applyDeferredSheetMutations` (the gated, one-shot block that runs at the
first successful batch) now ends with a `values.clear` over
`A2..{lastImportedColA1}{availableRowsCount}` — the entire imported
rectangle below the headers. Subsequent `writeReportDataBatch` calls just
write data into an already-empty region:

- Cells SQL fills with a value → overwritten with that value.
- Cells SQL produces `null` for → land on an already-empty cell (Sheets'
  "skip this cell" leaves the cleared cell empty — the desired outcome).
- Cells past the last written row → stay empty from the pre-clear (no
  need for a separate post-write truncate).

The failed-refresh contract is preserved by the same deferred-mutations
gate that already covered structural ops + headers: nothing touches the
sheet until the reader returns at least one successful batch.

## Before / After

```
BEFORE (post-#1229)                 AFTER (this PR)

applyDeferredSheetMutations         applyDeferredSheetMutations
  ├─ prepareSheetColumns              ├─ prepareSheetColumns
  ├─ applyStructuralColumnOps         ├─ applyStructuralColumnOps
  └─ writeHeaders                     ├─ writeHeaders
                                      └─ preClearImportedRectangle  ◄── new
writeReportDataBatch × N
  ├─ formatter normalizes nullish    writeReportDataBatch × N
  │   to "" in the payload  ◄── gone   └─ writes payload as-is
  └─ writes payload
                                     finalize
finalize                              ├─ metadata + fill-down
  ├─ truncateTrailingImportedRows   ◄── gone, covered by pre-clear
  └─ metadata + fill-down
```

## Architectural trade-offs (considered, documented)

- **+1 API call in the common case.** Old truncation was a no-op when
  `writtenRows >= availableRows`, so a steady-state refresh used N
  `values.update` calls. New pre-clear runs unconditionally — so N+1.
  When the dataset shrinks, both old and new use N+1 (same cost). The
  +1 is the price of removing the asymmetric-null footgun.
- **Mid-stream failure (reader fails after batch K of N).** Old: rows
  written so far hold new data, rows below hold stale data from previous
  refresh. New: rows written so far hold new data, rows below are empty
  (from pre-clear). Both are partial states; "empty below" is preferable
  for analytics consumers where stale numbers can lead to bad decisions
  without a visible signal.
- **Race window between pre-clear and first data write.** ~100 ms
  (vs. ~50 ms in the old architecture). User typing during this window
  could see a flicker; same fundamental Sheets-API limitation as before,
  marginally widened, considered acceptable.

### Modified
- `apps/backend/src/data-marts/data-destination-types/google-sheets/services/google-sheets-report-writer.ts`
  — added `preClearImportedRectangle`, wired into
  `applyDeferredSheetMutations`; removed `truncateTrailingImportedRows`
  and the `finalize` call site; updated JSDoc on
  `applyDeferredSheetMutations`, `writeReportDataBatch`, `finalize`.
- `apps/backend/src/data-marts/data-destination-types/google-sheets/services/sheet-formatters/sheet-values-formatter.ts`
  — reverted `formatRowsValuesByName` to pre-#1229 behavior (no nullish
  normalization). JSDoc explains why pre-clear makes the normalization
  unnecessary.
- `apps/backend/src/data-marts/data-destination-types/google-sheets/adapters/google-sheets-api.adapter.ts`
  — `clearValuesInRange` JSDoc made generic (no longer locked to the
  writer's specific use case).

### Tests
- `apps/backend/src/data-marts/data-destination-types/google-sheets/services/google-sheets-report-writer.spec.ts`
  — rewritten. 11 tests across two `describe` blocks:
  - **Pre-clear pipeline (5):** first batch triggers pre-clear; subsequent
    batches do not re-clear; reader-failed-before-batch → no pre-clear;
    `prepareToWriteReport` failure handling; zero-batch fallback.
  - **Pre-clear range invariants (6):** clear starts at row 2 (headers
    preserved); clear ends at last imported column (user content right
    untouched); A1 notation correct across the Z→AA boundary; pre-clear
    runs before any data write (sequence); writer passes nullish through
    unchanged (locks the no-normalize decision); pre-clear failure aborts
    the batch.
- `apps/backend/src/data-marts/data-destination-types/google-sheets/services/sheet-formatters/sheet-values-formatter.spec.ts`
  — deleted (it tested the normalize behavior introduced in #1229,
  which no longer exists).
- `apps/backend/test/integration/google-sheets-column-preservation.integration.ts`
  — `provisionFixture` now also returns `dataDestinationId`; the
  `limitConfig` shrink test sends a minimal PUT payload instead of
  round-tripping the full report through GET → PUT (the latter triggered
  a 400 from the global `forbidNonWhitelisted` ValidationPipe with
  certain null-shaped optional fields).

## Tests

- Unit: **83/83** passing across the google-sheets module.
- ESLint + Prettier + `tsc --noEmit`: clean on all touched files.
- Integration (real Google Sheets + BigQuery, gated on CI by
  `GOOGLE_SHEETS_SERVICE_ACCOUNT_JSON` / `TEST_GOOGLE_SPREADSHEET_ID` /
  `BQ_*` secrets): **12/12** in
  `google-sheets-column-preservation.integration.ts`. Every existing case
  passes against the new architecture, which is the strongest evidence
  that the user-facing contract is unchanged:
  - first run + OWOX_COLUMNS metadata persistence
  - user formulas right of imported range survive
  - user-driven column reorder wins
  - new SQL column appended at right edge
  - removed SQL column → `#REF!`
  - alias-aware mapping
  - auto fill-down
  - static values right of imported range not overwritten
  - column picker (`columnConfig`)
  - failed refresh leaves sheet untouched
  - **manual edits in NULL cells overwritten** (the #1229 contract)
  - **stale rows cleared when limitConfig shrinks** (the #1198 contract)

## Documentation

User-facing docs at
`docs/destinations/supported-destinations/google-sheets.md` updated to
state explicitly that the imported rectangle is fully overwritten on
every refresh — including manual edits placed inside it. Previously this
was implied ("OWOX owns only the columns it writes"); now it is called
out as its own bullet so users understand the contract before they put
notes or formulas inside the imported columns.

## Risks

- **Athena / Snowflake.** Pre-clear is storage-agnostic — it acts on the
  destination sheet, not on warehouse output. The integration suite still
  exercises BigQuery only; if a future adapter ever delivered SQL NULL
  as a non-`null`/`undefined` JavaScript value (e.g. `"null"` string),
  the writer would now write that string into the cell — same behavior
  as #1229's normalize approach.
- **API quota.** +1 `values.clear` per refresh. Negligible for typical
  daily-refresh customers; worth monitoring if a customer schedules many
  reports per minute.


kalnik-a-a (approved):
LGTM👍🏻

PR #1236, opened 2026-05-22
fix(connectors): auto-evolve Snowflake and Redshift table schema on field change

prokopetsv:
# Problem

Adding new fields to an existing Snowflake or Redshift data mart caused MERGE and INSERT failures with errors like `invalid identifier '"account_id"'` (Snowflake) or `column "account_id" does not exist` (Redshift). Both storage layers called `CREATE TABLE IF NOT EXISTS` which silently no-ops when the table already exists, so the schema check never detected that new columns were missing and `ALTER TABLE` was never executed.

The Snowflake fix was further blocked by two INFORMATION_SCHEMA query bugs: the `TABLE_CATALOG` filter used a lowercase database name (Snowflake stores it uppercase, returning 0 rows), and config values for schema and table already contained surrounding double-quotes that made `WHERE TABLE_SCHEMA = UPPER('"PUBLIC"')` match nothing. Redshift had no `getAListOfExistingColumns()` at all — schema state was rebuilt from the connector config on each run rather than queried from the live table.

Additionally, both `addNewColumns()` implementations tried to issue a single `ALTER TABLE … ADD COLUMN c1, ADD COLUMN c2` statement, which is not valid syntax in either Snowflake or Redshift.

# Solution

Both storage layers now query the live table schema on startup (`init()`) and derive which columns are new by comparing the result against the configured fields. `ALTER TABLE … ADD COLUMN` is issued once per new column in a loop. If the table does not exist yet, `createTable()` is called as before.

Key changes:
- **Snowflake** — fixed `getAListOfExistingColumns()`: removed `TABLE_CATALOG` filter, removed `UPPER()` from `TABLE_NAME`, and stripped surrounding double-quotes from config values before interpolating into `INFORMATION_SCHEMA` queries.
- **Redshift** — added `loadTableSchema()` called from `init()`, added `getAListOfExistingColumns()` querying `information_schema.columns` with lowercase normalization, and added `executeQueryWithResults()` to fetch rows via the Redshift Data API (async execute → wait → paginated fetch).
- **Both** — replaced batched `ALTER TABLE … ADD COLUMN c1, c2` with one statement per column in a `for` loop.

# Changes

- Fixed `SnowflakeStorage.getAListOfExistingColumns()` to strip surrounding double-quotes from schema/table config values and removed the broken `TABLE_CATALOG` and `UPPER(TABLE_NAME)` filters
- Fixed `SnowflakeStorage.addNewColumns()` to issue one `ALTER TABLE … ADD COLUMN` statement per column instead of a single batched statement
- Added `AwsRedshiftStorage.loadTableSchema()` called from `init()` to detect and apply schema changes at startup
- Added `AwsRedshiftStorage.getAListOfExistingColumns()` querying `information_schema.columns` with quote-stripping and lowercase normalization
- Added `AwsRedshiftStorage.executeQueryWithResults()` for SELECT queries via the Redshift Data API with pagination support
- Fixed `AwsRedshiftStorage.addNewColumns()` to issue one `ALTER TABLE … ADD COLUMN` statement per column
- Removed the lazy `createTable()` guard from `AwsRedshiftStorage.saveData()` (now handled in `init()`)

### DIFF
--- a/packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
+++ b/packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
@@ -100,13 +100,19 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
 
   //---- getAListOfExistingColumns ----------------------------------
   async getAListOfExistingColumns() {
-    // Strip surrounding double-quotes that may be present in config values,
-    // then compare case-insensitively via LOWER() so the query works regardless
-    // of whether enable_case_sensitive_identifier is on or off.
+    // Strip surrounding double-quotes that may be present in config values.
+    // Prefer exact matching first because quoted Redshift identifiers can be
+    // case-sensitive; fall back to LOWER() for clusters that fold identifiers.
     const rawSchema = stripQuotes(this.config.Schema.value);
     const rawTable = stripQuotes(this.config.DestinationTableName.value);
 
-    const sql = `SELECT column_name, data_type
+    const exactSql = `SELECT column_name, data_type
+      FROM information_schema.columns
+      WHERE table_schema = '${rawSchema}'
+        AND table_name = '${rawTable}'
+      ORDER BY ordinal_position`;
+
+    const fallbackSql = `SELECT column_name, data_type
       FROM information_schema.columns
       WHERE LOWER(table_schema) = LOWER('${rawSchema}')
         AND LOWER(table_name) = LOWER('${rawTable}')
@@ -114,7 +120,10 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
 
     let rows;
     try {
-      rows = await this.executeQueryWithResults(sql);
+      rows = await this.executeQueryWithResults(exactSql);
+      if (rows.length === 0) {
+        rows = await this.executeQueryWithResults(fallbackSql);
+      }
     } catch (error) {
       if (error.message && error.message.includes('does not exist')) {
         return {};
@@ -122,11 +131,12 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
       throw error;
     }
 
-    // Build a lowercase → configured-key map so that info_schema's lowercased names
-    // (Redshift folds unquoted identifiers to lowercase) resolve back to the
-    // connector's schema key (which may be mixed-case, e.g. CampaignId).
-    // This ensures existingColumns keys match getSelectedFields() output and
-    // loadTableSchema() does not treat already-present columns as missing.
+    // Redshift folds even quoted identifiers to lowercase by default, so a column
+    // created as "CampaignId" comes back as campaignid. Map info_schema names back
+    // to the configured schema key (case-insensitively) whenever the returned name
+    // doesn't already match a key exactly, so existingColumns keys line up with
+    // getSelectedFields() and loadTableSchema() doesn't re-add existing columns.
+    // An exact name always wins (handles case-sensitive clusters untouched).
     const schemaKeyByLower = {};
     for (const key of Object.keys(this.schema || {})) {
       schemaKeyByLower[key.toLowerCase()] = key;
@@ -134,7 +144,10 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
 
     const columns = {};
     for (const row of rows) {
-      const schemaKey = schemaKeyByLower[row.column_name.toLowerCase()] || row.column_name;
+      const columnName = row.column_name;
+      const schemaKey = (this.schema && columnName in this.schema)
+        ? columnName
+        : (schemaKeyByLower[columnName.toLowerCase()] || columnName);
       columns[schemaKey] = (this.schema && schemaKey in this.schema)
         ? this.getColumnType(schemaKey)
         : row.data_type;
--- a/packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
+++ b/packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
@@ -73,6 +73,100 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
   async init() {
     await this.checkConnection();
     this.config.logMessage('Connection to Redshift established');
+    await this.loadTableSchema();
+  }
+  //----------------------------------------------------------------
+
+  //---- loadTableSchema --------------------------------------------
+  async loadTableSchema() {
+    this.existingColumns = await this.getAListOfExistingColumns();
+
+    if (Object.keys(this.existingColumns).length === 0) {
+      await this.createTable();
+    } else {
+      const selectedFields = this.getSelectedFields();
+      const newFields = selectedFields.filter(col => !(col in this.existingColumns));
+      if (newFields.length > 0) {
+        await this.addNewColumns(newFields);
+      }
+    }
+  }
+  //----------------------------------------------------------------
+
+  //---- getAListOfExistingColumns ----------------------------------
+  async getAListOfExistingColumns() {
+    // Redshift stores unquoted identifiers as lowercase; strip any surrounding
+    // double-quotes from config values before comparing against information_schema.
+    const rawSchema = this.config.Schema.value.replace(/^"|"$/g, '').toLowerCase();
+    const rawTable = this.config.DestinationTableName.value.replace(/^"|"$/g, '').toLowerCase();
+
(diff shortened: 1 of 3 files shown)


packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
```diff
@@ -73,6 +73,100 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
...
+      }
+    }
+  }
+  //----------------------------------------------------------------
+
+  //---- getAListOfExistingColumns ----------------------------------
+  async getAListOfExistingColumns() {
+    // Redshift stores unquoted identifiers as lowercase; strip any surrounding
+    // double-quotes from config values before comparing against information_schema.
+    const rawSchema = this.config.Schema.value.replace(/^"|"$/g, '').toLowerCase();
```
max-voloshyn: [P2] Lowercasing the schema/table before querying `information_schema` makes Redshift miss quoted mixed-case targets when `enable_case_sensitive_identifier=true` ([AWS docs](https://docs.aws.amazon.com/redshift/latest/dg/r_enable_case_sensitive_identifier.html)). The DDL below still uses the original quoted config values, so `"MySchema"."MyTable"` can be created/referenced but never discovered here; `loadTableSchema()` then treats the table as absent and skips the real `addNewColumns()` path for existing tables. Please use the same identifier normalization for metadata lookup and DDL.

🤖 Reviewed by Codex (GPT-5.5)

packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
```diff
@@ -73,6 +73,100 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
...
+    } catch (error) {
+      if (error.message && error.message.includes('does not exist')) {
+        return {};
+      }
+      throw error;
+    }
+
+    const columns = {};
+    for (const row of rows) {
+      const columnName = row.column_name;
```
max-voloshyn, line 147: [P1] Redshift returns column names lowercase by default ([AWS docs](https://docs.aws.amazon.com/redshift/latest/dg/r_names.html)), but this stores `row.column_name` directly and later compares it to connector field names like `CampaignId` or `AdvertiserCost`. That makes `loadTableSchema()` think existing mixed-case fields are missing and issue `ALTER TABLE ... ADD COLUMN "CampaignId"` again, which reintroduces the failure this PR is meant to fix. Please compare via normalized names while preserving the configured schema key in `existingColumns`.

🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (commented):
See inline comments.

packages/connectors/src/Storages/Snowflake/SnowflakeStorage.js
```diff
@@ -217,11 +217,15 @@ var SnowflakeStorage = class SnowflakeStorage extends AbstractStorage {
...
+      // INFORMATION_SCHEMA stores bare identifiers, so strip them before comparing.
+      const rawSchemaName = this.config.SnowflakeSchema.value.replace(/^"|"$/g, '');
+      const rawTableName = this.config.DestinationTableName.value.replace(/^"|"$/g, '');
+
       let query = `SELECT COLUMN_NAME, DATA_TYPE, IS_NULLABLE
         FROM ${this.config.SnowflakeDatabase.value}.INFORMATION_SCHEMA.COLUMNS
-        WHERE TABLE_CATALOG = '${this.config.SnowflakeDatabase.value}'
-          AND TABLE_SCHEMA = UPPER('${this.config.SnowflakeSchema.value}')
-          AND TABLE_NAME = UPPER('${this.config.DestinationTableName.value}')
+        WHERE TABLE_SCHEMA = UPPER('${rawSchemaName}')
```
max-voloshyn: [P1] `UPPER(rawSchemaName)` still misses quoted Snowflake schemas with lowercase or mixed-case names. This storage creates/references schemas with `quoteIdentifier(...)`, so `"marketingRaw"` is stored as exactly `marketingRaw`, but this metadata query looks for `MARKETINGRAW` and returns no columns. `loadTableSchema()` then treats the table as absent and never calls `addNewColumns()` for the existing table, so the next MERGE can still fail. Please use the same casing rules here as the DDL path.

🤖 Reviewed by Codex (GPT-5.5)

packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
prokopetsv: Thanks! Fixed in the next commit.

packages/connectors/src/Storages/Snowflake/SnowflakeStorage.js
prokopetsv: Thanks! Fixed in the next commit.

packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
prokopetsv: Thanks! Fixed in the next commit.

max-voloshyn:
- The Snowflake findings look addressed.
- The Redshift column-name mapping finding looks addressed.
- The remaining Redshift `LOWER(table_schema/table_name)` ambiguity is a real edge risk, but I would not treat it as a PR blocker unless the product explicitly supports case-distinct Redshift schemas/tables.
- The potential blocker is still Redshift eager schema/table handling in `init()`: `init()` calls `loadTableSchema()`, and `loadTableSchema()` can `createTable()` or `addNewColumns()` before `saveData()` knows whether there is data or whether `CreateEmptyTables=false` should suppress table creation.

My call: if `CreateEmptyTables=false` is supposed to mean “do not create/evolve destination tables for empty runs,” this should be fixed before merge. If Redshift intentionally joins Snowflake/BigQuery-style eager initialization, then it is not a blocker, but it should be an explicit product decision because it changes prior Redshift behavior.

🤖 Reviewed by Codex (GPT-5.5)

prokopetsv:
@max-voloshyn Kindly ask to review.

max-voloshyn (commented):
See inline comment.

packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
```diff
@@ -389,16 +499,17 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
    * @returns {Promise}
    */
   async saveData(data) {
-    if (!data || data.length === 0) {
-      return Promise.resolve();
-    }
-
-    // Create table if it doesn't exist
+    // Create table before the empty-data guard so that CreateEmptyTables=true
+    // calls with data=[] still result in the table being created.
```
max-voloshyn: [P1] This now creates a Redshift table for any `saveData([])` call, even when `CreateEmptyTables=false`. Some callers can pass empty batches without checking that flag first, e.g. Facebook catalog pagination saves `pageData` from the page callback and Microsoft Ads can save an empty filtered batch. That means the PR still changes the user-facing `CreateEmptyTables=false` behavior for Redshift. Please keep the empty-data guard first, and only call `createTable()` for empty data when `this.config.CreateEmptyTables?.value` is true.

🤖 Reviewed by Codex (GPT-5.5)

packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
```diff
@@ -73,6 +77,112 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
...
+  async getAListOfExistingColumns() {
+    // Strip surrounding double-quotes that may be present in config values,
+    // then compare case-insensitively via LOWER() so the query works regardless
+    // of whether enable_case_sensitive_identifier is on or off.
+    const rawSchema = stripQuotes(this.config.Schema.value);
+    const rawTable = stripQuotes(this.config.DestinationTableName.value);
+
+    const sql = `SELECT column_name, data_type
+      FROM information_schema.columns
+      WHERE LOWER(table_schema) = LOWER('${rawSchema}')
```
max-voloshyn, line 117: [P2] This metadata lookup matches schema/table case-insensitively, while the DDL path references the quoted configured identifiers. On Redshift clusters that support case-sensitive quoted identifiers, two objects can differ only by case, so this can read columns from the wrong table and then skip needed `ALTER TABLE` work for the actual target. Please prefer an exact `table_schema = ... AND table_name = ...` lookup first, and only fall back to case-insensitive matching when the exact lookup returns no rows.

🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (approved):

PR #1238, opened 2026-05-25
feat(user-provisioning): add user provisioning settings and request access context

amarchenk0:


max-voloshyn (commented):
Superseded by standalone inline comments.

max-voloshyn (commented):
Superseded by standalone inline comments.

apps/web/src/features/contexts/types/context.types.ts
```diff
@@ -17,6 +17,7 @@ export interface ContextImpactDto {
   storageCount: number;
   destinationCount: number;
   memberCount: number;
+  userProvisioningDefaultsCount: number;
```
max-voloshyn, line 20: Line 20 adds `userProvisioningDefaultsCount`, but the delete guard in `ContextsTab.tsx:39` still computes the blocking total without it. A context referenced only by provisioning defaults will show the delete confirmation and then fail on the backend 409 instead of showing the blocking dialog. Please include this count in the total and in `renderAttachments`.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/project-members/update-user-provisioning-settings.service.ts
```diff
@@ -0,0 +1,82 @@
...
+      const members = await this.idpProjectionsFacade.getProjectMembersOrThrow(projectId);
+      await this.applyDefaultsService.preserveExistingMembersAsEntireProject(
+        projectId,
+        members.map(member => member.userId)
+      );
+    }
+
+    const shouldSaveLocalDefaultsFirst = contextDefaults.roleScope === RoleScope.SELECTED_CONTEXTS;
+    const savedContextDefaults = shouldSaveLocalDefaultsFirst
+      ? await this.contextSettingsService.saveDefaultSettings(projectId, contextDefaults)
```
max-voloshyn: Line 40 saves ODM-local selected-context defaults before the IDP update at line 43. If the IDP call fails or returns not-applicable, local defaults stay changed even though the request failed, so later members can receive scopes from a failed settings update. Please persist after a successful IDP update or add rollback/compensation.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/controllers/request-access.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+  RequestAccessContextApiDto,
+  RequestProjectAccessApiDto,
+  RequestProjectAccessResponseApiDto,
+} from '../dto/presentation/request-access-api.dto';
+import { CreateNewProjectService } from '../use-cases/project-members/create-new-project.service';
+import { GetRequestAccessContextService } from '../use-cases/project-members/get-request-access-context.service';
+import { RequestProjectAccessService } from '../use-cases/project-members/request-project-access.service';
+
+@Controller('user-provisioning')
+@ApiTags('User Provisioning')
```
romandubovyi, line 21: **Missing OpenAPI spec decorators**

All other controllers in this module (`ProjectMembersController`, `ReportController`, `DataMartController`, etc.) use per-endpoint spec decorator functions from `spec/*.api.ts` files to document response types, error codes, and operation summaries for Swagger. The three endpoints here (`getContext`, `requestAccess`, `createNewProject`) have no such decorators — only `@ApiTags` at the class level. These endpoints will be missing from the generated OpenAPI schema.

Consider adding a `spec/request-access.api.ts` file following the existing pattern (e.g. `GetUserProvisioningSettingsSpec`, `UpdateUserProvisioningSettingsSpec` in `spec/project-members.api.ts`).

apps/web/src/layouts/MainLayout.tsx
```diff
@@ -32,13 +135,27 @@ function MainLayoutContent() {
...
-          {showTrigger && <SidebarTrigger />}
-          <ProjectIdGuard>
-            <Outlet />
-          </ProjectIdGuard>
-        </SidebarInset>
+        <ProjectIdGuard>
+          <ProjectRoleGuard>
+            {hasEmptyProjectRoles ? (
+              <>
+                <RestrictedProjectSidebar user={user} />
```
romandubovyi, line 142: **`user` may be `null` here**

`useUser()` returns `User | null`. The intermediate boolean `hasEmptyProjectRoles` (line 128) uses optional chaining (`user?.roles`), but TypeScript cannot narrow `user` through a pre-computed boolean variable. In this truthy branch, `user` is still typed as `User | null` while `RestrictedProjectSidebar` expects `User`.

Under `strict: true` this is a compile error. Fix by inlining the condition or adding a null guard:
```ts
if (user && user.roles?.length === 0) {
  <RestrictedProjectSidebar user={user} />
```

apps/backend/src/data-marts/use-cases/project-members/update-user-provisioning-settings.service.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    if (contextDefaults.roleScope === RoleScope.SELECTED_CONTEXTS) {
+      const members = await this.idpProjectionsFacade.getProjectMembersOrThrow(projectId);
+      await this.applyDefaultsService.preserveExistingMembersAsEntireProject(
+        projectId,
+        members.map(member => member.userId)
+      );
+    }
+
+    const shouldSaveLocalDefaultsFirst = contextDefaults.roleScope === RoleScope.SELECTED_CONTEXTS;
```
romandubovyi: **Local defaults committed before IDP call — state inconsistency on IDP failure**

When `roleScope === SELECTED_CONTEXTS`, `saveDefaultSettings` is called here (line 39) before the IDP update (line 42). If IDP then returns `isApplicable: false` (line 51), the method returns early — but the local ODM defaults are already committed via the `@Transactional` boundary in `saveDefaultSettings`.

The caller receives `{ isApplicable: false, settings: null }`, implying no settings exist, while the database holds the saved defaults. Consider wrapping both operations in a single transaction, or rolling back the local save on IDP failure.

romandubovyi (commented):
Code review: found 3 issues.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1240, opened 2026-05-25
feat(data-marts): extend access-aware blending with SQL-build-time check and report-builder UX

kalnik-a-a:
## Summary

Extends the existing access-aware blending model with two planned improvements:

1. **Backend** — promotes the per-source access state into the report-builder schema (`isAccessibleForReporting` on `AvailableSource`, cascading along the join chain) and adds a SQL-build-time validation that mirrors the existing save-time orphan check. The error message names the accessor responsible for the run.
2. **Frontend** — the report builder now visualises inaccessible joined data marts (red joined-DM blocks with a no-access indicator), and the column picker keeps inaccessible columns visible when they are still referenced by the report's `value` / `filterConfig` / `sortConfig`, so the report owner can clear those references in-place.

## Why

The previous round of access-aware blending validated references at save time and gated the relationship canvas, but the column picker rendered every joined data mart uniformly and the run path relied on the save-time check alone. That left two gaps that we deliberately deferred to a follow-up:

- **No in-builder feedback for access state on joined data marts** — the picker did not distinguish accessible from inaccessible joined blocks, so authors with mixed access could not tell at a glance which fields they could actually use in a new report.
- **No SQL-build-time check** — a report whose creator lost access to a joined DM after save (manual run by another viewer, scheduled run, Looker Studio cached pull) had no second gate at SQL build time, so the failure surfaced opaquely from the storage engine.

This PR closes both, completing the originally planned access-aware blending UX and adding the SQL-build-time defence-in-depth.

### Backend (`apps/backend/src/data-marts/`)

- **`AvailableSourceDto.isAccessibleForReporting`** — computed by `BlendableSchemaService.applyReportingAccess` via `accessDecisionService.canAccessMany` with `Action.USE`. Cascades along `aliasPath` so a child reachable via a denied parent is consistently marked inaccessible, keeping the schema's access view consistent with the canvas / orphan validator.
- **`RelationshipDto.{source,target}DataMart.userHasAccess`** — populated by all relationship list / single-fetch endpoints, so the relationship canvas can mark unreachable nodes from the initial graph payload instead of inferring access from per-node 403 round-trips.
- **`BlendedReportDataService.assertAllRequestedSourcesAccessible`** — new defence-in-depth check. Runs after schema computation and before any DB I/O, walks the full join chain (`columnConfig` ∪ post-join filter columns ∪ sort columns, plus pre-join slice aliasPaths and all their ancestors), and throws a `BusinessViolationException` listing the inaccessible data-mart titles if any path is denied for the current accessor. The shared helper `collectNeededAliasPaths` is now the single derivation of the join-chain alias-path set, used both by this gate and by `buildRelationshipChains`.
- **Error message names the affected user** — `user "<label>" is missing access to data marts: "X", "Y"`. Label is resolved through `UserProjectionsFetcherService` with `fullName → email → userId` fallback; `userId` is also persisted in `errorDetails` for log consumers. The fetch only fires on the deny path, so the happy path adds no extra DB work.
- **Uniform behaviour across run contexts** — manual runs use the active viewer's accessor; scheduled runs and Looker Studio cached pulls use the report creator's accessor (the same accessor that already drives `isAccessibleForReporting` in the schema). The exception bubbles through `RunReportService`, `ReportSqlComposerService`, and `ReportDataCacheService` unchanged.
- **`RunReportService.executeReport()`** — defers reader/writer resolution until after access validation, so a denied run no longer allocates a storage reader or destination writer it never uses, and the `finalize()` calls are guarded against `null` on that path.
- **Pre-join slices participate in the same gate** — slices targeting an inaccessible aliasPath are validated alongside columns, filters, and sorts.

### Frontend (`apps/web/src/features/data-marts/`)

- **Relationship canvas** — uses `userHasAccess` on the relationship payload to mark unreachable nodes and skip descendant fetches from the initial graph payload. The new `NoAccessIndicator` component is shared between the canvas, the relationship accordion item, and the report column picker.
- **`ReportColumnPicker`** — joined-DM groups with `isAccessibleForReporting === false` render with a red border / red header / no-access icon. The group is expandable, but the master **Select all** sources only from native + accessible blended fields. **Show selected only** honours the same filter. Inaccessible fields stay visible inside the red group when they are currently selected, filtered, sliced, or sorted on — so the report owner can find and clear the rule in-place rather than guessing which column the run failure refers to.
- **Selection math** — `selectableFieldNames` (native + accessible blended) drives both `selectAll` / `deselectAll` and the counter denominator, while a separate `selectedInaccessibleBlendedCount` keeps already-selected inaccessible fields counted toward `X / Y`.
- **New tests** — `ReportColumnPicker.test.tsx` covers the inaccessible-group rendering, master-checkbox behaviour, counter math, and the filter / sort / slice visibility rule.

## Behaviour at a glance

- **Manual run** of a report whose creator has lost access to a joined DM: opaque storage-engine error → HTTP 400 `user "<label>" is missing access to data marts: "<title>"` with `errorDetails = { userId, deniedDataMartIds, deniedAliasPaths }`.
- **Scheduled run** of the same report: same `BusinessViolationException`, surfaces in the scheduled-run status.
- **Looker Studio pull** of the same report: same gate, propagated through `ReportDataCacheService`.
- **Column picker** for a user with mixed access: joined groups now distinguish accessible from inaccessible (red border + no-access indicator). Master **Select all** sources only from accessible fields.
- **Pre-existing rule on an inaccessible field**: the field stays visible inside the red joined block with its filter / slice / sort icon, so the report owner can clear the rule in-place.
- **Relationship canvas** initial load: greyed nodes come from the initial graph payload via `userHasAccess` instead of per-node 403 round-trips.

## Migration

- Reports that reference columns, filters, sorts, or slices on a joined data mart the **current run accessor** cannot reach now produce a clear HTTP 400 instead of a storage-engine error. The message names the responsible user and the inaccessible data-mart titles; the column picker surfaces those columns inside the red joined block so the report owner can clear them.
- Scheduled runs follow the same rule against the schedule creator's accessor. To rescue a schedule whose creator has lost access, either clear the inaccessible references from the report config or reassign the schedule to an accessor with the needed permission.
- No DB migrations, no new endpoints, no changes to the `Action` enum or roles.

## Tests

- Backend: `nest build` clean, targeted Jest specs green (`blended-report-data.service.spec.ts`, `report-sql-composer.service.spec.ts`, `run-report.service.spec.ts` — 47/47, covering the cascade, single-DM and multi-DM denial messages, native-only short-circuit, schedule / Looker accessor paths, and deny-side user-label fallback chain).
- Frontend: `tsc -b && vite build` clean, Vitest 41/41 (5 new cases in `ReportColumnPicker.test.tsx`).
- Lint clean for both workspaces.
- Manual QA against the running app (Playwright): a report with a filter on the `Orders` joined data mart (no `USE` for the current user) renders the `Orders` group in red with the filtered column visible inside, and `Run` produces the expected HTTP 400 with the user's display name.

## Notes

- The blendable-schema cache is keyed by the requesting accessor, so per-user access changes invalidate naturally — no new cache key is required.
- Pre-join slice columns are matched by `aliasPath + originalFieldName` in the visibility helper so the picker's slice popover and the cascade gate stay in sync.
- The Google Sheets add-on ships the mirror UI changes in a separate PR against `OWOX/google-sheets-extension`


max-voloshyn (commented):
Inline findings from Codex review.

apps/backend/src/data-marts/services/blendable-schema.service.ts
```diff
@@ -17,6 +18,21 @@ import { AggregateFunction } from '../dto/schemas/aggregate-function.schema';
...
+  roles: string[];
+}
+
+export async function resolveBlendableSchemaAccessor(
+  idpProjectionsFacade: IdpProjectionsFacade,
+  projectId: string,
+  userId: string
+): Promise<BlendableSchemaAccessor> {
+  const member = await idpProjectionsFacade.getProjectMember(projectId, userId);
+  return { userId, roles: member ? [member.role] : [] };
```
max-voloshyn: [P1] Missing project members should not fall back to viewer access

`resolveBlendableSchemaAccessor()` returns `roles: []` when `getProjectMember()` does not find the user. `AccessDecisionService.resolveRole()` then treats an empty roles array as `Role.VIEWER`, so scheduled runs / Looker cache pulls for a removed creator can still pass `USE` checks for data marts shared for reporting. That conflicts with the PR’s run-accessor model: if the responsible user is no longer a project member, the accessor should deny or throw instead of becoming a viewer.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/get-data-mart-relationship.service.ts
```diff
@@ -49,8 +32,18 @@ export class GetDataMartRelationshipService {
       );
     }
 
-    const createdByUser = await this.userProjectionsFetcherService.fetchCreatedByUser(relationship);
-
-    return this.mapper.toDomainDto(relationship, createdByUser);
+    const [accessByDataMartId, createdByUser] = await Promise.all([
+      this.accessDecisionService.canAccessMany(
```
max-voloshyn: [P1] Keep the source DataMart visibility gate here as well

This single-relationship path has the same regression as the list endpoint. The new `userHasAccess` flag is useful for inaccessible targets, but it should not replace the source DataMart SEE boundary. A user who cannot see `sourceDataMartId` should still get 403 before relationship metadata is returned.

🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -541,33 +580,64 @@ export function ReportColumnPicker({
...
       let group = groupMap.get(field.aliasPath);
       if (!group) {
+        const source = availableSourceByPath.get(field.aliasPath);
         group = {
           aliasPath: field.aliasPath,
           title: field.sourceDataMartTitle,
           alias: field.outputPrefix,
-          description: availableSourceDescriptionByPath.get(field.aliasPath),
+          description: source?.description,
+          isAccessibleForReporting: source?.isAccessibleForReporting ?? false,
```
max-voloshyn, line 633: [P2] Do not offer inaccessible sources as new slice targets

This code correctly carries `isAccessibleForReporting` into the rendered blended group, but the source list used by `OutputSettingsDropdown` for slices is still built from every included source (`joinedSources`) without applying this same flag. The PR description says inaccessible fields should stay visible when they are already selected, filtered, sliced, or sorted so the owner can clear those references; it should not let users create new pre-join filters against locked joined marts. Please restrict new slice targets to accessible sources while preserving existing inaccessible slice rules for display/removal.

🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -365,22 +386,38 @@ export function ReportColumnPicker({
...
+    for (const source of schema?.availableSources ?? []) {
+      map.set(source.aliasPath, source);
+    }
+    return map;
+  }, [schema?.availableSources]);
+
+  const accessibleBlendedFieldNames = useMemo(
+    () =>
+      includedBlendedFields
+        .filter(f => availableSourceByPath.get(f.aliasPath)?.isAccessibleForReporting === true)
```
max-voloshyn, line 416: [P2] Do not offer inaccessible blended fields in new filter controls

This computes the accessible blended field list for selection, but `dropdownColumns` still builds filter options from every `includedBlendedField`, regardless of source accessibility. That means `OutputSettingsDropdown` can offer new post-join filters against inaccessible joined marts, creating references the backend will now reject. The intended behavior from the PR description is to keep existing inaccessible references visible so users can clear them, not to allow adding new ones. Please filter add-control options to accessible blended fields while keeping already-referenced inaccessible fields visible in the picker.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/list-data-mart-relationships.service.ts
```diff
@@ -24,23 +24,25 @@ export class ListDataMartRelationshipsService {
...
 
-    const userProjectionsList =
-      await this.userProjectionsFetcherService.fetchRelevantUserProjections(relationships);
+    const involvedDataMartIds = [
+      command.dataMartId,
+      ...relationships.map(r => r.targetDataMart.id),
+    ];
+
+    const [accessByDataMartId, userProjectionsList] = await Promise.all([
+      this.accessDecisionService.canAccessMany(
```
max-voloshyn: [P1] Keep the source DataMart visibility gate

The PR description explains why inaccessible joined targets should be returned with `userHasAccess=false`, and that part makes sense for the canvas. But this endpoint also removed the visibility check for the source/root `command.dataMartId`. As written, a viewer who cannot SEE that data mart but knows its id can still fetch relationship metadata and target titles. Please keep the source `canAccess(... Action.SEE ...)` guard before loading relationships, while using `userHasAccess` only for target/descendant access state.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/blendable-schema.service.ts
kalnik-a-a: Fixed in 527689ca6. `resolveBlendableSchemaAccessor` now throws `BusinessViolationException` (with `userId` + `projectId` in `errorDetails`) when `getProjectMember` returns `undefined`, instead of silently degrading to viewer access. Scheduled runs and Looker cache pulls on behalf of a removed creator now surface as explicit failed runs. System triggers that legitimately use `roles: []` (`schema-actualize-trigger-handler`, `sql-dry-run-trigger-handler`) construct their accessor directly without going through this function and are unaffected.

apps/backend/src/data-marts/use-cases/get-data-mart-relationship.service.ts
kalnik-a-a: Addressed in 7ff048975 by a deeper refactor. The `GET /relationships/:id` and `GET /relationships` endpoints (along with their use-cases, specs, commands, mapper helpers and OpenAPI specs) are deleted entirely. They're replaced by a single bulk endpoint `GET /:dataMartId/relationships/graph` that:

1. resolves the root data mart inside the project,
2. loads all storage relationships in one query and walks the graph in memory (cycle detection, cumulative `isBlocked` cascade, alias paths),
3. resolves access via a single `canAccessMany` call over the union of root + every involved DM (cycle-stub targets are excluded from that union),
4. throws `ForbiddenException` if the **root** DM is not in the access map.

The frontend now consumes this single endpoint for both the table view (filtered to depth=1) and the canvas walker — the previous recursive HTTP traversal (`useTransientRelationships.ts:81`, `RelationshipCanvas.tsx:489`) is gone. No more action-at-a-distance: there is no longer an endpoint that returns relationship metadata for a DM the caller can't SEE. The UX requirement of "graph visible end-to-end once you have access to the root" is preserved because the server-side walker keeps going through inaccessible targets — they surface to the UI with `userHasAccess=false` flags (warning icon) instead of being hidden.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
kalnik-a-a: Fixed

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
kalnik-a-a: Fixed

apps/backend/src/data-marts/use-cases/list-data-mart-relationships.service.ts
kalnik-a-a: Same fix as on the get endpoint (above): 7ff048975 removes both `list` and `get/:id` and replaces them with a single bulk graph endpoint that gates SEE on the root only and exposes per-target `userHasAccess` through the inner `RelationshipResponseApiDto`. The transient traversal that drove the original "no SEE gate on root" workaround now lives entirely server-side - the frontend issues one HTTP request per data mart for the whole table + canvas + transient subtree.

apps/backend/src/data-marts/use-cases/run-report.service.ts
```diff
@@ -168,7 +175,13 @@ export class RunReportService {
       );
     }
 
-    await this.executeReportRunWithCleanup(reportRun, signal);
+    const accessor = await this.resolveAccessor(runByUserId, actualProjectId);
```
max-voloshyn: [P1] Resolve the run accessor inside the report-run cleanup path

`ReportRunTriggerHandlerService` claims the `DataMartRun` as `RUNNING` before calling `executeExistingRun()`. With the new missing-member behavior, `resolveAccessor()` can now throw before `executeReportRunWithCleanup()` is entered, so `RunReportService` never calls `handleReportRunError()`. The outer trigger handler then sees the same run is already `RUNNING` and treats it as a duplicate trigger, leaving the scheduled/manual trigger run stuck instead of marking it failed. This contradicts the PR description behavior for scheduled-run status when a creator is removed. Move accessor resolution into the cleanup try/catch, or otherwise make this pre-execution failure finish the claimed run as unsuccessful.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/blendable-schema.service.ts
```diff
@@ -17,6 +18,27 @@ import { AggregateFunction } from '../dto/schemas/aggregate-function.schema';
...
+  userId: string;
+  roles: string[];
+}
+
+export async function resolveBlendableSchemaAccessor(
+  idpProjectionsFacade: IdpProjectionsFacade,
+  projectId: string,
+  userId: string
+): Promise<BlendableSchemaAccessor> {
+  const member = await idpProjectionsFacade.getProjectMember(projectId, userId);
```
max-voloshyn: [P2] Do not use the non-strict project-member lookup for run accessors

`getProjectMember()` is backed by `getProjectMembers()`, whose implementation explicitly swallows IDP failures and returns an empty list. In this new run-accessor path, a transient IDP failure is therefore indistinguishable from a removed creator and becomes a `BusinessViolationException` saying the user is no longer a project member. Scheduled runs / Looker pulls should fail closed, but they should not persist a misleading business failure for an infrastructure lookup error. Please resolve the member through a strict path such as `getProjectMembersOrThrow()` and only convert an actual missing member into this business violation.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/blended-report-data.service.ts
```diff
@@ -41,6 +46,7 @@ export class BlendedReportDataService {
       filterConfig: report.filterConfig ?? null,
       sortConfig: report.sortConfig ?? null,
       limitConfig: report.limitConfig ?? null,
+      accessor,
     });
 
     if (columnConfig === null || columnConfig === undefined) {
```
max-voloshyn, line 64: [P2] Do not skip joined-filter validation when columnConfig is null

The new access gate is supposed to consider `columnConfig ∪ post-join filter columns ∪ sort columns`, but this early return exits before post-join filters are examined whenever the report uses the default column selection (`columnConfig === null`). The UI can still create post-join filters from `allColumns`, including accessible blended fields, while keeping `columnConfig` null. That report then bypasses the new joined-source/access check and falls through to native query composition with a joined-field filter, producing the opaque storage/query-builder failure this PR is meant to replace. Either reject blended output controls when `columnConfig` is null, or promote this case into the blended path before returning.

🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -244,31 +253,42 @@ function BlendedGroupItem({
   preJoinByAliasPathColumn,
 }: BlendedGroupItemProps) {
   const [isOpen, setIsOpen] = useState(() => group.selectedCount > 0);
+  const inaccessible = !group.isAccessibleForReporting;
```
max-voloshyn, line 271: [P2] Keep inaccessible blended rows remove-only

The dropdown add paths are filtered now, but visible fields inside an inaccessible group still receive `onToggleField`, `filterableTypeFor?.(field.name)`, and the normal add/replace handlers. That lets a referenced locked row create new backend-rejected references: checking the field adds it to `columnConfig`, and `RowFilterIcon` still allows adding or editing post-join filters and pre-join slices. The PR description says inaccessible fields remain visible so the owner can clear existing references in-place, not create or edit them. For `group.isAccessibleForReporting === false`, please keep removal paths available but block new selection and add/edit filter or slice actions.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/report-data-cache.service.ts
```diff
@@ -180,7 +201,8 @@ export class ReportDataCacheService {
   private async finalizeExpiredCacheEntry(cacheEntry: ReportDataCache): Promise<void> {
     try {
       if (cacheEntry.readerState) {
-        const { options } = await this.resolvePrepareOptions(cacheEntry.report);
+        const accessor = await this.resolveCreatorAccessor(cacheEntry.report);
```
max-voloshyn: [P3] Do not let creator access loss block cache finalization

Expired/invalidation cleanup now resolves the report creator accessor before restoring the cached reader. If the creator was removed from the project, this throws before `restoreReaderFromCache()` and `reader.finalize()` run, but the caller still deletes the cache row after logging the rejected finalization. That means the same missing-creator case introduced by this PR can permanently skip reader cleanup for cached Looker data, including storage-specific cleanup such as Athena output files. Cache finalization should not require current report-run access; it should either use stored prepare state or a cleanup path that can restore/finalize the reader without converting membership loss into skipped cleanup.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/run-report.service.ts
kalnik-a-a: Fixed in 1cce8c97c. `resolveAccessor` is now called from inside `executeReportRunWithCleanup`'s try/catch, so a removed-creator throw lands in `handleReportRunError` and marks the DataMartRun as failed instead of leaving it stuck in RUNNING. Scheduled and manual triggers no longer get stuck on this path.

apps/backend/src/data-marts/services/blendable-schema.service.ts
kalnik-a-a: Fixed in 1cce8c97c. Added `IdpProjectionsFacade.getProjectMemberOrThrow` (mirrors the existing strict `getProjectMembersOrThrow`) and switched `resolveBlendableSchemaAccessor` to it. Transient IDP failures now propagate as infrastructure errors instead of being persisted as a membership-loss business violation; only a real `undefined` member (verified missing) still raises the existing `BusinessViolationException`.

apps/backend/src/data-marts/services/blended-report-data.service.ts
kalnik-a-a: Fixed in 1cce8c97c. The `columnConfig === null` branch no longer short-circuits before inspecting filter/sort/slice references. If any post-join filter or sort column resolves to a blended field, or there is a pre-join slice present, the service throws `BusinessViolationException` (with the offending columns and alias paths in `errorDetails`) instead of falling through to native composition. Pure-native default-column reports keep the original cheap path.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
kalnik-a-a: Fixed in 1cce8c97c. `BlendedFieldRow` now takes a `removeOnly` prop that we pass for every row inside an inaccessible group. In that mode the checkbox is disabled while `!checked` (so the field can still be unchecked, but not added to `columnConfig`), and `onAdd` / `onAddSlice` / `onReplaceFilter` are not forwarded — so the popover only exposes the existing-references list with a remove action. `RowFilterIcon`'s `onAdd`/`onAddSlice` props became optional to support this.

apps/backend/src/data-marts/services/report-data-cache.service.ts
kalnik-a-a: Fixed in 1cce8c97c. `ReportDataCacheService` now finalizes expired entries with a private admin-roled `cleanupAccessor` sentinel instead of calling `resolveCreatorAccessor`. Cache cleanup is system-level work — `reader.finalize()` must not be skipped just because the original creator was removed from the project, otherwise storage-specific cleanup (Athena output files, BigQuery temp tables, etc.) leaks while the cache row gets deleted. The sentinel is private to the cleanup path and never reaches user-driven read flows.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
max-voloshyn: Thanks, this fixes the new selection and add/edit paths, but the existing pre-join slice removal path is still lost in remove-only mode.

For inaccessible rows, `BlendedFieldRow` still passes `preJoinSlices`, but omits `onAddSlice`. Then `RowFilterIcon` only renders `FilterOrSliceEditorPopover` when `handleSliceApply` exists; otherwise it falls back to `FilterEditorPopover`, which can only show/remove post-join filters. So a locked field that is visible only because it has an existing pre-join slice still shows the active filter icon, but opening it does not expose the slice list or the remove action.

Please keep the slice popover available when existing slices are present, even if add/replace handlers are omitted, so remove-only mode really allows clearing existing slice references in-place.

🤖 Reviewed by Codex (GPT-5.5)

kalnik-a-a: Good catch — fixed in b9e7db566. The slice tab now opens whenever there is something to manage in it (either an add path **or** existing slices). When the add path is locked, `addSlice?.(rule)` is a no-op, so the popover still renders the Active slices list with its remove buttons and a locked field that's only there because of an existing pre-join slice can have that slice cleared in-place. Mirrored in the Google Sheets extension via `601136e`.

max-voloshyn (approved):

PR #1241, opened 2026-05-26
feat(api): add Project Member API Key management entity

amarchenk0:


apps/backend/src/project-member-api-keys/services/project-member-api-key.service.ts
```diff
@@ -0,0 +1,117 @@
...
+import { ProjectMemberApiKeyMapper } from '../mappers/project-member-api-key.mapper';
+import { ProjectMemberApiKeyCryptoService } from './project-member-api-key-crypto.service';
+
+export type CreateProjectMemberApiKeyResult = {
+  apiKey: ProjectMemberApiKeyMetadata;
+  apiKeySecret: string;
+};
+
+@Injectable()
+export class ProjectMemberApiKeyService {
```
max-voloshyn, line 16: [P2] This service is marked injectable, but it is not registered in any Nest module, and `ProjectMemberApiKey` is not included in any `TypeOrmModule.forFeature(...)`. Any future controller or use case that injects `ProjectMemberApiKeyService` will fail at startup with a Nest DI resolution error. Please add a dedicated module or wire the entity, service, crypto service, and mapper into an existing module and export the service where it will be consumed.

🤖 Reviewed by Codex (GPT-5.5)

romandubovyi (approved):

PR #1242, opened 2026-05-27
fix(backend): Biz Owner no longer downgrades Tech User access

laskevych:
## Summary

- Fixes a Data Mart permissions bug where a Technical User assigned as Business Owner of a DM that was *Available for maintenance* could only **See** and **Use** the DM, losing the **Edit**, **Delete**, and **Manage Triggers** actions they would have had as a non-owner Technical User on the same DM. Ownership is now additive: it can only grant access, never reduce it.
- Restructures `AccessDecisionService.canAccess()` to evaluate two independent paths and union them: an **ownership floor** (matrix lookup at the user's actual owner status, bypasses the context gate) and a **non-owner sharing path** (matrix lookup at `NON_OWNER`, subject to role scope / context gate even when the user is an owner of the resource). The access matrix now encodes only the ownership floor for Business Owner; maintenance permissions for a Business Owner who is a Technical User come from the non-owner sharing path at decision time.
- Documents the underlying permission model in `docs/project/ownership-and-availability.md` — multiple access paths combine by union, subject to role, availability, role-scope/context, and owner-only/admin-only gates. The DM and DM Trigger tables now annotate which actions are granted through the non-owner sharing path and therefore subject to the context gate.

## Background

Reported via a real user: a Technical User (`Entire project` scope) tagged as Business Owner of a Data Mart that was `Available for maintenance = true` could not edit the Data Mart. The matrix collapsed `BIZ_OWNER + EDITOR` to `See, Use` regardless of sharing state, while the same user as a non-owner would have received full maintenance actions. The Product Manager confirmed this is a bug — owner-side actions must be a superset of, not a downgrade of, non-owner actions.

A first iteration of this fix baked the union directly into the matrix rows for `BIZ_OWNER + EDITOR`. CTO review caught the regression: because `canAccess()` skipped the context gate for any owner status, a Technical User with `Selected contexts only` scope and no context overlap could still gain Edit / Delete / Manage Triggers on a Data Mart they were Business Owner of. That violated the union model — maintenance should come from the non-owner Technical User path and remain gated by role scope / contexts.

The current implementation evaluates the two paths separately so the context gate applies to maintenance permissions even for owners.

### `apps/backend/src/data-marts/services/access-decision/access-matrix.config.ts`

- Restores `dmBizOwnerEditorRules` to a single rule: See + Use only across all sharing states. The matrix now expresses only the ownership floor for Business Owner; the union with the non-owner sharing path is computed in `canAccess()` so the context gate can apply correctly.

### `apps/backend/src/data-marts/services/access-decision/access-decision.service.ts`

- Replaces the single matrix lookup + post-lookup context gate with a two-path evaluation:
  1. **Ownership path** — for users with `ownerStatus !== NON_OWNER`, look up the matrix at the actual owner status. If allowed, return true without checking contexts.
  2. **Non-owner sharing path** — look up the matrix at `NON_OWNER`. If allowed, apply the context gate (`Selected contexts only` scope requires a context overlap). This now runs for both pure non-owners and owners whose ownership floor did not grant the action, so a Business Owner's maintenance permissions flow through here and are properly gated.

### Tests

- `access-decision.service.spec.ts` — parametrized matrix test now computes the expected result as `ownership_path || non_owner_path` under the default `Entire Project` mock, because the matrix is no longer a 1:1 spec of `canAccess()` output once the union is computed in code.
- `access-decision-context.spec.ts` — three new cases (12a, 12b, 12c) covering Business Owner under `entire_project` (Edit allowed), `selected_contexts` with no overlap (Edit denied — the regression CTO requested), and `selected_contexts` with no overlap on See (allowed via ownership floor).
- `access-decision-e2e.spec.ts` — adds end-to-end scenarios driving the real matrix: Business Owner + Selected contexts + no overlap denies maintenance but allows See; the same with overlap allows Edit; plus existing scenarios for `entire_project` and Business User as Business Owner (unchanged See / Use only).

### Documentation

- `docs/project/ownership-and-availability.md` — new "Permission Model" section explicitly describes the two paths, ownership floor (bypasses context gate), and non-owner sharing path (subject to context gate). Data Mart and Data Mart Trigger access tables are split by user role for Business Owner rows, with footnotes (†, ‡) noting which actions require a context overlap under `Selected contexts only` scope.

## Production deployment notes

This change expands access for existing rows: any Technical User who is currently a Business Owner of a Data Mart with `available_for_maintenance = true` will gain Edit / Delete / Manage Triggers immediately after deploy, **subject to their role scope and context assignments**. Through `canAccessReport`, they will also gain Edit / Delete / Run on all Reports of that Data Mart (Product confirmed this cascade is the intended scope).


max-voloshyn (commented):
This fixes the `Entire project` happy path, but I think the current implementation over-grants for `Selected contexts only` users.

The new `EDITOR + BIZ_OWNER + SHARED_FOR_MAINTENANCE/BOTH` matrix rows grant `EDIT`, `DELETE`, and `MANAGE_TRIGGERS` directly. `AccessDecisionService.canAccess()` skips the context gate for any owner status, so a Technical User scoped to selected contexts with no context overlap can still get maintenance access if they are Business Owner and the Data Mart is available for maintenance.

That seems to contradict the intended union model: Business Owner should grant the `SEE`/`USE` floor, while maintenance should come from the shared Technical User path and remain subject to role scope/context gates. I’d either evaluate those paths separately or apply the context gate before granting maintenance through the Business Owner + Technical User branch.

Please add regression coverage for:
- `editor + BIZ_OWNER + selected_contexts + no context overlap + availableForMaintenance=true + EDIT => false`
- `editor + BIZ_OWNER + entire_project + availableForMaintenance=true + EDIT => true`

🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (commented):
[P2] Report access table still excludes the new additive maintenance path

`docs/project/ownership-and-availability.md` still defines Data Mart maintenance access as “Technical Owner with Technical User role, or non-owner Technical User with DM available for maintenance,” and then says “Business Owner of parent Data Mart” cannot edit/delete/run reports. That contradicts the new union model in this PR: a Business Owner who is also a Technical User can receive Data Mart maintenance through the non-owner sharing path when the Data Mart is available for maintenance, and that maintenance access cascades to Report edit/delete/run.

Please update this section to describe the general principle instead of excluding owners from the sharing path. For example, “Has Data Mart maintenance access” can cover any Technical User who receives maintenance via ownership or the non-owner sharing path, and the Business Owner row should be limited to Business Owners without Data Mart maintenance access or removed.

🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (approved):

PR #1248, opened 2026-05-28
feat: HTTP Data API NDJSON streaming endpoint

kalnik-a-a:
## Summary

Adds **HTTP Data API** — a script-friendly endpoint that streams a published Data Mart's rows over HTTP as NDJSON, authenticated with the caller's ODM member token. No persisted `Report` / `DataDestination` is created; each pull is recorded as an `HTTP_DATA` Data Mart run and counted for consumption.

```
GET /api/external/http-data/data-marts/{dataMartId}.ndjson?column=date&column=Revenue%20Total
x-owox-authorization: <odm-member-token>
```

- **Auth/authz** — `@Auth(Role.viewer())` + `Action.USE` on the requested Data Mart. Access follows the platform's **authorized-view model** (same as Reports, Looker, and blend configuration): `USE` on the root mart grants read of its whole composed blend, because a blend is an `EDIT`-gated artifact on both sides — contributing/joined marts are not separately re-authorized. Published-only; cross-project / DRAFT → `404`.
- **Response** — pure NDJSON (one data-row object per line, no envelope). The `200 application/x-ndjson` response is committed only after the first storage batch reads successfully, so storage errors (permissions, invalid SQL, missing table) surface as a proper error status instead of a truncated `200`. `runId` is returned in the `x-owox-run-id` header; requested columns in `x-owox-columns` (base64url-encoded JSON). Final status is read back from run history.
- **Column selection** — required repeated `column`; optional base64url-encoded (URL-safe) `filter`/`sort` (same shape as Reports) and integer `limit`. `pageToken`/`offset` are rejected.
- **Schema freshness** — the Data Mart schema is actualized-if-expired (and persisted, 30-min TTL) before column/blending validation, so requests validate against the live storage schema instead of a stale snapshot — mirroring the existing AI-insights read paths.
- **Storage** — any storage with a ready `DataStorageReportReader` (BigQuery first-class); resolver-gated.
- **Run persistence** — the run row is written once at the end in its terminal status: `SUCCESS` (always, + consumption) or `FAILED` (when a started stream errors). There is no `RUNNING` phase, hence no orphaned runs to recover and no lifecycle/cleanup hook; `DataMartRunService` owns the persistence. The `x-owox-run-id` header uses a pre-generated UUID that becomes the row id.
- **Stream lifecycle hardening** — backpressure (`write`→`drain`) where the request-timeout / client-disconnect / shutdown `AbortController` also interrupts a stalled drain wait (a disconnect or socket error during backpressure fails the run — never a false `SUCCESS`); per-row abort checks; hard row/byte caps enforced before each write (a row that would exceed the byte cap is not written); no unhandled rejection if abort wins the read race.
- **Reader interface** — `prepareReportData`/`resolveBlendingDecision`/`compose` generalized from `Report` to a `ReportLike` union so HTTP Data reuses report planning without a persisted `Report`; non-BigQuery readers fail closed on parameterized `sqlOverride`.
- **Run metadata** — `additionalParams.httpData` stores `{ format, columns, filter?, sort?, limit?, dataDescription.dataHeaders, rowCount, bytesWritten, completed }`; surfaced (whitelisted) in run history and rendered in the web run-history view.

## Known limitations (follow-ups)

- The request timeout and the client-disconnect / shutdown aborts take effect between storage batches; an in-flight storage query is not itself cancelled (same as the Reports run path). Truly cancelling mid-query would need an `AbortSignal` on the storage reader interface.

## Test plan

- [x] `npm run lint` (all workspaces)
- [x] `@owox/backend` unit tests
- [x] `@owox/backend` e2e (`http-data`)
- [x] `@owox/web` tests
- [x] `@owox/backend` + `@owox/web` builds
- [x] Manual browser pass — real HTTP_DATA run (sort + limit) streamed 200 NDJSON; run history renders the icon/label and the restructured Configuration view shows `httpData` params incl. `dataDescription`, capped `limit`, `rowCount`, `bytesWritten`, `completed`


apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -0,0 +1,409 @@
...
+    const effectiveLimit =
+      query.limit == null ? undefined : Math.min(query.limit, HTTP_DATA_MAX_ROWS);
+    const ctx: AuthorizationContext = {
+      userId: command.userId,
+      projectId: command.projectId,
+      roles: command.roles,
+    };
+
+    const dataMart = await this.loadAccessibleDataMart(command.dataMartId, ctx);
+    await this.dataMartService.actualizeSchemaInEntity(dataMart);
```
max-voloshyn: This only mutates the in-memory `dataMart`, but the next validation paths call `computeBlendableSchema(dataMart.id, dataMart.projectId)`, which reloads the mart from the repository. Since the actualized schema is never saved here, HTTP Data can still validate against stale schema: newly added columns may be rejected, and removed columns may be allowed and emitted as null/placeholders. Persist the actualized mart before these validators run, or make the validators use the already-actualized entity.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -0,0 +1,409 @@
...
+
+      reader = await this.readerResolver.resolve(dataMart.storage.type);
+      const description = await reader.prepareReportData(readPlan, {
+        sqlOverride,
+        sqlOverrideParams,
+        columnFilter: decision.columnFilter,
+        blendedDataHeaders: decision.blendedDataHeaders,
+      });
+
+      this.streamWriter.initHeaders(res, { runId: run.getId(), columns: query.columns });
```
max-voloshyn: Headers are flushed before the first batch is read. For BigQuery/Athena/Redshift, query execution happens inside `readReportDataBatch()`, so storage permission errors, invalid generated SQL, missing tables, etc. happen after the response is already committed as `200 application/x-ndjson`; `handleStreamFailure()` can only destroy the socket instead of returning a proper error response. Please defer `initHeaders()` until after the first batch/query initialization succeeds, then stream that buffered first batch.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/http-data/http-data-stream-writer.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  private waitForDrain(res: Response): Promise<void> {
+    return new Promise<void>(resolve => {
+      const cleanup = () => {
+        res.off('drain', onDone);
+        res.off('close', onDone);
+        res.off('error', onDone);
+      };
+      const onDone = () => {
```
max-voloshyn: `close` and `error` are treated the same as `drain`, so `writeRow()` resolves successfully even when the client disconnected or the socket failed while backpressure was pending. If that happens on the last row, `streamRows()` returns success and the run is marked `SUCCESS` even though the client did not receive a complete stream; `error` is not connected to the abort signal at all. `close`/`error` should reject or otherwise propagate cancellation so the run is finalized as failed.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -0,0 +1,412 @@
...
+          if (bytesWritten >= HTTP_DATA_MAX_BYTES) {
+            throw new Error(`Byte limit reached (${HTTP_DATA_MAX_BYTES})`);
+          }
+
+          const obj: Record<string, unknown> = {};
+          for (let i = 0; i < requestedColumns.length; i++) {
+            const idx = fieldIndexMap[i];
+            obj[requestedColumns[i]] = idx === -1 ? null : row[idx];
+          }
+          bytesWritten += await this.streamWriter.writeRow(res, obj);
```
max-voloshyn: `HTTP_DATA_REQUEST_TIMEOUT_MS` is not enforced while `writeRow()` is waiting for backpressure to drain. A slow client can make `res.write()` return `false` and then keep the socket open without producing `drain`; the timeout aborts the controller, but `waitForDrain()` does not observe that signal, so the request can hang past the configured timeout while holding the reader open. Pass the `AbortSignal` into `writeRow()`/`waitForDrain()` or race the write wait against the same abort signal.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -0,0 +1,412 @@
...
+          this.streamWriter.initHeaders(res, { runId, columns: requestedColumns });
+          headersSent = true;
+        }
+
+        for (const row of batch.dataRows) {
+          this.throwIfAborted(abortController.signal);
+          if (rowCount >= HTTP_DATA_MAX_ROWS) {
+            throw new Error(`Row limit reached (${HTTP_DATA_MAX_ROWS})`);
+          }
+          if (bytesWritten >= HTTP_DATA_MAX_BYTES) {
```
max-voloshyn: The byte cap is checked only before writing the current row, so the response can exceed `HTTP_DATA_MAX_BYTES` by the full size of one row. With `bytesWritten` just below the limit, a very large JSON row is still serialized and written, and only the next iteration notices the cap. Compute the row chunk size before writing and reject when `bytesWritten + chunk.length` would exceed the limit.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -0,0 +1,412 @@
...
+    try {
+      await this.dataMartRunService.recordHttpDataRun({
+        runId,
+        dataMart,
+        createdById,
+        startedAt,
+        status: DataMartRunStatus.SUCCESS,
+        metadata,
+      });
+    } catch (err) {
```
max-voloshyn, line 192: If the final `recordHttpDataRun()` insert fails, this logs and still lets the HTTP response complete successfully with an `x-owox-run-id` that has no run-history row and no consumption event. That breaks the endpoint contract that every pull is recorded/countable, and clients cannot tell the trace id is unusable. Consider creating the audit row before streaming starts, or otherwise fail before sending data when the run id cannot be persisted.

🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn: Discussed on the daily: recording the HTTP Data run at the very end is an intentional trade-off for this streaming endpoint. Yes, there is a risk that the final run record is not persisted after a successful stream, but recording before streaming creates the opposite failure mode where interrupted requests can leave pending/running records hanging and needing recovery. We accept the current approach as the preferred trade-off here, with the persistence failure logged for operators. No change requested from this thread.

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
kalnik-a-a: Good catch — fixed. `writeChunk()` now receives the request `AbortSignal`, and `waitForDrain()` races `drain` against `close`/`error` **and** the abort signal (rejecting immediately if it is already aborted). So the request timeout / client-disconnect / shutdown now interrupt a stalled backpressure wait instead of holding the reader open past the timeout. Thanks!

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
kalnik-a-a: Fixed — the row is now serialized first and the cap is checked as `bytesWritten + chunk.length > HTTP_DATA_MAX_BYTES` before writing, so a row that would exceed the limit is rejected up front rather than written and only noticed on the next iteration. Thanks!

max-voloshyn (approved):

PR #1253, opened 2026-05-29
fix(backend): prevent OOM on large BigQuery reports by streaming query results

laskevych:
## Problem

Reports backed by a large BigQuery **VIEW** or SQL query (millions of rows)
crashed the worker with an out-of-memory error. The user-visible symptoms:

- The report sat in `RUNNING` for many minutes, then the worker process died
  with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap
  out of memory`.
- Because the process was OOM-killed mid-flight, `markAsUnsuccessful` never
  ran — the report stayed stuck in `RUNNING` indefinitely instead of failing
  cleanly.
- **No data ever reached the destination.** With a `TABLE`-backed data mart
  you could watch rows append to Google Sheets batch by batch; with a `VIEW`
  nothing appeared, because the crash happened before the first batch was
  written.

This reproduced on a clean checkout with **no recent Google Sheets changes**,
which ruled out the diff-based writer and pointed at the read path.

## Root cause

`BigQueryApiAdapter.executeQuery` ran the query via `bigQuery.query()`:

```ts
const [, , res] = await this.bigQuery.query(queryConfig); // maxResults: 0
return { jobId: res.jobReference.jobId };
```

`bigQuery.query()` calls `getQueryResults` under the hood, which **buffers the
entire result set into the process heap** before resolving — even though we
only need the job id and never read the returned rows here (the reader
paginates the job's destination table separately). For a multi-million-row
result that buffer alone exhausts the worker heap.

The crash stack confirmed it: `ArrayMap` → `PromiseFulfillReactionJob` →
`CompressionStream<ZlibContext>::AfterThreadPoolWork` — i.e. gzipped query
results being decompressed, JSON-parsed and `.map()`-ed into memory inside the
BigQuery client.

## Fix

Run the query as a job and **wait for completion without fetching any rows**:

```ts
const [job] = await this.bigQuery.createQueryJob(queryConfig);
await this.waitForJobToComplete(job); // polls jobs.get — no row data
return { jobId: job.id };
```

`waitForJobToComplete` polls the job's metadata (`jobs.get`, status only) every
2s until the job reaches `DONE`, and throws if the job finished with an error.
Memory stays flat regardless of result size. The actual row streaming is
unchanged — it still happens later, page by page (`maxResults: 5000`,
`autoPaginate: false`), in the report reader / SQL-run executor against the
job's anonymous destination table.

This mirrors the pattern already used by `executeDryRunQuery` (`createQueryJob`)
and `bigquery-sql-run.executor` (`job.promise()` to await completion without
fetching rows), so it's consistent with existing code in this module.

## Scope / blast radius

`executeQuery` has four callers; the contract (`(query, params?) → { jobId }`,
job finished before returning, throws on bad query) is preserved for all:

| Caller | Path | Verified |
| --- | --- | --- |
| `bigquery-report-reader` | SELECT → destination table → rows (GS export & all BQ reports) | new integration test + full report integration suite |
| `bigquery-sql-run.executor` | ad-hoc Run / preview | smoke checklist |
| `bigquery-create-view.executor` (×2) | DDL `CREATE VIEW` / `CREATE SCHEMA` | already exercised by this suite's `beforeAll`/`afterAll` |

`checkAccess` (`SELECT 1`) is intentionally left on `bigQuery.query()` — tiny
result, no memory concern.

Behaviour is identical for SQL-query and VIEW-backed data marts; both arrive as
a query string and only need the job to finish so the destination table is
materialized.

## Tests

- **New integration tests** (`bigquery.integration.ts`, real BigQuery, gated by
  `BQ_*` credentials, skip cleanly without them) locking the contracts the
  rewrite changed:
  1. SELECT → job → destination table → `getRows` returns the expected rows
     (also exercises autodetect-location, since the suite runs with
     `BIGQUERY_AUTODETECT_LOCATION`).
  2. NAMED query parameters end-to-end (used by output controls / blended SQL).
  3. Invalid SQL rejects (error now surfaces from job status, not from
     `bigQuery.query()`).
- DDL execution is already covered by the suite's setup/teardown.
- The end-to-end report path is covered by
  `google-sheets-column-preservation.integration.ts`.
- `tsc --noEmit`, ESLint, Prettier: clean on changed files.
- **Manual production smoke (all green):** large VIEW (1M+ rows) → GS export
  streams with flat memory and finishes `SUCCESS`; normal SQL data mart;
  report with filter/sort/limit (NAMED params); blended data mart; VIEW
  creation (DDL); ad-hoc Run; invalid SQL → clean `ERROR` (no stuck
  `RUNNING`); empty result set.

## Why no unit test

The adapter constructs its `BigQuery` client internally, so a unit test would
require mocking `@google-cloud/bigquery` — and the behaviour this change
depends on (does `createQueryJob().id` resolve via `getJob`, is the
destination table materialized after `DONE`, what shape is `errorResult`) is
exactly the SDK behaviour a mock cannot verify. Real-BigQuery integration is
the only meaningful net here.

## Risk / notes

- The poll loop has no explicit timeout — it polls until the job reports `DONE`
  or an error. BigQuery jobs always terminate, and run-level cancellation is
  handled upstream via `AbortSignal`, so there is no new "hang forever" state.
  Wall-clock is unchanged vs. the old code, which also blocked until the job
  finished — only the buffering is gone.
- Polling interval (2s) keeps `jobs.get` call volume low on long jobs while
  adding negligible latency to fast ones (a fast job is already `DONE` on the
  first poll).

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
@@ -14,6 +14,13 @@ import type { SqlParameter } from '../../utils/sql-clause-renderer';
  * query execution, dry-run, and job management.
  */
 export class BigQueryApiAdapter {
+  /**
+   * Interval between query-job status polls in {@link waitForJobToComplete}.
+   * Large enough to keep the `jobs.get` call rate low on long-running jobs,
+   * small enough that fast jobs add negligible latency.
+   */
+  private static readonly JOB_POLL_INTERVAL_MS = 2000;
+
   private readonly logger = new Logger(BigQueryApiAdapter.name);
   private readonly bigQuery: BigQuery;
   private readonly authClient: JWT | OAuth2Client;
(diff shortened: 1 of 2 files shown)


max-voloshyn:
Non-blocking: this regression coverage only runs in the manual integration workflow, so PR CI will not exercise the new polling/error behavior. A small unit spec with mocked `createQueryJob`/`getMetadata` would make this easier to protect in normal PR checks.

🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (approved):

PR #1254, opened 2026-05-29
feat: HTTP Data API — optional column selectors (*/**) and no streaming caps

kalnik-a-a:
# HTTP Data API — optional `column` with `*`/`**` selectors, no hard caps

Follow-up to the HTTP Data API NDJSON streaming endpoint. Removes the artificial streaming
limits and makes the `column` query parameter optional with two reserved expansion values,
while keeping column exposure aligned with the reporting column picker.

### Removed limits

- **`HTTP_DATA_MAX_ROWS`**, **`HTTP_DATA_MAX_BYTES`**, **`HTTP_DATA_REQUEST_TIMEOUT_MS`** — the
  hard streaming caps and the request timeout are gone. Streams now run until the data is
  exhausted, the client disconnects, the response errors, or the server enters shutdown
  (the `AbortController` still guards a stalled backpressure drain).
- **100-column cap** (`HTTP_DATA_MAX_COLUMNS`) — removed.
- **`limit` upper bound** (`HTTP_DATA_MAX_USER_LIMIT`, 10,000,000) — removed; `limit` is now any
  positive integer.
- `bytesWritten`/`rowCount` are still tracked and recorded in the run metadata.

### Optional `column` with reserved selectors

- `column` is now optional. Two reserved values expand to column sets:
  - `*` — all native columns.
  - `**` — all native columns plus all reporting-visible blended columns.
- `*` can be combined with explicit columns (e.g. `column=*&column=orders__revenue` = all native
  plus that blended field). Omitting `column` is equivalent to `*`. Overlaps are de-duplicated.
- The previous hard rejection of duplicate `column` values is replaced by de-duplication, since
  `*` plus explicit columns legitimately overlap.

### Reporting-visibility is enforced (not just for `**`)

Column exposure now mirrors the reporting column picker, derived entirely from the existing
`BlendableSchemaDto` flags — both for sentinel expansion and for validating explicit selections:

- **Native**: fields flagged `isHiddenForReporting` (at every nesting level) and fields with
  `status === DISCONNECTED` are excluded — matching the report header generators.
- **Blended**: fields flagged `isHidden`, and any field belonging to an excluded source
  (`availableSources` with `isIncluded === false`), are excluded.

As a result, hidden/excluded columns are rejected with a `400` even when requested explicitly —
previously the validator accepted any blended field name.

## Implementation notes

- New `http-data-column-sets.util.ts` exposes `nativeColumnNames(schema)` and
  `visibleBlendedColumnNames(schema)` — the single source of truth shared by the resolver and the
  validator. The blended predicate must stay in sync with the web `ReportColumnPicker`.
- New `HttpDataColumnResolver` turns a parsed `ColumnSelector` (`allNative` / `allBlendable` /
  `explicit`) into a concrete, de-duplicated column list; an empty result is a `400`.
- `HttpDataColumnValidator` no longer recomputes the blendable schema — it takes the precomputed
  `BlendableSchemaDto` (computed once per request in the use-case) and validates against the
  reporting-visible column universe.
- The resolved concrete column list drives blending detection, the run metadata, and the NDJSON
  row keys.
- The `x-owox-columns` response header is removed: with the column cap gone, an unbounded selector
  could push that header past proxy/client header-size limits. Clients infer columns from the row
  keys; the resolved column list is still recorded in the run metadata. `x-owox-run-id` stays.

## Tests

- Updated `http-data-request-validator` specs: `column` optional, `*`/`**` parsing, `*` + explicit
  mixing, `**` dominance; removed the now-invalid "missing/empty/duplicate column" cases.
- New `http-data-column-resolver` specs: `*`/`**` expansion, nested-native flattening, hidden-native
  exclusion, blended visibility, mixing, de-duplication, empty-result `400`.
- Updated `http-data-column-validator` specs: reporting-visible validation; explicit hidden-native,
  hidden-blended, and excluded-source columns are rejected.
- `stream-http-data` specs: dropped the env-driven row/byte-cap tests; added an all-blendable
  selector wiring test.

## Swagger

`column` is documented as optional with the `*`/`**` semantics and mixing rules; the `limit`
description drops the upper bound; the operation summary no longer claims columns must be
explicitly selected.

### DIFF
--- a/apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts
+++ b/apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts
@@ -123,6 +123,13 @@ export const HttpDataQuerySchema = z
         });
       }
     }
+    if (value.column.includes(ALL_BLENDABLE_COLUMNS) && value.column.length > 1) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        path: ['column'],
+        message: `"${ALL_BLENDABLE_COLUMNS}" cannot be combined with other column values`,
+      });
+    }
   })
   .transform(value => ({
     columnSelector: toColumnSelector(value.column),
--- a/apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts
+++ b/apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts
@@ -2,11 +2,28 @@ import { z } from 'zod';
 import { FilterConfigSchema } from './filter-config.schema';
 import { SortConfigSchema } from './sort-config.schema';
 
-export const HTTP_DATA_MAX_COLUMNS = 100;
-export const HTTP_DATA_MAX_USER_LIMIT = 10_000_000;
 export const HTTP_DATA_MAX_ENCODED_PARAM_LENGTH = 8192;
 
+export const ALL_NATIVE_COLUMNS = '*';
+export const ALL_BLENDABLE_COLUMNS = '**';
+
+export type ColumnSelector =
+  | { mode: 'allNative'; explicit: string[] }
+  | { mode: 'allBlendable' }
+  | { mode: 'explicit'; explicit: string[] };
+
 const REJECTED_PAGINATION_KEYS = ['pageToken', 'offset'] as const;
+
+function toColumnSelector(tokens: string[]): ColumnSelector {
+  if (tokens.includes(ALL_BLENDABLE_COLUMNS)) {
+    return { mode: 'allBlendable' };
+  }
+  const explicit = tokens.filter(token => token !== ALL_NATIVE_COLUMNS);
+  if (tokens.includes(ALL_NATIVE_COLUMNS) || tokens.length === 0) {
+    return { mode: 'allNative', explicit };
+  }
+  return { mode: 'explicit', explicit };
+}
 
 function decodeBase64UrlJson(raw: string): unknown {
   let decoded: string;
@@ -77,21 +94,16 @@ const SortParamSchema = z
 const LimitParamSchema = z.coerce
   .number({ invalid_type_error: 'limit must be an integer' })
   .int('limit must be an integer')
-  .min(1, 'limit must be ≥ 1')
-  .max(HTTP_DATA_MAX_USER_LIMIT, `limit must be ≤ ${HTTP_DATA_MAX_USER_LIMIT}`);
+  .min(1, 'limit must be ≥ 1');
 
 const ColumnsParamSchema = z
   .union([z.string(), z.array(z.string())])
-  .transform(value => (Array.isArray(value) ? value : [value]))
-  .pipe(
-    z
-      .array(z.string().min(1, 'column value must not be empty'))
-      .min(1, 'at least one column is required')
-      .max(HTTP_DATA_MAX_COLUMNS, `at most ${HTTP_DATA_MAX_COLUMNS} columns are allowed`)
-      .refine(arr => new Set(arr).size === arr.length, {
-        message: 'duplicate column names are not allowed',
-      })
-  );
+  .optional()
+  .transform(value => {
+    if (value === undefined) return [];
+    return Array.isArray(value) ? value : [value];
+  })
+  .pipe(z.array(z.string().min(1, 'column value must not be empty')));
 
 export const HttpDataQuerySchema = z
   .object({
@@ -113,7 +125,7 @@ export const HttpDataQuerySchema = z
     }
   })
   .transform(value => ({
-    columns: value.column,
+    columnSelector: toColumnSelector(value.column),
     filter: value.filter,
     sort: value.sort,
     limit: value.limit,
--- a/apps/backend/src/data-marts/services/http-data/http-data-column-sets.util.ts
+++ b/apps/backend/src/data-marts/services/http-data/http-data-column-sets.util.ts
@@ -1,0 +1,39 @@
+import { BlendableSchemaDto } from '../../dto/domain/blendable-schema.dto';
+
+export interface ReportingColumns {
+  native: string[];
+  blended: string[];
+}
+
+interface NativeSchemaField {
+  name: string;
+  isHiddenForReporting?: boolean;
+  fields?: NativeSchemaField[];
+}
+
+export function nativeColumnNames(schema: BlendableSchemaDto): string[] {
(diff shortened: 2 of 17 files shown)


apps/backend/src/data-marts/services/http-data/http-data-request-validator.service.spec.ts
```diff
@@ -4,32 +4,43 @@ import { HttpDataRequestValidator } from './http-data-request-validator.service'
...
-    expect(result.columns).toEqual(['date', 'Revenue Total', 'user_id']);
+    expect(result.columnSelector).toEqual({
+      mode: 'explicit',
+      explicit: ['date', 'Revenue Total', 'user_id'],
+    });
   });
 
-  it('rejects empty column list', () => {
-    expect(() => validator.validate({ column: [] })).toThrow(BusinessViolationException);
+  it('treats a missing column param as all-native', () => {
```
max-voloshyn, line 20: This updates unit coverage for missing `column`, but `apps/backend/test/http-data.e2e-spec.ts` still expects 400 for both missing `column` and duplicate columns. With this schema, missing `column` resolves to all native columns, and duplicate explicit columns are de-duped by `HttpDataColumnResolver`, so those e2e assertions will fail. Please update or remove the stale e2e cases.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/http-data/http-data-column-sets.util.ts
```diff
@@ -0,0 +1,39 @@
...
+  fields?: NativeSchemaField[];
+}
+
+export function nativeColumnNames(schema: BlendableSchemaDto): string[] {
+  return collectNative(schema.nativeFields as unknown as NativeSchemaField[]);
+}
+
+function collectNative(fields: NativeSchemaField[], prefix = ''): string[] {
+  const result: string[] = [];
+  for (const field of fields) {
```
max-voloshyn, line 22: `collectNative` filters hidden fields but not fields with `status === DISCONNECTED`. `column=*` or an omitted `column` can therefore select disconnected schema fields even though the report header generators skip them; with output controls the generated SELECT can reference dead columns, and without controls the stream emits placeholder/null columns. Please filter disconnected native fields before expanding `*`.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts
```diff
@@ -2,12 +2,29 @@ import { z } from 'zod';
...
+
+export type ColumnSelector =
+  | { mode: 'allNative'; explicit: string[] }
+  | { mode: 'allBlendable' }
+  | { mode: 'explicit'; explicit: string[] };
+
 const REJECTED_PAGINATION_KEYS = ['pageToken', 'offset'] as const;
 
+function toColumnSelector(tokens: string[]): ColumnSelector {
+  if (tokens.includes(ALL_BLENDABLE_COLUMNS)) {
```
max-voloshyn, line 18: Returning `allBlendable` as soon as `**` is present drops every other `column` token before `HttpDataColumnValidator` sees it. Requests like `column=**&column=orders__secret` or `column=**&column=typo` now succeed instead of rejecting the hidden/unknown column, even though the docs say only reporting-visible columns are selectable and only `*` may be combined. Please reject mixed `**` selectors or preserve and validate the extra tokens.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts
```diff
@@ -77,21 +94,16 @@ const SortParamSchema = z
...
-  .pipe(
-    z
-      .array(z.string().min(1, 'column value must not be empty'))
-      .min(1, 'at least one column is required')
-      .max(HTTP_DATA_MAX_COLUMNS, `at most ${HTTP_DATA_MAX_COLUMNS} columns are allowed`)
-      .refine(arr => new Set(arr).size === arr.length, {
-        message: 'duplicate column names are not allowed',
-      })
-  );
+  .optional()
```
max-voloshyn, line 101: Removing the column-count cap makes explicit lists and expanded `*`/`**` selections unbounded, but `HttpDataStreamWriter` still serializes the full expanded list into the `x-owox-columns` response header. Wide marts can exceed proxy/client header limits before any row is streamed. Please keep a bounded column/header size or stop putting the full expanded column list in a response header for unbounded selectors.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/http-data/http-data-request-validator.service.spec.ts
kalnik-a-a: Good catch, thanks — addressed in 29ccb20. Updated `http-data.e2e-spec.ts`: removed the stale `missing column → 400` and `duplicate columns → 400` cases, and added coverage for omitted / `*` / de-duplicated columns (now 200) and `**` combined with other columns (400).

apps/backend/src/data-marts/services/http-data/http-data-column-sets.util.ts
kalnik-a-a: Thanks — fixed in 29ccb20. `collectNative` now also skips `status === DISCONNECTED`, matching the report header generators and the blended `flattenSchemaFields` path, so `*`/`**` no longer expand disconnected fields. Added a unit-spec case.

apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts
kalnik-a-a: Thanks — fixed in 29ccb20. A `**` selector combined with any other `column` token is now rejected with 400 (in the schema `superRefine`) instead of silently dropping the extras. Covered by both unit and e2e tests.

apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts
kalnik-a-a: Agreed, thanks — fixed in 29ccb20. Removed the `x-owox-columns` response header entirely, since an unbounded selector could push it past proxy/client header-size limits once the column cap was lifted. The resolved column list is still recorded in the run metadata, and `x-owox-run-id` is unchanged.

max-voloshyn (approved):

PR #1260, opened 2026-05-30
Expose Data Mart data traversal with column controls in API client and owox-ctl

max-voloshyn:
## Summary

Adds API client and CLI support for reading rows from the existing HTTP Data API, with explicit column selection semantics for Data Mart output columns, report-available columns, and exact column names.

## Changes

- Introduces `dataMarts.traverseData()` in `@owox/api-client` for incremental NDJSON row traversal.
- Adds `DataMartDataTraversal.cancel()` so callers can close an opened traversal without iterating it.
- Adds `owox-ctl data-marts stream <dataMartId>` for writing streamed rows as newline-delimited JSON.
- Aligns `owox-ctl data-marts stream` with the automation-first JSON-only CLI surface introduced by #1266.
- Introduces the HTTP Data `columns` selector separately from repeated exact `column` values:
  - `columns=*` / `--columns '*'` selects Data Mart output columns.
  - `columns=**` / `--columns '**'` selects report-available columns, including joined fields.
  - repeated `column` values / `--column` select exact column names, so literal `*` and `**` remain addressable.
- Supports filter, sort, and limit controls through the client and CLI traversal path.
- Preserves stdout backpressure handling for CLI streaming.
- Aligns API client stream requests with authenticated JSON requests.
- Preserves Data Mart context on stream-open errors from the shared API transport.
- Maps storage-provider read failures to `424 Failed Dependency`, keeping OWOX authorization failures separate from third-party storage permission/read failures.
- Preserves app `HttpException`s before forced BigQuery storage-read mapping.
- Keeps storage-specific error parsing in storage-specific mappers, with a minimal shared helper only for building storage-read HTTP responses.
- Updates docs and release notes for NDJSON output and `columns=*` / `columns=**` selector semantics.

## Verification

- `npm run test -w @owox/backend -- storage-read-error.spec.ts bigquery-storage-error.mapper.spec.ts stream-http-data.service.spec.ts --runInBand`
- `npm run test -w @owox/ctl -- base-command.test.ts data-marts/stream.test.ts --runInBand`
- `npm run test -w @owox/api-client -- client.test.ts data-marts.test.ts --runInBand`
- `npm run test:e2e -w @owox/backend -- http-data.e2e-spec.ts --runInBand`
- `npm run typecheck -w @owox/api-client`
- `npm run typecheck -w @owox/ctl`
- `npm run lint -w @owox/api-client`
- `npm run lint -w @owox/ctl`
- `git diff --check`

Latest review update also re-ran:

- `npm run test -w @owox/api-client -- data-marts.test.ts --runInBand`
- `npm run test -w @owox/backend -- bigquery-storage-error.mapper.spec.ts --runInBand`
- `npm run typecheck -w @owox/api-client`
- `git diff --check`

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-storage-error.mapper.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-storage-error.mapper.ts
@@ -60,6 +60,8 @@ export function mapBigQueryStorageReadError(
   error: unknown,
   options: StorageReadErrorMappingOptions = {}
 ): unknown {
+  if (error instanceof HttpException) return error;
+
   const context = {
     storageType: type,
     providerName: toHumanReadable(type),
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-storage-error.mapper.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-storage-error.mapper.ts
@@ -1,0 +1,86 @@
+import { HttpException, Injectable } from '@nestjs/common';
+import { DataStorageType, toHumanReadable } from '../../enums/data-storage-type.enum';
+import {
+  createStorageReadError,
+  messageFromError,
+  StorageProviderReadError,
+} from '../../interfaces/storage-read-error';
+import {
+  DataStorageErrorMapper,
+  StorageReadErrorMappingOptions,
+} from '../../interfaces/data-storage-error-mapper.interface';
+
+type BigQueryErrorShape = {
+  errors?: Array<{ reason?: unknown; message?: unknown }>;
+  response?: {
+    status?: { errorResult?: { reason?: unknown; message?: unknown } };
+  };
+};
+
+function asString(value: unknown): string | undefined {
+  return typeof value === 'string' && value.length > 0 ? value : undefined;
+}
+
+function asBigQueryError(error: unknown): BigQueryErrorShape {
+  return typeof error === 'object' && error !== null ? (error as BigQueryErrorShape) : {};
+}
+
+function responseFieldsFor(
+  reason: string | undefined
+): Pick<StorageProviderReadError, 'code' | 'statusCode' | 'statusText'> {
+  if (reason === 'authError') {
+    return { code: 'STORAGE_UNAUTHORIZED', statusCode: 401, statusText: '401 Unauthorized' };
+  }
+  if (reason === 'accessDenied') {
+    return { code: 'STORAGE_PERMISSION_DENIED', statusCode: 403, statusText: '403 Forbidden' };
+  }
+  return { code: 'STORAGE_READ_FAILED' };
+}
+
+function extractBigQueryError(error: unknown): StorageProviderReadError | null {
+  if (error instanceof HttpException) return null;
+
+  const shaped = asBigQueryError(error);
+  const errorResult = shaped.response?.status?.errorResult;
+  const firstError = shaped.errors?.[0];
+  const reason = asString(errorResult?.reason) ?? asString(firstError?.reason);
+  const message = asString(errorResult?.message) ?? asString(firstError?.message);
+
+  if (!reason && !message) return null;
+
+  return {
+    ...responseFieldsFor(reason),
+    message: message ?? messageFromError(error),
+    ...(reason ? { reason } : {}),
+  };
+}
+
+export function mapBigQueryStorageReadError(
+  type: DataStorageType,
+  error: unknown,
+  options: StorageReadErrorMappingOptions = {}
+): unknown {
+  const context = {
+    storageType: type,
+    providerName: toHumanReadable(type),
+  };
+  const providerError = extractBigQueryError(error);
+
+  if (providerError) return createStorageReadError(context, providerError);
+  if (options.force) {
+    return createStorageReadError(context, {
+      code: 'STORAGE_READ_FAILED',
+      message: messageFromError(error),
+    });
+  }
+  return error;
+}
+
+@Injectable()
+export class BigQueryStorageErrorMapper implements DataStorageErrorMapper {
+  readonly type = DataStorageType.GOOGLE_BIGQUERY;
+
+  toStorageReadError(error: unknown, options?: StorageReadErrorMappingOptions): unknown {
+    return mapBigQueryStorageReadError(this.type, error, options);
+  }
+}
--- a/apps/backend/src/data-marts/data-storage-types/interfaces/storage-read-error.ts
+++ b/apps/backend/src/data-marts/data-storage-types/interfaces/storage-read-error.ts
@@ -1,0 +1,47 @@
+import { HttpException, HttpStatus } from '@nestjs/common';
+import { DataStorageType } from '../enums/data-storage-type.enum';
+
+export type StorageReadErrorContext = {
+  storageType: DataStorageType;
+  providerName: string;
+};
+
+export type StorageProviderReadError = {
+  code: string;
+  message: string;
+  reason?: string;
+  statusCode?: number;
+  statusText?: string;
+};
+
+export function messageFromError(error: unknown): string {
+  return error instanceof Error ? error.message : String(error);
+}
+
+export function createStorageReadError(
+  context: StorageReadErrorContext,
+  providerError: StorageProviderReadError
+): HttpException {
+  let providerStatus = ' returned an error';
+  if (providerError.statusText) {
+    providerStatus = ` returned ${providerError.statusText}`;
+  } else if (providerError.statusCode) {
+    providerStatus = ` returned HTTP ${providerError.statusCode}`;
+  }
+
+  return new HttpException(
+    {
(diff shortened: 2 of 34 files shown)


packages/api-client/src/data-marts.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    const client = new OWOXApiClient({
+      apiOrigin,
+      apiKeyId,
+      apiKeySecret,
+      fetchImpl: fetchMock.fetchImpl,
+    });
+
+    const data = await client.dataMarts.traverseData('dm-1');
+
+    await collectRows(data).catch(error => {
```
max-voloshyn: This assertion pattern passes if `collectRows(data)` unexpectedly resolves, because the `.catch()` callback never runs. The network-failure test below has the same issue. Please use `await expect(...).rejects...` so these tests fail when the stream stops rejecting.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -172,20 +181,37 @@ export class StreamHttpDataService {
...
+    reader: DataStorageReportReader | null,
+    dataMart: DataMart,
+    res: Response
+  ): Promise<unknown> {
+    if (res.headersSent || error instanceof StreamCancelledError) {
+      return error;
+    }
+
+    const mapper = await this.errorMapperResolver.resolve(dataMart.storage.type);
+    return mapper.toStorageReadError(error, { force: reader !== null });
```
max-voloshyn: This still lets pre-reader storage failures escape as 500s for Snowflake/Redshift/Athena/Databricks. `actualizeSchemaInEntityIfExpired()` runs before `reader` is assigned, so `force` is false here; those new mappers immediately return the original error when `force` is false. Please either make those mappers recognize provider errors in the schema-refresh path or intentionally mark schema actualization failures as storage dependency errors too.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/data-storage-types/interfaces/storage-read-error.ts
```diff
@@ -0,0 +1,47 @@
...
+      details: {
+        dependency: 'storage',
+        providerMessage: providerError.message,
+        providerName: context.providerName,
+        ...(providerError.reason ? { providerReason: providerError.reason } : {}),
+        ...(providerError.statusCode ? { providerStatusCode: providerError.statusCode } : {}),
+        storageType: context.storageType,
+      },
+    },
+    HttpStatus.FAILED_DEPENDENCY
```
max-voloshyn, line 45: This introduces `424 Failed Dependency` as a public HTTP Data response for storage read failures, but the Swagger spec for the endpoint still documents only 400/401/403/404. Please add the 424 response there so generated OpenAPI clients and docs know this is part of the API contract.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-storage-error.mapper.ts
```diff
@@ -0,0 +1,86 @@
...
+  options: StorageReadErrorMappingOptions = {}
+): unknown {
+  const context = {
+    storageType: type,
+    providerName: toHumanReadable(type),
+  };
+  const providerError = extractBigQueryError(error);
+
+  if (providerError) return createStorageReadError(context, providerError);
+  if (options.force) {
```
kalnik-a-a, line 72: [P1] Preserve app `HttpException`s before forced storage mapping

`extractBigQueryError()` returns `null` for existing `HttpException`s, but the forced fallback below still wraps them into a generic `424 STORAGE_READ_FAILED`. `StreamHttpDataService` now calls the mapper with `force: true` during schema actualization / after reader creation, so app-level errors such as missing OAuth credentials can lose their original status and code.

The non-BigQuery mappers already short-circuit this case with `if (error instanceof HttpException || !options?.force) return error;`. BigQuery should do the same before provider extraction / forced fallback, and the mapper spec should cover `HttpException` with `{ force: true }`.

packages/api-client/src/data-marts.ts
```diff
@@ -67,4 +239,29 @@ export class DataMartsApi {
...
   }
+
+  async traverseData(
+    dataMartId: string,
+    options: TraverseDataOptions = {}
+  ): Promise<DataMartDataTraversal> {
+    const query = buildTraverseDataQuery(options);
+    let response: Response;
+    try {
+      response = await this.requester.getStream(
```
kalnik-a-a, line 261: [P1] Avoid opening a stream that callers cannot close

`traverseData()` opens the HTTP response immediately, but cleanup only exists inside `rowChunks()` after `response.body.getReader()` is acquired. If a caller does:

```ts
const traversal = await client.dataMarts.traverseData(id);
// decides not to iterate
```

there is no public `cancel()` / `close()` and no lazy-open path, so the response body can remain open. Either expose an explicit cleanup method on `DataMartDataTraversal`, or defer `getStream()` until `rowChunks()` starts. Please add a regression test for opening traversal without iterating it.

kalnik-a-a:
Release-note / docs cleanup items from review:

- [P2] `.changeset/http-data-api-ndjson-endpoint.md` still describes the old column selector contract where repeated `column` values include reserved `*` / `**` selectors. The current API uses `columns=*` and `columns=**` as selectors, while repeated `column` values are exact names; `column=*` and `column=**` now mean literal column names. If this ships as-is, release notes will document the wrong public API.

- [P2] `.changeset/add-api-control-cli.md` lists the supported `owox-ctl` commands but omits the new `owox-ctl data-marts stream` command added by this PR. Since this is the release note for the API/CLI access layer, it should mention the stream command too.

- [P3] `docs/api/owox-ctl.md` says “All commands emit JSON”, but `owox-ctl data-marts stream` intentionally writes newline-delimited JSON rows to stdout. Each row is JSON, but the output is not a single JSON value. Suggested wording: “Commands emit JSON unless noted; `data-marts stream` emits NDJSON rows.”

kalnik-a-a (changes requested):

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-storage-error.mapper.ts
max-voloshyn: Addressed in the follow-up changes. `mapBigQueryStorageReadError()` now preserves existing app `HttpException`s before provider extraction or forced fallback, so app-level errors keep their original status/code even with `{ force: true }`. Added mapper spec coverage for `HttpException` with forced storage-read mapping.

packages/api-client/src/data-marts.ts
max-voloshyn: Addressed in the follow-up changes by adding `DataMartDataTraversal.cancel()`. This keeps eager stream opening and `runId` availability, while giving callers an explicit way to close an opened traversal they decide not to iterate. Added a regression test for opening a traversal, cancelling it without iteration, and rejecting later traversal attempts.

max-voloshyn:
Addressed release-note/docs cleanup from https://github.com/OWOX/owox-data-marts/pull/1260#issuecomment-4599668140 in the follow-up changes.

- The HTTP Data changeset now documents `columns=*` / `columns=**` as selectors and repeated `column` values as exact names.
- The API/CLI changeset now lists `owox-ctl data-marts stream`.
- The ctl docs now say `data-marts stream` emits NDJSON rows, and the nonexistent `--output` reference was removed.

packages/api-client/src/data-marts.ts
```diff
@@ -34,6 +47,175 @@ function parsePage(response: unknown): DataMartsPage {
...
+
+function withDataMartContext(error: OWOXApiError, dataMartId: string): OWOXApiError {
+  const details = isRecord(error.details)
+    ? { dataMartId, ...error.details }
+    : {
+        dataMartId,
+        ...(error.details === undefined ? {} : { details: error.details }),
+      };
+
+  return new OWOXApiError(error.message, {
```
max-voloshyn: `withDataMartContext()` adds the missing `dataMartId` context, but it currently does so by reconstructing every client error as a plain `OWOXApiError`. If `getStream()` fails during token exchange or the stream endpoint still returns 401 after retry, the original error is an `OWOXAuthError`; after this wrapper, `traverseData()` callers can no longer use the documented `instanceof OWOXAuthError` handling while other client methods still can. Please keep the added data mart context while preserving or recreating the original auth subclass, and add coverage for a `traverseData()` auth failure.
🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn: Addressed. `withDataMartContext()` now preserves `OWOXAuthError` when adding `dataMartId` context, so `traverseData()` callers can still use `instanceof OWOXAuthError` consistently with other API client methods. Added coverage for a `traverseData()` auth failure during token exchange.

kalnik-a-a (approved):

PR #1261, opened 2026-05-31
Improve API key credential dialogs and details sidebar

max-voloshyn:
## Summary

- Add copyable API Origin to the API key creation success dialog, alongside API Key ID and API Key Secret.
- Keep newly created secrets hidden by default with reveal/copy controls and a one-time visibility notice.
- Add URL-addressable API key details sidebar via `apiKeyId`, including General, Credentials, Documentation, and Danger zone sections.
- Open API key details from table row clicks and support revoking from the details sidebar.
- Expose `PUBLIC_ORIGIN` through public flags without adding route-level fallback logic.
- Align API key forms, labels, help hints, copy controls, docs links, and expiration display with existing sidebar patterns.

## Testing

- `npm exec --workspace owox -- mocha --forbid-only "test/web/flags-route.test.ts"`
- `npm test --workspace @owox/web -- CreateApiKeySheet.test.tsx EditApiKeySheet.test.tsx SecretRevealDialog.test.tsx ApiKeysTable.test.tsx MyApiKeysPage.test.tsx`
- `npm run lint --workspace owox`
- `npm exec --workspace owox -- tsc --noEmit`
- `npm run lint --workspace @owox/web`
- `npm run build --workspace @owox/web`
- `git diff --check`


apps/web/src/features/api-keys/components/SecretRevealDialog.tsx
```diff
@@ -10,34 +10,71 @@ import {
...
 
   useEffect(() => {
-    setSecretCopied(false);
-  }, [data]);
+    setSecretVisible(false);
+  }, [data?.apiKeySecret]);
 
   if (!data) return null;
 
+  const apiOrigin = typeof flags?.PUBLIC_ORIGIN === 'string' ? flags.PUBLIC_ORIGIN : '';
```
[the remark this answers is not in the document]
max-voloshyn: Checked this against the active self-hosted setup on `http://localhost:3127`: `/api/flags` includes `PUBLIC_ORIGIN: "http://localhost:3127"`, so the creation dialog has a usable API Origin.

The flags route itself intentionally does not synthesize this value; public origin resolution happens before flags are served. I’m leaving the implementation unchanged unless we identify a startup path where API keys are enabled and `PUBLIC_ORIGIN` is actually absent from `/api/flags`.

apps/web/src/features/api-keys/components/SecretRevealDialog.tsx
```diff
@@ -10,34 +10,71 @@ import {
...
+            type='button'
+            aria-label='Help information'
+            className='text-muted-foreground/60 hover:text-muted-foreground focus-visible:ring-ring/50 inline-flex size-4 shrink-0 items-center justify-center rounded-sm transition-colors focus-visible:ring-2 focus-visible:outline-none'
+          >
+            <Info className='size-3.5' aria-hidden='true' />
+          </button>
+        </TooltipTrigger>
+        <TooltipContent side='top' align='start' role='tooltip'>
+          {description}
+        </TooltipContent>
```
romandubovyi: `FieldLabel` here does the same job as `ApiKeyFormLabel` in `ApiKeyFormShared.tsx` — both render a label + info tooltip. Consider unifying them, or at minimum exporting `FieldLabel` from the shared file.

apps/web/src/features/api-keys/components/EditApiKeySheet.tsx
```diff
@@ -51,6 +105,16 @@ export function EditApiKeySheet({ apiKey, onClose, onUpdated }: EditApiKeySheetP
 
   const { control, handleSubmit, reset } = form;
 
+  const apiOrigin = typeof flags?.PUBLIC_ORIGIN === 'string' ? flags.PUBLIC_ORIGIN : '';
```
romandubovyi: The `apiOrigin` derivation (`typeof flags?.PUBLIC_ORIGIN === 'string' ? ...`) is copy-pasted identically in both `EditApiKeySheet` and `SecretRevealDialog`. Extracting it to a small selector or hook in `utils.ts`
   would prevent drift if the logic ever changes.

apps/web/src/features/api-keys/components/ApiKeysTable/ApiKeysTable.tsx
```diff
@@ -24,11 +32,27 @@ export function ApiKeysTable({ keys, onCreateKey, onEditName, onRevoke }: ApiKey
     enableRowSelection: false,
   });
 
+  const handleRowClick = (row: Row<ProjectMemberApiKey>, e: React.MouseEvent) => {
```
romandubovyi: `handleRowClick` is recreated on every render. Wrapping it with `useCallback` (depends on `onOpenDetails`) keeps it stable and avoids unnecessary re-renders of `BaseTable`.

romandubovyi (approved):
Good overall — clean separation into sections, good test coverage, and the URL-addressable sidebar is a nice UX improvement

apps/web/src/features/api-keys/components/SecretRevealDialog.tsx
```diff
@@ -53,30 +90,59 @@ export function SecretRevealDialog({ data, onDone }: SecretRevealDialogProps) {
...
+                  value={apiOrigin}
+                  readOnly
+                  tabIndex={-1}
+                  className='min-w-0 flex-1 bg-transparent font-mono text-sm outline-none'
+                />
+                <Button
+                  variant='ghost'
+                  size='icon'
+                  className='size-7'
+                  aria-label='Copy API Origin'
```
romandubovyi, line 112: When `PUBLIC_ORIGIN` is not configured, `apiOrigin` is `''` but the Copy button remains enabled — clicking it silently copies an empty string. Consider: (a) disabling with disabled={!apiOrigin}, or (b) hiding the row
  when absent. Affects both EditApiKeySheet and SecretRevealDialog.

max-voloshyn: Screenshot 2026-06-01 at 16 21 48

https://github.com/OWOX/owox-data-marts/pull/1261#discussion_r3331055095 ? 🤔

apps/web/src/features/api-keys/components/SecretRevealDialog.tsx
max-voloshyn: Removed the local `FieldLabel` abstraction. `SecretRevealDialog` is not inside React Hook Form, so it cannot safely reuse `ApiKeyFormLabel` / `FormLabel`. I kept the existing non-form `Label` + `Tooltip` + `Info` pattern inline instead of introducing another shared label component just for API keys.

Pushed in `e413fb230`.

apps/web/src/features/api-keys/components/EditApiKeySheet.tsx
max-voloshyn: Changed this by moving the guarantee to `/api/flags`: `PUBLIC_ORIGIN` is now always returned as a resolved value. With that in place, the client no longer needs origin derivation logic or a selector/hook; both components just consume the already-resolved flag.

Pushed in `e413fb230`.

apps/web/src/features/api-keys/components/ApiKeysTable/ApiKeysTable.tsx
max-voloshyn: Wrapped `handleRowClick` in `useCallback` with `onOpenDetails` as dependency. The existing `Element` / `closest()` guard is kept so clicks on nested SVG/action controls do not open the details sidebar.

Pushed in `e413fb230`.

apps/web/src/features/api-keys/components/SecretRevealDialog.tsx
max-voloshyn: Addressed this at the flags source instead of disabling or hiding the credential row. `/api/flags` now always includes resolved `PUBLIC_ORIGIN`, including when it is omitted from the public env whitelist. Added route tests for explicit origin, fallback origin, and custom whitelist behavior.

Pushed in `e413fb230`.

apps/owox/src/utils/public-origin.ts
```diff
@@ -0,0 +1,8 @@
+export function resolvePublicOrigin(): string {
+  const publicOrigin = process.env.PUBLIC_ORIGIN?.trim();
+  if (publicOrigin) {
+    return publicOrigin;
+  }
+
+  return `http://localhost:${process.env.PORT}`;
```
romandubovyi, line 7: Should we avoid hardcoding the http protocol here? It might break in production where https is used

romandubovyi (approved):

apps/owox/src/utils/public-origin.ts
max-voloshyn: The hardcoded `http://localhost:${process.env.PORT}` is only the fallback path when `PUBLIC_ORIGIN` is not configured. Production HTTPS deployments should set `PUBLIC_ORIGIN` explicitly, and that value is returned as-is before the fallback is used.

I’d avoid trying to infer `https` here because this resolver is also used outside request context, and behind proxies/load balancers the Node process usually cannot know the external scheme reliably. `PUBLIC_ORIGIN` remains the production source of truth; the localhost fallback is for local/self-hosted default startup only.

PR #1266, opened 2026-06-01
Simplify owox-ctl as an automation-first Control CLI

max-voloshyn:
## Summary

Simplifies `owox-ctl` around script, CI, and AI agent automation workflows.

- Keeps `owox-ctl` positioned as the OWOX Data Marts Control CLI
- Removes interactive auth login/logout and local credential storage
- Makes all command output JSON-only
- Keeps environment-based configuration with `.env` and `--env-file`
- Defaults `owox-ctl` to OWOX Data Marts Cloud at `https://app.owox.com`
- Adds top-level `owox-ctl status`
- Reports the resolved env file path as `envFile`, or `null` when none was used
- Improves API client request construction and network error messages
- Updates docs and the existing changeset for the simplified CLI surface

Supported commands:

- `owox-ctl status`
- `owox-ctl data-marts list`
- `owox-ctl storages list`
- `owox-ctl destinations list`

## Breaking Changes

- Removes `auth login`, `auth logout`, and `auth status`
- Removes table output, color output, and output format flags
- Removes dedicated `owox-ctl` JSON config storage
- Requires credentials through environment variables or env files

## Verification

- `npm run test -w @owox/ctl`
- `npm run typecheck -w @owox/ctl`
- `npm run lint -w @owox/ctl`
- `npm run test -w @owox/api-client`
- `npm run typecheck -w @owox/api-client`
- `npm run lint -w @owox/api-client`
- `./apps/ctl/bin/run.js --help`


apps/ctl/src/commands/status.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export default class StatusCommand extends BaseCommand {
+  static override description = 'Validate OWOX Data Marts API credentials';
+  static override flags = {
+    ...BaseCommand.baseFlags,
+  };
+
+  async run(): Promise<void> {
+    const { flags } = await this.parse(StatusCommand);
```
max-voloshyn: This parse call runs before the command’s JSON error handler, and the same pattern exists in the list commands. For example, `owox-ctl status --unknown` currently exits with oclif’s human-readable error/help text instead of JSON, which breaks the automation-first JSON error contract. Please move parsing under a JSON error boundary, or add a BaseCommand-level catch, so flag/argument errors also emit the normalized JSON shape.
🤖 Reviewed by Codex (GPT-5.5)

docs/superpowers/specs/2026-06-01-owox-ctl-automation-cli-design.md
```diff
@@ -0,0 +1,150 @@
+# owox-ctl Automation CLI Simplification Design
```
max-voloshyn: This adds a tracked file under `docs/superpowers/`, but that directory is explicitly ignored by `.gitignore`, so it looks like local planning output rather than maintained repo documentation. It also already diverges from the implementation/docs contract, for example saying status masks `apiKeyId` and preserves the user-provided env-file path while the command returns the full ID and absolute path. Please remove it or move the maintained content into non-ignored docs.
🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn: Fixed in d22e555ae. Removed the tracked docs/superpowers design spec from the PR; the maintained contract stays in the public docs and changeset.

PR #1270, opened 2026-06-02
fix(backend): preserve user column formats on Google Sheets refresh

laskevych:
## Problem

When a Report Run updates a Google Sheet, the column formatting a user set in
the Sheets UI (dates, numbers, currencies, custom/datetime formats) was reset
on every refresh. Expected behaviour — formatting is preserved, the way the old
OWOX Sheets extension worked.

## Root cause

The writer commits every data cell with `valueInputOption: 'USER_ENTERED'`
(`google-sheets-api.adapter.ts`). With `USER_ENTERED`, Sheets re-derives each
cell's number format from the parsed value and **overwrites** the user's
explicit format. It's worst for `TIMESTAMP` columns, which the writer itself
stringifies to `"yyyy-MM-dd HH:mm:ss"` (`sheet-values-formatter.ts`) — Sheets
reparses that to a datetime serial and stamps it with its **default** format —
and for `DATE` columns arriving as ISO strings. Plain integers usually survive,
which is why users mostly noticed it on date/currency columns.

Things that were _not_ the cause (ruled out): the header-row "clear format"
request only touches `textFormat`/`backgroundColor`, never `numberFormat`;
`values.clear` preserves formatting; and a stable schema produces zero
structural column ops.

## Fix — capture & restore (reconstructing the old extension's behaviour)

The old extension never re-derived formats. We reproduce that:

1. **Before** any write, capture each imported column's
   `userEnteredFormat.numberFormat` (sampled from the first data row), keyed by
   the column's **canonical name** so it survives reorder/delete.
2. **After** the data write, in the same atomic `finalize` `batchUpdate`,
   re-apply the captured formats over the written data rows. Re-applied format
   wins over whatever `USER_ENTERED` inferred.

Notable properties:
- Format is keyed by name → follows a column through user reorder and ODM
  delete ops (a dropped column's format is not re-applied to a neighbour).
- Columns the user never formatted are left untouched — we never _impose_ a
  format.
- First run captures/restores nothing; failed refreshes restore nothing
  (deferred-mutations contract preserved — the sheet is left exactly as it was).
- Bonus: the format now extends to rows that grew on this refresh.

## Changes

- `column-plan.dto.ts` / `column-plan-builder.ts` — expose
  `currentImportedNames` (canonical name per current row-1 position).
- `google-sheets-api.adapter.ts` — `getColumnNumberFormats()` (read) and
  `buildSetColumnNumberFormatRequest()` (a `repeatCell` touching only
  `userEnteredFormat.numberFormat`).
- `google-sheets-report-writer.ts` — `captureUserColumnFormats()` in
  `prepareToWriteReport`; restore requests appended to the `finalize` batch.
- `getColumnNumberFormats` selects the destination tab by `sheetId` (not
  `sheets[0]`) — `ranges` does not guarantee the requested tab is first, so a
  non-first destination sheet would otherwise read empty grid data and silently
  skip the format restore.

## Tests

- **Integration (real Google Sheets + BigQuery):** new `Test 11` in
  `google-sheets-column-preservation.integration.ts` — applies currency +
  custom number formats to data columns, runs the report twice, asserts
  `numberFormat` is preserved on both a pre-existing row and a freshly grown
  row. Verified passing against the live Sheets API (13/13 in the suite).
- **Unit:** 5 new cases in `google-sheets-report-writer.spec.ts` (capture
  before write, restore after write, first-run no-op, failed-refresh no-op,
  dropped-column skip) + `currentImportedNames` assertions in
  `column-plan-builder.spec.ts`.

## Verification

- Integration (real Sheets + BQ): 13/13 ✅
- Unit (writer / builder / adapter specs): 56/56 ✅
- `tsc` + ESLint on changed files: clean ✅

## Migration note

On the first refresh after deploy over an already-clobbered sheet, the writer
captures whatever format is currently present and preserves it. The user
re-applies their intended format once, and from then on it sticks across
refreshes.

### DIFF
--- a/apps/backend/src/data-marts/data-destination-types/google-sheets/adapters/google-sheets-api.adapter.ts
+++ b/apps/backend/src/data-marts/data-destination-types/google-sheets/adapters/google-sheets-api.adapter.ts
@@ -190,7 +190,14 @@ export class GoogleSheetsApiAdapter {
    * format (default / "Automatic" cell). Returns all-`undefined` (and never
    * throws) when the requested range lies outside the current grid.
    *
+   * The target sheet is selected by `sheetId`, not by position: `ranges` only
+   * limits which grid-data portions are included in the response, it does NOT
+   * guarantee that `sheets[0]` is the requested tab. Relying on `sheets[0]`
+   * would read empty/other-tab grid data for any destination that is not the
+   * first sheet, silently skipping the format restore.
+   *
    * @param spreadsheetId - ID of the spreadsheet
+   * @param sheetId - ID of the destination sheet to read from
    * @param sheetTitle - Title of the sheet (used in the A1 range)
    * @param row - 1-based row index to sample (typically the first data row)
    * @param fromCol - 1-based, inclusive
@@ -198,6 +205,7 @@ export class GoogleSheetsApiAdapter {
    */
   public async getColumnNumberFormats(
     spreadsheetId: string,
+    sheetId: number,
     sheetTitle: string,
     row: number,
     fromCol: number,
@@ -213,10 +221,12 @@ export class GoogleSheetsApiAdapter {
         spreadsheetId,
         ranges: [range],
         includeGridData: true,
-        fields: 'sheets(data(rowData(values(userEnteredFormat(numberFormat)))))',
+        fields:
+          'sheets(properties(sheetId),data(rowData(values(userEnteredFormat(numberFormat)))))',
       })
     );
-    const cells = resp.data.sheets?.[0]?.data?.[0]?.rowData?.[0]?.values ?? [];
+    const sheet = resp.data.sheets?.find(s => s.properties?.sheetId === sheetId);
+    const cells = sheet?.data?.[0]?.rowData?.[0]?.values ?? [];
     const formats: (sheets_v4.Schema$NumberFormat | undefined)[] = new Array(width).fill(undefined);
     for (let i = 0; i < width; i++) {
       formats[i] = cells[i]?.userEnteredFormat?.numberFormat ?? undefined;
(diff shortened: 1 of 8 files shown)


apps/backend/src/data-marts/data-destination-types/google-sheets/adapters/google-sheets-api.adapter.ts
```diff
@@ -177,6 +177,87 @@ export class GoogleSheetsApiAdapter {
...
+    const range = `'${sheetTitle}'!${GoogleSheetsApiAdapter.colToA1(fromCol)}${row}:${GoogleSheetsApiAdapter.colToA1(toCol)}${row}`;
+    const resp = await this.executeWithRetry(() =>
+      this.service.spreadsheets.get({
+        spreadsheetId,
+        ranges: [range],
+        includeGridData: true,
+        fields: 'sheets(data(rowData(values(userEnteredFormat(numberFormat)))))',
+      })
+    );
+    const cells = resp.data.sheets?.[0]?.data?.[0]?.rowData?.[0]?.values ?? [];
```
max-voloshyn: `spreadsheets.get` returns a spreadsheet resource, and `ranges` only limits which grid-data portions are included; it does not make `sheets[0]` the requested tab. For a destination tab that is not the first sheet, this can read empty/other-tab grid data and return all `undefined`, so the refresh silently skips restoring the user's formats. Please include `sheets(properties(sheetId,title),data(...))` and select the matching sheet, preferably by `sheetId` passed into this method, before reading `rowData`.
🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (approved):

PR #1272, opened 2026-06-03
feat(data-marts): output controls for Athena data marts (filters, sli…

zapolsky:
…ces, sort, limit)

Extend report-level output controls — already shipped for BigQuery — to AWS Athena, across both consumers (Reports and the HTTP Data API), plus live integration coverage and a scheduled CI workflow.

Backend
- AthenaClauseRenderer (Trino dialect): all 19 operators with positional `?` placeholders and double-quoted identifiers; substring/affix matchers use strpos/substr (never LIKE) so `%`/`_` can't smuggle wildcards; relative_date via date_add/date_trunc. nextParamName lifted to the SqlClauseRenderer base.
- Non-blended (AthenaQueryBuilder) and blended (AthenaBlendedQueryBuilder) now apply output controls; legacy {limit:0} schema probe preserved for SQL-defined marts; positional params emit in textual order (pre-join CTEs before post-join).
- AthenaApiAdapter binds values via Athena ExecutionParameters (rendered as SQL literals per AWS docs); AthenaReportReader forwards sqlOverride params.
- Capability enabled for AWS_ATHENA; renderer registered in DI.
- Fix: OutputControlsValidatorService only recognised BigQuery type names, so every string/number/bool filter on Athena (VARCHAR/INTEGER/BOOLEAN/…) was rejected with 400. Type sets now match what each provider emits (incl. zoned timestamps); is_null/is_not_null allowed for any column type. Caught by the live integration tests, not by generated-SQL checks.
- getColumnComments escapes single quotes in the admin-supplied FQN.

Frontend
- Unlock the output-controls UI for Athena in apps/web (and the Google Sheets extension, in its own repo) via the capability mirror.

Tests
- Unit: renderer operator matrix, builder (non-blended/blended/param-order), validator type matrix (incl. zoned timestamps + type-agnostic is_null), capability-negative guards, composer schema-drift, multi-filter/filter-only, run-config snapshots.
- HTTP Data e2e (mock): output-controls validation + applied wiring + run metadata.
- Live integration (credential-gated) against real Athena + BigQuery: adapter operator matrix + relative_date + wildcard-literal safety; full HTTP Data API (no-OC, eq/contains/between/sort/limit, SQL-def via real CreateView, multi-batch streaming, filter-only column); pre-join slices narrowing a joined mart; report read path. S3 prefixes isolated per run so parallel jobs don't collide; integration suite runs serially locally (maxWorkers:1).

CI
- test-integration.yml: daily at 08:00 & 16:00 (+03:00) and on demand, as 5 parallel jobs (BigQuery / Athena / HTTP Data / Report read / Google Sheets column preservation), credentials from GitHub secrets.


apps/backend/src/data-marts/data-storage-types/athena/adapters/athena-execution-parameters.utils.ts
```diff
@@ -0,0 +1,26 @@
...
+): string[] | undefined {
+  if (!params || params.length === 0) return undefined;
+  return params.map(p => formatLiteral(p.value));
+}
+
+function formatLiteral(value: string | number | boolean | null): string {
+  if (value === null) return 'NULL';
+  if (typeof value === 'number') return String(value);
+  if (typeof value === 'boolean') return value ? 'true' : 'false';
+  return `'${value.split("'").join("''")}'`;
```
max-voloshyn, line 25: Date/time scalar filters will fail for Athena because string values are always emitted as VARCHAR literals. The UI sends date inputs as strings, so a filter like `created_at >= '2024-01-01'` compares a DATE/TIMESTAMP column to VARCHAR; Athena/Trino requires typed literals or casts for those parameters. AWS calls out using `CAST('2014-07-05' AS DATE)` for date execution parameters: https://docs.aws.amazon.com/athena/latest/ug/querying-with-prepared-statements.html. Please carry field type into parameter formatting or render typed/cast parameters for DATE/TIME/TIMESTAMP operators, and add a real Athena integration test for date `eq`/`between`.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/output-controls-capability.service.ts
```diff
@@ -3,8 +3,10 @@ import { DataStorageType } from '../data-storage-types/enums/data-storage-type.e
 
 @Injectable()
 export class OutputControlsCapabilityService {
+  // Source of truth — keep frontend mirrors (web + extension output-controls-support.ts) in sync.
   private readonly supported: ReadonlySet<DataStorageType> = new Set([
     DataStorageType.GOOGLE_BIGQUERY,
+    DataStorageType.AWS_ATHENA,
```
max-voloshyn, line 9: Enabling Athena here exposes output controls through Looker Studio, but the cached-reader path still drops both native output controls and SQL params. `ReportDataCacheService.resolvePrepareOptions()` only passes `sqlOverride` for blended reports and never passes `decision.params`; non-blended Athena reports through Looker cache ignore filter/sort/limit, while blended/sliced Athena reports produce SQL containing `?` without `sqlOverrideParams`, so execution fails. Please route this cache path through `ReportSqlComposerService` like `RunReportService` does and pass `sqlOverrideParams`.

🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-marts/shared/utils/output-controls-support.ts
```diff
@@ -1,6 +1,10 @@
 import { DataStorageType } from '../../../data-storage/shared/model/types/data-storage-type.enum';
 
-const SUPPORTED: ReadonlySet<DataStorageType> = new Set([DataStorageType.GOOGLE_BIGQUERY]);
+// Mirror of backend OutputControlsCapabilityService — keep in sync.
+const SUPPORTED: ReadonlySet<DataStorageType> = new Set([
+  DataStorageType.GOOGLE_BIGQUERY,
+  DataStorageType.AWS_ATHENA,
```
max-voloshyn, line 6: The web app now advertises Athena output controls, but the operator/type matrix still only recognizes the BigQuery subset (`STRING`, `INTEGER`, `FLOAT`, `NUMERIC`, `BIGNUMERIC`, `DATE`, `DATETIME`, `TIMESTAMP`, `TIME`, `BOOLEAN`). Common Athena schema types such as `VARCHAR`, `CHAR`, `BIGINT`, `SMALLINT`, `TINYINT`, `REAL`, `DOUBLE`, `DECIMAL`, and zoned timestamps are treated as unfilterable in `operatorsForType()`, so users will see Athena support enabled but cannot add filters on normal Athena columns. Please update the frontend type sets and tests alongside this support flag.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/data-storage-types/athena/adapters/athena-execution-parameters.utils.ts
zapolsky: Good catch — fixed. The Athena renderer now carries the column's storage type and wraps date/time value comparisons (eq/neq/gt/lt/gte/lte/between) in `CAST(? AS <type>)` for DATE/TIME/TIMESTAMP and the zoned variants, so a string parameter is compared as the proper type instead of as VARCHAR. `relative_date` was already safe since it renders param-free.

This is done for both paths — non-blended (`AthenaQueryBuilder`) and blended (`AbstractBlendedQueryBuilder`, post-join filters and pre-join slices) — with column types sourced from the persisted/blendable schema, matching what `OutputControlsValidatorService` types against. BigQuery's named-parameter renderer is untouched.

Covered by renderer/builder unit specs, a live Athena integration test (eq/gte/between on a real TIMESTAMP column), and a `/generated-sql` e2e asserting `CAST(? AS TIMESTAMP)` is emitted.

apps/backend/src/data-marts/services/output-controls-capability.service.ts
zapolsky: Fixed. `ReportDataCacheService.resolvePrepareOptions()` now routes through `ReportSqlComposerService` exactly like `RunReportService`: it forwards `decision.params` for blended reports and composes SQL + params for non-blended reports with output controls. So the Looker Studio cached path no longer drops filter/sort/limit on non-blended reports, and blended/sliced reports no longer emit unbound `?` / `@p` placeholders. The per-minute cleanup path skips the extra compose since it only finalizes (never executes).

Added a unit spec, plus an e2e that drives `POST /looker/get-data` with the real `ReportDataCacheService` (only the storage reader is a spy) and asserts the reader receives the composed `sqlOverride` + bound `sqlOverrideParams`.

apps/web/src/features/data-marts/shared/utils/output-controls-support.ts
zapolsky: Fixed. Aligned the frontend `operatorsForType()` type sets in `output-controls-operators.ts` with the backend `OutputControlsValidatorService` source of truth, so `VARCHAR`, `CHAR`, `BIGINT`, `SMALLINT`, `TINYINT`, `REAL`, `DOUBLE`, `DECIMAL`, the zoned timestamps and `BOOL` now resolve to the correct operator group instead of being treated as unfilterable. Added a regression spec covering the Athena type names alongside the existing BigQuery ones.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/output-controls-operators.ts
```diff
@@ -26,10 +26,30 @@ export interface OperatorMeta {
...
 }
 
-const STRING_TYPES = new Set(['STRING']);
-const NUMBER_TYPES = new Set(['INTEGER', 'FLOAT', 'NUMERIC', 'BIGNUMERIC']);
-const DATE_TYPES = new Set(['DATE', 'DATETIME', 'TIMESTAMP', 'TIME']);
-const BOOL_TYPES = new Set(['BOOLEAN']);
+// Mirror of backend OutputControlsValidatorService type sets — cover every type
+// name each supported provider emits (BigQuery + Athena), keep the two in sync.
+const STRING_TYPES = new Set(['STRING', 'VARCHAR', 'CHAR']);
+const NUMBER_TYPES = new Set([
```
max-voloshyn, line 32: Adding Athena numeric types here makes the operator picker allow filters on `BIGINT`/`DOUBLE`/`DECIMAL`, but `FilterValueEditor` still has its own `NUMBER_TYPES` set with only `INTEGER`/`FLOAT`/`NUMERIC`/`BIGNUMERIC` (`FilterValueEditor.tsx:50`). Those Athena values are emitted as strings, and `toAthenaExecutionParameters` quotes strings, so a user-entered `BIGINT > 10` becomes a string literal instead of a numeric literal and Athena rejects the comparison. Please share these type sets or update the editor parsing/tests alongside this operator matrix.

🤖 Reviewed by Codex (GPT-5.5)

zapolsky: Good point — fixed. `FilterValueEditor` kept its own BigQuery-only `NUMBER_TYPES`/`DATE_TYPES`, so I extracted `isNumberType`/`isDateType` from `output-controls-operators.ts` as the single source of truth and switched the editor to them. An Athena `BIGINT`/`DOUBLE`/`DECIMAL`/`SMALLINT`/`TINYINT`/`REAL` value now coerces to a JS number instead of a string — so it's no longer quoted into `"col" > '10'` — and zoned-timestamp columns get the date input. Added predicate unit tests plus a `FilterValueEditor` regression over the Athena numeric type names.

zapolsky:
All review comments are addressed and pushed:

- **Date/time filters** — the Athena renderer now wraps value placeholders in `CAST(? AS <type>)` (non-blended + blended paths), so DATE/TIMESTAMP comparisons are no longer sent as varchar.
- **Looker cached-reader path** — `ReportDataCacheService` now composes via `ReportSqlComposerService` and forwards bound params, so the cache path no longer drops output controls or emits unbound placeholders.
- **FE type/operator matrix** — `operatorsForType()` is aligned with the backend `OutputControlsValidatorService`, and `FilterValueEditor` now shares the same `isNumberType`/`isDateType` predicates so Athena numeric values coerce to JS numbers (no more quoted `'10'`).

Coverage: unit + e2e + a live Athena integration test (`eq`/`gte`/`between` on a real `TIMESTAMP` column). Ready for another pass — thanks for the thorough review.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -20,13 +20,32 @@ export type ValidationError =
...
+  'BIGNUMERIC',
+  'DECIMAL',
+]);
+const DATE_TYPES = new Set([
+  'DATE',
+  'DATETIME',
+  'TIME',
+  'TIMESTAMP',
+  'TIMESTAMP WITH TIME ZONE',
+  'TIME WITH TIME ZONE',
```
max-voloshyn: This puts the newly added time-only Athena type in the same operator bucket as DATE/TIMESTAMP, so the API accepts `relative_date` for `TIME WITH TIME ZONE`. The Athena renderer later emits date predicates like `"tz_time" = current_date` / `date_add(..., current_date)`, which are invalid for a time-of-day column. Please split time-only types from date/timestamp types, or at least remove `relative_date` for time-only columns.

🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/output-controls-operators.ts
```diff
@@ -26,10 +26,30 @@ export interface OperatorMeta {
...
+  'BIGNUMERIC',
+  'DECIMAL',
+]);
+const DATE_TYPES = new Set([
+  'DATE',
+  'DATETIME',
+  'TIME',
+  'TIMESTAMP',
+  'TIMESTAMP WITH TIME ZONE',
+  'TIME WITH TIME ZONE',
```
max-voloshyn: Mirroring `TIME WITH TIME ZONE` as a date type makes the UI expose `relative_date` and makes `FilterValueEditor` render an HTML `date` input for a time-only column. That sends values like `2026-06-03`, which the backend then casts to `TIME WITH TIME ZONE` or compares against `current_date`. Please keep time-only types out of the date operator/input path.

🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/output-controls-validator.service.ts
zapolsky: Fixed. Split `TIME` / `TIME WITH TIME ZONE` out of `DATE_TYPES` into a dedicated `TIME_TYPES` bucket with its own `TIME_OPS` — the same comparison operators (`eq`/`neq`/`gt`/`lt`/`gte`/`lte`/`between` + `is_null`/`is_not_null`) minus `relative_date`. So the validator now rejects `relative_date` on a time-of-day column with `INVALID_OPERATOR_FOR_TYPE`, while `DATE`/`DATETIME`/`TIMESTAMP` (incl. the zoned timestamp) keep it — no more `current_date` / `date_add(..., current_date)` predicates emitted against a column with no date component.

Covered by a unit spec asserting `relative_date` is rejected on `TIME` / `TIME WITH TIME ZONE` and that `eq`/`between` stay allowed.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/output-controls-operators.ts
zapolsky: Fixed. Mirrored the backend split on the FE: `TIME` / `TIME WITH TIME ZONE` now resolve to a `TIME_OPERATORS` set (the date operators minus `relative_date`) via a separate `TIME_TYPES`, and `FilterValueEditor` renders an HTML `time` input (not `date`) for them through a new `isTimeType` predicate. Zoned `TIMESTAMP` still gets the `date` input and `relative_date`, so only the time-only types are kept out of the date operator/input path. The same split landed in the Google Sheets extension's copy of these files.

Vitest regression covers it: a `TIME` column withholds `relative_date` and renders the `time` input, while a zoned `TIMESTAMP` keeps the date path.

max-voloshyn (approved):

PR #1275, opened 2026-06-03
Package project member API keys as a single copyable credential

max-voloshyn:
## Summary
- Replace separate API Origin / API Key ID / API Key Secret setup with one backend-generated `owox_key_...` API Key value.
- Add API key encoding/parsing for backend creation, `@owox/api-client`, and `owox-ctl`.
- Update API key creation UI to show/copy the full API Key once, with visibility toggle and simplified details screens.
- Keep API Key ID visible as non-secret metadata for status, logs, support, and debugging.
- Update docs for API Keys, `owox-ctl`, `@owox/api-client`, and raw HTTP OpenAPI authentication flow.

## Testing
- `npm test -w @owox/backend -- project-member-api-key`
- `npm test -w @owox/api-client`
- `npm test -w @owox/ctl -- config-store.test.ts status.test.ts`
- `npm test -w @owox/web -- SecretRevealDialog.test.tsx`
- `npm test -w @owox/web -- EditApiKeySheet.test.tsx`
- `npm test -w @owox/web -- CreateApiKeySheet.test.tsx`
- `git diff --check`

Note: a full workspace `npm test` run was intentionally not used because it overloaded the local laptop during this work.


docs/superpowers/specs/2026-06-03-api-key-packaging-design.md
```diff
@@ -0,0 +1,211 @@
+# API Key Packaging Design
```
max-voloshyn: this file should not be part of PR

romandubovyi:
`PublicOriginService.getPublicOrigin()` falls back to `http://localhost:` when `PUBLIC_ORIGIN` is unset. Since the origin is now baked into the API key at creation time, a misconfigured self-managed deployment will generate permanently unusable keys. Consider adding startup validation that fails or warns when `PUBLIC_ORIGIN` defaults to localhost in production-like environments, or at least document this as a hard requirement for self-managed setups

apps/web/src/features/api-keys/components/CreateApiKeySheet.test.tsx
```diff
@@ -25,6 +25,7 @@ function renderCreateApiKeySheet() {
 describe('CreateApiKeySheet', () => {
   beforeEach(() => {
     localStorage.clear();
+    vi.restoreAllMocks();
```
romandubovyi: 🟢 **Nit:** In \`CreateApiKeySheet.test.tsx\`, \`vi.restoreAllMocks()\` resets mock implementations to \`undefined\`. If the intent is only to reset call counts between tests, \`vi.clearAllMocks()\` is safer and faster."

apps/backend/src/data-marts/use-cases/project-member-api-keys/create-project-member-api-key.service.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      false,
+      null
+    );
+    expect(publicOriginService.getPublicOrigin).toHaveBeenCalled();
+    expect(result.apiKeyId).toBe('pmk_AbCdEfGhIjKlMnOpQrStUv');
+    expect(result.apiKey).toMatch(/^owox_key_[A-Za-z0-9_-]+$/);
+    expect(result).not.toHaveProperty('apiKeySecret');
+
+    const payload = JSON.parse(
+      Buffer.from(result.apiKey.slice('owox_key_'.length), 'base64url').toString('utf8')
```
romandubovyi: 🟢 **Nit:** The create service spec manually decodes the API key payload. Extracting a shared test helper (or adding a \`decode\` method to the codec service) would reduce duplication and make the test more resilient to format changes

apps/ctl/src/config-store.ts
```diff
@@ -1,39 +1,18 @@
-import { OWOXConfigError } from '@owox/api-client';
-
-export const DEFAULT_API_ORIGIN = 'https://app.owox.com';
+import { parseOWOXApiKey } from '@owox/api-client';
 
 export type AuthConfig = {
+  apiKey: string;
   apiOrigin: string;
   apiKeyId: string;
   apiKeySecret: string;
```
romandubovyi: 🟢 **Nit:** `AuthConfig` includes `apiKeySecret` as an explicit field, but it is never read by any command — `createClient` only uses `config.apiKey`, and `status` only reads `config.apiOrigin` and `config.apiKeyId`. The field is    
  populated via `...parsed` spread but serves no purpose. Consider omitting it from the type and destructuring only the fields that are actually used. This would also eliminate one of the two places where the raw secret is held in
  memory simultaneously.

romandubovyi (approved):

max-voloshyn:
@romandubovyi checked the actual fallback path in code and locally. It does not produce bare `http://localhost:`. `PublicOriginService` resolves to `http://localhost:${PORT}`, and when `PORT` is unset it uses `DEFAULT_PORT = 3000`, so the full default is `http://localhost:3000`.

Verified locally:
- `PUBLIC_ORIGIN` unset, `PORT=3100` -> generated API key contains `http://localhost:3100`
- `PUBLIC_ORIGIN` unset, `PORT` unset -> generated API key contains `http://localhost:3000`

I do not think startup validation fits this PR because localhost is the intended local default. I added a targeted note in the API Keys docs for self-managed deployments: set `PUBLIC_ORIGIN` to the public deployment URL before creating API keys, because the origin is embedded into the key.

apps/web/src/features/api-keys/components/CreateApiKeySheet.test.tsx
max-voloshyn: Changed to `vi.clearAllMocks()`. The one test that mocks `Math.random` now restores that spy locally, so the mocked implementation does not leak into later tests.

apps/backend/src/data-marts/use-cases/project-member-api-keys/create-project-member-api-key.service.spec.ts
max-voloshyn: Added a shared backend test helper for decoding project member API keys and reused it in both the codec spec and the create-service spec. I kept this as a test helper rather than adding a production `decode` API because the backend runtime only needs to encode keys.

apps/ctl/src/config-store.ts
max-voloshyn: Removed `apiKeySecret` from `AuthConfig` and changed `resolveAuthConfig` to destructure only `apiOrigin` and `apiKeyId` from the parsed key. The raw secret remains inside the original combined `apiKey` value used by the API client.

PR #1276, opened 2026-06-04
feat(backend): added project id to triggers

laskevych:
## Summary

Follow-up to the notification-settings `projectId` fix. Several endpoints still
resolved entities by id alone, allowing one project to read/abort/refresh
another project's data. This PR binds those paths to `context.projectId`
(and `context.userId` where relevant).

Three tenant-boundary gaps are closed:

1. **UI trigger status / result / abort** fetched triggers by `triggerId` only.
2. **Connector OAuth status** returned credential status by `credentialId` without project context.
3. **Connector OAuth refresh** could read/copy a `_source_credential_id` from another project.

### UI triggers
- Lifted `projectId` onto the base `UiTrigger` entity (it was duplicated
  identically across all 8 concrete trigger entities) and removed the
  per-subclass declarations. No schema change — these use concrete-table
  inheritance, so the column already exists in every table.
- `UiTriggerService.getTriggerStatus` / `getTriggerResponse` / `abortTriggerRun`
  / `getTriggerById` now take `projectId` and filter
  `findOne({ where: { id, projectId } })`. A trigger from another project is
  treated as not-found before any read or removal. `abortTriggerRun` keeps its
  existing `userId` ownership check on top.
- `getTriggerById` **fails closed** when `projectId` is missing: TypeORM drops
  `undefined` keys from a `where` clause, which would otherwise collapse the
  query to `{ id }` and silently disable the tenant filter.
- Base + subclass controllers (`ui-trigger-controller`, `ai-helper`,
  `sql-dry-run`) inject `@AuthContext()` and pass `context.projectId`. The other
  6 trigger controllers inherit the base behavior automatically.

### Connector OAuth
- `ConnectorOauthService.getCredentialStatus` takes `projectId` and treats a
  credential from another project as non-existent (same error as not-found, so
  validity isn't leaked by probing expiry). `ConnectorController` passes
  `context.projectId`.
- `ConnectorService.refreshCredentials` rejects a credential whose `projectId`
  differs from the caller's **before** reading or copying its secrets — closing
  the path where a cross-project credential could be copied into the attacker's
  project. Mirrors the check already present on the injection path.

## Behavior

- Same-project access is unchanged.
- Cross-project access returns not-found instead of leaking or mutating data.
- Connector checks fail **closed** on a missing `projectId`; the trigger path is
  now hardened to do the same.

## Tests

- New `ui-trigger.service.spec.ts` — same-project success and cross-project
  not-found for status / result / abort, the same-project / other-user forbidden
  case, and a fail-closed case asserting `findOne` is never called when
  `projectId` is missing.
- `connector-oauth.service.spec.ts` — updated for the new signature plus a
  cross-project denial case that asserts expiry is not probed.
- `connector.service.spec.ts` — refresh same-project-success and
  cross-project-denial, asserting no credential is copied across the boundary.

## Notes / out of scope

- The `Strategy.PARSE` auth path does not validate `projectId` presence. This PR
  hardens the trigger lookup against an undefined `projectId`; validating it at
  the auth-guard level (to harden all endpoints uniformly) is left as a separate
  change.
- Pre-existing `tsc` errors in unrelated files (`@owox/idp-protocol` exports from
  the recent user-provisioning work, and some other specs) are not affected by
  this PR.

### DIFF
--- a/apps/backend/src/common/scheduler/shared/ui-trigger.service.ts
+++ b/apps/backend/src/common/scheduler/shared/ui-trigger.service.ts
@@ -108,6 +108,13 @@ export abstract class UiTriggerService<UiResponseType> {
     id: string,
     projectId: string
   ): Promise<UiTrigger<UiResponseType>> {
+    // Fail closed when projectId is missing: TypeORM drops `undefined` keys from
+    // the where clause, which would otherwise collapse the query to `{ id }` and
+    // silently disable the tenant boundary.
+    if (!projectId) {
+      throw new NotFoundException(`Trigger with id ${id} not found`);
+    }
+
     const trigger = await this.triggerRepository.findOne({
       where: { id, projectId },
     });
--- a/apps/backend/src/common/scheduler/shared/ui-trigger.service.ts
+++ b/apps/backend/src/common/scheduler/shared/ui-trigger.service.ts
@@ -29,20 +29,20 @@ export abstract class UiTriggerService<UiResponseType> {
   }
 
   /**
-   * Get trigger status by ID
+   * Get trigger status by ID, scoped to the authenticated project.
    */
-  async getTriggerStatus(triggerId: string): Promise<TriggerStatus> {
+  async getTriggerStatus(triggerId: string, projectId: string): Promise<TriggerStatus> {
     this.logger.debug(`Checking trigger status for trigger ${triggerId}`);
-    const trigger = await this.getTriggerById(triggerId);
+    const trigger = await this.getTriggerById(triggerId, projectId);
     return trigger.status;
   }
 
   /**
-   * Get trigger response and handle different status cases
+   * Get trigger response and handle different status cases, scoped to the authenticated project.
    */
-  async getTriggerResponse(triggerId: string): Promise<UiResponseType> {
+  async getTriggerResponse(triggerId: string, projectId: string): Promise<UiResponseType> {
     this.logger.debug(`Getting trigger response for trigger ${triggerId}`);
-    const trigger = await this.getTriggerById(triggerId);
+    const trigger = await this.getTriggerById(triggerId, projectId);
 
     if (trigger.status === TriggerStatus.SUCCESS) {
       const response = trigger.uiResponse;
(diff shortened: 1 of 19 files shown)


apps/backend/src/common/scheduler/shared/ui-trigger.service.ts
```diff
@@ -98,11 +98,25 @@ export abstract class UiTriggerService<UiResponseType> {
...
+    // Fail closed when projectId is missing: TypeORM drops `undefined` keys from
+    // the where clause, which would otherwise collapse the query to `{ id }` and
+    // silently disable the tenant boundary.
+    if (!projectId) {
+      throw new NotFoundException(`Trigger with id ${id} not found`);
+    }
+
     const trigger = await this.triggerRepository.findOne({
-      where: { id },
+      where: { id, projectId },
```
max-voloshyn, line 119: The lookup now filters by project, but the inherited lifecycle routes still ignore the parent resource id in the URL (`dataMartId`/`dataStorageId`). That bypasses the `AccessDecisionService` checks used by create/list endpoints for selected-context users: any project viewer who learns a trigger id can call any matching trigger route in the project and read/remove a result from a Data Mart or Storage they cannot access. Please bind the lookup to the route resource id, or run the same resource access check before returning/removing the trigger.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/connector/connector.service.ts
```diff
@@ -238,6 +238,12 @@ export class ConnectorService {
       throw new Error(`Credential with ID ${credentialId} not found`);
     }
 
+    // Tenant boundary: never read or copy a credential that belongs to another
+    // project, even if its id is referenced from this project's configuration.
+    if (credential.projectId !== projectId) {
```
max-voloshyn, line 243: This blocks cross-project credentials, but a same-project credential from a different connector still passes and is then fed into `createConnectorSource(connectorName).refreshCredentials(...)`. For providers that rotate tokens, that can copy/refresh the wrong connector’s credential under the requested connector name. Please also verify `credential.connectorName === connectorName` before refreshing.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/common/scheduler/shared/ui-trigger.service.ts
laskevych: Thanks for the comment. I don't see the point in doing that. Triggers are deleted after being read. Each trigger has a unique ID.

If you think it's worth doing, let me know. I'll try to implement it.

apps/backend/src/data-marts/services/connector/connector.service.ts
laskevych: Good catch. Fixed ✅

apps/backend/src/common/scheduler/shared/ui-trigger.service.ts
max-voloshyn: Thanks, that makes sense. After tracing the payloads, I agree this is a low-probability inner-project risk rather than a cross-project issue. The `projectId` filter blocks cross-project access, and exploiting the remaining gap would require a leaked valid trigger UUID before it is consumed/expired.

max-voloshyn (approved):

PR #1279, opened 2026-06-04
fix request access project switcher

amarchenk0:


apps/web/src/components/AppSidebar/ProjectMenu/ProjectMenuContent.tsx
```diff
@@ -8,12 +9,46 @@ import { ProjectSettingsSubmenu } from './ProjectSettingsSubmenu';
...
+function RestrictedProjectMenuContent() {
+  const { projects, loadProjects, callState, error, isLoading } = useProjects();
+
+  useEffect(() => {
+    if (callState === RequestStatus.IDLE) {
+      void loadProjects();
+    }
+  }, [callState, loadProjects]);
+
+  const shouldShowSwitchProject =
```
max-voloshyn: The restricted switcher treats any returned project as “another” project, but the regular menu assumes this list includes the current project (`canSwitchProject = projects.length > 1`). If `/auth/api/projects` includes the current denied project, a user with no real alternatives will see “Switch project” with only a link back to the same request-access project instead of “No other projects available”. Filter out `user?.projectId` before deciding `shouldShowSwitchProject` and before rendering the restricted switcher.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/components/AppSidebar/ProjectMenu/ProjectMenuContent.tsx
```diff
@@ -8,12 +9,46 @@ import { ProjectSettingsSubmenu } from './ProjectSettingsSubmenu';
...
+import { useProjects } from '../../../features/idp/hooks/useProjects.ts';
+import { RequestStatus } from '../../../shared/types/request-status.ts';
 
 interface ProjectMenuContentProps {
   onClose: () => void;
+  restricted?: boolean;
 }
 
-export function ProjectMenuContent({ onClose }: ProjectMenuContentProps) {
+export function ProjectMenuContent({ onClose, restricted = false }: ProjectMenuContentProps) {
```
romandubovyi, line 17: Nit: `onClose` is required in the interface but unused when `restricted` is true. Consider making it optional, or passing it through to `RestrictedProjectMenuContent` so the menu can close after a project switch.

apps/web/src/components/AppSidebar/ProjectMenu/ProjectMenuContent.tsx
```diff
@@ -8,12 +9,46 @@ import { ProjectSettingsSubmenu } from './ProjectSettingsSubmenu';
...
+  if (restricted) {
+    return <RestrictedProjectMenuContent />;
+  }
+
+  return <RegularProjectMenuContent onClose={onClose} />;
+}
+
+function RestrictedProjectMenuContent() {
+  const { projects, loadProjects, callState, error, isLoading } = useProjects();
```
romandubovyi: Both `RestrictedProjectMenuContent` and `SwitchProjectMenuInner` call `useProjects()` independently. Since they share `ProjectsProvider` context, data should stay in sync — but the `loadProjects()` call here may race with any loading logic inside `SwitchProjectMenu`. Consider whether the restricted variant could simply always render `<SwitchProjectMenu>` and let it handle its own empty/loading/error states, removing the need for the outer `useProjects()` + visibility logic.

apps/web/src/components/AppSidebar/ProjectMenu/ProjectMenuContent.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      projects: [{ id: 'project-2', title: 'Project 2' }],
+      callState: RequestStatus.LOADED,
+    };
+
+    render(<ProjectMenuContent restricted onClose={vi.fn()} />);
+
+    expect(screen.getByText('Switch project')).toBeInTheDocument();
+    expect(screen.queryByText('No other projects available')).not.toBeInTheDocument();
+  });
+});
```
romandubovyi, line 48: Missing test case: when `callState` is `LOADED` and `projects` is empty, the component should render "No other projects available". This is the only branch of `shouldShowSwitchProject` not covered by the current tests.

romandubovyi (changes requested):

romandubovyi (approved):

PR #1281, opened 2026-06-04
feat(web): make Google Sheets access emails visible and copyable for OAuth and Service Account methods

romandubovyi:
This pull request introduces functionality to display and enable copying of the access emails for Google Sheets integration. It applies to both OAuth and Service Account methods, improving usability for handling configuration needs.

## What's new
 - Added display of the email address that needs to be granted access to a Google Sheet document.
 - This now works for **both** authentication methods:
   - **Service Account** — `client_email` from the JSON key.
   - **Google OAuth** — email of the authenticated Google account.
 - The email is **copyable with one click** in both the destination form and the report form.
 - Updated the dropdown in the report form to show the email regardless of auth method.

 ## How to test

 1. **Service Account**
    - Create or edit a Google Sheets destination with a Service Account JSON.
    - Observe the email is shown as a link to Google Cloud Console with a copy button.
    - Create a report using this destination — the email is visible in the dropdown and below it.

 2. **OAuth**
    - Connect a Google Sheets destination via OAuth.
    - In the destination form, the "Authenticated email" appears under the OAuth button.
    - Create a report using this destination — the email is visible in the dropdown and below it.

 3. **Copy functionality**
    - Click the email field (or the copy icon for SA) — it should be copied to clipboard.

### DIFF
--- a/apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
+++ b/apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
@@ -68,9 +68,10 @@ export function GoogleSheetsFields({ form }: GoogleSheetsFieldsProps) {
   const credentialIdValue = form.watch('credentials.credentialId');
 
   useEffect(() => {
+    const abortController = new AbortController();
     if (authMethod === 'oauth' && credentialIdValue) {
       destinationOAuthApi
-        .getCredentialStatus(credentialIdValue)
+        .getCredentialStatus(credentialIdValue, { signal: abortController.signal })
         .then(status => {
           setOauthEmail(status.user?.email ?? null);
         })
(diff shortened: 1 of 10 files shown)


apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
```diff
@@ -67,6 +67,21 @@ export function GoogleSheetsFields({ form }: GoogleSheetsFieldsProps) {
 
   const credentialIdValue = form.watch('credentials.credentialId');
 
+  useEffect(() => {
+    if (authMethod === 'oauth' && credentialIdValue) {
+      destinationOAuthApi
```
max-voloshyn, line 73: Please guard this async status lookup against stale responses. If the user reconnects or switches credentials while an earlier `getCredentialStatus` request is still in flight, the older request can resolve last and overwrite `oauthEmail`, so the copyable field may show/share the previous Google account. Track an `isCurrent` flag or abort the request in the effect cleanup, and only call `setOauthEmail` for the latest `credentialIdValue`.

🤖 Reviewed by Codex (GPT-5.5)

romandubovyi: Good catch! Fixed — now aborting the in-flight request in the effect cleanup so stale responses can't overwrite the email

max-voloshyn (approved):

PR #1282, opened 2026-06-05
Add run cancellation for connector and report runs

kalnik-a-a:
## Motivation

Connector and standard report runs can be left in non-terminal states when a worker goes down or when an in-progress trigger cannot be stopped from the UI. Users need a Run History action that can request cancellation, clear stuck run records from the active state, and stop the underlying trigger work when it is still running.

## What changed

- Extends Data Mart run cancellation to connector runs and standard report runs: Google Sheets, Email, Slack, MS Teams, and Google Chat.
- Keeps HTTP Data and Looker Studio runs out of this flow because they use separate execution models.
- Moves cancellation state changes through the relevant services instead of updating repositories directly in the use case.
- Marks active Data Mart runs as `CANCELLED` conditionally, so completed runs are not cancelled by stale requests.
- Stops matching connector/report run triggers by moving queued triggers to `CANCELLED` and in-progress triggers to `CANCELLING`.
- Relies on the existing scheduler abort job to fetch `CANCELLING` triggers and call `runner.abortTriggerRuns(...)` on active trigger controllers.
- Persists trigger cancellation with atomic status updates, avoiding stale entity saves after abort.
- Keeps cancelled runs from being overwritten by graceful-shutdown `INTERRUPTED` recovery.
- Preserves successful connector completion, including consumption tracking and success event registration, when an abort arrives after the connector upload already succeeded.
- Updates report last-run state only from `RUNNING` to `CANCELLED`.
- Removes the old unused connector cancellation chain from `ConnectorExecutionService`/`ConnectorRunService`.
- Adds a Run History cancel action with confirmation and an explicit warning that cancelling a run may leave incomplete data.

## Behavior
CleanShot 2026-06-05 at 16 40 48

- The cancel action is shown only for connector and standard report runs in `PENDING` or `RUNNING` status.
- Cancellation is best-effort for live runs. The endpoint marks the Run History record as `CANCELLED` immediately, so stale/orphaned runs that no longer have a live worker stop showing as active.
- If the worker is still live, queued triggers are cancelled, in-progress triggers move to `CANCELLING`, and the scheduler abort job delivers the abort signal to the active runner.
- If a live worker has already passed the cancellation point and completes before the abort signal is delivered/observed, its normal finalizer may replace `CANCELLED` with `SUCCESS` or `FAILED`. In that case Run History shows the actual completed outcome instead of preserving a cancellation marker for work that already finished.
- If a connector upload already succeeded before a late abort is observed, the run remains successful and consumption tracking is still recorded.
- HTTP Data and Looker Studio runs remain non-cancellable.

## Verification

- `npm test -w @owox/backend -- --runInBand src/common/scheduler/facades/scheduler-facade.impl.spec.ts src/data-marts/services/connector/connector-run-trigger-handler.service.spec.ts src/data-marts/services/report-run-trigger-handler.service.spec.ts src/data-marts/services/connector/connector-executor.service.spec.ts src/data-marts/services/report.service.spec.ts src/data-marts/services/connector/connector-run.service.spec.ts src/data-marts/services/connector/connector-execution.service.spec.ts src/data-marts/use-cases/cancel-data-mart-run.service.spec.ts src/data-marts/services/data-mart-run.service.spec.ts`
- `npm test -w @owox/web -- RunItem.test.tsx data-mart.service.test.ts`
- `npx prettier --check <touched files>`
- `npm run lint -w @owox/web`

### DIFF
--- a/apps/backend/src/data-marts/services/base-run-trigger-handler.service.ts
+++ b/apps/backend/src/data-marts/services/base-run-trigger-handler.service.ts
@@ -1,8 +1,10 @@
 import { Logger, OnModuleDestroy, OnModuleInit } from '@nestjs/common';
-import { Repository } from 'typeorm';
+import { FindOptionsWhere, In, Repository } from 'typeorm';
+import { QueryDeepPartialEntity } from 'typeorm/query-builder/QueryPartialEntity';
 import { SchedulerFacade } from '../../common/scheduler/shared/scheduler.facade';
 import { TriggerHandler } from '../../common/scheduler/shared/trigger-handler.interface';
 import { Trigger } from '../../common/scheduler/shared/entities/trigger.entity';
+import { TriggerStatus } from '../../common/scheduler/shared/entities/trigger-status';
 import { DataMartRun } from '../entities/data-mart-run.entity';
 import { DataMartRunStatus } from '../enums/data-mart-run-status.enum';
 import { DataMartRunService } from './data-mart-run.service';
@@ -46,6 +48,42 @@ export abstract class BaseRunTriggerHandlerService<T extends Trigger>
    * Returns the field name used to link trigger to DataMartRun.
    */
   protected abstract getTriggerRunIdField(): string;
+
+  protected async cancelTriggerIfRunAlreadyCancelled(trigger: T): Promise<boolean> {
+    const dataMartRunId = this.getTriggerDataMartRunId(trigger);
+    const existingRun = await this.dataMartRunService.findById(dataMartRunId);
+    if (existingRun?.status !== DataMartRunStatus.CANCELLED) {
+      return false;
+    }
+
+    await this.markTriggerAsCancelled(trigger);
+    return true;
+  }
+
+  protected async markTriggerAsCancelled(trigger: T): Promise<void> {
+    trigger.status = TriggerStatus.CANCELLED;
+    trigger.isActive = false;
+
+    await this.getTriggerRepository().update(
+      {
+        id: trigger.id,
+        status: In([TriggerStatus.PROCESSING, TriggerStatus.CANCELLING]),
+      } as FindOptionsWhere<T>,
+      {
+        status: TriggerStatus.CANCELLED,
+        isActive: false,
+        version: () => 'version + 1',
+      } as QueryDeepPartialEntity<T>
+    );
+
+    this.logger.log(
+      `Skipping run trigger ${trigger.id}: DataMartRun ${this.getTriggerDataMartRunId(trigger)} is already CANCELLED`
+    );
+  }
+
+  private getTriggerDataMartRunId(trigger: T): string {
+    return (trigger as unknown as Record<string, string>)[this.getTriggerRunIdField()];
+  }
 
   /**
    * Safely marks a DataMartRun as FAILED, handling any errors during the operation.
--- a/apps/backend/src/data-marts/services/base-run-trigger-handler.service.ts
+++ b/apps/backend/src/data-marts/services/base-run-trigger-handler.service.ts
@@ -56,11 +56,14 @@ export abstract class BaseRunTriggerHandlerService<T extends Trigger>
       return false;
     }
 
-    await this.markTriggerAsCancelled(trigger);
+    await this.markTriggerAsCancelled(
+      trigger,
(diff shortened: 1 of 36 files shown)


apps/backend/src/data-marts/services/base-run-trigger-handler.service.ts
```diff
@@ -47,6 +49,42 @@ export abstract class BaseRunTriggerHandlerService<T extends Trigger>
...
+      {
+        status: TriggerStatus.CANCELLED,
+        isActive: false,
+        version: () => 'version + 1',
+      } as QueryDeepPartialEntity<T>
+    );
+
+    this.logger.log(
+      `Skipping run trigger ${trigger.id}: DataMartRun ${this.getTriggerDataMartRunId(trigger)} is already CANCELLED`
+    );
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in a45098226: trigger cancellation logging now uses an explicit reason, so abort-signal cancellation no longer logs as "run already CANCELLED". Added coverage for the abort path.

apps/backend/src/data-marts/use-cases/cancel-data-mart-run.service.ts
```diff
@@ -31,17 +48,57 @@ export class CancelDataMartRunService {
       }
     }
 
-    if (dataMart.definitionType !== DataMartDefinitionType.CONNECTOR) {
-      throw new Error('Only data marts with connector definition can be cancelled');
+    await this.cancelRunState(command);
```
romandubovyi, line 51: `cancelRunState` is public but the authorization check lives only in `run()`. Any caller that invokes `cancelRunState` directly would bypass the access decision. Since `typeorm-transactional`'s `@Transactional()` transforms the method descriptor (not a proxy), it works on private methods too — consider making this `private`.

apps/backend/src/data-marts/services/connector/connector-run-trigger.service.ts
```diff
@@ -36,4 +36,15 @@ export class ConnectorRunTriggerService {
     const saved = await this.repository.save(trigger);
     return saved.id;
   }
+
```
romandubovyi, line 40: nit: `stopTriggersForRun()` is duplicated verbatim in `ReportRunTriggerService`. If the trigger cancellation logic changes (e.g., new statuses), both places would need updating. Consider extracting to a shared helper or base class.

apps/web/src/features/data-marts/edit/model/context/DataMartContext.tsx
```diff
@@ -549,6 +549,7 @@ export function DataMartProvider({ children }: DataMartProviderProps) {
           action: 'CancelRunError',
           error: apiError.message,
         });
+        throw error;
```
max-voloshyn, line 544: [P2, non-blocking] Don’t rethrow refresh failures as cancellation failures. This `try` includes both the cancel POST and the follow-up `getDataMartRuns(id)` refresh, so if the POST returns 204 but the refresh fails, `CancelRunButton` reports “failed to cancel” and keeps the confirmation open even though the run was already cancelled. This should not block merge, but separating the cancel request from the refresh, or making the refresh best-effort/silent, would avoid misleading users.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/services/connector/connector-run-trigger.service.ts
kalnik-a-a: Addressed in 880b648f1: connector and report trigger cancellation now use a shared helper, typed to the two supported run-trigger entities.

apps/backend/src/data-marts/use-cases/cancel-data-mart-run.service.ts
kalnik-a-a: Addressed in 880b648f1: `cancelRunState` is now private, so cancellation state changes stay behind `run()` and its access check.

apps/web/src/features/data-marts/edit/model/context/DataMartContext.tsx
kalnik-a-a: Addressed in 880b648f1: the cancel POST is now handled separately from the Run History refresh, so a refresh failure no longer reports cancellation failure or keeps the dialog open. Added regression coverage for this path.

max-voloshyn (approved):

PR #1285, opened 2026-06-05
Improve API key expiration visibility and sorting

max-voloshyn:
## Summary

- Sort API key expiration dates chronologically, with `Never` treated as the furthest expiration date.
- Make expired and soon-to-expire API keys easier to scan in the table.
- Reuse the same expiration representation in the API key details sidebar.
- Show expiration state as a compact badge after the neutral date text.

## Verification

- Ran focused API key table/sidebar tests.
- Ran ESLint on touched files.
- Ran `git diff --check`.
- Ran web build.
- Verified table and sidebar rendering in the browser.

### DIFF
--- a/apps/web/src/features/api-keys/components/ApiKeysTable/ApiKeysTable.test.tsx
+++ b/apps/web/src/features/api-keys/components/ApiKeysTable/ApiKeysTable.test.tsx
@@ -121,8 +121,18 @@ describe('ApiKeysTable', () => {
     );
 
     const expirationDate = screen.getByText(formatDateOnly(expiresAt, { timeZone: 'UTC' }));
+    const expiresCell = expirationDate.closest('td');
+    expect(expiresCell).not.toBeNull();
+    const statusBadge = within(expiresCell as HTMLElement).getByText('Expires soon');
 
-    expect(expirationDate).toHaveClass('font-medium', 'text-amber-600');
+    expect(expirationDate).toHaveClass('text-foreground');
+    expect(expirationDate).not.toHaveClass('text-amber-600');
+    expect(expirationDate).not.toHaveClass('font-medium');
+    expect(statusBadge).toHaveAttribute('data-slot', 'badge');
+    expect(statusBadge).toHaveClass('text-[11px]', 'text-amber-700');
+    expect(
+      expirationDate.compareDocumentPosition(statusBadge) & Node.DOCUMENT_POSITION_FOLLOWING
+    ).toBeTruthy();
     expect(screen.getByRole('tooltip')).toHaveTextContent('This API key expires within 30 days.');
   });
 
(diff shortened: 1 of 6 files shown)


apps/web/src/features/api-keys/components/ApiKeysTable/ApiKeysTable.test.tsx
```diff
@@ -120,4 +125,85 @@ describe('ApiKeysTable', () => {
...
+
+    expect(expirationDate).toHaveClass('font-medium', 'text-destructive');
+    expect(screen.getByRole('tooltip')).toHaveTextContent('This API key has expired.');
+    expect(container.querySelector('svg.lucide-circle-alert')).toBeNull();
+  });
+
+  it('sorts keys that never expire as the furthest expiration date', () => {
+    const keys: ProjectMemberApiKey[] = [
+      { ...key, apiKeyId: 'pmk_never_1', name: 'Never 1', expiresAt: null },
+      {
```
romandubovyi, line 176: nit: Consider using `pastIso()`/`futureIso()` helpers (already defined above) instead of hardcoded ISO strings. This would keep the sorting test consistent with the other tests in this file and make the "Expired"/"Future soon"/"Future later" labels always semantically accurate regardless of when the test runs.

apps/web/src/features/api-keys/components/EditApiKeySheet.tsx
```diff
@@ -107,7 +110,18 @@ export function EditApiKeySheet({ apiKey, onClose, onUpdated, onRevoke }: EditAp
   const expiresAt = apiKey?.expiresAt
     ? formatDateOnly(apiKey.expiresAt, { timeZone: 'UTC' })
     : 'Never';
+  const expired = isApiKeyExpired(apiKey?.expiresAt);
```
romandubovyi: nit: Both this file and `columns.tsx` now share the same expired → expiring-soon → normal branching pattern. Fine with two call sites, but if a third appears consider extracting a shared `getApiKeyExpirationState()` helper that returns `{ className, notice } | null`.

romandubovyi (approved):

apps/web/src/features/api-keys/components/ApiKeysTable/ApiKeysTable.test.tsx
max-voloshyn: Adjusted in 00040dfaa. The sorting test now uses `pastIso()` / `futureIso()` for expired, soon, and later dates instead of fixed 2026 timestamps.

apps/web/src/features/api-keys/components/EditApiKeySheet.tsx
max-voloshyn: This became outdated after the shared `ApiKeyExpirationValue` component was introduced. The table and details sheet no longer duplicate the expired → expiring-soon → normal rendering branch, so I did not add a separate state helper.

romandubovyi (approved):

PR #1287, opened 2026-06-07
Feature suggestion: project-level Data Mart activity pages

kalnik-a-a:
## Feature Suggestion

This draft PR proposes adding project-level Data Mart activity pages. The idea is to keep the existing per-Data-Mart tabs, but also give users a faster way to understand what is configured and what is currently happening across the whole project.

Today the user often has to open individual Data Marts to answer project-level questions like "what is scheduled?", "which reports and insights exist?", or "what has failed recently?". These pages make those answers available from the left navigation while preserving the existing detail workflows.

## Access Model

The project-level lists keep the current access model in place.

- All new list endpoints require the existing viewer-level auth context.
- Admins can see the full project; non-admin users are filtered through the same role scope and Data Mart visibility rules used elsewhere.
- Reports are returned with existing action capabilities: `canRun`, `canManageTriggers`, and `canEditConfig`.
- Trigger edit/delete availability is computed per row, including report-operability checks for report runs and technical-user checks for connector runs.
- Insights compute delete availability through `AccessDecisionService` on the parent Data Mart.
- Run History is read-only and only lists runs from visible Data Marts.

This keeps the pages useful as project-wide overviews without turning them into permission bypasses or metadata leaks.

### Triggers

Adds a project-level Triggers page for scheduled Data Mart work. It shows the Data Mart first, then the schedule, run target, next/last run context, status, and row actions. The page also supports the same search/filter/table treatment as the Data Mart list, and trigger rows can open the same side panel used from the Data Mart tab.

Screenshots:

triggers-empty

triggers-demo

### Reports

Adds a project-level Reports page for all accessible reports. It keeps the Data Mart as the first column, shows the report title or a muted dash when the title is absent, keeps the existing report actions available from the row menu, and opens the report side panel for editing from this cross-project view.

Screenshots:

reports-empty
reports-demo

### Insights

Adds a project-level Insights page for accessible insight templates. The Data Mart is the first column, insight title links to the insight detail view, and row actions mirror the per-Data-Mart tab where applicable. The page intentionally omits columns that are not present on the tab, such as sources.

Screenshots:

insights-empty
insights-demo

### Run History

Adds a project-level Run History page for recent runs across visible Data Marts. Each row keeps the Data Mart association visible, uses connector/report icons for scanability, includes run target/status details, and continues polling while active runs are present so the overview stays current.

Screenshots:

run-history-empty
run-history-demo

## Verification

- `npm run test -w @owox/web -- DataMartSchedulesPage DataMartReportsPage DataMartInsightsPage DataMartRunsPage`
- `npm run test -w @owox/backend -- --runInBand --testPathPatterns='project|scheduled|report|insight|data-mart-run'`

### DIFF
--- a/apps/backend/src/data-marts/controllers/project-reports.controller.ts
+++ b/apps/backend/src/data-marts/controllers/project-reports.controller.ts
@@ -1,0 +1,42 @@
+import { Controller, Get, ParseEnumPipe, Query } from '@nestjs/common';
+import { ApiTags } from '@nestjs/swagger';
+import { Auth, AuthContext, AuthorizationContext, Role, Strategy } from '../../idp';
+import { ReportResponseApiDto } from '../dto/presentation/report-response-api.dto';
+import { OwnerFilter } from '../enums/owner-filter.enum';
+import { ReportMapper } from '../mappers/report.mapper';
+import { ListReportsByProjectService } from '../use-cases/list-reports-by-project.service';
+import { normalizeProjectListPagination } from '../utils/normalize-project-list-pagination';
+import { ListReportsByProjectSpec } from './spec/project-reports.api';
+
+@Controller('reports')
+@ApiTags('Reports')
+export class ProjectReportsController {
+  constructor(
+    private readonly listReportsByProjectService: ListReportsByProjectService,
+    private readonly mapper: ReportMapper
+  ) {}
+
+  @Auth(Role.viewer(Strategy.PARSE))
+  @Get()
+  @ListReportsByProjectSpec()
+  async list(
+    @AuthContext() context: AuthorizationContext,
+    @Query('limit') limit?: string | number,
+    @Query('offset') offset?: string | number,
+    @Query('ownerFilter', new ParseEnumPipe(OwnerFilter, { optional: true }))
+    ownerFilter?: OwnerFilter
+  ): Promise<ReportResponseApiDto[]> {
+    const pagination =
+      limit === undefined && offset === undefined
+        ? null
+        : normalizeProjectListPagination(limit, offset);
+    const command = this.mapper.toListByProjectCommand(
+      context,
+      ownerFilter,
+      pagination?.limit,
+      pagination?.offset
+    );
+    const reports = await this.listReportsByProjectService.run(command);
+    return this.mapper.toResponseList(reports);
+  }
+}
--- a/apps/backend/src/data-marts/use-cases/list-project-insight-templates.service.ts
+++ b/apps/backend/src/data-marts/use-cases/list-project-insight-templates.service.ts
@@ -1,0 +1,51 @@
+import { Injectable } from '@nestjs/common';
+import { ListProjectInsightTemplatesCommand } from '../dto/domain/list-project-insight-templates.command';
+import { ProjectInsightTemplateDto } from '../dto/domain/project-insight-template.dto';
+import { RoleScope } from '../enums/role-scope.enum';
+import { InsightTemplateMapper } from '../mappers/insight-template.mapper';
+import { ContextAccessService } from '../services/context/context-access.service';
+import { InsightTemplateService } from '../services/insight-template.service';
+import { UserProjectionsFetcherService } from '../services/user-projections-fetcher.service';
+
+@Injectable()
+export class ListProjectInsightTemplatesService {
+  constructor(
+    private readonly insightTemplateService: InsightTemplateService,
+    private readonly contextAccessService: ContextAccessService,
+    private readonly userProjectionsFetcherService: UserProjectionsFetcherService,
+    private readonly mapper: InsightTemplateMapper
+  ) {}
+
+  async run(command: ListProjectInsightTemplatesCommand): Promise<ProjectInsightTemplateDto[]> {
+    const isAdmin = command.roles.includes('admin');
+    const roleScope = isAdmin
+      ? RoleScope.ENTIRE_PROJECT
+      : await this.contextAccessService.getRoleScope(command.userId, command.projectId);
+
+    const insightTemplates = await this.insightTemplateService.listVisibleByProject({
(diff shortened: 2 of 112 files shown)


apps/web/src/features/data-marts/reports/shared/services/report.service.ts
```diff
@@ -68,8 +68,16 @@ export class ReportService extends ApiService {
...
+  async getReportsByProject(limit?: number, offset?: number): Promise<ReportResponseDto[]> {
+    const params =
+      limit === undefined
+        ? undefined
+        : {
+            limit,
+            offset: offset ?? 0,
+          };
+
+    return this.get<ReportResponseDto[]>('/', params);
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in 0349857cf. The service now builds query params when either limit or offset is provided, and unit tests cover offset-only pagination.

apps/backend/src/data-marts/use-cases/list-project-scheduled-triggers.service.ts
```diff
@@ -0,0 +1,180 @@
...
+      })
+    );
+  }
+
+  private async canManageTrigger(
+    trigger: DataMartScheduledTrigger,
+    command: ListProjectScheduledTriggersCommand
+  ): Promise<boolean> {
+    if (trigger.type !== ScheduledTriggerType.REPORT_RUN) {
+      return this.reportAccessService.isTechnicalUser(command.roles);
```
max-voloshyn: This marks every editor as able to edit/delete connector-run triggers as soon as the Data Mart is visible. For an editor who only has reporting access (`availableForReporting=true`, `availableForMaintenance=false`), the access matrix denies `MANAGE_TRIGGERS`, but the project Triggers page will show destructive actions and the write path can proceed through the SEE + technical-user checks. Please derive this from the Data Mart trigger access decision (`MANAGE_TRIGGERS`) and align the update/delete guard if needed.
🤖 Reviewed by Codex (GPT-5.5)

Obolrus:
Great proposal! We actually see a lot of user requests for the ability to view reports/triggers/insights in a single, filtered list—without having to open and search for them within each data mart individually.
The only suggestion I have is to adjust the nesting level. Let's make them sub-items under Data Marts (collapsible section), as these entities are essentially derived from the data marts themselves.

The order should be the following:
1. Report
2. Insights
3. Triggers
4. Run History
CleanShot 2026-06-08 at 11 42 22@2x

apps/backend/src/data-marts/use-cases/list-project-scheduled-triggers.service.ts
kalnik-a-a: Fixed in 136727a06. Project trigger capabilities and connector-run create/update/delete guards now use MANAGE_TRIGGERS through AccessDecisionService, with denied-editor coverage.

kalnik-a-a:
> make them sub-items under Data Marts (collapsible section)

@Obolrus  Implemented in 7ec00e7a3. Reports, Insights, Triggers, and Run History now live under a collapsible Data Marts section in that order; direct child routes auto-open it, manual expansion is toggle-only, and the state is persisted via storageService.

apps/backend/src/data-marts/use-cases/list-project-insight-templates.service.ts
```diff
@@ -0,0 +1,64 @@
...
+      roles: command.roles,
+      roleScope,
+      limit: command.limit,
+      offset: command.offset,
+    });
+
+    const userProjections =
+      await this.userProjectionsFetcherService.fetchRelevantUserProjections(insightTemplates);
+
+    return Promise.all(
```
romandubovyi, line 39: This `Promise.all(items.map(item => canAccess(...)))` pattern creates N+1 access checks — one per entity. Since `applyDataMartVisibilityFilter` already handles visibility at the query level, consider extending that approach to cover access decisions too, or adding a batch `canAccessMany()` method to `AccessDecisionService` to reduce the number of individual checks.

apps/backend/src/data-marts/controllers/project-reports.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+  @ListReportsByProjectSpec()
+  async list(
+    @AuthContext() context: AuthorizationContext,
+    @Query('limit') limit?: string | number,
+    @Query('offset') offset?: string | number,
+    @Query('ownerFilter', new ParseEnumPipe(OwnerFilter, { optional: true }))
+    ownerFilter?: OwnerFilter
+  ): Promise<ReportResponseApiDto[]> {
+    const pagination =
+      limit === undefined && offset === undefined
```
romandubovyi, line 30: nit: This endpoint returns a plain array while the other new project-level endpoints (triggers, insights, runs) return wrapped objects with metadata. Consider aligning the response shape for consistency and easier frontend consumption.

max-voloshyn (approved):

apps/backend/src/data-marts/use-cases/list-project-insight-templates.service.ts
kalnik-a-a: Agreed, there is a performance risk here. In this PR we kept the existing access/capability service contracts and applied visibility filtering at the query level. A proper fix should batch access decisions in the shared access layer, so I’d address that in a follow-up refactor rather than expanding this feature branch.

apps/backend/src/data-marts/controllers/project-reports.controller.ts
kalnik-a-a: This endpoint is an existing GET /api/reports contract that already returned a plain array before this PR. This branch extends it with pagination/project-list usage while preserving the response shape for existing consumers. I’d keep it unchanged here and consider a wrapped shape only as a separate API change or new endpoint.

PR #1289, opened 2026-06-08
Fix SQLite scheduled run transaction stability

kalnik-a-a:
## Summary
- Add SQLite-only serialization for top-level TypeORM transactions when the backend uses `better-sqlite3`.
- Preserve nested transaction behavior and both TypeORM transaction call forms.
- Add regression coverage for concurrent scheduled-run style transaction contention, idempotent wrapping, isolation-level overloads, and non-SQLite passthrough.

## Root Cause
Local SQLite deployments use TypeORM with the `better-sqlite3` driver. That driver keeps a single query runner for the data source, while scheduled trigger processing can run multiple ready triggers in parallel. When several scheduled run flows enter `@Transactional()` or `dataSource.transaction()` at the same time, TypeORM can try to manage multiple transaction and savepoint stacks on that same SQLite query runner.

Under unlucky interleavings this can surface as SQLite transaction/savepoint errors, for example `no such savepoint` during `RELEASE SAVEPOINT`, or similar nested transaction failures. MySQL does not have the same shape here because TypeORM can use separate query runners/connections from the pool for concurrent transactions.

## Fix
For `better-sqlite3` data sources, wrap `dataSource.transaction()` with a small FIFO queue. Independent top-level transactions are executed one at a time, while nested transactions inside the currently running transaction bypass the queue through `AsyncLocalStorage`. That preserves TypeORM nested transaction behavior and avoids a transaction waiting on its own parent.

Non-SQLite data sources are returned unchanged, so MySQL keeps its existing concurrent transaction behavior.

## Product Impact
This changes rare concurrent SQLite transaction collisions from immediate failures into bounded waiting before a transaction starts. In those rare cases another transaction may wait for the current transaction callback to finish, but the scheduled run flow can complete instead of failing and leaving the trigger unprocessed.

The serializer only covers the transactional callback itself. The expensive report or connector execution is not placed under one long global lock by this change, so the expected user-facing impact is limited to short SQLite transaction coordination windows during concurrent background activity.

## Test Plan
- `npm test -w @owox/backend -- sqlite-transaction-serializer.spec.ts --runInBand`
- `npx tsc -p apps/backend/tsconfig.build.json --noEmit`
- `npx eslint src/config/sqlite-transaction-serializer.ts src/config/sqlite-transaction-serializer.spec.ts src/app.module.ts --config ./eslint.config.mjs`

### DIFF
--- a/apps/backend/src/app.module.ts
+++ b/apps/backend/src/app.module.ts
@@ -16,6 +16,7 @@ import { validateConfig } from './config/env-validation.config';
 import { ClsModule } from 'nestjs-cls';
 import { DataSource } from 'typeorm';
 import { addTransactionalDataSource } from 'typeorm-transactional';
+import { serializeSqliteTransactions } from './config/sqlite-transaction-serializer';
 
 @Module({
   imports: [
@@ -40,7 +41,7 @@ import { addTransactionalDataSource } from 'typeorm-transactional';
         }
         const dataSource = new DataSource(options);
         await dataSource.initialize();
-        return addTransactionalDataSource(dataSource);
+        return serializeSqliteTransactions(addTransactionalDataSource(dataSource));
       },
     }),
     ScheduleModule.forRoot(),
--- a/apps/backend/src/config/sqlite-transaction-serializer.spec.ts
+++ b/apps/backend/src/config/sqlite-transaction-serializer.spec.ts
@@ -1,5 +1,13 @@
+import 'reflect-metadata';
 import { DataSource, EntityManager } from 'typeorm';
 import { IsolationLevel } from 'typeorm/driver/types/IsolationLevel';
+import {
+  addTransactionalDataSource,
+  deleteDataSourceByName,
+  initializeTransactionalContext,
+  StorageDriver,
+  Transactional,
+} from 'typeorm-transactional';
 import { serializeSqliteTransactions } from './sqlite-transaction-serializer';
 
 describe('serializeSqliteTransactions', () => {
@@ -33,11 +41,16 @@ describe('serializeSqliteTransactions', () => {
     } as unknown as DataSource;
   };
 
+  const wait = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
+  beforeAll(() => {
+    initializeTransactionalContext({ storageDriver: StorageDriver.AUTO });
+  });
+
   it('serializes concurrent better-sqlite3 transactions while preserving nested transactions', async () => {
     const dataSource = createDataSource('better-sqlite3');
     serializeSqliteTransactions(dataSource);
 
-    const wait = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
     const activeTopLevelTransactions = new Set<string>();
     const executionOrder: string[] = [];
     let maxActiveTopLevelTransactions = 0;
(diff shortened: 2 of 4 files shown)


apps/backend/src/config/sqlite-transaction-serializer.ts
```diff
@@ -0,0 +1,78 @@
...
+    isolationOrRunInTransaction: IsolationLevel | RunInTransaction<T>,
+    runInTransaction?: RunInTransaction<T>
+  ): Promise<T> => {
+    if (typeof isolationOrRunInTransaction === 'function') {
+      return originalTransaction(isolationOrRunInTransaction);
+    }
+
+    return originalTransaction(isolationOrRunInTransaction, runInTransaction!);
+  };
+
```
romandubovyi, line 60: This wraps `dataSource.transaction()`, but `@Transactional()` from `typeorm-transactional` may manage transactions via its own query runners (`queryRunner.startTransaction()`), bypassing this wrapper entirely. Could you verify that the scheduled-run transaction flows actually go through `dataSource.transaction()` rather than `@Transactional()`? If both paths are used, the serializer may need to also intercept `dataSource.createQueryRunner()` or hook into the `typeorm-transactional` internals.

apps/backend/src/config/sqlite-transaction-serializer.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  it('does not wrap non-sqlite data sources', async () => {
+    const dataSource = createDataSource('mysql');
+    const transaction = dataSource.transaction;
+
+    serializeSqliteTransactions(dataSource);
+
+    expect(dataSource.transaction).toBe(transaction);
+  });
+});
```
romandubovyi, line 207: nit: Consider adding a test that verifies queue recovery after a transaction rejection — e.g., the first transaction throws, and the second one still executes and resolves successfully. The `finally`/`catch` pattern in `AsyncTransactionQueue` is critical for liveness; a stuck queue would block all subsequent transactions indefinitely.

apps/backend/src/config/sqlite-transaction-serializer.ts
kalnik-a-a: Checked this against the installed `typeorm-transactional@0.5.0` implementation and our AppModule wiring.

`typeorm-transactional` opens managed transactions through `dataSource.transaction(...)`. In our setup, `addTransactionalDataSource(dataSource)` registers the DataSource object, and `serializeSqliteTransactions(...)` then wraps that same object, so `@Transactional()` flows and direct `dataSource.transaction(...)` flows both go through the SQLite serializer.

I also checked the codebase for direct `createQueryRunner().startTransaction()` usage. The regular application transaction paths I found use either `@Transactional()` or `dataSource.transaction(...)`; the direct query-runner transaction usage is isolated to the dump/apply utility. Because of that, wrapping `createQueryRunner()` would broaden the TypeORM monkey patch into a lower-level API without covering an active product transaction path today.

Added regression coverage for the `@Transactional` path as well: the test registers a fake sqlite DataSource through the real `typeorm-transactional` registry and verifies that concurrent decorated method calls are serialized through the wrapped `dataSource.transaction`.

apps/backend/src/config/sqlite-transaction-serializer.spec.ts
kalnik-a-a: Added a regression test for this liveness path: when the first queued transaction rejects, the queue is released and the next queued transaction still executes and resolves successfully.

apps/backend/src/app.module.ts
```diff
@@ -40,7 +41,7 @@ import { addTransactionalDataSource } from 'typeorm-transactional';
         }
         const dataSource = new DataSource(options);
         await dataSource.initialize();
-        return addTransactionalDataSource(dataSource);
+        return serializeSqliteTransactions(addTransactionalDataSource(dataSource));
```
max-voloshyn, line 44: Non-blocking: this queues every top-level SQLite transaction, not only the scheduled-run claim transactions. That means existing transactional flows that call remote services inside the transaction, such as `CreateReportService.run()` validating Google Sheets access, can hold the global SQLite queue longer than intended and delay unrelated scheduled-run claim transactions. This does not look like a correctness blocker for this PR, but it would be safer to shrink those existing transaction scopes or follow up with narrower serialization around the short SQLite transaction paths.

🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (approved):

apps/backend/src/app.module.ts
kalnik-a-a: Acknowledged. The entry point for this PR was scheduled runs, but the underlying failure mode is not scheduled-run-specific: with `better-sqlite3`/TypeORM, any two concurrent top-level transactions can interleave on the shared SQLite transaction machinery and break savepoint handling. For that reason the serializer is intentionally applied at the SQLite DataSource transaction boundary rather than only around scheduled-run claim transactions.

The trade-off is as described: if an existing transaction includes slower work, unrelated SQLite transactions may wait behind it. For this PR I think that is preferable to the current behavior, where the same concurrency can fail with SQLite savepoint errors. The wrapper is SQLite-only and top-level only; MySQL remains unchanged.

I agree that shrinking long transaction scopes, especially flows that perform remote validation inside a transaction, would be a good follow-up. I would keep that separate from this SQLite stability fix because it has a wider behavioral surface than serializing the current transaction boundary.

PR #1290, opened 2026-06-08
feat(data-marts): output controls for AWS Redshift data marts

zapolsky:
Enable filters / slices / sort / limit for AWS Redshift, matching the Athena + BigQuery support. Redshift uses "option B": RedshiftClauseRenderer emits inline SQL literals (the Redshift Data API has no bound-parameter channel), so it always returns zero params — compose / composeStatic / run / stream / Looker-cache / copy / preview and the blended builder all work with no param plumbing. Literal escaping ('->'') is the sole injection barrier; live-verified that backslash is literal and that all five date/time types (DATE, TIMESTAMP, TIMESTAMPTZ, TIME, TIMETZ) coerce a bare literal — no CAST needed.

Backend: RedshiftClauseRenderer (19 operators, relative_date half-open ranges), non-blended + blended query builders wired to it, DI registration; validator, capability and FE operator type-sets gain the Redshift type names (TEXT, BPCHAR, DOUBLE PRECISION, TIMESTAMPTZ, TIMETZ). Subquery FROM clauses get an alias (Redshift/Postgres requires it); create-view and the query-debug log hardened.

Cross-storage: this_month / this_year relative_date presets now carry an upper bound in all three renderers (Athena, BigQuery, Redshift); FE "Last N" inputs are bounded to max 3650. FE filter type-matrix nits from PR #15 mirrored to web.

Tests: renderer operator matrix + injection probes, query/blended builder specs, validator + capability contract, e2e SQL-emission + Looker cache path, and a live Redshift integration suite (coercion, escaping, operator matrix, relative_date).

### DIFF
--- a/.changeset/redshift-output-controls.md
+++ b/.changeset/redshift-output-controls.md
@@ -5,5 +5,6 @@
 # Add output controls for Redshift data marts
 
 Add output controls (filters, slices, sort, limit) for AWS Redshift data marts, and
-bound the `this_month`/`this_year` relative-date presets to the current period across
-all storages.
+bound the relative-date presets so they no longer match future-dated rows across all
+storages: `this_month`/`this_year` are clamped to the current period, and
+`last_n_days`/`last_n_months` now stop at today (previously lower-bound only).
--- a/.changeset/redshift-output-controls.md
+++ b/.changeset/redshift-output-controls.md
@@ -1,0 +1,9 @@
+---
+'owox': minor
+---
+
+# Add output controls for Redshift data marts
+
+Add output controls (filters, slices, sort, limit) for AWS Redshift data marts, and
+bound the `this_month`/`this_year` relative-date presets to the current period across
+all storages.
--- a/.changeset/redshift-output-controls.md
+++ b/.changeset/redshift-output-controls.md
@@ -5,6 +5,7 @@
 # Add output controls for Redshift data marts
 
 Add output controls (filters, slices, sort, limit) for AWS Redshift data marts, and
-bound the relative-date presets so they no longer match future-dated rows across all
-storages: `this_month`/`this_year` are clamped to the current period, and
-`last_n_days`/`last_n_months` now stop at today (previously lower-bound only).
+tighten the relative-date presets across all storages: `last_n_days`/`last_n_months`
+now stop at today (previously lower-bound only, so future-dated rows leaked in), and
+`this_month`/`this_year` are clamped to the current period (a future date still inside
+the current month/year continues to match).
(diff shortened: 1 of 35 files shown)


apps/backend/src/data-marts/data-storage-types/redshift/services/redshift-clause-renderer.ts
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * Formats a value as a Redshift SQL literal. This is the ONLY barrier between user
+ * filter input and executed SQL — the Redshift Data API path has no bound-param
+ * channel, so the renderer inlines literals and escaping must be airtight. Single
+ * quotes are doubled (standard SQL). Assumes `standard_conforming_strings = on`
+ * (backslash is an ordinary character) — verified by the integration suite.
+ */
+function formatRedshiftLiteral(value: string | number | boolean | null): string {
+  if (value === null) return 'NULL';
+  if (typeof value === 'number') return String(value);
```
max-voloshyn: Blocking: because Redshift now inlines numeric literals, this needs to reject non-finite numbers before `String(value)`. The current filter schema uses plain `z.number()`, and parsing JSON numeric text like `1e309` becomes `Infinity` that Zod accepts, so report/HTTP Data filters can render invalid SQL like `WHERE "amount" > Infinity` instead of rejecting the request. Please guard with `Number.isFinite(value)` here or switch the scalar number schema to `.finite()`.
🤖 Reviewed by Codex (GPT-5.5)

.changeset/redshift-output-controls.md
```diff
@@ -0,0 +1,10 @@
+---
+'owox': minor
+---
+
+# Add output controls for Redshift data marts
+
+Add output controls (filters, slices, sort, limit) for AWS Redshift data marts, and
+bound the relative-date presets so they no longer match future-dated rows across all
```
max-voloshyn: Non-blocking: this release note slightly overstates the behavior. The new `this_month`/`this_year` bounds clamp to the end of the current month/year, so a future date inside the current period still matches. Consider rewording this to say only `last_n_*` stops at today and `this_*` is limited to the current period.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/data-storage-types/redshift/services/redshift-clause-renderer.ts
zapolsky: Good catch — fixed. Switched the scalar value schema to `z.number().finite()` so Infinity/NaN are rejected at the request boundary for every storage (scalar values and `between` bounds alike), and added a `Number.isFinite` guard in `formatRedshiftLiteral` as defense-in-depth since the Redshift path inlines the literal. Covered by schema + renderer unit tests. Fixed in 3b629ae61.

.changeset/redshift-output-controls.md
zapolsky: Reworded — you're right, `this_month`/`this_year` clamp to the current period (a future date inside it still matches); only `last_n_*` stops at today. 924d43c17.

max-voloshyn (approved):

PR #1292, opened 2026-06-08
ci: bump checkout and setup-node to v5 (Node 24 runtime)

zapolsky:
## Why

These workflows pin GitHub Actions that run on the **deprecated Node.js 20 runtime**. The runner flags **any** action whose `runs.using` is `node20`, regardless of author — so every run currently emits a Node 20 deprecation warning, and clearing it means moving *every* node20 action off node20.

GitHub timeline:

- **June 16, 2026** — runners default to the Node 24 runtime.
- **Fall 2026** — Node 20 removed from runners entirely.

Nothing breaks before then — this is CI hygiene to clear the warning.

## What

Bump every node20 action across all 12 workflows to a node24 release. Each target version was verified to declare `runs.using: node24` in its `action.yml`.

**Drop-in runtime bumps (no config change):**

- ✅ `actions/checkout@v4` → `@v5` (×12)
- ✅ `actions/setup-node@v4` → `@v5` (×12)
- ✅ `actions/github-script@v7` → `@v8`
- ✅ `actions/upload-artifact@v4` → `@v7` (×2)
- ✅ `docker/setup-buildx-action@v3` → `@v4`
- ✅ `docker/login-action@v3` → `@v4`
- ✅ `docker/build-push-action@v5` → `@v7` (×2)

**Already node24 — no change needed:** `changesets/action@v1` and `JamesIves/github-pages-deploy-action@v4` — their floating major tags already resolve to a node24 release (`JamesIves` → `v4.8.0`).

## Risk

Most bumps are pure runtime/dependency updates. The two to keep an eye on:

- `docker/build-push-action` v5 → v7 spans two majors. All inputs in use (`context`, `push`, `platforms`, `build-args`, `tags`) are unchanged, so no config breakage — but this is the publish path, so the **next publish run should be confirmed** to push the multi-arch image to GHCR as before.
- `actions/upload-artifact` v4 → v7: artifact names are unique per upload (`playwright-report`, `playwright-traces`); no removed inputs in use.

## Verification

- All 12 changed workflow files validated as parseable YAML.
- No functional changes — runtime bumps only.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout 🛎️
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup Node.js 🔧
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
           node-version: '22'
 
--- a/.github/workflows/e2e-browser.yml
+++ b/.github/workflows/e2e-browser.yml
@@ -53,7 +53,7 @@ jobs:
           HEADERS_TIMEOUT_MS: '185000'
 
       - name: Upload Playwright report
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         if: ${{ !cancelled() }}
         with:
           name: playwright-report
@@ -61,7 +61,7 @@ jobs:
           retention-days: 7
 
       - name: Upload Playwright traces
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         if: failure()
         with:
           name: playwright-traces
--- a/.github/workflows/e2e-browser.yml
+++ b/.github/workflows/e2e-browser.yml
@@ -22,10 +22,10 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
           node-version: '22'
           cache: 'npm'
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,10 +15,10 @@ jobs:
           fetch-depth: 0 # Required for changesets to work properly
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
 
       - name: Log in to GitHub Container Registry
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
@@ -158,7 +158,7 @@ jobs:
 
       - name: Build and push "latest" container image
         if: steps.latest_image_exists.outputs.exists == 'false'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           push: true
@@ -277,7 +277,7 @@ jobs:
           exit 1
 
       - name: Build and push "next" container image
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           push: true
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,7 +10,7 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0 # Required for changesets to work properly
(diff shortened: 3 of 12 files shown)


.github/workflows/e2e-browser.yml
```diff
@@ -22,10 +22,10 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
```
max-voloshyn, line 28: This workflow still has two `actions/upload-artifact@v4` steps below, and `v4` declares `runs.using: node20`. Since this PR is moving workflows off Node 20 action runtimes, please bump both artifact uploads to a Node 24 release (`actions/upload-artifact@v7`, or at least `@v6`) in the same PR.
🤖 Reviewed by Codex (GPT-5.5)

.github/workflows/publish.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           echo "repo_lower=$(echo '${{ github.repository }}' | tr '[:upper:]' '[:lower:]')" >> $GITHUB_OUTPUT
           cat $GITHUB_OUTPUT
 
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
```
max-voloshyn, line 33: This publish workflow will still run Node 20 JavaScript actions after the checkout/setup-node bump: `docker/setup-buildx-action@v3`, `docker/login-action@v3`, and both `docker/build-push-action@v5` steps still declare `runs.using: node20`. Please bump them as part of this migration (`setup-buildx@v4`, `login-action@v4`, `build-push-action@v7`) so the publish job is actually clear of Node 20 action runtimes.
🤖 Reviewed by Codex (GPT-5.5)

.github/workflows/deploy-docs.yml
```diff
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout 🛎️
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup Node.js 🔧
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
```
max-voloshyn, line 15: This workflow still deploys with `JamesIves/github-pages-deploy-action@v4`, whose floating `v4` action metadata currently declares `runs.using: node20`; the Node 24 migration is in later fixed v4 tags such as `v4.8.0`. Please pin this deploy step to a Node 24 tag/SHA or move to a Node 24 Pages deployment action, otherwise this workflow keeps the Node 20 warning after this PR.
🤖 Reviewed by Codex (GPT-5.5)

.github/workflows/test-integration.yml
```diff
@@ -37,10 +37,10 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
```
max-voloshyn, line 43: The scheduled-failure notifier still uses `actions/github-script@v7`, and `v7` declares `runs.using: node20` while `v8+` uses Node 24. This means the integration workflow can still emit the Node 20 deprecation warning even after checkout/setup-node are bumped; please update this step too, likely to `actions/github-script@v8` for the minimal runtime-only bump.
🤖 Reviewed by Codex (GPT-5.5)

.github/workflows/e2e-browser.yml
zapolsky: Done — both `upload-artifact` steps are now on `@v7` (node24). I widened the scope of this PR so it clears every node20 action in one pass rather than leaving a follow-up.

.github/workflows/publish.yml
zapolsky: Done — `setup-buildx-action@v3→v4`, `login-action@v3→v4`, and both `build-push-action@v5→v7`, all node24. `build-push` crosses two majors, but every input in use (`context`/`push`/`platforms`/`build-args`/`tags`) is unchanged, so no config change is needed; the next publish run should be watched to confirm the multi-arch push to GHCR is unaffected.

.github/workflows/deploy-docs.yml
zapolsky: The floating `v4` tag already resolves to `v4.8.0`, whose `action.yml` declares `runs.using: node24`, so this deploy step is not on the node20 runtime and does not emit the deprecation warning. Leaving it as-is — no change needed here.

.github/workflows/test-integration.yml
zapolsky: Done — `github-script@v7→v8` (node24). Only the stable `github.rest.*` API surface is used, so the bump is a drop-in runtime change.

max-voloshyn (approved):

PR #1293, opened 2026-06-08
feat(reports): show SQL preview for all reports regardless of definition type

romandubovyi:
Remove the info-tooltip mode from `GeneratedSqlViewer` — always open the SQL dialog.
Previously, reports without blending or output controls showed a dimmed icon with a "no custom SQL" tooltip instead of actual SQL. Now every report gets the full SQL preview dialog.

**Key changes:**

- Replaced per-row `useBlendedFieldNames` + `reportHasBlending` + `reportHasOutputControls` checks with a single `isGeneratedSqlSupported(definitionType, storageType)` gate that hides the SQL preview button only for `TABLE_PATTERN` definitions on storages that don't support them (everything except non-legacy Google BigQuery)
- Added a backend guard in `GetReportGeneratedSqlService` that returns `400 GENERATED_SQL_NOT_SUPPORTED` for unsupported TABLE_PATTERN / storage combinations before reaching `composeStatic`
- Extended `DataMartReport` with `definitionType` and made `storage` required so action cells and edit forms can evaluate support client-side
- Removed `useBlendedFieldNames` hook, `reportHasBlending`, and `reportHasOutputControls` utilities (no longer needed)
- Extracted `BLENDABLE_SCHEMA_QUERY_KEY` into its own module to keep it importable without pulling in the deleted hook


apps/web/src/features/data-marts/edit/components/ReportColumnPicker/GeneratedSqlViewer.tsx
```diff
@@ -121,89 +108,8 @@ export function GeneratedSqlViewer({
...
-            Preview SQL
-          </Button>
-        </TooltipTrigger>
-        <TooltipContent side='bottom' role='tooltip' className='w-fit max-w-[270px] text-balance'>
-          {infoTooltipContent}
-        </TooltipContent>
-      </Tooltip>
-    );
-  }
+  const ariaLabel = reportTitle ? `Preview SQL: ${reportTitle}` : 'Preview SQL';
```
max-voloshyn, line 111: 🔴 Blocking: This now opens the generated SQL dialog for every report, but `/generated-sql` still delegates to storage query builders that throw for TABLE_PATTERN on Athena, Redshift, Snowflake, and Databricks, while the UI still validates/allows TABLE_PATTERN for those storages. A source-direct table-pattern report that previously showed the Data Setup tooltip will now call this endpoint and show "Failed to load generated SQL". Please either make generated SQL support those combinations or keep a gate/disabled state for unsupported definition/storage pairs.
🤖 Reviewed by Codex (GPT-5.5)

romandubovyi: Fixed: Added gate/disabled state for unsupported definition/storage pairs

apps/web/src/features/data-marts/reports/shared/utils/is-generated-sql-supported.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  definitionType: DataMartDefinitionType | null,
+  storageType: DataStorageType
+): boolean {
+  if (definitionType !== DataMartDefinitionType.TABLE_PATTERN) {
+    return true;
+  }
+
+  return (
+    storageType === DataStorageType.GOOGLE_BIGQUERY ||
+    storageType === DataStorageType.LEGACY_GOOGLE_BIGQUERY
```
max-voloshyn: 🟡 Non-blocking: This marks TABLE_PATTERN as supported for `LEGACY_GOOGLE_BIGQUERY`, and the backend guard mirrors that, but `LegacyBigQueryQueryBuilder.buildQuery()` only accepts SQL definitions and throws for anything else. If legacy table-pattern data is impossible because legacy marts are SQL-only, please remove legacy from the supported list/tests; otherwise add real builder support or reject it before `composeStatic`.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-marts/reports/shared/model/types/data-mart-report.ts
```diff
@@ -5,6 +5,7 @@ import {
 } from '../../enums';
 import { type DataDestination } from '../../../../../data-destination';
 import type { DataMart } from '../../../../edit';
+import type { DataMartDefinitionType } from '../../../../data-marts/shared/enums/data-mart-definition-type.enum.ts';
```
max-voloshyn: 🔴 Blocking: This import path is still one directory too deep, so `DataMartDefinitionType` resolves as an error type and `npm run lint -w @owox/web` fails here, then propagates unsafe-argument errors into the report action cells. Please import the enum from the existing shared enum path for this package, for example `../../../../shared/enums/data-mart-definition-type.enum.ts`, so the report action gates are strongly typed again.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-marts/reports/edit/components/LookerStudioReportEditForm/LookerStudioReportEditForm.tsx
```diff
@@ -245,19 +240,23 @@ export const LookerStudioReportEditForm = forwardRef<
                       form.setValue('sortConfig', config.sortConfig, { shouldDirty: true });
                       form.setValue('limitConfig', config.limitConfig, { shouldDirty: true });
                     }}
-                    onBlendedSelectionChange={setHasBlendedSelection}
                     onCountChange={setColumnsCount}
```
max-voloshyn, line 242: 🔴 Blocking: Removing `onBlendedSelectionChange={setHasBlendedSelection}` leaves the `hasBlendedSelection` state declared above unused, so `npm run lint -w @owox/web` now fails with no-unused-vars. Please remove the unused state tuple or rewire it if it is still needed.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-marts/reports/shared/utils/is-generated-sql-supported.utils.ts
romandubovyi: Fixed: removed `LEGACY_GOOGLE_BIGQUERY` from the supported list in both frontend (`isGeneratedSqlSupported`) and backend (`GetReportGeneratedSqlService`). Legacy data marts are SQL-only, so TABLE_PATTERN is impossible there. Updated the test to assert `false` for legacy.

apps/web/src/features/data-marts/reports/shared/model/types/data-mart-report.ts
romandubovyi: This was based on the original diff — the current code imports from `../../../../shared` (not `../../../../data-marts/shared/...`), which resolves correctly via the barrel export. Lint passes clean. No changes needed.

apps/web/src/features/data-marts/reports/edit/components/LookerStudioReportEditForm/LookerStudioReportEditForm.tsx
romandubovyi: Already removed — `hasBlendedSelection` state and `usesSourceDirectly` computation were both cleaned up in the same commit. Lint passes clean.

apps/web/src/features/data-marts/reports/edit/components/LookerStudioReportEditForm/LookerStudioReportEditForm.tsx
```diff
@@ -245,19 +239,20 @@ export const LookerStudioReportEditForm = forwardRef<
...
-                        dataMartId={dataMart.id}
-                        variant='outline-button'
-                        usesSourceDirectly={usesSourceDirectly}
-                      />
-                    </div>
-                  )}
+                  {mode === ReportFormMode.EDIT &&
+                    initialReport?.id &&
+                    dataMart.id &&
+                    isGeneratedSqlSupported(dataMart.definitionType, dataMart.storage.type) && (
```
max-voloshyn, line 247: 🔴 Blocking: The project-wide reports edit path loses the real `definitionType` before this check. `DataMartReportsPage` wraps the sheet with `buildProjectDataMartContextValue(editingReport.dataMart)`, but that helper still hard-codes `definitionType: null`, so `isGeneratedSqlSupported(...)` returns true for a `TABLE_PATTERN` report on Athena/Snowflake/Redshift/Databricks/legacy BigQuery and shows the Preview SQL button even though `get-report-generated-sql` now rejects it. Please carry `dataMartRef.definitionType` through the project context or base this check on `initialReport.dataMart.definitionType` when present.
🤖 Reviewed by Codex (GPT-5.5)

romandubovyi: Fixed: `buildProjectDataMartContextValue` now carries `definitionType` from the passed `dataMartRef` instead of hard-coding `null`. Since `DataMartReport.dataMart` already includes `definitionType`, the value propagates automatically through both the action cells context and the edit sheet context.

max-voloshyn (approved):

PR #1294, opened 2026-06-08
feat: anonymous opt-out CLI telemetry for owox serve and owox-ctl

zapolsky:
## Summary

Anonymous, opt-out usage telemetry for the OWOX CLIs, so we can understand how many people run
the product and prioritize accordingly.

**`owox serve`** sends a single `cli.serve.started` event on successful startup.
**`owox-ctl`** sends a single `cli.command.invoked` event per command (the oclif command id only,
e.g. `data-marts:list` — never arguments).

Shared design:
- Built on the existing `EventBus` in `@owox/internal-helpers`. A `TelemetryEvent` marker class +
  `PostHogTransport` that is a **no-op for any non-telemetry event**, so PII-bearing backend events
  can never reach PostHog.
- CLI-agnostic helpers (`isTelemetryEnabled`, `getOrCreateAnonymousId`) and a generic
  `emitTelemetry()` orchestrator live in `@owox/internal-helpers`; `trackServeStarted` (owox) and
  `trackCommand` (ctl) are thin wrappers. ctl is instrumented once at `BaseCommand.finally`, so all
  commands are covered.
- Anonymous random UUID (`crypto.randomUUID`, not machine-derived) persisted in the OS app-data dir.
- Opt-out via `OWOX_TELEMETRY_DISABLED` or `DO_NOT_TRACK`; auto-disabled in CI and in the dev script.
- Public PostHog key is empty in git, injected at release time by `scripts/inject-posthog-key.mjs`,
  wired into `publish.yml` before the build step.
- Fire-and-forget: never throws, never blocks startup or command completion.

What is collected (anonymous, no PII): a random id, CLI/Node versions, OS platform/arch, Docker/IDP/web
flags (serve), and the command id (ctl). Never: hostnames, paths, emails, env values, command args, IPs.

## ⚠️ Manual steps required on GitHub before this ships telemetry

Telemetry stays silently disabled until these are done (empty key → safe no-op):

1. **Add the repo Actions secret** `POSTHOG_API_KEY` = the OWOX PostHog project's **public** key
   (`phc_…`): Settings → Secrets and variables → Actions → New repository secret.
2. **If the PostHog region is not EU**, bump the `DEFAULT_POSTHOG_HOST` default in
   `packages/internal-helpers/src/integrations/event-bus/posthog-config.ts`
   (e.g. to `https://us.i.posthog.com`).
3. The publish workflow already runs the inject step before build; no further CI wiring needed.

## Test Plan

- [x] `@owox/internal-helpers`, `owox`, `@owox/ctl`: lint + typecheck + test + build all green
- [x] `posthog-key.ts` empty in git; no real key committed
- [x] telemetry tests are hermetic (no `telemetry.json` written outside a temp dir)
- [x] ctl payload sends only the command id — no argv/flags
- [ ] After merge + secret: confirm `cli.serve.started` and `cli.command.invoked` events arrive in PostHog
- [ ] Confirm `OWOX_TELEMETRY_DISABLED=1` suppresses events for both `owox serve` and `owox-ctl`


apps/ctl/src/base-command.ts
```diff
@@ -87,4 +87,18 @@ export abstract class BaseCommand extends Command {
...
     this.exit(1);
   }
+
+  protected async finally(error: Error | undefined): Promise<void> {
+    try {
+      const { trackCommand } = await import('./telemetry/track-command.js');
+      trackCommand({
+        cliVersion: this.config.version,
+        command: this.id ?? 'unknown',
+        log: (message: string) => this.log(message),
```
max-voloshyn: 🔴 Blocking: This routes the first-run telemetry notice through `this.log`, so released `owox-ctl` will append a human-readable line to stdout on the first telemetry-enabled invocation. Several commands emit JSON/NDJSON on stdout, so this can corrupt automation output even when the command succeeds. Please send the notice to stderr or suppress it for machine-readable commands.
🤖 Reviewed by Codex (GPT-5.5)

packages/internal-helpers/src/integrations/event-bus/transports/posthog-transport.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (!isTelemetryEvent(event)) return;
+    if (!this.config.apiKey) return;
+
+    const payload = { ...(event.payload as Record<string, unknown>) };
+    const anonymousId = payload.anonymousId;
+    if (typeof anonymousId !== 'string' || anonymousId.length === 0) return;
+    delete payload.anonymousId;
+
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), this.config.timeoutMs);
```
max-voloshyn: 🔴 Blocking: This timeout is a ref-counted timer inside an un-awaited telemetry send, so short-lived `owox-ctl` commands can stay alive until the PostHog request settles or this 3s abort fires. I verified this path with a pending fetch and the process exited after about 3006ms, which violates the “never blocks command completion” behavior. Please avoid leaving ref-counted timers/network handles on the CLI path, or explicitly bound/await the delay if it is intentional.
🤖 Reviewed by Codex (GPT-5.5)

apps/ctl/src/base-command.ts
zapolsky: Fixed in 33b11e3 — the first-run notice now goes to `process.stderr.write` instead of `this.log`, so stdout stays clean for owox-ctl's JSON/NDJSON output.

packages/internal-helpers/src/integrations/event-bus/transports/posthog-transport.ts
zapolsky: Fixed in dee7db6 — replaced `fetch` with `node:http`/`node:https` using `socket.unref()` and an unref'd bounded timeout that `req.destroy()`s. Root cause was `fetch`: its socket can't be unref'd. Verified an unreachable host now releases the process in ~680ms (bounded by timeoutMs) instead of 8s+, and a responsive host releases as soon as the socket is established.

max-voloshyn (approved):

PR #1297, opened 2026-06-08
fix(idp): invalidate other sessions on password reset and change

laskevych:
## Why

A security assessment found that some password reset/change paths left a user's
*other* active sessions valid. After a successful password reset or change, the
user's other sessions should be invalidated (the session performing the change
may remain valid). Scope: `idp-better-auth` and `idp-owox-better-auth`.
First-time password setup is out of scope.

### `idp-better-auth`
A "session" here is a Better Auth session (the `refreshToken` cookie is the
Better Auth session token; access tokens are short-lived), so deleting session
rows cuts off other devices' refresh ability.

- **Forced revoke on the exposed `/change-password`** — a `hooks.before`
  middleware sets `revokeOtherSessions: true`, so a password change always
  revokes the user's other sessions while a fresh session is issued for the
  current one (clients can no longer skip it).
- **Admin self-reset revokes other sessions, preserving the current one** —
  `resetUserPassword(userId, adminUserId, currentSessionToken?)`: resetting
  another user revokes all of that user's sessions; self-reset revokes every
  other session and keeps the acting one; missing token → fail-safe revoke-all.
- New store method `revokeOtherUserSessions(userId, exceptSessionToken)` on the
  SQLite and MySQL stores + interface; a failed revoke is logged rather than
  silently swallowed.
- Hardening: `deleteUserCascade` logs failed session/account/member deletes
  instead of swallowing them, and `resetUserPassword` issues the magic link
  before clearing the password so a failure can't leave the user password-less.

### `idp-owox-better-auth`
The Better Auth session layer is brought in line with `idp-better-auth`:

- Forced `revokeOtherSessions: true` on the proxied `/change-password` (same
  `hooks.before` middleware).
- `revokeSessionsOnPasswordReset: true`, so a password reset revokes the user's
  other Better Auth sessions.
- On a successful reset, the platform refresh token carried by the request is
  also revoked (best-effort, non-fatal) and its cookie cleared.

Broader multi-device revocation at the platform-session level is owned by the
OWOX Identity service and tracked separately.

## Tests

- `idp-better-auth`: unit tests cover the `resetUserPassword` branches and the
  forced `/change-password` revoke hook (a DB-free unit test that runs in CI).
- `idp-owox-better-auth`: controller tests assert the platform token is revoked
  on a successful reset and not on a failed one; the change-password revoke hook
  is unit-tested too.
- Real-SQLite suites (the `token != ?` revoke SQL and an end-to-end
  change-password session-invalidation check) run against an in-memory database,
  gated on a `better-sqlite3` binding probe so they run locally / in any
  native-enabled job and skip (not fail) in the `--ignore-scripts` `test-owox`
  job.
- Full suites green: idp-better-auth 132, idp-owox-better-auth 245.

## Not in scope

- First-time password setup.
- Platform-session multi-device revocation in `idp-owox` (handled in a separate
  OWOX Identity task).


packages/idp-owox-better-auth/src/controllers/password-flow-controller.ts
```diff
@@ -161,6 +173,10 @@ export class PasswordFlowController {
           body: { newPassword: password, token: resetToken },
           headers: convertExpressHeaders(req),
         });
+        // Revoke the OWOX platform refresh token carried by this request (if
+        // any) so the acting device's platform session is invalidated at the
+        // OWOX Identity level, not only the Better Auth UI session.
+        await this.revokePlatformSession?.(req, res);
```
max-voloshyn, line 179: 🔴 Blocking: This reset flow still leaves the user’s other Better Auth sessions valid. better-auth@1.6.13 only deletes sessions during resetPassword when emailAndPassword.revokeSessionsOnPasswordReset is enabled; the OWOX config does not set it, and the new callback only revokes or clears the current request platform cookie. A reset from a logged-out browser or while another Better Auth session exists will keep that other session usable. Please enable Better Auth reset-session revocation or explicitly delete the user’s Better Auth sessions before redirecting.
🤖 Reviewed by Codex (GPT-5.5)

packages/idp-better-auth/src/auth/auth-config.ts
```diff
@@ -110,6 +111,23 @@ export async function createBetterAuthConfig(
       enabled: true,
       requireEmailVerification: false,
     },
+    hooks: {
```
max-voloshyn, line 134: 🔴 Blocking: This hook only covers @owox/idp-better-auth. @owox/idp-owox-better-auth has a separate createBetterAuthConfig, and BetterAuthProxyHandler forwards every /auth/better-auth/* route there, so /auth/better-auth/change-password can still omit revokeOtherSessions and keep other sessions alive. Please add the same hook to the OWOX Better Auth config or centralize this config behavior.
🤖 Reviewed by Codex (GPT-5.5)

packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
```diff
@@ -621,6 +622,38 @@ export class OwoxBetterAuthIdp implements IdpProvider {
...
+  /**
+   * Best-effort revocation of the OWOX platform refresh token carried by the
+   * current request (the `refreshToken` cookie), then clears that cookie.
+   *
+   * Used after a password reset so the platform session tied to THIS request
+   * is invalidated at the OWOX Identity level — not only the Better Auth UI
+   * session. Revocation failures are non-fatal (logged, not thrown) so they
+   * never block completing the password flow.
+   *
+   * Scope note: this only revokes the token present in the request (the acting
```
max-voloshyn: 🔴 Blocking: This scope note means password reset still does not invalidate other active OWOX platform sessions. Only the refresh token on the current request is revoked; another device refreshToken can still refresh through accessTokenMiddleware. Since this PR is meant to invalidate other sessions on reset, it needs a revoke-by-user or session-family path or another mechanism that prevents old platform refresh tokens from continuing after reset.
🤖 Reviewed by Codex (GPT-5.5)

zapolsky (approved):
✅ Approved by automated OWOX review — no blocking issues. See top-level comment for non-blocking notes.

zapolsky:

Correct, well-scoped security fix. Both store dialects use identical, fully parameterized SQL (`DELETE FROM session WHERE userId = ? AND token != ?`), scoping is correctly by `userId`, the current session is preserved on self-reset (and the missing-token branch fail-safes to revoke-all), and both the `/change-password` hook and admin self-reset paths are wired. No blocking issues. Minor notes:

- `packages/idp-better-auth/src/store/SqliteDatabaseStore.ts:181` — `revokeOtherUserSessions` swallows all errors in an empty `catch {}`. A zero-row DELETE doesn't throw, so the catch only masks genuine DB failures (locked DB, connection loss) while the password-reset flow still reports success — a fail-open. Mirrors the pre-existing `revokeUserSessions` pattern, but for a security-correctness method consider `logger.error(...)` so a failed revoke is observable.
- `packages/idp-better-auth/src/store/MysqlDatabaseStore.ts:214` — Same swallowed-error concern in the MySQL `revokeOtherUserSessions` empty `catch {}` — log the error instead of discarding it.
- `packages/idp-better-auth/src/store/MysqlDatabaseStore.ts:207` — No SQL-level test for the MySQL revoke; only SQLite gets a real-DB test. The SQL string + params are byte-identical to SQLite and both dialects evaluate `!=` the same with a NOT-NULL `token`, so behavior should match — just noting the MySQL path is only structurally (not empirically) verified, consistent with the existing repo posture.
- `packages/idp-better-auth/src/auth/auth-config.ts:114` — No changeset included; both idp packages are in the `fixed` changeset group and this is a user-facing security fix — add a `patch`/`minor` changeset.
- `packages/idp-better-auth/src/auth/auth-config.session-revocation.test.ts:67` — The change-password test asserts `remaining.length === 1` + token rotated, which proves the other session was revoked and the current survived; consider also asserting `remaining[0]` maps to the acting user's `userId` for explicitness. Minor.

packages/idp-better-auth/src/store/SqliteDatabaseStore.session-revocation.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  function tokensFor(userId: string): string[] {
+    return rows()
+      .filter(r => r.userId === userId)
+      .map(r => r.token);
+  }
+
+  it('revokeOtherUserSessions deletes all of the user’s sessions except the current one', async () => {
+    await store.revokeOtherUserSessions('user-1', 'tok-current');
+
+    expect(tokensFor('user-1')).toEqual(['tok-current']);
```
romandubovyi, line 64: 🟡 **Important:** Reaching into the private `db` field via `as unknown` makes this test brittle. Use `store.getAdapter()` or expose a test-only helper instead, so the test survives internal refactors.

packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
```diff
@@ -621,6 +622,38 @@ export class OwoxBetterAuthIdp implements IdpProvider {
     await this.tokenFacade.revokeToken(token);
   }
 
+  /**
+   * Best-effort revocation of the OWOX platform refresh token carried by the
+   * current request (the `refreshToken` cookie), then clears that cookie.
+   *
+   * Used after a password reset so the platform session tied to THIS request
+   * is invalidated at the OWOX Identity level — not only the Better Auth UI
+   * session. Revocation failures are non-fatal (logged, not thrown) so they
```
romandubovyi, line 738: 💡 **Suggestion:** Consider adding structured context to the warning log — e.g. user id, request path, or a flag whether a token was present. Passing `undefined` as the context makes production debugging harder.

packages/idp-owox-better-auth/src/controllers/password-flow-controller.ts
laskevych: ✅ Fixed in 589119d01. The OWOX Better Auth config now sets
`emailAndPassword.revokeSessionsOnPasswordReset: true`, so `resetPassword`
deletes the user's other Better Auth sessions; the acting reset flow clears its
own cookies immediately after. A reset from a logged-out browser, or with
another Better Auth session present, no longer leaves that session usable.

packages/idp-better-auth/src/auth/auth-config.ts
laskevych: ✅ Fixed in 589119d01. The same `hooks.before` middleware is now in the OWOX
`createBetterAuthConfig` (packages/idp-owox-better-auth/src/config/idp-better-auth-config.ts),
forcing `revokeOtherSessions: true` on `/change-password`. Since
BetterAuthProxyHandler forwards every `/better-auth/*` route to this config, the
proxied change-password can no longer omit the flag. Kept as a duplicated hook
for now rather than centralizing the two configs — happy to extract a shared
helper if you'd prefer.

packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
laskevych: Agreed this is the key concern; we've split it deliberately (scope agreed with
the team).

- Better Auth layer (this PR): now closed for OWOX too —
  `revokeSessionsOnPasswordReset` plus the forced `/change-password` hook (see
  the two comments above). This prevents a stale Better Auth session from being
  used to obtain fresh platform tokens after a reset.
- Platform-session layer: broader multi-device revocation is owned by the OWOX
  Identity service and tracked in a separate task there. I'll share the
  technical details in the internal tracker rather than in this public thread.

I've also reworded the scope-note comment in this file to match.

packages/idp-better-auth/src/store/SqliteDatabaseStore.session-revocation.test.ts
laskevych: ✅ Applied your suggestion (switched the helper to `store.getAdapter()`), but the
suite has since been removed entirely: the `test-owox` CI job installs with
`--ignore-scripts`, so better-sqlite3's native binding isn't built and the
real-SQLite suites fail with "Could not locate the bindings file". So the
brittleness is moot now.

Trade-off: the raw revoke SQL no longer has a dedicated test. If we want it back,
the options are (a) build the binding in `test-owox` (`npm rebuild better-sqlite3`,
or drop `--ignore-scripts` as #1291 did for the integration job), or (b) leave it
removed. Open to your preference.

packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
laskevych: ✅ Done — the warning now logs `{ path: req.path }` instead of `undefined`.

packages/idp-better-auth/src/store/SqliteDatabaseStore.ts
```diff
@@ -167,8 +167,22 @@ export class SqliteDatabaseStore implements DatabaseStore {
...
+  }
+
+  async revokeOtherUserSessions(userId: string, exceptSessionToken: string): Promise<void> {
+    await this.connect();
+    try {
+      this.getDb()
+        .prepare('DELETE FROM session WHERE userId = ? AND token != ?')
+        .run(userId, exceptSessionToken);
+    } catch (error) {
+      logger.error('Failed to revoke other user sessions', { userId }, error as Error);
```
romandubovyi, line 185: 🔵 **Question:** This logs the revoke failure but doesn't propagate it, so `resetUserPassword` still clears the password, issues a magic link, and reports success even when the other sessions were *not* actually revoked. Is that the intended trade-off for now — i.e. we knowingly accept that a failed revocation won't fail the operation? If so, no change needed; just want to confirm it's deliberate rather than an oversight, since it weakens the guarantee this PR is meant to provide.

packages/idp-better-auth/src/store/SqliteDatabaseStore.ts
```diff
@@ -167,8 +167,22 @@ export class SqliteDatabaseStore implements DatabaseStore {
...
+      // so log it rather than swallowing.
+      logger.error('Failed to revoke user sessions', { userId }, error as Error);
+    }
+  }
+
+  async revokeOtherUserSessions(userId: string, exceptSessionToken: string): Promise<void> {
+    await this.connect();
+    try {
+      this.getDb()
+        .prepare('DELETE FROM session WHERE userId = ? AND token != ?')
```
romandubovyi, line 182: 🟡 **Important:** The new `token != ?` query (and the existing `token` column assumption) isn't exercised by any test — the unit tests mock the store, and the PR drops the real-SQLite revocation suites because of the `--ignore-scripts` CI constraint. That leaves the most fragile part (correct column name / filter semantics against the actual Better Auth schema) uncovered. Could we add at least one test that runs this query against a real (or in-memory) DB, even if gated to run outside the `--ignore-scripts` job, so a schema/column drift would be caught?

packages/idp-better-auth/src/auth/auth-config.ts
```diff
@@ -110,6 +111,23 @@ export async function createBetterAuthConfig(
...
+      // Security: the Better Auth `/change-password` endpoint is exposed through
+      // the request handler. By default it only revokes other sessions when the
+      // caller opts in via `revokeOtherSessions`. Force it on so a password
+      // change always invalidates the user's other sessions while Better Auth
+      // issues a fresh session for the current one.
+      before: createAuthMiddleware(async ctx => {
+        if (ctx.path === '/change-password') {
+          return {
+            context: {
+              body: { ...ctx.body, revokeOtherSessions: true },
```
romandubovyi: 🔵 **Question:** Have we verified end-to-end that returning `{ context: { body } }` from the `before` middleware actually reaches the `/change-password` handler (Better Auth's context-merge semantics), and that with `revokeOtherSessions: true` the *current* session stays valid (fresh token issued) rather than the password-changing user being logged out too? This path isn't covered by the unit tests, so a quick manual/integration confirmation would be reassuring.

packages/idp-better-auth/src/store/MysqlDatabaseStore.ts
```diff
@@ -199,8 +199,23 @@ export class MysqlDatabaseStore implements DatabaseStore {
     const pool = await this.getPool();
     try {
       await pool.execute('DELETE FROM session WHERE userId = ?', [userId]);
-    } catch {
-      // Non-fatal: sessions might not exist
+    } catch (error) {
+      // A zero-row delete does not throw, so this catch only fires on a real DB
+      // failure. A silently failed session revoke is a security-relevant event,
+      // so log it rather than swallowing.
+      this.logger.error('Failed to revoke user sessions', { userId, error });
```
romandubovyi, line 206: 🟡 **Important:** `deleteUserCascade` still swallows session deletion errors with `// Non-fatal cleanup`, while the new revocation methods log them as security-relevant. Consider applying the same logging standard to the cleanup deletes so a failed cascade session removal is not silently ignored.

packages/idp-better-auth/src/services/user-management-service.ts
```diff
@@ -270,6 +280,10 @@ export class UserManagementService {
 
       if (userId !== adminUserId) {
         await this.store.revokeUserSessions(userId);
+      } else if (currentSessionToken) {
+        await this.store.revokeOtherUserSessions(userId, currentSessionToken);
+      } else {
+        await this.store.revokeUserSessions(userId);
       }
 
       await this.store.clearUserPassword(userId);
```
romandubovyi: 🟡 **Important:** `resetUserPassword` performs session revocation, `clearUserPassword`, and `generateMagicLink` without a transaction. If `clearUserPassword` succeeds but `generateMagicLink` fails, the user is left password-less and without a magic link. Consider wrapping the destructive operations in a transaction or recovering on failure.

packages/idp-better-auth/src/auth/auth-config.ts
```diff
@@ -110,6 +111,23 @@ export async function createBetterAuthConfig(
...
       requireEmailVerification: false,
     },
+    hooks: {
+      // Security: the Better Auth `/change-password` endpoint is exposed through
+      // the request handler. By default it only revokes other sessions when the
+      // caller opts in via `revokeOtherSessions`. Force it on so a password
+      // change always invalidates the user's other sessions while Better Auth
+      // issues a fresh session for the current one.
+      before: createAuthMiddleware(async ctx => {
+        if (ctx.path === '/change-password') {
```
romandubovyi: 💡 **Suggestion:** Consider adding a unit test that asserts the `before` hook mutates the body to `revokeOtherSessions: true` for `/change-password` and leaves other paths untouched. The current tests cover the admin reset path but not this forced hook.

packages/idp-better-auth/src/store/SqliteDatabaseStore.ts
laskevych: Yes, deliberate. The revoke is best-effort but now observable: a zero-row delete
doesn't throw, so the catch only fires on a real DB failure, which we log as a
security-relevant event (previously swallowed). We don't fail the reset on a
session-revoke DB error so a transient DB issue can't block a user from resetting.
If you'd prefer the stronger guarantee, we can make the revoke in
resetUserPassword propagate and abort the reset on failure — happy to switch.

packages/idp-better-auth/src/store/SqliteDatabaseStore.ts
laskevych: Added. Re-added a real in-memory SQLite test for revokeOtherUserSessions /
revokeUserSessions exercising the `token != ?` filter and the token/userId column
names against an actual schema. It's gated on a one-time better-sqlite3 binding
probe: runs locally and in any native-enabled job, and skips (doesn't fail) in the
--ignore-scripts test-owox job. Verified it skips with the binding hidden and
passes with it present.

packages/idp-better-auth/src/auth/auth-config.ts
laskevych: Covered two ways: (1) a pure unit test on the extracted hook asserts it sets
revokeOtherSessions: true for /change-password and leaves other paths untouched
(runs in CI, no DB); (2) a gated in-memory integration test runs change-password
end-to-end and asserts exactly one session remains afterward with a rotated token
belonging to the acting user — i.e. the changer keeps a fresh session, not logged
out (skips when the native binding is unavailable).

packages/idp-better-auth/src/store/MysqlDatabaseStore.ts
laskevych: ✅ Done — deleteUserCascade now logs failed session/account/membership deletes
(logger.error with { userId }) instead of swallowing them, matching the revocation
methods; the session delete is flagged as security-relevant (orphaned sessions
outliving the user).

packages/idp-better-auth/src/services/user-management-service.ts
laskevych: Addressed by ordering rather than a DB transaction (the magic-link step spans the
store and the mail service, so one store transaction can't wrap it cleanly):
resetUserPassword now generates the magic link before clearing the password, so a
magic-link failure can't leave the user password-less and without a link —
clearing the credential is the last destructive step. Happy to wrap the remaining
DB writes in a transaction too if you'd like stronger atomicity.

packages/idp-better-auth/src/auth/auth-config.ts
laskevych: Added — extracted the change-password hook into
forceRevokeOtherSessionsOnChangePassword and unit-tested it in both packages:
asserts the body gets revokeOtherSessions: true for /change-password and is left
untouched for other paths. No DB, so it runs in the standard test-owox job.

romandubovyi (approved):

PR #1300, opened 2026-06-09
fix: follow up project data mart pages

kalnik-a-a:
## Summary
- Load Reports, Insights, and Triggers lists page-by-page from the API while keeping table pagination at 15 rows.
- Refresh Reports without refetching on side sheet close, and poll only running reports from the visible page.
- Set Run History fetch batches to 50.
- Fix the broken Reports link in the "Cannot delete destination" popup so it opens the new Reports page filtered by destination.

## Verification
- npm run lint -w @owox/web
- npm run type-check -w @owox/web
- npm run format:check -w @owox/web
- npm run test -w @owox/web -- src/features/data-destination/list/components/DataDestinationList/DataDestinationList.test.tsx src/features/data-marts/insights/shared/services/insight-templates.service.test.ts src/features/data-marts/reports/shared/services/report.service.test.ts src/features/data-marts/schedules/shared/services/scheduled-trigger.service.test.ts src/pages/data-marts/insights/DataMartInsightsPage.test.tsx src/pages/data-marts/reports/DataMartReportsPage.test.tsx src/pages/data-marts/runs/DataMartRunsPage.test.tsx src/pages/data-marts/schedules/DataMartSchedulesPage.test.tsx

### DIFF
--- a/apps/web/src/features/data-destination/list/components/DataDestinationList/DataDestinationList.tsx
+++ b/apps/web/src/features/data-destination/list/components/DataDestinationList/DataDestinationList.tsx
@@ -22,6 +22,14 @@ interface DataDestinationListProps {
   onCreateSheetClose?: () => void;
 }
 
+function buildReportsDestinationFilterPath(destinationTitle: string) {
+  const params = new URLSearchParams({
+    filters: JSON.stringify([{ f: 'destination', o: 'eq', v: [destinationTitle] }]),
+  });
+
+  return `/data-marts/reports?${params.toString()}`;
+}
+
 export const DataDestinationList = ({
   isCreateSheetInitiallyOpen = false,
   onCreateSheetClose,
@@ -219,7 +227,7 @@ export const DataDestinationList = ({
               <span className='block'>
                 <strong>&ldquo;{blocked.impact.destinationTitle}&rdquo;</strong> is referenced by{' '}
                 <Link
-                  to={scope('/data-marts')}
+                  to={scope(buildReportsDestinationFilterPath(blocked.impact.destinationTitle))}
                   className='text-primary hover:underline'
                   onClick={() => {
                     setBlocked(null);
--- a/apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
+++ b/apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
@@ -2,8 +2,6 @@ import { useCallback, useEffect, useMemo, useState } from 'react';
 import type { ReactNode } from 'react';
 import { useParams } from 'react-router-dom';
 import type { ColumnDef } from '@tanstack/react-table';
-import { RefreshCw } from 'lucide-react';
-import { Button } from '@owox/ui/components/button';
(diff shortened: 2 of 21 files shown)


apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
```diff
@@ -444,10 +388,26 @@ export default function DataMartReportsPage() {
...
 
+  const hasRunningVisibleReports = table
+    .getRowModel()
+    .rows.some(row => row.original.lastRunStatus === ReportStatusEnum.RUNNING);
+
+  useEffect(() => {
+    if (!hasRunningVisibleReports) return;
+
+    const intervalId = window.setInterval(() => {
+      void loadReports({ silent: true, firstPageOnly: true });
```
max-voloshyn, line 403: 🔴 Blocking: This polling path always refreshes API offset 0, but hasRunningVisibleReports is based on the client-side paginated/sorted table. A visible running report can be outside the first 100 reports returned by the backend, so the interval keeps merging the same stale row and the RUNNING status never clears. Poll the visible running report IDs or otherwise refresh the slice that actually contains the visible rows.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/features/data-destination/list/components/DataDestinationList/DataDestinationList.tsx
```diff
@@ -22,6 +22,14 @@ interface DataDestinationListProps {
   onCreateSheetClose?: () => void;
 }
 
+function buildReportsDestinationFilterPath(destinationTitle: string) {
+  const params = new URLSearchParams({
+    filters: JSON.stringify([{ f: 'destination', o: 'eq', v: [destinationTitle] }]),
```
max-voloshyn, line 27: 🔴 Blocking: This uses destinationTitle as the filter identity, but duplicate Data Destination titles are allowed, so the blocked-delete link can show reports for unrelated destinations with the same title. Please filter by destination id, or add an id-backed Reports filter, so the link opens exactly the affected reports for blocked.impact.destinationId.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
kalnik-a-a: Addressed in follow-up PR #1301: https://github.com/OWOX/owox-data-marts/pull/1301. Reports polling now refreshes only the currently visible RUNNING report rows by id instead of refetching API offset 0.

apps/web/src/features/data-destination/list/components/DataDestinationList/DataDestinationList.tsx
kalnik-a-a: Keeping this unchanged for this task. The Reports filters are title-based today, so this link follows the current filter contract. It is intentionally aligned with the existing "Cannot delete storage" popup, which also builds its Data Marts link by storage title.

PR #1301, opened 2026-06-09
fix: poll visible running reports by id

kalnik-a-a:
## Summary
- Replace Reports silent first-page polling with per-id refreshes for currently visible running report rows.
- Preserve loaded report order and unchanged row references during partial refreshes.
- Add regression coverage for hidden running rows, tail visible running rows, and polling stop after completion.

## Verification
- npm run test -w @owox/web -- src/pages/data-marts/reports/DataMartReportsPage.test.tsx
- npm run format:check -w @owox/web
- npm run type-check -w @owox/web
- npm run lint -w @owox/web

### DIFF
--- a/apps/web/src/pages/data-marts/reports/DataMartReportsPage.test.tsx
+++ b/apps/web/src/pages/data-marts/reports/DataMartReportsPage.test.tsx
@@ -487,9 +487,9 @@ describe('DataMartReportsPage', () => {
     expect(reportService.runReport).not.toHaveBeenCalled();
   });
 
-  it('polls project reports while any visible report is still running', async () => {
+  it('polls visible running reports by id', async () => {
     vi.useFakeTimers({ shouldAdvanceTime: true });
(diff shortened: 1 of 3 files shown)


apps/web/src/pages/data-marts/reports/DataMartReportsPage.test.tsx
```diff
@@ -502,8 +502,9 @@ describe('DataMartReportsPage', () => {
     await vi.advanceTimersByTimeAsync(5000);
```
max-voloshyn, line 502: 🟡 Non-blocking: This timer advancement pattern, also repeated in the new polling tests below, emits React `act(...)` warnings when running the targeted suite. The tests pass, but wrapping the fake-timer advancement/state flush in `act` would keep the output clean and make real warnings easier to spot.
🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (approved):

PR #1303, opened 2026-06-10
feat(data-marts): output controls for Snowflake data marts (filters, …

zapolsky:
…slices, sort, limit)

Adds filters / column slices / sort / limit to Snowflake reports, matching the Athena/BigQuery/Redshift feature already shipped. Uses "option B": the clause renderer inlines every value as a SQL literal (params always empty), so there is no bound-param plumbing and composeStatic's empty-params early-return covers it.

Snowflake-specific choices, all live-verified against a real account (28/28 integration):
- CONTAINS/STARTSWITH/ENDSWITH (never LIKE) so user %/_ stay literal
- REGEXP_INSTR(col, pat) > 0 for regex (RLIKE/REGEXP_LIKE full-anchor in Snowflake)
- backslash-first literal escaping (Snowflake honors backslash escape sequences)
- defensive CAST(<lit> AS DATE|TIME|TIMESTAMP) for date-typed comparisons
- LISTAGG(...) WITHIN GROUP (ORDER BY) for deterministic blended STRING_AGG
- two identifier escapers: the robust shared escaper for user-controlled column refs (injection-safe), the FQN-aware escaper for table names (hardened to throw on 4+-part input instead of returning raw)

Capability + web/extension mirrors updated; renderer/builder/blended unit specs, SQL-emission e2e, and a credential-gated live integration suite added. The live suite (snowflake.integration + the Snowflake block in http-data-real) is wired into the scheduled CI matrix in test-integration.yml; it skips green until the SNOWFLAKE_* secrets are configured. Changeset included.

### DIFF
--- a/.changeset/snowflake-output-controls.md
+++ b/.changeset/snowflake-output-controls.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Add output controls for Snowflake data marts
+
+Add report-level output controls (filters, slices, sort, limit) for Snowflake data marts. Filter values are inlined as escaped SQL literals (Snowflake-correct backslash + quote escaping); substring matchers use CONTAINS/STARTSWITH/ENDSWITH (not LIKE) so user `%`/`_` stay literal, and regex uses `REGEXP_INSTR(...) > 0` for partial matching; date/time comparisons are CAST to the column type. Works for the web app and the Google Sheets extension.
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -28,9 +28,11 @@ jobs:
             pattern: athena.integration
           - name: Redshift
             pattern: redshift.integration
+          - name: Snowflake
+            pattern: snowflake.integration
           - name: Databricks
             pattern: databricks.integration
-          - name: HTTP Data (Athena + BigQuery + Redshift + Databricks)
+          - name: HTTP Data (Athena + BigQuery + Redshift + Databricks + Snowflake)
             pattern: http-data-real
           - name: Report read (Athena)
             pattern: report-run-real
@@ -75,6 +77,14 @@ jobs:
           REDSHIFT_REGION: ${{ secrets.REDSHIFT_REGION }}
           REDSHIFT_WORKGROUP_NAME: ${{ secrets.REDSHIFT_WORKGROUP_NAME }}
           REDSHIFT_DATABASE: ${{ secrets.REDSHIFT_DATABASE }}
+          # Snowflake (PASSWORD auth = PAT in the password field; the account needs a
+          # network policy that permits GitHub Actions runners, or the auth will fail)
+          SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
+          SNOWFLAKE_WAREHOUSE: ${{ secrets.SNOWFLAKE_WAREHOUSE }}
+          SNOWFLAKE_USERNAME: ${{ secrets.SNOWFLAKE_USERNAME }}
+          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
+          SNOWFLAKE_DATABASE: ${{ secrets.SNOWFLAKE_DATABASE }}
+          SNOWFLAKE_SCHEMA: ${{ secrets.SNOWFLAKE_SCHEMA }}
           # Databricks (PERSONAL_ACCESS_TOKEN auth; catalog + schema host the seed table)
           DATABRICKS_HOST: ${{ secrets.DATABRICKS_HOST }}
           DATABRICKS_HTTP_PATH: ${{ secrets.DATABRICKS_HTTP_PATH }}
--- a/apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-blended-query-builder.ts
+++ b/apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-blended-query-builder.ts
@@ -2,17 +2,32 @@ import { Injectable } from '@nestjs/common';
 import { DataStorageType } from '../../enums/data-storage-type.enum';
 import { AbstractBlendedQueryBuilder } from '../../interfaces/abstract-blended-query-builder';
 import { SqlClauseRenderer } from '../../utils/sql-clause-renderer';
+import { SnowflakeClauseRenderer } from './snowflake-clause-renderer';
 
 @Injectable()
 export class SnowflakeBlendedQueryBuilder extends AbstractBlendedQueryBuilder {
   readonly type = DataStorageType.SNOWFLAKE;
   protected readonly identifierQuoteChar = '"';
 
-  protected get clauseRenderer(): SqlClauseRenderer | null {
-    return null;
+  constructor(private readonly _renderer: SnowflakeClauseRenderer) {
+    super();
+  }
+
+  protected get clauseRenderer(): SqlClauseRenderer {
+    return this._renderer;
+  }
+
+  // Snowflake folds UNQUOTED identifiers to UPPERCASE, so the base class's bare return
+  // for simple names would miss the lowercase columns the OWOX connector creates (as
+  // quoted). Always quote — case-stable, and consistent with the non-blended builder.
+  protected quoteIdentifier(name: string): string {
+    const q = this.identifierQuoteChar;
+    return `${q}${name.split(q).join(q + q)}${q}`;
   }
 
   protected buildStringAgg(fieldName: string): string {
-    return `LISTAGG(CAST(${fieldName} AS VARCHAR), ', ')`;
+    // WITHIN GROUP (ORDER BY) makes the concatenation deterministic — Snowflake allows
+    // omitting it (Redshift requires it), but a stable order matches the Redshift sibling.
+    return `LISTAGG(CAST(${fieldName} AS VARCHAR), ', ') WITHIN GROUP (ORDER BY ${fieldName})`;
   }
 }
--- a/apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-query.builder.ts
+++ b/apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-query.builder.ts
@@ -1,4 +1,4 @@
-import { Injectable, NotImplementedException } from '@nestjs/common';
+import { Injectable } from '@nestjs/common';
 import { DataMartDefinition } from '../../../dto/schemas/data-mart-table-definitions/data-mart-definition';
 import {
   isConnectorDefinition,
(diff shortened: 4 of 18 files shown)


zapolsky:

> ℹ️ This PR is authored by the review-bot identity, so no formal approval review is submitted (GitHub blocks self-approval). The notes below are informational.

Clean, well-tested, security-conscious PR that mirrors the established Redshift output-controls pattern. **SQL / identifier injection was verified safe** — filter values use Snowflake-correct backslash-then-quote escaping (proven inert against breakout payloads), user-controlled column identifiers go through the shared escaper that quotes every dotted part, substring/regex operators avoid `LIKE`, date comparisons CAST to the normalized column type, and the `params: []` invariant is triple-guarded. The `escapeSnowflakeIdentifier` 4+-part raw passthrough is also hardened to fail closed. No controllers/facades touched. Only minor notes:

- `apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-clause-renderer.ts:5` — The comment justifying the shared escaper says `escapeSnowflakeIdentifier` "returns 4+-part identifiers unescaped (a WHERE-injection)" — now stale, since this same PR changes that function to throw on 4+ parts (`snowflake-identifier.utils.ts:51`). The decision to use the shared escaper is still correct; just update the rationale so it doesn't imply live-injection behavior that no longer exists.
- `apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-clause-renderer.ts:617` — The local `relative_date` n guard permits `n === 0` while the zod schema restricts n to `.positive()` (n ≥ 1). Intentionally looser defense-in-depth and matches the Redshift sibling exactly — consistent, no action required.
- `.changeset/snowflake-output-controls.md:8` — Uses single-quoted `'owox'` whereas sibling changesets (e.g. `add-athena-output-controls.md`) use double-quoted `"owox"`. Both valid YAML; purely cosmetic.

apps/backend/src/data-marts/data-storage-types/snowflake/utils/snowflake-identifier.utils.ts
```diff
@@ -48,8 +48,9 @@ export function escapeSnowflakeIdentifier(identifier: string): string {
     return `${database}.${quoteIdentifierPart(parts[1])}.${quoteIdentifierPart(parts[2])}`;
   }
 
-  // Fallback for unexpected format
-  return identifier;
+  // 4+ parts is never a valid Snowflake identifier (max is database.schema.table).
+  // Returning it raw would emit unescaped user input into SQL, so fail closed.
+  throw new Error(`Invalid Snowflake identifier (too many parts): "${identifier}"`);
```
romandubovyi: 🔵 **Question:** Failing closed here is right, but `fullyQualifiedName` is user-typed and flows into `buildQuery` → this function, so a 4-part input would now surface as a plain `Error` (likely an HTTP 500) instead of a 400. Does the definition dry-run validation reliably catch this earlier, or should this throw `BusinessViolationException` so the user gets a proper validation error?

apps/backend/test/integration/snowflake.integration.ts
```diff
@@ -0,0 +1,484 @@
...
+    }, 30000);
+
+    // -------------------------------------------------------------------------
+    // §0 PROBE 2 — CAST necessity
+    // -------------------------------------------------------------------------
+    // Does Snowflake accept a bare string literal in a TIMESTAMP_NTZ comparison
+    // without a CAST? The renderer currently emits CAST(...), but if bare literals
+    // work the defensive CAST could be dropped.
+
+    it('§0 PROBE 2 — CAST necessity: bare-literal date comparison on TIMESTAMP_NTZ', async () => {
```
romandubovyi: 💡 **Suggestion:** PROBE 2 only logs the bare-literal outcome without asserting it, and the renderer comment still says "(the live integration may relax this)". Since the live run has already happened, consider finalizing the decision now: either drop the defensive CAST, or declare it permanent and turn this probe into a regular assertion (removing the "may relax" note). Otherwise every scheduled CI run keeps executing an open question that nobody reads the answer to.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-query.builder.ts
```diff
@@ -12,55 +12,113 @@ import {
...
+      'p',
+      resolveColumnType
+    );
+    const orderBy = this.clauseRenderer.renderOrderBy(queryOptions?.sort ?? []);
+    const limit = this.clauseRenderer.renderLimit(queryOptions?.limit ?? null);
+
+    // Snowflake inlines every literal — no path carries bound params. Fail fast if a
+    // fragment ever emitted one (the reader rejects parameterized sqlOverride).
+    const paramCount = where.params.length + orderBy.params.length + limit.params.length;
+    if (paramCount > 0) {
```
romandubovyi, line 60: ⚪ **Nit:** This guard looks unreachable: `SnowflakeClauseRenderer.validateFragment` already throws per-fragment inside `renderWhere`, and the base `renderOrderBy`/`renderLimit` never emit params. Fine to keep as defense-in-depth, but it could be dropped to avoid implying there's a path that produces params here.

apps/backend/test/integration/snowflake.integration.ts
```diff
@@ -0,0 +1,484 @@
...
+
+    // -------------------------------------------------------------------------
+    // §0 PROBE 1 — Backslash round-trip
+    // -------------------------------------------------------------------------
+    // Row 5 stores name = 'a\b' (one literal backslash).
+    // The renderer doubles backslashes → emits `'a\\b'` in SQL.
+    // Snowflake interprets `\\` as a single `\`, so the WHERE matches row 5.
+    // If escaping is wrong the result count is 0 and we fail loudly.
+
+    it('§0 PROBE 1 — backslash round-trip: eq "a\\b" → row 5 (count must be 1)', async () => {
```
romandubovyi, line 202: 🟢 **Praise:** Since option B makes literal escaping the only injection barrier, the depth of verification here is exactly right: live round-trips for backslash and single-quote escaping, dedicated malicious-value and malicious-column-name tests, and the fail-closed escaper change. This is the level of care inline-literal renderers usually lack.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-blended-query-builder.ts
```diff
@@ -2,17 +2,32 @@ import { Injectable } from '@nestjs/common';
...
+  }
+
+  protected get clauseRenderer(): SqlClauseRenderer {
+    return this._renderer;
+  }
+
+  // Snowflake folds UNQUOTED identifiers to UPPERCASE, so the base class's bare return
+  // for simple names would miss the lowercase columns the OWOX connector creates (as
+  // quoted). Always quote — case-stable, and consistent with the non-blended builder.
+  protected quoteIdentifier(name: string): string {
```
max-voloshyn, line 23: 🔴 Blocking: This now quotes the CTE alias as "main", but AbstractBlendedQueryBuilder still emits the final query as FROM main and qualifies native fields as main."col". Snowflake folds those unquoted references to MAIN, so blended Snowflake reports with output controls produce SQL that cannot resolve the "main" CTE. I confirmed this branch generates: WITH "main" AS (...) SELECT main."a" FROM main WHERE main."a" = 1. Please quote the hardcoded main references consistently or avoid quoting the main CTE alias.
🤖 Reviewed by Codex (GPT-5.5)

.changeset/snowflake-output-controls.md
```diff
@@ -0,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Add output controls for Snowflake data marts
+
+Add report-level output controls (filters, slices, sort, limit) for Snowflake data marts. Filter values are inlined as escaped SQL literals (Snowflake-correct backslash + quote escaping); substring matchers use CONTAINS/STARTSWITH/ENDSWITH and regex uses RLIKE so user `%`/`_` stay literal; date/time comparisons are CAST to the column type. Works for the web app and the Google Sheets extension.
```
max-voloshyn: 🟡 Non-blocking: The changeset says regex uses RLIKE, but SnowflakeClauseRenderer renders REGEXP_INSTR(...) > 0 / = 0. Please align the release note with the actual implementation so the published changelog does not document a different regex strategy.
🤖 Reviewed by Codex (GPT-5.5)

.github/workflows/test-integration.yml
```diff
@@ -73,6 +75,15 @@ jobs:
...
           REDSHIFT_DATABASE: ${{ secrets.REDSHIFT_DATABASE }}
+          # Snowflake (PASSWORD auth = PAT in the password field; the account needs a
+          # network policy that permits GitHub Actions runners, or the auth will fail)
+          SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
+          SNOWFLAKE_WAREHOUSE: ${{ secrets.SNOWFLAKE_WAREHOUSE }}
+          SNOWFLAKE_USERNAME: ${{ secrets.SNOWFLAKE_USERNAME }}
+          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
+          SNOWFLAKE_DATABASE: ${{ secrets.SNOWFLAKE_DATABASE }}
+          SNOWFLAKE_SCHEMA: ${{ secrets.SNOWFLAKE_SCHEMA }}
+          SNOWFLAKE_ROLE: ${{ secrets.SNOWFLAKE_ROLE }}
```
max-voloshyn: 🟡 Non-blocking: Setting SNOWFLAKE_ROLE here currently has no effect because SnowflakeApiAdapter never copies config.role into the Snowflake SDK connection options. If the CI user needs this role to access SNOWFLAKE_DATABASE/SNOWFLAKE_SCHEMA, the new live suite still runs under the default role. Please wire the role into the adapter connection or remove this env/config path from the tests.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-blended-query-builder.ts
zapolsky: Good catch — fixed. The root cause was in `AbstractBlendedQueryBuilder`: it defined the home CTE as `"main"` but emitted bare `FROM main` / `main."col"` references. Every `main` reference now goes through `quoteIdentifier('main')`, so the definition and all references quote consistently — `WITH "main" AS (...) SELECT "main"."a" FROM "main" WHERE "main"."a" = 1`. It's a no-op for the case-insensitive engines (`quoteIdentifier('main')` stays bare there), and there's a new regression test asserting the consistent quoting.

.changeset/snowflake-output-controls.md
zapolsky: Fixed — the changeset now documents `REGEXP_INSTR(...) > 0` for regex, matching the renderer.

.github/workflows/test-integration.yml
zapolsky: You're right — the adapter never copies `config.role` into the SDK connection, so the env was a no-op. Dropped `SNOWFLAKE_ROLE` from the workflow and both integration suites. Actually wiring `config.role` into the adapter is a real (pre-existing) gap, but it's out of this PR's scope — tracking it as a separate follow-up.

apps/backend/src/data-marts/data-storage-types/snowflake/utils/snowflake-identifier.utils.ts
zapolsky: Good point — it now throws `BusinessViolationException` so a 4-part `fullyQualifiedName` surfaces as a 400 instead of a 500. The definition dry-run usually catches a bad FQN earlier, but this makes the fail-closed path return a proper validation error too.

apps/backend/test/integration/snowflake.integration.ts
zapolsky: Finalized: the CAST stays as defensive-only — the live run confirmed Snowflake also coerces a bare literal, so it's explicit-not-required. PROBE 2 now asserts the bare-literal form works, and the renderer's "may relax" note is removed.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-query.builder.ts
zapolsky: Agreed it's unreachable today — keeping it as defense-in-depth since it documents the `params: []` invariant and matches the guard in the sibling builders.

PR #1304, opened 2026-06-10
feat(backend): show each column's description in Google Sheets header notes

laskevych:
## Problem

When a report is exported to Google Sheets, each imported column gets a note
on its header cell. We previously put the full provenance block (the user's
column description + ODM metadata: import time, data mart title, link) on the
**first** column only, and every other imported column got just a short
`--- Imported via OWOX Data Marts ---` marker.

That dropped the per-column **description** for every column except the first.
The Product Manager expects each column to keep its own `header.description`
in the note, so users can see what a column means from its header regardless
of its position in the range.

## Change

Non-first imported columns now carry their own `header.description` **before**
the ODM marker, mirroring how the first column places its description before
the full provenance block. The first column is unchanged (description + full
metadata); only the previously bare columns gain their description back.

Implementation:

- `SheetMetadataFormatter.buildImportedColumnMarker(description, isCommunityEdition)`
  now prepends the column description before the marker.
- Extracted a shared private `assembleColumnNote(description, infoBlock)` used
  by both `buildImportedColumnNote` (full block) and `buildImportedColumnMarker`
  (bare marker), so description placement, the blank-line separator, the
  description-length truncation, and the final note-size cap live in one place
  instead of being duplicated.
- `google-sheets-report-writer` passes `header.description` into the marker.

An empty/undefined description collapses to just the marker, so columns
without a description look exactly as before — no leading blank line.

## Result

Note on a **non-first** imported column:

**Before**
```
--- Imported via OWOX Data Marts ---
```

**After**
```
Total clicks per campaign

--- Imported via OWOX Data Marts ---
```

First-column note is unchanged:
```
Country of the user

--- Imported via OWOX Data Marts ---
Imported at 2026 Jun 10, 12:00:00 UTC
Data Mart: My Data Mart
Data Mart page: https://app.owox.com/ui/.../data-setup
```

## Tests

- `sheet-metadata-formatter.spec.ts`: `buildImportedColumnMarker` now covers
  description + marker, empty/undefined description → bare marker, the
  Community Edition suffix, and oversize-description truncation.
- `google-sheets-report-writer.spec.ts`: the per-column-note test now gives
  every column a description and asserts the first column's description reaches
  `buildImportedColumnNote` while each non-first column's description reaches
  `buildImportedColumnMarker`.
- All 48 unit tests in the two suites pass; ESLint and Prettier clean on the
  changed files.

### DIFF
--- a/.changeset/bigquery-large-view-memory.md
+++ b/.changeset/bigquery-large-view-memory.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Faster and more reliable Google Sheets import from BigQuery views
+
+Data marts based on a BigQuery VIEW now import into Google Sheets faster and more reliably, even for very large datasets. Previously, large views could stop responding partway through and deliver no data at all; now rows are loaded steadily and the report completes successfully.
--- a/apps/backend/src/data-marts/data-destination-types/google-sheets/services/google-sheets-report-writer.spec.ts
+++ b/apps/backend/src/data-marts/data-destination-types/google-sheets/services/google-sheets-report-writer.spec.ts
@@ -683,15 +683,18 @@ describe('GoogleSheetsReportWriter — preserves user column formats across refr
 });
 
 describe('GoogleSheetsReportWriter — per-column header notes', () => {
-  it('writes the full ODM note only on the first column and the short marker on the rest', async () => {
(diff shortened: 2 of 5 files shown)


prokopetsv (approved):
Awesome, thanks!

PR #1309, opened 2026-06-10
Handle disconnected report columns and output controls

kalnik-a-a:
## Summary

- Fail fast when a report references disconnected or hidden selected columns/output controls instead of letting stale references leak into generated SQL or fail later in storage with a cryptic error.
- Reuse recursive schema path collection for native fields so nested native report fields stay supported while disconnected/hidden schema subtrees are pruned consistently.
- Surface stale selected columns, filter-only references, and slices in the Report Column Picker disconnected block so they can be removed from the picker.
- Keep sort-only stale references in the Output Controls menu and highlight stale filters, slices, and sorts there in red.
- Add a red Output Controls badge/count when any filter, slice, or sort references a disconnected field, matching the unresolved state in the menu.
- Share the remove-only active-rules popover between disconnected and no-access rows so stale filter/slice rules can be reviewed and removed without add/edit controls.
- Tighten cache invalidation cleanup so it does not prepare a fresh blended reader while deleting cache entries.

## Related

- Google Sheets extension sync PR: https://github.com/OWOX/google-sheets-extension/pull/21

## Screenshots

Add screenshots of the updated Report Column Picker / Output Controls UI before moving this PR out of draft.

| State | Screenshot |
| --- | --- |
| Disconnected columns block | CleanShot 2026-06-11 at 00 00 20 |
| Output Controls red badge and menu | CleanShot 2026-06-11 at 00 35 39 |
| Stale slice/filter remove-only popover | CleanShot 2026-06-11 at 00 36 21 |

## Test Plan

- [x] `npm test -w @owox/backend -- output-controls-validator.service.spec.ts blended-report-data.service.spec.ts report-sql-composer.service.spec.ts report-data-cache.service.spec.ts`
- [x] `npm test -w @owox/web -- ReportColumnPicker.test.tsx OutputSettingsDropdown.test.tsx RowFilterIcon.test.tsx`
- [x] `npm run test:e2e -w @owox/backend -- --shard=1/3 --runInBand`
- [x] `npm run test:e2e -w @owox/backend -- --shard=2/3 --runInBand`
- [x] `npm run test:e2e -w @owox/backend -- --shard=3/3 --runInBand`
- [x] `npx eslint ...` for touched backend files
- [x] `npx eslint ...` for touched web ReportColumnPicker files
- [x] `npx tsc -p tsconfig.build.json --noEmit --incremental false`
- [x] `npm run type-check -w @owox/web`
- [x] `npx prettier --check ...` for touched backend/web/changeset files
- [x] `git diff --check`

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
+++ b/apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
@@ -1,10 +1,37 @@
 import { z } from 'zod';
 import { DataStorageCredentials } from './data-storage-credentials.type';
+import type { DataMartSchemaField } from './data-mart-schema.type';
 import { DataMartSchemaFieldStatus } from './enums/data-mart-schema-field-status.enum';
 import { DataStorageType } from './enums/data-storage-type.enum';
 import { Injectable } from '@nestjs/common';
 import { DataStoragePublicCredentialsFactory } from './factories/data-storage-public-credentials.factory';
 import { DataStorageCredentialsPublic } from '../dto/presentation/data-storage-response-api.dto';
+
+// A hidden-for-reporting or DISCONNECTED node prunes its whole subtree; container
+// nodes count as referenceable paths alongside their nested `a.b` leaves.
+export function collectSchemaFieldPaths(
+  fields: readonly DataMartSchemaField[],
+  prefix = ''
+): string[] {
+  return collectSchemaFieldPathTypes(fields, prefix).map(field => field.name);
+}
+
+export function collectSchemaFieldPathTypes(
+  fields: readonly DataMartSchemaField[],
+  prefix = ''
+): { name: string; type: string }[] {
+  const result: { name: string; type: string }[] = [];
+  for (const field of fields) {
+    if (field.isHiddenForReporting) continue;
+    if (field.status === DataMartSchemaFieldStatus.DISCONNECTED) continue;
+    const fullName = prefix ? `${prefix}.${field.name}` : field.name;
+    result.push({ name: fullName, type: String(field.type) });
+    if ('fields' in field && field.fields?.length) {
+      result.push(...collectSchemaFieldPathTypes(field.fields, fullName));
+    }
+  }
+  return result;
+}
 
 export function createBaseFieldSchemaForType<T extends z.ZodTypeAny>(schemaFieldType: T) {
   const typedSchema = z
--- a/apps/backend/src/data-marts/services/blended-report-data.service.ts
+++ b/apps/backend/src/data-marts/services/blended-report-data.service.ts
@@ -20,11 +20,17 @@ import { DataStorageType } from '../data-storage-types/enums/data-storage-type.e
 import { StorageFieldType } from '../dto/domain/storage-field-type';
 import { computeEffectiveType } from '../data-storage-types/field-aggregation';
 import { BlendingDecision } from '../dto/domain/blending-decision.dto';
+import { DataMart } from '../entities/data-mart.entity';
 import { DataMartRelationship } from '../entities/data-mart-relationship.entity';
+import {
+  collectSchemaFieldPaths,
+  collectSchemaFieldPathTypes,
+} from '../data-storage-types/data-mart-schema.utils';
 import { PublicOriginService } from '../../common/config/public-origin.service';
 import { BusinessViolationException } from '../../common/exceptions/business-violation.exception';
 import { buildDataMartUrl } from '../../common/helpers/data-mart-url.helper';
 import { UserProjectionsFetcherService } from './user-projections-fetcher.service';
+import { throwDisconnectedReportColumnsError } from '../errors/disconnected-report-columns.error';
 
 @Injectable()
 export class BlendedReportDataService {
@@ -107,6 +113,8 @@ export class BlendedReportDataService {
     );
 
     const blendedFieldsByName = new Map(blendableSchema.blendedFields.map(f => [f.name, f]));
+    this.assertNoOrphanedColumnReferences(dataMart, columnConfig, blendedFieldsByName);
+
     const referencedColumns = new Set<string>([
       ...columnConfig,
       ...postJoinFilterColumns,
@@ -194,7 +202,9 @@ export class BlendedReportDataService {
   // blended output aliases; pre-join slices target subsidiary raw columns by aliasPath.
   private buildBlendedColumnTypes(blendableSchema: BlendableSchemaDto): BlendedColumnTypes {
     const postJoin = new Map<string, string>();
-    for (const nf of blendableSchema.nativeFields) postJoin.set(nf.name, String(nf.type));
+    for (const nf of collectSchemaFieldPathTypes(blendableSchema.nativeFields)) {
+      postJoin.set(nf.name, nf.type);
+    }
     for (const bf of blendableSchema.blendedFields) postJoin.set(bf.name, bf.type);
 
     const preJoin = new Map<string, Map<string, string>>();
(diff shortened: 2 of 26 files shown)


zapolsky (approved):
✅ Approved by automated OWOX review — no blocking issues. See top-level comment for non-blocking notes.

zapolsky:

Solid, thoroughly-tested PR — the core disconnected/orphaned-column detection (schema walker, blended-report orphan assertion, validator reclassification, recursive `columnTypes` for nested filters) is correct against the spec, consistent between the run path and the SQL-preview path, and well covered by unit + e2e + React tests. No correctness, security, data-loss, auth, or architecture violations found. Only minor notes:

- `apps/backend/src/data-marts/errors/disconnected-report-columns.error.ts:3` — Implemented as a `throwDisconnectedReportColumnsError(): never` function throwing an inline `BusinessViolationException`, whereas sibling errors in `data-marts/errors/` are classes. Works and maps to HTTP 400 correctly, but can't be matched with `instanceof`. Consider a `DisconnectedReportColumnsException extends BusinessViolationException` for convention consistency.
- `apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx:2451` — The red "disconnected" badge predicate tests post-join filter columns against `dropdownColumnNames` (native + included-and-accessible blended fields), while the actual block and the backend orphan check treat any non-hidden blended field as known. A post-join filter on an excluded-but-non-hidden blended column would light the badge red while the block shows nothing and the backend accepts it — a harmless false-positive, but the badge and block/backend predicates should share the same "known" set.
- `apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx:2532` — Stale rows in the disconnected block pass `fieldType={fieldTypeByName.get(name) ?? 'STRING'}`; for a truly disconnected column the type is unknown, so defaulting to `'STRING'` can mislabel operators in the view-only popover. Cosmetic.
- `apps/backend/src/data-marts/services/report-data-cache.service.ts:233` — `finalizeExpiredCacheEntry` now restores+finalizes only for `AWS_ATHENA` and skips other storages (previously every storage with `readerState` finalized). The comment + tests justify this (Athena is the only reader with external S3 artifacts), but it's a real behavior change — worth confirming no other current/future reader has external artifacts to release on cleanup.
- `.changeset/fail-fast-orphaned-report-columns.md:5` — Changeset is valid; minor nit: it opens with a `#` heading whereas sibling changesets start directly with prose.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -588,6 +652,54 @@ export function ReportColumnPicker({
...
+        if (!rule.aliasPath || !joinedColumnKeys.has(`${rule.aliasPath}\0${rule.column}`)) {
+          return true;
+        }
+      } else if (!dropdownColumnNames.has(rule.column)) {
+        return true;
+      }
+    }
+
+    return effectiveOutputConfig.sortConfig.some(
+      rule => !selectedDropdownColumnNames.has(rule.column)
```
romandubovyi: 🟡 **Important:** This treats a sort rule on a deselected-but-existing column the same as a truly disconnected one: the badge turns red with the "Disconnected output controls" aria-label, and `SortRow` shows "Column not found in schema" with a strikethrough. The backend distinguishes these cases (`SORT_COLUMN_NOT_SELECTED` vs the disconnected-columns error), and the fix for the user is different too — re-selecting the column vs removing the rule. Consider either pruning sort rules in `toggleField` when their column is unchecked, or using a distinct message/state for "column not selected" sort rules.

apps/backend/src/data-marts/services/blended-report-data.service.ts
```diff
@@ -347,6 +357,32 @@ export class BlendedReportDataService {
...
+  // Hidden fields are equally unavailable: schema-hidden ones are excluded from
+  // reporting by definition, and blend-hidden ones are dropped from the final SELECT
+  // while their data headers are still emitted — selecting either must fail here.
+  private assertNoOrphanedColumnReferences(
+    dataMart: DataMart,
+    columnConfig: string[],
+    blendedFieldsByName: ReadonlyMap<string, BlendedFieldDto>
+  ): void {
+    const schemaFields = dataMart.schema?.fields ?? [];
+    if (schemaFields.length === 0) return;
```
romandubovyi, line 372: 🔵 **Question:** This orphan check deliberately skips when the schema is not actualized (`schemaFields.length === 0`), but `validateForReport` has no such guard — with an empty schema every filter column throws the disconnected-columns error, whose message suggests restoring "the previous schema or joined data marts setup", which is misleading for a mart that never had a schema actualized. The old e2e covered exactly that empty-schema path before the schema PUT was added in this PR. Is the asymmetry intentional, or should the validator skip (or use a different message) for non-actualized schemas too?

apps/backend/src/data-marts/services/report-data-cache.service.ts
```diff
@@ -235,19 +208,29 @@ export class ReportDataCacheService {
...
-        const { options } = await this.resolvePrepareOptions(
-          cacheEntry.report,
-          this.cleanupAccessor,
-          false
+      if (!cacheEntry.readerState) {
+        this.logger.debug(`No reader state to finalize for cache entry ${cacheEntry.id}`);
+        return;
+      }
+
+      if (cacheEntry.storageType !== DataStorageType.AWS_ATHENA) {
```
romandubovyi, line 225: 💡 **Suggestion:** Whether a reader can be safely finalized from persisted state is knowledge that belongs to the reader, but here it's hardcoded as a `storageType !== AWS_ATHENA` check in the cache service. When a new storage with external artifacts is added, this spot is easy to miss — cleanup would silently skip it with only a debug log, leaking artifacts. A capability flag on the reader interface (e.g. `supportsStateOnlyFinalize`) would keep this decision next to each implementation.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -330,4 +351,36 @@ export class OutputControlsValidatorService {
...
+  private collectDisconnectedOutputControlRefs(
+    errors: ValidationError[],
+    sort: SortRule[],
+    knownOutputColumns: ReadonlySet<string>
+  ): string[] {
+    const refs: string[] = [];
+
+    for (const error of errors) {
+      switch (error.code) {
+        case 'FILTER_COLUMN_UNKNOWN':
```
romandubovyi, line 369: ⚪ **Nit:** The `FILTER_COLUMN_UNKNOWN` case body is identical to the grouped `FILTER_ALIAS_PATH_UNKNOWN` / `FILTER_ALIAS_PATH_NOT_INCLUDED` cases below — all three can share one case group.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
```diff
@@ -1,11 +1,38 @@
...
+// A hidden-for-reporting or DISCONNECTED node prunes its whole subtree; container
+// nodes count as referenceable paths alongside their nested `a.b` leaves.
+export function collectSchemaFieldPaths(
+  fields: readonly DataMartSchemaField[],
+  prefix = ''
+): string[] {
+  return collectSchemaFieldPathTypes(fields, prefix).map(field => field.name);
+}
+
+export function collectSchemaFieldPathTypes(
```
romandubovyi, line 19: 🟢 **Praise:** Nice extraction — replacing the local walker copy with shared `collectSchemaFieldPaths`/`collectSchemaFieldPathTypes` reused across the blended service, validator, SQL composer, and http-data util, plus the explicit FE/BE sync comments on both walkers. The nested-struct test coverage on all the consuming paths is thorough too.

kalnik-a-a:
Text copy updated. New UI tooltip text: They are missing from the current Data Mart output schema. Uncheck them to remove them from the report, or contact your analyst to restore the schema. Backend validation error text now uses the same wording after the disconnected columns list: Cannot build report SQL. Disconnected columns: "<columns>". They are missing from the current Data Mart output schema. Uncheck them to remove them from the report, or contact your analyst to restore the schema.

apps/backend/src/data-marts/services/report-data-cache.service.ts
kalnik-a-a: Leaving this scoped as-is in 46a253a04. The current cleanup path intentionally only restores Athena readers from persisted state because Athena has external S3 result artifacts to clean up; the other current readers either have no cleanup work or require a live prepared cursor/session. A reader-level capability would be a good separate cleanup, but I would keep it out of this disconnected-fields follow-up.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
kalnik-a-a: Fixed in 46a253a04. FILTER_COLUMN_UNKNOWN now shares the same case group with FILTER_ALIAS_PATH_UNKNOWN and FILTER_ALIAS_PATH_NOT_INCLUDED.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
kalnik-a-a: Leaving this as-is in 46a253a04. This state is transient only: save validation rejects sort rules on non-selected columns with SORT_COLUMN_NOT_SELECTED, so it cannot persist through the normal API path. A more precise not-selected sort UX can be handled separately if needed.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
kalnik-a-a: Acknowledged, keeping this extraction as-is.

apps/backend/src/data-marts/services/blended-report-data.service.ts
kalnik-a-a: Fixed in 46a253a04. validateForReport now skips schema-based disconnected-column classification while the computed schema is empty/non-actualized, matching the orphan-column check behavior. Added regression coverage for filters/sorts on a not-yet-actualized schema.

romandubovyi (approved):

PR #1311, opened 2026-06-11
feat(data-marts): output controls for Databricks data marts (filters,…

zapolsky:
… slices, sort, limit)

Enable report-level output controls for the DATABRICKS storage — the last non-GBQ storage in the epic. Option B (same shape as Redshift/Snowflake): DatabricksClauseRenderer inlines every filter value as a SQL literal and always returns params: [], so there is no bound-param plumbing — the report reader (already option-B-ready), composeStatic (early- returns on empty params), and the adapter need no changes.

Per-dialect (Spark SQL): contains/startswith/endswith (never LIKE so %/_ stay literal), infix RLIKE (partial match), <> for neq, = TRUE/FALSE for bool, backslash-first literal escaping, and date_add/add_months/trunc for relative_date (half-open ranges, upper bounds on every preset). One robust backtick escaper for columns and FQNs; no quoteIdentifier override (Spark is case-insensitive). Defensive CAST(... AS DATE|TIMESTAMP|TIMESTAMP_NTZ).

Validator + web + extension type matrices gain INT and TIMESTAMP_NTZ (Databricks has no standalone TIME type). Capability + FE mirrors add DATABRICKS.

Verified: nest build, full backend unit suite, emission e2e (generated-sql + Looker cache path), web tsc + vitest, extension typecheck, and 31/31 LIVE integration (operator matrix
+ relative_date + escaping/regex/CAST probes + full HTTP Data API stack) against a real Databricks SQL warehouse. Live findings: Spark RLIKE is partial (no anchoring fix needed), CAST is defensive-only (Spark coerces bare literals), backslash round-trip confirmed.

### DIFF
--- a/apps/backend/test/integration/README.md
+++ b/apps/backend/test/integration/README.md
@@ -50,6 +50,17 @@ ATHENA_OUTPUT_BUCKET=my-athena-results-bucket
 # Athena database name (must already exist — create via: CREATE DATABASE name)
 ATHENA_DATABASE=my_test_database
 
+# === Databricks ===
+# Workspace host (with https://) of the SQL warehouse
+DATABRICKS_HOST=https://dbc-xxxxxxxx.cloud.databricks.com
+# SQL warehouse HTTP path (Warehouse → Connection details → HTTP path)
+DATABRICKS_HTTP_PATH=/sql/1.0/warehouses/abcdef1234567890
+# Personal access token (User Settings → Developer → Access tokens)
+DATABRICKS_TOKEN=dapi...
+# Catalog + schema that host the seed table (the token needs CREATE/DROP TABLE there)
+DATABRICKS_CATALOG=main
+DATABRICKS_SCHEMA=default
+
 # === Google Sheets ===
 # JSON string of a GCP service account key with Google Sheets API access
 # (Editor role on TEST_GOOGLE_SPREADSHEET_ID — sheet add/delete requires Editor).
@@ -79,6 +90,12 @@ TEST_GOOGLE_SPREADSHEET_ID=your_spreadsheet_id_here
 - `athena:StartQueryExecution`, `athena:GetQueryExecution`, `athena:GetQueryResults`
 - `s3:GetObject`, `s3:PutObject`, `s3:DeleteObject`, `s3:ListBucket` (on output bucket)
 - `glue:GetTable`, `glue:CreateTable`, `glue:DeleteTable`, `glue:GetDatabase` (on Glue catalog)
+
+**Databricks token:**
+
+- `CREATE TABLE` / `DROP TABLE` on `DATABRICKS_CATALOG`.`DATABRICKS_SCHEMA` (the suite
+  seeds and drops its own table)
+- `SELECT` on that schema; the SQL warehouse must be running (or auto-start enabled)
 
 ## CI Setup (GitHub Actions)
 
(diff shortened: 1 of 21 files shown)


zapolsky:

> ℹ️ This PR is authored by the review-bot identity, so no formal approval review is submitted (GitHub blocks self-approval). The notes below are informational.

High-quality PR that extends the existing, well-vetted Option-B output-controls pattern to Databricks/Spark SQL. **SQL injection was verified safe** — values use `formatDatabricksLiteral` (Spark-correct backslash-then-quote escaping, proven inert against breakout payloads), identifiers go through `escapeDatabricksIdentifier` (backtick-doubling per dotted part), numbers re-assert `Number.isFinite`, substring matchers use `contains`/`startswith`/`endswith` + `RLIKE` (not `LIKE`), and the `params: []` invariant is guarded at multiple layers. No controllers touched (so `@Auth` N/A), no facade impl added. Only minor notes:

- `apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-query.builder.ts:800` — The SQL-definition fallback returns `(${cleanQuery})` with no derived-table alias, whereas the Redshift sibling returns `(${cleanQuery}) AS subq`. Spark tolerates unaliased derived tables (asserted + claimed live-verified), so not a bug, but aligning to the sibling keeps the dialect builders uniform.
- `apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-query.builder.spec.ts:651` — The bare-SQL-wrap-with-output-controls fallback (SQL definition, no `mainTableReference`) isn't unit-tested under output controls; the live suite uses TABLE definitions only. Low risk under Spark's leniency, but the branch most likely to differ from siblings is the least exercised — consider a unit assertion on the wrapped form.
- `apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-clause-renderer.ts:475` — `text()` calls `formatDatabricksLiteral(String(v))` directly, duplicating logic the `lit`/`litCast` path already has. It's correct (text operators are validator-restricted to string columns) and mirrors the Redshift renderer — minor readability nit only.

apps/backend/test/integration/README.md
```diff
@@ -50,6 +50,17 @@ ATHENA_OUTPUT_BUCKET=my-athena-results-bucket
 # Athena database name (must already exist — create via: CREATE DATABASE name)
 ATHENA_DATABASE=my_test_database
 
+# === Databricks ===
+# Workspace host (with https://) of the SQL warehouse
+DATABRICKS_HOST=https://dbc-xxxxxxxx.cloud.databricks.com
```
max-voloshyn: 🟡 Non-blocking: This example should use the hostname only, not a full `https://` URL. The app stores `config.host` directly and passes it to `DBSQLClient.connect({ host })`, and the existing Databricks UI/docs use hostname-only examples. If CI secrets are created from this README, the Databricks jobs can be misconfigured. Please update this example and the matching `DATABRICKS_HOST` comment in `databricks.integration.ts`.

🤖 Reviewed by Codex (GPT-5.5)

max-voloshyn (approved):

apps/backend/test/integration/README.md
zapolsky: Fixed in 66aac045c — DATABRICKS_HOST is now hostname-only (no https://) in the README example + comment, in the databricks.integration.ts env doc, and in the emission e2e config, matching how config.host is passed straight to DBSQLClient.connect({ host }). Thanks for the catch!

zapolsky:
Addressed the non-blocking notes in 66aac045c:
- SQL-definition fallback now aliases the derived table as `(...) AS subq`, matching the Redshift builder (both the columns-only and the output-controls paths).
- Added a unit assertion for the SQL-def + output-controls path with no `mainTableReference` (`FROM (SELECT 1) AS subq` plus the inlined WHERE), so the branch that differs from the table-def path is now covered.
- Left `text()` as-is: it mirrors the Redshift renderer and text operators are validator-restricted to string columns, so the `String(v)` coercion is intentional — parity/readability only.

PR #1319, opened 2026-06-12
feat(web): Add drag-and-drop support for service account JSON inputs

romandubovyi:
## Description

Currently, users have to manually open their Google Service Account JSON files, copy the raw JSON content, and paste it into the credentials textarea. This PR simplifies the onboarding flow by allowing users to drag and drop their service account JSON file directly onto the input field.

## Key Changes

- **Reusable Component**: Created `FileDropTextarea` in `@owox/ui` with drag-and-drop event listeners (`onDragOver`, `onDragLeave`, `onDrop`) and local file reading using `FileReader`.
- **Validation**: Added client-side verification to ensure that only files with `.json` extensions are accepted and that their content is valid JSON. Fails gracefully with user-friendly toast notifications.
- **UX/UI Improvements**: Added a styled glassmorphic overlay with backdrop blur and a bouncing `FileDown` icon to indicate when a file is hovered over the textarea.
- **Forms Integration**: Integrated `FileDropTextarea` in `GoogleSheetsFields.tsx`, `GoogleBigQueryFields.tsx`, and `LegacyGoogleBigQueryFields.tsx`.

drag_over_overlay

## Verification

- Verified that ESLint and Prettier formatting checks pass successfully.
- Verified that all unit tests (`npm run test`) pass.
- Verified that `@owox/web` compiles successfully.

### DIFF
--- a/.changeset/service-account-json-drag-and-drop.md
+++ b/.changeset/service-account-json-drag-and-drop.md
@@ -1,0 +1,7 @@
+---
+"owox": minor
+---
+
+# Drag-and-drop support for service account JSON inputs
+
+Added drag-and-drop support for service account JSON files on credentials textareas. Users can now drop their credential files directly to populate the field, with JSON validation, file size limits, and user-friendly error feedback.
--- a/apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
+++ b/apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
@@ -7,7 +7,8 @@ import {
   FormLabel,
   FormMessage,
 } from '@owox/ui/components/form';
-import { Textarea } from '@owox/ui/components/textarea';
+import { FileDropTextarea } from '@owox/ui/components/file-drop-textarea';
+import { toast } from 'react-hot-toast';
 import { useState, useEffect } from 'react';
 import { type UseFormReturn } from 'react-hook-form';
 import { type DataDestinationFormData, DataDestinationType } from '../../../shared';
@@ -262,11 +263,20 @@ export function GoogleSheetsFields({ form }: GoogleSheetsFieldsProps) {
                         </button>
                       </div>
                     ) : (
-                      <Textarea
+                      <FileDropTextarea
                         {...field}
                         className='min-h-[150px] font-mono'
                         rows={8}
-                        placeholder='Paste your service account JSON here'
+                        placeholder='Paste your service account JSON here or drag & drop the file'
+                        onFileRead={content => {
+                          form.setValue('credentials.serviceAccount', content, {
+                            shouldDirty: true,
+                            shouldValidate: true,
+                          });
+                        }}
+                        onFileReject={error => {
+                          toast.error(error);
+                        }}
                       />
                     )}
                   </FormControl>
--- a/apps/web/src/features/data-storage/edit/components/DataStorageEditForm/GoogleBigQueryFields.tsx
+++ b/apps/web/src/features/data-storage/edit/components/DataStorageEditForm/GoogleBigQueryFields.tsx
@@ -20,7 +20,8 @@ import GoogleBigQueryLocationDescription from './FormDescriptions/GoogleBigQuery
 import { Tooltip, TooltipContent, TooltipTrigger } from '@owox/ui/components/tooltip';
 import { Button } from '@owox/ui/components/button';
 import { ExternalAnchor } from '@owox/ui/components/common/external-anchor';
-import { Textarea } from '@owox/ui/components/textarea';
+import { FileDropTextarea } from '@owox/ui/components/file-drop-textarea';
+import { toast } from 'react-hot-toast';
 import { useState, useEffect } from 'react';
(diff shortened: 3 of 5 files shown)


zapolsky:

Deep multi-agent review against the worktree (`reviewer`, `logic-hunter`, fresh-eyes generalist). **Security — the part that matters most here — is clean:** the dropped service-account private key is read with `FileReader.readAsText` only and routed into react-hook-form state via `form.setValue(..., { shouldDirty: true, shouldValidate: true })`; it is never logged, never sent anywhere, never retained after `JSON.parse`, and the error toasts use static English strings that never echo the file content or name. Form wiring is correct and identical across all three consumers.

**On CI/lint:** `tsc -b` initially reported 9 errors and ESLint 6 (`no-unsafe-argument` on `content`/`error`), but these are a *review-harness artifact*: my worktree symlinks `node_modules`, so `@owox/ui/components/file-drop-textarea` (the new file) resolved to the **base** `packages/ui`, which doesn't contain it yet → `FileDropTextarea` typed as `any` → its callback params implicitly `any`. The worktree component types them correctly (`onFileRead?: (content: string) => void`, `onFileReject?: (message: string) => void`), and re-running `tsc` with `@owox/ui` aliased to the PR's `packages/ui/src` reports **0 errors**. `packages/ui/file-drop-textarea.tsx` itself lints clean. CI installs the workspace at HEAD, so it will be green — flagging only so the note isn't mistaken for a real failure.

- **A missed drop navigates the tab to the local key file and loses unsaved form state.** `preventDefault` for dragover/drop is bound only to the `FileDropTextarea` wrapper, and there is no app/window-level drop guard, so releasing the file a few px outside the textarea triggers the browser default (navigate to `file:///…/service-account.json`). This is *pre-existing* default browser behavior app-wide (the PR strictly reduces the unguarded surface, never worsens it — verified, hence non-blocking), but since this PR is the feature that actively invites users to drag credential files onto the page, it's the natural place to add a document/form-level `preventDefault` on `dragover`/`drop` so a stray drop is a no-op.

- **Drag overlay flickers because `dragleave` isn't guarded against child elements.** `handleDragLeave` (`file-drop-textarea.tsx:21-24`) unconditionally `setIsDragging(false)`; since the wrapper contains the `<Textarea>` child, `dragleave` bubbles up as the cursor crosses onto it and strobes the overlay (and can leave it stale on Esc/out-of-bounds). It's only cosmetic (the overlay is `pointer-events-none`, so typing/paste are never blocked — verified), but the fix is small: `if (e.currentTarget.contains(e.relatedTarget as Node)) return;` (or an enter/leave counter), plus a reset on `dragend`.

- **Validation only checks JSON well-formedness, not service-account shape, and there's no success feedback.** Any valid JSON (`{}`, `[]`, `42`) passes (`file-drop-textarea.tsx:40-49`) and is written to the credentials field; combined with there being no success toast, dropping a non-service-account JSON gives the user zero signal that it isn't a real key (the failure only surfaces later downstream). Consider asserting a minimal shape after parse (`parsed.type === 'service_account' && parsed.client_email && parsed.private_key`) and/or a success toast.

- **Minor / robustness:** (a) the JSON-validity gate is hard-coded to `.json` while `allowedExtensions` is a configurable prop — a future consumer passing a broader list would get files accepted with no content validation (latent; the 3 current callers use the default); (b) no file-size cap before `readAsText`, so an accidental huge-file drop reads fully into memory and can freeze the tab (SA keys are ~2 KB — a small cap is cheap insurance); (c) a multi-file drop silently uses `files[0]` with no feedback; (d) the drop updates the field via `form.setValue` rather than `field.onChange` — functionally correct (dirty + validate fire) but it splits the update path; moving the `onChange` call and the drop-navigation guard *into* the shared component would keep the three identical call sites from drifting.

<sub>Automated OWOX review · worktree multi-agent passes · no blockers. The candidate "navigate-away" blocker was adversarially verified → refuted (pre-existing app-wide browser default, local file open, recoverable; the PR reduces the unguarded surface). Security traced clean end-to-end. `tsc`/ESLint green against the PR's `packages/ui` — the initial errors are a node_modules-symlink resolution artifact of the review harness, not CI.</sub>

zapolsky (approved):

romandubovyi:
Hi @zapolsky, thanks for the review! All notes have been addressed in the latest commit:

1. **Stray drop navigation guard:** Added window-level event listeners for `dragover`/`drop` inside `FileDropTextarea` (during mount) to call `e.preventDefault()`, preventing the tab from navigating away and losing unsaved state on misdirected drops.
2. **Overlay flickering fix:** Guarded the wrapper's `dragleave` handler by checking if the cursor is entering a child element (`e.currentTarget.contains(...)`), and added a `dragend` reset callback.
3. **Strict key validation:** Restructured file processing to check Google Service Account JSON keys (`type === 'service_account'`, `project_id`, `private_key`, `client_email`), rejecting malformed JSONs immediately with a specific toast message.
4. **Size limits & multi-file protection:** Enforced a maximum file size limit of 1MB (to prevent tab freezing on heavy files) and a single-file drop policy (rejecting multi-file uploads).
5. **Dynamic extension checking:** Replaced hardcoded `.json` extensions logic with checks against the `allowedExtensions` prop.
6. **Changeset:** Added minor changeset file.

PR #1320, opened 2026-06-12
fix(backend): escape identifiers in data mart sample-data SQL

laskevych:
## What changed

`DataMartSampleDataService.sampleColumns()` built its sampling query by interpolating the raw fully qualified table name and column names into SQL:

```ts
const sql = `SELECT ${columnList} FROM ${fqn} LIMIT ${limit}`;
```

For data marts whose table path contains dashes (e.g. `my-project.dataset.my-table-name`), BigQuery rejects the query with `Syntax error: Expected end of input but got "-"`, which breaks AI Helper metadata generation (`POST /data-marts/:id/ai-helper/triggers`) and the `SAMPLE_TABLE_DATA` AI tool.

This PR adds identifier escaping to that query, following the existing storage-type component pattern (interface → per-storage implementations → `TypeResolver` → facade):

- `IdentifierEscaper` interface (`TypedComponent<DataStorageType>`)
- Six thin implementations delegating to the existing per-storage escaping utils (`escapeBigQueryIdentifier`, `escapeAthenaIdentifier`, `escapeSnowflakeIdentifier`, `escapeRedshiftIdentifier`, `escapeDatabricksIdentifier`)
- `IDENTIFIER_ESCAPER_RESOLVER` registered in `data-storage-providers.ts`
- `IdentifierEscaperFacade` registered in `data-storage-facades.ts`
- `DataMartSampleDataService` now escapes both the table reference and the column list per the data mart's storage type, so the generated SQL becomes ``SELECT `report_date` FROM `my-project`.`dataset`.`my-table-name` LIMIT 5``

## Why

Every other SQL construction path (query builders for all storage types) already escapes identifiers; this service was the only exception. Escaping the column list also stops LLM-supplied tool arguments from being interpolated into SQL unquoted.

## Notes for reviewers

- The escapers handle already-quoted parts without double-quoting, so pre-escaped FQNs from the AI prefetch context remain safe.
- `resolveTableName()` intentionally stays unescaped — its raw output is also used outside SQL (LLM context, report composers).
- Known same-class issue left out of scope: the `${DATA_MART_TABLE}` macro replacement in `data-mart-sql-table.service.ts` also inserts the raw name, but needs double-quoting protection for user SQL that already wraps the macro in backticks; planned as a follow-up.

## Testing

- New `identifier-escaper.facade.spec.ts`: dashed FQN escaping per storage type (incl. Snowflake's unquoted database part), no double-quoting, unknown-type error.
- New `data-mart-sample-data.service.spec.ts`: generated SQL is fully escaped for resolved and explicitly passed table names.
- `jest src/data-marts`: 3268 passed; the 2 failures in `project-list-sql-dialects.spec.ts` reproduce on a clean tree without these changes (pre-existing parser issue).

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/facades/identifier-escaper.facade.spec.ts
+++ b/apps/backend/src/data-marts/data-storage-types/facades/identifier-escaper.facade.spec.ts
@@ -1,0 +1,48 @@
+import { TypeResolver } from '../../../common/resolver/type-resolver';
+import { AthenaIdentifierEscaper } from '../athena/services/athena-identifier.escaper';
+import { BigQueryIdentifierEscaper } from '../bigquery/services/bigquery-identifier.escaper';
+import { LegacyBigQueryIdentifierEscaper } from '../bigquery/services/legacy/legacy-bigquery-identifier.escaper';
+import { DatabricksIdentifierEscaper } from '../databricks/services/databricks-identifier.escaper';
+import { DataStorageType } from '../enums/data-storage-type.enum';
+import { IdentifierEscaper } from '../interfaces/identifier-escaper.interface';
+import { RedshiftIdentifierEscaper } from '../redshift/services/redshift-identifier.escaper';
+import { SnowflakeIdentifierEscaper } from '../snowflake/services/snowflake-identifier.escaper';
+import { IdentifierEscaperFacade } from './identifier-escaper.facade';
+
+describe('IdentifierEscaperFacade', () => {
+  const facade = new IdentifierEscaperFacade(
+    new TypeResolver<DataStorageType, IdentifierEscaper>([
+      new BigQueryIdentifierEscaper(),
+      new LegacyBigQueryIdentifierEscaper(),
+      new AthenaIdentifierEscaper(),
+      new SnowflakeIdentifierEscaper(),
+      new RedshiftIdentifierEscaper(),
+      new DatabricksIdentifierEscaper(),
+    ])
+  );
+
+  const fqn = 'test-project.TestDataset.dashed-table-name';
+
+  it.each([
+    [DataStorageType.GOOGLE_BIGQUERY, '`test-project`.`TestDataset`.`dashed-table-name`'],
+    [DataStorageType.LEGACY_GOOGLE_BIGQUERY, '`test-project`.`TestDataset`.`dashed-table-name`'],
+    [DataStorageType.AWS_ATHENA, '"test-project"."TestDataset"."dashed-table-name"'],
+    [DataStorageType.AWS_REDSHIFT, '"test-project"."TestDataset"."dashed-table-name"'],
+    [DataStorageType.DATABRICKS, '`test-project`.`TestDataset`.`dashed-table-name`'],
+    [DataStorageType.SNOWFLAKE, 'test-project."TestDataset"."dashed-table-name"'],
+  ])('escapes a dashed fully qualified name for %s', async (type, expected) => {
+    await expect(facade.escapeIdentifier(type, fqn)).resolves.toBe(expected);
+  });
+
+  it('does not double-quote already escaped parts', async () => {
+    await expect(
+      facade.escapeIdentifier(DataStorageType.GOOGLE_BIGQUERY, '`test-project`.TestDataset.`table`')
+    ).resolves.toBe('`test-project`.`TestDataset`.`table`');
+  });
+
+  it('throws for an unregistered storage type', async () => {
+    await expect(facade.escapeIdentifier('UNKNOWN' as DataStorageType, fqn)).rejects.toThrow(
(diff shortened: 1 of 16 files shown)


apps/backend/src/data-marts/services/data-mart-sample-data.service.ts
```diff
@@ -29,8 +31,12 @@ export class DataMartSampleDataService {
       fullyQualifiedTableName ??
       (await this.dataMartTableReferenceService.resolveTableName(dataMartId, projectId));
 
-    const columnList = columns.join(', ');
-    const sql = `SELECT ${columnList} FROM ${fqn} LIMIT ${limit}`;
+    const storageType = dataMart.storage.type;
+    const escapedColumns = await Promise.all(
+      columns.map(column => this.identifierEscaperFacade.escapeIdentifier(storageType, column))
```
kalnik-a-a, line 36: **Blocking:** This still leaves the Snowflake `SAMPLE_TABLE_DATA` path vulnerable to SQL breakout through `columns`. The new facade delegates Snowflake escaping to `escapeSnowflakeIdentifier`, but that helper returns the raw identifier for unexpected multi-part inputs and does not escape embedded double quotes. Since tool input currently accepts arbitrary non-empty column strings, a value such as `a.b.c.d FROM sensitive_table --` can become `SELECT a.b.c.d FROM sensitive_table -- FROM ...`, bypassing the intended Data Mart table. Please make the Snowflake escaper fail closed or quote/escape every accepted column segment safely before this interpolation path is used.

apps/backend/src/data-marts/data-storage-types/facades/identifier-escaper.facade.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const fqn = 'test-project.TestDataset.dashed-table-name';
+
+  it.each([
+    [DataStorageType.GOOGLE_BIGQUERY, '`test-project`.`TestDataset`.`dashed-table-name`'],
+    [DataStorageType.LEGACY_GOOGLE_BIGQUERY, '`test-project`.`TestDataset`.`dashed-table-name`'],
+    [DataStorageType.AWS_ATHENA, '"test-project"."TestDataset"."dashed-table-name"'],
+    [DataStorageType.AWS_REDSHIFT, '"test-project"."TestDataset"."dashed-table-name"'],
+    [DataStorageType.DATABRICKS, '`test-project`.`TestDataset`.`dashed-table-name`'],
+    [DataStorageType.SNOWFLAKE, 'test-project."TestDataset"."dashed-table-name"'],
```
kalnik-a-a: **Non-blocking:** This fixture encodes behavior that is not valid for Snowflake: `test-project` is left unquoted as the database segment, but Snowflake validation only allows `a-zA-Z0-9_$` for unquoted parts. This can give false confidence for the dashed-name case. Please use a Snowflake-specific valid fixture here, or update the expected output to match the actual Snowflake identifier contract after the escaping fix.

kalnik-a-a (changes requested):
I left two inline findings. One is blocking because it leaves the sample-data SQL path unsafe for Snowflake; the other is non-blocking test coverage cleanup.

apps/backend/src/data-marts/services/data-mart-sample-data.service.ts
laskevych: Fixed in 6b226fc8b. `escapeSnowflakeIdentifier` now fails closed: every part is always quoted with embedded double quotes doubled; inputs with more than three parts (or unparseable ones) get every part quoted instead of the raw passthrough; the database part stays unquoted only when it matches Snowflake's unquoted-identifier rule (`[A-Za-z_][A-Za-z0-9_$]*`), otherwise it is quoted too. `escapeSnowflakeSchema` got the same database-part hardening. Added coverage: util specs for >3-part inputs, embedded quotes, and the exact breakout payload, plus a service-level spec asserting the interpolated SQL for a Snowflake data mart with a malicious column (`a.b.c.d FROM sensitive_table --`) stays a single quoted identifier.

apps/backend/src/data-marts/data-storage-types/facades/identifier-escaper.facade.spec.ts
laskevych: Fixed in 6b226fc8b. With the fail-closed change above, a dashed database part is now quoted, so the fixture's expected output became `"test-project"."TestDataset"."dashed-table-name"` — matching the actual Snowflake contract. Also added a separate case verifying that a valid unquoted database part (`test_db`) stays unquoted.

kalnik-a-a (approved):
Looks good after the Snowflake escaping fix. The previously blocking sample-data SQL breakout path is now fail-closed, and the Snowflake fixture expectation was corrected as well.

PR #1321, opened 2026-06-12
feat(web): improve SwitchProjectMenu UX with search, truncation, and keyboard navigation

romandubovyi:
## Description

This PR improves the User Experience of the project switcher dropdown (`SwitchProjectMenu`) by introducing project list height limitations, proper text truncation, native tooltips, keyboard navigation, and search input layout alignment.

Screenshot 2026-06-12 at 17 03 43

### 1. `SwitchProjectMenu.tsx`
- **List Constraints**: Limited the project list width to `w-64` (256px) and maximum height to `max-h-[400px]` (fits ~10-12 items).
- **Text Truncation & Native Tooltips**: Long project names are now truncated using CSS ellipsis, and the full project name is displayed using the native HTML `title` attribute on hover (replacing the previous Radix tooltip component).
- **Sticky Search & Alignments**:
  - Aligned the search input with the dropdown items correctly.
  - The search input remains sticky at the top of the dropdown list on scroll, using a solid `bg-popover` background to prevent project list content from bleeding through.
- **Keyboard Navigation**:
  - Implemented `ArrowDown` and `ArrowUp` arrow key navigation to move the active focus through the filtered list of projects.
  - Pressing `Enter` triggers navigation to the currently active project.
  - Added automatic scrolling using `scrollIntoView({ block: 'nearest' })` to keep the keyboard-focused item in view.
  - The active index resets when the search query changes or when the project list changes.

### 2. `SwitchProjectMenu.test.tsx`
- **Keyboard Navigation Test**: Added a new test suite verifying that `ArrowDown`, `ArrowUp`, and `Enter` key presses work correctly.
- **Test Stability**: Used a stable `data-testid="project-list"` target instead of brittle DOM traversal to find list items.
- **State Cleanup**: Added `mockNavigate.mockClear()` in `beforeEach` to prevent navigation state pollution between tests.
- **ESLint & Prettier**: Fixed type castings and unused variables to fully comply with project-wide formatting and lint rules.

### DIFF
--- a/apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
+++ b/apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
@@ -69,7 +69,8 @@ function SwitchProjectMenuInner({
     itemRefs.current = [];
   }, [filteredProjects]);
 
-  // Scroll active item into view
+  // Scroll active item into view. `itemRefs` is a ref (stable), so omitting it from deps is intentional —
+  // the clear effect (above) always runs before this one when the list changes, so refs are never stale here.
   useEffect(() => {
     if (activeIndex >= 0 && itemRefs.current[activeIndex]) {
       itemRefs.current[activeIndex]?.scrollIntoView({ block: 'nearest' });
@@ -92,6 +93,7 @@ function SwitchProjectMenuInner({
       } else if (e.key === 'Enter') {
         const targetIndex = activeIndex >= 0 ? activeIndex : 0;
         const project = filteredProjects[targetIndex];
+        if (!project) return;
         void navigate(buildProjectPath(encodeURIComponent(project.id), '/'));
       }
     },
@@ -115,7 +117,7 @@ function SwitchProjectMenuInner({
       <DropdownMenuPortal>
         <DropdownMenuSubContent
           className='flex w-64 flex-col overflow-hidden p-0'
-          onKeyDown={showSearch ? undefined : handleKeyDown}
+          onKeyDown={handleKeyDown}
         >
           {showSearch && (
             <div
--- a/apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
+++ b/apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
@@ -1,5 +1,5 @@
-import { useEffect } from 'react';
-import { ArrowRightLeft, Loader2 } from 'lucide-react';
+import { useEffect, useState, useMemo, useRef, useCallback } from 'react';
+import { ArrowRightLeft, Check, Loader2 } from 'lucide-react';
 import {
   DropdownMenuSub,
   DropdownMenuSubTrigger,
@@ -8,8 +8,8 @@ import {
   DropdownMenuPortal,
   DropdownMenuSeparator,
 } from '@owox/ui/components/dropdown-menu';
-import { Check } from 'lucide-react';
-import { NavLink } from 'react-router-dom';
+import { NavLink, useNavigate } from 'react-router-dom';
+import { Input } from '@owox/ui/components/input';
 import { useAuth } from '../../../features/idp';
 import { useProjects } from '../../../features/idp/hooks/useProjects.ts';
 import { RequestStatus } from '../../../shared/types/request-status.ts';
@@ -30,6 +30,12 @@ function SwitchProjectMenuInner({
 }: SwitchProjectMenuProps) {
   const { projects, loadProjects, callState, error, isLoading } = useProjects();
   const { user } = useAuth();
+  const navigate = useNavigate();
+  const [searchQuery, setSearchQuery] = useState('');
+  const [activeIndex, setActiveIndex] = useState<number>(-1);
+  const listRef = useRef<HTMLDivElement>(null);
+  const itemRefs = useRef<(HTMLElement | null)[]>([]);
+
   const visibleProjects = excludeCurrentProject
     ? projects.filter(project => project.id !== user?.projectId)
     : projects;
@@ -41,6 +47,50 @@ function SwitchProjectMenuInner({
     }
   }, [isInitialLoad, loadProjects]);
 
+  const showSearch = projects.length > 10;
+
+  const filteredProjects = useMemo(() => {
+    if (!showSearch || !searchQuery.trim()) {
+      return visibleProjects;
+    }
+    const query = searchQuery.toLowerCase().trim();
+    return visibleProjects.filter(project => project.title.toLowerCase().includes(query));
+  }, [visibleProjects, searchQuery, showSearch]);
+
+  // Reset active index when filtered list changes
+  useEffect(() => {
+    setActiveIndex(-1);
+  }, [filteredProjects]);
+
+  // Scroll active item into view
+  useEffect(() => {
+    if (activeIndex >= 0 && itemRefs.current[activeIndex]) {
+      itemRefs.current[activeIndex]?.scrollIntoView({ block: 'nearest' });
+    }
+  }, [activeIndex]);
+
+  const handleKeyDown = useCallback(
+    (e: React.KeyboardEvent) => {
+      if (!filteredProjects.length) return;
+
+      if (e.key === 'ArrowDown') {
+        e.preventDefault();
+        e.stopPropagation();
+        setActiveIndex(prev => (prev + 1) % filteredProjects.length);
+      } else if (e.key === 'ArrowUp') {
+        e.preventDefault();
+        e.stopPropagation();
+        setActiveIndex(prev => (prev <= 0 ? filteredProjects.length - 1 : prev - 1));
+      } else if (e.key === 'Enter' && activeIndex >= 0) {
+        e.preventDefault();
+        e.stopPropagation();
+        const project = filteredProjects[activeIndex];
+        void navigate(buildProjectPath(encodeURIComponent(project.id), '/'));
+      }
+    },
+    [filteredProjects, activeIndex, navigate]
+  );
+
   return (
     <DropdownMenuSub>
       {showSeparator && <DropdownMenuSeparator />}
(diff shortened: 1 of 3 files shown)


apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
```diff
@@ -41,6 +47,50 @@ function SwitchProjectMenuInner({
...
+
+  // Reset active index when filtered list changes
+  useEffect(() => {
+    setActiveIndex(-1);
+  }, [filteredProjects]);
+
+  // Scroll active item into view
+  useEffect(() => {
+    if (activeIndex >= 0 && itemRefs.current[activeIndex]) {
+      itemRefs.current[activeIndex]?.scrollIntoView({ block: 'nearest' });
```
zapolsky, line 76: **[Bug — verified] The `scrollIntoView` auto-scroll feature is dead code — `itemRefs` is never populated.**

**Evidence chain:** `itemRefs` is declared (`SwitchProjectMenu.tsx:37`) and read here in the scroll effect (`:67-68`), but **no rendered item ever assigns it** — the `DropdownMenuItem`/`NavLink` map (`:175-198`) carries no `ref`, and the only `ref=` in the file is `listRef` on the list `div` (`:139`), which is itself never read. So `itemRefs.current` stays `[]`, `itemRefs.current[activeIndex]` is always `undefined`, the guard at `:67` is always false, and `scrollIntoView` never runs.

**Trigger:** open a switcher long enough to scroll, ArrowDown past the visible viewport — the highlighted row scrolls out of sight and is never brought back, the exact thing this feature (per the PR description) was added to do.

**Fix:** assign each item's node, e.g. on the `NavLink`: `ref={(el: HTMLAnchorElement | null) => { itemRefs.current[index] = el; }}`, and trim `itemRefs.current.length = filteredProjects.length` when the list changes so stale entries don't linger. (Also remove the now-unused `listRef`.)

apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
```diff
@@ -41,6 +47,50 @@ function SwitchProjectMenuInner({
...
+    if (!showSearch || !searchQuery.trim()) {
+      return visibleProjects;
+    }
+    const query = searchQuery.toLowerCase().trim();
+    return visibleProjects.filter(project => project.title.toLowerCase().includes(query));
+  }, [visibleProjects, searchQuery, showSearch]);
+
+  // Reset active index when filtered list changes
+  useEffect(() => {
+    setActiveIndex(-1);
```
zapolsky, line 64: **[Bug — verified] Keyboard navigation is broken in the `excludeCurrentProject` (restricted) menu — `activeIndex` is reset to -1 on every render.**

**Evidence chain:** `visibleProjects = excludeCurrentProject ? projects.filter(...) : projects` (`:39-41`) — in the restricted path `.filter()` returns a **new array reference every render**. `filteredProjects = useMemo(..., [visibleProjects, ...])` (`:52-58`) therefore recomputes to a **new reference every render**. This reset effect's dependency is `[filteredProjects]` (`:61-63`), so it fires on **every commit** → `setActiveIndex(-1)`. When the user presses ArrowDown, `setActiveIndex(0)` (`:79`) triggers a re-render whose reset effect immediately sets it back to -1; `Enter` requires `activeIndex >= 0` (`:84`) so it never navigates. (Verified empirically in the project's test harness: in the restricted path ArrowDown leaves 0 rows highlighted and `navigate` is never called; the non-restricted path — `visibleProjects = projects`, a stable reducer reference — works, which is why the existing test passes.) This path is live: `ProjectMenuContent.tsx:31` renders `<SwitchProjectMenu … excludeCurrentProject />`.

**Fix:** memoize `visibleProjects` (`useMemo(() => excludeCurrentProject ? projects.filter(p => p.id !== user?.projectId) : projects, [excludeCurrentProject, projects, user?.projectId])`), or key the reset effect on a stable value (`searchQuery`, `filteredProjects.length`) instead of the array identity.

zapolsky (commented):
🔴 **Blocked issues** — see inline comments. Two verified defects in the new keyboard-navigation feature. These should be addressed before merge, but this review does not formally block; a human owns the merge decision. Posted by automated OWOX review.

zapolsky:
## 🔴 Two blocking issues found — see the inline review

Deep multi-agent review against the worktree (`reviewer`, `logic-hunter`, fresh-eyes generalist, + a second wave since ≥2 blockers were confirmed). The two inline blockers are verified: the `scrollIntoView` auto-scroll is dead (`itemRefs` never populated), and keyboard navigation is broken in the `excludeCurrentProject` restricted menu (an unstable `visibleProjects` array re-fires the reset effect every render and pins `activeIndex` to -1 — confirmed empirically in the test harness). `tsc -b` 0 errors, ESLint clean, existing 8 tests pass — but the suite mocks Radix items as plain `<div>`s and only exercises the non-restricted ≤10-project path, so it masks both blockers.

**Root cause worth stepping back on:** the component layers a hand-rolled `activeIndex` highlight/keyboard system on top of Radix `DropdownMenuItem`, which already ships its own roving focus + typeahead + Enter-to-activate. The two systems never share state, which is what produces both blockers and several notes below. The cleanest direction is to pick **one** model — either drive selection through Radix (drop the custom `activeIndex`, style via the existing `focus:bg-accent`, keep only the search filter) or fully opt the items out of Radix focus — rather than patching the hybrid. Given the number of confirmed defects in the core feature, this likely needs a rework pass rather than incremental fixes.

- **Custom highlight vs Radix focus can diverge.** Because the items are real Radix menu items, the visually-highlighted row (`activeIndex` → `bg-accent`, `:179`) is not necessarily the row Radix has DOM-focused or that Radix's own Enter-to-`click()` would activate. Exact runtime behavior depends on focus state (the tests can't observe it — Radix is mocked), but it's the architectural root above. Unify the two systems.

- **Hover-then-Enter is a no-op.** `onPointerLeave` sets `activeIndex` back to -1 (`:183-185`) while `Enter` requires `activeIndex >= 0` (`:84`). Hover a row, move the mouse off, press Enter → nothing happens. Don't clear to -1 on leave (or separate hover state from keyboard state).

- **Question — Enter in the search box with nothing highlighted.** When `showSearch` is true and `activeIndex < 0`, the `Enter` branch is skipped without `preventDefault`/`stopPropagation` (`:84`, `:124-131`), so the event reaches Radix. Two of our passes disagreed on the outcome (a stray Radix typeahead/activation vs. an inert no-op since focus is in the input), so flagging it as a question: please confirm what Enter does after typing a filter but before arrowing. Safer to always `preventDefault`/`stopPropagation` the nav keys (and on empty results too — the `if (!filteredProjects.length) return;` at `:74` also returns before stopping the event).

- **`showSearch` counts the wrong population.** `showSearch = projects.length > 10` (`:50`) uses the unfiltered count, but the list renders `visibleProjects`; in `excludeCurrentProject` mode with 11 total / 10 switchable, the search box appears to filter a 10-item list (off-by-one vs. intent). Use `visibleProjects.length > 10`.

- **Stale search state across reopen.** `searchQuery` / `activeIndex` (`:34-35`) live in the always-mounted `SwitchProjectMenuInner`, so closing and reopening the submenu keeps the previous filter applied (reopen can show "no projects" with no obvious cause). Reset on `DropdownMenuSub`'s `onOpenChange`.

- **Accessibility.** The highlight is purely visual (`:179`) with no `role="listbox"`/`role="option"`, `aria-selected`, or `aria-activedescendant`, so screen-reader users get no announcement as they arrow through; the native `title` (`:190`) only helps mouse users read truncated names. Add listbox semantics + `aria-activedescendant`, and a non-`title` affordance for the truncated text.

- **Dead code:** `listRef` (`:36`, assigned `:139`) is never read.

- **Missing changeset.** This is a user-facing feature — add an `'owox': minor` changeset (`changeset-generator`). (Noted, not blocking.)

- **Tests mask the defects.** The only keyboard test (`SwitchProjectMenu.test.tsx:214`) uses 3 projects (no search) and mocks Radix items as bare `<div>`s, so it can't see the restricted-mode reset loop, the search-path Enter behavior, or the highlight-vs-focus divergence. Add a restricted-mode (`excludeCurrentProject`) ArrowDown→Enter test (fails today, proves blocker #2), a >10-project search-path Enter test, and assert the active highlight survives a re-render.

<sub>Automated OWOX review · worktree multi-agent passes + second wave · the 2 inline blockers were adversarially verified (dead-scroll = mechanical/grep; restricted-menu reset loop = skeptic-confirmed with an executed test). `tsc -b` 0 errors · ESLint clean.</sub>

apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
romandubovyi: Fixed in the latest commit. The ref callbacks have been assigned to the `DropdownMenuItem` elements:

```typescript
ref={el => {
  itemRefs.current[index] = el;
}}
```

Additionally, `itemRefs.current` is cleared when `filteredProjects` changes to ensure no stale references linger, and the unused `listRef` has been removed.

apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
romandubovyi: Fixed. Wrapped `visibleProjects` in a `useMemo` hook with proper dependencies (`[projects, excludeCurrentProject, user?.projectId]`). This ensures a stable reference and prevents the reset effect from re-running on every render.

romandubovyi:
Hi @zapolsky, thanks for the thorough review! All blockers and notes have been resolved in the latest commit:

1. **Fixed the two inline blockers:**
   - Memoized `visibleProjects` inside a `useMemo` block to ensure its reference is stable and `activeIndex` is not reset on every render in `excludeCurrentProject` mode.
   - Populated `itemRefs` inside the map loop and added a cleanup effect to clear stale refs when the list updates.
   - Removed the unused `listRef`.
2. **Addressed UX & navigation issues:**
   - Unified hover and keyboard highlights by removing the `onPointerLeave` handler, ensuring that moving the mouse off does not break `Enter` navigation.
   - Pressing `Enter` in the search box when nothing is explicitly highlighted now defaults to selecting/navigating to the first filtered project (`filteredProjects[0]`), while preventing the event from bubbling to Radix.
   - Corrected `showSearch` to depend on `visibleProjects.length > 10` instead of the unfiltered count.
   - Reset the search state (`searchQuery` and `activeIndex`) when the sub-menu is closed by hooking into `DropdownMenuSub`'s `onOpenChange`.
3. **Accessibility (a11y) improvements:**
   - Added standard ARIA roles and attributes (`role="combobox"`, `role="listbox"`, `role="option"`, `aria-autocomplete="list"`, `aria-activedescendant`, `aria-selected`).
4. **Added Unit Tests:**
   - Added test coverage in `SwitchProjectMenu.test.tsx` verifying keyboard navigation under `excludeCurrentProject` mode (restricted menu), `Enter` key execution on empty highlighted search state, and highlight persistence on unrelated component re-renders.
5. **Changeset:**
   - Added a minor `'owox': minor` changeset file.

apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
zapolsky: ✅ **Verified fixed.** The ref callback now populates `itemRefs` on each rendered item (`SwitchProjectMenu.tsx:201-203`, `ref={el => { itemRefs.current[index] = el; }}`), and via `DropdownMenuItem` `asChild` the ref is forwarded through Radix's Slot onto the `NavLink` `<a>` — the actual scrollable child of the `max-h-[400px] overflow-y-auto` `#project-list` container. The scroll effect's read is safe (truthiness guard + optional chaining at `:74-75`), and on a filter change `activeIndex` resets to `-1` so the effect short-circuits before reading the just-cleared refs — no stale-null deref. Auto-scroll now executes. Confirmed live against real Radix by the fresh-eyes pass.

apps/web/src/components/AppSidebar/ProjectMenu/SwitchProjectMenu.tsx
zapolsky: ✅ **Verified fixed.** `visibleProjects` is now memoized (`SwitchProjectMenu.tsx:38-42`) with deps `[projects, excludeCurrentProject, user?.projectId]` — complete and correct (the filter body references exactly those three). Its stability propagates to the `filteredProjects` memo, so the `activeIndex`-reset effect (dep `[filteredProjects]`) no longer re-fires every render in `excludeCurrentProject` mode. `activeIndex` survives renders and keyboard nav works in the restricted menu. Locked by the new regression test "retains active highlight index when unrelated re-render occurs" (`SwitchProjectMenu.test.tsx:310-339`).

zapolsky:
## ✅ Both previously-flagged blockers are fixed — no new blocking issues

Incremental re-review against the worktree (`reviewer`, `logic-hunter`, fresh-eyes generalist). The two blockers from the previous review are **verified fixed** (confirmed in their inline threads):

1. **`dead-scrollintoview-itemrefs`** — `itemRefs` is now populated via a ref callback forwarded (through `asChild`) to the scrollable `NavLink`; auto-scroll executes. ✅
2. **`restricted-keyboard-nav-reset-loop`** — `visibleProjects` is memoized with correct/complete deps, so `activeIndex` survives renders in `excludeCurrentProject` mode. ✅

- **Question — search-mode arrow nav depends on the search input keeping focus.** `SwitchProjectMenu.tsx:118` sets `onKeyDown={showSearch ? undefined : handleKeyDown}`, so when the search box is shown (>10 projects) the only keydown wiring is on the `Input` (`:138-145`). Radix `SubContent.onOpenAutoFocus` can land focus on the content rather than the input on keyboard-open, and if focus ever moves into the list the arrows stop working until the user clicks back into the search box. Consider keeping `handleKeyDown` on the container in search mode too (it early-returns on non-nav keys, and the input already `stopPropagation`s character keys, so no double-handling).
- **`SwitchProjectMenu.tsx:73-77`** — the scroll effect lists only `[activeIndex]` in its deps while reading `filteredProjects`-derived refs. Works today because the reset-to-`-1` effect drives re-runs, but it's a latent fragility and a likely `react-hooks/exhaustive-deps` flag — worth a clarifying comment so a future edit doesn't "fix" the deps and break the sequencing.
- **`SwitchProjectMenu.tsx:92-95`** — the Enter handler reads `filteredProjects[targetIndex]` then `project.id` with only an emptiness guard, not a bounds guard. Structurally safe today (the index is reset whenever the list changes), but a one-line `if (!project) return;` makes it robust against a future "sticky highlight" change.
- **Test coverage** — the unit tests mock `@owox/ui`'s DropdownMenu entirely (dropping `asChild`/`ref`), so auto-scroll and the Radix `preventDefault`/roving-focus suppression are not exercised by the suite. The fresh-eyes pass verified both work against real Radix; one integration test rendering the real DropdownMenu would lock the interaction in.

<sub>Automated OWOX review · worktree multi-agent passes · prior blockers re-verified fixed in their inline threads.</sub>

zapolsky (approved):

PR #1322, opened 2026-06-12
feat(criteo): add report endpoints, configurable currency, API 2026-01

prokopetsv:
# Problem

The Criteo connector exposed only a single `statistics` endpoint, pinned to API version `2025-07`, with the reporting currency hardcoded to USD. Customers could not break performance down by publisher placement or content category, could not pull attributed transaction-level data, and always received cost/revenue figures in USD regardless of their account currency. The response parser also failed silently: any unrecognized response shape was treated as "no data", so a misconfigured stream produced a successful run with empty tables instead of an error.

# Solution

Three new report streams were added on top of the verified Criteo `2026-01` OpenAPI spec, with the API version centralized in a single constant. During live testing the documented endpoints turned out to diverge from the spec (the dedicated categories endpoint returns 404, response envelopes differ per endpoint, and field casing is inconsistent), so the integration was hardened structurally rather than per-case:

- `placement_categories` reuses the spec-confirmed `/placements/report` endpoint with category dimensions, since `/categories/report` does not exist in `2026-01`.
- Response parsing recognizes all observed envelope shapes (bare array, JSON:API `data[].attributes.rows`, per-row `data[].attributes`, legacy `Rows`) and logs a PII-free shape description instead of silently returning zero rows when the shape is unknown.
- Response keys are matched case-insensitively against the schema (with optional `apiName` for request-side mapping), so unique keys cannot be silently dropped by casing drift — which would otherwise corrupt storage dedup.
- Unique-key and unknown-field validation runs before every fetch, failing fast with a clear error.
- The new `Currency` config validates format as a hard error but treats the known-currency list as advisory (warning only), since the list drifts and the API is the authority.
- Storage config is now built per node without mutating the shared connector config, scoping fields to the current stream.

# Changes

- Added `placements`, `placement_categories`, and `transactions` streams with field schemas, unique keys, and default fields (`Source.js`, `CriteoAdsFieldsSchema.js`, `placementFields.js`, `placementCategoryFields.js`, `transactionFields.js`)
- Bumped the Criteo API version from `2025-07` to `2026-01` via a single `CRITEO_API_VERSION` constant
- Added 10 dimensions to the statistics stream: Device, Os, Channel/ChannelId, Category/CategoryId, Coupon/CouponId, MarketingObjective/MarketingObjectiveId (`adStatisticsFields.js`)
- Added a `Currency` config parameter (default USD) with ISO 4217 format validation and an advisory supported-currency warning, replacing the hardcoded USD
- Added multi-shape response-row extraction with a logged, value-free shape warning for unrecognized responses
- Added case-insensitive response-key normalization and request-side `apiName` mapping to protect unique keys from API casing inconsistencies
- Added pre-fetch validation of unknown fields and missing unique keys for all streams
- Generalized `_makeApiRequest` to accept the endpoint URL and removed its redundant status check (handled by `urlFetchWithRetry`)
- Scoped storage config per node with unique keys always included, without mutating the shared connector config (`Connector.js`)
- Hardened `parseFields`/`parseAdvertiserIds` against malformed separators and empty entries (`Helper.js`)
- Documented all four endpoints and the unavailability of domain-level breakdown in `2026-01` (`README.md`, `GETTING_STARTED.md`)
- Added changeset `criteo-new-endpoints-and-currency`

### DIFF
--- a/.changeset/criteo-new-endpoints-and-currency.md
+++ b/.changeset/criteo-new-endpoints-and-currency.md
@@ -1,0 +1,11 @@
+---
+'owox': minor
+---
+
+# Criteo placement, category, and transaction reporting
+
+The Criteo connector now supports three new endpoints — placements, placement categories,
+and transactions — alongside the existing campaign statistics, plus extra statistics
+dimensions such as device, OS, and channel. You can also choose the reporting currency
+instead of always getting USD. Previously only campaign-level statistics were available;
+now you can break performance down by where ads ran and pull individual attributed orders.
--- a/packages/connectors/src/Sources/CriteoAds/Connector.js
+++ b/packages/connectors/src/Sources/CriteoAds/Connector.js
@@ -78,7 +78,7 @@ var CriteoAdsConnector = class CriteoAdsConnector extends AbstractConnector {
 
       if (data.length || this.config.CreateEmptyTables?.value) {
         const preparedData = data.length ? this.addMissingFieldsToData(data, fields) : data;
-        const storage = await this.getStorageByNode(nodeName);
+        const storage = await this.getStorageByNode(nodeName, fields);
         await storage.saveData(preparedData);
       }
 
@@ -92,9 +92,10 @@ var CriteoAdsConnector = class CriteoAdsConnector extends AbstractConnector {
   /**
    * Get storage instance for a node
    * @param {string} nodeName - Name of the node
+   * @param {Array<string>} fields - Fields selected for this node
    * @returns {Object} Storage instance
    */
-  async getStorageByNode(nodeName) {
+  async getStorageByNode(nodeName, fields = []) {
     if (!("storages" in this)) {
       this.storages = {};
     }
@@ -105,12 +106,10 @@ var CriteoAdsConnector = class CriteoAdsConnector extends AbstractConnector {
       }
 
       const uniqueFields = this.source.fieldsSchema[nodeName].uniqueKeys;
+      const storageConfig = this._buildStorageConfig({ nodeName, fields, uniqueFields });
 
       this.storages[nodeName] = new globalThis[this.storageName](
-        this.config.mergeParameters({
-          DestinationSheetName: { value: this.source.fieldsSchema[nodeName].destinationName },
-          DestinationTableName: { value: this.getDestinationName(nodeName, this.config, this.source.fieldsSchema[nodeName].destinationName) },
-        }),
+        storageConfig,
         uniqueFields,
         this.source.fieldsSchema[nodeName].fields,
         `${this.source.fieldsSchema[nodeName].description} ${this.source.fieldsSchema[nodeName].documentation}`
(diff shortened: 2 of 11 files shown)


zapolsky:

Deep multi-agent review against the worktree (`reviewer`, `logic-hunter`, `architect`, `tester`, fresh-eyes generalist). One candidate blocker was raised (the `_extractRows` "garbage row" path) and went through the adversarial gate — it was **refuted/unproven** (see the question below). The per-node storage-config isolation, the case-insensitive normalization for the *current* schema, the currency validation, and the removed `_makeApiRequest` status check were all traced end-to-end and verified correct.

- **Latent silent-drop trap — `_buildKeyResolver` ignores `apiName`.** Requests rename fields via `apiName` (`Source.js:328`, e.g. `adsetId`→`adSetId`), but the response resolver is built only from the field name + `fieldConfig.aliases` (`Source.js:451-460`) — `apiName` is never registered, and **no schema declares `aliases`**, so that branch is dead code. It works today *only* because every current `apiName` differs from its canonical name by case alone, which `.toLowerCase()` folds (`Source.js:434`). The first field whose `apiName` differs by more than case (a real rename / snake_case key) would be silently dropped from the response; if it were a unique key, dedup would break silently. Suggest registering `apiName` in the resolver (`if (fieldConfig.apiName) resolver[fieldConfig.apiName.toLowerCase()] = fieldName;`) and removing the unused `aliases` path.

- **Question (refuted/unproven blocker) — `_extractRows` is shape-based, not stream-aware.** For placements / placement_categories (which expect `{data:[{attributes:{rows:[…]}}]}`), an entry whose `attributes` is an object *without* a `rows` array falls into the per-row branch (`Source.js:488-490`) and is treated as a single row. We traced the consequence: it does **not** hard-throw (the `_validateUniqueKeys` guard at `Source.js:138-152` plus `addMissingFieldsToData` ensure the unique-key columns are present-as-null, so `getUniqueKeyByRecordFields` at `AbstractStorage.js:62` does not throw) — it would instead persist a spurious all-null-key record. **However, the trigger is unproven**: there is no evidence the `2026-01` placements endpoint ever returns an `attributes`-without-`rows` entry (the docstring at `Source.js:465-467` and your live testing say placements use `attributes.rows`, and per-entry `attributes` is the *transactions* shape). Not a blocker — but cheap defensive hardening: in the placements/categories path require `attributes.rows` and route a bare-`attributes` entry into the unrecognized-shape branch rather than treating it as a row.

- **No automated coverage.** The entire `packages/connectors` package has no test harness (no `*.test.js`, no test runner config, no `test` script, not linted in CI — `eslint.config.js` ignores `src/**/*`). The new response parser (`_extractRows`) and key normalization (`_filterBySchema`/`_buildKeyResolver`) are the highest-risk untested logic because their failure modes are silent (wrong/missing rows, broken dedup) rather than exceptions. Worth a minimal fixture-based test for the 4 envelope shapes + the unknown-shape path, even outside CI.

- **Question — Currency now applies to the `statistics` stream too.** It was hardcoded `"USD"` and is now `_getCurrency()` (`Source.js:303`). Benign for existing users (default + validate fill `"USD"`), but a user who sets a non-USD currency will now also get `statistics` in that currency — confirm that's intended.

- **Maintainability (optional).** The per-node `Fields` scoping in `_buildStorageConfig` (`Connector.js:134-152`) is a genuine fix for a bug that also exists in sibling multi-stream connectors (they pass the full multi-node `Fields` to every per-node storage); and the `parseFields` hardening in `Helper.js` is a fork of the still-brittle `Core/Utils/ConnectorUtils.parseFields`. Both would be better lifted into `Core` so siblings benefit and Criteo doesn't silently diverge. Also: the four near-identical `_fetch*` methods could collapse into one `_fetchReport({nodeName, endpoint, buildBody})`.

<sub>Automated OWOX review · worktree multi-agent passes · the one candidate blocker was adversarially verified and refuted/unproven. `tsc`: n/a (plain JS) · `eslint`: n/a (config ignores `src/**`, connectors not linted in CI).</sub>

zapolsky (approved):

PR #1323, opened 2026-06-12
fix(web): surface validation errors hidden in collapsed form sections

romandubovyi:
## Problem

In the Storage, Destination, and Report forms, required fields with validation
errors could sit inside a collapsed accordion section. Clicking **Save** then
failed silently: react-hook-form kept the errors on unmounted fields, no message
was rendered, and the built-in focus on the invalid field could not work. Two
adjacent issues made it worse:

- **Copy credentials from another storage** always failed validation for
  non-Google storage types (Athena, Redshift, Snowflake, Databricks), because
  the hidden credential fields were still required by the schema — with nowhere
  to display the errors.
- The **Create & Run report** button was disabled while the form was invalid,
  leaving the user with a dead button and no hint about what was missing.

### `packages/ui` — `FormSection`
- Sections track their fields automatically (every `FormField` rendered inside
  registers itself via context); sections that may start collapsed can declare
  them explicitly with the new optional `fields` prop.
- On a failed submit, a collapsed section containing errors auto-opens once per
  submit attempt — the user can re-collapse it, and their persisted (localStorage)
  preference is never overwritten.
- A red `CircleAlert` indicator (optically centered with the uppercase title)
  appears in the section header while it contains errors, with an `sr-only`
  description for screen readers.

### `apps/web` — forms
- New `focusFirstInvalidField` helper is wired as the `onInvalid` handler in all
  five forms (Storage, Destination, Google Sheets / Email / Looker Studio
  reports): it polls a few frames until the reopened section mounts the invalid
  field, then scrolls to it and focuses it.
- New `createDataStorageFormResolver` bypasses credential validation while a
  copy source is selected (the credentials are stripped from the payload and
  copied server-side via `sourceStorageId`), replacing the previous
  Google-only placeholder hack — Copy From now works for every storage type.
- Report create buttons (`ReportFormActions` and the Looker Studio form) no
  longer disable on `!isValid` in CREATE mode, so submitting reveals the
  validation errors. EDIT mode behavior (disabled until dirty) is unchanged.
- The Report Columns section now participates in validation: it declares its
  fields, the picker is wrapped in `FormField`/`FormControl` (aria-invalid +
  scroll target), and the error clears immediately once a column is selected
  again.

## Testing
- Unit tests cover the new behavior: `FormSection` auto-open, the focus helper,
  the storage form resolver for all storage types, and `ReportFormActions`
  disabled states.
- `tsc -b` and ESLint clean.
- Verified manually end-to-end in the running app: section auto-open + focus on
  Storage save, Copy From on a new Redshift storage (success toast), report
  creation with collapsed General section, and the Report Columns error flow.

### DIFF
--- a/apps/web/src/features/data-marts/reports/edit/components/shared/ReportFormActions.tsx
+++ b/apps/web/src/features/data-marts/reports/edit/components/shared/ReportFormActions.tsx
@@ -9,13 +9,13 @@ import {
 } from '@owox/ui/components/dropdown-menu';
 import { ChevronDown } from 'lucide-react';
 import { ReportFormMode } from '../../../shared';
+import { useRef, useEffect } from 'react';
 import type { RefObject } from 'react';
 
 export interface ReportFormActionsProps {
   mode: ReportFormMode;
   isSubmitting: boolean;
   isDirty: boolean;
-  isValid: boolean;
   triggersDirty: boolean;
   ownersDirty?: boolean;
   runAfterSaveRef: RefObject<boolean>;
@@ -27,16 +27,27 @@ export const ReportFormActions = ({
   mode,
   isSubmitting,
   isDirty,
-  isValid,
   triggersDirty,
   ownersDirty = false,
   runAfterSaveRef,
   onSubmit,
   onCancel,
 }: ReportFormActionsProps) => {
+  // Guards against double-submit during the window between the first click and
+  // the parent updating isSubmitting to true (async resolver latency).
+  const submitPendingRef = useRef(false);
+  useEffect(() => {
+    if (!isSubmitting) {
+      submitPendingRef.current = false;
+    }
+  }, [isSubmitting]);
+
+  // In CREATE mode the button stays clickable even while the form is invalid:
+  // submitting runs validation, which opens collapsed sections with errors and
+  // focuses the first invalid field. Disabling on !isValid would leave the user
+  // with no hint about what is missing.
   const disabledPrimary =
-    isSubmitting ||
-    (mode === ReportFormMode.CREATE ? !isValid : !(isDirty || triggersDirty || ownersDirty));
+    isSubmitting || (mode === ReportFormMode.EDIT && !(isDirty || triggersDirty || ownersDirty));
 
   const primaryLabel =
     mode === ReportFormMode.CREATE ? 'Create & Run report' : 'Save changes to report';
@@ -56,7 +67,12 @@ export const ReportFormActions = ({
           type='submit'
           disabled={disabledPrimary}
           className='flex-1'
-          onClick={() => {
+          onClick={e => {
+            if (submitPendingRef.current) {
+              e.preventDefault();
+              return;
+            }
+            submitPendingRef.current = true;
             // For CREATE: run after save. For EDIT: don't run
             runAfterSaveRef.current = mode === ReportFormMode.CREATE;
           }}
@@ -82,6 +98,8 @@ export const ReportFormActions = ({
           <DropdownMenuContent align='end' side='top'>
             <DropdownMenuItem
               onClick={() => {
+                if (submitPendingRef.current || isSubmitting) return;
+                submitPendingRef.current = true;
                 // For CREATE: don't run. For EDIT: run after save
                 runAfterSaveRef.current = mode === ReportFormMode.EDIT;
                 onSubmit();
--- a/.changeset/report-form-invalid-submit-guard.md
+++ b/.changeset/report-form-invalid-submit-guard.md
@@ -1,0 +1,7 @@
+---
+"owox": minor
+---
+
+# Surface validation errors hidden in collapsed form sections
+
+Improved form validation UX so submitting a form opens collapsed sections that contain invalid fields and focuses the first validation error.
--- a/apps/web/e2e/selectors/testids.ts
+++ b/apps/web/e2e/selectors/testids.ts
@@ -54,6 +54,9 @@ export const TESTIDS = {
   notifToggle: 'notifToggle',
   notifEditSheet: 'notifEditSheet',
 
+  // Form validation
+  formSectionErrorIndicator: 'form-section-error-indicator',
+
   // FloatingPopover domain
   floatingPopoverClose: 'floatingPopoverClose',
 
--- a/apps/web/e2e/specs/storage.spec.ts
+++ b/apps/web/e2e/specs/storage.spec.ts
@@ -232,6 +232,48 @@ describeIfCredentials(['AWS_ACCESS_KEY_ID', 'AWS_REGION'], 'Athena Config Save',
 });
 
 // ---------------------------------------------------------------------------
+// STOR-08: Collapsed section with errors auto-opens on failed submit
+// ---------------------------------------------------------------------------
+
+test.describe('Storage Form — Collapsed Section Validation', () => {
+  test('auto-opens collapsed General section and shows error indicator on failed submit', async ({
+    page,
+    apiHelpers,
+  }) => {
+    await apiHelpers.createStorage('GOOGLE_BIGQUERY');
+    await page.goto('/ui/0/data-storages');
+    await expect(page.getByTestId(TESTIDS.storageListPage)).toBeVisible();
+
+    await page.getByRole('button', { name: 'Open menu' }).first().click();
+    await page.getByRole('menuitem', { name: 'Edit' }).click();
+    await expect(page.getByTestId(TESTIDS.storageEditForm)).toBeVisible();
+
+    const form = page.getByTestId(TESTIDS.storageEditForm);
+    const titleInput = form.getByLabel('Title');
+
+    // Clear title to create a validation error inside "General"
+    await titleInput.click();
+    await titleInput.fill('');
+
+    // Collapse the "General" section
+    await form.getByRole('button', { name: /general/i }).click();
+    await expect(titleInput).not.toBeVisible();
+
+    // Submit with a collapsed section that contains errors
+    await page.getByRole('button', { name: 'Save' }).click();
+
+    // Section must auto-open and expose the invalid field
+    await expect(titleInput).toBeVisible({ timeout: 3000 });
+
+    // Error indicator must appear in at least one section header (General, and possibly others)
+    await expect(form.getByTestId(TESTIDS.formSectionErrorIndicator).first()).toBeVisible();
+
+    // Inline validation message must be visible
+    await expect(form.getByText('Title is required')).toBeVisible();
+  });
+});
+
+// ---------------------------------------------------------------------------
 // STOR-06: Delete storage with confirmation dialog
 // ---------------------------------------------------------------------------
 
--- a/apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/DataDestinationForm.tsx
+++ b/apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/DataDestinationForm.tsx
@@ -39,7 +39,7 @@ import {
   AccordionTrigger,
 } from '@owox/ui/components/accordion';
 
-import { createFormPayload } from '../../../../../utils';
+import { createFormPayload, focusFirstInvalidField } from '../../../../../utils';
 import { COPY_SOURCE_CREDENTIAL_PLACEHOLDER } from '../../../../../shared/utils/credential-identity-utils';
 import {
   DataDestinationType,
@@ -220,7 +220,7 @@ export function DataDestinationForm({
     <Form {...form}>
       <AppForm
         onSubmit={e => {
-          void form.handleSubmit(handleSubmit)(e);
+          void form.handleSubmit(handleSubmit, focusFirstInvalidField)(e);
         }}
       >
         <FormLayout>
--- a/apps/web/src/features/data-marts/reports/edit/components/EmailReportEditForm/EmailReportEditForm.tsx
+++ b/apps/web/src/features/data-marts/reports/edit/components/EmailReportEditForm/EmailReportEditForm.tsx
@@ -1013,7 +1013,7 @@ export const EmailReportEditForm = forwardRef<HTMLFormElement, EmailReportEditFo
             </FormLayout>
             <ReportFormActions
               mode={mode}
-              isSubmitting={isSubmitting}
+              isSubmitting={isSubmitting || form.formState.isSubmitting}
               isDirty={isDirty}
               triggersDirty={triggersDirty}
               ownersDirty={ownersDirty}
--- a/apps/web/src/features/data-marts/reports/edit/components/EmailReportEditForm/EmailReportEditForm.tsx
+++ b/apps/web/src/features/data-marts/reports/edit/components/EmailReportEditForm/EmailReportEditForm.tsx
@@ -1,5 +1,6 @@
 import { forwardRef, useEffect, useMemo, useState, useRef } from 'react';
 import { useOwnerState } from '../../../../../../shared/hooks/useOwnerState';
+import { focusFirstInvalidField } from '../../../../../../utils';
 import { UserReference } from '../../../../../../shared/components/UserReference/UserReference';
 import { useUser } from '../../../../../idp/hooks/useAuthState';
 import { useNavigate, Link } from 'react-router-dom';
@@ -220,7 +221,6 @@ export const EmailReportEditForm = forwardRef<HTMLFormElement, EmailReportEditFo
 
     const {
       isDirty,
-      isValid,
       reset,
       form,
       isSubmitting,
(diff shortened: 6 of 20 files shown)


zapolsky:

Deep multi-agent review against the worktree (`reviewer`, `logic-hunter`, `architect`, `tester`, fresh-eyes generalist). The core mechanism — context-based field tracking, auto-open-once-per-submit (without clobbering the persisted collapse preference), the error indicator, the bounded focus poll, and the per-type copy-credentials resolver — was traced end-to-end and is **correct**.

**On CI/tests:** an initial run of the new suites showed 6 failures, but I confirmed this is a *review-harness artifact*, not a defect: my review worktree symlinks `node_modules`, so `@owox/ui/components/form` resolved to the **base** `packages/ui` (without this PR's `FormSection` changes), and `tsc -b` raised the same false `Property 'fields' does not exist on FormSectionProps` for the same reason. Re-running with `@owox/ui` aliased to the PR's `packages/ui/src`: **FormSection + form-utils suites pass 9/9, and `tsc` reports 0 errors.** CI installs the workspace at HEAD, so it will be green. ESLint is clean on all changed files. Nothing to action here — flagging only so the note isn't mistaken for a real failure.

- **Make the copy-credential placeholder switch exhaustive.** `buildCopySourcePlaceholderCredentials` (`data-storage-form-resolver.ts`) hard-codes the credential shape per `DataStorageType` and ends in `default: return undefined`. It re-derives what the credential sub-schemas in `data-storage.schema.ts` already encode, so a future storage type (or a changed credential schema) silently breaks "Copy From" for that type **with no compile error**. Add a `never` exhaustiveness check on the switch so adding a type forces a compile failure here.

- **Placeholder secrets live in the resolver's returned values; safety hinges on one `delete`.** When a copy source is selected the resolver injects placeholder credentials (e.g. `accessKeyId: "00000000-…"`) into the values handed to `onValid`; the only thing preventing them from being persisted is `if (… || selectedSource) delete payload.credentials` (`DataStorageForm.tsx:198-200`). Correct today (the strip condition and `selectedSource` are always set together), but fragile — any new submit path or a tweak to that condition would persist placeholder secrets. Safer to strip `credentials` in the resolver's returned `values` so the placeholder never exists in form state, and/or add a regression test asserting the submitted payload has no `credentials` when copying.

- **`focusFirstInvalidField` falls back to `document` scope on the eventless submit path.** The split-button/dropdown path calls `form.handleSubmit(onValid, onInvalid)()` with no event, so `event?.target` is undefined and the query scope becomes the whole `document` (`form-utils.ts:1300`). It then focuses/scrolls to the first `[aria-invalid="true"]` anywhere on the page. Low impact in practice (single-open Sheets), but worth scoping to the form element explicitly on that path.

- **Redshift `workgroupName` / `clusterIdentifier` errors aren't reached by the new focus/indicator machinery.** Those inputs are `form.register(...)`-ed inside a `FormField name="config.connectionType"` (`RedshiftFields.tsx`), so the zod error key `config.workgroupName` has no `FormControl` emitting `aria-invalid` and no inline `FormMessage`. `focusFirstInvalidField` can't find/scroll to it and the message doesn't render. Pre-existing wiring (Connection Settings is always open, so the field is at least visible), but the PR advertises "focus the first invalid field" and it silently no-ops for this one. Giving those two inputs their own `FormField`/`FormControl` would fix it.

- **CREATE-mode double-submit window widened.** Removing `!isValid` leaves the primary button gated only by the *local* `isSubmitting`, which flips to `true` only after the async zod resolver passes inside `onSubmit`. Two fast clicks can fire two `createReport` calls. Consider also gating on `form.formState.isSubmitting` (RHF sets it before validation) or a synchronous submitting ref.

- **Minor / consistency:** (a) a section that is `defaultOpen={false}` *and* contains validated fields but omits `fields` will silently swallow errors again — no active offenders today, but a dev-mode `console.warn` would turn that footgun into a visible warning; (b) `aria-invalid` for the Report Columns field lands on a non-focusable `<div>`, so `.focus()` is a no-op (scroll still works) — `tabIndex={-1}` on the picker container would let focus land; (c) `DataDestinationForm` still uses the old Google-only `credentialId` placeholder rather than the new resolver — benign (only Google-based destinations are copyable) but inconsistent with the 5-form story; (d) consider moving `focusFirstInvalidField` into `packages/ui` next to the form primitives, since its retry loop exists specifically because of the lib's auto-open behavior.

<sub>Automated OWOX review · worktree multi-agent passes · no blockers. Verified green: FormSection + form-utils suites (9/9) and `tsc` (0 errors) against the PR's `packages/ui` · ESLint clean. The 6 "failures" / 5 `tsc` errors seen initially are a node_modules-symlink resolution artifact of the review harness, not CI.</sub>

romandubovyi:
> In response to @zapolsky's [review](https://github.com/OWOX/owox-data-marts/pull/1323#issuecomment-4717017106):

**Addressed:**

- **Exhaustive switch** — added `case undefined:` and a `default: { const _exhaustive: never = type; }` block; a new `DataStorageType` without a matching case will now produce a compile error.

- **Placeholder credentials in resolver values** — the resolver now strips `credentials` from the returned `values` when copying, so placeholders never reach `onValid`. Added a regression test asserting `result.values` has no `credentials` key when a copy source is selected.

- **Redshift `workgroupName` / `clusterIdentifier`** — each input now has its own `FormField`/`FormControl`, so `aria-invalid` lands on the `<input>` element and `focusFirstInvalidField` can scroll/focus those fields on validation failure. The tabs still appear inline next to the label via `form.setValue` + `form.watch`.

- **CREATE-mode double-submit** — added `submitPendingRef` in `ReportFormActions`; the dropdown item is blocked if a submit is already pending (before `isSubmitting` prop updates), and the primary button calls `e.preventDefault()` on a second click. Added a regression test.

- **`aria-invalid` on non-focusable div (Report Columns)** — added `tabIndex={-1}` to the picker container in all three report forms (Google Sheets, Email, Looker Studio) so `.focus()` lands correctly.

apps/web/src/features/data-marts/reports/edit/components/shared/ReportFormActions.tsx
```diff
@@ -56,7 +67,12 @@ export const ReportFormActions = ({
           type='submit'
           disabled={disabledPrimary}
           className='flex-1'
-          onClick={() => {
+          onClick={e => {
+            if (submitPendingRef.current) {
+              e.preventDefault();
+              return;
+            }
+            submitPendingRef.current = true;
```
zapolsky, line 75: **[Bug — verified] The primary submit button locks permanently after the first *invalid* click on the Email & Google Sheets report forms (CREATE and EDIT).**

This PR removes the `!isValid` disable so the user *can* click Submit on an incomplete form to surface the errors — but after that first failed-validation click the button (and the dropdown) become un-clickable, defeating the fix. The only escape is the Enter key.

**Evidence chain (production wiring, verified hop-by-hop incl. an adversarial skeptic pass):**
1. `ReportFormActions.tsx:75` — the primary `type='submit'` button's `onClick` sets `submitPendingRef.current = true`, and `:71-74` `e.preventDefault(); return;` on the next click while the ref is `true`. The dropdown item is gated by the same ref (`:101`).
2. `ReportFormActions.tsx:39-43` — the **only** code that resets the ref is `useEffect(() => { if (!isSubmitting) submitPendingRef.current = false; }, [isSubmitting])`; it runs only when the `isSubmitting` **prop** changes value.
3. The `isSubmitting` prop is the hook's **manual** `useState` — `EmailReportEditForm.tsx:226` destructures it from `useEmailReportForm` and passes `isSubmitting={isSubmitting}` at `:1016` (Google Sheets: `GoogleSheetsReportEditForm.tsx:160,530`). It is **not** RHF's `formState.isSubmitting`.
4. That manual flag is set `true` **only** inside the valid `onSubmit` callback — `useEmailReportForm.ts:149` `setIsSubmitting(true)` (Google Sheets `useGoogleSheetsReportForm.ts:90`).
5. The form wires `form.handleSubmit(handleFormSubmit, focusFirstInvalidField)` (`EmailReportEditForm.tsx:394,1021`; Google Sheets `:245,535`). On an **invalid** submit RHF calls only the onInvalid handler (`focusFirstInvalidField`, which touches no React state — `form-utils.ts:20-36`), never `handleFormSubmit` → `setIsSubmitting(true)` never runs → the manual `isSubmitting` stays `false`.
6. So after the first invalid click: ref is `true`, the manual `isSubmitting` never changed (`false`→`false`), the `[isSubmitting]` effect never re-runs, the ref stays `true` forever → every subsequent primary-button and dropdown click is swallowed.

**Trigger:** Create an Email or Google Sheets report, leave a required field empty, click **Create & Run report** (mouse) → errors appear (works once) → fix the field → click again → nothing happens. (Looker Studio is unaffected — it uses a plain inline `<Button type='submit'>` with no ref guard, `LookerStudioReportEditForm.tsx:320-337`.)

**Why existing tests miss it:** `ReportFormActions.test.tsx:45-62` only drives the *dropdown* with a no-op `onSubmit`, so `isSubmitting` never cycles and the latch-release path is never exercised — it asserts the latch *engages*, never that it *releases*.

**Fix:** reset `submitPendingRef` on the onInvalid path too (e.g. wrap `focusFirstInvalidField` to also clear it), or key the guard off RHF's `formState.isSubmitting` (which toggles `true→false` on every submit, including invalid), or drop the manual ref and rely on `formState.isSubmitting` for the in-flight guard. Add a test that clicks the primary button on an invalid form, then clicks again after correction, asserting the second submit fires.

zapolsky (commented):
🔴 **Blocked issue** — see the inline comment. This is a new defect introduced by this PR's incremental change (removing the `!isValid` disable in CREATE mode), not present at the previously-reviewed head. It should be addressed before merge, but this review does not formally block — a human owns the merge decision. Posted by automated OWOX review.

zapolsky:
## 🔴 One blocking issue found — see the inline review

Incremental re-review against the worktree (`reviewer`, `logic-hunter`, fresh-eyes generalist, `tester`, + an adversarial skeptic pass on the blocker). Scope: the changes since the previously-reviewed head `bc8036a3`.

**Correction to my earlier review:** the previous pass (at `bc8036a3`) reported *no blocking issues*. The change made since then — removing the `!isValid` disable on the CREATE-mode submit button — exposed a real submit-button lockup. Details and the full evidence chain are in the inline comment on `ReportFormActions.tsx`. Affects Email & Google Sheets report forms (CREATE and EDIT); Looker Studio is unaffected.

**Previously-raised items — confirmed resolved (verified in code):**
- **Exhaustive switch** in `createDataStorageFormResolver` — the `default: { const _exhaustive: never = type }` block is genuinely exhaustive; a new `DataStorageType` without a case will fail to compile. ✅
- **Placeholder credentials** — double-guarded: the resolver strips `credentials` from `result.values` while copying (`data-storage-form-resolver.ts:71-82`) *and* `handleSubmit` deletes them from the payload (`DataStorageForm.tsx:198-200`); per-type mappers only emit credentials when present. No placeholder reaches the wire. ✅

- **Verified safe (the security-sensitive parts):** the credential-bypass resolver only bypasses when a copy-source is selected and fully enforces credentials otherwise — no storage type can be saved credential-less without a copy source. The `focusFirstInvalidField` frame-poll is bounded (`MAX_FOCUS_ATTEMPTS = 20`) and leak-free. The auto-open-once-per-submit logic opens via `setIsOpen(true)` directly, so it never overwrites the persisted localStorage collapse preference.
- **`apps/web/src/utils/form-utils.ts`** — the frame-poll has no cancellation, so rapid repeated invalid submits spawn overlapping `requestAnimationFrame` loops (cosmetic scroll/focus thrash, bounded). Optional: keep a single `cancelAnimationFrame` handle.
- **`apps/web/src/utils/form-utils.ts:22`** — on a *programmatic* submit (the report dropdown path passes no event) `container` falls back to `document`, so the focus query isn't form-scoped. Harmless with a single form mounted; worth scoping by form id/ref if two forms can ever co-exist.
- **Changeset:** this is a user-facing fix but I don't see a changeset for it in the PR (the changesets in the tree are from other merged PRs). Add a `'owox': patch`/`minor` changeset (`changeset-generator`).
- **Coverage gap:** no test exercises invalid-submit → fix → resubmit on `ReportFormActions` — exactly the path the inline blocker breaks. Adding it would have caught this and guards against regressions.

<sub>Automated OWOX review · worktree multi-agent passes · the blocker was adversarially verified (one earlier agent's "disproved" was traced to a faulty repro that used RHF's `formState.isSubmitting` instead of the production manual-state prop).</sub>

apps/web/src/features/data-marts/reports/edit/components/shared/ReportFormActions.tsx
romandubovyi: Addressed in `850b2c85d`.

Email and Google Sheets report forms now pass `isSubmitting || form.formState.isSubmitting` into `ReportFormActions`. The local pending guard still blocks the double-submit race before the manual submit state updates, while RHF's `formState.isSubmitting` now cycles for invalid submits too, allowing the existing effect to release `submitPendingRef` after validation errors.

Added a regression test covering the broken path: primary submit on an invalid form, submitting state cycle, then a second primary submit is accepted. Also added the changeset for the main PR behavior: surfacing validation errors hidden in collapsed form sections.

Verified:
- `npm --workspace @owox/web test -- src/features/data-marts/reports/edit/components/shared/ReportFormActions.test.tsx`
- `npm --workspace @owox/web run lint -- src/features/data-marts/reports/edit/components/shared/ReportFormActions.test.tsx`
- `npm --workspace @owox/web run type-check`

romandubovyi: Small follow-up: amended the fix commit to `532b9c0c3` to avoid the DOM `SubmitEvent` vs React submit handler type mismatch in `ReportFormActions.test.tsx`. The test now uses an inferred inline form submit handler and keeps the mock event-free.

Re-verified:
- `npm --workspace @owox/web test -- src/features/data-marts/reports/edit/components/shared/ReportFormActions.test.tsx`
- `npm --workspace @owox/web run lint -- src/features/data-marts/reports/edit/components/shared/ReportFormActions.test.tsx`
- `npm --workspace @owox/web run type-check`

zapolsky: ✅ **Verified fixed** at `532b9c0c3`. Passing `isSubmitting || form.formState.isSubmitting` is exactly the right fix: because the forms use the async `zodResolver`, RHF's `handleSubmit` commits `formState.isSubmitting = true` (before the awaited resolver) and then `= false` (after `onInvalid`) as **two separate renders**, so the combined prop now transitions on every submit — including an invalid one. The `[isSubmitting]` reset effect (`ReportFormActions.tsx:39-43`) re-runs on the trailing `→ false` edge and clears `submitPendingRef`, so the primary button **and** the dropdown release after a failed validation. Confirmed for both Email (`EmailReportEditForm.tsx:1016`) and Google Sheets (`GoogleSheetsReportEditForm.tsx:530`); Looker Studio was never affected (it already used `form.formState.isSubmitting`). Thanks for the quick turnaround.

One heads-up, non-blocking: the added unit test (`ReportFormActions.test.tsx`) hand-feeds the `isSubmitting` prop `true→false` directly, so it passes even if both call sites are reverted to the buggy `isSubmitting={isSubmitting}` — it exercises the (already-correct) component, not the parent wiring that this fix actually changed. An integration-level test on `EmailReportEditForm`/`GoogleSheetsReportEditForm` (submit invalid → assert a second submit fires) would be the one that truly guards this regression. Not blocking — the production fix itself is solid.

zapolsky:

Incremental re-review at `532b9c0c3` (targeted run). Scope: the changes since the previously-reviewed head `4fb9bc3b6`.

**Previously-reported blocker — confirmed fixed.** The submit-button lockup (`report-submit-button-locks-after-invalid-click`) is resolved: both Email and Google Sheets forms now pass `isSubmitting || form.formState.isSubmitting`, which gives the `submitPendingRef` reset effect a real `true→false` edge on every submit (verified against the installed RHF source — the async `zodResolver` guarantees two committed renders). Verification chain and an in-thread note are on the [original blocker thread](https://github.com/OWOX/owox-data-marts/pull/1323#discussion_r3428145707). No new blockers in the delta; the fix is minimal, correctly scoped, and introduces no regressions.

<sub>Not attributable to this PR: `tsc -b` in the review worktree reports `FormSectionProps.fields` errors because the worktree's `@owox/ui` symlink resolves to the main-branch `packages/ui` (which lacks this PR's `fields` prop); CI installs the workspace coherently and is unaffected. · Automated OWOX review · worktree multi-agent passes · the fix was adversarially re-verified.</sub>

zapolsky (approved):

PR #1324, opened 2026-06-12
refactor(backend): allow saving data mart definition without storage validation

romandubovyi:
### Description
Currently, saving a data mart definition of type `VIEW`, `TABLE`, or `TABLE_PATTERN` is blocked if the associated storage is not fully configured or valid. This PR bypasses this validation during the definition saving stage (`UpdateDataMartDefinitionService`), allowing users to save their definition drafts regardless of the storage status.

### Changes
- Removed `definitionValidatorFacade` injection and `checkIsValid` check from `UpdateDataMartDefinitionService`.
- Validation still runs on publishing (`PublishDataMartService`), schema actualization (`ActualizeDataMartSchemaService`), and metadata generation.

### Testing
- Executed and passed backend e2e tests.

### DIFF
--- a/apps/backend/src/data-marts/use-cases/actualize-data-mart-schema.service.spec.ts
+++ b/apps/backend/src/data-marts/use-cases/actualize-data-mart-schema.service.spec.ts
@@ -1,0 +1,81 @@
+jest.mock('typeorm-transactional', () => ({
+  Transactional: () => (_target: unknown, _key: string, descriptor: PropertyDescriptor) =>
+    descriptor,
+}));
+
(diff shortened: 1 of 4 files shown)


kalnik-a-a (approved):
Non-blocking: please consider adding a focused regression test for the new contract: saving a TABLE/VIEW/TABLE_PATTERN definition should succeed even when storage validation would fail, while publish/schema actualization must still reject it. The behavior looks intentional, but a test would make it much harder to accidentally move this validation back into the save path later.

PR #1327, opened 2026-06-15
feat(web): add quick run functionality for reports in visible area, don't hide them

romandubovyi:
- Introduced a quick run feature for reports, allowing users to execute reports more efficiently.
- Added an undo toast notification to provide users with the option to revert the quick run action.

### DIFF
--- a/apps/web/src/features/data-marts/reports/shared/components/ReportQuickRunCell.tsx
+++ b/apps/web/src/features/data-marts/reports/shared/components/ReportQuickRunCell.tsx
@@ -43,6 +43,10 @@ export function ReportQuickRunCell({ report, onRunSuccess }: ReportQuickRunCellP
     mountedRef.current = true;
     return () => {
       mountedRef.current = false;
+      if (toastIdRef.current) {
+        toast.dismiss(toastIdRef.current);
+        toastIdRef.current = null;
+      }
     };
   }, []);
 
--- a/apps/web/src/features/data-marts/reports/shared/components/ReportQuickRunCell.tsx
+++ b/apps/web/src/features/data-marts/reports/shared/components/ReportQuickRunCell.tsx
@@ -1,0 +1,137 @@
+import { useState, useEffect, useCallback, useRef } from 'react';
+import { Loader2, Play } from 'lucide-react';
+import { toast } from 'react-hot-toast';
+import { Button } from '@owox/ui/components/button';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from '@owox/ui/components/tooltip';
+import type { DataMartReport } from '../model/types/data-mart-report';
+import { ReportStatusEnum } from '../enums';
+import { useReport } from '../model';
+import { RunUndoToast } from './RunUndoToast';
+
+const GRACE_PERIOD_MS = 3000;
+
+interface ReportQuickRunCellProps {
+  report: DataMartReport;
+  onRunSuccess?: () => void | Promise<void>;
+}
+
+export function ReportQuickRunCell({ report, onRunSuccess }: ReportQuickRunCellProps) {
+  const canRun = report.canRun;
+  const [isRunning, setIsRunning] = useState(report.lastRunStatus === ReportStatusEnum.RUNNING);
+  const [isPending, setIsPending] = useState(false);
+  const [isOptimisticRunning, setIsOptimisticRunning] = useState(false);
+  const { runReport } = useReport();
+
+  useEffect(() => {
+    const running = report.lastRunStatus === ReportStatusEnum.RUNNING;
+    setIsRunning(running);
+    if (running) {
+      setIsOptimisticRunning(false);
+    }
+  }, [report.lastRunStatus]);
+
+  const mountedRef = useRef(true);
+  const hasRunRef = useRef(false);
+  const toastIdRef = useRef<string | null>(null);
+
+  useEffect(() => {
+    mountedRef.current = true;
+    return () => {
+      mountedRef.current = false;
+    };
+  }, []);
+
+  const executeRun = useCallback(async () => {
+    if (hasRunRef.current) return;
+    hasRunRef.current = true;
+
+    if (toastIdRef.current) {
+      toast.dismiss(toastIdRef.current);
+      toastIdRef.current = null;
+    }
+
+    setIsPending(false);
+    setIsOptimisticRunning(true);
+
+    try {
+      await runReport(report.id);
+      if (onRunSuccess) await onRunSuccess();
+    } catch (error) {
+      console.error('Failed to run report:', error);
+    } finally {
+      if (mountedRef.current) {
+        setIsOptimisticRunning(false);
+      }
+    }
+  }, [runReport, report.id, onRunSuccess]);
+
+  const handleRun = useCallback(
+    (e: React.MouseEvent) => {
+      e.stopPropagation();
+      if (!canRun || isPending || isRunning || isOptimisticRunning) return;
+
+      hasRunRef.current = false;
+      setIsPending(true);
+
+      const id = toast.custom(
+        t => (
+          <RunUndoToast
+            toastId={t.id}
+            reportName={report.title}
+            gracePeriodMs={GRACE_PERIOD_MS}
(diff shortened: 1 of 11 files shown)


apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
```diff
@@ -138,7 +139,6 @@ function ProjectReportActionsCell({
     row: { original: report },
     onDeleteSuccess: onReportActionComplete,
     onEditReport,
-    onRunSuccess: onReportActionComplete,
   };
```
kalnik-a-a, line 143: Criticality: Blocking.

This project-wide actions cell no longer passes `onRunSuccess` into `EmailActionsCell` / `GoogleSheetsActionsCell`, while those menu actions still expose `Run report`. Running from the dropdown now starts the backend job but does not refresh the local `reports` state on this page, so `lastRunStatus` stays stale, `visibleRunningReportIds` never starts polling, and the new quick-run button can remain enabled for a second run. Please either pass `onReportActionComplete` through for the dropdown run path again or remove that duplicate run action, and add coverage for the dropdown path.

Reviewed by Codex (GPT-5.5).

apps/web/src/features/data-marts/reports/shared/components/ReportQuickRunCell.tsx
```diff
@@ -0,0 +1,137 @@
...
+  }, [report.lastRunStatus]);
+
+  const mountedRef = useRef(true);
+  const hasRunRef = useRef(false);
+  const toastIdRef = useRef<string | null>(null);
+
+  useEffect(() => {
+    mountedRef.current = true;
+    return () => {
+      mountedRef.current = false;
```
kalnik-a-a, line 45: Criticality: Non-blocking.

The pending toast is never dismissed on unmount. If the row or page disappears during the grace period, the global toast can still call `executeRun`; that path calls `setIsPending(false)` and `setIsOptimisticRunning(true)` before checking `mountedRef`, then starts the report after the user has left this context. Please dismiss or cancel `toastIdRef.current` in cleanup so navigation or filtering cannot trigger a delayed run.

Reviewed by Codex (GPT-5.5).

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
```diff
@@ -273,6 +273,16 @@ export default function DataMartReportsPage() {
 
   const columns = useMemo<ColumnDef<DataMartReport>[]>(
     () => [
+      {
+        id: 'quickRun',
```
kalnik-a-a, line 278: Criticality: Non-blocking.

This new utility column is hideable and has no `meta.title`, so the column toggle shows it as raw `quickRun` and lets users hide the primary quick-run affordance. If the goal is to keep quick run visible, set `enableHiding: false`; otherwise add a user-facing title. The same applies to the Email and Google Sheets report tables.

Reviewed by Codex (GPT-5.5).

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
romandubovyi: Fixed. Added `onRunSuccess: onReportActionComplete` to `actionProps` in `ProjectReportActionsCell`. Also added `onRunSuccess` prop to `EmailActionsCell` and wired the callback after `runReport` — it was missing from that component entirely, unlike `GoogleSheetsActionsCell`.

apps/web/src/features/data-marts/reports/shared/components/ReportQuickRunCell.tsx
romandubovyi: Fixed. The `mountedRef` cleanup now calls `toast.dismiss(toastIdRef.current)` on unmount, preventing the deferred `executeRun` from firing after the component is gone.

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
romandubovyi: Fixed. Added `enableHiding: false` to the `quickRun` column in all three tables (`DataMartReportsPage`, `GoogleSheetsReportsTable` columns, `EmailReportsTable` columns).

kalnik-a-a (approved):
Approved.

zapolsky:

Deep multi-agent review against the worktree (`reviewer`, `logic-hunter`, `tester`, fresh-eyes generalist). The undo/grace-period lifecycle is solid: the `activeRef`/`hasRunRef` guards correctly prevent double-runs and post-cancel firing, the cell-unmount toast-dismiss + `mountedRef` setState guard are right, and Cancel is a true deferral (no backend call) rather than a fake undo. The 3 issues @kalnik-a-a flagged are **verified present and correct** in the current head (`onRunSuccess` threaded into the project-page actions cells, `toast.dismiss` on unmount via `mountedRef`, `enableHiding:false` on the quickRun column in all three tables) — not re-raised here. CI checks: `tsc -b` 0 errors, ESLint clean, all 39 unit tests pass.

- **Rows are index-keyed, so a list re-sort/refresh during the 3s countdown misattributes the pending UI to the wrong row.** `useBaseTable` calls `useReactTable` with no `getRowId`, and `BaseTable.tsx:176` uses `key={row.id}` → react-table falls back to the row index. If the report list reorders within the grace window (the 5s auto-refresh re-sorting by status, a higher row deleted, or a filter/search change), the `ReportQuickRunCell` instance at a given position is re-rendered with a *different* report, so its spinner/disabled state visually lands on the wrong report's row. **The correct report still runs** (`executeRun` closes over the click-time `report.id`, `ReportQuickRunCell.tsx:53-75`), so there's no data harm — but the feedback is misleading. Cheap fix: add `getRowId: row => row.id` in `useBaseTable` so cell identity tracks the report across re-sorts.

- **A failed run is never surfaced — the user sees "Report run started" and the countdown completing.** `runReport` (`useReport.ts`, not modified by this PR) catches its own errors, does `console.error` + `trackEvent`, and **never rethrows**, so `await runReport(report.id)` in `executeRun` always resolves and the `catch` at `ReportQuickRunCell.tsx:68` is dead code. On a backend failure (permissions/quota/500) there is no error toast at the quick-run entry point. The root cause is pre-existing (the dropdown "Run report" shares it), but the quick-run's strong "it's starting" affordance makes the silent failure more misleading. Worth having `runReport` rethrow (or return a result) so both entry points can show a `toast.error`.

- **Unmount during the grace window has inconsistent semantics.** `toast.dismiss` only hides the toast and schedules removal after `removeDelay` (~1s), so unmounting the row in the **last ~1s** of the 3s window still lets the interval fire `onConfirm` → the run executes with no visible toast; unmounting earlier cancels it silently. Neither is wrong per se, but the behavior depends on *when* in the countdown the row unmounts. Consider owning the timer in the cell (clear on unmount) and making "unmount = commit" or "unmount = cancel" an explicit decision.

- **Minor:** (a) the per-data-mart Email/GoogleSheets quick-run cells are wired without `onRunSuccess`, so after a run they rely only on the 5s poll while the in-row dropdown refreshes immediately — inconsistent freshness; (b) `react-hot-toast`'s `toastLimit` (20) means that if ≥20 toasts appear within the 3s window the undo toast is evicted without firing `onConfirm`/`onCancel`, leaving that cell's `isPending` stuck (edge case, recovers on remount); (c) every row now exposes two run paths — the immediate dropdown "Run report" (no undo) and the quick-run icon (3s undo) — so the undo guarantee is only partial.

- **Test coverage (informational):** all 39 tests pass, and the cancel-race / re-entrancy / disabled-state guards are well covered. Gaps are at integration/lifecycle boundaries: the cell never drives the real `RunUndoToast` timer (the toast is mocked in the cell tests), and the failed-run branch, unmount-dismiss, `onRunSuccess`, and concurrent quick-runs are untested.

<sub>Automated OWOX review · worktree multi-agent passes · no blockers. The 2 candidate blockers raised by the fresh-eyes pass were adversarially checked: the "wrong report runs" claim is refuted (the click-time closure runs the correct report), and the silent-failure path is a pre-existing weakness of `runReport`, not introduced here. `tsc -b` 0 errors · ESLint clean · 39/39 tests pass.</sub>

PR #1329, opened 2026-06-16
feat(data-marts): unified column identifiers for output control filters

zapolsky:
## Summary

Pre-join (slice) filters now reference a column by the **same fully qualified identifier** as post-join filters — e.g. `category_details__item_event_count` — instead of a raw column name plus a separate `aliasPath`. A single blended-field index resolves that unified name back to its source / CTE / raw column for both validation and SQL generation across all storages.

### DoD
- [x] Leave the fully qualified column name as-is (e.g. `category_details__item_event_count`)
- [x] Pre-join filter uses the same column format as post-join (no `aliasPath`)

## Why

Previously a slice identified a column with **two** fields (`column` = raw name + `aliasPath`), while post-join filters used the fully qualified blended name. The frontend also reconstructed composite keys (`aliasPath\0column`) in several places, so FE and BE could disagree on how a slice column is named. Unifying on the existing blended-field `name` removes that whole class of mismatch — every filter / slice / sort now refers to a column the same way, on both sides.

## Example

Stored slice rule, before → after:

```jsonc
// before — two fields carry the identifier
{ "column": "item_event_count", "aliasPath": "category.details", "placement": "pre-join", "operator": "gt", "value": 0 }

// after — one unified identifier, no aliasPath
{ "column": "category_details__item_event_count", "placement": "pre-join", "operator": "gt", "value": 0 }
```

Unified name format: `<aliasPath with dots→'_'>__<originalFieldName with dots→'_'>`.

## What changed

**Backend**
- New `buildBlendedFieldIndex` — one flat index `name → { aliasPath, cteName, originalFieldName, type, isIncluded }`; replaces the two parallel two-level maps (`buildKnownPaths` in the validator and `BlendedColumnTypes.preJoin`).
- Filter Zod schema **and** the hand-written OpenAPI DTO drop `aliasPath`.
- Validator + blended query builder resolve pre-join filters through the index (BigQuery, Athena/Trino, Snowflake, Redshift, Databricks). Emitted SQL is unchanged — the unified name resolves back to the raw column.
- Review hardening: collision guard in the index (two fields folding to the same unified name now throws instead of silently mis-resolving); a pre-join filter on a non-actualized schema returns **400** (was a 500); excluded-source slices are gated on the run path too.

**Frontend**
- Slice UI emits the unified identifier and no longer reconstructs composite keys client-side.

**Migration** — `1779200000000-migrate-prejoin-filter-to-unified-column`
- Converts stored reports: folds `aliasPath` + raw column into the unified name, drops `aliasPath`. Idempotent (already-unified rows are skipped).
- Self-contained: the transform is a private method and the file exports only the migration class (no shared helper).

## Migration notes (rolling deploy)

- This is a **data-shape change**, not pure expand-contract. During a rolling deploy, an old pod that runs a just-migrated report's slice would reject it (the old schema still expects `aliasPath`). Blast radius is **small and transient** — only reports that use slices (a new, low-usage feature), only during the deploy window. Recommendation: drain-then-migrate, or accept the transient window (the migration runs last).
- `down()` is intentionally a **no-op**: the unified name is lossy (nested-field dots collapse to `_`), so the reverse split can't reliably reconstruct `aliasPath` + raw column — there is no clean data rollback.

## Testing

- Backend unit (data-marts) + web suites green; `nest build` + `tsc -b` clean.
- e2e: `output-controls` + `blended-output-controls` 30/30 — acceptance, excluded-source / unknown-column rejection, disconnected-columns.
- Integration vs **real cloud**: Athena + BigQuery 92/92 (slices execute end-to-end); Snowflake / Redshift / Databricks output-controls.
- Migration round-trip (`up → down → up`, 96 migrations) **and** the actual JSON data transform verified on **both SQLite and MySQL** — including nested-struct dot handling and MySQL native JSON read/write of `filterConfig`.
- Changeset included (`'owox': minor`).

### DIFF
--- a/.changeset/unified-column-identifiers.md
+++ b/.changeset/unified-column-identifiers.md
@@ -1,0 +1,11 @@
+---
+'owox': minor
+---
+
+# Unified column identifiers for output control filters
+
+Slice (pre-join) filters now reference a column by the same fully qualified identifier as
+regular output filters — for example `category_details__item_event_count` — instead of a raw
+column name plus a separate `aliasPath`. This makes every filter, slice, and sort refer to a
+column the same way across the API and the Web/Extension report editor. Existing saved reports
+are migrated automatically, so no manual changes are needed.
--- a/apps/backend/src/data-marts/controllers/spec/report-openapi.ts
+++ b/apps/backend/src/data-marts/controllers/spec/report-openapi.ts
@@ -127,7 +127,11 @@ export class ReportRelativeDateFilterValueApiDto {
 }
 
 export class ReportFilterRuleApiDto {
-  @ApiProperty({ minLength: 1 })
+  @ApiProperty({
+    minLength: 1,
+    description:
+      'Output column name. For slice filters (placement=pre-join), use the fully qualified blended column identifier, e.g. category_details__item_event_count.',
+  })
   column: string;
 
   @ApiProperty({
@@ -156,14 +160,6 @@ export class ReportFilterRuleApiDto {
     description: 'Use pre-join for slice filters. Omit for normal output filters.',
   })
   placement?: 'pre-join' | 'post-join';
-
-  @ApiPropertyOptional({
-    minLength: 1,
-    maxLength: 255,
-    pattern: '^[a-z0-9_]+(\\.[a-z0-9_]+)*$',
-    description: 'Required for slice filters when placement is pre-join.',
-  })
-  aliasPath?: string;
 }
 
 export class ReportSortRuleApiDto {
@@ -232,7 +228,8 @@ const commonReportRequestProperties = {
     nullable: true,
     maxItems: 50,
     items: { $ref: getSchemaPath(ReportFilterRuleApiDto) },
-    description: 'Output filters. Use placement=pre-join and aliasPath for slice filters.',
+    description:
+      'Output filters. Use placement=pre-join for slice filters (column = unified blended identifier).',
   },
   sortConfig: {
     type: 'array',
--- a/apps/backend/src/data-marts/controllers/spec/report.openapi.spec.ts
+++ b/apps/backend/src/data-marts/controllers/spec/report.openapi.spec.ts
@@ -127,7 +127,10 @@ describe('ReportController OpenAPI', () => {
     );
     expect(filterRule.properties.value.oneOf).toHaveLength(3);
     expect(filterRule.properties.placement.enum).toEqual(['pre-join', 'post-join']);
-    expect(filterRule.properties.aliasPath.description).toContain('slice');
+    expect(filterRule.properties.aliasPath).toBeUndefined();
+    expect(filterRule.properties.column.description).toContain(
+      'category_details__item_event_count'
+    );
 
     expect(updateProperties.sortConfig).toMatchObject({
       type: 'array',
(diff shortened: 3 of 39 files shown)


zapolsky:

Deep multi-agent review against the worktree (`reviewer`, `logic-hunter`, `architect`, `tester`, fresh-eyes generalist). The core mechanism is sound and verified end-to-end: the unified name `<aliasPath dots→_>__<originalFieldName dots→_>` is encoded identically by the three independent producers (`blendable-schema.service.ts:249`, the migration, and the FE pass-through of `field.name`); the collision guard in `buildBlendedFieldIndex` (`blended-field-index.ts:28-43`) is reachable on every path that resolves a name (validator + run-path both build the index before any SQL); emitted SQL and identifier quoting are provably unchanged (raw column still routed through `quoteFieldRef`, values still parameterized); the 500→400 conversions and excluded-source run-path gating are correctly placed and test-locked; and the FE / OpenAPI / validator contracts are consistent after the `aliasPath` drop. Migration idempotency, post-join-untouched, null/malformed-skip, and nested-dot handling were each traced and confirmed.

- **Migration has no automated test.** The data-shape rewrite of stored `filterConfig` JSON (`1779200000000-…ts`) is covered only by a manual claim ("verified on SQLite and MySQL"); a repo-wide search finds no spec that imports or exercises it. The `migrateRules`/`parse` logic is pure and trivially testable — a unit test over idempotency (re-run skips already-unified rows), pre-/post-join discrimination, malformed-skip, and dot-collapse would close a real gap for a data-mutating migration.
- **"Round-trip up→down→up" wording is misleading.** `down()` is an intentional no-op (`:60-62`), so no reverse transform was (or could be) exercised — the second `up` is just an idempotency re-run. Worth rewording in the PR/testing notes.
- **Rolling-deploy blast radius is broader than the PR note states.** The PR documents that an old pod "running a just-migrated report's slice would reject it." In fact, because `Report.filterConfig`'s Zod `from()` transformer parses on **every load** and the old schema's `superRefine` *requires* `aliasPath` for pre-join rules, an old pod will throw (500) when **listing/loading** any report that contains a pre-join filter — e.g. `list-reports-by-project` / `list-reports-by-data-mart` — not only when running one. New pods are safe (migrate-before-serve, `bootstrap.ts`), so the exposure is the rollout window + a code rollback past this migration. Low real blast radius (slices are new/low-usage) and a conscious forward-only tradeoff — just worth stating the read-path impact explicitly in the release note.
- **`migration:48-50` comment is inaccurate.** It says the write "bumps version/modifiedAt", but a QueryBuilder `.update().set()` does **not** touch `@VersionColumn()`/`@UpdateDateColumn()` (those only update via `repository.save()`). Behavior is fine; the comment is wrong.
- **Dead code:** `isPreJoinFilter` (`apps/web/.../output-config.ts`) has no callers, and `FILTER_ALIAS_PATH_UNKNOWN` (`output-controls-validator.service.ts:23` + its `collectDisconnectedOutputControlRefs` case) is no longer emitted. Remove or keep intentionally for API back-compat.

<sub>Automated OWOX review · worktree multi-agent passes · all candidate findings traced to closure; no blocker survived verification.</sub>

kalnik-a-a (approved):

PR #1331, opened 2026-06-17
feat: add MCP OAuth integration

amarchenk0:


apps/backend/src/ee/mcp/mcp-boundary.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { readdirSync, readFileSync, statSync } from 'node:fs';
```
romandubovyi, line 1: 🟢 **Praise:** This boundary test is an excellent architectural safeguard — a static import check that fails fast in CI if EE/MCP code ever accidentally couples to a specific IDP implementation. Much better than documentation or review-only enforcement.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  private async loadProjectsOrEmpty(
+    provider: ReturnType<IdpProviderService['getProvider']>,
+    accessToken: string
+  ) {
+    try {
+      return await this.projectSelectionService.loadProjects(provider, accessToken);
+    } catch {
```
romandubovyi, line 120: 💡 **Suggestion:** The empty `catch` block silently swallows any project-loading failure (IDP unavailable, timeout, network error) with no log entry — not a blocker, but makes production debugging harder. When this happens and the user has only one project in their token, they are silently auto-authorized without seeing the project selection screen. Consider adding `this.logger.warn('Failed to load projects, falling back to token context', error)` so issues are visible in logs.

apps/backend/src/ee/mcp/sdk/mcp-streamable-http-session.registry.ts
```diff
@@ -0,0 +1,259 @@
+import { randomUUID } from 'node:crypto';
+import { Injectable, Logger, NotFoundException, UnauthorizedException } from '@nestjs/common';
+import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
+import type { Request, Response } from 'express';
+import type { McpAuthContext } from '../auth/mcp-auth-context';
+import { McpSdkServerFactory } from './mcp-sdk-server.factory';
+
+const MCP_SESSION_IDLE_TTL_MS = 30 * 60 * 1000;
```
romandubovyi, line 8: 🔵 **Question:** Sessions are stored in a `Map` in process memory. This means all sessions are lost on restart (clients get `404 Unknown MCP session` until they re-initialize), and stateful sessions won't work across multiple backend replicas. Is single-instance deployment an explicit constraint for MCP? If horizontal scaling is on the roadmap, this registry will need an external store (Redis/DB). Worth a comment in the code either way.

apps/backend/src/idp/oauth/oauth-request.validator.ts
```diff
@@ -0,0 +1,116 @@
...
+  private async getClient(clientId: string): Promise<OAuthRegisteredClient> {
+    const client = await this.clientRegistry.get(clientId);
+    if (!client) {
+      throw new BadRequestException('unknown client_id');
+    }
+    return client;
+  }
+
+  private parseScopes(raw: unknown, client: OAuthRegisteredClient): McpScope[] {
+    const value = this.requiredString(raw, 'scope');
```
romandubovyi, line 94: 🔵 **Question:** `scope` is treated as required here (via `requiredString`), but RFC 6749 §3.3 defines it as optional — if omitted, the server should default to the client's registered scopes. Some OAuth libraries omit `scope` in the authorization request. Is this intentional? If so, a comment explaining the deviation (e.g. "MCP clients are expected to always specify scope") would help future maintainers.

packages/idp-protocol/src/types/oauth.ts
```diff
@@ -0,0 +1,93 @@
...
+  codeVerifier: z.string().min(1).optional(),
+});
+export type OAuthTokenExchangeRequest = z.infer<typeof OAuthTokenExchangeRequestSchema>;
+
+export const OAuthTokenExchangeResultSchema = z.object({
+  access_token: z.string().min(1),
+  refresh_token: z.string().min(1).optional(),
+  token_type: z.enum(['Bearer', 'bearer']),
+  expires_in: z.number().int().positive(),
+  scope: z.string().min(1),
```
romandubovyi, line 86: 🔵 **Question:** `scope` is required here (`z.string().min(1)`), but RFC 6749 §5.1 makes it optional in token responses when identical to the requested scope. The refresh_token grant type is declared as supported, but there are no tests for that path. If the identity service omits `scope` on token refresh, `McpOAuthTokenExchangeResponseSchema.parse(data)` in `IdentityOwoxClient.exchangeMcpOAuthToken` will throw. Is the identity service guaranteed to always return `scope`? If not, this field should be `.optional()`.

romandubovyi (approved):

PR #1333, opened 2026-06-17
web(table-filters): set default table filter operator to is

romandubovyi:
## Summary

Changed shared table filters so fields that support both `contains` and `is` now default to `is`.

## Why

For fields with a fixed option list, users usually want to select an exact value from the dropdown. When `contains` was selected by default, the value control switched to a free-text input and users had to manually change the condition to `is` before selecting from the list.

## What changed

- Added shared default operator selection for table filters.
- If a filter field supports both `contains` and `is` (`eq` internally), the default operator is now `is`.
- Fields that do not offer both operators keep their existing first configured operator as the default.
- Added regression tests for the default operator selection.

## Validation

- `npm --workspace apps/web test -- filter-utils.test.ts`
- `npm --workspace apps/web run lint -- src/shared/components/TableFilters/FiltersForm.tsx src/shared/components/TableFilters/filter-utils.ts src/shared/components/TableFilters/filter-utils.test.ts`

### DIFF
--- a/.changeset/table-filters-default-is-condition.md
+++ b/.changeset/table-filters-default-is-condition.md
@@ -1,0 +1,5 @@
+---
+"owox": minor
+---
+
+Set default table filter operator to `is` for fields supporting both `contains` and `is`. Reset the operator when switching filter fields to ensure the new default takes effect.
--- a/apps/web/src/shared/components/TableFilters/FiltersForm.tsx
+++ b/apps/web/src/shared/components/TableFilters/FiltersForm.tsx
@@ -11,7 +11,7 @@ import {
 import { Button } from '@owox/ui/components/button';
 import { Plus, X } from 'lucide-react';
 import type { FilterConfigItem, FilterOperator, FiltersState } from './types';
-import { isFilterRowValid } from './filter-utils';
+import { getDefaultFilterOperator, isFilterRowValid } from './filter-utils';
 import { SelectValueControl, InputValueControl } from './index';
 
 /* ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


zapolsky:

Deep multi-agent review against the worktree (`reviewer`, `logic-hunter`, fresh-eyes generalist). The change is small and correct in isolation: `getDefaultFilterOperator` (`filter-utils.ts:5-13`) returns `eq` only when the field supports **both** `contains` and `eq`, otherwise the first configured operator — all branches verified, and the empty-list case (`undefined`) is guarded at the call site (`FiltersForm.tsx:177`). Existing saved filters are not rewritten (the `if (currentOperator) return` guard and `usePersistentFilters` both preserve the stored operator), and no stale free-text value can carry into the `is` dropdown via this PR's path (the value is `[]` whenever the default flips).

- **Question — the new `is` default only takes effect on a row's *first* operator assignment, not on field switches.** The auto-set effect is gated by `if (currentOperator) return` (`FiltersForm.tsx:172`), and the field selector's `onValueChange` clears the value but not the operator (`:217-220`). So the headline intent ("default to `is` for any field supporting both") is only realized on row creation; switching to a different both-capable field keeps the prior operator. It happens to be benign today (all real configs include `eq`/`neq`), but if "per-field default" is the actual requirement, reset the operator in the field `onValueChange`. This is pre-existing behavior the PR re-touched — worth confirming intent.
- **Changeset:** this changes default filter UX (user-facing) but I don't see a changeset in the PR. Add a `'owox': patch`/`minor` changeset (`changeset-generator`).
- **Test gap:** the only-`is` branch (a field with `eq` but no `contains`, e.g. `['eq','neq']` → `eq`) isn't asserted. Adding it would lock the contract — without it, a future "simplify to `operators.includes('eq')`" refactor would silently flip `['neq','eq']` from `neq` to `eq` while every current test still passes.
- **NIT:** the comment at `FiltersForm.tsx:168` ("when field changes") slightly overstates — the effect runs only when no operator is set yet.

<sub>Automated OWOX review · worktree multi-agent passes · all branches of the new helper traced; no blocker found.</sub>

zapolsky (approved):

romandubovyi:
Thanks for the review @zapolsky !                                                                                                                                                                                                              
                                                                                                                                                                                                                                       1. **Operator reset on field switch**: Added `setValue(opPath, '')` to the field `onValueChange` handler in `FiltersForm.tsx`. This clears the previous operator and correctly triggers the `useEffect` auto-set hook to compute the default operator for the newly selected field.                                                                                                                                                                                        
2. **Missing test**: Added an assertion for the case `['eq', 'neq']` -> `eq` to `filter-utils.test.ts`.                                                                                                                             
3. **Changeset**: Added a patch changeset for `@owox/web` under `.changeset/table-filters-default-is-condition.md`.

PR #1334, opened 2026-06-17
feat(web): readable Output Control fields with search

zapolsky:
## Summary
Makes report Output Control field names readable for business users, stops them overflowing the sidebar, and adds search to the field pickers.

- Filters / Slices / Sort pickers, the column picker, and active-rule chips now show a business-readable field name (alias → leaf) instead of the raw nested column identifier
- Blended fields get the joined data mart name as a muted second line; hovering a field reveals its full path as a tree
- Each picker is searchable — by readable name, data mart name, or the technical field name
- Long nested names truncate instead of overflowing; section info tooltips stay within view in narrow sidebars
- Filter/slice editor popover header shows the readable name as its heading
- Presentation only: stored identifiers (`FilterRule.column`, `aliasPath`, `SortRule.column`, slice keys) are unchanged

### Review follow-ups folded in
- Centralised the pre-join slice-key delimiter in a `makePreJoinKey` helper (one named home for the `\0` delimiter)
- `toPreJoin` helper replaces ad-hoc `as FilterRule` casts when re-stamping slice rules
- Added invariant tests proving the add-filter/slice/sort paths store the raw identifier, not the business label

## Test plan
- [x] `tsc -b` clean
- [x] `vitest` ReportColumnPicker — 105 passing (incl. new invariant tests)
- [x] Filters/Slices/Sort pickers show readable names + data mart line; search matches both readable and technical names
- [x] Long names don't overflow; hover shows the path tree
- [x] Existing filters/slices/sorts still apply correctly (stored identifiers intact)

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FieldSearchPicker.tsx
+++ b/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FieldSearchPicker.tsx
@@ -1,0 +1,128 @@
+import { useState } from 'react';
+import { Plus } from 'lucide-react';
+import { Button } from '@owox/ui/components/button';
+import { Popover, PopoverContent, PopoverTrigger } from '@owox/ui/components/popover';
+import {
+  Command,
+  CommandEmpty,
+  CommandInput,
+  CommandItem,
+  CommandList,
+} from '@owox/ui/components/command';
+import { Tooltip, TooltipContent, TooltipTrigger } from '@owox/ui/components/tooltip';
+
+// Substring match over the readable field name, data-mart name, and the raw
+// field identifier (the item keywords) — so power users can still search by the
+// technical name. cmdk's default fuzzy scorer normalises the value, which makes
+// dotted-name search erratic.
+function commandFilter(value: string, search: string, keywords?: string[]): number {
+  const target = keywords?.join(' ') ?? value;
+  return target.toLowerCase().includes(search.toLowerCase()) ? 1 : 0;
+}
+
+export interface FieldPickerItem {
+  /** Stored identifier handed back on select. */
+  value: string;
+  /** Business-readable field name (line 1). */
+  label: string;
+  /** Joined data mart name (muted line 2); omit for home-mart fields. */
+  dataMartName?: string;
+  /** Full path segments, shown as a vertical tree on hover. */
+  path?: string[];
+}
+
+/** Vertical, indented path so a deep join chain never overflows the sidebar. */
+export function PathTree({ segments }: { segments: readonly string[] }) {
+  return (
+    <div className='font-mono text-[11px] leading-snug'>
+      {segments.map((seg, i) => (
+        <div key={i} style={{ paddingLeft: i * 12 }}>
+          {i > 0 && <span className='opacity-50'>› </span>}
+          {seg}
+        </div>
+      ))}
+    </div>
+  );
+}
+
+interface FieldSearchPickerProps {
+  items: readonly FieldPickerItem[];
+  /** Trigger text, e.g. 'Add filter'. */
+  placeholder: string;
+  searchPlaceholder?: string;
+  onSelect: (value: string) => void;
+}
+
+export function FieldSearchPicker({
+  items,
+  placeholder,
+  searchPlaceholder = 'Search fields…',
+  onSelect,
+}: FieldSearchPickerProps) {
+  const [open, setOpen] = useState(false);
+  return (
+    <Popover open={open} onOpenChange={setOpen} modal={true}>
+      <PopoverTrigger asChild>
+        <Button
+          variant='outline'
+          size='sm'
+          className='text-muted-foreground h-7 w-full justify-start text-xs'
+          aria-expanded={open}
+          aria-haspopup='listbox'
+        >
+          <Plus className='mr-1 h-4 w-4' />
+          {placeholder}
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent align='start' className='w-[var(--radix-popover-trigger-width)] p-0'>
+        <Command filter={commandFilter}>
+          <CommandInput placeholder={searchPlaceholder} className='h-8' />
+          <CommandList>
+            <CommandEmpty>No fields found.</CommandEmpty>
+            {items.map(item => {
+              const content = (
+                <span className='flex min-w-0 flex-1 flex-col'>
+                  <span className='truncate'>{item.label}</span>
+                  {item.dataMartName && (
+                    <span className='text-muted-foreground truncate text-[11px]'>
+                      {item.dataMartName}
+                    </span>
+                  )}
+                </span>
+              );
+              const tree = item.path && item.path.length > 1 ? item.path : null;
+              return (
+                <CommandItem
+                  key={item.value}
+                  value={item.value}
+                  keywords={[item.label, item.dataMartName ?? '', item.value]}
+                  // Use the closure value, not cmdk's normalized arg.
+                  onSelect={() => {
+                    onSelect(item.value);
+                    setOpen(false);
+                  }}
+                >
+                  {tree ? (
+                    <Tooltip>
+                      <TooltipTrigger asChild>{content}</TooltipTrigger>
+                      <TooltipContent
+                        side='bottom'
+                        align='start'
+                        collisionPadding={8}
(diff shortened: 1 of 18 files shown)


apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FieldSearchPicker.tsx
```diff
@@ -0,0 +1,128 @@
...
+} from '@owox/ui/components/command';
+import { Tooltip, TooltipContent, TooltipTrigger } from '@owox/ui/components/tooltip';
+
+// Substring match over the readable field name, data-mart name, and the raw
+// field identifier (the item keywords) — so power users can still search by the
+// technical name. cmdk's default fuzzy scorer normalises the value, which makes
+// dotted-name search erratic.
+function commandFilter(value: string, search: string, keywords?: string[]): number {
+  const target = keywords?.join(' ') ?? value;
+  return target.toLowerCase().includes(search.toLowerCase()) ? 1 : 0;
```
romandubovyi: [Non-blocking] Consider trimming the search query to handle accidental leading/trailing whitespace introduced by the user. For example:

```typescript
return target.toLowerCase().includes(search.trim().toLowerCase()) ? 1 : 0;
```

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/OutputSettingsDropdown.tsx
```diff
@@ -228,15 +256,29 @@ function SlicesSection({
   const [pending, setPending] = useState<PendingSliceColumn | null>(null);
 
   const fieldTypeMap = new Map<string, string>();
```
romandubovyi: [Non-blocking] To avoid reconstructing the lookup maps (`fieldTypeMap`, `labelMap`, and `dataMartByAliasPath`) on every render of `SlicesSection`, we can wrap their initialization in a `useMemo` hook dependent on `joinedSources`. For example:

```typescript
const { fieldTypeMap, labelMap, dataMartByAliasPath } = useMemo(() => {
  const fieldTypeMap = new Map<string, string>();
  const labelMap = new Map<string, string>();
  const dataMartByAliasPath = new Map<string, string | undefined>();
  for (const src of joinedSources) {
    dataMartByAliasPath.set(src.aliasPath, src.dataMartName);
    for (const col of src.columns) {
      fieldTypeMap.set(makePreJoinKey(src.aliasPath, col.name), col.type);
      labelMap.set(makePreJoinKey(src.aliasPath, col.name), fieldDisplayLabel(col.alias, col.name));
    }
  }
  return { fieldTypeMap, labelMap, dataMartByAliasPath };
}, [joinedSources]);
```

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/OutputSettingsDropdown.tsx
```diff
@@ -139,6 +154,11 @@ function filterRowKey(rule: FilterRule, index: number): string {
   return `${rule.placement ?? 'post'}|${rule.aliasPath ?? ''}|${rule.column}|${rule.operator}|${index}`;
 }
 
+// FilterEditorPopover strips placement/aliasPath; re-stamp them for a slice.
+function toPreJoin(rule: FilterRule, aliasPath: string | undefined): FilterRule {
```
romandubovyi, line 158: 🔵 **Question:** `toPreJoin` accepts `aliasPath: string | undefined` and silences the type gap with `as FilterRule`. At the `onUpdateAt(index, toPreJoin(next, rule.aliasPath))` call site, `rule.aliasPath` can be `undefined` if the rule was somehow created without one. Is there a guarantee it is always defined here? If so, narrowing the type or adding a guard would make that explicit and remove the cast.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FieldSearchPicker.tsx
zapolsky: Done in `8d56a6b88` — `commandFilter` now trims the search query before matching (`search.trim().toLowerCase()`). Mirrored in the extension.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/OutputSettingsDropdown.tsx
zapolsky: Done in `8d56a6b88` — the three SlicesSection lookup maps are now built inside a `useMemo` keyed on `joinedSources`, so they no longer rebuild every render. Mirrored in the extension.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/OutputSettingsDropdown.tsx
zapolsky: Good catch — the cast was actually unnecessary. With the `: FilterRule` return type the object literal is contextually typed, so the spread is checked against the discriminated union without `as FilterRule`; dropped in `91bed237b` (TS now verifies the shape rather than trusting it).

On `aliasPath`: it is always defined at both call sites. The edit path’s `rule` comes from `preJoinIndexed` (`placement === "pre-join"`), and the Zod `superRefine` requires `aliasPath` whenever placement is pre-join, so any persisted pre-join rule carries one; the add path passes `pending.aliasPath` (a required `string`). I kept the param as `string | undefined` only so the edit call site does not need a non-null assertion on the optionally-typed `rule.aliasPath`.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FieldSearchPicker.tsx
```diff
@@ -0,0 +1,128 @@
...
+                    </span>
+                  )}
+                </span>
+              );
+              const tree = item.path && item.path.length > 1 ? item.path : null;
+              return (
+                <CommandItem
+                  key={item.value}
+                  value={item.value}
+                  keywords={[item.label, item.dataMartName ?? '', item.value]}
```
romandubovyi: ⚪ **Nit:** `item.dataMartName ?? ''` adds an empty string to `keywords` when `dataMartName` is absent. Harmless, but `[item.label, item.value, ...(item.dataMartName ? [item.dataMartName] : [])]` (or `.filter(Boolean)`) would keep the array clean.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FieldSearchPicker.tsx
```diff
@@ -0,0 +1,128 @@
...
+                  // Use the closure value, not cmdk's normalized arg.
+                  onSelect={() => {
+                    onSelect(item.value);
+                    setOpen(false);
+                  }}
+                >
+                  {tree ? (
```
romandubovyi, line 109: 💡 **Suggestion:** The path-tree tooltip fires immediately on hover, which can obscure adjacent list items while the user is scanning or clicking — making selection feel a bit awkward. Two options worth considering: (1) add a `delayDuration` (e.g. 600ms) to the `<Tooltip>` so it only shows after intentional pause; or (2) move the tree into a small `ⓘ` info icon placed after the label, so the tooltip is only triggered by an explicit hover on that icon rather than the whole row.

romandubovyi (approved):

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FieldSearchPicker.tsx
zapolsky: Done in `4354796c7` — keywords are now `[item.label, item.value, ...(item.dataMartName ? [item.dataMartName] : [])]`, so no empty string is pushed when a field has no data mart name. Mirrored in the extension.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FieldSearchPicker.tsx
zapolsky: Went with option (1) in `4354796c7` — added `delayDuration={600}` to the path-tree `<Tooltip>`, so it only opens after an intentional pause and no longer pops while the user scans or clicks adjacent rows. I kept the whole-row hover as the trigger (the agreed design) rather than moving to an ⓘ icon, but happy to switch to option (2) if you would prefer the explicit trigger. Mirrored in the extension.

PR #1335, opened 2026-06-17
fix(deps): resolve js-yaml & markdown-it audit advisories + prune stale overrides

zapolsky:
## Summary
- Resolves the two open Dependabot advisories surfaced by `npm audit`:
  - **js-yaml** GHSA-h67p-54hq-rp68 (runtime) — the production-reachable copy under `@nestjs/swagger` is pinned to patched `^4.2.0`.
  - **markdown-it** GHSA-6v5v-wf23-fmfq (dev) — bumped to patched `^14.2.0` under `markdownlint-cli2`.
- Drops nine `overrides` a clean-room audit proved are no longer load-bearing: `lodash`, `fast-xml-parser`, `http-proxy-agent`, `handlebars`, `h3`, `kysely`, `path-to-regexp`, `smol-toml`, `happy-dom`.

## Result
- `npm audit --omit=dev`: **0 vulnerabilities** (was 2).
- Full `npm audit`: 34 → 31 moderate. The remaining 31 are all the same js-yaml DoS advisory reaching only **dev** tooling via three `js-yaml@3.14.2` copies (gray-matter, changesets/read-yaml-file, ts-jest/istanbul). Those use the js-yaml v3 API (`safeLoad`), so a 3→4 bump would break the docs build and changesets; left intentionally as a dev-only, non-shipped DoS.

## Validation
- Full workspace `npm run build` — green.
- `@owox/web` tests (happy-dom 20.10): 430 passed.
- `@owox/idp-better-auth` + `@owox/idp-owox-better-auth` tests (kysely 0.29): 239 passed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -225,9 +225,9 @@
       }
     },
     "apps/backend/node_modules/@nestjs/typeorm": {
-      "version": "11.0.1",
-      "resolved": "https://registry.npmjs.org/@nestjs/typeorm/-/typeorm-11.0.1.tgz",
-      "integrity": "sha512-8rw/nKT0S+L+MkzgE9F2/mox7mAgsPlwfzmW9gsESN1lmQtIrVEfiiBwC2O8+guS1jBfQehJIdcdUj2OAp4VUQ==",
+      "version": "11.0.2",
+      "resolved": "https://registry.npmjs.org/@nestjs/typeorm/-/typeorm-11.0.2.tgz",
+      "integrity": "sha512-KQsmOKqIbfq+I1fe88rDq8QNfTU3C4b8MA2qOImhvIbIDOStcC+m+z2itf7vnWGQK0LZBFjWlYdRBID12qgZeg==",
       "license": "MIT",
       "peerDependencies": {
         "@nestjs/common": "^10.0.0 || ^11.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -91,25 +89,25 @@
...
-    "kysely": "^0.28.14",
-    "path-to-regexp": "^8.4.0",
-    "smol-toml": "^1.6.1",
-    "happy-dom": "^20.8.9",
     "picomatch": "^4.0.4",
     "file-type": "^21.3.2",
     "thrift": "^0.23.0",
     "uuid": "^11.1.1",
     "esbuild": "^0.28.1",
+    "markdown-it": "^14.2.0",
```
romandubovyi: 🔵 **Question:** This is a top-level (global) override, unlike the scoped js-yaml fixes above. The changeset claims "same semver-major line" — can you confirm every consumer in the dependency tree was already on `markdown-it@14.x` before this PR? If any package pinned `markdown-it@13`, this forces a cross-major bump which could silently change parsing behaviour. A scoped `"markdownlint-cli2": { "markdown-it": "^14.2.0" }` would be safer if the vulnerability path is only through `markdownlint-cli2`.

package-lock.json
```diff
@@ -776,10 +776,8 @@
       "version": "4.0.3",
       "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-4.0.3.tgz",
       "integrity": "sha512-Qgzu8kfBvo+cA4962jnP1KkS6Dop5NS6g7R5LFYJr4b8Ub94PPQXUksCw9PvXoeXPRRddRNC5C1JQUR2SMGtnA==",
-      "dev": true,
+      "extraneous": true,
```
romandubovyi: 🔵 **Question:** `chokidar` and `readdirp` are now marked `"extraneous": true` — they are orphaned in `node_modules` after removing the `happy-dom` override. Was `npm prune` intentionally skipped? As-is, `npm ci` in CI will still install these two unused packages. A `npm prune` (or a clean `rm -rf node_modules && npm install`) before committing the lockfile would remove them.

package.json
zapolsky: Good call — switched it to a scoped override: `markdownlint-cli2 -> { js-yaml ^4.2.0, markdown-it ^14.2.0 }` (force-pushed).

To answer the question directly: `markdownlint-cli2` is the only `markdown-it` consumer in the tree (`npm why markdown-it`), and it was already on `14.1.1` — there is no `markdown-it@13` anywhere — so the global override didn't actually force a cross-major bump. But scoping it is safer and future-proof (a future dep pinning v13 would no longer be silently bumped), and it matches the js-yaml approach. Verified `npm audit --omit=dev` stays at 0 and the full build passes.

package-lock.json
zapolsky: Fixed via `npm prune` — the lockfile is back to 0 extraneous (force-pushed).

One clarification on the cause: it isn't from removing the `happy-dom` override. A clean `rm -rf node_modules package-lock.json && npm install` on **main's unmodified** `package.json` reproduces the exact same 6 extraneous entries under npm 11.7.0 — main's committed lockfile (0 extraneous) was simply generated with an older npm. `chokidar@4` stays hoisted at the root for `@nestjs/cli` / `@angular-devkit/core` / `@swc/cli`; the nested copies under the schematics packages were redundant dedupes, which is why `npm prune` removes them cleanly without affecting the build.

romandubovyi (approved):

PR #1336, opened 2026-06-17
fix: publish owox declaration files

Pugsin:
## Summary
- include generated `dist/**/*.d.ts` files in the `owox` package publish allowlist
- keep existing JS, bin, manifest, and shrinkwrap publish entries unchanged

## Why
The package advertises `types: "dist/index.d.ts"` and the TypeScript config emits declarations, but the published `owox@0.27.1` tarball only includes `dist/**/*.js` because the package `files` allowlist excludes `.d.ts` files. TypeScript consumers resolve a missing declaration entrypoint.

## Validation
- `npm pack owox@0.27.1 --silent` confirmed `package/dist/index.js` exists and `package/dist/index.d.ts` is absent
- `node -e "const p=require('./apps/owox/package.json'); if(!p.files.includes('./dist/**/*.d.ts')) process.exit(1)"`
- `git diff --check`

### DIFF
--- a/apps/owox/package.json
+++ b/apps/owox/package.json
@@ -46,7 +46,13 @@
   "engines": {
     "node": ">=22.16.0"
   },
-  "files": ["./bin", "./dist/**/*.js", "./oclif.manifest.json", "./npm-shrinkwrap.json"],
+  "files": [
+    "./bin",
+    "./dist/**/*.js",
+    "./dist/**/*.d.ts",
+    "./oclif.manifest.json",
+    "./npm-shrinkwrap.json"
+  ],
   "homepage": "https://github.com/OWOX/owox-data-marts",
   "keywords": [
     "data-marts",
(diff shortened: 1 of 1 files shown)


max-voloshyn:
Thanks for catching this. The inconsistency is real: `owox` advertised `types: "dist/index.d.ts"` but the published package did not include declaration files.

After reviewing the package contract, I think the better root-cause fix is slightly different from this PR. `owox` is intended to be an app-level CLI package, not a supported importable library, so we should stop advertising `import "owox"` rather than publish declarations for that accidental surface.

I opened #1389 to handle that by removing the `main`/`types` entrypoints and the unused `src/index.ts` re-export, while keeping the CLI `bin` entrypoint unchanged.

So I’m not planning to merge this PR as-is, but the report was valid and helped identify the cleaner fix. Thanks for raising it.

PR #1337, opened 2026-06-17
chore(deps): remove dev-only vulnerable js-yaml@3 to clear audit (#166)

zapolsky:
## Summary
Closes the last open Dependabot alert (#166, GHSA-h67p-54hq-rp68 — js-yaml quadratic-complexity DoS) by removing the three **dev-only** `js-yaml@3.14.2` copies that kept it open after #1335. The earlier PR patched the production-reachable copy; this one clears the remaining dev tooling.

Why they couldn't be patched in #1335: there is no patched `3.x` release (the fix is only in `4.2.0`), and a blanket override to `^4.2.0` breaks any consumer using the removed `safeLoad`/`safeDump`. So each path is handled individually.

## Changes
- **read-yaml-file** → override to `^2.1.0`. v2.1.0 is the last CommonJS release and depends on `js-yaml@4`, so `@manypkg/get-packages` (→ changesets) no longer pulls `js-yaml@3`. v3 is ESM-only and would break `@manypkg`'s `require`, so it is intentionally avoided.
- **@istanbuljs/load-nyc-config** → scope `js-yaml` to `^4.2.0`. It calls `js-yaml`'s `.load()`, which is unchanged in v4 (only `safeLoad`/`safeDump` were removed), so this is safe.
- **gray-matter** → dropped from `@owox/docs`. Its only use was front-matter parse/stringify in `scripts/sync-docs.js`; gray-matter@4 pins `js-yaml@3` (via `safeLoad`/`safeDump`) and has no js-yaml@4 release. Replaced with a small parse/stringify backed by the `js-yaml@4` already imported in that script. The generated front matter format is unchanged.

## Result
- `npm audit` — **0 vulnerabilities** (both `--omit=dev` and full).
- No vulnerable `js-yaml` left in the tree (only `4.2.0`).

## Validation
- Full workspace `npm run build` — green (incl. the docs build through the rewritten sync script; generated front matter verified).
- `@manypkg/get-packages` discovers all 16 packages with `read-yaml-file@2.1.0`.
- `jest --coverage` runs clean (exercises `@istanbuljs/load-nyc-config` with `js-yaml@4`).
- 0 extraneous entries in the lockfile.

No changeset: changes are limited to dev tooling / the (private) docs build; no published package behaviour changes.

### DIFF
--- a/apps/docs/scripts/sync-docs.js
+++ b/apps/docs/scripts/sync-docs.js
@@ -4,7 +4,6 @@ import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { glob } from 'glob';
 import yaml from 'js-yaml';
-import matter from 'gray-matter';
 import Papa from 'papaparse';
 import { getConfig } from './env-config.js';
 import {
@@ -12,6 +11,35 @@ import {
   normalizePathToKebabCase,
   normalizeSuffixForDirectoryStyleURL,
 } from './utils.js';
+
+const FRONT_MATTER_RE = /^\uFEFF?---\r?\n([\s\S]*?)\r?\n---\r?\n?/;
+
+/**
+ * Parses YAML front matter from markdown content. Drop-in for the gray-matter
+ * subset used here, backed by js-yaml v4 (which exposes `load`/`dump`, not the
+ * removed `safeLoad`/`safeDump`).
+ * @param {string} content - Markdown content, optionally prefixed with front matter
+ * @returns {{ data: Record<string, any>, content: string }} Parsed front matter and the body
+ */
+function matter(content) {
+  const match = FRONT_MATTER_RE.exec(content);
+  if (!match) return { data: {}, content };
+  const data = yaml.load(match[1]) ?? {};
+  return { data: typeof data === 'object' ? data : {}, content: content.slice(match[0].length) };
+}
+
+/**
+ * Serializes a front matter object onto markdown content, replacing any existing
+ * front matter. Mirrors `gray-matter`'s `stringify` output for our usage.
+ * @param {string} content - Markdown content, optionally prefixed with front matter
+ * @param {Record<string, any>} data - Front matter object to serialize
+ * @returns {string} Markdown content with the serialized front matter prepended
+ */
+matter.stringify = function stringifyFrontmatter(content, data) {
+  const body = matter(content).content;
+  const frontMatter = yaml.dump(data, { lineWidth: -1 }).trimEnd();
+  return `---\n${frontMatter}\n---\n${body.endsWith('\n') ? body : `${body}\n`}`;
+};
 
 // CONFIGURATION
 const APP_LOCATION = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
--- a/package.json
+++ b/package.json
@@ -101,6 +101,10 @@
     "thrift": "^0.23.0",
     "uuid": "^11.1.1",
     "esbuild": "^0.28.1",
+    "read-yaml-file": "^2.1.0",
+    "@istanbuljs/load-nyc-config": {
+      "js-yaml": "^4.2.0"
+    },
     "@nestjs/swagger": {
       "js-yaml": "^4.2.0"
     },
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -101,6 +101,11 @@
     "thrift": "^0.23.0",
     "uuid": "^11.1.1",
     "esbuild": "^0.28.1",
+    "multer": "^2.2.0",
```
romandubovyi, line 104: 🟡 **Important:** `multer@^2.2.0` here (plus the broad `@aws-sdk`/`@smithy`/`@google-cloud` bumps in `package-lock.json`) is unrelated to the js-yaml audit fix and isn't mentioned in the PR description. Looks like a full lockfile refresh rather than a targeted change. Could you either split these out or call them out in the description, so the audit fix stays reviewable in isolation?

apps/docs/scripts/sync-docs.js
```diff
@@ -13,6 +12,35 @@ import {
...
+ * subset used here, backed by js-yaml v4 (which exposes `load`/`dump`, not the
+ * removed `safeLoad`/`safeDump`).
+ * @param {string} content - Markdown content, optionally prefixed with front matter
+ * @returns {{ data: Record<string, any>, content: string }} Parsed front matter and the body
+ */
+function matter(content) {
+  const match = FRONT_MATTER_RE.exec(content);
+  if (!match) return { data: {}, content };
+  const data = yaml.load(match[1]) ?? {};
+  return { data: typeof data === 'object' ? data : {}, content: content.slice(match[0].length) };
```
romandubovyi, line 28: 💡 **Suggestion:** `typeof data === 'object'` is also true for arrays, so YAML list front matter would be returned as an array (gray-matter coerces non-plain-objects to `{}`). Consider tightening to a plain-object check, e.g. `data && typeof data === 'object' && !Array.isArray(data) ? data : {}`, so downstream `frontmatter.title = …` always writes to an object.

apps/docs/scripts/sync-docs.js
```diff
@@ -13,6 +12,35 @@ import {
...
+/**
+ * Serializes a front matter object onto markdown content, replacing any existing
+ * front matter. Mirrors `gray-matter`'s `stringify` output for our usage.
+ * @param {string} content - Markdown content, optionally prefixed with front matter
+ * @param {Record<string, any>} data - Front matter object to serialize
+ * @returns {string} Markdown content with the serialized front matter prepended
+ */
+matter.stringify = function stringifyFrontmatter(content, data) {
+  const body = matter(content).content;
+  const frontMatter = yaml.dump(data, { lineWidth: -1 }).trimEnd();
```
romandubovyi, line 40: 🔵 **Question:** `{ lineWidth: -1 }` disables wrapping, whereas gray-matter relied on js-yaml's default (80). For long `description`/`metaTitle` values, did you confirm the generated front matter stays byte-identical to the previous output? (Disabling wrapping is arguably the better choice — just want to make sure it matches the "format unchanged" claim.)

apps/docs/scripts/sync-docs.js
```diff
@@ -13,6 +12,35 @@ import {
   normalizeSuffixForDirectoryStyleURL,
 } from './utils.js';
 
+const FRONT_MATTER_RE = /^\uFEFF?---\r?\n([\s\S]*?)\r?\n---\r?\n?/;
```
romandubovyi, line 15: ⚪ **Nit:** The BOM is only stripped when front matter is present (the regex requires `---`). A file that starts with a BOM but has no front matter keeps the BOM in its body, whereas gray-matter stripped it unconditionally. Almost certainly irrelevant for these docs, just noting the behavioral delta.

romandubovyi (approved):
Approving ✅

The core fix is clean and well-reasoned — each remaining dev-only `js-yaml@3` path is handled individually rather than with a blanket override, and the validation is thorough (`npm audit` 0 vulnerabilities, full build green, `jest --coverage` clean). No blockers — left a few non-blocking notes inline. Nice work clearing the last alert. 🎉

PR #1338, opened 2026-06-17
feat: record executed SQL on report runs and show it in Run History

zapolsky:
## Summary

Report runs now capture the exact SQL that ran — output controls (filters, sorting, limit) applied and filter parameter values inlined as literals, identical to the generated-SQL preview — recorded on `reportDefinition.executionSqlQuery` and surfaced in Run History as a dedicated **Executed SQL** block.

Previously the run stored only the raw Data Mart query, so the recorded SQL did not reflect the parameters used at run time.

- Extracted a shared `ReportSqlComposerService.inlineStaticSql(...)` from `composeStatic`, so the run-history SQL is byte-identical to the generated-SQL preview.
- Wired into every run-creating path: push destinations (Google Sheets, Email, Slack, Google Chat, MS Teams) via `RunReportService.executeReport`, and Looker Studio (streaming + non-streaming) via `ReportDataCacheService` + `LookerStudioConnectorApiService` (deduped into one helper, set before the read so it persists on success **and** failure).
- Populated **only** when the run applied output controls or blended SQL.
- Frontend: a dedicated "Executed SQL" block in the Run History Configuration view, rendered as plain-text SQL.
- HTTP Data API path intentionally excluded (its run has no `reportDefinition`).

No DB migration (additive optional field on the existing `reportDefinition` JSON column) and no backfill — the field appears only on new runs.

## Test Plan

- [ ] Backend unit: `inlineStaticSql` (Athena/BigQuery inlining, no-params passthrough, unsupported-dialect throw); push + Looker paths set `executionSqlQuery` when controls present and omit it otherwise; failure-path persistence; blended-with-params consistency.
- [ ] Frontend unit: Executed SQL block renders when present, omitted when absent, not duplicated in other blocks.
- [ ] Manual: run a scheduled/manual report with a date filter → Run History → Configuration tab shows the **Executed SQL** block with the filter inlined.
- [ ] Manual: run a report with no output controls → no Executed SQL block.
- [ ] Manual: Looker Studio fetch with a filter → run record carries the executed SQL.

Verified locally: `nest build` ✅, web `tsc -b` + vite ✅, affected backend suites (89 tests) ✅, web ConfigurationView ✅.

### DIFF
--- a/apps/backend/src/data-marts/use-cases/run-report.service.ts
+++ b/apps/backend/src/data-marts/use-cases/run-report.service.ts
@@ -20,6 +20,7 @@ import { DataStorageReportReader } from '../data-storage-types/interfaces/data-s
 import { RunReportCommand } from '../dto/domain/run-report.command';
 import { RunType } from '../../common/scheduler/shared/types';
 import { DataMart } from '../entities/data-mart.entity';
+import { DataMartRun } from '../entities/data-mart-run.entity';
 import { Report } from '../entities/report.entity';
 import { ReportRun } from '../models/report-run.model';
 import { logBlendedSqlIfNeeded } from '../report-run-logging/log-blended-sql';
@@ -212,7 +213,8 @@ export class RunReportService {
     report: Report,
     accessor: BlendableSchemaAccessor,
     signal?: AbortSignal,
-    reportRunLogger?: ReportRunLogger
+    reportRunLogger?: ReportRunLogger,
+    dataMartRun?: DataMartRun
   ): Promise<ReportWriteFinalizeResult | undefined> {
     signal?.throwIfAborted();
     const { dataMart, dataDestination } = report;
@@ -253,6 +255,19 @@ export class RunReportService {
         const composed = await this.reportSqlComposerService.compose(report, accessor);
         sqlOverride = composed.sql;
         sqlOverrideParams = composed.params;
+      }
+
+      // Persist the exact executed SQL (output controls applied, params inlined as
+      // literals — same render as the generated-SQL preview) onto the run record so
+      // Run History can show it. Only when an override exists (output controls or
+      // blending); a plain report's executed SQL == the raw definition sqlQuery.
+      if (sqlOverride && dataMartRun?.reportDefinition) {
+        dataMartRun.reportDefinition.executionSqlQuery =
+          this.reportSqlComposerService.inlineStaticSql(
+            dataMart.storage.type,
+            sqlOverride,
+            sqlOverrideParams
+          );
       }
 
       const reportDataDescription = await reportReader.prepareReportData(report, {
@@ -332,7 +347,8 @@ export class RunReportService {
         reportRun.getReport(),
         accessor,
         signal,
-        reportRunLogger
+        reportRunLogger,
+        reportRun.getDataMartRun()
       );
       const { logs, errors } = reportRunLogger.asArrays();
       await this.handleReportRunSuccess(reportRun, logs, errors, finalizeResult);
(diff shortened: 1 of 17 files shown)


apps/backend/src/data-marts/use-cases/run-report.service.ts
```diff
@@ -255,6 +257,19 @@ export class RunReportService {
         sqlOverrideParams = composed.params;
       }
 
+      // Persist the exact executed SQL (output controls applied, params inlined as
+      // literals — same render as the generated-SQL preview) onto the run record so
+      // Run History can show it. Only when an override exists (output controls or
+      // blending); a plain report's executed SQL == the raw definition sqlQuery.
+      if (sqlOverride && dataMartRun?.reportDefinition) {
+        dataMartRun.reportDefinition.executionSqlQuery =
+          this.reportSqlComposerService.inlineStaticSql(
```
romandubovyi: 🟡 **Important:** Recording `executionSqlQuery` runs `inlineStaticSql(...)` before `prepareReportData`, and that call throws `BusinessViolationException` for a dialect that produces bound params but has no inliner. This makes a display-only field able to fail an otherwise-successful run. It is unreachable today (only Athena/BigQuery produce params, both have inliners), but it is a latent coupling — a future output-controls dialect without an inliner would break real runs, not just the SQL preview. Consider computing this best-effort (try/catch + log) so run-history metadata can never abort the run. Same applies to the Looker path in `ReportDataCacheService.executeGetOrCreateOperation`.

apps/backend/src/data-marts/use-cases/run-report.service.ts
```diff
@@ -255,6 +257,19 @@ export class RunReportService {
         sqlOverrideParams = composed.params;
       }
 
+      // Persist the exact executed SQL (output controls applied, params inlined as
+      // literals — same render as the generated-SQL preview) onto the run record so
+      // Run History can show it. Only when an override exists (output controls or
+      // blending); a plain report's executed SQL == the raw definition sqlQuery.
+      if (sqlOverride && dataMartRun?.reportDefinition) {
```
romandubovyi, line 263: 💡 **Suggestion:** The output-controls check just above (`filterConfig`/`sortConfig`/`limitConfig`, lines ~247-249) is duplicated inline here and as `hasOutputControls` helpers in `ReportDataCacheService` and `ReportSqlComposerService.compose`. Since this PR builds on top of it, it might be worth extracting one shared helper (e.g. on `ReportLike`) so the three copies do not drift. Non-blocking.

apps/backend/src/data-marts/use-cases/run-report.service.ts
zapolsky: Fixed in 7e40e0e82. `executionSqlQuery` is now recorded best-effort — try/catch + `logger.warn` on both the push path (`RunReportService.executeReport`) and the Looker path (`ReportDataCacheService.executeGetOrCreateOperation`) — so a failure to render this display-only field can never abort the run/fetch. Added a test on each path asserting the run proceeds and the field stays unset when `inlineStaticSql` throws.

apps/backend/src/data-marts/use-cases/run-report.service.ts
zapolsky: Done in 6c7f72c4d. Extracted `hasOutputControls(report: ReportLike)` into `report-like-read-plan.ts` and replaced the four report-shaped copies (`RunReportService`, `ReportDataCacheService`, `ReportSqlComposerService.compose`, `DataMartRunService.createReportRunFromReport`). The query-builder / validator / http-data copies operate on a different (post-mapping `options`/`args`) shape, so they're left as-is.

romandubovyi (approved):
Looks good. The previously raised concerns have been addressed, and I found no new issues in the latest revision.

PR #1341, opened 2026-06-18
 fix(backend): optimize project-wide list queries

romandubovyi:
## Summary

- reduce MySQL sort memory usage for project-wide Runs, Reports, Insights, and Triggers
- split pagination into a narrow ID query followed by full entity loading
- preserve the requested ordering after loading full entities
- handle unknown report statuses without crashing the Reports page
- fall back to user labels when avatar images fail to load

## Problem

Project-wide lists could fail with `Out of sort memory` because MySQL sorted wide rows containing JSON fields and joined entities.

The Reports page could also crash when it encountered an unknown status or an unavailable avatar image.

## Solution

Project-wide list queries now:

1. select, sort, and paginate only entity IDs
2. load complete entities and required relations in a separate query
3. restore the original page order in application code

Report rendering now provides fallbacks for unknown statuses and failed avatar requests.

## Testing

- Backend: 5 test suites, 22 tests passed
- Web Reports page: 25 tests passed

### DIFF
--- a/apps/backend/src/data-marts/services/insight-template.service.ts
+++ b/apps/backend/src/data-marts/services/insight-template.service.ts
@@ -85,18 +85,13 @@ export class InsightTemplateService {
   async listVisibleByProject(
     options: ListVisibleProjectInsightTemplatesOptions
   ): Promise<InsightTemplate[]> {
-    const qb = this.repository
+    const pageQb = this.repository
       .createQueryBuilder('insightTemplate')
-      .innerJoinAndSelect('insightTemplate.dataMart', 'dataMart')
-      .leftJoinAndSelect('insightTemplate.sourceEntities', 'sourceEntities')
+      .innerJoin('insightTemplate.dataMart', 'dataMart')
       .where('dataMart.projectId = :projectId', { projectId: options.projectId })
-      .andWhere('dataMart.deletedAt IS NULL')
-      .orderBy('insightTemplate.modifiedAt', 'DESC')
-      .addOrderBy('insightTemplate.id', 'DESC')
-      .take(options.limit ?? 100)
-      .skip(options.offset ?? 0);
+      .andWhere('dataMart.deletedAt IS NULL');
 
-    applyDataMartVisibilityFilter(qb, {
+    applyDataMartVisibilityFilter(pageQb, {
       dataMartAlias: 'dataMart',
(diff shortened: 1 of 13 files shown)


apps/web/src/shared/components/UserAvatar/UserAvatar.tsx
```diff
@@ -11,6 +12,7 @@ export function UserAvatar({
   displayName: string;
   size?: UserAvatarSize;
 }) {
+  const [hasAvatarError, setHasAvatarError] = useState(false);
```
kalnik-a-a: Non-blocking: after `onError` sets `hasAvatarError` to true, this component will not try a new `avatar` prop while the same component instance stays mounted. That can leave initials visible even after a valid URL arrives. Consider resetting this state when `avatar` changes.

kalnik-a-a (approved):
Approved. I left one non-blocking note about the avatar fallback state.

apps/backend/src/data-marts/services/insight-template.service.ts
```diff
@@ -85,26 +85,47 @@ export class InsightTemplateService {
...
+    const insightTemplateIds = page.map(({ id }) => id);
+    if (insightTemplateIds.length === 0) {
+      return [];
+    }
+
+    const insightTemplates = await this.repository
+      .createQueryBuilder('insightTemplate')
+      .innerJoin('insightTemplate.dataMart', 'dataMart')
+      .leftJoinAndSelect('insightTemplate.sourceEntities', 'sourceEntities')
+      .select(['insightTemplate', 'dataMart.id', 'dataMart.title'])
```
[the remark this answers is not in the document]
romandubovyi: Fixed in b545fa426. The row query now explicitly selects `sourceEntities`, so TypeORM hydrates the relation and `sourcesCount` remains accurate. I also added a focused assertion covering the selected relation.

apps/web/src/shared/components/UserAvatar/UserAvatar.tsx
romandubovyi: Fixed in b545fa426. The component now tracks the specific failed avatar URL, so a changed `avatar` prop is attempted immediately. Added a regression test covering an error followed by rerendering with a new URL.

PR #1342, opened 2026-06-18
fix(connectors): defer incremental checkpoint until all accounts succeed

prokopetsv:
# Problem

Two related bugs caused an incorrect state after a partial connector run:

1. **Checkpoint advanced prematurely**: For connectors with multiple accounts/advertisers (MicrosoftAds, CriteoAds, GoogleAds, RedditAds, XAds), the incremental `lastRequestedDate` checkpoint was updated after each processed day *per account*. If account 2 failed mid-run, the checkpoint had already advanced past dates that account 2 never finished fetching, so those dates were silently skipped on the next run.

2. **Executor treated any non-error status as success**: The `ConnectorExecutorService` set `success = true` for any `STATUS` message that wasn't `ERROR` (including `IMPORT_IN_PROGRESS`). A connector process that exited cleanly after emitting only an in-progress status was incorrectly recorded as a successful run.

# Solution

**Connector fix**: Introduced an `updateRequestedDate` flag on `processNode` / `processTimeSeriesNode` / `processAsyncTimeSeriesNode`. `startImportProcess` passes `updateRequestedDate: false` for all per-node calls, collects the maximum processed date across all accounts and nodes, and writes a single checkpoint only after all iterations complete successfully. If any account throws, the checkpoint is never advanced.

**Executor fix**: Changed the `STATUS` message handler to only set `success = true` on `IMPORT_DONE` (status `3`). Intermediate statuses (e.g. `IMPORT_IN_PROGRESS`) are now logged without marking the run successful. Added a fallback: if `spawnConnector` resolves without `success = true` and no explicit error was recorded, a synthetic error is injected so the run is correctly marked as failed.

The trade-off for the connector fix is that on a partial failure, already-completed accounts will be re-fetched on the next run (instead of only the failed account resuming). This is acceptable because writes are idempotent, and correctness of the checkpoint outweighs redundant re-work.

# Changes

- **MicrosoftAds, CriteoAds, GoogleAds, RedditAds, XAds connectors**: Added `updateRequestedDate` parameter (default `true`) to `processNode`, `processTimeSeriesNode`, and `processAsyncTimeSeriesNode`
- **All 5 connectors**: `startImportProcess` now passes `updateRequestedDate: false`, tracks `lastProcessedDate` as the max across all iterations, and writes the checkpoint once at the end
- **`connector-executor.service.ts`**: Replaced unconditional `success = true` for non-error statuses with an explicit check against `IMPORT_DONE`; added `isSuccessfulConnectorStatus` private method
- **`connector-executor.service.ts`**: Added fallback after `spawnConnector` resolves: if `success` is still `false` and no errors were recorded, injects a synthetic error message
- **`connector-executor.service.spec.ts`**: Updated mocked `EXECUTION_STATUS` enum to use numeric values matching the real enum (`IMPORT_IN_PROGRESS: 1`, `IMPORT_DONE: 3`, `ERROR: 5`)
- **`connector-executor.service.spec.ts`**: Added `emitInProgressMessage` helper and a new test asserting that a run emitting only `IMPORT_IN_PROGRESS` is marked as failed

### DIFF
--- a/.changeset/connector-partial-run-checkpoint.md
+++ b/.changeset/connector-partial-run-checkpoint.md
@@ -1,0 +1,7 @@
+---
+"owox": minor
+---
+
+# Reliable incremental sync after partial connector run
+
+Previously, if a connector run failed partway through multiple accounts, the incremental date checkpoint could advance past data that was never fully fetched — causing those dates to be silently skipped on the next run. Now the checkpoint is written only after all accounts complete successfully, so a failed run always retries from the correct starting point. This affects MicrosoftAds, CriteoAds, GoogleAds, RedditAds, and XAds connectors.
--- a/apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
+++ b/apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
@@ -7,8 +7,9 @@ jest.mock('@owox/connectors', () => ({
       Object.assign(this as object, data);
     }),
     EXECUTION_STATUS: {
-      ERROR: 'ERROR',
-      SUCCESS: 'SUCCESS',
+      IMPORT_IN_PROGRESS: 1,
+      IMPORT_DONE: 3,
+      ERROR: 5,
     },
   },
 }));
@@ -38,7 +39,7 @@ describe('ConnectorExecutorService', () => {
       findOne: jest.fn().mockResolvedValue(null),
     } as unknown as Repository<DataMartRun>;
 
-    // Capture the onMessage callback so spawnConnector can invoke it with a success message
+    // Capture the onMessage callback so spawnConnector can emit connector status messages.
     let capturedOnMessage: ((msg: unknown) => void) | null = null;
 
     const outputCaptureService = {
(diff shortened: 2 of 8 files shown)


zapolsky:

Reviewed against a full worktree checkout at `bfe4c25` with parallel specialist agents (reviewer, logic-hunter, architect, tester, fresh-eyes generalist). Both fixes trace cleanly:

- **Checkpoint deferral** is applied consistently across all 5 connectors. The per-connector max-date is computed correctly (Date-vs-Date comparisons, null guards, GoogleAds' extra `customerIds` level, XAds' async path all check out). On full success the single deferred `updateLastRequstedDate` write equals the old per-day final write; on a throw nothing is written. Within-run date computation is unaffected because `NodeJsConfig.updateLastRequstedDate` only logs (it never mutated in-memory state). Re-fetch on partial failure is safe — `GoogleBigQueryStorage.saveData` is an idempotent `MERGE … ON uniqueKeys` upsert.
- **Executor success-gating**: every connector reaches the terminal `IMPORT_DONE` (3) via the shared `AbstractConnector.run()` (no source overrides `run()`), so gating on it is safe across *all* connectors, not just the 5 changed. `isInProgress()` is always `false` in `NodeJsConfig`, so the early-return-without-`IMPORT_DONE` branch is dead in backend runs. Graceful-shutdown / cancellation still resolve to `INTERRUPTED` / `CANCELLED` (not `FAILED`). The synthetic-error fallback is correctly guarded by `configErrors.length === 0`.

CI-parity: `tsc -b` and `eslint` on the changed files are clean; `connector-executor.service.spec.ts` passes (12/12, run in the worktree). Pre-existing `tsc` errors exist only in untouched files (`project-member-api-keys/*` specs, `permissions-*.e2e`, `report-operate-access.e2e`) — **not attributable to this PR**.

- **Question — does the executor need newline-buffered stdout now?** The success gate now hinges on the `IMPORT_DONE` (3) status line being parsed cleanly (`connector-executor.service.ts:272` + `isSuccessfulConnectorStatus` at `:395-397`); if it's never observed, the new fallback marks the run failed (`:349-364`). But child stdout is read **without cross-chunk line buffering** — `connector-process-spawner.service.ts:75-79` hands each raw `data` chunk straight to the capture, and `connector-output-capture.service.ts:45-52` only splits on `\n`/`}{` per call (no buffer carried across calls). If the `IMPORT_DONE` JSON line is ever split across two pipe `data` events, both halves fail `JSON.parse` → both become `UNKNOWN` → the run is marked failed despite a fully successful import. Previously this was masked because the old code set `success=true` on *any* non-ERROR status (including the early `IMPORT_IN_PROGRESS`), so a mangled `IMPORT_DONE` was harmless; this change narrows success to that one specific line. The buffering gap is pre-existing and the failure mode is safe (data is already persisted via `saveData`, and the checkpoint either advanced correctly or not at all → idempotent re-fetch next run — never a silent data skip), and a split of a small `console.log` line is low-probability. Flagging so the trade-off is conscious: is it worth buffering stdout to whole lines before parsing, or is "spurious failed → idempotent retry" acceptable? Not a blocker either way.

<sub>Automated OWOX review · worktree multi-agent passes · all candidate blockers adversarially verified.</sub>

zapolsky (approved):

PR #1343, opened 2026-06-18
fix(backend): exclude disconnected fields from AI Helper sample query

laskevych:
## What changed

AI Helper alias/description generation (`POST /api/data-marts/:id/ai-helper/triggers/:triggerId`) failed when a data mart's **Output Schema** contained a field with status **Disconnected** ("Field is not connected to the data source").

`GenerateDataMartMetadataOrchestratorService` built the 30-row sample query from **every** schema field:

```ts
const schemaColumns = schema.fields.map(f => f.name);
```

A `DISCONNECTED` field exists in the saved schema but not in the underlying table/view, so the storage rejected the whole query. On BigQuery:

```
Unrecognized name: impressions at [1:41]
SELECT `campaign`, `country`, `clicks`, `impressions` FROM `...view...` LIMIT 30
```

This PR filters disconnected fields out of the sample query, and skips the sample fetch entirely when no connected fields remain (which would otherwise emit an empty `SELECT ... FROM`):

```ts
const schemaColumns = schema.fields
  .filter(f => f.status !== DataMartSchemaFieldStatus.DISCONNECTED)
  .map(f => f.name);
```

## Why

The `DISCONNECTED` status exists to mark schema fields no longer present in the data source. The sample query is the one place that selects raw columns from the live view, so it must respect that status — every other schema consumer already excludes disconnected fields (e.g. `collectSchemaFieldPaths`, the report header generators).

`isHiddenForReporting` fields are **intentionally kept**: they are real source columns, and the user may still want the AI to generate an alias/description for them. Unlike the report header/SQL generators, this path is not producing report output, so it does not filter them.

## Testing

- New `generate-data-mart-metadata.orchestrator.service.spec.ts` — 4 cases: disconnected fields excluded, `CONNECTED_WITH_DEFINITION_MISMATCH` kept, hidden-for-reporting kept, and the all-disconnected skip path (no sample fetch; agent receives `sampleColumns: null`).
- `jest src/data-marts/ai-insights/generate-data-mart-metadata.orchestrator.service.spec.ts`: **4 passed**.
- `eslint` + `prettier` clean (also enforced by the pre-commit lint-staged hook).

### DIFF
--- a/apps/backend/src/data-marts/ai-insights/generate-data-mart-metadata.orchestrator.service.ts
+++ b/apps/backend/src/data-marts/ai-insights/generate-data-mart-metadata.orchestrator.service.ts
@@ -5,6 +5,7 @@ import { ToolRegistry } from '../../common/ai-insights/agent/tool-registry';
 import { BusinessViolationException } from '../../common/exceptions/business-violation.exception';
 import { DataMartDefinitionValidatorFacade } from '../data-storage-types/facades/data-mart-definition-validator-facade.service';
 import type { DataMartSchema } from '../data-storage-types/data-mart-schema.type';
+import { DataMartSchemaFieldStatus } from '../data-storage-types/enums/data-mart-schema-field-status.enum';
 import { DataMartDefinitionType } from '../enums/data-mart-definition-type.enum';
 import { DataMartService } from '../services/data-mart.service';
 import { DataMartSampleDataService } from '../services/data-mart-sample-data.service';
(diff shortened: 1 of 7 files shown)


apps/backend/src/data-marts/ai-insights/generate-data-mart-metadata.orchestrator.service.ts
```diff
@@ -73,19 +74,28 @@ export class GenerateDataMartMetadataOrchestratorService {
...
-        METADATA_SAMPLE_ROW_LIMIT
-      );
-      sampleColumns = sample.columns;
-      sampleRows = sample.rows.map(row => this.toQueryRow(sample.columns, row));
-      this.logger.log(
-        `Fetched ${sample.rows.length} sample row(s) across ${sample.columns.length} column(s) for data mart ${request.dataMartId}`
-      );
+      // DISCONNECTED fields aren't in the underlying table/view, so selecting them fails the query.
+      const schemaColumns = schema.fields
+        .filter(f => f.status !== DataMartSchemaFieldStatus.DISCONNECTED)
```
romandubovyi: 🔵 **Question:** This filters only top-level field status, whereas `collectSchemaFieldPathTypes` recurses and prunes DISCONNECTED nested subtrees. Since the sample query selects whole top-level columns, a CONNECTED RECORD containing a DISCONNECTED nested leaf still selects fine — so I think this is correct. Just confirming the non-recursive divergence is intentional and there's no case where a top-level column is CONNECTED but the underlying view dropped it.

romandubovyi: 💡 **Suggestion:** The `status !== DISCONNECTED` predicate now lives both here and in `collectSchemaFieldPathTypes`. A small shared helper (e.g. `isConnected(field)`) in `data-mart-schema.utils.ts` would keep the "field exists in the source" semantics in one place, so a future status that also means "gone from the source" only needs updating once. Non-blocking.

laskevych: Confirmed — the non-recursive filter is intentional.

The sample projects whole top-level columns (`schema.fields.map(f => f.name)` →
`SELECT \`col\``), so nested `a.b` leaves are never projected and there's nothing to
prune below the top level. Using `collectSchemaFieldPaths` here would be wrong — it
emits container + nested-leaf paths (`a`, `a.b`), which would change the projected
column set for RECORD/STRUCT fields. And as you noted, a CONNECTED RECORD with a
DISCONNECTED nested leaf still selects fine, because `SELECT \`record\`` returns
whatever the struct actually contains in the view.

On "CONNECTED top-level but dropped from the view": the merge rules it out at
actualization time. A top-level field is marked DISCONNECTED exactly when it's absent
from the freshly-fetched live schema:
- BigQuery: `updateExistingFields` → `!newField` → `markFieldAsDisconnected`
  (bigquery-schema-merger.ts:57-61)
- Athena/Snowflake/Redshift/Databricks: flat-data-mart-schema-merger.ts:67-68

So `status !== DISCONNECTED` at top level ⟺ the column was present in the last live
fetch. CONNECTED_WITH_DEFINITION_MISMATCH also implies presence (name exists, only
type/mode differs), so keeping it is safe — the SELECT won't hit "Unrecognized name".

One caveat for completeness: the schema only re-fetches past a 30-min TTL
(`actualizeSchemaIfExpired` / `AI_INSIGHTS_SCHEMA_EXPIRES_AFTER_MS`), so a column
dropped within that window could still be cached as CONNECTED. That's not specific to
this path — the actual report-run SELECT is built from the same persisted schema — so
it's a pre-existing staleness property rather than something this change introduces.

laskevych: Good call — done in ad010d77c. Added `isConnected(field)` to data-mart-schema.utils.ts
and routed both this path and `collectSchemaFieldPathTypes` through it, so the
"field exists in the source" check lives in one place. Also added a small unit test
pinning the contract (CONNECTED / CONNECTED_WITH_DEFINITION_MISMATCH → present,
DISCONNECTED → gone).

I scoped the refactor to the two spots called out here. The report-header generators
(`flat-report-headers-generator`, `bigquery-report-headers-generator`) and
`blendable-schema` have the same predicate but some operate on looser field types
(`status?: string`), so I left them for a separate pass rather than widen this fix.

romandubovyi (approved):

PR #1344, opened 2026-06-18
fix(storage): surface expired Google OAuth as a reconnect prompt

prokopetsv:
# Problem

When a data storage's Google OAuth authorization expired or was revoked, the product handled it poorly. Connector runs were created and then failed mid-execution with an opaque BigQuery `401 — Request had invalid authentication credentials`, and schema actualization and definition saves surfaced raw "Failed to refresh OAuth tokens" text. The root cause was twofold: there was no machine-readable signal distinguishing "needs re-authorization" from a generic failure, and nothing validated storage access *before* a run was started — so the user only discovered the broken connection after a run had already failed. The token refresh path also mis-handled Google's `invalid_grant` response and could persist a null access token.

# Solution

Introduced an explicit `OAUTH_REAUTH_REQUIRED` result code that flows from the backend validators through to a dedicated "Reconnect Storage" UI state, and added a pre-run guard so expired authorization is caught before a run is created.

- **Backend signal:** Added `ValidationResultCode.OAUTH_REAUTH_REQUIRED` and `ValidationResult.oauthReauthRequired()`. The BigQuery access validator maps Google auth errors to it, and `CredentialsExpiredException` now returns a clear, user-facing message with a `401`.
- **Robust token refresh:** `google-oauth-flow.service` now detects `invalid_grant` from Google's response body (not just the message), throws `CredentialsExpiredException` on a dead refresh token, separates refresh failures from token-persistence failures, and guards against a missing `access_token`.
- **Pre-run check:** `RunDataMartService.run()` validates storage access before delegating to the connector run. It lives outside the `@Transactional()` boundary (credential resolution can refresh and persist tokens) and is the single chokepoint for both manual and scheduled runs. It blocks **only** on a definitive re-authorization condition — transient failures (network, rate limits, temporary API errors) are logged and the run proceeds, as before.
- **Frontend:** Added a `REAUTH_REQUIRED` health status with a "Reconnect Storage" banner (BigQuery view), invalidated the cached storage health status when an OAuth refresh error is returned, revalidated health on tab focus/visibility, and propagated the new error `code` through the schema-actualize trigger path.

# Changes

- Added `OAUTH_REAUTH_REQUIRED` validation code, `ValidationResult.oauthReauthRequired()`, and exposed it in the validation response DTO.
- Mapped BigQuery OAuth auth errors to the reauth code in `bigquery-access.validator`.
- Reworked `CredentialsExpiredException` / `TokenRefreshFailedException` messages and returned `401` for expired credentials.
- Hardened `refreshTokensByCredentialId`: structured `invalid_grant` detection, separated refresh vs. persistence failures, and added a null `access_token` guard.
- Added a pre-run storage access check in `RunDataMartService` that blocks only on definitive reauth, covering manual and scheduled runs, outside the DB transaction.
- Returned the reauth case from `validate-data-storage-access` and propagated an error `code` through the schema-actualize trigger handler and response DTO.
- Added `REAUTH_REQUIRED` health status, the "Reconnect Storage" banner, and tab focus/visibility revalidation in the web app.
- Invalidated cached storage health on OAuth refresh errors from definition save, publish, and schema-actualize flows.
- Added `storage-oauth-refresh-error.utils` (with tests) and guarded the health-status service against empty storage IDs.
- Added unit coverage in `run-data-mart.service.spec` for reauth-blocking, transient pass-through, and non-OAuth skip paths.

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-access.validator.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-access.validator.ts
@@ -14,6 +14,18 @@ import { DataStorageConfig } from '../../data-storage-config.type';
 import { DataStorageCredentials } from '../../data-storage-credentials.type';
 import { BigQueryApiAdapter } from '../adapters/bigquery-api.adapter';
 
+const GOOGLE_OAUTH_REAUTH_MESSAGE =
+  'Google authorization could not be refreshed. Reconnect this Storage to restore access.';
+
+function isInvalidGoogleAuthError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error);
+
+  return (
+    message.includes('Request had invalid authentication credentials') ||
+    message.includes('Expected OAuth 2 access token')
+  );
+}
+
 @Injectable()
 export class BigQueryAccessValidator implements DataStorageAccessValidator {
   readonly logger = new Logger(BigQueryAccessValidator.name);
@@ -41,6 +53,10 @@ export class BigQueryAccessValidator implements DataStorageAccessValidator {
         return new ValidationResult(true);
       } catch (error) {
         this.logger.warn('OAuth access validation failed', error);
+        if (isInvalidGoogleAuthError(error)) {
+          return ValidationResult.oauthReauthRequired(GOOGLE_OAUTH_REAUTH_MESSAGE);
+        }
+
         const errorMessage = error instanceof Error ? error.message : String(error);
         return new ValidationResult(false, errorMessage);
       }
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-access.validator.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-access.validator.ts
@@ -17,10 +17,34 @@ import { BigQueryApiAdapter } from '../adapters/bigquery-api.adapter';
 const GOOGLE_OAUTH_REAUTH_MESSAGE =
   'Google authorization could not be refreshed. Reconnect this Storage to restore access.';
 
+interface GoogleAuthErrorBody {
+  code?: number;
+  response?: {
+    status?: number;
+    data?: {
+      error?: unknown;
+      error_description?: unknown;
+    };
+  };
+}
+
 function isInvalidGoogleAuthError(error: unknown): boolean {
+  const body = error as GoogleAuthErrorBody | null;
+  const httpStatus = body?.code ?? body?.response?.status;
+  const data = body?.response?.data;
+  const googleError = data?.error;
+  const googleErrorDescription = data?.error_description;
   const message = error instanceof Error ? error.message : String(error);
 
   return (
+    // Structured: HTTP 401 from the BigQuery API (invalid or revoked access token)
+    httpStatus === 401 ||
+    // Structured: OAuth token endpoint signals a revoked/expired refresh token
+    googleError === 'invalid_grant' ||
+    (typeof googleErrorDescription === 'string' &&
+      googleErrorDescription.includes('invalid_grant')) ||
+    // Fallback: message-based detection for clients that surface the error differently
+    message.includes('invalid_grant') ||
     message.includes('Request had invalid authentication credentials') ||
     message.includes('Expected OAuth 2 access token')
   );
--- a/apps/backend/src/data-marts/use-cases/run-data-mart.service.spec.ts
+++ b/apps/backend/src/data-marts/use-cases/run-data-mart.service.spec.ts
@@ -29,7 +29,7 @@ describe('RunDataMartService', () => {
       canAccess: jest.fn().mockResolvedValue(true),
     };
     const credentialsResolver = {
-      resolve: jest.fn().mockResolvedValue({ type: 'oauth' }),
+      resolve: jest.fn().mockResolvedValue({ type: 'bigquery_oauth' }),
     };
     const validationFacade = {
       validateAccess: jest.fn().mockResolvedValue({ valid: true }),
@@ -126,7 +126,7 @@ describe('RunDataMartService', () => {
   });
 
   describe('pre-run storage access check', () => {
-    const command = () => new RunDataMartCommand('dm-1', 'proj-1', '', 'scheduled' as never);
+    const command = () => new RunDataMartCommand('dm-1', 'proj-1', '', 'manual' as never);
 
     it('blocks the run when validation requires re-authorization', async () => {
       const { service, validationFacade, connectorExecutionService } = createService();
@@ -172,7 +172,7 @@ describe('RunDataMartService', () => {
       expect(connectorExecutionService.run).toHaveBeenCalled();
     });
 
-    it('skips the check for storage without OAuth credentials', async () => {
+    it('skips the check for storage without a credentialId', async () => {
       const { service, dataMartService, credentialsResolver, connectorExecutionService } =
         createService();
       dataMartService.getByIdAndProjectId.mockResolvedValue({
@@ -184,6 +184,37 @@ describe('RunDataMartService', () => {
 
       expect(result).toBe('run-id-1');
       expect(credentialsResolver.resolve).not.toHaveBeenCalled();
+      expect(connectorExecutionService.run).toHaveBeenCalled();
+    });
+
+    it('skips validateAccess for non-OAuth credentials (e.g. service account)', async () => {
+      const { service, credentialsResolver, validationFacade, connectorExecutionService } =
+        createService();
+      credentialsResolver.resolve.mockResolvedValue({ type: 'service_account' });
+
+      const result = await service.run(command());
+
+      expect(result).toBe('run-id-1');
+      expect(validationFacade.validateAccess).not.toHaveBeenCalled();
+      expect(connectorExecutionService.run).toHaveBeenCalled();
+    });
+
+    it('skips the check for scheduled runs so they still produce a failed-run record', async () => {
+      const { service, validationFacade, credentialsResolver, connectorExecutionService } =
+        createService();
+      // Even when the storage would require re-authorization, a scheduled run must proceed
+      // (and fail during execution) rather than being silently dropped before any record exists.
+      validationFacade.validateAccess.mockResolvedValue({
+        valid: false,
+        code: ValidationResultCode.OAUTH_REAUTH_REQUIRED,
+      });
+
+      const scheduledCommand = new RunDataMartCommand('dm-1', 'proj-1', '', 'scheduled' as never);
+      const result = await service.run(scheduledCommand);
+
+      expect(result).toBe('run-id-1');
+      expect(credentialsResolver.resolve).not.toHaveBeenCalled();
+      expect(validationFacade.validateAccess).not.toHaveBeenCalled();
       expect(connectorExecutionService.run).toHaveBeenCalled();
     });
   });
--- a/apps/backend/src/data-marts/use-cases/run-data-mart.service.spec.ts
+++ b/apps/backend/src/data-marts/use-cases/run-data-mart.service.spec.ts
@@ -2,12 +2,20 @@ import { ForbiddenException } from '@nestjs/common';
 import { RunDataMartService } from './run-data-mart.service';
 import { RunDataMartCommand } from '../dto/domain/run-data-mart.command';
 import { DataMartDefinitionType } from '../enums/data-mart-definition-type.enum';
+import { ValidationResultCode } from '../data-storage-types/interfaces/data-storage-access-validator.interface';
+import { CredentialsExpiredException } from '../exceptions/google-oauth.exceptions';
 
 describe('RunDataMartService', () => {
   const mockDataMart = {
     id: 'dm-1',
     projectId: 'proj-1',
     definitionType: DataMartDefinitionType.CONNECTOR,
+    storage: {
+      id: 'storage-1',
+      type: 'GOOGLE_BIGQUERY',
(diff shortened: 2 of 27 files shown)


apps/web/src/features/data-marts/edit/model/context/DataMartContext.tsx
```diff
@@ -34,12 +34,22 @@ import type {
...
+import { extractApiError, type ApiError } from '../../../../../app/api';
 import type { DataMartSchema } from '../../../shared/types/data-mart-schema.types';
 import toast from 'react-hot-toast';
 import { pushToDataLayer, trackEvent } from '../../../../../utils';
 import { DATA_MART_RUNS_PAGE_SIZE } from '../../constants';
 import { useRefreshSetupProgress } from '../../../../../components/AppSidebar/SetupChecklist/useSetupProgress';
+import { invalidateDataStorageHealthStatus } from '../../../../data-storage/shared/services/data-storage-health-status.service';
+import { isStorageOAuthRefreshError } from '../../../shared/utils/storage-oauth-refresh-error.utils';
+
+function invalidateStorageHealthOnOAuthRefreshError(error: ApiError, storageId?: string): void {
```
romandubovyi, line 46: 🟡 **Important:** The new pre-run guard can return `CREDENTIALS_EXPIRED` from `runDataMart`, but this helper is only used by definition and publish failures. Please also invalidate the current storage in the `runDataMart` catch block; otherwise a cached `VALID` status remains visible after the run is rejected and the reconnect prompt does not appear.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-access.validator.ts
```diff
@@ -14,6 +14,18 @@ import { DataStorageConfig } from '../../data-storage-config.type';
 import { DataStorageCredentials } from '../../data-storage-credentials.type';
 import { BigQueryApiAdapter } from '../adapters/bigquery-api.adapter';
 
+const GOOGLE_OAUTH_REAUTH_MESSAGE =
+  'Google authorization could not be refreshed. Reconnect this Storage to restore access.';
+
+function isInvalidGoogleAuthError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error);
+
+  return (
```
romandubovyi, line 39: 🟡 **Important:** Please also recognize `invalid_grant` here. The OAuth client may refresh automatically during the BigQuery request, particularly when the stored credentials have no `expiry_date`; a revoked refresh token then surfaces as `invalid_grant`, falls through as a generic validation failure, and the pre-run guard proceeds with creating the run. Please cover this path with a validator test as well.

apps/backend/src/data-marts/use-cases/run-data-mart.service.ts
```diff
@@ -36,11 +45,52 @@ export class RunDataMartService {
...
+    try {
+      const credentials = await this.credentialsResolver.resolve(storage);
+      const result = await this.validationFacade.validateAccess(
+        storage.type,
+        storage.config,
+        credentials
+      );
+
+      if (!result.valid && result.code === ValidationResultCode.OAUTH_REAUTH_REQUIRED) {
+        throw new CredentialsExpiredException(storage.id, 'storage');
```
romandubovyi, line 92: 🔵 **Question:** For scheduled runs, this exception is thrown before `connectorExecutionService.run()`, so no run record is ever created — `scheduled-triggers-handler.service.ts` just logs and swallows it. Previously an expired-OAuth scheduled run produced a visible failed run in history. Is silently dropping the scheduled run intended? Without a failed-run record or notification, the only signal is the health indicator, which a user only sees when they open the UI.

apps/backend/src/data-marts/use-cases/run-data-mart.service.ts
```diff
@@ -36,11 +45,52 @@ export class RunDataMartService {
...
+   *
+   * Runs outside any DB transaction (resolving credentials may refresh and persist tokens).
+   * Only a definitive re-authorization condition blocks the run — transient problems
+   * (network blips, rate limits, temporary API/token-endpoint failures) must not prevent a
+   * run from starting, since those are surfaced during execution as they were before.
+   */
+  private async checkStorageAccess(dataMart: DataMart): Promise<void> {
+    const { storage } = dataMart;
+
+    if (!storage.config || !storage.credentialId) {
```
romandubovyi, line 78: 🟡 **Important:** This guard runs a full `validateAccess()` (a live `checkAccess()` round-trip) plus a credential resolve on *every* connector run — including scheduled runs and non-OAuth/service-account storages — yet only `OAUTH_REAUTH_REQUIRED` can ever block. For non-OAuth storages this is pure overhead added to each run, and credentials get resolved twice (here and again in `connectorExecutionService.run()`). Consider scoping the check to OAuth credentials only (e.g. gate on the credential/storage being Google OAuth) to avoid the extra latency and API quota per run.

apps/web/src/features/data-storage/shared/model/hooks/useDataStorageHealthStatus.ts
```diff
@@ -43,11 +44,32 @@ export function useDataStorageHealthStatus(storageId: string): UseDataStorageHea
     setCached(prev => (prev === next ? prev : next));
   }, [readCached]);
 
-  // Auto-fetch on mount / storageId change
+  // Auto-fetch on mount / storageId change (cache-respecting — avoids redundant requests)
   useEffect(() => {
     fetchAndCacheDataStorageHealthStatus(storageId);
```
romandubovyi, line 48: 🟡 **Important:** These `focus`/`visibilitychange` listeners are registered per hook instance, and each one calls `invalidateDataStorageHealthStatus` → `notifySubscribers()`, which re-fetches health for every subscribed hook. With multiple indicators mounted for the same storage (e.g. `DataStorageHealthIndicator` + the new `DataStorageHealthError`), a single tab focus fans out into repeated invalidations and refetches. Consider a single shared listener (module-level) or debouncing the revalidation.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-access.validator.ts
```diff
@@ -14,6 +14,18 @@ import { DataStorageConfig } from '../../data-storage-config.type';
 import { DataStorageCredentials } from '../../data-storage-credentials.type';
 import { BigQueryApiAdapter } from '../adapters/bigquery-api.adapter';
 
+const GOOGLE_OAUTH_REAUTH_MESSAGE =
+  'Google authorization could not be refreshed. Reconnect this Storage to restore access.';
+
+function isInvalidGoogleAuthError(error: unknown): boolean {
```
romandubovyi, line 31: 💡 **Suggestion:** This relies on matching the exact English text of Google's error message, so a wording change upstream would silently break the reauth mapping and the user would get a raw 401 again. Where possible, prefer a structured signal (HTTP status / error code) like `isInvalidRefreshTokenError` in `google-oauth-flow.service` already does, keeping the substring match only as a fallback.

apps/web/src/features/data-marts/shared/utils/storage-oauth-refresh-error.utils.ts
```diff
@@ -0,0 +1,20 @@
...
+export function isStorageOAuthRefreshError(error: StorageOAuthRefreshError): boolean {
+  if (error.code && STORAGE_OAUTH_REFRESH_ERROR_CODES.has(error.code)) {
+    return true;
+  }
+
+  const message = error.message ?? '';
+  return (
+    message.includes('Failed to refresh OAuth tokens') ||
+    message.includes('Google authorization could not be refreshed') ||
+    message.includes('Google access is no longer active') ||
```
romandubovyi: ⚪ **Nit:** `'Google access is no longer active'` doesn't appear to be produced by any backend message (the exceptions emit "Google authorization could not be refreshed…", "Google access could not be refreshed…", and "Google access was refreshed, but…"). If it's not intentionally guarding a known/legacy message, consider dropping it to avoid implying a string that never matches.

apps/backend/src/data-marts/use-cases/run-data-mart.service.spec.ts
```diff
@@ -101,4 +124,67 @@ describe('RunDataMartService', () => {
     expect(accessDecisionService.canAccess).not.toHaveBeenCalled();
     expect(connectorExecutionService.run).toHaveBeenCalled();
   });
```
romandubovyi, line 126: 🟢 **Praise:** Great coverage of the pre-run guard — the reauth-block, transient pass-through, and non-OAuth skip cases are all exercised, and they directly encode the "block only on definitive reauth" contract documented in `checkStorageAccess`. This is exactly the kind of edge-case coverage that prevents regressions here.

romandubovyi:
**Overall:** Solid, well-structured change — clear separation of backend signal / token-refresh hardening / pre-run guard / frontend state, with good test coverage. **No blocking issues found.**

Two things worth aligning on before merge:
- **Scheduled-run visibility:** the pre-run guard throws before a run record is created, and the scheduled trigger handler swallows the error — so an expired-OAuth scheduled run now disappears from history instead of showing as failed. Is that intended? (see inline)
- **Per-run cost:** the guard runs a live `checkAccess()` + a second credential resolve on *every* connector run, including non-OAuth/service-account storages where it can never block. Consider scoping it to Google OAuth credentials. (see inline)

Everything else is suggestions/nits.

apps/web/src/features/data-marts/edit/model/context/DataMartContext.tsx
prokopetsv: In `DataMartContext.tsx`, the `runDataMart` catch block now calls `invalidateStorageHealthOnOAuthRefreshError(apiError, state.dataMart?.storage.id)` — the same line the `updateDataMartDefinition` and `publishDataMart` flows already use — and `state.dataMart?.storage.id` was added to the `useCallback` dependency array to match.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-access.validator.ts
prokopetsv: Both addressed in the latest push:

**1. `runDataMart` health invalidation** — Added `invalidateStorageHealthOnOAuthRefreshError(apiError, state.dataMart?.storage.id)` to the `runDataMart` catch block (and `storage.id` to the dep array), matching the definition/publish flows. A run rejected with `CREDENTIALS_EXPIRED` now clears the cached status, so the indicator re-fetches and the reconnect prompt appears.

**2. `invalid_grant` in the BQ validator** — `isInvalidGoogleAuthError` now also recognizes `invalid_grant`, both in the error message and in a structured Google error body (`response.data.error` / `error_description`), so an on-demand refresh failure during the BigQuery call is classified as reauth-required instead of falling through as a generic failure. Covered with a new `bigquery-access.validator.spec.ts` (invalid_grant via message, via structured body, the invalid-token case, and a non-reauth failure).

apps/backend/src/data-marts/use-cases/run-data-mart.service.ts
prokopetsv: No — that was a bug in the original implementation, now fixed. The pre-run check is gated to `RunType.manual` only:

```typescript
if (command.runType === RunType.manual) {
  await this.checkStorageAccess(dataMart);
}
```

Scheduled runs skip the check entirely and reach `connectorExecutionService.run()` as before, so they still produce a visible failed-run record and trigger notifications. The silent-drop scenario you described no longer applies.

apps/backend/src/data-marts/use-cases/run-data-mart.service.ts
prokopetsv: Fixed — after resolving credentials, we now check `isBigQueryOAuthCredentials()` (the existing guard in `data-storage-credentials.guards.ts`) and return early if the credentials aren't Google OAuth. `validateAccess()` is never called for service accounts or any non-OAuth storage type.

apps/web/src/features/data-storage/shared/model/hooks/useDataStorageHealthStatus.ts
prokopetsv: Update: on reflection I removed the focus/visibilitychange revalidation entirely instead of batching it. It was added by this branch and isn't needed — health status is already revalidated on mount, on storageId change, and via explicit invalidation after a run error/reconnect. That eliminates the fan-out at the source with less code.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-access.validator.ts
prokopetsv: Fixed — `isInvalidGoogleAuthError` now checks `error.code === 401` (the structured HTTP status Google's API client sets on the error object) first, then the `invalid_grant` OAuth body fields, with the English message strings kept only as fallbacks. Added a spec test for the structured 401 path.

apps/web/src/features/data-marts/shared/utils/storage-oauth-refresh-error.utils.ts
prokopetsv: Dropped — confirmed no backend exception produces that string.

apps/backend/src/data-marts/use-cases/run-data-mart.service.spec.ts
prokopetsv: Thank you! The contract felt important to pin down explicitly, given how easy it would be for a future change to accidentally widen the blocking condition.

romandubovyi (approved):

PR #1349, opened 2026-06-19
fix(web): serve connector OAuth callback pages via SPA fallback

prokopetsv:
# Problem

Signing in to any connector that uses OAuth (Microsoft Ads, Google Ads, TikTok, LinkedIn) failed at the final step: after the user authorized with the provider, the browser was redirected to a callback page such as `/oauth/microsoft-ads/callback` and got a NestJS `404 Not Found` (`Cannot GET /oauth/microsoft-ads/callback`) instead of the app. The callback pages are front-end (SPA) routes, but the web server's SPA fallback excluded the entire `/oauth` prefix from serving `index.html`, so every `/oauth/*` request fell through to the backend, which only serves `/api/*`. The broad exclusion was originally added to keep the MCP/IDP OAuth authorization-server endpoints routed to the backend, but it unintentionally swallowed the connector callback pages too. This affected both self-hosted and cloud deployments.

# Solution

Narrow the SPA-fallback exclusion list from the catch-all `/oauth` to only the specific backend OAuth endpoints, so connector callback pages once again fall through to the SPA while MCP/IDP OAuth keeps working.

- Updated `DEFAULT_EXCLUDED_ROUTES` in `apps/owox/src/web/static-config.ts` to exclude `/oauth/authorize`, `/oauth/token`, `/oauth/register`, and `/oauth/jwks` instead of all of `/oauth`. These four mirror the IDP OAuth controllers (`apps/backend/src/idp/oauth/controllers/*`); discovery metadata stays covered by the unchanged `/.well-known` and `/mcp` excludes.
- Connector callback paths (`/oauth/<provider>/callback`) no longer match any exclusion, so they are served `index.html` and React Router can capture the auth code.
- Added a regression test that asserts both directions of the split, so re-broadening to `/oauth` (which breaks connector sign-in) or dropping a backend endpoint (which breaks MCP OAuth) fails CI.

# Changes

- Replaced the `/oauth` catch-all in `DEFAULT_EXCLUDED_ROUTES` with the explicit backend endpoints `/oauth/authorize`, `/oauth/token`, `/oauth/register`, `/oauth/jwks`.
- Added tests in `apps/owox/test/web/static-config.test.ts` verifying MCP/IDP backend OAuth and discovery endpoints are not served the SPA shell, while connector callback pages are.
- Added a changeset (`owox`, minor) describing the restored OAuth connector sign-in.

### DIFF
--- a/apps/owox/src/web/static-config.ts
+++ b/apps/owox/src/web/static-config.ts
@@ -14,7 +14,15 @@ export interface StaticAssetsOptions {
   packageName?: string;
 }
 
-const DEFAULT_EXCLUDED_ROUTES = ['/api', '/.well-known', '/oauth', '/mcp'];
+const DEFAULT_EXCLUDED_ROUTES = [
+  '/api',
+  '/.well-known',
+  '/oauth/authorize',
+  '/oauth/token',
+  '/oauth/register',
+  '/oauth/jwks',
+  '/mcp',
+];
 
 /**
  * Configures Express application to serve static web interface files
--- a/apps/owox/test/web/static-config.test.ts
+++ b/apps/owox/test/web/static-config.test.ts
@@ -143,4 +143,65 @@ describe('setupWebStaticAssets', () => {
       );
     });
   });
+
+  // Guards the split between backend OAuth endpoints (MCP/IDP authorization
+  // server + discovery) and front-end OAuth connector callback pages. Both
+  // live under `/oauth/*`, so the exclude list must be specific enough to keep
+  // the backend endpoints routed to NestJS while letting connector callback
+  // pages fall through to the SPA. Re-broadening the excludes to `/oauth` (the
+  // regression this guards) breaks every connector sign-in; dropping a backend
+  // endpoint from the list breaks MCP OAuth.
+  describe('OAuth route exclusions (MCP/IDP backend vs connector callbacks)', () => {
+    beforeEach(() => {
+      // No backend handlers are registered here, so if the SPA fallback wrongly
+      // catches a backend route it returns index.html (200) instead of a 404.
+      const configured = setupWebStaticAssets(app);
+      expect(configured, '@owox/web must be built for these tests').to.be.true;
+    });
+
+    // Must stay routed to the NestJS backend (never served the SPA shell).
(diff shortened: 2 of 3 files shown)


apps/owox/src/web/static-config.ts
```diff
@@ -14,7 +14,15 @@ export interface StaticAssetsOptions {
   packageName?: string;
 }
 
-const DEFAULT_EXCLUDED_ROUTES = ['/api', '/.well-known', '/oauth', '/mcp'];
+const DEFAULT_EXCLUDED_ROUTES = [
```
romandubovyi, line 17: 💡 **Suggestion:** Since `setupSpaFallback` uses `req.path.startsWith(route)` to check exclusions, adding routes like `/oauth/token` or `/oauth/register` here means that any future connector callback paths starting with those prefixes (e.g., `/oauth/token-manager/callback` or `/oauth/register-me/callback`) will be incorrectly excluded from the SPA fallback. To prevent partial prefix matching, we could update the matching logic in `setupSpaFallback` to check for exact match or boundaries (e.g., `req.path === route || req.path.startsWith(route + /)`).

apps/owox/test/web/static-config.test.ts
```diff
@@ -143,4 +143,65 @@ describe('setupWebStaticAssets', () => {
...
+    beforeEach(() => {
+      // No backend handlers are registered here, so if the SPA fallback wrongly
+      // catches a backend route it returns index.html (200) instead of a 404.
+      const configured = setupWebStaticAssets(app);
+      expect(configured, '@owox/web must be built for these tests').to.be.true;
+    });
+
+    // Must stay routed to the NestJS backend (never served the SPA shell).
+    // Mirrors apps/backend/src/idp/oauth + ee/mcp metadata controllers.
+    const backendOnlyRoutes = [
```
romandubovyi, line 164: 💡 **Suggestion:** We could add a non-existent API path (e.g., `/api/non-existent-endpoint`) to the `backendOnlyRoutes` array. This would explicitly assert that any API requests that do not have a registered backend handler correctly fall through to a 404 instead of being captured by the SPA fallback.

romandubovyi (approved):
LGTM. No blocking issues found, regression tests look solid.

PR #1355, opened 2026-06-24
fix(web): prevent sheet dismissal loop when interacting with confirmation dialog

romandubovyi:
### Description
Clicking elements inside the unsaved-changes `ConfirmationDialog` was trapping the user in an infinite loop. The dialog was rendered as a portal **sibling** of the `Sheet`, so Radix's dismissable layer treated interactions with it as happening "outside" the sheet. The sheet then tried to close and — because the form was dirty — immediately re-opened the dialog, so "No, stay here" just showed it again.

### Solution
Render the `ConfirmationDialog` **inside `SheetContent`** (a React descendant of the sheet's dismissable layer) instead of as a sibling of `Sheet`. Radix then treats it as a nested dismissable layer and no longer considers interactions with it as "outside" the sheet — breaking the loop while keeping Escape / overlay-click / close-button dismissal intact. No event-interception workaround is needed.

### Changes
* Moved `<ConfirmationDialog>` inside `<SheetContent>` in all 9 sheets that use `useUnsavedGuard`.
* Documented the rule in `useUnsavedGuard`'s JSDoc (render the dialog inside `SheetContent`).
* Added unit tests for the dirty-guard flow (`useUnsavedGuard.test.ts`).
* Removed the now-redundant wrapping `<>…</>` fragments left after the move.

### Testing
* `npm run lint` — clean
* `tsc --noEmit` — clean
* `useUnsavedGuard.test.ts` — 4 passing
* Manually verified in the browser on the Storage config sheet: editing a field, closing, and clicking inside the confirmation dialog no longer dismisses the sheet or loops.


apps/web/src/hooks/useUnsavedGuard.ts
```diff
@@ -7,6 +7,11 @@ import { useCallback, useState } from 'react';
  * - intercept close and show confirmation dialog
  * - confirm close and reset state
  * - handle successful submit (reset + close)
+ *
+ * Note: render the unsaved-changes `ConfirmationDialog` *inside* the sheet's `SheetContent`
```
max-voloshyn, line 11: 🔴 Blocking: This adds the Radix nesting rule, but the PR does not apply it to all sheet-hosted confirmation dialogs. `MembershipRequestSheet` still opens its decline confirmation from inside `SheetContent` while rendering `ConfirmationDialog` as a sibling of `Sheet`, so clicking in that confirmation can dismiss/reset the underlying membership-request sheet. Please move that dialog into that sheet content as part of this audit, or explicitly narrow the rule to `useUnsavedGuard` consumers.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/hooks/useUnsavedGuard.test.ts
```diff
@@ -0,0 +1,75 @@
+import { describe, it, expect, vi } from 'vitest';
+import { act, renderHook } from '@testing-library/react';
+import { useUnsavedGuard } from './useUnsavedGuard';
+
+/**
+ * Tests for the "unsaved changes" guard: a dirty form must intercept close and show the
+ * confirmation dialog instead of closing immediately. The re-open loop the dialog used to
```
max-voloshyn: 🟡 Non-blocking: These hook-only tests do not cover the regression described here: the Radix re-open loop depends on rendering `ConfirmationDialog` inside `SheetContent`, but the test never renders a sheet/dialog consumer or verifies the nested dismissable-layer interaction. Moving the dialog back outside `SheetContent` would leave all four tests passing while reintroducing the bug, so this should be covered by at least one component/integration test for a representative consumer.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/hooks/useUnsavedGuard.ts
romandubovyi: Confirmed — `MembershipRequestSheet` had the same sibling-layer bug: its decline `ConfirmationDialog` was a sibling of `Sheet` while opened from inside `SheetContent`, so clicking in it dismissed/reset the underlying sheet. Moved it inside `SheetContent`, matching the pattern used in the other sheets. `AddContextSheet` is left as a sibling since it's a standalone `Sheet`, not a confirmation dialog. lint / `tsc` clean, tests green.

max-voloshyn (approved):

apps/web/src/hooks/useUnsavedGuard.test.ts
romandubovyi: Agreed — the hook-only test didn't guard this regression: it only exercised the hook's state machine, so moving the dialog back outside `SheetContent` would have kept it green while reintroducing the bug. A component test can't cover the real behaviour either, since the suite runs on happy-dom and every sheet/dialog test in this repo mocks Radix `Sheet`/`Dialog` away (portal issues) — with Radix mocked, the dismissable-layer logic the regression depends on isn't exercised regardless of nesting. So I removed that test as misleading rather than keep a false sense of coverage. We'll add proper e2e (Playwright) coverage for these sheet-hosted confirmation dialogs separately — outside happy-dom, against real Radix — as a follow-up.

PR #1357, opened 2026-06-24
feat: add advanced search indexing and MCP tool

kalnik-a-a:
## Summary

This PR adds advanced search as a first-class data-marts capability:

- Backend search API, semantic/keyword indexing, scoring, and reindex scheduling for data marts, data storages, and data destinations.
- Web search entry point, route, service, hook, and result UI without runtime feature flags.
- MCP tool `get_relevant_data_marts_by_prompt` for prompt-based data mart discovery inside the current OWOX project.

The implementation is intentionally placed under the data-marts domain. It does not require the EE module loader or a public-origin URL dependency.

## Backend API

Adds `GET /search` through the data-marts search module.

Request parameters:

- `q` - required natural-language query. It is trimmed, must be non-empty, and is bounded by search query length configuration.
- `limit` - optional result limit. API accepts values from `1` to `50`.
- `entityTypes` - optional list of searchable entity types. Supported values are `DATA_MART`, `DATA_STORAGE`, and `DATA_DESTINATION`.
- `excludeDrafts` - optional boolean to hide draft data marts from the result set.

Response fields per result:

- `entityType` - type of matched entity.
- `entityId` - stable entity id.
- `title` - display title used by the UI and MCP consumers.
- `description` - indexed description text when available.
- `finalScore` - final score after relevance and extendability are combined.
- `kwScore` - keyword component of the match.
- `vecScore` - vector/embedding component of the match.
- `extendability` - extra score based on how useful the entity is for follow-up analysis.

## Search Logic

The search engine combines keyword and semantic candidates:

1. The query is validated and embedded through the configured embedding provider.
2. Candidate rows are loaded from entity-specific index sources.
3. Keyword and vector scores are normalized and blended into a relevance score.
4. Extendability is added to produce `finalScore`.
5. Results below the minimum relevance threshold are filtered out before returning matches.
6. Remaining results are sorted by `finalScore` and capped by the requested limit.

Entity-specific behavior:

- Data marts keep draft status and field count in the index because those values affect filtering and usefulness.
- Data storages and data destinations do not store physical `is_draft` or `field_count` columns because those values are constant for these entity types.
- Storage and destination search documents do not include data mart context, keeping those index rows focused on the entity itself.
- Data mart relationship signals are indexed so first-level linked data marts can be reindexed when relationships change.

## Indexing And Reindexing

The backend adds search index entities and migrations for:

- Data mart search index rows.
- Data storage search index rows.
- Data destination search index rows.
- Data mart relationship search index rows.
- Targeted entity reindex triggers.
- Project-level reindex triggers split by searchable entity type.

Reindexing is split into targeted and project-level flows:

- Targeted entity triggers are created for direct changes to individual entities and relationship changes.
- Targeted entity triggers run every 2 seconds, process up to 2 triggers per batch, and have a 2-hour TTL.
- Project-level triggers are used for broader reconciliation and drift handling.
- Project-level triggers are split by entity type, so data mart indexing does not block storage or destination indexing.
- Project-level handlers process one trigger at a time and are staggered by entity type to limit broad reindex load.
- Data mart project indexing runs at seconds `0,30`, storage project indexing at seconds `10,40`, and destination project indexing at seconds `20,50`.

Covered invalidation paths include:

- Data mart create/update/delete.
- Data mart schema, definition, title, description, and publish changes.
- Data storage create/update/delete.
- Data destination create/update/delete.
- Data mart relationship create/update/delete.
- First-level related data mart reindexing when relationship context changes.

The default drift scheduler runs hourly through `ADVANCED_SEARCH_DRIFT_CRON` unless overridden.

## Configuration

Search API configuration:

- `SEARCH_QUERY_MAX_LENGTH` - maximum accepted query length. Falls back to `ADVANCED_SEARCH_QUERY_MAX_LENGTH`; default is `256`.
- `SEARCH_QUERY_MIN_LENGTH` - minimum accepted query length; default is `2`.
- `SEARCH_TOP_K` - default API result count; default is `25`.

Advanced search/indexing configuration:

- `ADVANCED_SEARCH_MODEL_CACHE_DIR` - optional local model cache directory.
- `ADVANCED_SEARCH_DRIFT_CRON` - cron schedule for drift checks; default is `0 * * * *`.
- `ADVANCED_SEARCH_TOP_K` - default semantic result count; default is `10`.
- `ADVANCED_SEARCH_INDEX_BATCH_SIZE` - indexing batch size; default is `50`.
- `ADVANCED_SEARCH_VECTOR_CANDIDATE_MULTIPLIER` - expands vector candidate pool before final ranking; default is `2`.
- `ADVANCED_SEARCH_MIN_RELEVANCE` - minimum relevance threshold before extendability is applied; default is `45`.
- `ADVANCED_SEARCH_CANDIDATE_LIMIT` - maximum candidate rows considered by a search request; default is `500`.
- `ADVANCED_SEARCH_QUERY_MAX_LENGTH` - maximum advanced-search query length; default is `256`.
- `ADVANCED_SEARCH_EMBEDDING_CONCURRENCY` - concurrent embedding/indexing work; default is `2`.
- `ADVANCED_SEARCH_EMBEDDING_PROVIDER` - embedding provider, `local` or `openrouter`; default is `local`.
- `ADVANCED_SEARCH_EMBEDDING_MODEL` - embedding model name; default is `google/gemini-embedding-2`.
- `ADVANCED_SEARCH_EMBEDDING_DIMENSIONS` - embedding vector dimension. Falls back to `ADVANCED_SEARCH_EMBEDDING_DIM`; default is `768`.
- `ADVANCED_SEARCH_OPENROUTER_API_KEY` - required only when `ADVANCED_SEARCH_EMBEDDING_PROVIDER=openrouter`.
- `ADVANCED_SEARCH_OPENROUTER_ALLOWED_PROVIDERS` - optional comma-separated provider allowlist. Falls back to `AI_ALLOWED_PROVIDERS`.
- `ADVANCED_SEARCH_OPENROUTER_DATA_COLLECTION` - OpenRouter data collection policy; default is `deny`.
- `ADVANCED_SEARCH_OPENROUTER_ZDR` - enables zero-data-retention routing preference; default is `true`.
- `ADVANCED_SEARCH_OPENROUTER_BATCHING_ENABLED` - enables OpenRouter embedding batching; default is `false`.
- `ADVANCED_SEARCH_OPENROUTER_BATCH_SIZE` - OpenRouter embedding batch size; default is `20`.
- `ADVANCED_SEARCH_OPENROUTER_REQUEST_TIMEOUT_MS` - OpenRouter request timeout; default is `60000`.

## MCP Tool

Adds MCP tool `get_relevant_data_marts_by_prompt`.

Input schema:

```json
{
  "prompt": "string"
}
```

Additional accepted parameter:

- `limit` - optional integer from `1` to `25`, defaulting to `10`.

Important contract details:

- `prompt` is required, trimmed, and validated with the same practical query bounds as backend search.
- The tool searches only non-draft data marts.
- The tool uses the authenticated MCP context for project and user scope; there is no `project_id` input parameter.
- The tool returns `relevance_score` instead of a URL so it does not depend on public-origin configuration.
- Legacy or unrelated inputs such as `query` and `project_id` are rejected by the strict schema.

Output shape:

```json
{
  "data_marts": [
    {
      "id": "...",
      "title": "...",
      "description": "...",
      "relevance_score": 87.5
    }
  ]
}
```

## Web UI

Adds the web search experience:

- Sidebar search entry.
- `/search` route.
- Search page, hook, service, and DTOs.
- Query submission, result rendering, loading, empty, and error states.
- Debounced requests and URL query synchronization.
- Result navigation to the matched entity page.
- Keyboard navigation entry through the existing command/search shortcut behavior.
- No runtime flag dependency; the route is wired directly into the app navigation.

## Verification

Local verification run before publishing this draft:

- `npm test -w @owox/backend -- search --runInBand` - 26 suites, 393 tests passed.
- `npm test -w @owox/web -- SearchButton SearchPage useSearch search.service` - 4 files, 42 tests passed.
- `npm run test:e2e -w @owox/backend -- search.e2e-spec.ts --runInBand` - 1 suite, 4 tests passed.
- `npx tsc -p apps/backend/tsconfig.build.json --noEmit` - passed.
- `npm run type-check -w @owox/web` - passed.
- `git diff --check origin/main...HEAD` - passed.


kalnik-a-a:
Review follow-up fixes are included in https://github.com/OWOX/owox-data-marts/commit/8f63df962a20778762e70969c971237c79402f08.

What changed:
- Made search index state project-aware: `projectId` is now part of staleness checks, document signatures, orphan cleanup, and access joins, so moved/cross-project entities do not leave stale or mismatched index rows behind.
- Centralized data mart schema search invalidation in `DataMartService.saveActualizedSchema()`, covering schema actualization from reports, Looker Studio schema reads, and manual actualization while keeping invalidation failures non-blocking.
- Added search reindex scheduling after legacy storage moves for the moved storage and affected first-level data marts in the new project.
- Improved search resilience: if prompt embedding is unavailable, search now falls back to keyword-only scoring instead of returning no results; the local embedding provider also retries initialization after transient load failures.
- Tightened scoring/indexing behavior: keyword scoring now uses the strongest matching slot, negative cosine scores are clamped to zero, embedding failures preserve existing index rows, and project reindex handlers fail/retry when embedding generation reports failures.
- Fixed keyset cursor formatting for SQLite timestamp precision while preserving non-zero milliseconds.
- Removed the obsolete context-loader path and simplified relationship edge loading to fetch only raw ids needed for indexing.
- Added/updated regression coverage around project-aware cleanup, schema invalidation, storage moves, cursor precision, keyword fallback, embedding retries, and scoring behavior.

Validation:
- `npm test -w @owox/backend -- --runInBand ...` for the affected search/report specs passed.
- Full `npm run test` passed.

apps/backend/src/data-marts/search/embedding/local-transformers.provider.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  get modelId(): string {
+    return `local:${EMBEDDING_MODEL}:${EMBEDDING_DTYPE}:${EMBEDDING_DIMENSIONS}`;
+  }
+
+  get dimensions(): number {
+    return EMBEDDING_DIMENSIONS;
+  }
+
+  async embed(texts: string[], _options?: EmbeddingOptions): Promise<(Float32Array | null)[]> {
```
romandubovyi, line 51: ⚪ **Nit:** Generating embeddings locally via ONNX runtime is a CPU-intensive operation. Running this directly on the main thread (even with concurrency limited to 2) can block the Node.js event loop under high search load or during batch reindexing, degrading overall application responsiveness. Consider documenting this CPU-bound risk or moving the embedding generation to a Worker Thread if performance tests indicate bottlenecks.

apps/backend/src/data-marts/search/indexing/search-reindex-trigger-handler.service.ts
```diff
@@ -0,0 +1,187 @@
...
+import { SearchReindexTrigger } from '../../entities/search/search-reindex-trigger.entity';
+import {
+  SearchDataDestinationProjectReindexTrigger,
+  SearchDataMartProjectReindexTrigger,
+  SearchDataStorageProjectReindexTrigger,
+  SearchProjectReindexTrigger,
+} from '../../entities/search/search-project-reindex-trigger.entity';
+import { SearchIndexerService } from './search-indexer.service';
+
+const ENTITY_PROCESSING_CRON = '*/2 * * * * *';
```
romandubovyi: 🟡 **Important:** The cron expressions for trigger processing (e.g., `ENTITY_PROCESSING_CRON = '*/2 * * * * *'`) are currently hardcoded. It would be better to load these values from the environment configuration (e.g., via `ADVANCED_SEARCH_CONFIG`), similar to how `driftCron` is handled. This allows tuning the reindexing frequencies in different environments (dev/staging/production) without rebuilding the application.

apps/backend/src/data-marts/search/embedding/local-transformers.provider.ts
```diff
@@ -0,0 +1,123 @@
...
+        if (pipe === null) {
+          this.pipelinePromise = null;
+        }
+        return pipe;
+      });
+    }
+    return this.pipelinePromise;
+  }
+
+  private async initPipeline(): Promise<Pipe | null> {
```
romandubovyi, line 104: 💡 **Suggestion:** The `initPipeline` method dynamically loads `@huggingface/transformers` and downloads the model from Hugging Face Hub if it is not cached. On slow networks or if the Hub is down, this can hang indefinitely or take a very long time during startup/first search. Consider adding a timeout or warning logging if the pipeline initialization takes longer than a few seconds, so operators can quickly identify model-download delays.

apps/backend/src/data-marts/search/engine/in-memory-paginated.search.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  const descriptor = {
+    entityType,
+    entityId: row.entityId,
+    projectId: row.projectId,
+    title: parsed.title,
+    description: parsed.description,
+    richTextSlots: parsed.richTextSlots,
+    atomicTokenSlots: parsed.atomicTokenSlots,
```
romandubovyi, line 36: 🟡 **Important:** When `promptVec` is available but `row.embedding` is null, this returns null and drops the entity entirely—even if the keyword score alone would make it a strong match. Newly created or not-yet-indexed entities become invisible in search. Consider falling back to keyword-only scoring instead of discarding the row.

apps/backend/src/data-marts/controllers/search.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+  async search(
+    @AuthContext() context: AuthorizationContext,
+    @Query() query: SearchQueryDto
+  ): Promise<SearchResultResponseApiDto[]> {
+    const normalizedQuery = query.q.trim();
+
+    if (normalizedQuery.length < this.config.queryMinLength) {
+      throw new BadRequestException(
+        `Query must be at least ${this.config.queryMinLength} characters`
+      );
```
romandubovyi, line 31: 💡 **Suggestion:** `query.q` is already trimmed by the `SearchQueryDto` `@Transform` decorator. The controller's manual `trim()` here is redundant and could diverge from DTO logic if the transform changes. Consider using `query.q` directly and trusting the DTO transformation.

apps/backend/src/data-marts/search/schema/search-index.repository.ts
```diff
@@ -0,0 +1,551 @@
...
+        return await this.executeSearchCandidatesQuery(
+          `
+            SELECT ${selectColumns},
+                   COSINE_DISTANCE(idx.embedding, :promptVec) AS vector_distance
+            FROM ${table} idx
+            ${accessPredicate.joinSql}
+            WHERE idx.project_id = :projectId
+              ${accessPredicate.whereSql ? `AND ${accessPredicate.whereSql}` : ''}
+              ${draftFilter}
+              AND idx.embedding IS NOT NULL
```
romandubovyi, line 285: 🔵 **Question:** When `query.tokens` is empty, the fallback branch executes a query with no `LIKE` filters and returns up to `candidateLimit` rows ordered only by `updated_at`. Is this intentional for empty queries, or should it return an empty result set immediately to avoid surfacing irrelevant matches?

apps/backend/src/data-marts/dto/presentation/search-result-response-api.dto.ts
```diff
@@ -0,0 +1,37 @@
...
+  @ApiProperty({
+    type: Number,
+    nullable: true,
+    example: 92,
+    description: 'Vector similarity score; null when no vector score contributed to the result',
+  })
+  vecScore: number | null;
+
+  @ApiProperty({ example: 0, description: 'Semantic engine extendability bonus' })
+  extendability: number;
```
romandubovyi: 💡 **Suggestion:** `extendability` is an internal scoring bonus that can change when the algorithm is recalibrated. Consider removing it from the public API DTO to avoid leaking internal scoring mechanics and keep the API contract stable.

romandubovyi (commented):
Review complete. Overall assessment: high-quality implementation with thorough test coverage and robust error handling. Key points raised around unindexed entity fallback behavior, DTO trim duplication, empty query fallback behavior, and API contract stability for internal scoring details.

apps/backend/src/data-marts/search/indexing/search-indexer.service.ts
```diff
@@ -0,0 +1,194 @@
...
+    entityType: SearchableEntityType,
+    descriptor: EntityScoringDescriptor
+  ): Promise<void> {
+    const document = buildDocument(descriptor);
+    const hash = docHash(this.provider.modelId, indexSignature(descriptor, document));
+    const vecs = await this.provider.embed([embeddingText(descriptor)], {
+      inputType: 'search_document',
+    });
+    const vec = vecs[0] ?? null;
+    if (vec === null) {
```
max-voloshyn, line 191: 🔴 Blocking: Returning success when `embed()` yields `null` leaves any existing index row untouched, so update invalidations can complete while search keeps serving stale documents/titles until a later drift run happens to repair them. This path should either clear/mark the row missing or throw so the trigger retries instead of being marked successful.
🤖 Reviewed by Codex (GPT-5.5)

apps/backend/src/data-marts/search/engine/advanced-search.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  async search(projectId: string, prompt: string, options: SearchOptions): Promise<SearchResult[]> {
+    if (!options?.accessScope) {
+      throw new Error('Search accessScope is required');
+    }
+
+    const start = performance.now();
+
+    const requestedTypes = options.entityTypes ?? Object.values(SearchableEntityType);
+    const activeTypes = requestedTypes.filter(t => this.registry.has(t));
```
max-voloshyn: 🟡 Non-blocking: Deduplicate `requestedTypes` before searching. The public DTO accepts repeated or comma-duplicated `entityTypes`, and this currently runs the same search multiple times and can return duplicate results for one entity.
🤖 Reviewed by Codex (GPT-5.5)

apps/web/src/pages/search/SearchPage.tsx
```diff
@@ -0,0 +1,123 @@
...
+  },
+  DATA_DESTINATION: {
+    label: 'Destination',
+    icon: ArchiveRestore,
+    to: entityId => `/data-destinations?id=${entityId}`,
+  },
+};
+
+export function SearchPage() {
+  const [query, setQuery] = useState('');
```
max-voloshyn: 🟡 Non-blocking: The search query only lives in local component state, so `/search?q=...` links/bookmarks do not hydrate the input/results and refresh/back-forward loses the current search. Please sync the input with `useSearchParams` and update the query param as the user types.
🤖 Reviewed by Codex (GPT-5.5)

zapolsky:

Worktree multi-agent review (5 parallel passes — reviewer · logic-hunter · architect · tester · fresh-eyes — plus an adversarial cross-check) against PR head `8f63df9`. I traced and **disproved** the high-risk hypotheses against the checked-out code:

- **No cross-project / cross-role data leak.** Every candidate query pins `WHERE idx.project_id = :projectId` taken from the auth context (never a client param) and reuses the existing `applyData{Mart,Storage,Destination}VisibilityFilter` predicates; index rows JOIN the live entity on both `id` **and** `projectId`. HTTP (`@Auth(Role.viewer(Strategy.PARSE))`) and the MCP tool (`mcp:read`, project/user from `McpAuthContext`) are both scoped correctly.
- **No secret leakage.** The OpenRouter API key only ever appears in the `Authorization` header (`openrouter-embedding.provider.ts:71`); error paths log the response body, never the request/key. Persisted index docs contain titles/descriptions/field names only — no credentials.
- **No SQL injection.** Named→positional binding, `LIKE` pattern escaping, and a tokenizer that strips boolean operators (`search-index.repository.ts` / `search-query-builder.ts`).
- **No new nested-transaction risk on single-connection SQLite.** Trigger claiming uses a version-checked `UPDATE` (optimistic lock); the indexer/sync paths introduce no `@Transactional` nesting.

Migrations are additive, existence-guarded, and reversible. CI is green on this head (ESLint + tsc + all test suites pass), and the search test suite is real — I ran it locally too (22 search suites / 368 tests pass under `--runInBand`). Solid, well-tested work.

- **Keyword-only fallback is defeated at the SQL layer** (extends the still-open review thread on `in-memory-paginated.search.ts`). All three candidate queries add `AND idx.embedding IS NOT NULL` — vector (`search-index.repository.ts:283`), full-text (`:312`), and the LIKE fallback (`:418`, `executeFallbackCandidateQuery`, 387–425). That LIKE branch is exactly the keyword-only path used when `promptVec === null`, yet it still excludes any row whose embedding is `null` / `embedding_status='MISSING'`. So an entity indexed during an embedding-provider outage (or a fresh corpus indexed while the provider is misconfigured) stays invisible **even to keyword search** — the opposite of the "fall back to keyword scoring instead of returning nothing" resilience the follow-up commit advertises. The suggested keyword-fallback fix only fully lands if `embedding IS NOT NULL` is also dropped from the full-text + LIKE branches (keep it only on the vector branch, where `COSINE_DISTANCE` requires it). Self-heals once embeddings succeed, so non-blocking.

- **Drift project-sync can clobber a concurrent targeted reindex (stale index for up to ~1h).** `syncTypeProject` reads the entity page (`search-indexer.service.ts:86`), runs the slow `embed()` (`:119`), then `upsertMany` (`:143`) — and the upsert is a plain `INSERT … ON CONFLICT(entityId)` with no freshness guard (`search-index.repository.ts:167-176`). If a user edits entity X inside that embedding window, the 2-second targeted reindex writes the fresh row and the drift batch then overwrites it with the older snapshot. It recovers on the next hourly drift (the hash mismatch is re-detected), but search serves a stale title/field-count for X until then. Consider guarding the upsert with `WHERE updated_at <= :rowUpdatedAt`, or re-reading index state immediately before the write.

- **A single native-vector error disables vector search process-wide until restart.** `vectorAvailable` is a per-singleton latch (`search-index.repository.ts:138`); `markVectorUnavailable` flips it to `false` on *any* thrown `COSINE_DISTANCE` query (`:298` → `:535`) with no TTL/reset, and `supportsVector()` short-circuits on it forever (`:489-491`). One stale-dimension or corrupt blob — e.g. during an embedding model/provider switch on Cloud SQL — drops every project to the keyword candidate path for the whole process lifetime, which changes recall, not just latency. Consider time-boxing the unavailability or scoping the failure to the offending query.

<sub>Automated OWOX review · 5-agent worktree passes (reviewer · logic-hunter · architect · tester · fresh-eyes) · all candidate blockers adversarially resolved against the checked-out code. For transparency: I investigated the "missing `search.facade.impl.spec.ts`" angle and it is **not** CI-blocking here — no ESLint/coverage gate enforces facade-impl tests and two facade impls already ship spec-less on `main` — so it isn't flagged.</sub>

apps/backend/src/data-marts/search/config/advanced-search.config.ts
```diff
@@ -0,0 +1,125 @@
...
+        .string()
+        .optional()
+        .default('45')
+        .transform(v => parseInt(v, 10))
+        .pipe(z.number().int().min(0)),
+      candidateLimit: positiveIntFromString('500'),
+      queryMaxLength: positiveIntFromString(String(DEFAULT_ADVANCED_SEARCH_QUERY_MAX_LENGTH)),
+      embeddingConcurrency: positiveIntFromString('2'),
+      embeddingProvider: z.enum(['local', 'openrouter']).optional().default('local'),
+      embeddingModel: z.string().optional().default('google/gemini-embedding-2'),
```
max-voloshyn: 🔴 Blocking: `ADVANCED_SEARCH_EMBEDDING_MODEL` and `ADVANCED_SEARCH_EMBEDDING_DIMENSIONS` look provider-agnostic, but the local provider ignores them and uses a fixed Xenova model/dimension pair. Since these settings only affect OpenRouter today, please rename them to OpenRouter-specific env vars, e.g. `ADVANCED_SEARCH_OPENROUTER_EMBEDDING_MODEL` and `ADVANCED_SEARCH_OPENROUTER_EMBEDDING_DIMENSIONS`, or make the local provider consume provider-specific defaults.
🙋‍♂️ Reviewed by Max

apps/backend/src/data-marts/search/config/advanced-search.config.ts
```diff
@@ -0,0 +1,125 @@
...
+    candidateLimit: env['ADVANCED_SEARCH_CANDIDATE_LIMIT'],
+    queryMaxLength: env['ADVANCED_SEARCH_QUERY_MAX_LENGTH'],
+    embeddingConcurrency: env['ADVANCED_SEARCH_EMBEDDING_CONCURRENCY'],
+    embeddingProvider: env['ADVANCED_SEARCH_EMBEDDING_PROVIDER'],
+    embeddingModel: env['ADVANCED_SEARCH_EMBEDDING_MODEL'],
+    embeddingDimensions:
+      env['ADVANCED_SEARCH_EMBEDDING_DIMENSIONS'] ?? env['ADVANCED_SEARCH_EMBEDDING_DIM'],
+    openRouterApiKey: nonBlankOrNull(env['ADVANCED_SEARCH_OPENROUTER_API_KEY']),
+    openRouterAllowedProviders:
+      env['ADVANCED_SEARCH_OPENROUTER_ALLOWED_PROVIDERS'] ?? env['AI_ALLOWED_PROVIDERS'],
```
max-voloshyn, line 130: 🔴 Blocking: The OpenRouter embedding config partially duplicates existing OpenRouter AI settings. It already falls back to `AI_ALLOWED_PROVIDERS`, but not `AI_DATA_COLLECTION`/`AI_ZDR`; `AI_API_KEY` reuse would only be safe when `AI_PROVIDER=openrouter`. Please make the reuse/separation policy explicit so operators do not configure conflicting OpenRouter settings.
🙋‍♂️ Reviewed by Max

apps/backend/src/ee/mcp/tools/search-data-marts.tool.ts
```diff
@@ -0,0 +1,89 @@
...
+        userId: context.userId,
+        roles: context.roles,
+      },
+    });
+
+    const structuredContent = {
+      data_marts: results
+        .filter(result => result.entityType === SearchableEntityType.DATA_MART)
+        .map(result => ({
+          id: result.entityId,
```
max-voloshyn, line 77: 🔴 Blocking: MCP data mart discovery results should include a direct Data Mart `url`, not only identifiers and scores. Please add an actionable `/ui/{projectId}/data-marts/{id}/data-setup` URL field to this search result shape, and apply the same output-shape update to `ListDataMartsTool` so both MCP data mart discovery tools return direct links.
🙋‍♂️ Reviewed by Max

kalnik-a-a:
Addressed the three follow-up notes in https://github.com/OWOX/owox-data-marts/commit/c07669472648d2b607d60a6779a48ad731b0b69b.

- Keyword fallback now includes rows with missing embeddings in the full-text and LIKE candidate paths, while the native vector branch still filters to rows with embeddings before calling `COSINE_DISTANCE`.
- In-memory scoring now falls back to keyword-only scoring when the prompt vector exists but a row embedding is missing.
- Project drift writes now re-check the index state immediately before upsert, so a slow project reindex does not overwrite a fresher targeted reindex.
- Native vector query failures are now time-boxed with a retry cooldown instead of disabling vector search for the whole process lifetime.

Verified with the targeted search suites, backend production typecheck, `advanced-search` Jest gate, and the full `npm run test` run.

apps/backend/src/data-marts/search/embedding/local-transformers.provider.ts
kalnik-a-a: Leaving Worker Threads out of this follow-up because that would be a larger runtime architecture change and should be driven by profiling data. The local provider remains bounded by the configured indexing concurrency/batch settings.

I did add operator visibility around the slow local-model path: pipeline initialization now logs a warning if it takes longer than the threshold, without hard-failing legitimate first downloads. The trigger schedules are also now configurable, so production can reduce local embedding pressure without rebuilding.

apps/backend/src/data-marts/search/indexing/search-reindex-trigger-handler.service.ts
kalnik-a-a: Addressed in https://github.com/OWOX/owox-data-marts/commit/c07669472648d2b607d60a6779a48ad731b0b69b. The entity trigger cron and all three project-trigger crons now come from `ADVANCED_SEARCH_CONFIG` with the previous hardcoded values preserved as defaults:

- `ADVANCED_SEARCH_ENTITY_PROCESSING_CRON`
- `ADVANCED_SEARCH_DATA_MART_PROJECT_PROCESSING_CRON`
- `ADVANCED_SEARCH_DATA_STORAGE_PROJECT_PROCESSING_CRON`
- `ADVANCED_SEARCH_DATA_DESTINATION_PROJECT_PROCESSING_CRON`

This keeps current behavior unchanged while allowing environment-specific tuning.

apps/backend/src/data-marts/search/embedding/local-transformers.provider.ts
kalnik-a-a: Addressed with warning visibility rather than a hard timeout. `LocalTransformersEmbeddingProvider.initPipeline()` now logs when pipeline initialization takes longer than the threshold, then clears the timer once the model load completes.

I avoided a hard timeout here because a first local model download can be slow but still valid; failing it automatically could make default local setup less reliable.

apps/backend/src/data-marts/search/engine/in-memory-paginated.search.ts
kalnik-a-a: Addressed in https://github.com/OWOX/owox-data-marts/commit/c07669472648d2b607d60a6779a48ad731b0b69b. Missing row embeddings no longer drop otherwise relevant keyword matches when the prompt vector exists. The in-memory scorer now falls back to keyword-only scoring for those rows.

The SQL candidate layer was updated as well: full-text and LIKE fallback queries no longer require `idx.embedding IS NOT NULL`; that filter remains only on the native vector branch where `COSINE_DISTANCE` needs an embedding.

apps/backend/src/data-marts/controllers/search.controller.ts
kalnik-a-a: Addressed. The controller now uses the DTO-transformed `query.q` directly and no longer applies a second manual `trim()`.

apps/backend/src/data-marts/search/schema/search-index.repository.ts
kalnik-a-a: Addressed at the service boundary. `AdvancedSearchService.search()` now tokenizes the prompt first and returns an empty result set when there are no searchable tokens, so punctuation-only or otherwise empty prompts do not fall through to arbitrary recent rows.

This keeps the lower-level repository focused on candidate retrieval while preventing irrelevant public/API results.

apps/backend/src/data-marts/dto/presentation/search-result-response-api.dto.ts
kalnik-a-a: Addressed. `extendability` is now kept internal to the scoring model and is no longer exposed through the public backend response DTO, facade result type, or frontend API response type.

apps/backend/src/data-marts/search/indexing/search-indexer.service.ts
kalnik-a-a: Addressed in https://github.com/OWOX/owox-data-marts/commit/c07669472648d2b607d60a6779a48ad731b0b69b. Targeted reindex no longer succeeds silently when `embed()` returns `null`.

The indexer now writes the current descriptor document with `embedding = null` / `embeddingStatus = MISSING`, then throws so the trigger remains retryable. That keeps keyword search fresh for the updated entity while preserving retry pressure for the missing vector.

apps/backend/src/data-marts/search/engine/advanced-search.service.ts
kalnik-a-a: Addressed. `AdvancedSearchService` now deduplicates requested entity types before resolving sources and running searches, so repeated `entityTypes` values do not duplicate work or results.

apps/web/src/pages/search/SearchPage.tsx
kalnik-a-a: Addressed. The search page now hydrates from `/search?q=...`, updates the `q` search param as the user types, removes it when the input is cleared, and stays coherent with back/forward navigation.

apps/backend/src/data-marts/search/config/advanced-search.config.ts
kalnik-a-a: Addressed in https://github.com/OWOX/owox-data-marts/commit/c07669472648d2b607d60a6779a48ad731b0b69b. The provider-agnostic embedding env names were removed for this path, and the OpenRouter-only settings are now explicitly named:

- `ADVANCED_SEARCH_OPENROUTER_EMBEDDING_MODEL`
- `ADVANCED_SEARCH_OPENROUTER_EMBEDDING_DIMENSIONS`

The default provider remains `local`; the local provider still uses its fixed bundled Xenova model/dimension pair, so these envs no longer imply local-provider behavior they do not control.

apps/backend/src/data-marts/search/config/advanced-search.config.ts
kalnik-a-a: Addressed. The OpenRouter reuse/separation policy is now explicit in config and tests:

- search-specific OpenRouter envs win when set;
- `AI_ALLOWED_PROVIDERS`, `AI_DATA_COLLECTION`, and `AI_ZDR` are used as fallback routing/privacy policy values;
- `AI_API_KEY` is intentionally not reused for embeddings, even when `AI_PROVIDER=openrouter`.

This keeps embedding credentials separated while avoiding duplicated policy configuration where reuse is safe.

apps/backend/src/ee/mcp/tools/search-data-marts.tool.ts
kalnik-a-a: Addressed in https://github.com/OWOX/owox-data-marts/commit/c07669472648d2b607d60a6779a48ad731b0b69b. Both MCP data-mart discovery tools now include a direct `url` field:

- `get_relevant_data_marts_by_prompt`
- `list_data_marts`

The URL is now an absolute browser link built from `PublicOriginService.getPublicOrigin()` plus the data-mart UI path, and search results still keep `relevance_score`.

apps/backend/src/data-marts/search/indexing/search-indexer.service.ts
```diff
@@ -0,0 +1,234 @@
...
+      this.logger.warn(
+        `reindexEntity: ${entityType} ${descriptor.entityId} not indexed because embedding could not be generated`
+      );
+      await this.repository.upsert(entityType, {
```
max-voloshyn, line 195: 🔴 Blocking: Persisting a `MISSING` row here still does not make the entity discoverable in the common Cloud SQL vector path. When prompt embedding exists and `cloudsql_vector` is enabled, `SearchIndexRepository.searchCandidates()` returns from the vector branch with `AND idx.embedding IS NOT NULL`, so rows saved with `embedding: null` are skipped and keyword candidates are never queried. Please merge keyword candidates with vector candidates or otherwise route matching `MISSING` rows into scoring before returning.
🤖 Reviewed by Codex (GPT-5.5)

.env.example
```diff
@@ -89,3 +89,69 @@ GOOGLE_SHEETS_EXTENSION_ORIGIN=
...
+# Local directory for caching the downloaded ONNX model files
+# Default: uses @huggingface/transformers default cache; in production set to a vendored path inside the image
+# ADVANCED_SEARCH_MODEL_CACHE_DIR=
+
+# Embedding provider: local (fixed ONNX model, default) or openrouter
+# ADVANCED_SEARCH_EMBEDDING_PROVIDER=local
+
+# OpenRouter embedding model and output dimensions.
+# These settings are used only when ADVANCED_SEARCH_EMBEDDING_PROVIDER=openrouter.
+# ADVANCED_SEARCH_EMBEDDING_MODEL=google/gemini-embedding-2
```
max-voloshyn: 🔴 Blocking: The example still documents the old provider-agnostic `ADVANCED_SEARCH_EMBEDDING_MODEL` / `ADVANCED_SEARCH_EMBEDDING_DIMENSIONS` names, but `loadAdvancedSearchConfig()` now ignores those env vars and only reads `ADVANCED_SEARCH_OPENROUTER_EMBEDDING_MODEL` / `ADVANCED_SEARCH_OPENROUTER_EMBEDDING_DIMENSIONS`. Operators following this template will think they changed the OpenRouter model/dimensions while the app silently keeps the defaults. Please update the example names to match the new config keys.
🤖 Reviewed by Codex (GPT-5.5)

kalnik-a-a: Addressed in https://github.com/OWOX/owox-data-marts/commit/25bc9988369bfd28c60d763d1e58c29b6373b837. `.env.example` now uses the OpenRouter-specific embedding env names and includes the missing advanced-search trigger cron envs as well. I also added a config spec that checks the example file documents all active advanced-search env vars and does not reintroduce the old provider-agnostic model/dimension names.

apps/backend/src/data-marts/search/indexing/search-indexer.service.ts
kalnik-a-a: Addressed in https://github.com/OWOX/owox-data-marts/commit/25bc9988369bfd28c60d763d1e58c29b6373b837. The Cloud SQL vector path now supplements vector candidates with keyword candidates where `embedding_status` is `MISSING`, then deduplicates before in-memory scoring. That keeps normal vector ranking for ready rows while still allowing missing-embedding rows to be discovered by keyword relevance.

max-voloshyn (approved):

PR #1364, opened 2026-06-26
Add API key exchange and auth context support

max-voloshyn:
## Summary

- Add project-member API key exchange support for Better Auth and Null IDP
- Add auth context resolution through `@owox/api-client` and `owox-ctl status`
- Bind exchanged access tokens to the originating API key ID
- Reject API-key based auth from API-key management, project-member administration, user provisioning, Intercom identity, and MCP OAuth authorization endpoints
- Allow API-key based auth to manage domain resources such as contexts, data marts, storages, destinations, and reports through the existing role and capability checks
- Include project titles in Better Auth exchanged API-key tokens for auth-context/status output
- Add a 15-minute Better Auth exchanged-token TTL and expose optional `accessTokenExpiresIn` from API-key exchange
- Add Null IDP and Better Auth app smoke coverage for API-key exchange, restricted auth/admin surfaces, domain management, and `owox-ctl status`
- Use an explicit Oclif manifest fixture for app smoke tests that is compatible with sharded e2e runs
- Preserve compatibility with older backends by falling back only when `/api/auth/context` is unavailable

## Verification

- `git diff --check`
- `npm run test -w @owox/backend -- --runTestsByPath src/idp/controllers/auth-context.controller.spec.ts src/project-member-api-keys/use-cases/exchange-project-member-api-key.service.spec.ts src/data-marts/controllers/context.controller.spec.ts src/idp/guards/idp.guard.spec.ts`
- `npm run test -w @owox/idp-better-auth -- --runTestsByPath src/services/token-service.test.ts src/providers/better-auth-provider.test.ts`
- `npm run build --workspace owox`
- `npm run test:e2e -w @owox/backend -- --runTestsByPath test/api-key-null-idp-app-smoke.e2e-spec.ts test/api-key-better-auth-app-smoke.e2e-spec.ts`

## Review

- Subagent security and correctness reviews found no blockers for the auth-context and exchanged-token changes
- Domain resource management remains allowed for API-key auth per product scope; account/auth administration surfaces remain rejected
- Role/readOnly API-key restrictions remain out of scope for this PR

### DIFF
--- a/apps/backend/src/data-marts/controllers/data-storage.controller.ts
+++ b/apps/backend/src/data-marts/controllers/data-storage.controller.ts
@@ -13,7 +13,6 @@ import {
 } from '@nestjs/common';
 import { ApiTags } from '@nestjs/swagger';
 import { Auth, AuthContext, AuthorizationContext } from '../../idp';
-import { AuthorizationError } from '@owox/idp-protocol';
 import { Role, Strategy } from '../../idp/types/role-config.types';
 import { CreateDataStorageApiDto } from '../dto/presentation/create-data-storage-api.dto';
 import { DataStorageAccessValidationResponseApiDto } from '../dto/presentation/data-storage-access-validation-response-api.dto';
@@ -107,21 +106,6 @@ export class DataStorageController {
     }
   }
 
-  private rejectApiKeyAuthForAccessControlUpdate(
-    context: AuthorizationContext,
-    dto: Pick<CreateDataStorageApiDto | UpdateDataStorageApiDto, 'ownerIds'> &
-      Partial<Pick<UpdateDataStorageApiDto, 'contextIds'>>
-  ): void {
-    if (
-      context.authFlow === 'api_key' &&
-      (dto.ownerIds !== undefined || dto.contextIds !== undefined)
-    ) {
-      throw new AuthorizationError(
-        'API key authentication is not allowed for access control updates'
-      );
-    }
-  }
-
   @Auth(Role.viewer(Strategy.PARSE))
   @Get()
   @ListDataStoragesSpec()
@@ -143,8 +127,6 @@ export class DataStorageController {
     @Param('id') id: string,
     @Body() dto: UpdateDataStorageApiDto
   ): Promise<DataStorageResponseApiDto> {
-    this.rejectApiKeyAuthForAccessControlUpdate(context, dto);
-
     const command = this.mapper.toUpdateCommand(id, context, dto);
     const dataStorageDto = await this.updateService.run(command);
     return this.mapper.toApiResponse(dataStorageDto);
@@ -157,8 +139,6 @@ export class DataStorageController {
     @AuthContext() context: AuthorizationContext,
     @Body() dto: CreateDataStorageApiDto
   ): Promise<DataStorageResponseApiDto> {
-    this.rejectApiKeyAuthForAccessControlUpdate(context, dto);
-
     const command = this.mapper.toCreateCommand(context, dto);
     const dataStorageDto = await this.createService.run(command);
     return this.mapper.toApiResponse(dataStorageDto);
--- a/apps/backend/src/data-marts/controllers/data-storage.controller.ts
+++ b/apps/backend/src/data-marts/controllers/data-storage.controller.ts
@@ -13,6 +13,7 @@ import {
 } from '@nestjs/common';
 import { ApiTags } from '@nestjs/swagger';
 import { Auth, AuthContext, AuthorizationContext } from '../../idp';
+import { AuthorizationError } from '@owox/idp-protocol';
 import { Role, Strategy } from '../../idp/types/role-config.types';
 import { CreateDataStorageApiDto } from '../dto/presentation/create-data-storage-api.dto';
 import { DataStorageAccessValidationResponseApiDto } from '../dto/presentation/data-storage-access-validation-response-api.dto';
@@ -106,6 +107,21 @@ export class DataStorageController {
     }
   }
 
+  private rejectApiKeyAuthForAccessControlUpdate(
+    context: AuthorizationContext,
+    dto: Pick<CreateDataStorageApiDto | UpdateDataStorageApiDto, 'ownerIds'> &
+      Partial<Pick<UpdateDataStorageApiDto, 'contextIds'>>
+  ): void {
+    if (
+      context.authFlow === 'api_key' &&
+      (dto.ownerIds !== undefined || dto.contextIds !== undefined)
+    ) {
+      throw new AuthorizationError(
+        'API key authentication is not allowed for access control updates'
+      );
+    }
+  }
+
   @Auth(Role.viewer(Strategy.PARSE))
   @Get()
   @ListDataStoragesSpec()
@@ -127,6 +143,8 @@ export class DataStorageController {
     @Param('id') id: string,
     @Body() dto: UpdateDataStorageApiDto
   ): Promise<DataStorageResponseApiDto> {
+    this.rejectApiKeyAuthForAccessControlUpdate(context, dto);
+
     const command = this.mapper.toUpdateCommand(id, context, dto);
     const dataStorageDto = await this.updateService.run(command);
     return this.mapper.toApiResponse(dataStorageDto);
@@ -139,6 +157,8 @@ export class DataStorageController {
     @AuthContext() context: AuthorizationContext,
     @Body() dto: CreateDataStorageApiDto
   ): Promise<DataStorageResponseApiDto> {
+    this.rejectApiKeyAuthForAccessControlUpdate(context, dto);
+
     const command = this.mapper.toCreateCommand(context, dto);
     const dataStorageDto = await this.createService.run(command);
     return this.mapper.toApiResponse(dataStorageDto);
--- a/apps/backend/src/data-marts/controllers/project-members.controller.ts
+++ b/apps/backend/src/data-marts/controllers/project-members.controller.ts
@@ -1,6 +1,13 @@
 import { Body, Controller, Delete, Get, HttpCode, Param, Post, Put } from '@nestjs/common';
 import { ApiTags } from '@nestjs/swagger';
-import { Auth, AuthContext, AuthorizationContext, Role, Strategy } from '../../idp';
+import {
+  Auth,
+  AuthContext,
+  AuthorizationContext,
+  RejectApiKeyAuth,
+  Role,
+  Strategy,
+} from '../../idp';
 import {
   ApproveMembershipRequestApiDto,
   ApproveMembershipRequestResponseApiDto,
@@ -66,6 +73,7 @@ export class ProjectMembersController {
   ) {}
 
   @Auth(Role.viewer(Strategy.PARSE))
+  @RejectApiKeyAuth()
   @Get()
   @ListProjectMembersSpec()
   async list(@AuthContext() context: AuthorizationContext): Promise<ProjectMemberResponseApiDto[]> {
@@ -74,6 +82,7 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
+  @RejectApiKeyAuth()
   @Post('invite')
   @HttpCode(202)
   @InviteProjectMemberSpec()
@@ -112,6 +121,7 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.viewer(Strategy.PARSE))
+  @RejectApiKeyAuth()
   @Get('user-provisioning-settings')
   @GetUserProvisioningSettingsSpec()
   async getProvisioningSettings(
@@ -122,6 +132,7 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
+  @RejectApiKeyAuth()
   @Put('user-provisioning-settings')
   @UpdateUserProvisioningSettingsSpec()
   async updateProvisioningSettings(
@@ -140,6 +151,7 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
+  @RejectApiKeyAuth()
   @Put(':userId')
   @UpdateProjectMemberSpec()
   async update(
@@ -168,6 +180,7 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
+  @RejectApiKeyAuth()
   @Delete(':userId')
   @HttpCode(204)
   @RemoveProjectMemberSpec()
@@ -181,6 +194,7 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
+  @RejectApiKeyAuth()
   @Get('requests')
   @ListMembershipRequestsSpec()
   async listRequests(
@@ -191,6 +205,7 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
+  @RejectApiKeyAuth()
   @Post('requests/:requestId/approve')
   @HttpCode(200)
   @ApproveMembershipRequestSpec()
@@ -218,6 +233,7 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
+  @RejectApiKeyAuth()
   @Post('requests/:requestId/decline')
   @HttpCode(204)
   @DeclineMembershipRequestSpec()
--- a/apps/backend/src/data-marts/controllers/project-members.controller.ts
+++ b/apps/backend/src/data-marts/controllers/project-members.controller.ts
@@ -52,6 +52,7 @@ import {
 
 @Controller('members')
 @ApiTags('Project Members')
+@RejectApiKeyAuth()
 /**
  * Self-protection (admin cannot remove or change-role on themselves) is
  * enforced inside `RemoveProjectMemberService.run` and
@@ -73,7 +74,6 @@ export class ProjectMembersController {
   ) {}
 
   @Auth(Role.viewer(Strategy.PARSE))
-  @RejectApiKeyAuth()
   @Get()
   @ListProjectMembersSpec()
   async list(@AuthContext() context: AuthorizationContext): Promise<ProjectMemberResponseApiDto[]> {
@@ -82,7 +82,6 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
-  @RejectApiKeyAuth()
   @Post('invite')
   @HttpCode(202)
   @InviteProjectMemberSpec()
@@ -121,7 +120,6 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.viewer(Strategy.PARSE))
-  @RejectApiKeyAuth()
   @Get('user-provisioning-settings')
   @GetUserProvisioningSettingsSpec()
   async getProvisioningSettings(
@@ -132,7 +130,6 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
-  @RejectApiKeyAuth()
   @Put('user-provisioning-settings')
   @UpdateUserProvisioningSettingsSpec()
   async updateProvisioningSettings(
@@ -151,7 +148,6 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
-  @RejectApiKeyAuth()
   @Put(':userId')
   @UpdateProjectMemberSpec()
   async update(
@@ -180,7 +176,6 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
-  @RejectApiKeyAuth()
   @Delete(':userId')
   @HttpCode(204)
   @RemoveProjectMemberSpec()
@@ -194,7 +189,6 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
-  @RejectApiKeyAuth()
   @Get('requests')
   @ListMembershipRequestsSpec()
   async listRequests(
@@ -205,7 +199,6 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
-  @RejectApiKeyAuth()
   @Post('requests/:requestId/approve')
   @HttpCode(200)
   @ApproveMembershipRequestSpec()
@@ -233,7 +226,6 @@ export class ProjectMembersController {
   }
 
   @Auth(Role.admin(Strategy.INTROSPECT))
-  @RejectApiKeyAuth()
   @Post('requests/:requestId/decline')
   @HttpCode(204)
   @DeclineMembershipRequestSpec()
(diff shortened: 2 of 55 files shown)


zapolsky:

Reviewed against a full worktree checkout of `230caf9` with parallel specialist passes (reviewer, logic-hunter, architect, tester, fresh-eyes generalist) plus two adversarial skeptics on the security candidates. The core machinery is sound and well-tested: token integrity (HS256-signed encrypted payload), the `X-OWOX-Api-Key-Id` header binding, the `@RejectApiKeyAuth` rejection on key-management + the `authFlow === 'api_key'` rejection on MCP OAuth `authorize`, and `owox-ctl status` omitting secrets are all covered by green unit specs. Two security candidates were traced end-to-end and **did not survive the blocker gate** — both resolved to design/intent questions (below), so nothing is posted inline.

Primary notes (criticality ≥ 80), highest first — none block merge.

**1. API-key tokens carry the member's full current role — admin keys can administer members programmatically.** This PR newly enables Better Auth api-key tokens (`better-auth-provider.ts:254-280`, previously `throw IdpOperationNotSupportedError`) and rejects them only on key-management (`project-member-api-keys.controller.ts`) and MCP OAuth `authorize` (`oauth-authorization.controller.ts:196-201`). The project-members admin endpoints — `invite` (`project-members.controller.ts:76`), `update` role (`:142`), `remove` (`:170`), `updateProvisioningSettings` (`:124`), `approve/declineRequest` (`:193`/`:220`) — carry `@Auth(Role.admin(Strategy.INTROSPECT))` but **not** `@RejectApiKeyAuth`, and `checkApiKeyUsageRestrictions` early-returns when the flag is unset (`idp.guard.ts:146-154`). So an admin member's exchanged key can invite new admins, change roles, and remove members non-interactively, and `invite` returns the magic link in the response body (`project-members.controller.ts:102-108`). This looks intentional given the "key inherits current membership role" design (`provider.ts` doc + changeset), and `create-project-member-api-key.service.ts` binds new keys to the caller's own `userId` (a key cannot directly mint another member's key), so it is not a clean key-management bypass. **Question:** is programmatic org administration (especially `invite` disclosing a magic link) intended for admin api-keys, or should the project-members admin endpoints also carry `@RejectApiKeyAuth()` so keys are scoped to data operations? If intended, a one-line note/test would make the boundary explicit.

**2. A revoked or expired API key keeps working for up to ~1h via its already-exchanged token.** `verifyCredential` checks `isRevoked`/`isExpired` only at exchange time (`project-member-api-key.service.ts:118`). The issued token is a self-contained encrypted JWT with a fixed 1-hour TTL (`crypto-service.ts:23`), and the guard path (`introspectToken`/`parseToken`, `token-service.ts:32-55`) only decrypts + shape-checks — there is **no** DB re-lookup of the key's `revokedAt`/`expiresAt` on subsequent requests. **Question:** is up-to-1h revocation latency acceptable for API keys, or should the guard re-check key status (or the token TTL be shorter)? Worth an explicit decision since revoking a leaked key won't take effect immediately.

**3. Rollout coupling — a newer `owox-ctl` against an older self-hosted backend reports a valid key as unauthenticated.** `owox-ctl status` now calls `auth.getContext()`, which hits the **new** `/api/auth/context` route (`status.ts:81`). Against an older backend that lacks that route, the 404 is caught at `status.ts:82-90` and reported as `authenticated: false` with the error — i.e. a valid API key looks broken. `@owox/api-client` and `@owox/ctl` ship to npm, so an updated CLI can run against an older backend; the `fixed` changeset group keeps a single deploy in lockstep but not the mixed npm-CLI-vs-self-hosted case. **Suggestion:** in `getStatus`, treat a 404 from `/api/auth/context` as a fallback to the exchange-only check (`authenticate()`) so a valid key still reports authenticated, just without the project/member context block.

**4. (Pre-existing, not introduced here) Magic-link invite role is not bound to the invitee.** The encrypted role on `/auth/magic-link-success/:role` is `encrypt(role)` with no email/user/invite claim (`magic-link-service.ts:25`), and `magicLinkSuccess` applies the decrypted role to whoever holds the session, overwriting an existing member's role (`page-service.ts:124-148` → `addMemberToOrganization` → `UPDATE member SET role`). A holder of any live `encrypt('admin')` ciphertext (1h TTL) plus their own session could self-promote. The skeptic confirmed this is **identical pre- and post-PR** — this PR only relocates the value from `?role=` query to the `:role` path (to survive callback query normalization), so it is **not a regression and not a blocker**, and practical exploitation needs a leaked live admin ciphertext (log access or invite interception). Flagging only because you're touching this exact code: consider binding the role ciphertext to the invitee's email and verifying it against `session.user.email` in `magicLinkSuccess`.

<details>

- `null-provider.ts:172-178` — **Null IDP fails open to admin** (dev-only, ~55): unknown tokens resolve to `defaultPayload` (admin, no `authFlow`), so after a restart an api_key token string silently loses its api-key restrictions. Confined to `IDP_PROVIDER=none`, which is already single-admin; document or return `null` for unknown tokens.
- `eslint.config.mjs:78` — **new auth endpoint outside the `require-auth-decorator` lint net** (~55): the rule is scoped to `src/data-marts/controllers/**`, so `idp/controllers/auth-context.controller.ts` isn't covered. It does carry `@Auth`, but the guarantee is manual; consider broadening the glob to `src/**/controllers/**/*.controller.ts`.
- `project-member-api-keys.controller.ts:3-4`, `auth-context.controller.ts:3-4` — **deep imports instead of the `../../idp` barrel** (~50): the pmak controller regressed an existing barrel import to a deep path for no functional reason; all symbols are re-exported from `../../idp`.
- minor (0–49): `null-provider.ts:67` `issuedAccessTokens` Map has no eviction (bounded by distinct `apiKeyId`, dev-only); `null-provider.test.ts` asserts header-aware `userApiMiddleware` behavior the impl doesn't have (passes coincidentally — misleading coverage); `magic-link-service.ts:27` + `page-service.ts:170` double-encode/decode the role segment (no-op for the JWT charset).

</details>

<sub>CI-parity: `eslint` clean on touched backend sources; affected unit specs green (backend targeted 27/27, idp-better-auth 150, idp-protocol 18, api-client 32, ctl 35). e2e smokes not run (heavyweight, separate gate). Backend `tsc -b` over test files surfaces pre-existing baseline errors across many untouched specs — **not** a CI gate here (ts-jest runs with `diagnostics:false`; CI runs jest), so not attributable to this PR. Changeset is correctly scoped — the `fixed` group in `.changeset/config.json` versions api-client/ctl/idp/backend in lockstep with `owox`.</sub>

<sub>Automated OWOX review · worktree multi-agent passes · both security candidates adversarially verified (CONFIRMED-BUT-PREEXISTING / UNPROVEN → posted as questions, not blockers).</sub>

max-voloshyn:
@zapolsky thank you, regarding https://github.com/OWOX/owox-data-marts/pull/1364#issuecomment-4809797283:

**1. Project-member admin endpoints via API keys**  
   Agreed, this should be addressed here. API keys should cover data/API operations, not project-member administration. I added `@RejectApiKeyAuth()` to the project-member admin endpoints and covered the boundary with a metadata test.

**2. Revoked/expired key keeps working until exchanged token TTL**  
   Out of scope for this PR. I’m not changing token revocation semantics here.

**3. New `owox-ctl` against older backend**  
   Agreed, worth addressing here. I added a narrow fallback: if `/api/auth/context` returns 404, `owox-ctl status` falls back to exchange-only authentication and reports `authenticated: true` without project/member context.

**4. Magic-link role not bound to invitee**  
   Out of scope for this PR. I’m not changing magic-link invite payload semantics here.

**Other observations**  
Out of scope for this PR; I’m not changing them here.

zapolsky:

Incremental re-review of `51acd19` (previous pass was `230caf9`). Re-ran a full worktree checkout with parallel specialist passes (reviewer, logic-hunter, architect, tester, fresh-eyes generalist). Both review-feedback fixes landed correctly:

- **Note 1 (project-member admin via api-keys) — fixed & verified.** All 9 `ProjectMembersController` endpoints now carry `@RejectApiKeyAuth()`, enforced in `idp.guard.ts:146-154` (`authFlow === 'api_key' && rejectApiKeyAuth` → `AuthorizationError`). The new metadata spec is non-vacuous — removing the decorator from any one endpoint makes the test fail (verified by mutation).
- **Note 3 (newer `owox-ctl` vs older backend) — fixed & verified.** `status.ts` now treats a `404` from `/api/auth/context` (`OWOXApiError`, not `OWOXAuthError`) as "endpoint missing" and falls back to exchange-only auth → reports `authenticated: true` without the context block. The fallback cannot misfire into a false positive: it still requires the key *exchange* to independently succeed, and exchange failures are forced to `OWOXAuthError` (`auth.ts:91` → `errors.ts:98`), which the 404 guard explicitly excludes. Every branch is covered in `status.test.ts`.
- Notes 2 & 4 (revocation latency, magic-link role binding) — acknowledged as deferred/out-of-scope; still tracked in the previous comment.

**1. `user-provisioning` endpoints remain reachable by exchanged api-key tokens — the same "not data-ops" class you just scoped out for `/members`.** `RequestAccessController` has no `@RejectApiKeyAuth()` on `POST /user-provisioning/request-access` (`request-access.controller.ts:47`) or `POST /user-provisioning/create-new-project` (`:63`), so an exchanged api-key token (carrying the member's live role) can call them. I traced both end-to-end and this is **not** privilege escalation — both run as the token's own `userId`, `request-access` only files a pending request an admin must still approve, `create-new-project` makes a brand-new isolated project the key cannot then act on (header binding keeps it on its original project), and the null/self-hosted provider throws `IdpOperationNotSupportedError`. But it *is* a scope-consistency gap against the stated principle ("API keys cover data/API operations, not administration/provisioning"). Codex flagged the same routes inline. **Question:** intended, or should `request-access` + `create-new-project` also carry `@RejectApiKeyAuth()`? (The same low-severity question applies to `POST /intercom/jwt`, which mints an Intercom identity token — `intercom.controller.ts:20`.)

**2. Consider applying `@RejectApiKeyAuth()` at the class level on the two privileged controllers.** The guard already resolves it via `getAllAndOverride([handler, getClass()])` (`idp.guard.ts:52-55`), so a class-level decorator on `ProjectMembersController` and `ProjectMemberApiKeysController` would auto-protect any *future* method added to them. The current per-method application + metadata spec only catches a *removed* decorator on a known method — a newly-added unguarded endpoint would slip through. Lower annotation cost, strictly safer.

<details>
<summary>Additional observations (non-blocking, for completeness)</summary>

- `better-auth-provider.ts:293-301` / `auth-context.controller.ts` — for api-key auth the Better Auth payload omits `projectTitle`, so `owox-ctl status` renders `"project": { "title": null }`, yet `docs/api/owox-ctl.md` and the `status.test.ts` fixtures show `"title": "Demo Project"`. Cosmetic, but the advertised output is misleading.
- **Codex's inline changeset comment is a false positive** — `.changeset/config.json` puts `@owox/api-client`, `@owox/ctl`, and the idp packages in the same `fixed` group as `owox`, so the single `owox: minor` changeset versions and publishes all of them in lockstep. No frontmatter change is needed.
- CI-parity: `eslint` clean on touched backend/ctl/api-client sources; targeted unit specs green (project-members 9/9, `idp.guard` suite 18, ctl `status` 7/7). `tsc` in the shared review worktree reports 3 errors in `apps/ctl/src/commands/status.ts` against `@owox/api-client` (`OWOXAuthContext`, `auth`) — this is a **worktree artifact only** (the shared `node_modules` resolves `@owox/api-client` to the `main`-branch build, which predates this PR's new exports); the source exports are present (`api-client/src/index.ts:2`, `client.ts:22`) and CI builds api-client topologically before ctl. **Not attributable to this PR.**

</details>

<sub>Automated OWOX review · incremental worktree multi-agent pass · no blockers (zero candidates survived to the blocker gate; the user-provisioning item was adversarially confirmed as scope, not escalation, so it is posted as a question).</sub>

zapolsky (approved):

packages/idp-better-auth/src/services/page-service.ts
```diff
@@ -164,10 +166,21 @@ export class PageService {
     }
   }
 
+  private getEncryptedMagicLinkRole(req: ExpressRequest): string {
+    if (typeof req.params?.role === 'string' && req.params.role) {
+      return decodeURIComponent(req.params.role);
```
romandubovyi: 💡 **Suggestion:** Express automatically decodes path parameters, so `req.params.role` is already decoded. Calling `decodeURIComponent` again is redundant and can potentially throw a `URIError` if the string contains a `%` character that is not part of a valid escape sequence. You can safely return `req.params.role` directly.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
```diff
@@ -185,6 +194,12 @@ export class OAuthAuthorizationController {
   }
 
   private toAuthorizationContext(payload: Payload): AuthorizationContext {
+    if (payload.authFlow === 'api_key') {
```
romandubovyi, line 188: 🟡 **Important:** This endpoint rejects API-key tokens with `UnauthorizedException` (401), while `@RejectApiKeyAuth` routes reject them with `AuthorizationError` (403). Consider using a consistent 403 response for the same policy so clients don't treat this as an authentication problem rather than an authorization restriction.

apps/backend/src/idp/guards/idp.guard.ts
```diff
@@ -88,7 +94,11 @@ export class IdpGuard implements CanActivate {
         void this.idpProjectionsService.updateProjectionsFromIdpPayload(tokenPayload);
       }
     } catch (error) {
-      if (error instanceof UnauthorizedException || error instanceof AuthorizationError) {
+      if (
```
romandubovyi: 🟡 **Important:** Rethrowing `AuthenticationError` directly instead of wrapping it as `AuthenticationError('Authentication failed')` changes the error message exposed to clients/logs for all routes, not just API-key rejections. Please confirm this is intentional and that no existing tests or alerts rely on the previous generic message.

apps/backend/src/idp/guards/idp.guard.ts
```diff
@@ -62,6 +67,7 @@ export class IdpGuard implements CanActivate {
     try {
       const tokenPayload = await this.authenticateUser(request, roleConfig.strategy);
       this.checkApiKeyHeaderBinding(request, tokenPayload);
```
romandubovyi: 💡 **Suggestion:** Consider running `checkApiKeyUsageRestrictions` before `checkApiKeyHeaderBinding`. On `@RejectApiKeyAuth` routes, this would surface the more descriptive "API key authentication is not allowed for this endpoint" message instead of the generic "Access denied by api key" when a valid API-key token is presented.

apps/backend/test/utils/api-key-app-smoke.ts
```diff
@@ -0,0 +1,678 @@
...
+    return;
+  }
+
+  for (const packageRoot of [owoxRoot, ctlRoot]) {
+    const manifestPath = join(packageRoot, 'oclif.manifest.json');
+    if (existsSync(manifestPath)) {
+      continue;
+    }
+
+    const result = spawnSync('npm', ['run', 'prepack'], {
```
romandubovyi: 🟡 **Important:** Running `npm run prepack` synchronously inside `ensureCliManifests()` can make smoke tests slow and flaky in CI. Consider generating the oclif manifests during test setup or the build phase instead of blocking the test runtime with npm.

apps/backend/test/utils/api-key-app-smoke.ts
```diff
@@ -0,0 +1,678 @@
...
+  expect(status.authenticated).toBe(true);
+  expect(status.apiOrigin).toBe(parsedApiKey.apiOrigin);
+  expect(status.apiKeyId).toBe(parsedApiKey.apiKeyId);
+  expect(status.project).toEqual(expect.any(Object));
+  expect(status.member).toEqual(expect.any(Object));
+
+  const renderedStatus = JSON.stringify(status);
+  expect(renderedStatus.includes(fullApiKey)).toBe(false);
+  expect(renderedStatus.includes(parsedApiKey.apiKeySecret)).toBe(false);
+  expect(renderedStatus.includes('"authFlow"')).toBe(false);
```
romandubovyi: ⚪ **Nit:** The substring check `renderedStatus.includes('authFlow')` is brittle — a future field name containing that substring could break the assertion. Consider asserting on object keys directly, e.g. `expect(status).not.toHaveProperty('authFlow')`.

apps/backend/src/idp/controllers/auth-context.controller.ts
```diff
@@ -0,0 +1,57 @@
+import { Controller, Get } from '@nestjs/common';
+import { ApiHeader, ApiOkResponse, ApiOperation, ApiTags } from '@nestjs/swagger';
+import { Auth, AuthContext } from '../decorators';
+import { AuthorizationContext, Role, Strategy } from '../types';
+
+type AuthContextResponse = {
```
romandubovyi: 💡 **Suggestion:** `AuthContextResponse` duplicates the fields already defined in `AuthorizationContext`. Consider deriving it from the existing type (e.g. `type AuthContextResponse = AuthorizationContext;`) to keep the shape in sync and avoid drift.

apps/backend/src/data-marts/controllers/project-members.controller.ts
```diff
@@ -66,6 +73,7 @@ export class ProjectMembersController {
   ) {}
 
   @Auth(Role.viewer(Strategy.PARSE))
+  @RejectApiKeyAuth()
```
romandubovyi: 🟢 **Praise:** Blocking API-key authentication from member administration, key management, user provisioning, and OAuth authorization is a well-scoped security improvement. The combination of `@RejectApiKeyAuth`, explicit OAuth rejection, header binding, and e2e smoke tests for both IDPs gives strong confidence that exchanged tokens cannot be misused.

packages/idp-better-auth/src/services/page-service.ts
max-voloshyn: Good catch. This is already addressed in the current branch: `req.params.role` is returned directly now, with the query fallback kept for compatibility.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
max-voloshyn: Agreed, and addressed. The OAuth authorization path now throws `AuthorizationError` for API-key auth, so it follows the same 403 authorization-restriction semantics as `@RejectApiKeyAuth` routes.

apps/backend/src/idp/guards/idp.guard.ts
max-voloshyn: Agreed, and addressed. The guard now preserves explicit authorization errors but wraps generic authentication failures back to `AuthenticationError('Authentication failed')`, so the previous generic behavior is retained.

apps/backend/src/idp/guards/idp.guard.ts
max-voloshyn: Agreed, and addressed. `checkApiKeyUsageRestrictions` now runs before header binding, so routes that reject API-key auth surface the endpoint-policy error first.

apps/backend/test/utils/api-key-app-smoke.ts
max-voloshyn: Agreed, and addressed. The smoke helper no longer generates oclif manifests at runtime, avoiding the synchronous npm call during the e2e smoke path.

apps/backend/test/utils/api-key-app-smoke.ts
max-voloshyn: Agreed, and addressed. The smoke assertion now checks JSON structure directly with `expectJsonValueNotToHaveKey(status, 'authFlow')`; string rendering is only used for secret-leak checks.

apps/backend/src/idp/controllers/auth-context.controller.ts
max-voloshyn: Agreed, and addressed. `AuthContextResponse` is now derived from the existing `AuthorizationContext` type.

apps/backend/src/data-marts/controllers/project-members.controller.ts
max-voloshyn: Thanks. The boundary has been kept explicit: API-key auth is rejected from key/member/provisioning/Intercom/OAuth authorization flows while remaining allowed for normal domain-resource APIs.

apps/backend/src/idp/guards/idp.guard.ts
```diff
@@ -62,6 +67,7 @@ export class IdpGuard implements CanActivate {
...
+      this.checkApiKeyUsageRestrictions(tokenPayload, Boolean(rejectApiKeyAuth));
```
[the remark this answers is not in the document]
max-voloshyn: This is intentionally not changed. Contexts, data marts, storages, destinations, and reports are normal domain-resource APIs for this feature, so API-key auth should be allowed there and governed by existing role/capability checks.

packages/idp-better-auth/src/services/token-service.ts
```diff
@@ -34,6 +54,50 @@ export class TokenService {
...
 
+  async issueProjectMemberApiKeyAccessToken(
+    payload: ProjectMemberApiKeyPayload
+  ): Promise<AuthResult> {
+    this.assertProjectMemberApiKeyPayload(payload);
+
+    try {
+      const encryptedToken = await this.cryptoService.encrypt(JSON.stringify(payload));
+      return {
+        accessToken: encryptedToken,
```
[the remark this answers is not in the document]
max-voloshyn: Partially addressed within this PR’s intended scope. Better Auth API-key tokens now carry a short 15-minute expiry, and `/api/auth/context` uses introspection to refresh current member/user context. Immediate revocation of already-issued tokens beyond that TTL window is out of scope for this change.

apps/backend/src/data-marts/controllers/data-storage.controller.ts
```diff
@@ -106,6 +107,21 @@ export class DataStorageController {
...
 
+  private rejectApiKeyAuthForAccessControlUpdate(
+    context: AuthorizationContext,
+    dto: Pick<CreateDataStorageApiDto | UpdateDataStorageApiDto, 'ownerIds'> &
+      Partial<Pick<UpdateDataStorageApiDto, 'contextIds'>>
+  ): void {
+    if (
+      context.authFlow === 'api_key' &&
+      (dto.ownerIds !== undefined || dto.contextIds !== undefined)
+    ) {
```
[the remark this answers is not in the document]
max-voloshyn: This is intentionally not changed for the same reason as the context-membership thread. Sharing/configuration changes on storages and destinations are part of the domain-resource management surface that API-key auth should be able to use, subject to normal role/capability checks.

packages/idp-better-auth/src/providers/better-auth-provider.ts
```diff
@@ -234,7 +236,7 @@ export class BetterAuthProvider
   }
 
   async introspectToken(token: string): Promise<Payload | null> {
-    return this.tokenService.introspectToken(token);
+    return this.refreshIntrospectedPayload(await this.tokenService.introspectToken(token));
```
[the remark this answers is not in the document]
max-voloshyn: This is an accepted tradeoff for this PR. `/api/auth/context` intentionally uses introspection and returns fresh user/member context; regular `Strategy.PARSE` routes continue to use token claims until the short 15-minute token expiry. Forcing all API-key route authorization through introspection would change the existing strategy semantics and is out of scope here.

PR #1365, opened 2026-06-26
docs: improve output controls, sharing, and Ads credential guides

prokopetsv:
# Problem

Several documentation pages had gaps and inaccuracies that slowed readers. The Output Controls guide explained the post-schema-change validation error in a single sentence, with no coverage of the **Disconnected columns** warning users actually see first in the column picker. The Ownership and Sharing reference had no section on Joinable Data Mart relationships, leaving the create/edit/delete permission rules undocumented, and it used the undefined abbreviation "DM" in several tables. The Microsoft Ads credentials guide had broken step numbering (jumping 1, 2, 5, 3, 4), missing Body content-type instructions, and multiple typos. The X Ads credentials guide listed a "Generate Access Token" step (Step 4) whose output was never consumed anywhere in the flow, and its Step 5 error guidance referenced fields that step no longer sent.

# Solution

- **Output Controls:** Split the schema-change section into two clearly separated error states — the inline "Disconnected columns" warning (with fix options) and the save-time "Output controls validation failed" error — each with an annotated screenshot and clear next steps.
- **Ownership and Sharing:** Added a "Joinable Data Mart" section documenting the permission model verified against the backend services (create requires `Edit` on both source and target; edit/delete requires `Edit` on source only; using joined fields follows normal Data Mart visibility). Spelled out "DM" as "Data Mart" across all affected tables.
- **Microsoft Ads credentials:** Reordered and renumbered the steps into a logical 1–7 sequence (Azure app → credentials → auth code → token exchange → account/developer tokens), fixed heading levels, added the `x-www-form-urlencoded` Body instruction, added a "Before you start" tooling note, and corrected typos.
- **X Ads credentials:** Removed the orphaned "Generate Access Token" step (its tokens were never used; the connector consumes the permanent tokens from the final OAuth exchange), renumbered the remaining steps, and rewrote the Step 5 error guidance to reference only the Consumer Key/Secret that step actually sends. Updated the matching step references in `GETTING_STARTED.md`.
- Applied a consistent style pass to all new prose: max 20-word sentences, active voice, no nominalizations or filler, while preserving verbatim product error strings.

# Changes

- Added "Disconnected columns warning" and "Validation error on save" subsections to `output-controls.md` with screenshots and alt text
- Added a "Joinable Data Mart" permission section to `ownership-and-sharing.md`
- Replaced the abbreviation "DM" with "Data Mart" throughout `ownership-and-sharing.md` tables
- Reordered and renumbered the Microsoft Ads credential steps into a sequential 1–7 flow
- Added an `x-www-form-urlencoded` Body-type instruction and a "Before you start" tooling note to the Microsoft Ads guide
- Fixed heading levels and typos (e.g. "Reqibin", "authorixation", "Clicks") in the Microsoft Ads guide
- Removed the orphaned "Generate Access Token" step from the X Ads guide and renumbered Steps 4–6
- Rewrote the X Ads Step 5 error guidance to reference only the credentials that request sends
- Updated step references in `XAds/GETTING_STARTED.md` to match the renumbered guide
- Rewrote all new prose to meet the readability rules (short active sentences, no passive voice or filler)

### DIFF
--- a/docs/getting-started/setup-guide/output-controls.md
+++ b/docs/getting-started/setup-guide/output-controls.md
@@ -162,7 +162,24 @@ OWOX applies output controls in this order on every report run:
 
 ## Keep Rules Valid After Schema Changes
 
-Output controls reference columns by name. If you remove or rename a column that a filter, slice, or sort rule points to, the next run fails with a validation error. Open the report, then update or delete the affected rule.
+Output controls reference columns by name. Rename or remove a column in the Data Mart schema, and every report that uses it breaks. The column picker flags the report with a **Disconnected columns** warning. Saving the report then fails with a validation error.
+
+### Disconnected columns warning
+
+Open **Edit report**. The column picker groups the missing columns under a red **Disconnected columns** label with a ⚠ icon. Hover the icon to see:
+
+> *They are missing from the current Data Mart output schema. Uncheck them to remove them from the report, or contact your analyst to restore the schema.*
+
+You have two options:
+
+- **Uncheck the disconnected columns** — this drops them from the report and removes any filter, slice, or sort rule that uses them. The report then runs on the remaining valid columns.
+- **Restore the schema** — does the column still belong? Ask whoever manages the Data Mart to add it back. Then reopen the report.
+
+![Edit report panel with a red "Disconnected columns" group at the top of the column list, containing order_date with a checked checkbox. A tooltip is open showing "They are missing from the current Data Mart output schema. Uncheck them to remove them from the report, or contact your analyst to restore the schema." The remaining columns (order_id, customer_id, order_timestamp, product_id, product_name, category, customer_name, country) are listed below and appear valid.](https://imagedelivery.net/zKr-4bdC5CBGL2DuuEmvYw/6616af2b-e216-406b-b11a-e876b17df900/public)
+
+### Validation error on save
+
+A filter, slice, or sort rule may still point to a disconnected column. Save the report, and OWOX blocks it with **"Output controls validation failed"**. Open the report, remove the affected rules from the Filters, Slices, or Sort sections, then save again.
 
 !["Output controls validation failed" error banner at the top of the page. The edit report panel shows the Sort section with two rules — category and payment_method — both highlighted in red with warning icons.](https://imagedelivery.net/zKr-4bdC5CBGL2DuuEmvYw/2342964b-37d0-4123-32c8-2a1ddbe6d400/public)
 
--- a/docs/getting-started/setup-guide/output-controls.md
+++ b/docs/getting-started/setup-guide/output-controls.md
@@ -172,7 +172,7 @@ Open **Edit report**. The column picker groups the missing columns under a red *
 
 You have two options:
 
-- **Uncheck the disconnected columns** — this drops them from the report and removes any filter, slice, or sort rule that uses them. The report then runs on the remaining valid columns.
+- **Uncheck the disconnected columns** — this removes them from the report selection only. It does not clear filter, slice, or sort rules that reference them. Open the Filters, Slices, and Sort sections, delete those rules too, then save.
 - **Restore the schema** — does the column still belong? Ask whoever manages the Data Mart to add it back. Then reopen the report.
 
 ![Edit report panel with a red "Disconnected columns" group at the top of the column list, containing order_date with a checked checkbox. A tooltip is open showing "They are missing from the current Data Mart output schema. Uncheck them to remove them from the report, or contact your analyst to restore the schema." The remaining columns (order_id, customer_id, order_timestamp, product_id, product_name, category, customer_name, country) are listed below and appear valid.](https://imagedelivery.net/zKr-4bdC5CBGL2DuuEmvYw/6616af2b-e216-406b-b11a-e876b17df900/public)
--- a/docs/project/ownership-and-sharing.md
+++ b/docs/project/ownership-and-sharing.md
@@ -136,8 +136,13 @@ The following actions can be granted or restricted by the combination of ownersh
 | Who | Not shared | Shared for use | Shared for maintenance | Shared for both |
 |---|---|---|---|---|
 | **Owner (Technical User)** | All actions | All actions | All actions | All actions |
-| **Non-owner Technical User** | No access | See, Use | See, Use, Copy Credentials, Edit, Delete | See, Use, Copy Credentials, Edit, Delete |
-| **Any Business User** | No access | No access | No access | No access |
+| **Owner (Business User)** | No access [1] | No access [1] | No access [1] | No access [1] |
+| **Non-owner Technical User** | No access | See, Use [2] | See, Use, Copy Credentials, Edit, Delete [2] | See, Use, Copy Credentials, Edit, Delete [2] |
+| **Any Business User (non-owner)** | No access | No access | No access | No access |
+
+[1] A Business User assigned as Storage Owner has no access until their role changes to Technical User. A warning appears on the [Notification Settings](../notifications/notification-settings.md) page.
+
+[2] Under `Selected contexts only` scope, these actions require a context overlap between the member and the Storage.
 
 > ☝️ Business Users do not have direct access to Storages. They interact with data through Data Marts and Destinations shared with them.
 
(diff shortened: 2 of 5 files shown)


docs/project/ownership-and-sharing.md
```diff
@@ -173,16 +173,44 @@ Data Mart Triggers have no dedicated ownership. Who can see them and who can man
...
+**Using joined fields in a report** does not require maintenance access. Any member who sees the source Data Mart can pick joined fields in the Report Columns picker. Two conditions apply:
+
+- The target Data Mart is also visible to them, through ownership or sharing.
+- The relationship is complete: it has at least one join condition, and the target Data Mart is published (not a draft).
+
+| Who | Can see relationships | Can create | Can edit or delete |
+|---|---|---|---|
+| **Technical Owner (Technical User)** of source Data Mart | Yes | Yes, if also has `Edit` on target Data Mart | Yes |
+| **Technical Owner (Business User)** of source Data Mart | Yes | No | No |
+| **Business Owner (any role)** of source Data Mart | Yes | No | No |
```
max-voloshyn: 🔴 Blocking: This row says Business Owners of any role cannot create, edit, or delete relationships, but a Business Owner who is also a Technical User can receive Data Mart Edit through the shared-for-maintenance non-owner path documented above. Relationship create/update/delete checks use `Action.EDIT`, so this row should split Business Owner (Technical User) from Business Owner (Business User), or make the cells conditional.
🤖 Reviewed by Codex (GPT-5.5)

packages/connectors/src/Sources/XAds/CREDENTIALS.md
```diff
@@ -147,18 +132,20 @@ This is a single string with values separated by `&`. Extract both values and ke
...
 - `oauth_token_secret` — the value between `oauth_token_secret=` and `&oauth_callback_confirmed`. In the example above: `UlDQaqOoJHj1VvLQ8fQH6Iq686rEFww2`
 
-![Postman showing POST request to oauth/request_token with OAuth 1.0 selected, all fields filled, and the 200 OK response containing oauth_token and oauth_token_secret](https://imagedelivery.net/zKr-4bdC5CBGL2DuuEmvYw/0065f031-a718-41db-ceae-a5951d862f00/public)
+![Postman showing POST request to oauth/request_token with OAuth 1.0 selected, all fields filled, and the 200 OK response containing oauth_token and oauth_token_secret](https://imagedelivery.net/zKr-4bdC5CBGL2DuuEmvYw/5f022030-e93b-45c0-cd89-831180b19f00/public)
 
 **Possible errors:**
 
-- **"Invalid or expired token."** — Your Access Token or Token Secret is wrong or expired. [Regenerate them in Step 4](#step-4-generate-access-token-and-token-secret) and try again.
-- **"Could not authenticate you."** — Your Consumer Key or Consumer Secret is incorrect. Re-enter them and try again.
+This request sends only your Consumer Key and Consumer Secret. So an authentication error points to one of those two values:
```
max-voloshyn, line 141: 🔴 Blocking: This says the request sends only the Consumer Key and Consumer Secret, but the PIN-based OAuth 1.0a flow needs an OAuth callback, typically `oauth_callback=oob`, and the app must have OAuth/user authentication configured. Without that setup, users may not get the temporary token and verifier needed for Steps 5-6. Please add the callback/OOB instructions before telling users to send this request.
🤖 Reviewed by Codex (GPT-5.5)

packages/connectors/src/Sources/MicrosoftAds/CREDENTIALS.md
```diff
@@ -57,31 +59,34 @@ At this point, you have:
...
+```
 
-Click **Request Token**, and copy the generated **Developer Token**.  
+![Microsoft Make Query](res/microsoft_makequery.png)
 
-![Microsoft Request](res/microsoft_request.png)
+![Microsoft Ampersand](res/microsoft_ampersand.png)
 
-## Step 5: Generate an Authorization Code
+## Step 4: Get the Authorization Code
```
max-voloshyn, line 87: 🔴 Blocking: Step 4 now runs before users collect the Account ID and Customer ID, but it only says to log in and authorize the app. The refresh token represents the signed-in Microsoft Advertising user, and the connector later uses it with the configured CustomerID/AccountIDs. Please either move the account/customer lookup before OAuth again or explicitly tell users to sign in here with the same Microsoft Ads user that has access to those IDs.
🤖 Reviewed by Codex (GPT-5.5)

docs/project/ownership-and-sharing.md
```diff
@@ -173,16 +173,44 @@ Data Mart Triggers have no dedicated ownership. Who can see them and who can man
...
 
+### Joinable Data Mart
+
+Relationships between Data Marts have no dedicated ownership. They belong to the source Data Mart and follow its access rules. See [Joinable Data Marts](../getting-started/setup-guide/joinable-data-marts.md) to configure them.
+
+**Creating a relationship** requires `Edit` access on both the source and target Data Marts. The user needs maintenance access on each Data Mart, through one of two paths. First: a Technical Owner with the Technical User role. Second: a Technical User on a Data Mart *Shared for maintenance*.
+
+**Editing or deleting a relationship** requires `Edit` access on the source Data Mart only.
+
+**Using joined fields in a report** does not require maintenance access. Any member who sees the source Data Mart can pick joined fields in the Report Columns picker. Two conditions apply:
```
max-voloshyn: 🟡 Non-blocking: These two conditions are not sufficient for "can pick joined fields." Even when the target Data Mart is visible and the relationship is complete, joined fields disappear if Allow for reporting is off or individual fields are hidden. Please add that caveat or scope this paragraph to access-only conditions.
🤖 Reviewed by Codex (GPT-5.5)

docs/getting-started/setup-guide/output-controls.md
```diff
@@ -162,7 +162,24 @@ OWOX applies output controls in this order on every report run:
...
+
+### Disconnected columns warning
+
+Open **Edit report**. The column picker groups the missing columns under a red **Disconnected columns** label with a ⚠ icon. Hover the icon to see:
+
+> *They are missing from the current Data Mart output schema. Uncheck them to remove them from the report, or contact your analyst to restore the schema.*
+
+You have two options:
+
+- **Uncheck the disconnected columns** — this drops them from the report and removes any filter, slice, or sort rule that uses them. The report then runs on the remaining valid columns.
```
max-voloshyn: 🟡 Non-blocking: Unchecking a disconnected selected column only removes that column from the report selection; it does not remove filters, slices, or sort rules that reference it. Users still need to clear those output-control rules before save succeeds. Please replace the automatic-removal claim with explicit cleanup guidance.
🤖 Reviewed by Codex (GPT-5.5)

packages/connectors/src/Sources/MicrosoftAds/CREDENTIALS.md
```diff
@@ -57,31 +59,34 @@ At this point, you have:
...
 
-## Step 4: Get Your Developer Token
+Open the **Body** tab. Set the body type to **x-www-form-urlencoded**. In ReqBin, choose **Form** as the content type. Add these parameters, replacing _YOUR_CLIENT_ID_, _YOUR_CLIENT_SECRET_, and _YOUR_AUTHORIZATION_CODE_ with real values:
 
-In the Microsoft Ads interface, go to **Settings → Developer Settings**.  
+> **Note:** You'll get the **Authorization Code** from the URL in Step 4. Prepare the request now. The code is short-lived, so a ready request lets you paste and send right away.
 
-![Microsoft Developer](res/microsoft_developer.png)
+``` code
+client_id=YOUR_CLIENT_ID&
```
max-voloshyn: 🟡 Non-blocking: These newly added lines contain trailing whitespace, and `git diff --check origin/main...HEAD` fails on this file. Please remove the hard-break spaces from the request example and nearby moved lines so standard git whitespace checks pass.
🤖 Reviewed by Codex (GPT-5.5)

prokopetsv: Done — removed the trailing whitespace from the request-example lines and the moved prose lines. `git diff --check main...HEAD` now passes clean on this file.

docs/getting-started/setup-guide/output-controls.md
prokopetsv: Confirmed in `ReportColumnPicker` — `toggleField` on uncheck only filters the column out of the selection array; it never touches `filterConfig` or sort rules. I replaced the automatic-removal claim with explicit guidance: unchecking removes the column from the selection only, and the user must clear the referencing rules in the Filters, Slices, and Sort sections before save succeeds.

docs/project/ownership-and-sharing.md
prokopetsv: You're right — visibility plus a complete relationship are necessary but not sufficient. I scoped those two bullets as the access conditions and added a caveat that a field also needs **Allow for reporting** on and must not be hidden via **Hide from reports**, pointing to the Joinable Data Marts guide for those field-exposure controls (kept the mechanics there rather than duplicating them in the permissions article).

packages/connectors/src/Sources/MicrosoftAds/CREDENTIALS.md
prokopetsv: Good catch — the refresh token is tied to the signed-in user, and the connector uses it with the configured CustomerId/AccountIds. Rather than revert the step order, I added a warning in Step 4 telling users to sign in with the Microsoft Advertising user that can access the Account ID and Customer ID they collect in Step 6, and explaining that the token only reaches accounts that user can access.

packages/connectors/src/Sources/XAds/CREDENTIALS.md
prokopetsv: The flow works as written because the app created in Step 1 already has OAuth 1.0a user authentication configured — X confirms this with `oauth_callback_confirmed=true` in the Step 4 response, and Step 5 already covers both the PIN and the `oauth_verifier` redirect. To cover apps that aren't set up that way, I added a note in Step 4 pointing to the user-authentication settings and the `oauth_callback=oob` fallback, rather than adding mandatory callback steps to the verified path.

docs/project/ownership-and-sharing.md
prokopetsv: Done — fixed.

You're right: relationship create/update/delete gate on `Action.EDIT`, so a Business Owner with the **Technical User** role gets that `Edit` via the shared-for-maintenance non-owner path and can manage relationships. I split the `Business Owner (any role)` row into Technical User (shared-for-maintenance / not) and Business User cases, matching the **Data Mart** and **Data Mart Trigger** tables.

Also audited the other tables for the same pattern: Data Mart, Data Mart Trigger, Report, and Report Trigger are already correct; Destination's `Owner (any role)` collapse is intentional; Storage had a related gap (missing `Owner (Business User)` case + context-gate caveat), now fixed too.

max-voloshyn (approved):

PR #1366, opened 2026-06-26
feat(mcp): add OpenAI Apps domain-verification endpoint

zapolsky:
## Summary

Adds a public endpoint so the OWOX MCP server can pass **OpenAI Apps domain verification** during app submission. OpenAI pings an origin-root well-known URL on the MCP host and expects the verification token back as plain text with a 200.

- **Route:** `GET /.well-known/openai-apps-challenge` → returns the token as `text/plain` (HTTP 200), served at the **host root** (not under `/api`).
- **Public:** no auth — added to `PROTOCOL_ROUTE_EXCLUSIONS` so it bypasses the `/api` prefix, same as the existing OAuth well-known metadata routes.
- **Config-driven:** the token comes from the new `MCP_OPENAI_APPS_CHALLENGE_TOKEN` env var (rotatable without a code change). Returns **404** until the token is configured, so we never publish an empty challenge.

## Out of scope

This PR is **only** the domain-verification endpoint. The separate `GET /oauth/authorize` HTTP 500 seen in the ChatGPT Apps OAuth flow is **not** addressed here — its root cause is upstream in the Java Identity backend (`Data too long for column 'state'` when persisting the MCP OAuth authorization code for OpenAI's long relay `state`), which needs a fix in that repo.

## Testing

- Unit (`mcp-metadata.controller.spec.ts`): returns the configured token verbatim; responds 404 when unset.
- E2E (`mcp-oauth.e2e-spec.ts`): serves the token as `text/plain` at the origin-root path; confirms it is **not** exposed under `/api`.
- `eslint` clean; `tsc -b` introduces no new errors.

## Rollout (infra — required before OpenAI can verify)

- [ ] Set `MCP_OPENAI_APPS_CHALLENGE_TOKEN` (value from OpenAI's submission UI) in the production `mcp.owox.com` backend
- [ ] Deploy
- [ ] Ensure `/.well-known/openai-apps-challenge` is publicly reachable on the `mcp.owox.com` root (no auth / no WAF block)

### DIFF
--- a/apps/backend/src/ee/mcp/controllers/mcp-metadata.controller.ts
+++ b/apps/backend/src/ee/mcp/controllers/mcp-metadata.controller.ts
@@ -32,6 +32,7 @@ export class McpMetadataController {
    */
   @Get(OPENAI_APPS_CHALLENGE_PATH)
   @Header('Content-Type', 'text/plain; charset=utf-8')
+  @Header('Cache-Control', 'no-store')
   getOpenaiAppsChallenge(): string {
     const token = this.config.openaiAppsChallengeToken;
     if (!token) {
--- a/apps/backend/src/ee/mcp/controllers/mcp-metadata.controller.ts
+++ b/apps/backend/src/ee/mcp/controllers/mcp-metadata.controller.ts
@@ -1,4 +1,4 @@
-import { Controller, Get } from '@nestjs/common';
+import { Controller, Get, Header, NotFoundException } from '@nestjs/common';
 import { OAuthProtectedResourceMetadataSchema } from '@owox/idp-protocol';
 import { McpConfigService } from '../config/mcp.config';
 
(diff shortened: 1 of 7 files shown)


max-voloshyn (approved):

apps/backend/src/ee/mcp/controllers/mcp-metadata.controller.ts
```diff
@@ -21,4 +23,20 @@ export class McpMetadataController {
...
   }
+
+  /**
+   * OpenAI Apps domain verification. OpenAI pings this origin-root well-known URL
+   * and expects the verification token back as plain text with a 200. Public (no
+   * auth, no `/api` prefix — see PROTOCOL_ROUTE_EXCLUSIONS). Returns 404 until the
+   * token is configured so we never publish an empty challenge.
+   */
+  @Get(OPENAI_APPS_CHALLENGE_PATH)
+  @Header('Content-Type', 'text/plain; charset=utf-8')
```
romandubovyi, line 34: ⚪ Nit:  Add `@Header('Cache-Control', 'no-store')` to prevent proxies or clients from caching the verification token. A cached stale token could cause future domain-verification failures or leak a rotated token.

romandubovyi (approved):

apps/backend/src/ee/mcp/controllers/mcp-metadata.controller.ts
zapolsky: Done — added `@Header('Cache-Control', 'no-store')` to the challenge handler in 3c8e833, plus an e2e assertion that the response carries `Cache-Control: no-store`.

PR #1368, opened 2026-06-26
fix(scripts): make npm scripts cross-platform on Windows

zapolsky:
## Problem
Several npm scripts set environment variables using the Unix-only `VAR=value command` form, which fails on Windows (cmd/PowerShell). A contributor running Windows 11 hit this on the first `npm run dev -w owox` and had to patch the script by hand.

## Fix
Wrap the affected scripts with `cross-env`, which is already a root devDependency (hoisted onto PATH for every workspace) — so there is no new dependency and no lockfile change:

| File | Script |
|------|--------|
| `apps/owox` | `dev` |
| `apps/ctl` | `test` |
| `packages/internal-helpers` | `test` |
| `packages/idp-better-auth` | `test` |
| `packages/idp-protocol` | `test` |
| `packages/api-client` | `test` |
| `packages/idp-owox-better-auth` | `test` |

The `test` scripts carried the same Unix-only `NODE_OPTIONS=...` pattern and would fail on Windows too, so they are fixed in the same pass.

## Verification
- All changed `package.json` files parse as valid JSON.
- `npm run test -w @owox/idp-protocol` passes with cross-env (`NODE_OPTIONS` forwarded correctly, 17 tests green).
- Changeset added (`owox`: minor).

### DIFF
--- a/.changeset/windows-cross-platform-npm-scripts.md
+++ b/.changeset/windows-cross-platform-npm-scripts.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Cross-platform npm scripts on Windows
+
+Wrap inline environment-variable assignments in npm scripts with `cross-env` so they run on Windows (cmd/PowerShell), not only on Unix shells. Previously `npm run dev -w owox` and the package `test` scripts failed on Windows because the `VAR=value command` form is Unix-only syntax, forcing contributors to edit the scripts by hand. Windows contributors can now run development and test scripts out of the box.
--- a/.changeset/windows-cross-platform-npm-scripts.md
+++ b/.changeset/windows-cross-platform-npm-scripts.md
@@ -1,7 +1,0 @@
----
-'owox': minor
----
-
-# Cross-platform npm scripts on Windows
-
-Wrap inline environment-variable assignments in npm scripts with `cross-env` so they run on Windows (cmd/PowerShell), not only on Unix shells. Previously `npm run dev -w owox` and the package `test` scripts failed on Windows because the `VAR=value command` form is Unix-only syntax, forcing contributors to edit the scripts by hand. Windows contributors can now run development and test scripts out of the box.
--- a/packages/idp-protocol/package.json
+++ b/packages/idp-protocol/package.json
@@ -14,7 +14,7 @@
   "scripts": {
     "build": "tsc",
     "clean": "shx rm -rf dist",
-    "test": "NODE_OPTIONS=--experimental-vm-modules jest --passWithNoTests",
+    "test": "cross-env NODE_OPTIONS=--experimental-vm-modules jest --passWithNoTests",
     "lint": "eslint . --config ./eslint.config.js",
     "lint:fix": "eslint . --fix --config ./eslint.config.js",
     "format": "prettier --write \"**/*.{ts,js,json,md}\" --ignore-path ../../.prettierignore",
--- a/tools/setup-husky.mjs
+++ b/tools/setup-husky.mjs
@@ -7,7 +7,7 @@
  * and sets up the necessary files for pre-commit validation.
  */
 
-import { chmodSync, existsSync, mkdirSync, writeFileSync } from 'fs';
+import { chmodSync, existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
 import { platform } from 'os';
 import { dirname, join } from 'path';
 import { fileURLToPath } from 'url';
@@ -42,36 +42,24 @@ function makeExecutable(filePath) {
 }
 
 /**
- * Generate cross-platform hook content
+ * Generate hook content
  * @param {string} command - Command to run in the hook
  * @returns {string} Hook content
+ *
+ * Git runs hooks through sh on every OS (including the sh.exe bundled with
+ * Git for Windows), so a hook is always a POSIX shell script — never a Windows
+ * batch file. Husky v9+ hooks contain only the command; the old
+ * `#!/usr/bin/env sh` + `. "$(dirname -- "$0")/_/husky.sh"` boilerplate is
+ * deprecated and is removed in husky v10.
  */
 function generateHookContent(command) {
-  if (isWindows) {
-    // Windows batch script style
-    return `@echo off
-:: Husky hook for Windows
-cd /d "%~dp0"
-call husky.cmd
-${command}
-`;
-  } else {
-    // Unix shell script style
-    return `#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-# ${command.split('\n')[0].replace('# ', '')}
-${command}
-`;
-  }
+  return `${command}\n`;
 }
 
 /**
  * Create pre-commit hook
  */
 function createPreCommitHook() {
-  console.log('🪝 Creating pre-commit hook...');
-
   const huskyDir = join(repoRoot, '.husky');
   if (!existsSync(huskyDir)) {
     mkdirSync(huskyDir, { recursive: true });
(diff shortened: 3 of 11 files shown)


.changeset/windows-cross-platform-npm-scripts.md
```diff
@@ -0,0 +1,7 @@
+---
+'owox': minor
```
laskevych: **Changeset level/scope: this PR changes nothing in the published `owox` artifact, yet declares a `minor` bump.**

`apps/owox`'s `files` field is `["./bin", "./dist/**/*.js", "./oclif.manifest.json", "./npm-shrinkwrap.json"]`, so the edited `package.json` scripts and `tools/setup-husky.mjs` never ship — a consumer who installs `owox` gets zero behavioral change from this PR.

Two consequences of `'owox': minor`:
- `.changeset/config.json` puts `owox` in a single `fixed` group with all 10 publishable packages, so this forces a synchronized **minor** bump + re-publish of every one of them (`@owox/backend`, `@owox/web`, `@owox/api-client`, the idp packages, …) even though their shipped contents are byte-for-byte identical.
- `@changesets/cli/changelog` writes a user-facing "Minor Changes" bullet into `apps/owox/CHANGELOG.md` (whose other entries are all real end-user features) describing internal Windows dev-tooling.

Suggest downgrading to `patch` — or dropping the changeset entirely, since no published artifact changes. Also note this PR adds a **second** changeset (`windows-cross-platform-git-hooks.md`) that also declares `'owox': minor`; both emit separate changelog bullets, so consider merging them into one.

tools/setup-husky.mjs
```diff
@@ -42,36 +42,24 @@ function makeExecutable(filePath) {
...
-  } else {
-    // Unix shell script style
-    return `#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-# ${command.split('\n')[0].replace('# ', '')}
-${command}
-`;
-  }
+  return `${command}\n`;
```
laskevych, line 37: Dropping the shebang + `husky.sh` sourcing is correct: with `core.hooksPath=.husky/_`, git runs the husky `_/h` wrapper, which executes the user hook via `sh -e "$s"` — so a shebang-less command file is valid husky-v9 format, and removing the `. "$(dirname -- "$0")/_/husky.sh"` line legitimately clears the `husky - DEPRECATED` (v10) warning. 👍

Follow-up cleanup now that the hook is always POSIX `sh` and never exec'd directly: the `isWindows` branch and `makeExecutable()` / `chmodSync(…, 0o755)` on `.husky/pre-commit` (lines 22, 28-42, 82) are no longer load-bearing — the `+x` bit is never consulted because git reads the file through `sh -e`. Your own new doc comment ("Git runs hooks through sh on every OS") makes the leftover `chmod` self-inconsistent. Optional: drop `makeExecutable`/`isWindows` (keep `platform()` for the log on line 94). Pre-existing and non-blocking.

packages/idp-protocol/package.json
```diff
@@ -14,7 +14,7 @@
   "scripts": {
     "build": "tsc",
     "clean": "shx rm -rf dist",
-    "test": "NODE_OPTIONS=--experimental-vm-modules jest --passWithNoTests",
+    "test": "cross-env NODE_OPTIONS=--experimental-vm-modules jest --passWithNoTests",
```
laskevych, line 17: `cross-env` is only a **root** devDependency; none of the 7 packages that now invoke it declare it locally. Inside the workspace it resolves via root `node_modules/.bin` hoisting (that's why `npm run test -w @owox/idp-protocol` passes), so this is fine in practice.

Worth noting the asymmetry though: `jest` is a local devDependency here while `cross-env` is not, so running this package's `test` in isolation (e.g. the package extracted from the monorepo) would fail with exit 127. If you want each package's scripts to be self-contained, add `cross-env` to their `devDependencies`; otherwise this matches the repo's existing root-tooling convention and is fine to leave. Low priority — applies to all 7 converted scripts.

laskevych:
## Code review — no correctness issues found ✅

Reviewed both commits (cross-env for inline env vars across 7 `package.json` files, and the POSIX `sh` husky hook rewrite). The fix is well-reasoned and correct:

- **husky hook** — With `core.hooksPath=.husky/_`, git runs the husky `_/h` wrapper, which executes the user hook via `sh -e "$s"`. So the shebang-less `npm run pre-commit` content is valid husky-v9 format, and dropping the `. "$(dirname -- "$0")/_/husky.sh"` line legitimately clears the `husky - DEPRECATED`/v10 warning. The self-heal comparison is stable (`.husky/` is gitignored, so `autocrlf` can't perturb the string compare).
- **cross-env** — Resolves in every workspace via root `node_modules/.bin` hoisting. I confirmed all 7 inline `VAR=value` scripts in the repo are covered — none were missed, and none use chained (`&&`/`;`) inline assignments that would need extra care.

Nothing blocks merge. A few non-blocking notes left inline:

1. **Changeset scope/level** (main point) — none of the changed files are part of the published `owox` artifact (`files` field excludes scripts + `tools/`), yet both changesets declare `'owox': minor`. Via the `fixed` group this re-publishes all 10 packages with identical contents and writes user-facing changelog bullets for internal tooling. Consider `patch` (or no changeset), and merging the two changesets into one.
2. **Vestigial platform code** — `isWindows`/`makeExecutable`/`chmod` in `setup-husky.mjs` are no longer load-bearing now that the hook runs via `sh -e`. Optional cleanup.
3. **`cross-env` not declared locally** — works via hoisting today; flagged only for isolation/self-containment. Low priority.

*Reviewed with a multi-angle pass (line-by-line, removed-behavior, cross-file/lifecycle, shell/JS pitfalls, cleanup/altitude, conventions) with adversarial verification of each finding.*

tools/setup-husky.mjs
```diff
@@ -81,29 +69,36 @@ function createPreCommitHook() {
...
+  // Self-healing: only (re)write when the hook is missing or stale, so a hook
+  // left over from an older setup (the broken Windows batch file or the
+  // deprecated husky v9 boilerplate) is repaired on the next install.
+  if (existsSync(hookPath) && readFileSync(hookPath, 'utf8') === hookContent) {
+    console.log('ℹ️  Pre-commit hook is already up to date.');
+    return false;
+  }
+
+  console.log('🪝 Writing pre-commit hook...');
   writeFileSync(hookPath, hookContent);
```
romandubovyi, line 91: 🔴 **Blocking:** This overwrites any customized pre-commit hook on every `npm install`, not only the known stale formats described above. Please only replace recognized legacy generated content; otherwise preserve the existing hook or warn the user.

zapolsky: Fixed in 4f00eea5f. `createPreCommitHook` now reads the existing hook and overwrites it only when it's missing or its content matches one of the formats this script itself produced before — collected in a new `knownGeneratedHooks()` helper (current husky-v9 command-only, the legacy `. "$(dirname -- "$0")/_/husky.sh"` boilerplate, and the old `@echo off` Windows batch). A hand-edited hook matches none of them, so it's now left untouched with a warning instead of being clobbered on every `npm install`.

Verified across six cases: missing → create, current → no-op, both legacy formats → self-heal, and two hand-edited variants → preserved.

.changeset/windows-cross-platform-npm-scripts.md
zapolsky: Merged the two changesets into one (`windows-cross-platform-support.md`) in 4f00eea5f. Keeping the bump at `minor` deliberately as a platform-support entry — acknowledged that the shipped `owox` artifact bytes don't change; the single changeset keeps this to one changelog bullet instead of two.

tools/setup-husky.mjs
zapolsky: Done in 4f00eea5f. Dropped `isWindows`, `makeExecutable`, the `chmodSync(…, 0o755)` call and the now-unused `chmodSync` import, with a short note in its place on why no chmod is needed (git reads the hook through `sh -e`). `platform()` stays for the startup log line.

packages/idp-protocol/package.json
zapolsky: Leaving these root-hoisted rather than adding `cross-env` to each package's `devDependencies`. Two reasons: it preserves this PR's "no new dependency, no lockfile change" property (declaring it in 7 packages would touch `package-lock.json`), and it matches the repo's existing root-tooling convention — `shx`, `tsc` etc. resolve the same way. The isolated-extraction case you flagged is real, but running a single package fully detached from the monorepo isn't a workflow we support today, so this is a conscious trade-off rather than an oversight.

amarchenk0 (approved):

PR #1369, opened 2026-06-29
feat(backend/web): Google Sheets auto-creation with Drive folder support

laskevych:
## What & why

Adds a "Create document" flow that auto-creates a Google Sheet for a Google
Sheets destination and fills the report's Document Link — no more manually
creating a spreadsheet and pasting its URL. Auto-created sheets can be placed
in a chosen Google Drive folder.

## How it works

Two auth paths, resolved explicitly by the destination's credential type:

- **OAuth** — the sheet is created in the connected user's Drive. With the
  `drive.file` scope (now added to the default OAuth scopes) the file is created
  via the Drive API so it's app-authorized and shareable; a folder can be
  selected through the **Google Drive Picker**. Tokens issued before this change
  fall back to a Sheets-API create in Drive root (sharing skipped) until the user
  reconnects.
- **Service Account** — the sheet is created inside a configured **Shared Drive**
  folder (the SA must be a member with Content Manager). The folder is validated
  at save time so misconfiguration fails fast.

After creation the document is shared (best-effort, writer) with the requesting
user so non-owners can open it.

## Backend
- `POST /data-destinations/:id/google-sheets/documents` (Action.USE) →
  `CreateGoogleSheetDocumentService`.
- New `data_destination.config` JSON column (`{ folderUrl, folderId }`);
  `folderId` is derived server-side from a validated `folderUrl` (migration
  included).
- `GoogleSheetsApiAdapter`: Drive-API create / create-in-folder / share /
  folder-access methods; separate Drive-scoped SA client so existing SA
  operations keep their narrow scope.
- `GoogleSheetsFolderValidator` validates the Shared Drive folder on save.

## Frontend
- "Create document" button on the Google Sheets report form.
- Drive Picker folder selection for OAuth destinations; pasteable Shared Drive
  folder URL for Service Account destinations.
- Picker renders in `<body>`; the destination Sheet stays open during picking.

## Review hardening (commit "harden Google Sheets auto-creation after review")
- Persist `config` on the fresh-credentials create path (a new SA destination no
  longer silently drops its folder).
- Validate the Drive folder only when its id changes on update (skips a live
  Drive round-trip on rename/availability edits).
- `getFolderAccess` rethrows transient errors (429/5xx) instead of reporting
  "folder not found / SA not a member".
- Don't swallow SA credential-resolver errors (infra faults surface instead of
  "not configured").
- Path-specific error hint for folder-create failures (no "service account"
  advice on the OAuth path).
- Clear stale folder config on OAuth revoke.
- Drop client-supplied `folderId` from the request DTO (server-derived only).
- Drive Picker doesn't cache a rejected script-load promise (a transient load
  failure can be retried).

## Testing
- Unit: `create-google-sheet-document.service`, `google-sheets-folder-validator`,
  `create-/update-data-destination.service`, `drive-folder-url.utils` (57 tests).
- e2e: auto-create endpoint wiring smoke test.
- Production typecheck (`tsconfig.build.json`) and lint pass.

## Follow-ups (not in this PR)
- Move the folder-validation Drive call outside the `@Transactional()` boundary.
- Decide fail-vs-fallback when an OAuth token lacks a Drive scope but a folder is
  configured (currently falls back to Drive root with a log only).
- Probe granted (not just requested) OAuth scopes before attempting to share.
- Enforce/validate that the Picker folder is chosen with the connected Google
  account (the `hintEmail` is advisory).
- Consolidate the duplicated `extractDriveFolderId` (backend vs web).


zapolsky:

Reviewed at head `8a9bac4` via a worktree checkout with a parallel multi-agent pass (security/leak reviewer, spec-vs-impl logic hunter, architecture, tester, and an un-checklisted fresh-eyes generalist). Every would-be blocker was traced boundary-to-boundary and **none survived as a blocker** — the credential/folder flow is solid: `folderId` is strictly server-derived (not client-injectable), no token/secret/SA-key reaches a log line or the HTTP body (`GoogleOAuthException` keeps `internalDetails` out of the response), best-effort sharing cannot fail an already-successful creation, the new endpoint is properly authed (`@Auth(Role.viewer(Strategy.INTROSPECT))` + `Action.USE` + project scoping), and the migration is additive/guarded/reversible.

CI-parity: `tsc -b` (prod `tsconfig.build.json`) ✓ · `eslint` on changed files ✓ · `jest` 57/57 unit + 1 e2e ✓.

The items below are **non-blocking** — judgement calls and sharp edges, highest-criticality first. None of them gate merge.

- `apps/backend/src/data-marts/use-cases/create-data-destination.service.ts:55,220` · `update-data-destination.service.ts:54,322` — **Live Drive API call held open inside the `@Transactional()` boundary.** Both save paths run `folderValidator.validateConfiguredFolder()` (→ `getFolderAccess` → `drive.files.get`, wrapped in `executeWithRetry` with up to ~31s of backoff and **no client `timeout`**) while a write transaction is open. On the single-connection `better-sqlite3` deployments this widens the window for the project's known concurrent-transaction 500s; on MySQL it pins a pooled connection + row locks for the network round-trip. Blast radius is limited to Service-Account destinations whose folder is created/changed (validator early-returns otherwise), and it fails gracefully (no data loss). You already list "move the folder-validation Drive call outside `@Transactional()`" as the #1 follow-up — flagging the SQLite severity specifically so it's weighted. Fix: validate before opening the tx (or after commit), or at minimum set a short `timeout` on the validation Drive client.

- `apps/backend/.../create-google-sheet-document.service.ts:100-118,220-225` · web `GoogleSheetsFields.tsx:158,255` · `GoogleSheetsReportEditForm.tsx:267` — **Pre-`drive.file` OAuth tokens: the folder picker is offered, but the folder is silently dropped and sharing is skipped at create — while the UI toasts success.** The picker gate (`canPickFolder = !!pickerApiKey && !!oauthClientId`) never consults the *stored* token's granted scope. A destination connected before this PR can pick+save a folder, but at create time `canShare=false` → `folderId=undefined` → create lands in Drive root, `shareWithRequester` no-ops, yet the report form shows `toast.success('Google Sheet created')` (the response DTO carries no `placedInRoot`/`shared` flag). Recoverable by reconnecting, but the behavior is silently wrong for a different requesting user (who gets a link they can't open). Fix: surface the stored token's `drive.file` capability to the FE (gate/relabel the picker) and/or return a `shared`/`placedInRoot` flag so the form warns instead of reporting plain success.

- `apps/backend/.../update-data-destination.service.ts:204-210` · `services/google-oauth/google-oauth-flow.service.ts:347-354` — **`config` clearing is inconsistent across the two OAuth-disconnect paths.** The standalone revoke (`DELETE :id/oauth`) clears `config: null` (and does so unconditionally — it would also wipe a still-valid Shared-Drive folder if ever invoked on a Service-Account destination, since there's no credential-type guard), but the edit-form disconnect (`PUT` with `credentialId: null`) sets `credentialId=null` and never touches `entity.config`, contradicting the spec's "on OAuth revoke, config is cleared." Net: a stale OAuth-picked folder can survive a disconnect+reconnect with a different account (then fails gracefully at create). Fix: clear the folder config on the update disconnect path too, and scope the revoke's `config: null` to `GOOGLE_OAUTH` credentials.

- `apps/backend/.../update-data-destination.service.ts:99-104,322` — **Credential-type switch with an unchanged `folderId` bypasses save-time SA folder validation.** `folderChanged` keys only on `folderId`, so switching an OAuth destination (whose folder was never validated — possibly a My Drive folder) to Service Account, or copying SA credentials, while keeping the same folder, skips `validateConfiguredFolder`. An unusable SA folder is persisted and the error is deferred to create time (graceful `SheetFolderCreateFailedException`, but no fail-fast on save). Fix: treat a credential-type transition as `folderChanged` (i.e. validate whenever the post-update credential is SA and a folder is set).

<details>

- `apps/backend/.../google-sheets-api.adapter.ts:187-215` — Non-atomic Drive create: `drive.files.create` succeeds, then a separate `spreadsheets.get` resolves the `sheetId`; if that second call fails post-retry the spreadsheet is already created (orphan) but the caller gets an error and no link. Affects both the OAuth Drive-create and SA folder-create paths. Consider a best-effort `files.delete` on failure, or tolerate a missing `sheetId`.
- `apps/backend/.../create-google-sheet-document.service.ts:105-118` — Asymmetric error handling: the folder branch wraps failures in `SheetFolderCreateFailedException` (400 + hint), but the root-create branch (`createSpreadsheetViaDrive`/`createSpreadsheet`) is unwrapped, so a transient Drive/Sheets error there surfaces as a raw 500. Wrap it for a consistent, friendlier error.
- `apps/backend/.../google-sheets/utils/drive-folder-url.utils.ts:17-35` · `mappers/data-destination.mapper.ts:92-101` — A non-empty `folderUrl` that yields no `folderId` (backend regex returns null) is still persisted as a half-configured destination (validator no-ops on null `folderId`), failing only at create time. The web form validates with the stricter `isValidGoogleDriveFolderUrl`, so this only bites direct API clients — consider rejecting on save when `folderUrl` is present but no `folderId` could be derived.
- `apps/backend/.../controllers/data-destination.controller.ts:168-172` · `services/google-oauth/google-oauth-config.service.ts:189-191` — The settings endpoint returns `pickerApiKey` (`GOOGLE_PICKER_API_KEY`) to any authed viewer. A Picker `developerKey` is public by design, but it's only safe if HTTP-referrer/API-restricted in GCP — worth a one-line note in the deployment docs.
- `apps/backend/.../controllers/data-destination.controller.ts:325-343` — A `viewer` with `Action.USE` can trigger Drive file creation on every click, with no idempotency or rate limit, into the owner's/SA's Drive. The auth bar itself is correct for the feature; just confirm `Action.USE` is the intended bar and consider a basic per-user/destination throttle.
- minor: `executeWithRetry` (`google-sheets-api.adapter.ts:859-884`) retries only the literal `"Quota exceeded"` and reads `error.message` unguarded (a non-`Error` throw → `TypeError`) — **pre-existing on `main`**, but now exercised by the new Drive calls; consider broadening to 429/5xx and guarding the message read. · `@HttpCode(200)` on the resource-creating POST (`controller.ts:327`) — 201 is conventional (cosmetic). · The new `create-google-sheet-document.service.spec.ts:15` introduces a `tsc` error (TS2556 on the zero-arg `jest.fn(() => ({}))` mock spread) — harmless under ts-jest `diagnostics:false` + `tsconfig.build.json` excluding specs (consistent with the existing baseline), but trivially fixed by typing the impl `(...args: unknown[]) => ({})`.

</details>

**Not attributable to this PR:** `tsc -b` over the full project (incl. specs) reports ~42 pre-existing type errors in untouched test files (`oauth-project-selection.service.spec.ts`, `bigquery-query.builder-edge-cases.spec.ts`, `run-report.service.spec.ts`, `permissions-*.e2e-spec.ts`, etc.). These exist on `main` and are tolerated by CI (backend `nest build` uses `tsconfig.build.json` which excludes `**/*spec.ts`, and `jest` runs ts-jest with `diagnostics:false`) — not introduced here.

<sub>Automated OWOX review · worktree multi-agent passes · all candidate blockers adversarially traced (0 confirmed).</sub>

apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
```diff
@@ -89,6 +91,18 @@ export class UpdateDataDestinationService {
...
+    // transaction) when the configured folder actually changes. Updates that
+    // leave the folder untouched — e.g. toggling availability or renaming the
+    // destination — skip the network call. Computed before `entity.config` is
+    // mutated below so it reflects the previously persisted folder.
+    const existingFolderId = entity.config?.folderId?.trim() || undefined;
+    const incomingFolderId =
+      command.config !== undefined
+        ? command.config?.folderId?.trim() || undefined
+        : existingFolderId;
+    const folderChanged = incomingFolderId !== existingFolderId;
```
romandubovyi, line 107: 🟡 **Important:** Folder validation also depends on the effective credential, not only on the folder ID. Replacing/copying credentials or switching from OAuth to a service account while keeping the same folder skips validation, so the update can succeed even though the new service account cannot write there. Please revalidate when the credential changes as well.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
```diff
@@ -288,6 +374,71 @@ export function GoogleSheetsFields({ form }: GoogleSheetsFieldsProps) {
...
+                              ? 'Open folder in new tab'
+                              : 'Paste a valid Drive folder URL to enable link'}
+                          </TooltipContent>
+                        </Tooltip>
+                      </div>
+                    </FormControl>
+                    <FormDescription>
+                      Paste a Google Drive folder URL. New documents created with “Create document”
+                      are placed here. Use a Shared Drive folder and add the service account email
+                      above as a member with the Content Manager role. Leave empty to create files
```
romandubovyi: 🟡 **Important:** This fallback is not implemented: the backend throws `ServiceAccountRequiresFolderException` when no folder is configured. Please state that a Shared Drive folder is required for service-account document creation, or implement the described fallback.

apps/web/src/features/data-destination/shared/hooks/useGoogleDrivePicker.ts
```diff
@@ -0,0 +1,200 @@
...
+        .addView(view)
+        .enableFeature(picker.Feature.SUPPORT_DRIVES)
+        .setCallback((response: PickerResponse) => {
+          const doc = response.action === picker.Action.PICKED ? response.docs?.[0] : undefined;
+          if (doc) {
+            onPicked({ id: doc.id, name: doc.name, url: doc.url });
+          }
+        })
+        .build();
+      built.setVisible(true);
```
romandubovyi: 💡 **Suggestion:** `openPicker()` resolves immediately after showing the picker, so the caller clears `isPickingFolder` while the dialog is still open and allows duplicate picker/token flows. Consider resolving only when the callback receives `PICKED` or `CANCEL`.

apps/web/src/features/data-destination/edit/components/DataDestinationEditSheet/DataDestinationConfigSheet.tsx
```diff
@@ -104,6 +104,19 @@ export function DataDestinationConfigSheet({
...
+  // Sheet's DOM), so interacting with it would be treated as an outside
+  // interaction and close the Sheet. Keep the Sheet open while a Picker is
+  // present (matched by the event target OR by any Picker node in the DOM).
+  const PICKER_SELECTOR =
+    '.picker-dialog, .picker-dialog-bg, .picker, .pac-container, iframe[src*="docs.google.com/picker"]';
+  const shouldKeepOpenForPicker = (target: EventTarget | null): boolean => {
+    if (target instanceof Element && target.closest(PICKER_SELECTOR)) {
+      return true;
+    }
+    return !!document.querySelector(PICKER_SELECTOR);
```
romandubovyi: 🟡 **Important:** The DOM-wide `document.querySelector(PICKER_SELECTOR)` fallback fires for *any* outside interaction. Google Picker renders in `<body>` and does not always remove every node on close (a hidden `.picker-dialog-bg` / iframe can linger). If any matched node remains, this guard permanently prevents the Sheet from closing on outside-click/focus-out — the user can no longer dismiss it that way. Consider scoping the fallback to a *visible* picker node, or explicitly removing residual picker nodes after the picker closes.

apps/backend/src/data-marts/controllers/data-destination.controller.ts
```diff
@@ -315,4 +321,24 @@ export class DataDestinationController {
...
+  @Auth(Role.viewer(Strategy.INTROSPECT))
+  @Post(':id/google-sheets/documents')
+  @HttpCode(200)
+  @CreateGoogleSheetDocumentSpec()
+  async createGoogleSheetDocument(
+    @AuthContext() context: AuthorizationContext,
+    @Param('id') id: string,
+    @Body() dto: CreateGoogleSheetDocumentRequestDto
+  ): Promise<CreateGoogleSheetDocumentResponseDto> {
+    await this.checkDestinationAccess(id, context, Action.USE);
```
romandubovyi, line 334: 🔵 **Question:** Is `Action.USE` intentionally sufficient for an unbounded external side effect? Users with a destination shared only “for use” can call this endpoint repeatedly and create persistent files in the owner’s Drive or Shared Drive. Should this operation have a dedicated permission, throttling, or another abuse safeguard?

apps/backend/src/data-marts/mappers/data-destination.mapper.ts
```diff
@@ -74,10 +78,28 @@ export class DataDestinationMapper {
...
+   */
+  private normalizeDestinationConfig(config?: DestinationConfigDto): DestinationConfig | undefined {
+    if (config === undefined) {
+      return undefined;
+    }
+    const folderUrl = config.folderUrl?.trim() || null;
+    // folderId is server-derived only — never taken from the client payload —
+    // so it always reflects a validated folder URL (or null when cleared).
+    const folderId = folderUrl ? extractDriveFolderId(folderUrl) : null;
+    return { folderUrl, folderId };
```
romandubovyi, line 108: 💡 **Suggestion:** This Google Sheets-specific config is normalized and persisted for every destination type. Consider rejecting or ignoring `folderUrl` for non-Google-Sheets destinations, or model `config` as a type-discriminated structure before other providers add their own settings.

apps/web/src/features/data-destination/shared/utils/drive-folder-url.utils.ts
```diff
@@ -0,0 +1,25 @@
...
+const driveFolderUrlRegex =
+  /^https:\/\/drive\.google\.com\/drive\/(?:u\/\d+\/)?folders\/([A-Za-z0-9_-]+)/;
+
+/** True when the string is a valid Google Drive folder URL. */
+export function isValidGoogleDriveFolderUrl(url: string): boolean {
+  return driveFolderUrlRegex.test(url.trim());
+}
+
+/** Extracts the folder ID from a Drive folder URL, or '' when not found. */
+export function extractDriveFolderId(url: string): string {
```
romandubovyi: ⚪ **Nit:** `extractDriveFolderId` is exported but unused anywhere in the web app — the frontend only sends `folderUrl` and the id is derived server-side. It's dead code (and a near-duplicate of the backend util with different behavior). Consider dropping it, or leaving a note that it's kept for parity.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
```diff
@@ -288,6 +374,71 @@ export function GoogleSheetsFields({ form }: GoogleSheetsFieldsProps) {
...
+                      in the service account’s Drive.
```
romandubovyi: 🔴 **Blocking:** This description says "Leave empty to create files in the service account’s Drive", but the backend `CreateGoogleSheetDocumentService.createViaServiceAccount` throws `ServiceAccountRequiresFolderException` when no folder is configured. Update the copy to require a Shared Drive folder for Service Account destinations, or change the backend behavior if leaving it empty is intentional.

apps/backend/src/data-marts/use-cases/google-sheets/create-google-sheet-document.service.ts
```diff
@@ -0,0 +1,244 @@
...
+    // on it would fail. Use Drive create only when the token actually has a Drive scope.
+    const canShare = this.oauthTokenHasDriveScope(destination);
+    // Folder placement is possible only with a Drive scope and a folder the user
+    // granted via the Picker (drive.file gives access to picker-selected folders).
+    const folderId = canShare ? destination.config?.folderId?.trim() : undefined;
+    let result: CreateGoogleSheetDocumentResponseDto;
+    if (folderId) {
+      try {
+        result = await adapter.createSpreadsheetInFolder(title, folderId);
+      } catch (error) {
```
romandubovyi, line 115: 🟡 **Important:** Wrapping every `createSpreadsheetInFolder` error in `SheetFolderCreateFailedException` may mask transient Google API errors (e.g., 503, quota) as a folder-access problem. Consider distinguishing 403/404 from transient failures, or log the original error details before throwing a user-facing exception.

apps/backend/src/data-marts/use-cases/google-sheets/create-google-sheet-document.service.ts
```diff
@@ -0,0 +1,244 @@
...
+        `Best-effort share of ${spreadsheetId} with ${email} failed: ${
+          error instanceof Error ? error.message : String(error)
+        }`
+      );
+    }
+  }
+
+  /** True when the destination's stored OAuth token already includes a Drive scope. */
+  private oauthTokenHasDriveScope(destination: DataDestination): boolean {
+    const credentials = destination.credential?.credentials as { scope?: string } | undefined;
```
romandubovyi: 🟡 **Important:** Use the existing `GoogleOAuthTokens` type from `google-oauth-flow.service.ts` instead of an inline `as { scope?: string }` cast. It already declares `scope: string` and is the canonical shape for OAuth tokens, making this code easier to refactor and safer.

apps/backend/src/data-marts/use-cases/google-sheets/create-google-sheet-document.service.ts
```diff
@@ -0,0 +1,244 @@
...
+        return this.createViaServiceAccount(destination, title, requesterEmail);
+      default:
+        throw new OAuthNotConnectedException(destination.id);
+    }
+  }
+
+  /**
+   * OAuth path: creates the file in the connected user's Drive root, then shares
+   * it with the requester so they can open it. Folder placement is NOT supported
+   * for OAuth (would require a restricted Drive scope or the Drive Picker).
```
romandubovyi: 💡 **Suggestion:** The JSDoc says OAuth folder placement is not supported, but the implementation below creates sheets in `destination.config.folderId` when the OAuth token has a Drive scope. Update the comment to reflect that folder placement is supported via the Drive Picker and `drive.file`.

apps/backend/src/data-marts/use-cases/google-sheets/create-google-sheet-document.service.ts
```diff
@@ -0,0 +1,244 @@
...
+    const adapter = new GoogleSheetsApiAdapter(undefined, jwt);
+
+    let result: CreateGoogleSheetDocumentResponseDto;
+    try {
+      result = await adapter.createSpreadsheetInFolder(title, folderId);
+    } catch (error) {
+      throw new SheetFolderCreateFailedException(
+        destination.id,
+        error,
+        'Make sure it is a Shared Drive folder shared with the service account as Editor.'
```
romandubovyi: ⚪ **Nit:** The error message says "shared with the service account as Editor", but the validator uses "Content Manager" role. Use the same role name in both messages to avoid confusing users.

apps/backend/src/data-marts/controllers/spec/data-destination.api.ts
```diff
@@ -165,6 +167,25 @@ export function OAuthRevokeSpec() {
   );
 }
 
+export function CreateGoogleSheetDocumentSpec() {
+  return applyDecorators(
+    ApiOperation({
+      summary: 'Auto-create a new Google Sheet for a Data Destination (OAuth)',
```
romandubovyi: ⚪ **Nit:** The Swagger summary says "(OAuth)" but the endpoint also supports Service Account destinations. Remove the qualifier or rename to "(Google Sheets)" to match the implementation.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
laskevych: Fixed in `377772b4a` — kept this FE-only. The Service Account folder field is now labeled **(required)** and the description states a Shared Drive folder is required (add the service account as a Content Manager). The backend behavior is intentional and stays: `ServiceAccountRequiresFolderException` on an empty folder is deliberate (an SA file in My Drive would be invisible) and is covered by a unit test, so I updated the copy rather than relaxing the backend.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/GoogleSheetsFields.tsx
laskevych: Addressed together with the blocking copy comment — fixed in `377772b4a`: the Service Account description no longer says "leave empty to create files in the service account's Drive"; it now states a Shared Drive folder is required.

apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
laskevych: Good catch — fixed in `377772b4a`. Validation no longer keys only on `folderId`: a `credentialChanged` flag is set inside the branches that actually mutate credential data (copy / OAuth switch / SA re-upload), and the folder validator now runs on `folderChanged || credentialChanged`. The validator already no-ops for non-Service-Account credentials, so OAuth changes don't add a Drive round-trip.

Note I deliberately did **not** key on `credentialId` equality — in-place updates (SA key re-upload, copy-into-existing) reuse the same id, so that would miss exactly the rotation cases you describe. Added specs for the copy-in-place case.

apps/web/src/features/data-destination/shared/hooks/useGoogleDrivePicker.ts
laskevych: Fixed in `377772b4a` — `openPicker()` now wraps the dialog in a promise that resolves only on `PICKED` or `CANCEL` (intermediate actions ignored), so the caller's `isPickingFolder` guard stays set until the dialog actually closes and a duplicate picker/token flow can't be triggered.

apps/web/src/features/data-destination/edit/components/DataDestinationEditSheet/DataDestinationConfigSheet.tsx
laskevych: Fixed in `377772b4a` — the `document.querySelectorAll(PICKER_SELECTOR)` fallback now counts only a **visible** picker node (`checkVisibility` + a non-zero bounding box; iframe matches treated as visible), so a residual hidden `.picker-dialog-bg` can no longer permanently block the Sheet from closing on outside-click / focus-out. Disposing the picker nodes in the hook to remove the root cause is a sensible follow-up.

apps/backend/src/data-marts/controllers/data-destination.controller.ts
laskevych: Intentional for now. `Action.USE` is the same bar as running a report against the destination, which also performs external writes — so a viewer who can use the destination can already trigger Google-side effects. We didn't add a dedicated permission. A basic per-user/destination throttle or idempotency key is a reasonable abuse safeguard, but there's no throttling infrastructure today, so I'm tracking it as a follow-up rather than adding it in this PR.

apps/backend/src/data-marts/mappers/data-destination.mapper.ts
laskevych: Partially addressed in `5a99062a1`: a non-empty `folderUrl` that can't be parsed into a `folderId` is now rejected at save time (400) instead of persisting a half-configured destination that only fails at create time. The broader change — modeling `config` as a type-discriminated structure / ignoring `folderUrl` for non-Google-Sheets destinations — I left as a follow-up: today only the Google Sheets services read `config.folderId` (so it's latent), and the update path doesn't carry the destination type at the mapper layer, so doing it properly is a larger change.

apps/web/src/features/data-destination/shared/utils/drive-folder-url.utils.ts
laskevych: Removed in `377772b4a` — the web `extractDriveFolderId` was unused (the FE only sends `folderUrl`; the id is derived server-side).

apps/backend/src/data-marts/use-cases/google-sheets/create-google-sheet-document.service.ts
laskevych: Fixed in `377772b4a` — a new `throwFolderCreateError` classifies by HTTP status: 403/404 (or unknown) stay `SheetFolderCreateFailedException` (400, with the folder hint), while 429/5xx now surface as `GoogleApiException` (502) so a transient Google outage isn't reported as a folder-access problem. The original error is logged by the exception constructor either way.

One known limitation: a quota-exhaustion that survives all retries throws a plain `Error` with no status, so it still maps to the folder case.

apps/backend/src/data-marts/use-cases/google-sheets/create-google-sheet-document.service.ts
laskevych: Done in `377772b4a` — now uses `GoogleOAuthTokens` from `google-oauth-flow.service.ts` instead of the inline `as { scope?: string }` cast.

apps/backend/src/data-marts/use-cases/google-sheets/create-google-sheet-document.service.ts
laskevych: Done in `377772b4a` — updated the `createViaOAuth` JSDoc (and the class-level JSDoc) to state that folder placement IS supported when the token has a Drive scope and the folder was picked via the Drive Picker (`drive.file`).

apps/backend/src/data-marts/use-cases/google-sheets/create-google-sheet-document.service.ts
laskevych: Done in `377772b4a` — aligned to "Content Manager" in the create-service hint, and also in `ServiceAccountRequiresFolderException`, which had the same "Editor" wording.

apps/backend/src/data-marts/controllers/spec/data-destination.api.ts
laskevych: Done in `377772b4a` — the Swagger summary is now "Auto-create a new Google Sheet for a Data Destination (Google Sheets)".

laskevych:
Thanks for the thorough review 🙏 — addressed across three commits on this branch:

- **`377772b4a`** — review comments (Tier 1 + 2): the blocking Service Account folder copy, credential-change revalidation, Drive Picker promise timing, the DOM-wide picker guard, transient-vs-folder error classification, the `GoogleOAuthTokens` type, the stale JSDoc, the "Content Manager" wording, the dead `extractDriveFolderId`, and the Swagger summary.
- **`5a99062a1`** — backend hardening: reject an unparseable `folderUrl` at save time; add a 10s client timeout to the folder-validation Drive call (it runs inside the save transaction); best-effort delete of an orphaned spreadsheet if `sheetId` resolution fails post-create.
- **`29a85c17e`** — surface a downgraded auto-create: the response now carries optional `placedInRoot` / `sharedWithRequester` flags, and the report form warns (instead of plain success) when a pre-`drive.file` OAuth token drops the chosen folder or skips sharing. Older backends omit the flags, so the success toast is unchanged for them.

Each inline thread has a specific reply.

**Intentionally deferred to follow-ups** (all fail gracefully today):
- gate the Drive Picker on the stored token's granted scope (so the picker isn't offered for a token that can't honor it);
- move the folder-validation Drive call outside the `@Transactional()` boundary (the timeout above bounds it for now);
- a type-discriminated `config` model;
- a per-user/destination throttle for the create endpoint (`Action.USE`);
- the 200 → 201 status nit.

Checks green: backend prod typecheck (`tsconfig.build.json`), backend jest incl. new specs, web typecheck, eslint, web vitest.

romandubovyi (approved):
Re-reviewed at 9799f452. The previous review comments were either addressed in code and tests or explicitly evaluated and reasonably deferred as non-blocking follow-ups. No new blocking issues found. All CI checks are green.

PR #1370, opened 2026-06-29
Fix search indexing sort memory exhaustion

kalnik-a-a:
## Summary

- Narrow search-indexing pagination queries so MySQL does not sort wide JSON/text payloads while building the search index.
- Split data-storage and data-destination indexing into a narrow keyset page query followed by bounded credential hydration.
- Adjust data-mart pagination to `createdAt DESC, id ASC` so it can use the existing `idx_dm_project_deleted_created` composite index instead of adding another data-mart pagination index.
- Add regression coverage for the new pagination shapes and include an `owox` changeset.

## Why

The production failure came from TypeORM building ordered queries over wide rows and eager relation joins. On MySQL this can force a filesort over large JSON/text payloads and hit `ER_OUT_OF_SORTMEMORY` / `Out of sort memory`.

For storages and destinations, the sorted query now only carries `{ id, createdAt }`; the wider credential data is loaded after the page boundary is known. For data marts, the query still needs `schema` to build embeddings, so the change narrows the selected fields, disables eager relation loading, and intentionally orders by `createdAt DESC, id ASC` to match the existing data-mart composite index.

## Verification

- `npm test -w @owox/backend -- data-mart.source data-storage.source data-destination.source typeorm-data-mart-catalog.adapter --runInBand`
- `npm test -w @owox/backend -- advanced-search --runInBand`
- `npx tsc -p apps/backend/tsconfig.build.json --noEmit`
- `git diff --check`

### DIFF
--- a/.changeset/add-search-functionality.md
+++ b/.changeset/add-search-functionality.md
@@ -1,0 +1,7 @@
+---
+"owox": minor
+---
+
+# Add search functionality
+
+Add search functionality for finding project data marts, storages, and destinations from the project search index.
--- a/apps/backend/src/data-marts/search/catalog/typeorm-data-mart-catalog.adapter.ts
+++ b/apps/backend/src/data-marts/search/catalog/typeorm-data-mart-catalog.adapter.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
-import { FindOptionsWhere, IsNull, MoreThan, Raw, Repository } from 'typeorm';
+import { In, IsNull, Repository } from 'typeorm';
 import { DataMart } from '../../entities/data-mart.entity';
 import { DataMartRelationship } from '../../entities/data-mart-relationship.entity';
 import { DataMartStatus } from '../../enums/data-mart-status.enum';
@@ -12,7 +12,7 @@ import type {
   SearchableDataMart,
 } from './data-mart-catalog.port';
 import type { PageCursor, SearchablePage } from '../sources/indexable-source.port';
-import { nextPageCursor } from '../sources/indexable-source.port';
+import { buildKeysetWhere, nextPageCursor } from '../sources/indexable-source.port';
 import { buildDataMartScoringDescriptor } from './data-mart-scoring-descriptor.builder';
 
 type SchemaFieldLike = {
@@ -29,40 +29,22 @@ type SearchableDataMartProjection = Pick<
   'id' | 'projectId' | 'title' | 'description' | 'schema' | 'status' | 'modifiedAt'
 >;
 
-type SearchableDataMartPageRow = SearchableDataMartProjection & Pick<DataMart, 'createdAt'>;
+type SearchableDataMartPageRow = Pick<DataMart, 'id' | 'createdAt'>;
 
-const DATA_MART_SEARCHABLE_PAGE_SELECT = {
+const DATA_MART_SEARCHABLE_SELECT = {
   id: true,
   projectId: true,
   title: true,
   description: true,
   schema: true,
   status: true,
-  createdAt: true,
   modifiedAt: true,
+} satisfies Record<keyof SearchableDataMartProjection, true>;
+
+const DATA_MART_SEARCHABLE_PAGE_SELECT = {
+  id: true,
+  createdAt: true,
 } satisfies Record<keyof SearchableDataMartPageRow, true>;
-
-function buildDataMartSearchPageWhere(
-  projectId: string,
-  cursor: PageCursor | null
-): FindOptionsWhere<DataMart> | FindOptionsWhere<DataMart>[] {
-  const baseWhere: FindOptionsWhere<DataMart> = { projectId, deletedAt: IsNull() };
-  if (!cursor) return baseWhere;
-
-  return [
-    {
-      ...baseWhere,
-      createdAt: Raw(alias => `${alias} < :createdBefore`, {
-        createdBefore: cursor.createdAt,
-      }),
-    },
-    {
-      ...baseWhere,
-      createdAt: Raw(alias => `${alias} = :createdAt`, { createdAt: cursor.createdAt }),
-      id: MoreThan(cursor.id),
-    },
-  ];
-}
 
 function collectFieldLabels(fields: SchemaFieldLike[]): string[] {
   const labels: string[] = [];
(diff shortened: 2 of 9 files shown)


apps/backend/src/data-marts/search/catalog/typeorm-data-mart-catalog.adapter.ts
```diff
@@ -77,11 +117,13 @@ export class TypeOrmDataMartCatalogAdapter implements DataMartCatalogPort {
...
-    const where = buildKeysetWhere<DataMart>({ projectId, deletedAt: IsNull() }, cursor);
+    const where = buildDataMartSearchPageWhere(projectId, cursor);
 
-    const marts = await this.dataMartRepo.find({
+    const marts: SearchableDataMartPageRow[] = await this.dataMartRepo.find({
       where,
-      order: { createdAt: 'ASC', id: 'ASC' },
+      select: DATA_MART_SEARCHABLE_PAGE_SELECT,
+      loadEagerRelations: false,
+      order: { createdAt: 'DESC', id: 'ASC' },
```
romandubovyi, line 109: 🟡 **Important:** This `createdAt DESC, id ASC` order is intentionally coupled to the `idx_dm_project_deleted_created (projectId, deletedAt, createdAt DESC, id ASC)` index — it's the only thing keeping MySQL from filesorting rows that still carry the wide `schema` JSON. Unlike the storage/destination sources (which sort `ASC`), it must not be "normalized" back to ASC. Please add a comment here referencing the index migration so a future change doesn't silently reintroduce `ER_OUT_OF_SORTMEMORY`.

apps/backend/src/data-marts/search/catalog/typeorm-data-mart-catalog.adapter.ts
```diff
@@ -24,6 +24,46 @@ type SchemaFieldLike = {
...
+  projectId: true,
+  title: true,
+  description: true,
+  schema: true,
+  status: true,
+  createdAt: true,
+  modifiedAt: true,
+} satisfies Record<keyof SearchableDataMartPageRow, true>;
+
+function buildDataMartSearchPageWhere(
```
romandubovyi: 💡 **Suggestion:** `buildDataMartSearchPageWhere` duplicates the shared `buildKeysetWhere` from `indexable-source.port.ts`, differing only in the comparison direction (`<` vs `>`). Consider extending the shared helper to accept a sort direction so the keyset logic lives in one place — two near-identical implementations are easy to fix in one and miss in the other.

apps/backend/src/data-marts/search/sources/data-storage.source.ts
```diff
@@ -102,15 +102,40 @@ export class DataStorageIndexableSource implements IndexableSource {
...
+      loadEagerRelations: false,
       order: { createdAt: 'ASC', id: 'ASC' },
       take: limit,
     });
-    const descriptors = storages.map(toDescriptor);
+    if (pageRows.length === 0) return { descriptors: [], nextCursor: null };
+
+    const pageIds = pageRows.map(storage => storage.id);
+    const storages = await this.dataStorageRepo.find({
+      where: { id: In(pageIds), projectId, deletedAt: IsNull() },
```
romandubovyi, line 117: 💡 **Suggestion:** The second query reapplies `projectId` and `deletedAt` filters that are already guaranteed by the keyset page query. Since `pageIds` come from the first query, these conditions are redundant. Consider removing them to make the hydration intent clearer, or add a comment if they're kept deliberately (e.g., for race-condition safety).

apps/backend/src/data-marts/search/catalog/typeorm-data-mart-catalog.adapter.ts
```diff
@@ -77,11 +117,13 @@ export class TypeOrmDataMartCatalogAdapter implements DataMartCatalogPort {
     cursor: PageCursor | null,
     limit: number
   ): Promise<SearchablePage> {
-    const where = buildKeysetWhere<DataMart>({ projectId, deletedAt: IsNull() }, cursor);
+    const where = buildDataMartSearchPageWhere(projectId, cursor);
 
-    const marts = await this.dataMartRepo.find({
+    const marts: SearchableDataMartPageRow[] = await this.dataMartRepo.find({
```
romandubovyi: 💡 **Suggestion:** Since `schema` is a wide JSON column, we could apply the same split-hydration pattern here as in `DataStorage` and `DataDestination`. By first querying only `{ id, createdAt }`, the database can satisfy the query entirely from the index (covering index query). Then, we can hydrate the full fields (including `schema`) for the paginated subset of IDs.

.changeset/add-search-functionality.md
```diff
@@ -0,0 +1,7 @@
+---
+"owox": minor
+---
+
+# Add search functionality
```
romandubovyi, line 5: ⚪ **Nit:** The changeset description says "Add search functionality", but this PR is a performance fix for "Fix search indexing sort memory exhaustion". It might be cleaner to adjust the text to match the PR's actual purpose.

romandubovyi:
Reviewed — no blockers. 👍

The fix is correct and well covered: I verified the DESC keyset logic and confirmed the `idx_dm_project_deleted_created` migration indexes `(projectId, deletedAt, createdAt DESC, id ASC)`, so the `id` tie-break is covered and the data-mart filesort/`ER_OUT_OF_SORTMEMORY` is genuinely avoided. The two-phase narrow-then-hydrate approach for storages/destinations is sound and keeps page order via the `pageIds` map.

One thing worth attention before merge (non-blocking): the data-mart `order: { createdAt: DESC, id: ASC }` is silently coupled to that composite index — please add a code comment so it isn't later "normalized" back to ASC and silently reintroduces the OOM. See the inline comment for details.

romandubovyi (approved):

apps/backend/src/data-marts/search/catalog/typeorm-data-mart-catalog.adapter.ts
kalnik-a-a: Updated in aa7ddc9cd. `buildKeysetWhere` now accepts the createdAt sort direction, so data marts use the shared helper with `DESC` while storage and destination sources keep the default `ASC` path.

apps/backend/src/data-marts/search/catalog/typeorm-data-mart-catalog.adapter.ts
kalnik-a-a: Updated in aa7ddc9cd. Data marts now use the split-hydration path, and the narrow page query keeps `createdAt DESC, id ASC` aligned with `idx_dm_project_deleted_created` for the covering page query. Added a short inline note near the order.

apps/backend/src/data-marts/search/catalog/typeorm-data-mart-catalog.adapter.ts
kalnik-a-a: Implemented in aa7ddc9cd. Data-mart indexing now first pages only `{ id, createdAt }`, then hydrates the selected searchable fields including `schema`, restores descriptor order from `pageIds`, and derives the cursor from the narrow page rows.

apps/backend/src/data-marts/search/sources/data-storage.source.ts
kalnik-a-a: Kept these filters intentionally. They re-check tenant and soft-delete scope in the race window between page selection and hydration; added short comments to make that explicit.

.changeset/add-search-functionality.md
kalnik-a-a: This changeset intentionally describes the search capability because the original search PR did not include a release note. This PR is where the missing changeset is added, so the user-facing text stays about search being introduced rather than this internal indexing hardening.

PR #1371, opened 2026-06-29
test(backend): stop Redshift integration timeouts from flaking on transient latency

zapolsky:
## Problem

The `Redshift` job in **Integration Tests (Real DB)** fails on roughly half of the
nightly scheduled runs. The failure signature is consistent and points to
environment latency, not a code bug:

- Always **1 test failed, 40 passed** — never a wrong result, always
  `thrown: "Exceeded timeout of 30000 ms for a test."` at ~30001 ms.
- The failing test **roams** between runs (`gt: amount > 20`, `relative_date today`,
  `TIMESTAMP: gte`, `DATE: between`, `TIMETZ: gte`, …) — whichever test happens to
  run during a latency spike gets killed at the 30s ceiling.
- Data API polling logs show queries intermittently taking ~30s to execute a
  trivial statement; latency climbs mid-run (e.g. `neq` took 17s, the next test
  timed out) and drops back to ~1.5s immediately after.

### Root cause

The Serverless workgroup runs on **minimal Base RPU** (cost cap) and is queried
**concurrently** by the parallel `http-data-real` suite (which also hits Redshift)
on a separate matrix runner. Under that contention, individual Data API queries
occasionally spike past the **hardcoded 30s per-test timeout** that every test in
`redshift.integration.ts` set explicitly — even though the suite's global timeout
in `jest-integration.json` is already 60s.

## Change

- Remove the hardcoded `30000` per-test timeout from all 41 tests so they inherit
  the **global 60s** from `jest-integration.json`. (The intentional longer
  timeouts — 60s/120s for seeding/heavy ops — are left untouched.)
- Add `jest.retryTimes(1, { logErrorsBeforeRetry: true })` so a single transient
  spike no longer fails the whole run, while still logging the original error.

No production code changes — test/CI reliability only.

## Notes

This addresses the symptom cheaply (no RPU spend). It does **not** remove the
underlying contention between the `Redshift` and `http-data-real` jobs; if 60s
still proves too tight on occasion, the follow-up is to serialize the two
Redshift-touching jobs via a shared concurrency group.

### DIFF
--- a/apps/backend/test/integration/redshift.integration.ts
+++ b/apps/backend/test/integration/redshift.integration.ts
@@ -59,6 +59,13 @@ if (!REDSHIFT_CREDENTIALS_AVAILABLE) {
 
 const describeIfCredentials = REDSHIFT_CREDENTIALS_AVAILABLE ? describe : describe.skip;
 
+// The Serverless workgroup runs on minimal Base RPU (cost cap) and is hit
+// concurrently by the parallel http-data-real suite, so individual Data API
+// queries occasionally spike to ~30s+ under contention. Per-test timeouts now
+// inherit the global 60s (jest-integration.json) instead of a hardcoded 30s, and
+// we retry once so a single transient latency spike doesn't fail the whole run.
+jest.retryTimes(1, { logErrorsBeforeRetry: true });
+
 // ---------------------------------------------------------------------------
 // Access validation + dry-run
 // ---------------------------------------------------------------------------
@@ -87,7 +94,7 @@ describeIfCredentials('Redshift Integration Tests', () => {
   describe('Access Validation', () => {
     it('should accept valid credentials', async () => {
       await expect(adapter.checkAccess()).resolves.not.toThrow();
-    }, 30000);
+    });
 
     it('should reject invalid credentials', async () => {
       const invalidAdapter = new RedshiftApiAdapter(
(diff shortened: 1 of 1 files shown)


laskevych (approved):

apps/backend/test/integration/redshift.integration.ts
```diff
@@ -55,6 +55,13 @@ if (!REDSHIFT_CREDENTIALS_AVAILABLE) {
 
 const describeIfCredentials = REDSHIFT_CREDENTIALS_AVAILABLE ? describe : describe.skip;
 
+// The Serverless workgroup runs on minimal Base RPU (cost cap) and is hit
+// concurrently by the parallel http-data-real suite, so individual Data API
+// queries occasionally spike to ~30s+ under contention. Per-test timeouts now
+// inherit the global 60s (jest-integration.json) instead of a hardcoded 30s, and
+// we retry once so a single transient latency spike doesn't fail the whole run.
+jest.retryTimes(1, { logErrorsBeforeRetry: true });
```
romandubovyi, line 67: 🟡 **Important:** `jest.retryTimes(1)` retries on *any* failure, not just the latency-timeout spikes it targets. Since this is the only live suite asserting query-builder correctness, a genuinely intermittent assertion regression (e.g. nondeterministic row ordering) could be silently flipped green by the retry and hidden from CI. The trade-off is probably acceptable here, but it's worth stating explicitly in the comment above — or scoping the retry to timeout errors if feasible — so a future reader knows non-timeout flakiness is also being masked.

romandubovyi (approved):

PR #1372, opened 2026-06-29
Feat(web): Add search bar to report column picker

artemfenoga:
Frame 23137615

# Improved the Report Column Picker with a faster and more intuitive column selection experience.

### Added column search
Frame 23137616
 
###  Added an empty state for search results &  Improved the toolbar layout
Frame 23137617

### Search works with the **Show selected only** option
Frame 23137618

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
+++ b/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
@@ -3,6 +3,8 @@ import { useQuery } from '@tanstack/react-query';
 import { cn } from '@owox/ui/lib/utils';
 import { Badge } from '@owox/ui/components/badge';
 import { Button } from '@owox/ui/components/button';
+import { Input } from '@owox/ui/components/input';
+import { Search } from 'lucide-react';
 import { Checkbox } from '@owox/ui/components/checkbox';
 import { Collapsible, CollapsibleContent } from '@owox/ui/components/collapsible';
 import { Switch } from '@owox/ui/components/switch';
@@ -13,9 +15,10 @@ import { NoAccessIndicator } from '../DataMartRelationships/NoAccessIndicator';
 import { dataMartRelationshipService } from '../../../shared/services/data-mart-relationship.service';
 import { BLENDABLE_SCHEMA_QUERY_KEY } from '../../../shared/hooks/blendable-schema-query-key';
 import type {
-  AggregationRole,
   AvailableSource,
   BlendedField,
+  BlendedGroup,
+  NativeField,
   ReportAggregateFunction,
 } from '../../../shared/types/relationship.types';
 import { DataStorageType } from '../../../../data-storage/shared/model/types/data-storage-type.enum';
@@ -47,20 +50,8 @@ import {
   functionsForColumn,
   timeZoneForColumn,
 } from './aggregation-config';
-
-interface NativeField {
-  name: string;
-  type?: string;
-  alias?: string;
-  description?: string;
-  isHiddenForReporting?: boolean;
-  status?: string;
-  fields?: NativeField[];
-  isPrimaryKey?: boolean;
-  // Aggregation governance (optional; absent → type-derived defaults on the web).
-  aggregationRole?: AggregationRole;
-  allowedAggregations?: ReportAggregateFunction[];
-}
+import { buildColumnSearchResult, matchesColumnSearch } from './report-column-search';
+import { SearchButton } from './SearchButton';
 
 // Must stay in sync with the backend collectSchemaFieldPaths walker: hidden and
 // DISCONNECTED nodes (with their subtrees) are unavailable for reporting, so they
@@ -84,16 +75,6 @@ function flattenNativeFields(fields: NativeField[], prefix = ''): NativeField[] 
     }
   }
   return result;
-}
-
-interface BlendedGroup {
-  aliasPath: string;
-  title: string;
-  alias: string;
-  description?: string;
-  isAccessibleForReporting: boolean;
-  visibleFields: BlendedField[];
-  selectedCount: number;
 }
 
 export interface ReportColumnSelectionCount {
@@ -372,6 +353,7 @@ interface BlendedGroupItemProps {
   preJoinByAliasPathColumn?: Map<string, ColumnFilters>;
   aggregationByColumn?: Map<string, ColumnAggregation>;
   onApplyAggregation?: ApplyAggregationFn;
+  hasSearchQuery?: boolean;
 }
 
 function BlendedGroupItem({
@@ -386,11 +368,21 @@ function BlendedGroupItem({
   preJoinByAliasPathColumn,
   aggregationByColumn,
   onApplyAggregation,
+  hasSearchQuery = false,
 }: BlendedGroupItemProps) {
   const [isOpen, setIsOpen] = useState(() => group.selectedCount > 0);
   const inaccessible = !group.isAccessibleForReporting;
   const Chevron = isOpen ? ChevronDown : ChevronRight;
   const accentClass = inaccessible ? 'text-destructive' : 'text-muted-foreground';
+  const previousHasSearchQuery = useRef(false);
+
+  useEffect(() => {
+    if (!previousHasSearchQuery.current && hasSearchQuery) {
+      setIsOpen(true);
+    }
+
+    previousHasSearchQuery.current = hasSearchQuery;
+  }, [hasSearchQuery]);
 
   return (
     <Collapsible
@@ -463,8 +455,16 @@ export function ReportColumnPicker({
   const outputControlsSupported = storageType ? supportsOutputControls(storageType) : false;
   const outputControlsAvailable: boolean = outputControlsSupported && !!onOutputConfigChange;
   const effectiveOutputConfig: OutputConfig = outputConfig ?? EMPTY_OUTPUT_CONFIG;
-  const [settingsOpen, setSettingsOpen] = useState(false);
-  const [aggSettingsOpen, setAggSettingsOpen] = useState(false);
+
+  type ActivePanel = 'aggregation' | 'output' | 'search' | null;
+  const [activePanel, setActivePanel] = useState<ActivePanel>(null);
+  const settingsOpen = activePanel === 'output';
+  const aggSettingsOpen = activePanel === 'aggregation';
+  const isSearchOpen = activePanel === 'search';
+  const togglePanel = (panel: Exclude<ActivePanel, null>) => {
+    setActivePanel(current => (current === panel ? null : panel));
+  };
+
   const { data: schema, isLoading } = useQuery({
     queryKey: [BLENDABLE_SCHEMA_QUERY_KEY, dataMartId],
     queryFn: () => dataMartRelationshipService.getBlendableSchema(dataMartId),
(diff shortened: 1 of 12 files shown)


apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -966,7 +1016,7 @@ export function ReportColumnPicker({
           />
         ))}
 
-        {groupedBlendedFields.map(group => (
+        {searchedBlendedGroups.map(group => (
```
zapolsky, line 1321: **[Bug — verified]** Blended-column search matches render inside a **collapsed** group, so the matched field is invisible — the feature silently fails for blended columns, which is the case the changeset explicitly promises ("Search works across native and blended Data Mart columns").

**Evidence chain:**
- `report-column-search.ts:54-73` — a blended group is included in the results when one of its fields matches (`matchesBlendedField`), not only when the group title matches; the field-match branch returns `{ ...group, visibleFields }`, preserving `selectedCount`.
- `ReportColumnPicker.tsx:269` — `const [isOpen, setIsOpen] = useState(() => group.selectedCount > 0)` is a lazy initializer that runs only once on mount; a group with no *selected* fields starts collapsed (`isOpen === false`).
- `ReportColumnPicker.tsx:1019` — `searchedBlendedGroups.map(...)` renders `<BlendedGroupItem key={group.aliasPath} ...>` with no `open`/`forceOpen` prop. The stable `aliasPath` key preserves the same instance across keystrokes, so the initializer never re-runs and `isOpen` stays `false`; `selectedCount` is never recomputed from search matches (`groupedBlendedFields` memo deps, `:748-755`, exclude `searchQuery`).
- `ReportColumnPicker.tsx:276,307-326` — the matched fields render only inside `<CollapsibleContent>` under `<Collapsible open={isOpen}>`. `packages/ui/src/components/collapsible.tsx:9` re-exports the raw `@radix-ui/react-collapsible` primitive with no `forceMount`, so when `open` is false the content is not shown.

**Trigger:** with the default `showSelectedOnly = false`, open search and type a column name that lives in a blended group with no selected fields (the common "find a column I haven't picked yet" flow). The group header appears in the results but the matching field stays hidden behind the collapsed chevron until the user manually expands it. Matching by group name has the same problem.

**Fix:** force groups open while a query is active — e.g. pass `forceOpen={searchQuery.trim().length > 0}` into `BlendedGroupItem` and use `open={forceOpen || isOpen}` for the `Collapsible` (keep `setIsOpen` for manual toggling when no query is active).

zapolsky (commented):
🔴 **Blocked issue** — see the inline comment. This should be addressed before merge, but this review does not formally block; a human owns the merge decision. Posted by automated OWOX review.

zapolsky:

One blocking issue is posted inline (blended-column search matches hidden inside collapsed groups). The notes below do **not** block merge.

Primary notes (criticality ≥ 80):

- `apps/web/src/features/data-marts/edit/components/ReportColumnPicker/report-column-search.ts:10` & `:15` — **Locale-casing mismatch can drop real matches.** `normalizeSearchText` folds the *query* with `value.trim().toLowerCase()` (locale-independent), but `containsSearchText` folds the *field value* with `value.toLocaleLowerCase()`. In Turkic locales (`tr`/`az`) these diverge on the dotted/dotless `I` — e.g. value `"ID"` → `"ıd"`, query `"i"` → `"i"`, so `"ıd".includes("i")` is `false` and the field never matches. The sibling helper `FieldSearchPicker.tsx:19` uses `toLowerCase()` on both sides. Fix: use the same fold on both sides.
- `apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx:914-999` — **The "Disconnected columns" block is not filtered by the search and can contradict the empty state.** It renders from `unresolvedColumns`/`unresolvedFilterOnlyColumns`/`unresolvedSlices`, none of which the query touches, while `hasVisibleColumns` (`:777`) only counts native/blended results. So a query that matches nothing native/blended but where disconnected columns exist shows the full disconnected block **and** "No matching columns found." together — a self-contradicting screen. Fix: either filter the disconnected rows by the query too, or fold them into `hasVisibleColumns` so the empty-state message is suppressed when that block is shown.

<details>

- `ReportColumnPicker.tsx:858-874` — search `<Input>` has only a `placeholder`, no accessible name; `role='search'` on the wrapper does not name the field. Add `aria-label="Search columns"`.
- `ReportColumnPicker.tsx:489-491` → `:762-766` — `visibleNativeFields` is a fresh `.filter(...)` array each render when `showSelectedOnly` is on, so the search `useMemo` recomputes every render in that mode. Wrap `visibleNativeFields` in `useMemo` (or drop the memo — cheap either way).
- `ReportColumnPicker.tsx:790-792`, `:472-483` — "Select all" / `allSelected` operate on the full `selectableFieldNames`, ignoring the active search (pre-existing behavior, already true for "Show selected only"; writes to local form state with a dirty-guard, so recoverable). The new search makes this affordance more ambiguous — consider scoping it to the filtered set or clarifying the copy when a query is active.
- `report-column-search.ts:18-19`, `:28-29` — search matches some strings the user does not see (native dotted `name` when an `alias` is shown; `group.title`/`sourceDataMartTitle` which is never rendered in the group header — only `group.alias` is). This errs toward *more* results, not fewer, so it's not harmful, but it's a small deviation from "match what is shown." Intentional? If so, fine to ignore.
- `EmailReportEditForm.tsx:923` & `GoogleSheetsReportEditForm.tsx:442` — both drop the card wrapper styling (`border-border ... rounded-md border-b bg-white px-4 py-3 dark:...`) down to `space-y-3`. Looks intentional and consistent, but it is an unremarked visual change — worth a design glance.
- minor: `ReportColumnPicker.tsx:760`/`:859` `searchInputRef` is dead code (`.current` never read; `autoFocus` handles focus) — remove it · `:1002` empty-state label keys off raw `searchQuery` instead of the trimmed value, so a whitespace-only query in a genuinely empty list reads "No matching columns found." instead of "No fields available." · `:7` and `:12` are two separate `lucide-react` imports that can be merged · adjacent 28×28 icon buttons use different base variants (`OutputSettingsButton` `size='sm'` + `h-7 w-7` vs the search button `size='icon'` + `h-7 w-7`).

</details>

CI-parity (run in a worktree checkout): `tsc -b` ✓, `eslint` ✓. The 19 existing `ReportColumnPicker` tests pass. No issues attributable elsewhere.

<sub>Automated OWOX review · worktree multi-agent passes · the blocker was adversarially verified (chain closed); the "Select all" finding was adversarially refuted as a blocker and downgraded.</sub>

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/report-column-search.ts
```diff
@@ -0,0 +1,79 @@
...
+
+function matchesGroup(group: BlendedGroup, query: string): boolean {
+  return containsSearchText(group.alias, query) || containsSearchText(group.title, query);
+}
+
+/**
+ * Builds a searchable view of report columns without mutating
+ * the original native fields or blended groups.
+ */
+export function buildColumnSearchResult(
```
romandubovyi, line 39: 🟡 **Important:** Please add unit tests for this search logic, covering native field names and aliases, blended group and field matches, empty queries, and filtering without mutating the original groups.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/report-column-search.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  const visibleNativeFields = nativeFields.filter(field =>
+    matchesNativeField(field, normalizedQuery)
+  );
+
+  const visibleBlendedGroups = blendedGroups
+    .map(group => {
+      if (matchesGroup(group, normalizedQuery)) {
+        return group;
```
romandubovyi, line 60: 🟢 **Praise:** Nice separation of the search logic from the component. Preserving the original groups and creating filtered copies keeps this predictable and easy to test.

romandubovyi (dismissed):
No blocking issues found. The implementation is well structured; the remaining feedback is non-blocking.

romandubovyi (commented):
No blocking issues found. The implementation is well structured; the remaining feedback is non-blocking.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -823,9 +834,48 @@ export function ReportColumnPicker({
...
+          <Input
+            autoFocus
+            value={searchQuery}
+            placeholder='Search columns...'
+            aria-label='Search columns'
+            className='pl-8'
+            onChange={event => {
+              setSearchQuery(event.target.value);
+            }}
+            onKeyDown={event => {
```
romandubovyi: 🔴 **Blocking:** This `Escape` handler doesn't call `event.stopPropagation()`, so the keydown bubbles up to the enclosing `SheetContent` (Radix Dialog), which closes on `Escape` by default. Since this component is rendered inside `EmailReportEditSheet`/`GoogleSheetsReportEditSheet` (both guarded by `useUnsavedGuard`), pressing `Escape` to dismiss the search box will also trigger the sheet's close/unsaved-changes flow. Add `event.stopPropagation()` alongside the existing logic.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -966,7 +1018,7 @@ export function ReportColumnPicker({
           />
         ))}
 
-        {groupedBlendedFields.map(group => (
+        {searchedBlendedGroups.map(group => (
```
romandubovyi, line 1321: 🔴 **Blocking:** `BlendedGroupItem` decides whether a group is expanded once at mount time (`useState(() => group.selectedCount > 0)` in `BlendedGroupItem`, line 269). Since the component is keyed by `group.aliasPath`, the same instance is reused when `searchedBlendedGroups` changes on every keystroke — that initial `isOpen` never re-evaluates. When a search matches fields inside a group that started collapsed (the common case: searching for a not-yet-selected field), `buildColumnSearchResult` correctly narrows `visibleFields` here, but `CollapsibleContent` will not render them while `isOpen` stays false. The user has to manually expand the group to see the results the search just found. Consider forcing the group open while a search is active and it has matches, e.g. `open={isOpen || (hasSearchQuery && group.visibleFields.length > 0)}`.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/report-column-search.ts
```diff
@@ -0,0 +1,79 @@
...
+
+function matchesGroup(group: BlendedGroup, query: string): boolean {
+  return containsSearchText(group.alias, query) || containsSearchText(group.title, query);
+}
+
+/**
+ * Builds a searchable view of report columns without mutating
+ * the original native fields or blended groups.
+ */
+export function buildColumnSearchResult(
```
romandubovyi, line 39: 🟡 **Important:** This is pure, easily testable logic (query normalization, group-level vs field-level matching, partial group filtering), but there is no `report-column-search.test.ts` — unlike sibling helpers in this folder (`output-controls-display.test.ts`, `aggregation-config.test.ts`). Worth covering at least: empty query returns originals unchanged, a group-title match keeps all its fields, a field-level match returns only the matching subset, and no-match groups are dropped. Would also help catch UI-level regressions like the collapsed-group issue noted on `ReportColumnPicker.tsx`.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
artemfenoga: Thanks! I investigated this. Unfortunately event.stopPropagation() doesn't prevent Radix Dialog from closing on Escape because the dialog handles the key itself. A proper fix would require handling onEscapeKeyDown on SheetContent, but the search state currently lives inside ReportColumnPicker. I've removed the input-level Escape handler instead of leaving a handler that still closes the sheet.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/report-column-search.ts
artemfenoga: Good point, thanks! Added a dedicated report-column-search.test.ts covering the core search logic (empty query, group/title matching, field-level filtering, dropping non-matching groups, alias matching, and matchesColumnSearch), plus integration tests for the search behavior in ReportColumnPicker.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
artemfenoga: Good catch, thanks! I addressed this by automatically expanding groups when a search starts. Instead of forcing them open while a search is active, the group is opened once on search activation and remains user-controlled afterwards, so users can still collapse it if they want. I also added tests covering the search behavior to help prevent regressions.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/report-column-search.ts
artemfenoga: Done

romandubovyi (approved):
All previous comments and issues have been successfully resolved and tested. The implementation looks clean and works as expected. Approving the pull request.

PR #1373, opened 2026-06-29
feat(data-marts): report aggregations and totals

zapolsky:
## Summary

Server-side report aggregations and totals for Data Marts across all five storages (BigQuery, Athena, Snowflake, Redshift, Databricks) — group by dimensions, aggregate metrics, bucket dates, count unique entities, and compute totals, with no hand-written SQL. The matching report-picker UI ships in the `google-sheets-extension` repo (separate PR).

## What's included

- **Aggregate functions** — `SUM`/`AVG`/`MIN`/`MAX`/`COUNT`/`COUNT_DISTINCT`/`STRING_AGG`/`ANY_VALUE` and percentiles `P25`/`P50`/`P75`/`P95`, rendered per storage dialect. Multiple functions per column, each its own output column.
- **Output column naming** — `<column> | <TOKEN>` (e.g. `revenue | SUM`, `customer_id | COUNTUNIQUE`), an uppercase spreadsheet-style token. The `|` separator was empirically verified legal as an output alias on all five real warehouses.
- **Group-by-all** — every selected column without an aggregate function becomes a `GROUP BY` key.
- **Date bucketing** — `DAY`/`WEEK`/`MONTH`/`QUARTER`/`YEAR` with an optional per-rule IANA time zone (convert-then-truncate).
- **Auto Row Count** (`COUNT(*)`) for aggregated reports; opt-in **Unique Count** (`COUNT(DISTINCT primary key)`, composite keys supported).
- **Totals** — a separate no-`GROUP BY` warehouse query: every selected numeric field (main and joined) aggregated by all of its allowed functions; surfaced at the top level of the run response. Row Count / Unique Count are not part of totals.
- **Data-mart-level governance** — each field carries a role plus a supported-vs-default aggregation set per type, with per-field overrides; reports may only request a function the field allows.
- **Post-aggregation filtering (`HAVING`)** — filters on an aggregated output column are auto-routed apart from row-level `WHERE`.
- **Joined (blended) Data Marts** — report-level aggregation over the join roll-up, plus joined numeric fields in totals.
- **API + docs** — request/response API and OpenAPI contract, save-time validation, and a user guide (`docs/getting-started/setup-guide/report-aggregations.md`).

## Migrations

Three additive **nullable** JSON columns on `report` (`aggregationConfig`, `dateTruncConfig`, `uniqueCountConfig`). Expand-only, rolling-deploy safe, with reversible `down()`.

## Verification

- `nest build` + the full `data-marts` jest suite green; web `tsc -b` + vitest green.
- Real-warehouse integration coverage (aggregation + totals + blended) on all five storages (credential-gated).
- Output-alias `|` legality confirmed on all five real warehouses (BigQuery — the strictest — accepts it; the known-illegal parens are rejected, confirming the probe).

## Known limitations / follow-ups

- Blended `AVG` over a joined field is unweighted (avg-of-avgs) — accepted limitation, flagged in code.
- Percentiles are approximate on BigQuery/Athena and exact (continuous-interpolated) on Redshift/Snowflake/Databricks — documented.
- MCP `query_data_mart` and the `HAVING` filter UI are separate work items.


apps/backend/src/data-marts/data-storage-types/utils/sql-clause-renderer.ts
```diff
@@ -73,14 +102,248 @@ export abstract class SqlClauseRenderer {
...
+    const ref = (col: string): string => (qualify ? qualify(col) : this.quoteIdentifier(col));
+    if (pkColumns.length === 1) {
+      return `COUNT(DISTINCT ${ref(pkColumns[0])})`;
+    }
+    // Multi-column PK: concatenate with a unit-separator so distinct tuples stay distinct.
+    const SEP = "'␟'";
+    const castType = this.textCastType();
+    const parts = pkColumns.map(col => `COALESCE(CAST(${ref(col)} AS ${castType}), '')`);
+    const concatArgs = parts.join(`, ${SEP}, `);
+    return `COUNT(DISTINCT CONCAT(${concatArgs}))`;
```
zapolsky, line 209: **[Bug — verified] Redshift composite-PK Unique Count generates invalid SQL → the report run fails.**

For a composite primary key (≥2 PK columns) this emits a single N-ary `CONCAT` — for a 2-col PK, `COUNT(DISTINCT CONCAT(COALESCE(CAST(c1 AS VARCHAR),''), '␟', COALESCE(CAST(c2 AS VARCHAR),'')))` = **3 arguments**. Amazon Redshift's `CONCAT` is **binary-only** (exactly 2 args; per AWS docs, to join more you nest `CONCAT` or use `||`), so Redshift rejects this at run time. BigQuery / Snowflake / Athena / Databricks have variadic `CONCAT`, which is why the shared base is safe there but not on Redshift.

**Evidence chain:**
- `utils/sql-clause-renderer.ts:208-209` — base joins parts with `, '␟', ` into one `CONCAT(...)` (3+ args).
- `redshift/services/redshift-clause-renderer.ts:36-83` — overrides `textCastType`/`renderPercentile`/`renderStringAgg`/`renderDateTrunc` but **not** `renderCountDistinctPrimaryKey` (no override anywhere under `redshift/`), so it uses the base 3-arg form.
- Reached via `services/report-sql-composer.service.ts:126-140` (uniqueCount + primaryKeyColumns) → `redshift/services/redshift-query.builder.ts` → `renderAggregatedSelect` includeUniqueCount branch (`sql-clause-renderer.ts:260-263`). This is the **main** report query, not the best-effort totals query, so the run FAILS (`use-cases/run-report.service.ts:311-313`).
- Save-time validation only rejects **zero** PK fields (`services/output-controls-validator.service.ts:639-647`); a composite PK passes.
- `redshift-clause-renderer.aggregation.spec.ts:142-149` locks the broken 3-arg SQL via a string-only `toContain` (never executed against real Redshift).

**Trigger:** a Redshift data mart with a composite primary key (≥2 PK fields) + a report with Unique Count enabled.

**Fix:** override `renderCountDistinctPrimaryKey` in `RedshiftClauseRenderer` to join parts with `||` (`COALESCE(CAST(a AS VARCHAR),'') || '␟' || COALESCE(CAST(b AS VARCHAR),'')`), or nest `CONCAT` pairwise; convert the Redshift spec to an execution-level assertion.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-clause-renderer.ts
```diff
@@ -30,6 +31,27 @@ export class BigQueryClauseRenderer extends SqlClauseRenderer {
...
+
+  // DATE() coerces TIMESTAMP/DATETIME to DATE so one form covers all date/time types.
+  // With a time zone, DATE(timestamp, tz) converts to the date in that zone first.
+  protected override renderDateTrunc(
+    columnRef: string,
+    unit: DateTruncUnit,
+    timeZone?: string
+  ): string {
+    this.assertSafeDateTrunc(unit, timeZone);
+    const dateExpr = timeZone ? `DATE(${columnRef}, '${timeZone}')` : `DATE(${columnRef})`;
```
zapolsky: **[Bug — verified] BigQuery date bucketing emits an invalid `DATE()` call → the report run fails.** Two trigger cases, one root cause: `renderDateTrunc` is type-blind and always wraps the column in `DATE(...)`.

**Case 1 — common path: a pure `DATE` dimension, no time zone** → `DATE_TRUNC(DATE(\`d\`), MONTH)`. BigQuery's `DATE()` overloads are `DATE(timestamp[, tz])`, `DATE(datetime)`, `DATE(y,m,d)` — there is **no `DATE(DATE)` overload** and no implicit DATE→DATETIME/TIMESTAMP coercion, so `DATE(date_col)` errors (`No matching signature for function DATE for argument types: DATE`). Any BigQuery report grouping a `DATE`-typed column by DAY/WEEK/MONTH/QUARTER/YEAR hits this.

**Case 2 — a `DATETIME` column + time zone** → `DATE(datetime_col, 'tz')`. There is no `DATE(DATETIME, STRING)` overload either, so it errors the same way. The validator's `DATE_TRUNC_TIMEZONE_REQUIRES_TIMESTAMP` guard fires only for pure `DATE`; `DATETIME` is in `TIMESTAMP_TYPES` and slips through (`services/output-controls-validator.service.ts:51, 387-398`).

**Evidence chain:**
- `bigquery/services/bigquery-clause-renderer.ts:51-52` — `dateExpr = timeZone ? DATE(col, 'tz') : DATE(col)`, then wrapped in `DATE_TRUNC(dateExpr, unit)`. No column type is consulted, so `DATE(col)` is emitted for every date/time type.
- `dto/schemas/field-type-category.ts` — `DATE_TYPES` includes both `DATE` and `DATETIME`; the validator admits date-trunc on both (`output-controls-validator.service.ts:371`) and admits tz on `DATETIME` (`:387-398`).
- Same renderer on flat and blended paths. Only string-only specs cover it (`bigquery-clause-renderer.aggregation.spec.ts:179-186`); the credential-gated real-BQ integration test buckets only a `TIMESTAMP` column, so neither failing path is ever executed.

**Trigger:** Case 1 — any BigQuery report date-bucketing a `DATE` column (the common case); Case 2 — date-bucketing a `DATETIME` column with a time zone.

**Fix:** make `renderDateTrunc` type-aware — don't wrap a column already typed `DATE` (`DATE_TRUNC` accepts a `DATE` directly); for `DATETIME`+tz convert via `TIMESTAMP` first (`DATE(TIMESTAMP(col), 'tz')`) or reject tz on tz-naive `DATETIME` at validation. Worth a real-BQ probe + an execution-level test for the `DATE`-column path. (Even if BQ tolerated `DATE(DATE)`, dropping the redundant wrap is the right fix.)

zapolsky (commented):
🔴 **Blocked issues** — see the inline comments. These should be addressed before merge, but this review does not formally block; a human owns the merge decision. Posted by automated OWOX review (worktree multi-agent passes; every blocker adversarially verified).

Three verified, user-reachable runtime failures / traps in the new aggregation path. CI-parity is green (`tsc -b` ✓, `eslint` ✓, jest `data-storage-types`+validator+totals 1148 ✓) — these are semantic bugs the type-checker and string-only specs don't catch. Non-blocking notes (incl. a confirmed HAVING-path 500 that is API-only until the HAVING UI ships, and a totals cost-bleed) are in the top-level comment.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -306,6 +603,69 @@ export class OutputControlsValidatorService {
...
+
+        // Unique Count emits a COUNT(DISTINCT <pk tuple>) column only when the data
+        // mart has at least one primary-key field. Without a PK the SQL no-ops but
+        // the header is still appended → header/column mismatch (silent data
+        // corruption). Reject at save time so the bad state can never be persisted.
+        if (args.uniqueCountConfig === true) {
+          const pkFields = getPrimaryKeyFields(blendableSchema.nativeFields);
+          if (pkFields.length === 0) {
+            errors.push({
+              code: 'UNIQUE_COUNT_REQUIRES_PRIMARY_KEY',
```
zapolsky, line 674: **[Bug — verified] Unique Count is a one-way trap: once the primary key is removed, the report can no longer be saved or edited — a broken recovery path the user cannot escape through the product.**

**Evidence chain:**
- A report saved with `uniqueCountConfig: true` is valid only while the mart has a PK.
- If the PK is later removed (an analyst unsets `isPrimaryKey`, or drops the PK column), the web toggle disappears — it renders only under `hasPrimaryKey && outputControlsAvailable` (`apps/web/src/.../ReportColumnPicker/ReportColumnPicker.tsx:1186`), so there is no UI to clear the stored `true`.
- The form round-trips the stored value back on **every** save: read mapper `apps/web/.../reports/shared/model/mappers/report.mapper.ts:43` → form init → submit `useGoogleSheetsReportForm.ts:155` (same for the Email/Looker forms). Update is a full-replace PUT that validates the **command** value (`use-cases/update-report.service.ts:121-133`), not gated on whether uniqueCount changed.
- The backend rejects `uniqueCountConfig: true` with 0 PK fields on every save here (line 642, `UNIQUE_COUNT_REQUIRES_PRIMARY_KEY`, 400). So even an unrelated edit (e.g. renaming the report) is rejected, and the only `false`-setter is the now-hidden toggle (the `Clear` fallback renders only when output controls are *unsupported*, which can't be the case here).

**Trigger:** save a report with Unique Count → remove the data mart's primary key → every subsequent save of that report 400s, with no in-product way to clear the flag.

**Fix:** auto-reset `uniqueCountConfig` to `false` (web) when `hasPrimaryKey` is false, or render a clear/reset affordance whenever `uniqueCountConfig === true` even without a PK; optionally relax the backend so uniqueCount-without-PK degrades to a no-op on update rather than a hard 400.

zapolsky:

Worktree multi-agent review of an unusually well-guarded PR — the SQL injection surface (identifier escaper round-trip guard, IANA-tz allowlist, date-trunc unit allowlist, parameterized values, `assertSafeDateTrunc` in all 5 dialects) and the alias↔header contract were traced and **disproved** as bug sources. CI-parity is green: `tsc -b` ✓, `eslint` ✓, jest (`data-storage-types` + `output-controls-validator` + `report-totals`) **1148 passed**. The three 🔴 inline items are the verified user-reachable failures. Below are the non-blocking findings, highest criticality first — none block merge.

- **HAVING on `COUNT`/`COUNT_DISTINCT`/`STRING_AGG` of a date/time column → run 500 on BigQuery, Legacy BigQuery, Athena, Snowflake, Databricks (Redshift safe).** `validateHavingFilters` validates the operator against the aggregate's **effective** type (`COUNT_DISTINCT(date)`→integer, `output-controls-validator.service.ts:215-246` via `computeEffectiveType`), so `COUNT_DISTINCT(visit_date) >= 5` saves clean. But at run time each builder hands `renderHaving` a **raw**-type resolver (`bigquery-query.builder.ts:94-99`, athena/snowflake/databricks similarly; blended `abstract-blended-query-builder.ts:216`), so `renderFilterFragment` casts the RHS to the date type → `HAVING COUNT(DISTINCT \`visit_date\`) >= CAST(@h0 AS TIMESTAMP)` (integer vs timestamp). *Not reachable via the product UI in this release* (the HAVING filter UI is separate work — no filter editor sets `function`), but it is **API-reachable now and latent in merged backend code that the upcoming HAVING UI will expose.** Fix: resolve the cast type from the aggregate's effective type in the HAVING path (run `computeEffectiveType` before `renderFilterFragment`, or pass no resolver and cast on effective type). Worth fixing before the HAVING UI ships.
- **Totals fire an extra full-dataset aggregate query on every server-side run, but the result is surfaced to no Sheets/Email user → wasted warehouse cost.** `persistTotalsBestEffort` is called unconditionally and `await`ed before streaming (`run-report.service.ts:294-295`), and `composeTotals` returns non-null whenever the projection has ≥1 numeric column (`report-sql-composer.service.ts:257-301`) — even on non-aggregated exports. The result lands on `run.totals` (REST top-level via `data-mart.mapper.ts` `extractTotals`) but **no web view renders it and no destination writer emits it** (a repo-wide `apps/web/src` grep for `.totals` hits only the mapper). Only the HTTP/Looker path (`stream-http-data.service.ts:182`) actually consumes totals. Net: an extra scan (BQ bytes / warehouse compute) + latency on every Sheets/Email run, for zero visible output. Fix: gate totals to contexts that surface them (skip for Sheets/Email destinations), or finish wiring the web run-history view to read `run.totals`.
- **Blended HAVING + `placement:'pre-join'` is silently dropped → wrong (unfiltered) results.** The filter schema allows a rule to carry both `function` and `placement:'pre-join'` (`filter-config.schema.ts:66-75`, no refinement), and neither validator branch rejects the combo (`output-controls-validator.service.ts:180, 224`). On the blended path the rule routes to the pre-join CTE by placement (`abstract-blended-query-builder.ts:118,135`), where `renderWhere` drops function-carrying rules (`sql-clause-renderer.ts:81`); `renderHaving` only sees post-join rules → the HAVING constraint is applied **nowhere** (silent wrong rows when the HAVING targets a joined field; a 500 when it targets a native field). API-reachable only (the UI keeps slices and metric-filters separate), flat path unaffected. Fix: reject `function` + `placement:'pre-join'` in the validator (a HAVING is inherently post-aggregation).
- **`COUNT_DISTINCT` is on-by-default for the `other` field-type category, but `COUNT(DISTINCT <non-groupable>)` is rejected by every warehouse → save-clean-then-500.** `DEFAULTS_BY_CATEGORY.other` includes `COUNT_DISTINCT` (`field-aggregation-governance.ts:44`) and the validator's only hard type-floor is for `SUM`/`AVG` (`output-controls-validator.service.ts:316-327`), so `COUNT_DISTINCT` on a `JSON`/`GEOGRAPHY`/`ARRAY`/`STRUCT`/`SUPER`/`VARIANT` column passes and then errors at run time (these types aren't groupable/comparable). Reachable when such a type is a selectable leaf. Fix: add a validator floor rejecting `COUNT_DISTINCT` (and `STRING_AGG`) on the `other` category, or narrow the `other` governance set to `COUNT`/`ANY_VALUE`.
- **Cross-dialect timezone date-trunc semantics diverge for tz-naive columns — the same report buckets differently per warehouse.** All dialects convert-then-truncate, but the assumed *source* zone differs: BigQuery `DATE(ts,'tz')` treats a `TIMESTAMP` as an absolute UTC instant; Athena `col AT TIME ZONE 'tz'` interprets a tz-naive `timestamp` in the **session** zone (`athena-clause-renderer.ts:85-93`); Snowflake 2-arg `CONVERT_TIMEZONE('tz', col)` on `TIMESTAMP_NTZ` is session-dependent (`snowflake-clause-renderer.ts:80-88`); Redshift/Databricks assume UTC. "Revenue by month in America/New_York" can yield different buckets across storages. Fix: pin a UTC source-zone contract and make each dialect explicit (e.g. Snowflake `CONVERT_TIMEZONE('UTC','<tz>',col)`), or restrict tz date-trunc to genuinely tz-aware column types.
- **Cross-repo version-skew / rollout coupling.** The aggregate-function list must stay in sync with the external `google-sheets-extension` report-picker (separate repo), which has no compile-time guard against drift (the in-monorepo `apps/web` is protected by paired sync comments + an exhaustive `never` switch, but the external repo is not). The API change is additive and safe new-server/old-client; the asymmetric risk is new-client/old-server. Document **backend-first deploy ordering** for this feature, and consider a shared constant / contract test for the cross-repo seam rather than prose "keep in sync" comments.

<details>
<summary>Additional observations (below the high-signal bar — non-blocking, listed for completeness)</summary>

- **Question (unproven — needs a live dry-run):** percentile / `STRING_AGG` inside `HAVING` re-emits the ordered-set / `LISTAGG` aggregate expression; some engines (notably Redshift) restrict ordered-set aggregates and `LISTAGG` in `HAVING`. All HAVING specs are string-only, so this path is unverified against real Redshift/Snowflake/Databricks. Worth a per-dialect dry-run.
- `services/output-controls-validator.service.ts:661-668` — the totals validation pass passes `includeRowCount = (parsedAggregations.length > 0)` while the totals SQL is built with `rowCount: false`, so the validator's collision set contains a `Row Count` the totals query never emits. Harmless except for a numeric column literally named `Row Count` (spurious `OUTPUT_COLUMN_NAME_COLLISION`, swallowed since totals are best-effort). Tighten the validator's `includeRowCount` to reflect the plan's `rowCount` flag.
- An aggregation on a `columnConfig` entry that is absent from the resolved schema bypasses both the SUM/AVG type-floor and governance (`resolveType`/`resolveAllowed` return `undefined`, `output-controls-validator.service.ts:316-337`); on the flat path this surfaces as a DWH 500 rather than a 400.
- `apps/web/.../SourceFieldsTable.tsx:319` shows "None" for an un-overridden joined field's available aggregations (`field.postJoinAggregations ?? []`), whereas the native editor falls back to `resolveFieldGovernance(...).allowedAggregations` (`BaseSchemaTable.tsx:265-266`) — asymmetric/misleading UX.
- `field-aggregation.ts:46-60` (`integerTypeFor`) and `:84-98` (`getStringType`) lack the exhaustiveness `never`-guard that `getFloatType` (`:75-80`) has — a future 6th storage would return `undefined` silently. Not reachable for the current 5.
- Composite-PK Unique Count can undercount when a PK part is NULL: `COALESCE(CAST(x AS …),'')` maps `NULL` and `''` to the same token (`sql-clause-renderer.ts:204-209`), so `(NULL,'a')` and `('','a')` collide. Minor (PK parts are normally NOT NULL).
- Same recovery-trap class as the Unique Count soft-lock, but milder: a `dateTruncConfig` tz entry becomes unsaveable if the column type later changes `TIMESTAMP`→`DATE` (`DATE_TRUNC_TIMEZONE_REQUIRES_TIMESTAMP`) — but it IS clearable via the per-rule AGG control, so it's recoverable.
- minor: COUNT family annotated as 32-bit `INTEGER` though `COUNT`/`COUNT(DISTINCT)` return 64-bit on these engines (`field-aggregation.ts:46-60`) — header metadata only, values not truncated; percentiles approximate on BigQuery/Athena vs exact-interpolated on Redshift/Snowflake/Databricks — same label, different number (documented limitation); `uniqueCountConfig` stored as a bare boolean in a `json` column (future-proofing, slightly unusual); three separate migration files for three columns (consistent with repo precedent); dotted/nested column names yield an illegal aggregated alias `"user"."id | SUM"` (acknowledged in `aggregation-labels.ts:57-62`); blended `AVG`/percentile re-aggregation is avg-of-avgs (acknowledged TODO #6680); `number` category cannot be `COUNT`-ed (governance excludes it — confirm intended); a date-trunc-only report (no `aggregationConfig`) gets no auto Row Count (confirm intended vs the "Row Count for aggregated reports" spec); minor web UX in the aggregation picker ("Add aggregation" pending flow can strand the picker; stale `timeZone` re-emitted if a TIMESTAMP bucket's column becomes DATE; `uniqueCountConfig` omitted from `FormSection` field lists).

</details>

<sub>Automated OWOX review · worktree multi-agent passes (reviewer · logic-hunter · architect · tester · fresh-eyes generalist; second wave on bug-rich PR) · all blockers adversarially verified with an independent skeptic gate (8 skeptic passes; 0 refuted). Pure commenter — a maintainer owns the merge decision.</sub>

zapolsky:
Thanks for the thorough pass. Addressed in two follow-up commits on top of the reviewed SHA:

- `034102e1e` — blockers + non-blocking fixes
- `827b99fb6` — totals gate

### 🔴 Blockers

**1. Redshift composite-PK Unique Count (N-ary `CONCAT`)** — Fixed. `RedshiftClauseRenderer` now overrides `renderCountDistinctPrimaryKey` to join the key parts with `||` instead of an N-ary `CONCAT`. Verified live against real Redshift: the 3-arg `CONCAT` is rejected (`function concat(character varying, "unknown", character varying) does not exist`), and the `||` chain returns the joined key and works inside `COUNT(DISTINCT …)`.

**2. BigQuery date bucketing (`DATE(...)`)** — one of the two cases reproduced.

- **Case 2 (DATETIME + time zone) is real** and fixed: `DATE(DATETIME, tz)` has no overload, so a tz-naive `DATETIME` with a time zone is now normalized via `TIMESTAMP(col, tz)` first. `renderDateTrunc` is type-aware now — the column type is threaded through `renderAggregatedSelect` on all flat + blended builders.
- **Case 1 (a plain `DATE` column, no tz) does NOT reproduce.** I probed it on real BigQuery (the comment recommended a probe): `DATE(DATE '…')`, `DATE_TRUNC(DATE(d), MONTH)`, and `DATE_TRUNC(d, MONTH)` against a `DATE`-typed **column** (CTE, not a folded literal) all execute successfully under both a dry-run and real execution. BigQuery accepts `DATE(DATE)` as a no-op, so the common path was never failing. I still dropped the redundant `DATE()` wrap for `DATE` columns as suggested, and added the execution-level coverage that was missing (the integration test only bucketed `TIMESTAMP`).

**3. Unique Count one-way trap** — Fixed. The report column picker auto-clears `uniqueCountConfig` once the schema loads without a primary key (gated on the loaded schema so it can't clear a legitimately PK-backed config before load), so the report stays editable.

### 🔵 Non-blocking (also addressed)

- **HAVING cast type** — Fixed. The HAVING value is now cast to the aggregate's **effective** type (`COUNT_DISTINCT(date)` → integer) rather than the raw column type, via a shared `effectiveComparisonType` helper wired into the BigQuery / Snowflake / Athena / Databricks + blended resolvers (Redshift inlines literals, so it was already safe). Verified live on BigQuery: `COUNT(DISTINCT date) >= 5` compares as an integer; the old raw-DATE cast (`>= CAST('…' AS TIMESTAMP)`) is rejected.
- **Blended HAVING + `placement:'pre-join'`** — Fixed. The validator now rejects `function` + `placement:'pre-join'` (a HAVING is inherently post-aggregation).
- **`COUNT_DISTINCT` / `STRING_AGG` on the `other` category** — Fixed. The validator type-floor now rejects them on non-groupable types (JSON / geography / array / struct / …), turning the save-clean-then-500 into a clean 400.
- **Totals cost on push destinations** — Fixed. `persistTotalsBestEffort` is gated to destinations that surface totals; Google Sheets and the message-based destinations (Email/Slack/Teams/Chat) skip the extra full-dataset scan.
- **Cross-repo version skew** — Documented backend-first deploy ordering in the changeset.

### Left as-is (by design / follow-up)

avg-of-avgs blended re-aggregation, percentile approx-vs-exact (documented), `ANY_VALUE` per-dialect render test, WEEK start-day / tz source-zone divergence, composite-PK `NULL`↔`''` collision. The cross-dialect HAVING param-binding on the non-BigQuery dialects (the "unproven" item) is worth a live dry-run before the HAVING filter UI ships.

Full `nest build` + 4677 data-marts jest + web `tsc -b` / vitest are green; the date-trunc, composite-PK, and HAVING forms were re-verified on real BigQuery and Redshift.

romandubovyi:
🔴 **Blocking:** `LegacyBigQueryQueryBuilder.buildQuery` — its local `hasOutputControls` (legacy-bigquery-query.builder.ts:30) only checks `filters`/`sort`/`limit`, but this PR expanded the parent's notion of output controls to include `aggregations`, `dateTruncs`, `rowCount`, and `uniqueCount`. An aggregated or Totals request without a filter/sort/limit (e.g. `composeTotals`, which sets sort/limit empty and `rowCount:false`) evaluates this to `false`, bypasses `super.buildQuery`, and silently drops the GROUP BY/aggregation — returning ungrouped rows. `LEGACY_GOOGLE_BIGQUERY` is in the supported set, so this is reachable and uncovered by legacy specs. Please mirror the parent's full output-controls condition here (or delegate to the shared `hasOutputControls(report)` helper).

_(Posted as a PR-level comment because the `hasOutputControls` block is a context line untouched by this PR and cannot take an inline comment.)_

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
```diff
@@ -176,16 +191,62 @@ export abstract class AbstractBlendedQueryBuilder implements BlendedQueryBuilder
...
+    // semantics). The CTE machinery and pre-join rollup are unchanged.
+    if (aggregated && renderer) {
+      const agg = renderer.renderAggregatedSelect(
+        context.columns,
+        context.aggregations ?? [],
+        this.buildDateTruncMap(context.dateTruncs),
+        {
+          includeRowCount: context.rowCount === true,
+          includeUniqueCount: context.uniqueCount === true,
+          primaryKeyColumns: context.primaryKeyColumns,
```
romandubovyi, line 213: 🔴 **Blocking:** `primaryKeyColumns` are passed into `renderAggregatedSelect` but are never added to `referencedColumns`, so the `main` CTE's explicit projection omits any PK column that is not also a selected/filtered/sorted column. The outer unique-count expression emits `COUNT(DISTINCT main.<pk>)`, which then references a column the CTE doesn't project. This breaks a common config — blended report + Unique Count where the PK isn't chosen as a dimension. The validator only checks that a PK *exists*, not that it's projected, and the specs only assert `.toContain(...)` the fragment without asserting the `main` projection. Please include `context.primaryKeyColumns` in the set feeding the `main` projection when `uniqueCount` is active.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -838,6 +1056,16 @@ export function ReportColumnPicker({
         </div>
       )}
 
+      {aggSettingsOpen && onOutputConfigChange && outputControlsSupported && (
+        <div className='rounded-md border'>
+          <AggregationSettingsDropdown
+            value={effectiveOutputConfig}
+            onChange={onOutputConfigChange}
```
romandubovyi: 🔴 **Blocking:** Adding an aggregation through the Aggregations dropdown panel only calls `onOutputConfigChange`, so it doesn't materialize an implicit `columnConfig` selection. For a new report where `columnConfig` starts as `null`, this leaves `aggregationConfig` set while `columnConfig` is still `null`, which the backend rejects with `AGGREGATION_REQUIRES_COLUMN_CONFIG`. The row-level Σ path already handles this in `handleApplyAggregation`; the panel path needs the same materialization (e.g. a wrapper that also calls `onChange(effectiveValue)` when aggregations/dateTruncs become non-empty and `value === null`).

apps/backend/src/data-marts/data-storage-types/utils/sql-clause-renderer.ts
```diff
@@ -73,14 +102,262 @@ export abstract class SqlClauseRenderer {
...
+    const groupByParts: string[] = [];
+    const selectParts = columns.flatMap(c => {
+      const ref = qualify ? qualify(c) : this.quoteIdentifier(c);
+      const fns = aggregationFunctionsForColumn(aggregations, c);
+      if (fns.length > 0) {
+        // One SELECT item per function, in rule order. The column is an aggregated
+        // metric — never a GROUP BY key. aliasByColumn points at the FIRST function's
+        // alias so ORDER BY on the column resolves to its first aggregation.
+        const items = fns.map(fn => {
+          const alias = this.quoteIdentifier(aggregatedColumnLabel(c, fn));
```
romandubovyi, line 241: 🟡 **Important:** `quoteIdentifier` splits dotted identifiers into a qualified reference (`` `a`.`b` ``), so using it to build a single-token output **alias** is wrong for any column containing a dot. `aggregatedColumnLabel('metrics.revenue','SUM')` → alias `` `metrics`.`revenue | SUM` ``, which is invalid SQL after `AS` on BigQuery, and also diverges from the un-split header name in `resolveReportDataHeaders` (so the reader nulls the value). Since `deriveNumericTotalsAggregations` auto-aggregates nested numeric leaves, a mart with a STRUCT/RECORD numeric subfield silently breaks the Totals query. The unit test misses this because `StubRenderer.quoteIdentifier` doesn't split on dots. Consider a dedicated alias-quoting routine that treats the whole label as one identifier (never splits on `.`), add a dotted-column spec case, and apply the same fix to the qualified dimension alias at lines 247/264.

apps/backend/src/data-marts/dto/schemas/field-aggregation-governance.ts
```diff
@@ -0,0 +1,86 @@
...
+ * supported (above) but never a default. Single source so the governance map and any UI
+ * hint cannot diverge.
+ */
+const DEFAULTS_BY_CATEGORY: Record<FieldTypeCategory, CategoryDefault> = {
+  number: { role: 'metric', allowedAggregations: ['SUM', 'AVG', 'MIN', 'MAX'] },
+  string: { role: 'dimension', allowedAggregations: ['COUNT', 'COUNT_DISTINCT'] },
+  date: { role: 'dimension', allowedAggregations: ['MIN', 'MAX'] },
+  time: { role: 'dimension', allowedAggregations: ['MIN', 'MAX'] },
+  boolean: { role: 'dimension', allowedAggregations: ['COUNT', 'COUNT_DISTINCT'] },
+  other: { role: 'dimension', allowedAggregations: ['COUNT', 'COUNT_DISTINCT'] },
```
romandubovyi: 🟡 **Important:** The `other` category lists `COUNT_DISTINCT` as both supported (line 29) and default-on (line 44), but `validateAggregations` hard-rejects `COUNT_DISTINCT`/`STRING_AGG` on any `other`-typed field via a type-floor that runs and `continue`s before the governance check (output-controls-validator.service.ts:348-358). The web mirror (`aggregation-governance.ts`) offers it too, so a user who picks this UI-offered default always gets a 400 `AGGREGATION_FUNCTION_NOT_ALLOWED_FOR_TYPE`. Either drop `COUNT_DISTINCT` from `other` in both `SUPPORTED_BY_CATEGORY`/`DEFAULTS_BY_CATEGORY` (and the web mirror), or relax the validator — right now the governance map and the validator disagree and the menu advertises an option that can never save.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
```diff
@@ -30,14 +31,41 @@ export function collectSchemaFieldPathTypes(
...
+    if ('fields' in field && field.fields?.length) {
+      result.push(...collectSchemaFieldPathDescriptors(field.fields, fullName));
+    }
+  }
+  return result;
+}
+
+// Returns the subset of fields (recursively through nested containers) where
+// isPrimaryKey is true. Mirrors the traversal in collectSchemaFieldPathDescriptors.
+export function getPrimaryKeyFields(fields: readonly DataMartSchemaField[]): DataMartSchemaField[] {
```
romandubovyi: 🟡 **Important:** Despite the "Mirrors the traversal in `collectSchemaFieldPathDescriptors`" comment, this skips both the `isConnected`/`isHiddenForReporting` pruning and the dotted-path (`prefix.name`) construction that the sibling traversal performs. Since the result feeds `primaryKeyColumns` for `COUNT(DISTINCT …)`, a disconnected PK references a dropped column (SQL error) and a nested PK emits only its leaf name (wrong column). Consider reusing the descriptor traversal (filtered on `isPrimaryKey`) so PK paths stay consistent and pruned.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
```diff
@@ -543,6 +613,7 @@ export abstract class AbstractBlendedQueryBuilder implements BlendedQueryBuilder
       case 'ANY_VALUE':
         return 'MAX';
       default:
+        // TODO(#6680 joined-DM aggregation): AVG re-aggregation is incorrect (avg of avgs); handle when joined-DM aggregation lands.
```
romandubovyi, line 626: 🟡 **Important:** `COUNT_DISTINCT` re-aggregates to `SUM`, so for a 2+ level blend the result is a sum of per-group distinct counts, not a global distinct count — it over-counts values appearing in more than one child group. This is as semantically wrong as the `AVG` avg-of-avgs case this TODO documents, but unlike `AVG` it's neither flagged nor documented, and a spec enshrines it as correct (`COUNT_DISTINCT becomes SUM at parent level`). Please either document it as a known limitation alongside this AVG TODO or compute it correctly, so users don't silently get inflated "unique" metrics on transitive blends.

apps/web/src/features/data-marts/shared/types/relationship.types.ts
```diff
@@ -97,6 +110,11 @@ export interface BlendedField {
   transitiveDepth: number;
   aliasPath: string;
   outputPrefix: string;
+  /** Aggregation governance — absent fields fall back to type-derived defaults. */
+  aggregationRole?: AggregationRole;
+  allowedAggregations?: ReportAggregateFunction[];
+  /** Analyst-allowed post-join aggregation set. Empty means no restriction. */
```
romandubovyi: 🔵 **Question:** This comment ("Empty means no restriction") contradicts the implemented governance — everywhere else (`resolveColumnAllowedAggregations`, the backend resolvers, and `cleanBlendedFieldOverride`'s own comment) an explicit `[]` means "none allowed", distinct from `undefined` (fall back to type defaults). The code looks correct and consistent with the server; could you fix the comment so it doesn't invite an inverted change later?

zapolsky:
Thanks @romandubovyi — all eight addressed in two commits: `3f3349fa4` (items 1–3, 5–8) and `4dbd5620a` (item 4). Full `nest build` + 4676 data-marts jest + web `tsc -b` / vitest green; the date-trunc / composite-PK / HAVING / dotted-alias forms re-verified on real BigQuery & Redshift.

### 🔴 Blocking

1. **Legacy BigQuery `hasOutputControls`** — fixed. The local condition now mirrors the parent's full notion (aggregations / date-truncs / Row Count / Unique Count, not just filter/sort/limit), so an aggregated or Totals request with none of the latter delegates to `super.buildQuery` instead of dropping the GROUP BY. (The shared `hasOutputControls(report)` helper already covers these, so the view/`mainTableReference` is created for the aggregated case.)

2. **Blended Unique Count — PK not projected** — fixed. `context.primaryKeyColumns` is added to the main-CTE `referencedColumns` set when `uniqueCount` is active, so `COUNT(DISTINCT main.<pk>)` references a column the CTE projects, even when the PK isn't a selected/filtered/sorted dimension.

3. **Aggregations panel doesn't materialize `columnConfig`** — fixed. The panel `onChange` now runs the same materialization as the row-level Σ path (`onChange(effectiveValue)` when aggregations/date-truncs become non-empty while columns are still implicit), so a brand-new report no longer 400s with `AGGREGATION_REQUIRES_COLUMN_CONFIG`.

### 🟡 Important

4. **Dotted-column alias** — fixed, but a heads-up: the suggested *"quote the whole label as one identifier (never split on `.`)"* **does not work** — I probed it on real BigQuery and a dot in an output alias is rejected **entirely**, even when the whole thing is back-tick-quoted:
   - `` `metrics.revenue | SUM` `` → rejected
   - `` `metrics`.`revenue | SUM` `` → rejected
   - `` `metrics_revenue | SUM` `` → accepted

   So the fix sanitizes the column path (dots → `_`) in `aggregatedColumnLabel` (`metrics.revenue | SUM` → `metrics_revenue | SUM`). The alias and the header both come from that function so they stay in sync, and the aggregate **argument** keeps the real dotted struct reference (`` `metrics`.`revenue` ``). Collisions are caught by `OUTPUT_COLUMN_NAME_COLLISION`. Added a dotted-column unit case. NB: a dotted column used as a **grouping dimension** (the qualified/date-trunc `AS` alias, and the flat-path nested dimension) is a separate, pre-existing round-trip gap I left as a known limitation — happy to take it as a follow-up if you'd like it in scope.

5. **`other` category advertises `COUNT_DISTINCT` but the validator rejects it** — fixed by narrowing governance (the validator floor stays as defense-in-depth). Dropped `COUNT_DISTINCT` from `SUPPORTED_BY_CATEGORY.other` / `DEFAULTS_BY_CATEGORY.other` in the backend and the in-repo web mirror, so the menu only offers `COUNT` / `ANY_VALUE` for non-groupable types. The external `google-sheets-extension` mirror needs the same one-line change; I'll fold it into that repo's PR (it isn't open yet).

6. **PK path traversal** — fixed. `getPrimaryKeyFields` now reuses the pruned descriptor traversal (`collectSchemaFieldPathDescriptors` filtered on `isPrimaryKey`), so a disconnected/hidden PK is dropped and a nested PK keeps its full `parent.child` path instead of just the leaf. Added disconnected/hidden/nested spec cases.

7. **`COUNT_DISTINCT` → `SUM` re-aggregation over-counts on transitive blends** — documented as a known limitation in `getReAggregateFunction` (and split from the correct `COUNT` → `SUM` case), alongside the existing AVG avg-of-avgs note; the spec title/comment no longer imply it's semantically correct. A correct transitive distinct needs the raw values at the parent level — same follow-up bucket as AVG.

### 🔵 Question

8. **`postJoinAggregations` comment** — fixed. It now states `[]` = none allowed, `undefined` = fall back to type defaults, consistent with `resolveColumnAllowedAggregations` / `cleanBlendedFieldOverride`.

> Context on Totals: per a team decision they're now computed only on the HTTP Data API path (and an upcoming MCP task), not on the push run path — `RunReportService.executeReport` no longer computes them (`447d95cb3`), which is why several of the above touch `composeTotals` / `deriveNumericTotalsAggregations` but not the push path.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
zapolsky: Fixed in `3f3349fa4`: `context.primaryKeyColumns` is added to the main-CTE `referencedColumns` set when `uniqueCount` is active, so `COUNT(DISTINCT main.<pk>)` references a column the CTE projects even when the PK isn't a selected/filtered/sorted dimension.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
zapolsky: Fixed in `3f3349fa4`: the panel `onChange` now runs the same materialization as the row-level Σ path — it calls `onChange(effectiveValue)` when aggregations/date-truncs become non-empty while columns are still implicit — so a brand-new report no longer 400s with `AGGREGATION_REQUIRES_COLUMN_CONFIG`.

apps/backend/src/data-marts/data-storage-types/utils/sql-clause-renderer.ts
zapolsky: Fixed in `4dbd5620a` — but a heads-up: the "quote the whole label as one identifier" approach **doesn't work**. Real-BQ probe (dry-run):

- `` `metrics.revenue | SUM` `` → rejected
- `` `metrics`.`revenue | SUM` `` → rejected
- `` `metrics_revenue | SUM` `` → accepted

BigQuery rejects a dot in an output alias **entirely**, even fully back-tick-quoted. So the fix sanitizes the column path (dots → `_`) in `aggregatedColumnLabel` (`metrics.revenue | SUM` → `metrics_revenue | SUM`); alias and header both come from there so they stay in sync, and the aggregate **argument** keeps the real dotted struct ref (`` `metrics`.`revenue` ``). Added a dotted-column unit case. A dotted column used as a **grouping dimension** (the qualified/date-trunc `AS` alias + the flat-path nested dimension) is a separate, pre-existing round-trip gap I left as a known limitation — happy to take it as a follow-up.

apps/backend/src/data-marts/dto/schemas/field-aggregation-governance.ts
zapolsky: Fixed in `3f3349fa4` by narrowing governance: dropped `COUNT_DISTINCT` from `SUPPORTED_BY_CATEGORY.other` and `DEFAULTS_BY_CATEGORY.other` (backend + in-repo web mirror), so the menu offers only `COUNT`/`ANY_VALUE` for non-groupable types. The validator floor stays as defense-in-depth. The external `google-sheets-extension` mirror needs the same one-line change — I'll fold it into that repo's PR (not open yet).

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
zapolsky: Fixed in `3f3349fa4`: `getPrimaryKeyFields` now reuses the pruned descriptor traversal (`collectSchemaFieldPathDescriptors` filtered on `isPrimaryKey`), so a disconnected/hidden PK is dropped and a nested PK keeps its full `parent.child` path instead of just the leaf. Added disconnected / hidden / nested spec cases.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
zapolsky: Documented in `3f3349fa4`: split the `COUNT` (correct) and `COUNT_DISTINCT` (over-counts on a transitive blend) cases in `getReAggregateFunction` with a known-limitation note alongside the existing AVG one, and reworded the spec title/comment so it no longer implies semantic correctness. A correct transitive distinct count needs the raw values at the parent level — same follow-up bucket as the AVG avg-of-avgs case.

apps/web/src/features/data-marts/shared/types/relationship.types.ts
zapolsky: Fixed in `3f3349fa4`: the comment now states `[]` = none allowed, `undefined` (absent) = fall back to type defaults — consistent with `resolveColumnAllowedAggregations` and `cleanBlendedFieldOverride`.

romandubovyi (approved):
Re-reviewed at 4dbd5620. All eight findings from the previous review were addressed or explicitly handled as documented limitations. The blocking Legacy BigQuery and blended Unique Count issues are fixed, the remaining corrections match the author responses, and all CI checks are green. No new blocking issues found.

PR #1374, opened 2026-06-30
feat(web): protect unsaved Output Schema changes from being lost

romandubovyi:
## Summary

When a data mart's Output Schema has unsaved changes, users are now asked how to proceed before any action that would replace the schema — so edits are never discarded without warning.

## What's changed

- Added `useSchemaUnsavedGuard` hook with a Save/Discard/Cancel dialog.
- Guarded actions: AI field metadata generation, schema refresh, publish, input source update, and in-app navigation.
- `SchemaGuardRegistration.isDirty` is now a getter to avoid race conditions.
- Registration is marked clean immediately after save/discard so subsequent navigation isn't blocked.
- Dialog buttons show saving state and are disabled during async save.
- Removed redundant `onClick={onCancel}` to prevent double cancel invocation.
- `DataMartContext` now re-throws schema/definition update errors so the guard can keep the dialog open on failure.
- Added/extended tests for the guard, dialog, and `useSchemaState`.

### DIFF
--- a/.changeset/6085-protect-unsaved-schema-changes.md
+++ b/.changeset/6085-protect-unsaved-schema-changes.md
@@ -1,0 +1,17 @@
+---
+"owox": minor
+---
+
+# Protect unsaved Output Schema changes from being lost
+
+When a data mart's Output Schema has unsaved changes, you're now asked how you'd like to proceed before any action that would replace the schema — so your edits are never discarded without warning.
+
+You'll see a clear prompt with three choices — **Save & continue**, **Discard & continue**, or **Cancel** — when you have unsaved schema changes and you:
+
+- generate field aliases or descriptions with AI,
+- refresh the schema,
+- publish the data mart,
+- update the input source, or
+- leave the page.
+
+This makes it obvious what will happen to your changes and lets you keep your work instead of losing it.
--- a/apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
@@ -56,7 +56,7 @@ export function DataMartDefinitionSettings({
   setDefinitionType,
   sqlRevalidateVersion,
 }: DataMartDefinitionSettingsProps) {
-  const { dataMart, updateDataMartDefinition, runSchemaActualization } =
+  const { dataMart, updateDataMartDefinition, runSchemaActualization, runGuarded } =
     useOutletContext<DataMartContextType>();
   const preset = useDataMartPreset();
 
@@ -177,9 +177,21 @@ export function DataMartDefinitionSettings({
   const handleFormSubmit = useCallback(
     (e?: React.SyntheticEvent<HTMLFormElement>) => {
       e?.preventDefault();
-      void handleSubmit(onSubmit)(e);
+      const submit = () => {
+        void handleSubmit(onSubmit)();
+      };
+      if (runGuarded) {
+        runGuarded(
+          () => {
+            submit();
+          },
+          { intent: 'definition' }
+        );
+      } else {
+        submit();
+      }
     },
-    [handleSubmit, onSubmit]
+    [handleSubmit, onSubmit, runGuarded]
   );
 
   const handleReset = useCallback(() => {
--- a/apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
@@ -174,24 +174,29 @@ export function DataMartDefinitionSettings({
     }
   }, [shouldActualizeSchema, runSchemaActualization]);
 
+  const guardedSubmit = useCallback<SubmitHandler<DataMartDefinitionFormData>>(
+    data => {
+      const runSubmit = () => {
+        void onSubmit(data);
+      };
+      if (runGuarded) {
+        runGuarded(runSubmit, { intent: 'definition' });
+      } else {
+        runSubmit();
+      }
+    },
+    [onSubmit, runGuarded]
+  );
+
   const handleFormSubmit = useCallback(
     (e?: React.SyntheticEvent<HTMLFormElement>) => {
       e?.preventDefault();
-      const submit = () => {
-        void handleSubmit(onSubmit)();
-      };
-      if (runGuarded) {
-        runGuarded(
-          () => {
-            submit();
-          },
-          { intent: 'definition' }
-        );
-      } else {
-        submit();
-      }
+      // Validate the definition form BEFORE involving the schema guard. Only a
+      // valid submission opens the guard dialog, so we never save/discard schema
+      // edits for a definition change that fails validation and never fires.
+      void handleSubmit(guardedSubmit)();
     },
-    [handleSubmit, onSubmit, runGuarded]
+    [handleSubmit, guardedSubmit]
   );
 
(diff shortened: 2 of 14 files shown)


zapolsky:

Reviewed against a fresh worktree checkout with parallel multi-agent passes (line-by-line reviewer, logic/data-flow hunter, un-checklisted fresh-eyes generalist) plus independent verification. **CI-parity: `tsc -b` clean, `eslint` clean, the 3 touched test files pass (15/15).** Changeset present and correct (`owox: minor`). Nothing attributable to this PR fails CI.

The guard state machine is genuinely well-built. Several plausible failure modes were traced and **disproved**: the re-throws in `DataMartContext` (`updateDataMartSchema`/`deleteDataMart`) are handled by every caller (the list-side delete is a separate `useDataMartList` hook, unaffected); collapsing the Output Schema card uses a CSS-grid hide and does **not** unmount the editor, so dirty edits + guard registration survive; `useBlocker` is valid in prod (`createBrowserRouter`); the `skipNextInitialSchemaResetRef` flag is provably consumed by exactly one reset and can't leak; the `beforeunload` listener stays in sync with the dirty flag; "Cancel" fully cancels and never touches the schema; save-failure correctly keeps the dialog open with edits intact.

- **Deleting a data mart with a dirty schema routes the delete through the schema guard, which is contradictory.** `DataMartDetails.tsx:610-623` — the delete `onConfirm` calls `schemaGuard.runGuarded(deleteAction, { intent: 'navigation' })`. With unsaved schema edits this stacks the `SchemaUnsavedChangesDialog` (Radix `AlertDialog`) on top of the still-open delete `ConfirmationDialog` (Radix `Dialog` — its confirm button never closes it, `ConfirmationDialog.tsx:60`), and offers **"Save & leave"**, which persists (`guardSave` → `updateDataMartSchema`, `DataMartSchemaSettings.tsx:115-122`) the schema of a mart that's about to be destroyed. If that save 422s, the now-rethrowing `updateDataMartSchema` (`DataMartContext.tsx:658`) makes `onSaveAndContinue` keep the dialog open and never run the delete (`use-schema-unsaved-guard.ts:112-117`) — the user has to discover that "Discard & leave" is the only way through. Delete also isn't in the PR's stated guard list, and the navigation copy ("Leaving this page will discard them") reads oddly for a delete. *Fix:* don't guard delete (it discards schema edits regardless) — run it directly, or use a discard-only intent with no Save option.

- **Question (unproven): for `refresh` / `publish` / `definition`, "Save & continue" may not actually preserve the edits — and the copy says they'd be lost.** The dialog tells the user their changes "would be lost" (`SchemaUnsavedChangesDialog.tsx:24-27`) yet still offers a Save button. For these three intents the guarded action refreshes/actualizes the schema from source after the save (`definition`: `DataMartDefinitionSettings.tsx:170-175`; `publish`: `DataMartDetails.tsx:182`; `refresh`: `DataMartSchemaSettings.tsx:140-142`), so on the frontend the just-saved edits are reset by the follow-up `dataMart` dispatch. Whether the edits survive in the DB depends on whether backend schema actualization **merges** field metadata (descriptions/aliases) by name or **replaces** the schema — I can't close that from the frontend. *What was checked:* the save → actualize → reset chain across `use-schema-unsaved-guard.ts:113-126` → context → `useSchemaState.ts:49-57`. *Ask:* confirm actualization merge semantics; if it replaces, "Save & continue" is a no-op here and the button should be relabeled/removed; if it merges, the "would be lost" copy is steering users to the wrong choice.

- **The `definition` guard runs `guardSave`/`guardDiscard` before the input-source form is validated.** `DataMartDefinitionSettings.tsx:177-195` — `handleFormSubmit` calls `runGuarded(() => submit(), { intent: 'definition' })` where `submit = () => void handleSubmit(onSubmit)()`, so react-hook-form validation happens *inside* the guarded action, after the schema has already been saved or discarded. On an invalid definition the schema is mutated but `onSubmit` never fires — and on **"Discard & continue"** the user loses their schema edits for a definition save that didn't even happen. The standard Save button is `disabled={!isValid || !isDirty}` (`:217`), but `handleFormSubmit` is also wired straight into the form as `saveDataMartDefinition` (`:213`), so that disable doesn't fully protect it. *Fix:* validate first — only call `runGuarded` from inside `onSubmit` (post-validation), or gate `handleFormSubmit` on `isValid`.

<details>
<summary>Additional observations (below the bar — non-blocking, listed for completeness)</summary>

- `use-schema-unsaved-guard.ts:48-52,94,126,141` — **`runAction` is invoked with no `try/catch`.** Safe today only because every wired action swallows its own errors (AI returns `{kind:'failed'}`, actualize/publish/delete have internal try/catch, the definition action `void`s the form promise). A future guarded action that rejects would surface as an unhandled rejection. Consider wrapping the `runAction` body in try/catch so the guarantee doesn't depend on every caller being defensive.
- `SchemaUnsavedChangesDialog.tsx:43-47` + `use-schema-unsaved-guard.ts:144-150` — **Escape bypasses the `isSaving` disabled state.** The buttons are disabled while saving, but `onOpenChange`/Escape still routes to `onCancel`, which nulls `pendingActionRef`. Pressing Escape during the (sub-second) save window persists the schema but silently drops the follow-up publish/refresh/navigation. Consider gating Escape/`onOpenChange` on `isSaving` too.
- minor: `use-schema-unsaved-guard.ts:55-59` — the navigation blocker keys on `pathname` only, so same-path search/hash navigations aren't guarded (fine for this app's path-based tabs). `DataMartSchemaSettings.tsx:187-201` → `useSchemaState.ts:84-90` — bulk-AI writes results back via `updateSchema`, whose closure captures the click-time (dirty) `schema`; currently harmless since `fields` is fully overwritten and `type` is identical, but it's latent fragility. `use-schema-unsaved-guard.ts:64-72` — a non-navigation dialog can be re-targeted to `navigation` intent if a programmatic navigation/Back fires while it's open (mitigated by the modal blocking user-initiated nav).

</details>

<sub>Automated OWOX review · worktree multi-agent passes · no blockers found (none to verify).</sub>

romandubovyi:
Thanks for the thorough pass. Addressed all three review notes:

**1. Delete routed through the schema guard.** Fixed. Deleting destroys the whole data mart (schema included), so guarding unsaved schema edits there was contradictory — it stacked a second dialog over the delete confirmation and risked blocking the delete on a doomed schema save. `DataMartDetails.tsx` now runs the delete directly (no `runGuarded`) and clears the guard registration (`registerSchemaGuard(null)`) before navigating, so the post-delete navigation isn't intercepted by the dirty-schema blocker.

**2. "Save & continue" semantics for refresh/publish/definition.** Confirmed on the backend: actualization **merges by field name**, it does not replace. `FlatDataMartSchemaMerger` and `BigQuerySchemaMerger` preserve `alias`, `description`, `isPrimaryKey` by name, and fields no longer present in the source are kept and marked `DISCONNECTED` rather than dropped. So "Save & continue" genuinely preserves the user's edits — it's not a no-op. The real issue was the dialog copy implying unconditional loss; I reworded all intents so they reflect that Save keeps the changes and Discard loses them.

**3. `definition` guard ran before form validation.** Fixed. `handleFormSubmit` now validates first via `handleSubmit(guardedSubmit)()`; the guard dialog only opens for a valid submission, so we no longer save/discard schema edits for a definition change that fails validation and never fires.

Also picked up the **Escape-bypasses-`isSaving`** observation: `onOpenChange` is now gated on `!isSaving`, so Escape/outside-click can't drop the pending follow-up action mid-save. Added two dialog tests covering both Escape cases.

Left the remaining below-the-bar items (no `try/catch` around `runAction`, pathname-only blocker, bulk-AI closure) as-is — they're documented as safe today.

CI-parity green locally: `tsc -b` clean, `eslint` clean, full web suite 555/555.

kalnik-a-a (approved):

PR #1375, opened 2026-07-01
feat/mcp-data-mart-details

amarchenk0:


apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
```diff
@@ -0,0 +1,83 @@
...
+  .object({
+    name: z.string(),
+    type: z.string(),
+    description: z.string().optional().nullable(),
+    businessName: z.string().optional(),
+  })
+  .passthrough();
+
+@Injectable()
+export class GetDataMartDetailsTool implements McpToolDefinition<GetDataMartDetailsInput> {
```
romandubovyi, line 29: 🟡 **Important:** The description promises "sample values for categorical fields," but `getDataMartDetails` never returns sample values (the schema field types have no sample-value concept). Since MCP tool descriptions directly steer LLM behavior, this false promise will make the model expect data it won't receive and construct filters blindly. Either implement sample values or drop this sentence from the description (and the assertion in the spec).

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
```diff
@@ -26,4 +42,48 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
...
+  private async ensureDataMartAccessible(request: McpGetDataMartDetailsRequest): Promise<void> {
+    await this.getDataMartService.run(
+      new GetDataMartCommand(request.dataMartId, request.projectId, request.userId, request.roles)
+    );
+  }
+
+  private filterAvailableFields(fields: DataMartSchemaField[]): DataMartSchemaField[] {
+    return fields.filter(isConnected).map(field => {
+      if (!('fields' in field) || !Array.isArray(field.fields)) {
+        return field;
```
romandubovyi: 🔵 **Question:** `filterAvailableFields` drops only DISCONNECTED fields, but keeps `isHiddenForReporting` ones. The sibling helper `collectSchemaFieldPathTypes` skips both. Since this tool exposes the schema to an LLM for query building, is exposing hidden-for-reporting fields intentional here, or should this align with `collectSchemaFieldPathTypes` and also skip `isHiddenForReporting`?

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
```diff
@@ -26,4 +42,48 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
...
+  async getDataMartDetails(
+    request: McpGetDataMartDetailsRequest
+  ): Promise<McpDataMartDetailsResponse> {
+    await this.ensureDataMartAccessible(request);
+
+    const dataMart = await this.dataMartService.actualizeSchemaIfExpired(
+      request.dataMartId,
+      request.projectId,
+      AI_INSIGHTS_SCHEMA_EXPIRES_AFTER_MS
+    );
```
romandubovyi, line 55: 💡 **Suggestion:** The double cast (`as DataMartSchema` in, `as { fields: ... }` out) around `prepareSchema` defeats type checking — a change to `prepareSchema`'s output shape wouldn't be caught here. Consider a shape-preserving generic or a small typed helper to avoid casting on both sides.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
```diff
@@ -26,4 +42,48 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
       })),
     };
   }
+
+  async getDataMartDetails(
+    request: McpGetDataMartDetailsRequest
+  ): Promise<McpDataMartDetailsResponse> {
+    await this.ensureDataMartAccessible(request);
+
+    const dataMart = await this.dataMartService.actualizeSchemaIfExpired(
```
romandubovyi, line 51: 💡 **Suggestion:** This loads the same Data Mart twice per call — once in `ensureDataMartAccessible` (`GetDataMartService.run`, result discarded) and again inside `actualizeSchemaIfExpired` → `getByIdAndProjectId`. The access check is needed (it enforces roles, which the project-scoped fetch does not), so the ordering is right, but if `GetDataMartService` returns the entity you could reuse it and call `actualizeSchemaInEntityIfExpired` to drop the second fetch. Minor — fine to leave if the extra query is not a concern.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
```diff
@@ -26,4 +42,48 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
...
+    };
+  }
+
+  private async ensureDataMartAccessible(request: McpGetDataMartDetailsRequest): Promise<void> {
+    await this.getDataMartService.run(
+      new GetDataMartCommand(request.dataMartId, request.projectId, request.userId, request.roles)
+    );
+  }
+
+  private filterAvailableFields(fields: DataMartSchemaField[]): DataMartSchemaField[] {
```
romandubovyi, line 77: ⚪ **Nit:** `filterAvailableFields` recurses into `field.fields` (nested/RECORD-style schemas), but no spec exercises that branch — every mock schema in the facade test is flat. The schema-less path (`dataMart.schema === undefined` → `fields: []`) is also uncovered. A couple of small cases (one nested field where a `DISCONNECTED` child is pruned, one data mart with no schema) would guard this recursion against silent regressions.

romandubovyi:
**Overall: no blockers.** 👍

Clean, well-scoped change and access control is handled correctly. Left a few non-blocking inline notes (tool description accuracy, hidden-field filtering, a double fetch, and a couple of test gaps) — all optional. Nice work.

romandubovyi (approved):

romandubovyi:
💡Suggestion (docs): The user-facing MCP docs — docs/getting-started/setup-guide/mcp.md (mirrored in apps/docs/src/content/docs/docs/getting-started/setup-guide/mcp.md) — have an "Available tools" section that still says the server "exposes two read-only tools" and documents only get_project_context and list_data_marts. This PR adds get_data_mart_details_by_id but does not update that page.

PR #1376, opened 2026-07-01
feat(mcp): add list_destinations MCP tool

laskevych:
## Summary

Adds a `list_destinations` MCP tool (PRD §15.15) so an AI assistant connected to OWOX over MCP can see the destinations in the active project and pick a target before creating a report. This is the first of the report/destination MCP tools; today the MCP surface only had read tools for data marts.

Read-only, requires the `mcp:read` scope, and is **free** (no credits consumed).

## What changed

- **New MCP tool** `ListDestinationsTool` (`ee/mcp/tools/list-destinations.tool.ts`) — no input params; output `{ destinations: [{ id, name, type, owner }] }`; `readOnlyHint: true`.
- **New facade** `McpDataDestinationsFacade` (+ impl, + DI symbol) wrapping the existing `ListDataDestinationsService` — registered and exported from `DataMartsModule`. No business logic reimplemented.
- **Registration** in `MCP_TOOL_PROVIDER_CLASSES`; updated the provider-list assertion in `data-mart-catalog.tool.spec.ts`.

## Design notes

- **`type`** is mapped to the lowercase MCP vocabulary via an exhaustive `Record<DataDestinationType, …>` (`MS_TEAMS → teams`, `GOOGLE_CHAT → google_chat`), so any future destination type fails the build until it's mapped.
- **`owner`** is derived from the destination creator's email.
- Destinations with `availableForUse === false` are omitted (the tool exists to pick a target for `add_report`).
- Row-level visibility is delegated to the existing permission model — the user only sees destinations they can access.

## Testing

- Unit tests for the facade (command args, type mapping for all 6 types, owner fallback, `availableForUse` filter) and the tool (output shape, `.strict()` input rejection, scope/annotations, registry).
- `nest build` (type-check, incl. the exhaustive `Record`), eslint, and the full MCP module suite — all green.
- Verified live in **Claude Desktop** (via `mcp-remote`) against a local `IDP_PROVIDER=owox-better-auth` server: the tool appears and returns the project's destinations.

## Out of scope / follow-ups

Sibling MCP report tools (`get_data_mart_reports`, `add_report`, `update_report`, `delete_report`) are tracked separately.

### DIFF
--- a/apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
+++ b/apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
@@ -36,9 +36,10 @@ export class McpDataDestinationsFacadeImpl implements McpDataDestinationsFacade 
           id: destination.id,
           name: destination.title,
           type: DESTINATION_TYPE_MAP[destination.type],
-          // Prefer an assigned owner; fall back to the creator when a
-          // destination has no explicit owner.
-          owner: (destination.ownerUsers[0] ?? destination.createdByUser)?.email ?? null,
+          // Report the creator as the owner. The `owners` relation has no stable
+          // ordering, so picking one of potentially several owners would be
+          // non-deterministic; the creator is a single, stable value.
+          owner: destination.createdByUser?.email ?? null,
         })),
     };
   }
--- a/apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
+++ b/apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
@@ -1,0 +1,45 @@
+import { Injectable } from '@nestjs/common';
+import { DataDestinationType } from '../data-destination-types/enums/data-destination-type.enum';
+import { ListDataDestinationsCommand } from '../dto/domain/list-data-destinations.command';
+import { ListDataDestinationsService } from '../use-cases/list-data-destinations.service';
+import {
+  McpDataDestinationsFacade,
+  McpDestinationType,
+  McpListDestinationsRequest,
+  McpListDestinationsResponse,
+} from './mcp-data-destinations.facade';
+
+const DESTINATION_TYPE_MAP: Record<DataDestinationType, McpDestinationType> = {
+  [DataDestinationType.GOOGLE_SHEETS]: 'google_sheets',
+  [DataDestinationType.LOOKER_STUDIO]: 'looker_studio',
+  [DataDestinationType.EMAIL]: 'email',
+  [DataDestinationType.SLACK]: 'slack',
+  [DataDestinationType.MS_TEAMS]: 'teams',
+  [DataDestinationType.GOOGLE_CHAT]: 'google_chat',
+};
+
+@Injectable()
+export class McpDataDestinationsFacadeImpl implements McpDataDestinationsFacade {
+  constructor(private readonly listDataDestinationsService: ListDataDestinationsService) {}
+
+  async listDestinations(
+    request: McpListDestinationsRequest
+  ): Promise<McpListDestinationsResponse> {
+    const items = await this.listDataDestinationsService.run(
+      new ListDataDestinationsCommand(request.projectId, request.userId, request.roles)
+    );
+
+    return {
+      destinations: items
+        .filter(destination => destination.availableForUse)
+        .map(destination => ({
+          id: destination.id,
+          name: destination.title,
+          type: DESTINATION_TYPE_MAP[destination.type],
+          // Prefer an assigned owner; fall back to the creator when a
+          // destination has no explicit owner.
+          owner: (destination.ownerUsers[0] ?? destination.createdByUser)?.email ?? null,
+        })),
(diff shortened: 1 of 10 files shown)


apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
```diff
@@ -0,0 +1,45 @@
...
+    return {
+      destinations: items
+        .filter(destination => destination.availableForUse)
+        .map(destination => ({
+          id: destination.id,
+          name: destination.title,
+          type: DESTINATION_TYPE_MAP[destination.type],
+          // Prefer an assigned owner; fall back to the creator when a
+          // destination has no explicit owner.
+          owner: (destination.ownerUsers[0] ?? destination.createdByUser)?.email ?? null,
```
romandubovyi: 🔵 **Question:** Destinations can have multiple owners, while the `owners` relation has no explicit ordering, so `ownerUsers[0]` may select an arbitrary user. The PR description also says this field is derived from the creator. Should the contract return the creator consistently, expose all owners, or define a deterministic primary-owner rule?

romandubovyi: 🟡 Important: ownerUsers[0] reads from the owners @OneToMany relation, which ListDataDestinationsService loads via leftJoinAndSelect('d.owners', 'owners') with no ORDER BY. For a destination with multiple owners, Postgres doesn't guarantee row order, so [0] can surface a different owner across calls, and the singular owner field silently drops the rest. Consider making the pick deterministic (e.g. a stable sort, or explicitly preferring the creator) — or  returning all owners if that's more useful for choosing a target.

apps/backend/src/ee/mcp/tools/list-destinations.tool.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { McpToolDefinition, McpToolResult } from './mcp-tool.definition';
+
+type ListDestinationsInput = Record<string, never>;
+
+@Injectable()
+export class ListDestinationsTool implements McpToolDefinition<ListDestinationsInput> {
+  readonly name = 'list_destinations';
+  readonly description =
+    'List destinations in the active OWOX project to target when creating a report.';
+  readonly zodSchema = {};
```
romandubovyi: ⚪ **Nit:** `zodSchema = {}` and `inputSchema = z.object({}).strict()` are two independent declarations of the same empty schema. Sibling tools (e.g. `SearchDataMartsTool`) keep a single schema and expose `zodSchema = inputSchema.shape`. Harmless for empty input, but deriving `zodSchema` from the one strict schema would match the existing convention. Fully optional.

romandubovyi:
💡 **Suggestion (docs):** The user-facing MCP docs — `docs/getting-started/setup-guide/mcp.md` (mirrored in `apps/docs/src/content/docs/docs/getting-started/setup-guide/mcp.md`) — have an **"Available tools"** section that still says the server *"exposes two read-only tools"* and documents only `get_project_context` and `list_data_marts`. This PR adds `list_destinations` but does not update that page. Since the changeset already ships user-facing copy for the new tool, it would be consistent to add a `list_destinations` subsection there and correct the "two" count.

Heads-up: that page is already behind — it also omits the existing `get_relevant_data_marts_by_prompt` (search) tool — so this is **non-blocking**. Flagging it in case the docs are meant to stay in sync with the registered tool set.

romandubovyi (approved):

apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
laskevych: Good catch on the ordering. `ownerUsers[0]` is indeed non-deterministic since the `owners` relation is loaded without an `ORDER BY`. Switched to reporting the destination **creator** — a single, stable value — so the singular `owner` field is deterministic. Fixed in f20ee9d6e.

laskevych: Went with returning the creator consistently: it keeps the singular `owner` contract, is deterministic (single stable value vs. the unordered owners relation — see the sibling thread), and matches the PR description. Fixed in f20ee9d6e.

apps/backend/src/ee/mcp/tools/list-destinations.tool.ts
laskevych: Applied — the tool now uses a single strict `inputSchema` with `zodSchema = inputSchema.shape`, matching `SearchDataMartsTool`. Fixed in f20ee9d6e.

laskevych:
Thanks — updated the MCP setup guide in `3f8665fb5`. Added a `list_destinations` subsection, and since the page was behind I also documented the existing `get_relevant_data_marts_by_prompt` tool and corrected the count from two to four read-only tools. Note: the `apps/docs/...` copy is generated from `docs/…/mcp.md` via `npm run sync` (gitignored), so only the source file needed editing.

PR #1377, opened 2026-07-01
feat(mcp): add get_data_mart_reports MCP tool

laskevych:
## Summary

Adds a `get_data_mart_reports` MCP tool (PRD §15.8) that lists the reports tied to a data mart in the active project, so an AI assistant can review a data mart's reports — their destination, owner, run schedules, last run, and its status — without leaving the chat. Read-only, requires the `mcp:read` scope, and is **free** (no credits).

Second of the report/destination MCP tools, after `list_destinations` (#1376).

## What changed

- **New MCP tool** `GetDataMartReportsTool` (`ee/mcp/tools/get-data-mart-reports.tool.ts`) — input `{ data_mart_id }`; output `{ reports: [{ report_id, name, destination_id, destination_type, owner, schedules, last_run_at, last_run_status }] }`; `readOnlyHint: true`.
- **New facade** `McpReportsFacade` (+ impl, + DI symbol), wrapping the existing `ListReportsByDataMartService`. Registered and exported from `DataMartsModule`. No business logic reimplemented.
- **Shared destination-type module** `facades/mcp-destination-type.ts` — extracts the `DataDestinationType → MCP vocabulary` mapping that `list_destinations` introduced, so both tools map from one source (no drift). `list_destinations` is refactored to use it (behavior unchanged).

## Design notes

- **A report's schedules are exposed as-is, all of them.** A report can have any number of `REPORT_RUN` triggers and the product does not enforce a single-schedule rule, so the tool does not pretend there is one: each report carries a `schedules` array with one item per trigger — `{ trigger_id, cron_expression, time_zone, is_active, next_run_at, last_run_at }` — ordered by trigger creation time (deterministic rather than dependent on DB row order), empty when the report is unscheduled. The field vocabulary matches the report-run-schedule tools from #1378, so `trigger_id` values can be passed to `update_report_run_schedule` / `delete_report_run_schedule` directly.
- **`last_run_status`** is the report's own `lastRunStatus` (`SUCCESS` / `ERROR` / `RUNNING` / `CANCELLED` / `RESTRICTED`), passed through verbatim rather than folded into a derived status. `null` when the report never ran.
- The facade loads the data mart's triggers once (`ScheduledTriggerService.getAllByDataMartIdAndProjectId`) and groups `REPORT_RUN` triggers by `triggerConfig.reportId`.
- **`owner`** is the report creator's email — consistent with `list_destinations` (the owners relation has no stable ordering).
- Row-level access is delegated to `ListReportsByDataMartService` (checks `DATA_MART.SEE`; returns an empty list for a data mart you cannot see).

## Testing

- Facade unit tests: field mapping (including per-schedule fields); multi-trigger reports return every schedule ordered by creation time, disabled ones included; unscheduled reports get an empty `schedules` array; connector and other-report triggers are ignored; `last_run_status` pass-through (`SUCCESS` / `ERROR` / never ran → `null`); empty-reports short-circuit (no trigger query); and `owner` null.
- Tool unit tests: output shape, input validation (`data_mart_id` required, `.strict()` rejects extras), scope/annotations, registry.
- `nest build` (type-check), eslint, prettier, and the full MCP + facades suites — all green (52 tests).

## Out of scope / follow-ups

Sibling report tools (`add_report`, `update_report`, `delete_report`) are tracked separately. Schedule management itself (create/update/delete of report run schedules) is #1378.

### DIFF
--- a/apps/backend/src/data-marts/facades/mcp-destination-type.ts
+++ b/apps/backend/src/data-marts/facades/mcp-destination-type.ts
@@ -1,0 +1,36 @@
+import { DataDestinationType } from '../data-destination-types/enums/data-destination-type.enum';
+
+/**
+ * Canonical MCP destination-type vocabulary and single source of truth. The
+ * `McpDestinationType` union, the `DESTINATION_TYPE_MAP` target type, and every
+ * tool's output `z.enum` are derived from this tuple, so they cannot drift out
+ * of sync (a rename here updates every consumer at once).
+ */
+export const MCP_DESTINATION_TYPES = [
+  'google_sheets',
+  'looker_studio',
+  'email',
+  'slack',
+  'teams',
+  'google_chat',
+] as const;
+
+export type McpDestinationType = (typeof MCP_DESTINATION_TYPES)[number];
+
+const DESTINATION_TYPE_MAP: Record<DataDestinationType, McpDestinationType> = {
+  [DataDestinationType.GOOGLE_SHEETS]: 'google_sheets',
+  [DataDestinationType.LOOKER_STUDIO]: 'looker_studio',
+  [DataDestinationType.EMAIL]: 'email',
+  [DataDestinationType.SLACK]: 'slack',
+  [DataDestinationType.MS_TEAMS]: 'teams',
+  [DataDestinationType.GOOGLE_CHAT]: 'google_chat',
+};
+
+/**
+ * Maps a domain `DataDestinationType` to the lowercase MCP vocabulary exposed to
+ * clients. The `Record` is exhaustive over the enum, so a new destination type
+ * fails the build until it is mapped here.
+ */
+export function toMcpDestinationType(type: DataDestinationType): McpDestinationType {
+  return DESTINATION_TYPE_MAP[type];
+}
--- a/apps/backend/src/data-marts/facades/mcp-destination-type.ts
+++ b/apps/backend/src/data-marts/facades/mcp-destination-type.ts
@@ -29,8 +29,14 @@ const DESTINATION_TYPE_MAP: Record<DataDestinationType, McpDestinationType> = {
 /**
  * Maps a domain `DataDestinationType` to the lowercase MCP vocabulary exposed to
  * clients. The `Record` is exhaustive over the enum, so a new destination type
- * fails the build until it is mapped here.
+ * fails the build until it is mapped here; the runtime guard covers values that
+ * bypass the compiler (e.g. a raw DB string outside the enum), which would
+ * otherwise leak `undefined` into an MCP response.
  */
 export function toMcpDestinationType(type: DataDestinationType): McpDestinationType {
-  return DESTINATION_TYPE_MAP[type];
+  const mapped = DESTINATION_TYPE_MAP[type];
+  if (!mapped) {
+    throw new Error(`Unsupported destination type for MCP: ${type}`);
+  }
+  return mapped;
 }
--- a/apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
+++ b/apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
@@ -1,0 +1,117 @@
+import { Injectable } from '@nestjs/common';
+import { DataMartScheduledTrigger } from '../entities/data-mart-scheduled-trigger.entity';
+import { ReportRunStatus } from '../enums/report-run-status.enum';
+import { ScheduledTriggerType } from '../scheduled-trigger-types/enums/scheduled-trigger-type.enum';
+import { ListReportsByDataMartCommand } from '../dto/domain/list-reports-by-data-mart.command';
+import { ReportDto } from '../dto/domain/report.dto';
+import { ListReportsByDataMartService } from '../use-cases/list-reports-by-data-mart.service';
+import { ScheduledTriggerService } from '../services/scheduled-trigger.service';
+import { toMcpDestinationType } from './mcp-destination-type';
+import {
+  McpGetDataMartReportsRequest,
+  McpGetDataMartReportsResponse,
+  McpReportStatus,
+  McpReportsFacade,
+} from './mcp-reports.facade';
+
+@Injectable()
+export class McpReportsFacadeImpl implements McpReportsFacade {
+  constructor(
+    private readonly listReportsByDataMartService: ListReportsByDataMartService,
+    private readonly scheduledTriggerService: ScheduledTriggerService
+  ) {}
+
+  async getDataMartReports(
+    request: McpGetDataMartReportsRequest
+  ): Promise<McpGetDataMartReportsResponse> {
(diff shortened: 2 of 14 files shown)


romandubovyi (commented):
Reviewing PR — inline comments will follow point by point.

apps/backend/src/data-marts/facades/mcp-destination-type.ts
```diff
@@ -0,0 +1,36 @@
...
+  [DataDestinationType.MS_TEAMS]: 'teams',
+  [DataDestinationType.GOOGLE_CHAT]: 'google_chat',
+};
+
+/**
+ * Maps a domain `DataDestinationType` to the lowercase MCP vocabulary exposed to
+ * clients. The `Record` is exhaustive over the enum, so a new destination type
+ * fails the build until it is mapped here.
+ */
+export function toMcpDestinationType(type: DataDestinationType): McpDestinationType {
```
romandubovyi, line 36: 🟡 **Important:** If a new `DataDestinationType` is added to the enum but not mapped in `DESTINATION_TYPE_MAP`, the function returns `undefined` at runtime despite the TypeScript type claiming `McpDestinationType`. Add a runtime throw or fallback so that unmapped types fail loudly rather than leaking `undefined` into the MCP response.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (a.isActive !== b.isActive) {
+      return a.isActive ? a : b;
+    }
+    return a.createdAt <= b.createdAt ? a : b;
+  }
+
+  private deriveStatus(
+    report: ReportDto,
+    trigger: DataMartScheduledTrigger | undefined
+  ): McpReportStatus {
```
romandubovyi: 🔵 **Question:** A manual run that failed returns `error` even when the report has no schedule at all. Is that intentional? For an unscheduled report it might be surprising to see `error` as the status; some consumers might expect `paused` (or `error` only when a scheduled run fails). Could you clarify the intended behavior for unscheduled reports?

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -0,0 +1,106 @@
...
+        byReportId.set(reportId, [trigger]);
+      }
+    }
+    for (const reportTriggers of byReportId.values()) {
+      reportTriggers.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime());
+    }
+    return byReportId;
+  }
+
+  private toScheduleItem(trigger: DataMartScheduledTrigger): McpReportScheduleItem {
```
romandubovyi, line 105: 💡 **Suggestion:** The tie-breaker on `createdAt` is correct but non-obvious — a one-line comment explaining *why* the earliest trigger wins would save future readers from second-guessing whether `createdAt` refers to creation or update, or whether a newer trigger might be a deliberate override.

apps/backend/src/data-marts/facades/mcp-reports.facade.ts
```diff
@@ -0,0 +1,49 @@
...
+  /** ISO 8601 timestamp of the next scheduled run, or `null`. */
+  next_run_at: string | null;
+  /** ISO 8601 timestamp of the trigger's last run, or `null` when it never ran. */
+  last_run_at: string | null;
+}
+
+export interface McpReportListItem {
+  report_id: string;
+  name: string;
+  destination_id: string;
```
romandubovyi, line 34: 💡 **Suggestion:** Consider including the `data_mart_id` in the response so downstream consumers can correlate reports back to their parent data mart without having to preserve the input separately. A small metadata field like `data_mart_id` or `data_mart_id` in the output would improve traceability.

romandubovyi (commented):
Review completed — 4 inline comments published, 2 skipped. Great work overall: clean architecture, good test coverage, and the shared `mcp-destination-type` module is exactly the right refactor.

apps/backend/src/data-marts/facades/mcp-destination-type.ts
laskevych: Agreed — fixed in a4a2335a4. `toMcpDestinationType` now throws `Unsupported destination type for MCP: <type>` instead of returning `undefined`: the `Record` still gives compile-time exhaustiveness for enum growth, and the runtime guard covers values that bypass the compiler (e.g. a raw DB string outside the enum). Added a dedicated spec (`mcp-destination-type.spec.ts`) covering the full mapping and the throw.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Good question — it applied to the earlier revision, and the schedules-array rework resolved it by removing the derived status entirely. The response now returns `last_run_status` verbatim (explicitly a *run* status, `null` when the report never ran) plus the full `schedules` array with per-schedule `is_active`. So an unscheduled report with a failed manual run reads unambiguously as `schedules: [], last_run_status: "ERROR"` — no conflation with a report-level state. No code change needed on top of the rework.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Done in a4a2335a4 — the comment now spells out that oldest-first is purely a stable chronological presentation with no precedence semantics (a newer trigger is an additional schedule, not an override), that `createdAt` is creation time and never changes on update, and why the id tie-break exists (equal timestamps from batch creation).

apps/backend/src/data-marts/facades/mcp-reports.facade.ts
laskevych: Done in a4a2335a4 — each report item now echoes `data_mart_id`, so items are self-describing for downstream consumers. Facade contract, tool output schema, and tests updated.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -0,0 +1,115 @@
...
+    };
+  }
+
+  /**
+   * Loads the data mart's schedule triggers once and groups the REPORT_RUN ones
+   * by their target report. A report can have any number of schedules; they are
+   * all returned, ordered by creation time so the output is deterministic
+   * rather than dependent on DB row order.
+   */
+  private async loadReportTriggers(
```
romandubovyi, line 66: 💡 **Suggestion:** You could narrow the trigger load to reports that actually exist in the response. Build a `Set` of `report.id`s first and skip `REPORT_RUN` triggers whose `reportId` is not in that set. This avoids loading/processing schedules for reports the caller cannot see, especially useful when a data mart has many connector-run triggers.

romandubovyi (approved):
LGTM. All previous comments have been addressed, no blockers remain.

PR #1378, opened 2026-07-01
feat(mcp): add report run schedule tools

kalnik-a-a:
## Summary

Adds MCP support for managing report run schedules in the current OWOX project. The branch introduces four report-run schedule tools and keeps schedule creation additive: MCP can create another schedule for the same report instead of replacing existing ones or enforcing a product-level uniqueness rule.

### `list_report_run_schedules`

Input:

- No fields. The tool uses the authenticated MCP context: current project, user, and roles.

Output:

- `schedules`: array of visible report run schedules.
- Each schedule includes:
  - `trigger_id`
  - `report`: `{ id, title }`
  - `data_mart`: `{ id, title }`
  - `cron_expression`
  - `time_zone`
  - `is_active`
  - `next_run_at`
  - `last_run_at`
  - `can_edit`
  - `can_delete`

Logic:

- Reads `REPORT_RUN` scheduled triggers for the current project.
- Applies the same user/role visibility rules as the scheduled-trigger use case.
- Paginates internally with a 500-item page size until all visible report run schedules are collected.
- Enriches each response item with the report reference, data mart reference, trigger state, timestamps, and permission flags.
- Returns `trigger_id` values that clients pass to `update_report_run_schedule` or `delete_report_run_schedule`.

### `create_report_run_schedule`

Input:

- `report_id`: required report id.
- `cron_expression`: required standard 5-field cron expression.
- `time_zone`: optional IANA timezone, defaults to `UTC`.
- `is_active`: optional boolean, defaults to `true`.

Output:

- `trigger_id`
- `report_id`
- `cron_expression`
- `time_zone`
- `is_active`
- `next_run_at`

Logic:

- Validates the cron expression and timezone through `CronTime`.
- Resolves the target report inside the current project and uses its data mart for the scheduled trigger.
- Creates a new `REPORT_RUN` scheduled trigger with `{ type: "scheduled-report-run", reportId }` config.
- Does not update, delete, or replace any existing schedule for the same report.

### `update_report_run_schedule`

Input:

- `trigger_id`: required scheduled-trigger id returned by `list_report_run_schedules`.
- `cron_expression`: required standard 5-field cron expression.
- `time_zone`: optional IANA timezone, defaults to `UTC`.
- `is_active`: optional boolean, defaults to `true`.

Output:

- `trigger_id`
- `report_id`
- `cron_expression`
- `time_zone`
- `is_active`
- `next_run_at`

Logic:

- Validates the cron expression and timezone through `CronTime`.
- Loads the scheduled trigger by `trigger_id` within the current project.
- Rejects non-`REPORT_RUN` triggers.
- Requires the existing report-run trigger config to contain a report id.
- Updates only the existing trigger's schedule fields: cron expression, timezone, and active state.
- Does not change the report target and does not create another schedule.

### `delete_report_run_schedule`

Input:

- `trigger_id`: required scheduled-trigger id returned by `list_report_run_schedules`.

Output:

- `trigger_id`
- `report_id`
- `schedule: null`

Logic:

- Loads the scheduled trigger by `trigger_id` within the current project.
- Rejects non-`REPORT_RUN` triggers.
- Deletes only that one scheduled trigger through the existing delete use case.
- Leaves the report and any other schedules for that report intact.
- Echoes the deleted `trigger_id` and `report_id`, with `schedule: null`, so the MCP client can confirm which schedule was removed.

## Unified MCP tool response formatting

Adds `jsonToolResult(structuredContent)` in `mcp-tool.definition.ts`.

The helper standardizes JSON-like MCP tool responses by returning the same payload in both:

- `structuredContent`, for clients that read structured MCP output.
- `content[0].text`, formatted as pretty-printed JSON for clients that display textual tool output.

The new report schedule tools use this helper, and the existing MCP tools touched by this branch were moved to the same formatting path. During the rebase onto current `main`, the newly added `list_destinations` tool from `main` was also aligned with this helper.

## Why create and update are separate

The initial MCP-facing design looked like an upsert: find the current user's schedule for a report, then update it or create one. That would have pushed an artificial product rule into MCP: one user can have only one schedule for one report.

That is not a product invariant. A user may have multiple triggers for the same report, and this branch does not add any database uniqueness constraint for that pair.

The final API is explicit:

- `create_report_run_schedule` always creates another schedule.
- `update_report_run_schedule` requires a concrete `trigger_id` and updates exactly that schedule.

This keeps MCP behavior aligned with the product model, avoids introducing new limits on the number of triggers a user can have for a report, and avoids race-prone "find then replace" behavior where concurrent calls could accidentally create duplicates, update the wrong schedule, or delete/replace a schedule the user did not mean to touch.

## Verification

- `npm test -w @owox/backend -- src/data-marts/facades/mcp-scheduled-triggers.facade.impl.spec.ts src/data-marts/facades/mcp-data-destinations.facade.impl.spec.ts src/data-marts/services/scheduled-trigger.service.spec.ts src/data-marts/services/report.service.spec.ts src/data-marts/use-cases/list-project-scheduled-triggers.service.spec.ts src/ee/mcp/tools/data-mart-catalog.tool.spec.ts src/ee/mcp/tools/list-destinations.tool.spec.ts src/ee/mcp/tools/create-report-run-schedule.tool.spec.ts src/ee/mcp/tools/update-report-run-schedule.tool.spec.ts src/ee/mcp/tools/list-report-run-schedules.tool.spec.ts src/ee/mcp/tools/delete-report-run-schedule.tool.spec.ts --runInBand`
- `npx prettier --check` on the changed MCP/DataMarts surface
- `npx eslint` on the changed MCP/DataMarts surface
- `git diff --check HEAD`


apps/backend/src/data-marts/facades/mcp-scheduled-triggers.facade.impl.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  private reportRefOf(config: DataMartScheduledTrigger['triggerConfig']): {
+    id: string;
+    title: string;
+  } {
+    if (!isScheduledReportRunConfig(config)) {
+      return { id: '', title: '' };
+    }
+    const enrichedTitle = (config as { report?: { title?: string } }).report?.title;
+    return { id: config.reportId, title: enrichedTitle ?? '' };
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in 373a99e73. Invalid report-run configs are no longer treated as valid report refs, so list output falls back to { id: "", title: "" }. Added regression coverage.

apps/backend/src/data-marts/facades/mcp-scheduled-triggers.facade.impl.ts
```diff
@@ -0,0 +1,212 @@
...
+    if (!isScheduledReportRunConfig(trigger.triggerConfig)) {
+      throw new BadRequestException('Report ID is required for REPORT_RUN schedules');
+    }
+
+    return trigger.triggerConfig.reportId;
+  }
+
+  private reportIdOrNull(config: DataMartScheduledTrigger['triggerConfig']): string | null {
+    return isScheduledReportRunConfig(config) ? config.reportId : null;
+  }
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in 373a99e73. The schema-backed guard now makes reportIdOrNull return null for invalid report-run config instead of leaking undefined. Added regression coverage.

apps/backend/src/data-marts/facades/mcp-scheduled-triggers.facade.impl.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  private assertReportRunWithConfig(trigger: DataMartScheduledTrigger): string {
+    this.assertReportRun(trigger);
+
+    if (!isScheduledReportRunConfig(trigger.triggerConfig)) {
+      throw new BadRequestException('Report ID is required for REPORT_RUN schedules');
+    }
+
+    return trigger.triggerConfig.reportId;
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in 373a99e73. isScheduledReportRunConfig now validates against ScheduledReportRunConfigSchema, so assertReportRunWithConfig rejects missing or empty reportId before update. Added regression coverage for this path.

apps/backend/src/ee/mcp/tools/delete-report-run-schedule.tool.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+import { DeleteReportRunScheduleTool } from './delete-report-run-schedule.tool';
+
+describe('DeleteReportRunScheduleTool', () => {
+  const context: McpAuthContext = {
+    clientId: 'mcp-client-1',
+    userId: 'user-1',
+    projectId: 'project-1',
+    roles: ['editor'],
+    resource: 'https://mcp.owox.com/mcp',
+    scopes: ['mcp:write'],
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in 373a99e73. The delete tool test context now includes both required scopes and asserts it stays aligned with requiredScopes.

apps/backend/src/ee/mcp/tools/update-report-run-schedule.tool.ts
```diff
@@ -0,0 +1,78 @@
...
+    context: McpAuthContext
+  ): Promise<McpToolResult> {
+    const parsed = this.parseInput(input);
+    const ctx = { projectId: context.projectId, userId: context.userId, roles: context.roles };
+
+    const result = await this.facade.updateReportRunSchedule(ctx, {
+      triggerId: parsed.trigger_id,
+      cronExpression: parsed.cron_expression,
+      timeZone: parsed.time_zone ?? DEFAULT_TIME_ZONE,
+      isActive: parsed.is_active ?? true,
```
laskevych: 🟡 **Important:** `update_report_run_schedule` clobbers `time_zone` and `is_active` when they're omitted.

An update is a full replace of cron + timezone + active state, but the handler substitutes defaults for omitted optional fields instead of preserving the trigger's current values:

```ts
timeZone: parsed.time_zone ?? DEFAULT_TIME_ZONE, // → 'UTC'
isActive: parsed.is_active ?? true,
```

Concrete failure: a schedule is paused (`is_active=false`) in `Europe/Kyiv`. The user says *"change it to run at 10am"*, so the model calls `update` with only `trigger_id` + `cron_expression`. The result: the schedule silently moves to `UTC` (runs shift by hours) **and** the paused schedule is re-activated — neither of which the user asked for. The tool description even tells the model to default an omitted timezone to UTC, so this is the expected path, not an edge case.

Suggested fix: load the existing trigger and only overwrite the fields the caller actually supplied (fall back to the trigger's current `timeZone`/`isActive`, not to `UTC`/`true`). Alternatively, make `time_zone` and `is_active` required so the intent is always explicit.

apps/backend/src/data-marts/facades/mcp-scheduled-triggers.facade.impl.ts
```diff
@@ -0,0 +1,212 @@
...
+  async deleteReportRunSchedule(
+    ctx: McpScheduledTriggersContext,
+    input: { triggerId: string }
+  ): Promise<{ triggerId: string; reportId: string | null }> {
+    const trigger = await this.scheduledTriggerService.getByIdAndProjectId(
+      input.triggerId,
+      ctx.projectId
+    );
+
+    this.assertReportRun(trigger);
```
laskevych, line 152: 🔵 **Question:** Is the type-assertion-before-access-check ordering intentional here?

`deleteReportRunSchedule` (and `updateReportRunSchedule`) load *any* trigger in the project via `getByIdAndProjectId` and run `assertReportRun` before the report-access check happens in the use case. The mutation itself is still correctly blocked by `ReportAccessService.checkOperateAccess`, but the ordering means a project member whose role scope excludes some data marts can probe guessed `trigger_id`s and distinguish `NotFound` (doesn't exist) vs `BadRequest`/`Forbidden` (exists + is/isn't `REPORT_RUN`) — an existence/type oracle across data marts they can't see.

Low severity (information disclosure only, no unauthorized write), and it mirrors the existing report tools, so this may be acceptable — just flagging in case the access check should run before revealing the trigger's existence/type.

apps/backend/src/data-marts/facades/mcp-scheduled-triggers.facade.impl.ts
```diff
@@ -0,0 +1,212 @@
...
+  constructor(
+    private readonly listProjectScheduledTriggersService: ListProjectScheduledTriggersService,
+    private readonly createScheduledTriggerService: CreateScheduledTriggerService,
+    private readonly deleteScheduledTriggerService: DeleteScheduledTriggerService,
+    private readonly updateScheduledTriggerService: UpdateScheduledTriggerService,
+    private readonly reportService: ReportService,
+    private readonly scheduledTriggerService: ScheduledTriggerService
+  ) {}
+
+  async listReportRunSchedules(
```
laskevych, line 45: 💡 **Suggestion:** `listReportRunSchedules` does heavyweight per-page enrichment for data it doesn't use.

It reuses `ListProjectScheduledTriggersService`, which fully hydrates every trigger's report into a complete `ReportResponseApiDto` (awaited dataMart + destination mapping, secret handling) for each 500-item page — but the MCP list only reads `report.id` and `report.title` (plus cron/tz/flags). On a project with thousands of schedules this repeats the full report-response pipeline for every page. Consider a lighter list path that fetches only `id`/`title` for the report reference.

laskevych (commented):
Automated review of the report-run schedule tools. One issue worth fixing before merge (the `update` clobber), plus one access-ordering question and one efficiency suggestion.

apps/backend/src/ee/mcp/tools/update-report-run-schedule.tool.ts
kalnik-a-a: Fixed in 2581e59c0. The tool no longer injects UTC/true for omitted optional fields; it sends only the fields the caller supplied, and the facade loads the existing trigger and preserves the current timeZone/isActive values. I also updated the tool description and added regression coverage for the omitted-fields path.

apps/backend/src/data-marts/facades/mcp-scheduled-triggers.facade.impl.ts
kalnik-a-a: Fixed in 2581e59c0 by normalizing MCP update/delete failures for this path. Non-REPORT_RUN trigger ids now return NotFound instead of exposing type via BadRequest, and Forbidden from the underlying update/delete use cases is translated to NotFound as well. The actual mutation still goes through the shared use case access checks.

apps/backend/src/data-marts/facades/mcp-scheduled-triggers.facade.impl.ts
kalnik-a-a: Acknowledged. I kept listReportRunSchedules on the shared ListProjectScheduledTriggersService for now, so we continue reusing the existing visibility/access/capability/enrichment semantics instead of adding a parallel optimized list path. The performance concern is valid, but I would avoid deeper tuning here until we see this become an actual bottleneck in schedule-list volume or latency.

laskevych (approved):

PR #1379, opened 2026-07-01
fix(connectors): retry transient errors and clarify import failure logs

prokopetsv:
# Problem

The Criteo Ads connector failed a data mart run on a transient `HTTP 500` from Criteo's `/statistics/v1/report` endpoint, even though `MaxFetchRetries`/`InitialRetryDelay` were configured. The root cause: `CriteoAdsSource` never overrode `isValidToRetry()`, so it inherited `AbstractSource`'s default, which unconditionally returned `false` — meaning no source that relied on this default ever retried anything. On top of that, the only thing logged on failure was a raw stack trace, with no indication of the HTTP status, the provider's own error message, or whether the failure was even worth retrying manually.

# Solution

- Hoisted the "retry on 5xx / 429 / network-level errors" policy into `AbstractSource.isValidToRetry` as the shared default, since four different sources (XAds, RedditAds, Shopify, and now Criteo) all needed the identical behavior. Per-source overrides are now reserved for genuine provider-specific exceptions to that policy.
- Removed the now-redundant `isValidToRetry` overrides in XAds and Shopify (byte-for-byte equivalent to the new default) and shrank RedditAds' override to a one-line addition of 401-retry on top of the shared default.
- Preserved Criteo's existing token-refresh flow: 401 is intentionally *not* retried by the shared default, since Criteo already handles it via a forced-refresh-and-retry-once path in `_makeApiRequest`. `getAccessToken` now also rethrows the original `HttpRequestException` (with its `statusCode`) instead of masking it in a generic `Error`, so token-fetch failures get the same readable summary.
- Added `AbstractConnector._describeError()` to build a one-line human-readable failure summary — HTTP status, the provider's own error message/code/trace ID (when present), and a note when the failure looks transient/server-side — logged immediately before the raw stack trace instead of replacing it.
- Extended `AbstractSource._extractErrorInfo` to also read `title`/`detail` from a provider's `errors[]` array (not just `message`), matching Criteo's actual error shape and giving the new summary real content to show instead of falling back to the raw response body.

# Changes

- Changed `AbstractSource.isValidToRetry` default from always-`false` to retry on 5xx, 429, and network-level (no `statusCode`) errors
- Extended `AbstractSource._extractErrorInfo` to fall back to `title`/`detail` on a provider's `errors[]` entries
- Added `_describeError` and `_formatErrorDetail` to `AbstractConnector`, logged ahead of the stack trace on import failure
- Removed the duplicate `isValidToRetry` overrides in `XAds/Source.js` and `Shopify/Source.js`
- Simplified `RedditAds/Source.js`'s `isValidToRetry` to add only its 401 case on top of the shared default, dropping leftover debug `console.log` calls
- Updated `CriteoAds/Source.js`'s `getAccessToken` to rethrow `HttpRequestException` with its `statusCode` intact instead of wrapping it in a generic `Error`

### DIFF
--- a/.changeset/6714-criteo-ads-retry-and-error-logging.md
+++ b/.changeset/6714-criteo-ads-retry-and-error-logging.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Automatic retry and clearer error messages for Criteo Ads imports
+
+Previously, a temporary server error from Criteo would fail the whole import immediately, and the logs only showed a raw stack trace with no explanation of what went wrong. Now, Criteo Ads imports automatically retry on server errors, rate limits, and network issues, so brief outages on Criteo's side no longer stop your import. When an import does fail, the logs now include a short summary — the HTTP status, the provider's own error message, and a note when the failure looks temporary — so you can tell at a glance whether it's worth just re-running the import.
--- a/packages/connectors/src/Core/AbstractConnector.js
+++ b/packages/connectors/src/Core/AbstractConnector.js
@@ -128,65 +128,14 @@ var AbstractConnector = class AbstractConnector {
           status: EXECUTION_STATUS.ERROR,
           error: error
         });
-        const summary = this._describeError(error);
-        this.config.logMessage(summary ? `${summary}\n${error.stack}` : `${error.stack}`);
+        this.config.logMessage(`${error.stack}`);
         throw error;
 
       }
 
     }
-    //----------------------------------------------------------------
-
-  //---- _describeError -----------------------------------------------
-    /**
-     * Builds a short, human-readable summary of an import failure to log
-     * ahead of the raw stack trace. Returns null when the error carries no
-     * HTTP status code, since the stack trace's first line already states
-     * the error name and message in that case.
-     * @param {Error} error - The error that ended the import
-     * @return {string|null} A one-line summary of the failure, or null
-     * @private
-     */
-    _describeError(error) {
-      if (typeof error?.statusCode !== 'number') {
-        return null;
-      }
-
-      const isServerSide = error.statusCode >= HTTP_STATUS.SERVER_ERROR_MIN;
-      const hint = isServerSide
-        ? " This is a server-side error from the API provider, not your configuration; it may be transient, so re-running the import later can help."
-        : "";
-      const detail = this._formatErrorDetail(error);
-      return `Import failed: the API request failed with HTTP ${error.statusCode}.${detail}${hint}`;
-    }
-    //----------------------------------------------------------------
-
-  //---- _formatErrorDetail ------------------------------------------
-    /**
-     * Extracts readable provider context for the high-level failure summary.
-     * @param {Error} error - The error that ended the import
-     * @return {string} A formatted detail suffix, or an empty string
-     * @private
-     */
-    _formatErrorDetail(error) {
-      const providerError = Array.isArray(error?.payload?.errors) ? error.payload.errors[0] : null;
-
-      if (providerError) {
-        const detailParts = [
-          providerError.message || providerError.title || providerError.detail,
-          providerError.code ? `Provider code: ${providerError.code}.` : "",
-          providerError.traceId ? `Provider trace ID: ${providerError.traceId}.` : ""
-        ].filter(Boolean);
-
-        if (detailParts.length) {
-          return ` ${detailParts.join(" ")}`;
-        }
-      }
-
-      return error.message ? ` ${error.message}` : "";
-    }
-    //----------------------------------------------------------------
-
+    //----------------------------------------------------------------  
+    
   //---- startImportProcess ------------------------------------------
     /**
      * A method for calling from Root script for determining parameters needed to fetch new data.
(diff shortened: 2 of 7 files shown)


zapolsky:

Automated multi-agent worktree review (4 parallel passes — deep reviewer, logic-hunter, architecture/blast-radius, un-checklisted fresh-eyes generalist). Every candidate blocker was adversarially traced and cleared — see **Verified & cleared** below. Nothing here blocks merge.

**Primary note (criticality ≥ 80):**

- `packages/connectors/src/Core/AbstractSource.js:211-215` — **Undocumented blast radius: Google Ads silently changes from never-retry to retry.** Flipping the `isValidToRetry` default from always-`false` to "retry on 5xx / 429 / network" affects *every* source that calls `urlFetchWithRetry` without its own override — not only the four the PR names. **Google Ads** (`packages/connectors/src/Sources/GoogleAds/Source.js:471`, no override) now auto-retries transient errors where it previously failed immediately. It's safe in practice — the retried `googleAds:search` call is an idempotent report read that re-sends the same `pageToken`, so no duplicate rows — but the changeset (`.changeset/6714-criteo-ads-retry-and-error-logging.md`) is worded Criteo-only, while this change (a) enables retries for Google Ads and (b) adds the new failure-summary log line to **all** connectors via `AbstractConnector.run()`. Consider broadening the changelog to state that all default-policy connectors (incl. Google Ads) now retry transient errors and get the new failure summary.

<details>

- `packages/connectors/src/Core/AbstractConnector.js:186` — the `_formatErrorDetail` fallback returns the full `error.message`, which is the entire raw response body when no structured field matched (e.g. an HTML gateway page on a 5xx). That defeats the "one-line summary" goal and bloats the persisted run log (and it duplicates the first line of the `error.stack` appended right after). Consider truncating, e.g. `.slice(0, 300)`.
- `packages/connectors/src/Core/AbstractConnector.js:155-157` — the summary hard-labels every `>= 500` as "may be transient … not your configuration." Correct for the transient case this PR targets, but it also fires for non-transient 5xx (501/505) and is shown *after* the auto-retry already exhausted `MaxFetchRetries`. Minor UX wording.
- **Question (unproven, no realistic regression):** `packages/connectors/src/Core/AbstractSource.js:140-144` → `packages/connectors/src/Sources/CriteoAds/Source.js:499` — preferring `errors[0].title`/`detail` over the raw response text weakens Criteo's *secondary* expired-token detection (`String(error.message).includes('authorization-token-expired')`). The **primary** check reads `error.payload.errors[].code` (unchanged), so there's no realistic break today, but the string fallback now silently depends on `_extractErrorInfo`'s field ordering — worth a comment or a `payload`-based check.
- minor: (a) `AbstractSource.js:212` — the `!error?.statusCode` clause retries *any* status-code-less throwable (e.g. a `TypeError` from `_validateResponse`), not only network errors, for the two newly-affected sources (Google Ads, Criteo) — bounded waste (retries × backoff), but it delays surfacing genuine programming errors. (b) The "clarify logs" cleanup is partial — the same `console.log('isValidToRetry() called')` debug noise remains in `packages/connectors/src/Sources/FacebookMarketing/Source.js:258-259`, and `_shouldRetry` still logs every attempt at `AbstractSource.js:171`. (c) `title`-before-`detail` ordering can surface the generic JSON:API label over the more specific `detail`.

</details>

<details>
<summary>Verified & cleared (adversarially checked — no action needed)</summary>

- **No secret leak.** All four passes independently traced the token path: exceptions are built only from the *response* body (`_extractErrorInfo` → `HttpRequestException{message,statusCode,payload}`), never from the request. Criteo's `client_secret`/bearer token live in request headers/body and never reach a logged or persisted string — including the network-failure path (`TypeError: fetch failed`) and the new `getAccessToken` rethrow.
- **Override removals are behavior-preserving.** The deleted XAds and Shopify overrides are set-equivalent to the new default (503/504 ⊆ `>= 500`; 429 covered). RedditAds' `401 || super()` reproduces its old `{401, 5xx, 429, network}` set exactly.
- **Criteo 401 flow intact, no retry storm.** 401 is (correctly) not retried at the `urlFetchWithRetry` layer, so it reaches `_makeApiRequest`'s forced-refresh-and-retry-once; a second 401 throws cleanly (no loop). Worst-case attempts are bounded.
- **No cross-package version skew.** `_describeError` is not a new message type — it's the same `{type:'log'}` envelope with a longer `message`; the backend parser (`connector-message-parser.service.ts`) validates `message` as an unconstrained `z.string()` and degrades gracefully to UNKNOWN, so an older backend cannot choke on it.
- **No `HTTP_STATUS` ReferenceError** in `AbstractConnector._describeError` — `HTTP_STATUS` is a bundle-global (the same pattern `AbstractSource` already uses), collected before the class files by the vite build.

CI-parity: eslint ✓ *(note: the connectors eslint config ignores `src/**` — pre-existing, not attributable to this PR)*, prettier ✓. No `tsc` (plain JS). No unit tests exist in `packages/connectors`, so none were run.
</details>

<sub>Automated OWOX review · worktree multi-agent passes · all candidate blockers adversarially verified.</sub>

prokopetsv:
@zapolsky, ask you kindly to review.

zapolsky (approved):

PR #1382, opened 2026-07-02
feat(mcp): add update_report MCP tool

laskevych:
## Summary

Adds an `update_report` MCP tool (PRD §15.10, task #6701): an AI assistant connected over MCP can partially update an existing report — rename it and/or replace which data mart fields it exports. Everything not mentioned stays exactly as it was.

Write tool (`mcp:write`, `readOnlyHint: false`), free per the PRD billing model.

> **Stacked on #1381** (`add_report`), which is stacked on #1377 (`get_data_mart_reports`) — all three extend the same `McpReportsFacade`. GitHub will retarget as the stack merges.

## What changed

- **New MCP tool** `UpdateReportTool` (`ee/mcp/tools/update-report.tool.ts`) — input `{ report_id, fields?, name? }` (at least one of `fields`/`name` required; `name` trimmed; `fields: ['*']` = all columns); output exactly the PRD shape `{ report_id, status: 'updated' }`.
- **`McpReportsFacade.updateReport`** — load-then-merge over the existing domain services; nothing reimplemented.

## Design notes

- **Load-then-merge is the core.** `UpdateReportCommand` carries the *full* report state and `UpdateReportService` overwrites every output control, while the PRD input is a partial PATCH. The facade therefore loads the current report via `GetReportService` (which also enforces SEE access and 404s cleanly before any mutation) and merges the requested changes into a full command.
- **What is preserved, and how:** destination id is passed unchanged (so the service skips its destination re-auth path), `destinationConfig`, `filterConfig`, `sortConfig`, `limitConfig`, `aggregationConfig`, `dateTruncConfig`, `uniqueCountConfig` are passed through from the current report, and `ownerIds` stays `undefined` — `UpdateReportService` only touches owners when `ownerIds !== undefined`, so ownership is untouched.
- **Authorization** is delegated to the domain services: `GetReportService` checks SEE, `UpdateReportService` checks mutate access, validates the column selection (`OutputControlsValidatorService`) and invalidates the report data cache when configs change. No external side effects here, so no pre-flight duplication is needed (unlike `add_report`).
- **Scope, consistent with the rest of the MCP track:** `slices`/`filters` and `description` are deferred (PRD filter vocabulary doesn't map 1:1 onto the domain filter model; `description` has no field on the Report entity). Schedules are not touched — they belong to the dedicated schedule tools (#1378), per the "MCP mirrors ODM" team decision.
- Input schema detail: the client-facing `zodSchema` exposes the raw shape (both change fields optional); the server-side parse additionally enforces "at least one of `fields`/`name`" with a clear message, since a no-op update is a caller mistake worth surfacing to the LLM.

## Testing

- Facade: rename-only preserves columns/filters/sort/limit/aggregations/destination/owners; fields-only preserves the title; `['*']` → no projection; name+fields together; get-failure short-circuits (no update call).
- Tool: happy path payload/output, at-least-one-change refine, strict input, name trimming + whitespace-only rejection, `mcp:write` scope + annotations, registry.
- `npx jest src/data-marts/facades src/ee/mcp/tools` — 60 tests, 12 suites green; `nest build` and eslint clean.

## Out of scope / follow-ups

- `delete_report` MCP tool — last one on this track.
- `slices`/`filters` input support — shared design with the future `query_data_mart` filter vocabulary.

### DIFF
--- a/apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
+++ b/apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
@@ -537,6 +537,11 @@ describe('McpReportsFacadeImpl.updateReport', () => {
     expect(updateReportService.run).toHaveBeenCalledWith(
       expect.objectContaining({
         id: 'report-1',
+        // Authorization inputs must be forwarded verbatim — UpdateReportService
+        // derives mutate-access decisions from them.
+        projectId: 'project-1',
+        userId: 'user-1',
+        roles: ['editor'],
         title: 'New name',
         dataDestinationId: 'dest-1',
         destinationConfig: currentReport.destinationConfig,
@@ -550,6 +555,16 @@ describe('McpReportsFacadeImpl.updateReport', () => {
       })
     );
     expect(result).toEqual({ report_id: 'report-1', status: 'updated' });
+  });
+
+  it('rejects a call with nothing to update before touching any service', async () => {
+    const { facade, getReportService, updateReportService } = buildUpdateFacade();
+
+    await expect(facade.updateReport(updateRequest)).rejects.toThrow(
+      'Nothing to update: provide fields and/or name'
+    );
+    expect(getReportService.run).not.toHaveBeenCalled();
+    expect(updateReportService.run).not.toHaveBeenCalled();
   });
 
   it('replaces the column selection while preserving the name and filters', async () => {
--- a/apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
+++ b/apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
@@ -9,6 +9,8 @@ jest.mock('../use-cases/google-sheets/create-google-sheet-document.service', () 
   CreateGoogleSheetDocumentService: jest.fn(),
 }));
 jest.mock('../services/data-mart.service', () => ({ DataMartService: jest.fn() }));
+jest.mock('../use-cases/get-report.service', () => ({ GetReportService: jest.fn() }));
+jest.mock('../use-cases/update-report.service', () => ({ UpdateReportService: jest.fn() }));
 jest.mock('../services/output-controls-validator.service', () => ({
   OutputControlsValidatorService: jest.fn(),
 }));
@@ -31,6 +33,8 @@ import type { CreateGoogleSheetDocumentService } from '../use-cases/google-sheet
 import type { AccessDecisionService } from '../services/access-decision';
 import type { DataMartService } from '../services/data-mart.service';
 import type { OutputControlsValidatorService } from '../services/output-controls-validator.service';
+import type { GetReportService } from '../use-cases/get-report.service';
+import type { UpdateReportService } from '../use-cases/update-report.service';
 import { McpReportsFacadeImpl } from './mcp-reports.facade.impl';
 
 function buildReport(overrides: {
@@ -108,6 +112,12 @@ function buildFacade(reports: ReportDto[], triggers: DataMartScheduledTrigger[])
   const outputControlsValidator = {
     validateForReport: jest.fn().mockResolvedValue(undefined),
   } as unknown as jest.Mocked<OutputControlsValidatorService>;
+  const getReportService = {
+    run: jest.fn(),
+  } as unknown as jest.Mocked<GetReportService>;
+  const updateReportService = {
+    run: jest.fn(),
+  } as unknown as jest.Mocked<UpdateReportService>;
   const facade = new McpReportsFacadeImpl(
     listReportsByDataMartService,
     scheduledTriggerService,
@@ -115,7 +125,9 @@ function buildFacade(reports: ReportDto[], triggers: DataMartScheduledTrigger[])
     createGoogleSheetDocumentService,
     dataMartService,
     accessDecisionService,
-    outputControlsValidator
+    outputControlsValidator,
+    getReportService,
+    updateReportService
   );
   return {
     facade,
@@ -126,6 +138,8 @@ function buildFacade(reports: ReportDto[], triggers: DataMartScheduledTrigger[])
     dataMartService,
     accessDecisionService,
     outputControlsValidator,
+    getReportService,
+    updateReportService,
   };
 }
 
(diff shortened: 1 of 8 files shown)


apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -43,9 +49,43 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+      new UpdateReportCommand(
+        request.reportId,
+        request.projectId,
+        request.userId,
+        request.roles,
+        request.name ?? current.title,
+        current.dataDestinationAccess.id,
+        current.destinationConfig,
+        undefined,
+        request.fields ? this.toColumnConfig(request.fields) : (current.columnConfig ?? null),
```
romandubovyi: 💡 **Suggestion:** This uses a truthy check (`request.fields ? ... : ...`), which relies on the Zod tool-layer schema (`min(1)`) to keep an empty array from reaching here. Since this facade method is a public interface not exclusively bound to that Zod-validated path, `request.fields !== undefined` would be more robust and consistent with the `command.ownerIds !== undefined` pattern used for the same "was this explicitly provided" check elsewhere.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
```diff
@@ -477,3 +491,108 @@ describe('McpReportsFacadeImpl.addReport', () => {
...
+
+    expect(getReportService.run).toHaveBeenCalledWith(
+      expect.objectContaining({
+        id: 'report-1',
+        projectId: 'project-1',
+        userId: 'user-1',
+        roles: ['editor'],
+      })
+    );
+    expect(updateReportService.run).toHaveBeenCalledWith(
```
romandubovyi, line 537: 🟡 **Important:** The test asserts most fields of the `UpdateReportCommand`, but it does not verify that `projectId`, `userId`, and `roles` are forwarded correctly. These fields drive authorization inside `UpdateReportService`, so please add assertions for them to prevent silent regressions.

apps/backend/src/ee/mcp/tools/update-report.tool.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type UpdateReportInput = z.infer<typeof inputSchema>;
+
+@Injectable()
+export class UpdateReportTool implements McpToolDefinition<UpdateReportInput> {
+  readonly name = 'update_report';
+  readonly description =
+    'Update an existing report: rename it and/or replace which data mart fields it exports. Anything not provided stays unchanged.';
+  readonly zodSchema = baseInputSchema.shape;
+  readonly outputSchema = {
```
romandubovyi, line 48: 💡 **Suggestion:** Add `.describe()` to the `outputSchema` fields (e.g., `report_id` and `status`) to stay consistent with other tools like `AddReportTool` and help MCP clients understand the response shape.

apps/backend/src/data-marts/facades/mcp-reports.facade.ts
```diff
@@ -79,4 +95,11 @@ export interface McpReportsFacade {
    * rejected by the underlying document-creation service.
    */
   addReport(request: McpAddReportRequest): Promise<McpAddReportResult>;
+  /**
+   * Partially updates a report (name and/or column selection). The domain
```
romandubovyi, line 99: 🔵 **Question:** The "at least one of `fields`/`name`" invariant is only enforced by the Zod `.refine()` in `UpdateReportTool`. Should this JSDoc call that out as a precondition (or should the facade itself guard against a no-op call), since `McpReportsFacade` is a general interface and not exclusively reached through the validated tool input?

apps/backend/src/ee/mcp/tools/update-report.tool.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { McpAuthContext } from '../auth/mcp-auth-context';
+import { jsonToolResult, type McpToolDefinition, type McpToolResult } from './mcp-tool.definition';
+
+// The raw shape (exposed to MCP clients) has both change fields optional; the
+// parsed schema additionally requires at least one of them, since an update
+// with nothing to change is a caller mistake worth surfacing.
+const baseInputSchema = z
+  .object({
+    report_id: z.string().min(1),
+    fields: z
```
romandubovyi, line 17: 💡 **Suggestion:** Consider adding a small example to the `fields` parameter description, e.g., `["field_name_1", "field_name_2"]`. This helps LLM clients understand the array structure more clearly.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -43,9 +49,43 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+    private readonly getReportService: GetReportService,
+    private readonly updateReportService: UpdateReportService
   ) {}
 
+  async updateReport(request: McpUpdateReportRequest): Promise<McpUpdateReportResult> {
+    // UpdateReportCommand carries the FULL report state and the service
+    // overwrites every output control, so merge the partial MCP input into the
+    // current report. Keeping the current destination id skips the destination
+    // re-auth path, and leaving ownerIds undefined keeps owners untouched.
+    const current = await this.getReportService.run(
```
romandubovyi, line 68: 🔵 **Question:** `GetReportService` enforces `Action.SEE` on the data mart before `UpdateReportService` checks mutate access on the report itself. Is this intentional for MCP updates? A user who can mutate the report but lacks `SEE` on the data mart would be rejected here rather than by the update access check.

apps/backend/src/ee/mcp/tools/update-report.tool.ts
```diff
@@ -0,0 +1,76 @@
...
+);
+
+type UpdateReportInput = z.infer<typeof inputSchema>;
+
+@Injectable()
+export class UpdateReportTool implements McpToolDefinition<UpdateReportInput> {
+  readonly name = 'update_report';
+  readonly description =
+    'Update an existing report: rename it and/or replace which data mart fields it exports. Anything not provided stays unchanged.';
+  readonly zodSchema = baseInputSchema.shape;
```
romandubovyi, line 47: 💡 **Suggestion:** The cross-field requirement "at least one of `fields`/`name`" is enforced by `inputSchema` in `parseInput`, but `zodSchema` exposed to MCP clients is `baseInputSchema.shape`, which presents both fields as optional. Consider moving the `.refine()` into `baseInputSchema` or documenting the requirement in the tool/field descriptions so MCP clients can surface it before calling.

romandubovyi (approved):
No blocking issues. The load-then-merge approach in the facade cleanly preserves unmentioned report settings, authorization stays delegated to domain services, and test coverage is solid. I left a few non-blocking suggestions around test assertions and MCP schema consistency.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Done in 34fb5e72c — switched to `request.fields !== undefined`, consistent with the `ownerIds !== undefined` pattern. Same spirit applied one level up: the facade no longer relies on the tool-layer schema at all (see the no-op guard thread).

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
laskevych: Done in 34fb5e72c — the assertion now verifies `projectId`, `userId`, and `roles` are forwarded into `UpdateReportCommand` verbatim, with a comment noting they drive authorization.

apps/backend/src/ee/mcp/tools/update-report.tool.ts
laskevych: Done in 34fb5e72c — `report_id` and `status` now carry `.describe()` in the output schema.

apps/backend/src/data-marts/facades/mcp-reports.facade.ts
laskevych: Went with the stronger option in 34fb5e72c: the facade now guards the invariant itself (`BadRequestException: Nothing to update…` before any service call), so it holds for any caller, not only the Zod-validated tool path; the interface JSDoc states the precondition as well. Covered by a new spec (`rejects a call with nothing to update before touching any service`).

apps/backend/src/ee/mcp/tools/update-report.tool.ts
laskevych: Done in 34fb5e72c — the `fields` description now includes the example `["field_name_1", "field_name_2"]` alongside the `["*"]` form.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Intentional. The load is inherent to load-then-merge (the full current state is needed to build the command), and SEE on the data mart is the minimal precondition for reading it. A user with mutate access but no SEE would be an inconsistent ACL state in ODM — and either way the call is denied with an access error; only the message differs. Failing on the weaker check first also avoids leaking report state to a caller who cannot see the data mart.

apps/backend/src/ee/mcp/tools/update-report.tool.ts
laskevych: You are right that the `.refine()` is invisible to clients — it cannot live in the raw shape, because the MCP SDK consumes a `ZodRawShape` and `.refine()` produces a `ZodEffects` with no `.shape`. Addressed in 34fb5e72c on two fronts: the requirement is now stated in the tool description and in both field descriptions (so clients see it before calling), and the facade enforces the same invariant server-side independent of the tool schema.

PR #1383, opened 2026-07-02
feat(mcp): add delete_report MCP tool

laskevych:
## Summary

Adds a `delete_report` MCP tool (PRD §15.12, task #6703): an AI assistant connected over MCP can delete a report. The report stops running and disappears from the project; the underlying data mart, destination, and any already-exported documents are not affected.

Destructive write tool: `mcp:write` scope, `readOnlyHint: false`, **`destructiveHint: true`** — MCP clients prompt the user for confirmation before calling it (and the annotation matters for the Anthropic connector-directory review, PRD §19). Free per the PRD billing model.

**This completes the report/destination MCP tool track** (`list_destinations` #1376 ✅, `get_data_mart_reports` #1377, `add_report` #1381, `update_report` #1382, `delete_report` — this PR).

> **Stacked on #1382** (`update_report`); full chain #1377 ← #1381 ← #1382 ← this. GitHub will retarget as the stack merges.

## What changed

- **New MCP tool** `DeleteReportTool` (`ee/mcp/tools/delete-report.tool.ts`) — input `{ report_id }`, output exactly the PRD shape `{ report_id, status: 'deleted' }`.
- **`McpReportsFacade.deleteReport`** — a thin delegation to the existing `DeleteReportService`; nothing reimplemented.

## Design notes

- **Authorization and existence checks stay in the domain:** `DeleteReportService` 404s on an unknown/foreign-project id and enforces mutate access (`ReportAccessService.checkMutateAccess`) before deleting. The facade adds nothing on top — deletion has no pre-side-effect hazard (unlike `add_report`).
- **Deleting an unknown id is a not-found error, not a no-op** — an LLM retrying a failed call gets an honest signal instead of a fabricated success.
- **The confirmation is synthesized:** the service returns `void`, so the facade builds `{ report_id, status: 'deleted' }` after the call succeeds.
- **External cleanup is async by design:** the service emits `report.deleted` (e.g. Google Sheets developer-metadata cleanup) after the DB delete; the tool response does not wait for it.

## Testing

- Facade: command payload + synthesized confirmation; not-found propagation (no fabricated success).
- Tool: happy path, strict input (`report_id` required, extras rejected), `mcp:write` + `destructiveHint: true` annotations, registry.
- `npx jest src/data-marts/facades src/ee/mcp/tools` — 65 tests, 13 suites green; `nest build` and eslint clean.

## Out of scope / follow-ups

- `query_data_mart` and destination-connect tools — the remaining PRD §15 surface, tracked separately.
- `run_report` / report-run-schedule tools live in #1378.

### DIFF
--- a/apps/backend/src/ee/mcp/tools/delete-report.tool.spec.ts
+++ b/apps/backend/src/ee/mcp/tools/delete-report.tool.spec.ts
@@ -49,6 +49,15 @@ describe('DeleteReportTool', () => {
     expect(() => tool.parseInput({ report_id: 'report-1', extra: true })).toThrow();
   });
 
+  it('propagates facade errors instead of swallowing them', async () => {
+    const facade = {
+      deleteReport: jest.fn().mockRejectedValue(new Error('Report with ID report-1 not found')),
+    } as unknown as jest.Mocked<McpReportsFacade>;
+    const tool = new DeleteReportTool(facade);
+
+    await expect(tool.handler({ report_id: 'report-1' }, context)).rejects.toThrow('not found');
+  });
+
   it('is registered as a destructive write tool with the mcp:write scope', () => {
     const registry = new McpToolRegistry([new DeleteReportTool({} as McpReportsFacade)]);
 
--- a/apps/backend/src/ee/mcp/tools/delete-report.tool.spec.ts
+++ b/apps/backend/src/ee/mcp/tools/delete-report.tool.spec.ts
@@ -1,0 +1,68 @@
+import type { McpReportsFacade } from '../../../data-marts/facades/mcp-reports.facade';
+import type { McpAuthContext } from '../auth/mcp-auth-context';
+import { McpToolRegistry } from './mcp-tool.registry';
+import { DeleteReportTool } from './delete-report.tool';
+import { MCP_TOOL_PROVIDER_CLASSES } from './mcp-tool.providers';
+
+describe('DeleteReportTool', () => {
+  const context: McpAuthContext = {
+    clientId: 'mcp-client-1',
+    userId: 'user-1',
+    projectId: 'project-1',
+    email: 'ann@owox.com',
+    roles: ['editor'],
+    resource: 'https://mcp.owox.com/mcp',
+    scopes: ['mcp:write'],
+    authFlow: 'mcp',
+  };
+
+  it('deletes a report and returns the confirmation shape', async () => {
+    const facade = {
+      deleteReport: jest.fn().mockResolvedValue({ report_id: 'report-1', status: 'deleted' }),
+    } as unknown as jest.Mocked<McpReportsFacade>;
+    const tool = new DeleteReportTool(facade);
+
+    const structuredContent = { report_id: 'report-1', status: 'deleted' };
+
+    await expect(tool.handler({ report_id: 'report-1' }, context)).resolves.toEqual({
+      structuredContent,
+      content: [
+        {
+          type: 'text',
+          text: JSON.stringify(structuredContent, null, 2),
+        },
+      ],
+    });
+    expect(facade.deleteReport).toHaveBeenCalledWith({
+      reportId: 'report-1',
+      projectId: 'project-1',
+      userId: 'user-1',
+      roles: ['editor'],
+    });
+  });
+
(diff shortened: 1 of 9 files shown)


apps/backend/src/ee/mcp/tools/delete-report.tool.ts
```diff
@@ -0,0 +1,54 @@
...
+
+type DeleteReportInput = z.infer<typeof inputSchema>;
+
+@Injectable()
+export class DeleteReportTool implements McpToolDefinition<DeleteReportInput> {
+  readonly name = 'delete_report';
+  readonly description =
+    'Permanently delete a report by id. The report stops running and disappears from the project; the underlying data mart, destination, and any already-exported documents are not affected. This cannot be undone.';
+  readonly zodSchema = inputSchema.shape;
+  readonly outputSchema = {
```
romandubovyi, line 21: 💡 **Suggestion:** Consider adding `.describe()` to `report_id` and `status` so the output schema matches the detail level of `UpdateReportTool` and helps MCP clients surface field meaning.

apps/backend/src/ee/mcp/tools/delete-report.tool.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+    expect(facade.deleteReport).toHaveBeenCalledWith({
+      reportId: 'report-1',
+      projectId: 'project-1',
+      userId: 'user-1',
+      roles: ['editor'],
+    });
+  });
+
+  it('requires report_id and rejects unexpected input', () => {
```
romandubovyi, line 44: 💡 **Suggestion:** Consider adding a test that verifies facade errors (e.g., not found or forbidden) propagate uncaught through `handler`. This protects against a future change accidentally swallowing them.

apps/backend/src/ee/mcp/tools/delete-report.tool.ts
```diff
@@ -0,0 +1,54 @@
...
+    report_id: z.string(),
+    status: z.literal('deleted'),
+  };
+  readonly annotations = {
+    title: 'Delete Report',
+    readOnlyHint: false,
+    destructiveHint: true,
+    openWorldHint: false,
+  };
+  readonly requiredScopes: McpScope[] = ['mcp:write'];
```
romandubovyi, line 31: 🔵 **Question:** I notice that `requiredScopes` here is set to `['mcp:write']`, whereas in schedule-related write tools (like `delete-report-run-schedule.tool.ts`), it is set to `['mcp:read', 'mcp:write']`. Is this scope mismatch deliberate, or should all write tools require both scopes for consistency?

romandubovyi (approved):
No blocking issues found. Approving.

apps/backend/src/ee/mcp/tools/delete-report.tool.ts
laskevych: Done in 07e13f873 — `report_id` and `status` now carry `.describe()`, matching `update_report`.

apps/backend/src/ee/mcp/tools/delete-report.tool.spec.ts
laskevych: Done in 07e13f873 — added `propagates facade errors instead of swallowing them`: a rejected facade call is asserted to reject through `handler` unchanged.

apps/backend/src/ee/mcp/tools/delete-report.tool.ts
laskevych: Deliberate on this side, and you have spotted a real cross-tool inconsistency. `assertScopes` requires EVERY listed scope, so declaring `[mcp:read, mcp:write]` would additionally reject a hypothetical write-only token — the report write tools (`add_report`, `update_report`, and this one, all following the same convention) declare only the capability they exercise, per least-required-scope. In practice there is no behavioral difference today: clients request both scopes together, so every `mcp:write` token also carries `mcp:read`. Since three merged report tools already use `[mcp:write]` while the schedule tools use both, I would harmonize in a single small follow-up rather than flip one tool here — my lean is `[mcp:write]`-only for all write tools, but happy to go the other way if the team prefers the explicit pair. Flagged it as a follow-up task.

PR #1384, opened 2026-07-02
feat(mcp): add query_data_mart tool + blended-field discovery

zapolsky:
## What

Adds a `query_data_mart` MCP tool and makes the data-mart MCP tools blended-aware, so an AI assistant connected over MCP can query a data mart's data — not just its metadata.

- **`query_data_mart`** — structured query over the existing report engine (`ReportSqlComposerService.compose` + `ReportTotalsService.computeTotals`; no MCP-custom SQL, no LLM-generated SQL, no invented joins). Fields, pre-join `slices` / post-join `filters`, aggregations (SUM/COUNT/COUNT_DISTINCT/AVG/MIN/MAX + P25/P50/P75/P95), `date_buckets`, `limit` (1–1000, default 20). Header-once TSV rows with a 128 KiB byte-cap, plus server-side `totals` computed over all matching rows (ignoring the limit). Per-user access + billing gating; each call is recorded in Run History (query definition + executed SQL only — never row values).
- **`get_data_mart_details_by_id`** now returns `joined_fields` — the fields contributed by blended/joined data marts, with their qualified `<alias>__<field>` names, source data mart, type, and allowed aggregations. Filtered to the sources the caller may report on.

## Design notes

- `readOnlyHint: false` + `mcp:read` scope is intentional: a billable, non-idempotent read (billing gates it; a separate `mcp:write` scope would be semantically wrong for a read).
- `rows` is a header-once TSV string (token efficiency), documented in the tool description.
- Public MCP docs updated: the data-sharing note now states that query result rows/totals are sent to the AI provider (they weren't before).

## Validation

- Three independent reviews (comprehensive line-by-line, adversarial bug-hunt, PR merge-readiness); all findings addressed in the `fix(mcp): address code-review findings` commit — notably gating `joined_fields` by per-source reporting access, and making `reader.finalize()` best-effort.
- Live end-to-end tested through Claude: discovery → query → aggregations/date-buckets → blended-field group-by → pre-join slice / post-join filter, all matching expected results.
- `nest build` clean; backend unit + real-BigQuery integration tests pass.

## Follow-ups (non-blocking, post-merge)

- `nl_prompt` is not captured in Run History (no natural-language text is present at the tool boundary).
- Formalize the `limit`-on-aggregations semantics in the docs.

Fibery: query_data_mart #6698


apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -0,0 +1,217 @@
...
+- aggregations: SUM, COUNT, COUNT_DISTINCT, AVG, MIN, MAX, and percentiles P25/P50/P75/P95. Group-by is implied by the non-aggregated fields you select.
+- date_buckets: bucket a date/timestamp field by DAY/WEEK/MONTH/QUARTER/YEAR (e.g. "revenue by month").
+- fields must list every column the query uses, INCLUDING any field named in aggregations or date_buckets — a field you aggregate or bucket but omit from fields is rejected. Example — "revenue by month": fields ["ts", "revenue"], aggregations [{field: "revenue", function: "SUM"}], date_buckets [{field: "ts", unit: "MONTH"}]. (Filters are the exception: a filter may reference a field that is not in fields.)
+
+Choosing between slices and filters:
+- slices (pre-join): narrow a JOINED data mart before it is blended in — criteria on a joined data mart's own fields only. Slices do NOT apply to the main data mart. More efficient — they reduce the joined volume before the join.
+- filters (post-join): criteria on the blended result — use for anything on the MAIN data mart's fields, on a joined field, or on an aggregated value.
```
zapolsky: **[Bug — verified] Aggregate-value `filters` silently become a row-level `WHERE`, not `HAVING` — wrong numbers, no error.**

This line (and the `:46` example) instructs the model to express an aggregate threshold as a `filter` ("…or on an aggregated value" / "the >100 threshold on the aggregated total → filters"). But the MCP `filters` surface has no `HAVING` marker, so such a filter is mistranslated into a *pre-aggregation* `WHERE`, producing a plausible-looking but silently wrong result.

**Evidence chain (adversarially re-verified — CONFIRMED):**
- `query-data-mart.input.ts:44-48` — `McpFilterSchema` is `{ field, operator, value }`; there is no `function` field (only `McpAggregationSchema` carries `function`, and it feeds `aggregations`, not `filters`).
- `query-data-mart.input.ts:159-193` — `mapOne` always builds `{ column, placement, operator, value }` and never sets `function`. So a `HAVING` rule is **unreachable** from the MCP tool entirely.
- `output-controls-validator.service.ts:185` + `sql-clause-renderer.ts:81` — a function-less rule always renders into `WHERE`; `:229` / `:124` — only `function`-carrying rules render into `HAVING`.
- `output-controls-validator.service.ts:201-208` — a post-join `WHERE` rule on a column that is *also* in `aggregationConfig` is accepted. There is no WHERE-vs-aggregation cross-check anywhere, so no error is raised.
- `bigquery-query.builder.ts:103` — the composed SQL is `… WHERE revenue > @p0 … GROUP BY <dims>` with `SUM(revenue)` in the SELECT and an empty `HAVING`.

**Trigger:** `aggregations:[{field:'revenue',function:'SUM'}]` + `filters:[{field:'revenue',operator:'gt',value:100}]` (the exact shape this description recommends). Intended `HAVING SUM(revenue) > 100`; actual `WHERE revenue > 100 … GROUP BY dims` → a SUM over only the raw rows where `revenue > 100`, not groups whose total exceeds 100. Returned with `truncated:false` and **no error** — silently wrong analytics surfaced through the AI assistant.

**Fix (any of):**
1. Correct/remove the `:44` and `:46` guidance so the tool stops recommending aggregate-value thresholds as `filters`.
2. Defense-in-depth: reject a post-join `WHERE` filter whose column is also in `aggregationConfig` with a clear `field_*` error instead of silently building the wrong SQL.
3. To actually support it, expose the `function` marker on the MCP filter schema so an aggregate threshold can emit `HAVING`.

zapolsky (commented):
🔴 **Blocked issue** — see the inline comment. This should be addressed before merge, but this review does not formally block; a human owns the merge decision. Posted by automated OWOX review (worktree multi-agent pass; the blocker was adversarially re-verified).

zapolsky:

One 🔴 blocking issue was posted inline (aggregate-value `filters` → `WHERE` instead of `HAVING`). Everything below is **non-blocking** — none of it gates merge.

Scope: worktree checkout of the PR head, 5 parallel agents (line-by-line reviewer, logic-hunter, architect, tester, un-checklisted fresh-eyes) + one adversarial skeptic on the blocker. **Verified non-issues** (traced and disproved): per-user joined-source reporting access *is* re-enforced on the query path (`assertAllRequestedSourcesAccessible`, not just discovery); `Action.SEE` ≡ `Action.USE` for data marts in the access matrix, so the gate matches report execution; totals correctly ignore the row limit (`composeTotals` rebuilds the plan with `limitConfig: null`); Run History persists query definition + executed SQL only (never row values); execution is fully parameterized (no SQL injection via field/filter/timezone); no cross-tenant access; no double-billing; frontend handles the new `MCP_QUERY` run type without crashing.

Primary notes (criticality ≥ 80), highest first — all printed:

- `apps/backend/src/ee/mcp/mappers/mcp-error.mapper.ts:9` (via `blended-report-data.service.ts:443`) — **Access-denial and raw DWH errors are forwarded verbatim to the third-party AI provider.** The source-access `BusinessViolationException` carries `deniedDataMartIds`/`userId` (not `unknownColumns`), so it misses every `mapError` branch and falls through to `toToolError(err)`, which returns `err.message` unchanged. That message is `Cannot build report SQL, user "<fullName/email>" is missing access to data marts: "<restricted mart title>"` — i.e. a **restricted data-mart title** (a field the discovery tool deliberately hides) plus the **caller's name/email** are sent to the provider. The same fallback also forwards raw BigQuery/DWH error text (can echo SQL fragments/identifiers). Fix: map source-access violations (and unclassified errors) to a generic sanitized `permission_denied`/`query_failed` message that omits titles, user identity, and raw driver text.

- `apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts` (`mapError`) + `apps/backend/src/data-marts/dto/schemas/field-aggregation-governance.ts:41-48` — **Advertised aggregations are off by default and their rejection is opaque.** The description and public docs promise "COUNT, MIN, MAX … and percentiles P25/P50/P75/P95", but default per-field governance enables only numeric `['SUM','AVG','MIN','MAX']` and string `['COUNT','COUNT_DISTINCT']` — **percentiles/COUNT-on-numeric require an explicit admin per-field override**. When they aren't enabled, `output-controls-validator.service.ts:363-369` raises `AGGREGATION_FUNCTION_NOT_ALLOWED_FOR_FIELD` → `BadRequestException('Output controls validation failed')`, and `mapError` does **not** handle that code, so the LLM gets the bare string `"Output controls validation failed"` with no `error_code` and no field/function detail — it cannot self-correct. Fix: map `AGGREGATION_FUNCTION_NOT_ALLOWED_FOR_FIELD` (and the sibling `..._FOR_TYPE`, `DUPLICATE_AGGREGATION`, HAVING/operator-type codes) to a structured error naming the field + allowed functions, and/or narrow the description to the default-enabled set.

- `apps/backend/src/data-marts/use-cases/query-data-mart.service.ts:55` — **Existence oracle: hidden-DM vs missing-DM are distinguishable by message.** The comment says the NotFound is thrown "so the caller cannot tell it apart from a missing one", but the missing path (`data-mart.service.ts:48`) throws `Data Mart with id <id> and projectId <pid> not found` while the hidden path (`:65`) throws `Data Mart not found`. Both map to `permission_denied`, but the text passes through verbatim, so a caller can distinguish "exists-but-hidden" from "doesn't exist" within their project. Fix: throw an identical constant message on both paths (or catch the `getByIdAndProjectId` NotFound and rethrow the same `'Data Mart not found'`).

<details>

- `apps/backend/src/data-marts/use-cases/query-data-mart.service.ts:139,164,204,214` & `facades/mcp-data-marts.facade.impl.ts:128` — `this.logger.warn('msg', { error: err })` misuses the Nest `Logger`: the 2nd arg is a `context` **string**, so the actual error object is dropped/garbled — exactly on the degraded paths (totals-degraded, swallowed audit, finalize failure) where you most need it. Interpolate the message instead (as already done at `:171-173`).
- `apps/backend/src/ee/mcp/tools/tabular-serializer.ts:8-14` — `cell()` uses `String(v)`, so RECORD/STRUCT/ARRAY/object column values serialize to `[object Object]` in the rows sent to the LLM. Consider `JSON.stringify` for non-primitive cells.
- `apps/backend/src/data-marts/use-cases/query-data-mart.service.ts:143-174` — billing and Run-History recording are independent best-effort steps: if the SUCCESS audit write fails (swallowed) but `registerMcpQueryRunConsumption` succeeds, the user is **billed with no Run-History record** (and the consumption `reportRunId` resolves to nothing); the reverse is also possible. By design, but worth surfacing to the billing team for reconciliation.
- `apps/backend/src/ee/mcp/tools/query-data-mart.input.ts:44-48,159-193` — MCP filter `value` is `z.unknown()` and `mapOne` casts it `as never`, bypassing `ScalarValueSchema`/the `.finite()` guard. Not exploitable (params are bound; inlined SQL is display-only), but a non-scalar / `Infinity` value yields an opaque driver error + a FAILED run instead of a clean `invalid_input`. Validate mapped values against `ScalarValueSchema`.
- `apps/backend/src/ee/mcp/tools/tabular-serializer.ts:29` — the 128 KiB cap is soft: a single oversized first row is always emitted in full (then embedded a second time in `content[0].text`), so per-row payload/cost is effectively unbounded. Bounded to one row; worth documenting the caveat.
- `apps/backend/src/data-marts/services/blended-report-data.service.ts:416-437` — the query path enforces source **exclusion** (`isIncluded === false`) only for pre-join slices; for `fields`/post-join filters it checks only `isAccessibleForReporting`. Discovery hides excluded-source fields, so an excluded-but-accessible `<alias>__<field>` named directly is queryable although discovery hides it. Not a cross-user leak (the caller holds USE access) and it matches the pre-existing report-run path — but discovery and query now disagree on exclusion governance.
- `apps/backend/src/data-marts/use-cases/query-data-mart.service.ts:127,156` — persisted `metadata.truncated` reflects only row-limit truncation, never the serializer's byte-cap, so a byte-capped-but-not-row-limited run records `truncated:false` in Run History while the client correctly sees `truncated:true`. Audit fidelity only.
- minor: `tabular-serializer.ts:22` header row is not escaped like data cells (asymmetry vs the file's own escaping contract; not reachable with today's identifier-derived headers) · `mcp-data-marts.facade.impl.ts:123-125` drops `allowedAggregations` for an explicit empty `[]` ("none permitted"), making it indistinguishable from "unrestricted" to the LLM · `query-data-mart.input.ts` `McpAggregationSchema` accepts `STRING_AGG`/`ANY_VALUE` (12 enum values) though the description lists 10, and `McpOperatorEnum` advertises the four always-rejected operators (`in`/`not_in`/`in_next_n_days`/`this_week`) — intentional per comments, but the LLM sees them as valid · `query-data-mart.service.ts` never bounds-checks `r.limit` (unreachable via the tool, which clamps 1–1000; a direct facade caller with `limit <= 0` would read + bill, then silently lose the Run-History row when `McpQueryRunMetadataSchema.parse` throws inside the best-effort audit).

</details>

### CI-parity (Phase D)
- `eslint` (backend + web changed files): **clean** (0 problems).
- `jest` (6 affected suites, run in the worktree): **108 passed / 108**.
- `tsc -b`: **not attributable / not verifiable locally** — the local toolchain produces **91 identical baseline errors on both `main` and this PR head** (unrelated spec/e2e files; the 6 in the touched `mcp-data-marts.facade.impl.spec.ts` are byte-identical on `main`, only line-shifted), i.e. **zero net new type errors from this PR**. CI type-checks test files, so the local run isn't CI-parity; relying on the author's stated "`nest build` clean".
- changeset present (`owox: minor`); no HTTP controllers changed (no `@Auth` surface).

<sub>Automated OWOX review · worktree multi-agent passes · the blocker was adversarially verified.</sub>

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Fixed in `8cb11fd60` via **option 1** (correct the guidance).

The slices/filters block now states that both are row-level predicates applied to raw values **before any aggregation**, that there is **no HAVING**, and that filtering by an aggregated total is not supported. Removed the misleading `">100 threshold on the aggregated total → filters"` example, and the `filters` line now spells out that a filter on a field you also aggregate restricts *which raw rows feed the aggregate*, not the group total.

Deliberately **not** taking option 2 (reject a post-join `WHERE` whose column is also in `aggregationConfig`): that produces a false positive on the legitimate `WHERE revenue > 0` + `SUM(revenue)` ("sum over positive rows") pattern, which is a valid query. Option 3 (expose a `function`/HAVING marker on the MCP filter schema) is a real feature and is tracked as a follow-up, not part of this review round.

zapolsky:
Addressed the review in `8cb11fd60` (new commit — no amend/force). The 🔴 blocker reply is inline on its thread. The three ≥80 notes:

**1. Access-denial / raw DWH errors forwarded to the AI provider** — fixed. The unclassified fallback no longer returns `err.message`; it returns a generic `query_failed`. Source-access `BusinessViolationException` (`deniedDataMartIds`/`excludedDataMartIds`) now maps to a generic `permission_denied` that omits the restricted data-mart titles and the caller name/email its message embeds. The full error is still recorded in Run History for operators.

**2. Advertised aggregations off by default + opaque rejection** — fixed. `AGGREGATION_FUNCTION_NOT_ALLOWED_FOR_FIELD`, `..._FOR_TYPE`, and `DUPLICATE_AGGREGATION` now map to a structured `aggregation_not_allowed` error naming `FN(field)` and telling the model to pick another function or ask an admin (no schema re-fetch). The tool description and public docs now also note that per-field output controls decide which functions are available.

**3. Existence oracle (hidden vs missing DM distinguishable by message)** — fixed. The MCP query path now catches the `getByIdAndProjectId` NotFound and rethrows the same constant `Data Mart not found` the hidden-DM path uses, so the id/projectId no longer leak whether a DM exists.

**Minors also fixed:** the `Logger.warn(msg, {error})` misuse on the degraded paths (message interpolated so the error is actually logged); `cell()` now JSON-encodes RECORD/STRUCT/ARRAY instead of `[object Object]`.

**Deliberately deferred (non-blocking), with rationale:**
- MCP filter `value` `z.unknown()` bypassing `ScalarValueSchema` — params are bound, and a bad value now yields the sanitized `query_failed` + a FAILED run rather than a leak; a clean `invalid_input` is a nice-to-have follow-up.
- Soft 128 KiB cap on a single oversized first row — bounded to one row; documenting the caveat rather than changing cap semantics here.
- Discovery-vs-query disagreement on **excluded** (vs inaccessible) sources — the query path matches the pre-existing report-run behaviour and is not a cross-user leak (caller holds USE access); aligning the two is a separate governance decision.
- `metadata.truncated` not reflecting the byte-cap — audit fidelity only; the client response is correct.
- Billing ↔ Run-History independent best-effort writes — by design; flagged for the billing team to reconcile.
- Micro items (header-row escaping asymmetry, `allowedAggregations:[]` vs unrestricted, the enum advertising the 4 always-rejected operators / 12-vs-10 agg values, direct-facade `limit<=0`) — noted, not blocking.

`nest build` + eslint clean; affected jest suites green (added/updated tests for each fix).

zapolsky:
Follow-up on deferred minor #3 (billing ↔ Run-History divergence): closed the worst half in `6115e9e6c`.

Consumption is now gated on the Run History write — a run is billed only when `recordMcpQueryRun(SUCCESS)` persisted. If the audit write fails, billing is suppressed (with a warn), so we never over-charge a user for a run that has no Run History record and whose consumption `reportRunId` would resolve to nothing. A consumption failure still never fails an already-successful read.

The reverse case (recorded but not billed) intentionally stays best-effort — full atomicity across the DB row and the PubSub consumption event is a billing-subsystem outbox/reconciliation concern, not this tool.

Test added (`suppresses billing when the SUCCESS audit save fails`): the read still resolves, but `registerMcpQueryRunConsumption` is not called. `nest build` + eslint clean; service spec green.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -0,0 +1,250 @@
...
+        executionSqlQuery = this.composer.inlineStaticSql(
+          dataMart.storage.type,
+          composed.sql,
+          composed.params
+        );
+      } catch {
+        executionSqlQuery = composed.sql;
+      }
+      reader = await this.readerResolver.resolve(dataMart.storage.type);
+      const description = await reader.prepareReportData(readPlan, {
```
laskevych, line 146: 🟡 **Important:** `prepareReportData` is called without `rowCount: false`, so every aggregated query gets a synthetic `Row Count` column appended to `columns` and to each row (here via the composed `COUNT(*)`, and in `resolveReportDataHeaders`). It's not in the tool description or `docs/mcp.md`'s Returns table, and it contradicts the tool's own "request only the fields relevant… never request all fields" guidance — the model may treat the count as data, echo it, or waste a retry thinking the shape is wrong. Pass `rowCount: false` to suppress it, or document the column. Note `totals` already opts out (`rowCount: false`), so the two blocks are inconsistent today.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -0,0 +1,250 @@
...
+        sqlOverrideParams: composed.params,
+        columnFilter: r.fields,
+        aggregationConfig: readPlan.aggregationConfig ?? undefined,
+      });
+      const columns = description.dataHeaders.map(header => header.name);
+
+      const rows: unknown[][] = [];
+      let batchId: string | undefined;
+      do {
+        const batch = await reader.readReportDataBatch(batchId, overReadLimit - rows.length);
```
laskevych, line 157: 🟡 **Important:** This loop exits only on a falsy `nextDataBatchId` or on reaching `overReadLimit`. If a reader ever returns `dataRows: []` with a truthy `nextDataBatchId`, `rows.length` never advances and the loop spins forever, hanging the request thread (never returns, never records a run). Reachable on Redshift, whose reader ignores the `maxDataRows` arg (`readReportDataBatch(batchId?, _maxDataRows = 1000)`) and forwards `results.NextToken` verbatim — so termination relies entirely on the token going empty (Athena forwards `NextToken` raw too). Please add a guard, e.g. `if (batch.dataRows.length === 0) break;`.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
```diff
@@ -65,9 +76,66 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
...
+        { userId: request.userId, roles: request.roles }
+      );
+
+      // Expose only fields from included sources the caller may report on — mirror the report
+      // UI's gate (isIncluded + isAccessibleForReporting). computeBlendableSchema resolves access
+      // per source but leaves it on availableSources; without this filter we would leak the
+      // schema of joined data marts the caller has no reporting access to.
+      const accessiblePaths = new Set(
+        blendable.availableSources
+          .filter(s => s.isIncluded && s.isAccessibleForReporting)
```
laskevych, line 113: 🔵 **Question:** Discovery hides a join source unless `isIncluded && isAccessibleForReporting` (this filter). On the query path, `assertAllRequestedSourcesAccessible` (`blended-report-data.service.ts`) rejects an `isIncluded === false` source **only for pre-join slices** — and its own comment notes the run path's field index still resolves excluded (`isIncluded:false`) fields. What stops a caller from naming such an excluded source's `<alias>__<field>` directly in `fields`/`filters` (not `slices`)? If that source is `isAccessibleForReporting: true` for the caller's role, the reporting-access check passes and the `isIncluded` guard doesn't cover this path — so a field discovery deliberately hides could still be queryable if the model guessed/cached the name. Could you confirm this is rejected elsewhere (column-config validation) and add a test for "excluded-but-role-accessible source referenced via `fields`", mirroring the existing pre-join-slice defence?

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -0,0 +1,261 @@
...
+      err instanceof BusinessViolationException &&
+      (err.errorDetails?.['deniedDataMartIds'] || err.errorDetails?.['excludedDataMartIds'])
+    ) {
+      return toStructuredToolError(
+        'permission_denied',
+        'This query references one or more data marts you do not have reporting access to. Remove the joined/blended field(s) you cannot access, or ask an admin to grant access.'
+      );
+    }
+
+    if (err instanceof BadRequestException) {
```
laskevych, line 198: 💡 **Suggestion:** `compose()` throws `BadRequestException` with code `PRE_JOIN_FILTERS_REQUIRE_JOINED_DATA_MART` when `slices` are used on a data mart with no joins, but this branch only special-cases `FILTER_COLUMN_UNKNOWN`, the `*_NOT_SELECTED` codes, and the aggregation codes — so it falls through to the generic `query_failed` ("verify the field names, filters, and aggregations…"). That sends the model re-checking the schema instead of learning the real rule. Consider mapping this code to a dedicated `error_code` with actionable guidance (e.g. "move these predicates to `filters`"), the same way `*_NOT_SELECTED` is handled.

apps/backend/src/ee/mcp/tools/query-data-mart.input.ts
```diff
@@ -0,0 +1,233 @@
...
+
+const McpFilterSchema = z.object({
+  field: z.string().min(1),
+  operator: McpOperatorEnum,
+  value: z.unknown().optional(),
+});
+
+const McpAggregationSchema = z.object({
+  field: z.string().min(1),
+  function: z.enum(REPORT_AGGREGATE_FUNCTIONS),
```
laskevych: 💡 **Suggestion:** `z.enum(REPORT_AGGREGATE_FUNCTIONS)` accepts `STRING_AGG` and `ANY_VALUE`, but the tool description and `docs/mcp.md` advertise only SUM/COUNT/COUNT_DISTINCT/AVG/MIN/MAX/P25–P95. The accepted input set and the documented set have drifted — these two are structurally accepted (and will run if a field's output controls allow them) yet documented nowhere. Either add them to the description/docs, or restrict the MCP aggregation enum to the advertised functions so the contract and the schema agree.

apps/backend/src/ee/mcp/tools/tabular-serializer.ts
```diff
@@ -0,0 +1,39 @@
...
+export const ROWS_PAYLOAD_BYTE_CAP = 131072; // 128 KiB — safety ceiling for the serialized rows payload.
+// Comfortably above a max-limit (1000) narrow result (~56 KB), so normal queries pass untouched;
+// only pathological wide-value payloads get trimmed (then `truncated: true`).
+
+function cell(v: unknown): string {
+  if (v === null || v === undefined) return '';
+  // RECORD/STRUCT/ARRAY warehouse cells are objects — String() would emit "[object Object]".
+  // JSON-encode them so the LLM sees the actual value. Date keeps its String() form to preserve
+  // the existing timestamp rendering.
+  const s = typeof v === 'object' && !(v instanceof Date) ? JSON.stringify(v) : String(v);
```
laskevych: 🔵 **Question:** `JSON.stringify` throws on a `BigInt` (or a circular ref) nested inside a RECORD/STRUCT/ARRAY cell. A top-level BigInt is safe (`String()`), but nested struct fields pass through as raw objects. If that's reachable from any reader, the throw happens *after* the query ran and the run was billed, and `mapError` surfaces it as the misleading generic `query_failed`. Is a nested BigInt possible from the readers here? If so, consider a `JSON.stringify` replacer / safe-stringify so a serialization edge doesn't turn a successful, billed read into a field-name error.

apps/backend/src/data-marts/dto/schemas/mcp-query-run-metadata.schema.ts
```diff
@@ -0,0 +1,21 @@
...
+  executionSqlQuery: z.string().optional(),
+  filterCount: z.number().int().nonnegative().optional(),
+  aggregationCount: z.number().int().nonnegative().optional(),
+  query: z
+    .object({
+      fields: z.array(z.string()),
+      filters: z.array(z.unknown()).optional(),
+      aggregations: z.array(z.unknown()).optional(),
+      dateBuckets: z.array(z.unknown()).optional(),
+      limit: z.number().int().positive(),
```
laskevych, line 16: 💡 **Suggestion (defense-in-depth):** `query.limit` requires `.positive()`. The tool input schema enforces `limit ≥ 1`, so this is safe through a real MCP client, but the facade contract (`McpQueryDataMartRequest.limit: number`) and the service don't re-guard it. If any direct facade caller ever passes `limit: 0`, the query still runs and returns rows, but `recordMcpQueryRun`'s `McpQueryRunMetadataSchema.parse` throws here and the best-effort catch swallows it — silently dropping the Run History record **and** suppressing billing for a successful read. Validate `limit ≥ 1` at the service boundary, or relax this to `nonnegative()`.

laskevych (commented):
Code review of `query_data_mart` + blended-field discovery. 2 correctness concerns, 2 questions, 3 suggestions — details inline. Verified clean: filter `value` is re-validated via `FilterConfigSchema.safeParse` before SQL binding; `dataMart.storage` is eagerly loaded; not-found normalization hides id/projectId; totals correctly ignore the row limit; all 22 MCP operators map exactly once.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
zapolsky: Documented rather than suppressed (kept the per-group count intentionally) — fixed in `953184518`. Added a note to the tool description's "Using the results" section and to `docs/mcp.md`'s Returns table: the `Row Count` column is appended on aggregated queries and is grouping metadata, not a requested field.

Left it on because it mirrors the report run path (per-group `COUNT(*)`). The `totals` block still opts out (`rowCount: false`) on purpose — totals is a grand total, so a per-group count there would be meaningless; the two blocks answer different questions, so the asymmetry is intended.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
zapolsky: Fixed in `953184518`. Added `if (batch.dataRows.length === 0) break;` after each page, so an empty page with a truthy `nextDataBatchId` (Redshift/Athena forward the warehouse `NextToken` verbatim) can no longer spin the loop forever. Test added: a reader that always returns `([], 'token')` now resolves with 0 rows after a single call instead of hanging.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
zapolsky: Traced the shared code — the regular **report run path behaves identically**, so this isn't MCP-specific:

- `RunReportService.executeReport` → the same `reportSqlComposerService.compose` → the same `assertAllRequestedSourcesAccessible` the MCP tool uses.
- The output-controls validator checks `isIncluded` **only for pre-join filters** (`FILTER_ALIAS_PATH_NOT_INCLUDED`, `validateFilters`). `columnConfig` (fields) and post-join filters are **not** checked against `isIncluded` on either path.
- `isIncluded` is a **per-data-mart blend-config** flag (`BlendedSource.isExcluded`), not a per-user access gate. The actual role gate — `isAccessibleForReporting` — **is** enforced on both paths, so this is **not** a cross-user leak.
- The FE hides excluded fields in both the report builder picker and MCP discovery — a **UI-level** gate, not backend-enforced.

So MCP mirrors the platform exactly. Adding an MCP-only exclusion check on `fields`/post-join would make MCP stricter than reports and touch shared blended-report code. I'd rather keep them consistent and treat real exclusion enforcement as a **platform-wide decision** for the blended-reporting owners (one ticket covering reports + MCP together). Happy to file that ticket and/or add a test documenting the current behavior — let me know which you'd prefer.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Fixed in `953184518`. `PRE_JOIN_FILTERS_REQUIRE_JOINED_DATA_MART` now maps to a dedicated `slices_not_applicable` error ("this data mart has no joined/blended sources, so slices don't apply — move these predicates to filters") instead of falling through to the generic `query_failed`. Same actionable-error pattern as the `*_NOT_SELECTED` codes. Test added.

apps/backend/src/ee/mcp/tools/query-data-mart.input.ts
zapolsky: Fixed in `953184518` by restricting the schema (option 2). The MCP aggregation enum is now `MCP_AGGREGATE_FUNCTIONS` — the 10 advertised functions only (`STRING_AGG` and `ANY_VALUE` dropped) — declared `as const satisfies readonly ReportAggregateFunction[]`, so it's a compile-time-checked subset of `REPORT_AGGREGATE_FUNCTIONS` and can't silently drift. `STRING_AGG`/`ANY_VALUE` now reject at parse → `invalid_input`. Schema and docs now agree. Test added.

apps/backend/src/ee/mcp/tools/tabular-serializer.ts
zapolsky: Good catch — that's an edge my `[object Object]` fix introduced. Fixed in `953184518`: object/array cells now go through `stringifyObjectCell`, which encodes `BigInt` as its decimal string via a `JSON.stringify` replacer and falls back to `String(v)` if stringify still throws (circular ref). So a nested `BigInt` (e.g. BigQuery INT64 inside a RECORD) or a circular struct can no longer turn a successful, already-billed read into a misleading `query_failed`. Test added for a nested BigInt in a STRUCT and in an ARRAY.

apps/backend/src/data-marts/dto/schemas/mcp-query-run-metadata.schema.ts
zapolsky: Fixed in `953184518`. Added a `Number.isInteger(limit) && limit >= 1` guard at the service boundary, before any read or billing, so a direct facade caller passing `0` gets a clean `BadRequestException` instead of reading and then silently dropping the Run History record when `McpQueryRunMetadataSchema.parse` throws. (Combined with the billing gate from `6115e9e6c`, a dropped record already suppresses billing — so no untraceable charge either.) Test asserts no read, no run record, and no consumption on `limit: 0`.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -0,0 +1,262 @@
...
+    // facade contract types it as a bare number. A limit < 1 would still read and then fail to
+    // persist the run (McpQueryRunMetadataSchema requires limit.positive()), silently dropping the
+    // audit row inside the best-effort catch. Reject it up front, before any read or billing.
+    if (!Number.isInteger(r.limit) || r.limit < 1) {
+      throw new BadRequestException('query_data_mart: limit must be an integer >= 1');
+    }
+
+    let dataMart: DataMart;
+    try {
+      dataMart = await this.dataMartService.getByIdAndProjectId(r.dataMartId, r.projectId);
```
romandubovyi, line 72: 🔴 **Blocking:** This path does not require the data mart to be `PUBLISHED`, so a visible draft/unpublished mart can be queried directly by ID. Please enforce the same published-state boundary as `StreamHttpDataService` before executing the warehouse query.

apps/backend/src/ee/mcp/tools/tabular-serializer.ts
```diff
@@ -0,0 +1,51 @@
...
+  maxBytes: number
+): { tsv: string; rowCount: number; capped: boolean } {
+  const header = columns.join('\t');
+  let total = Buffer.byteLength(header, 'utf8');
+  const lines = [header];
+  let count = 0;
+  for (const r of rows) {
+    const line = r.map(cell).join('\t');
+    const add = Buffer.byteLength('\n' + line, 'utf8');
+    if (count > 0 && total + add > maxBytes) break; // always keep at least the first row
```
romandubovyi: 🟡 **Important:** The byte cap is bypassed for the first row, so a single large warehouse value can produce an arbitrarily oversized MCP response. Please truncate or omit an oversized first row so the advertised 128 KiB safety ceiling is actually enforced.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -0,0 +1,272 @@
...
+} from './query-data-mart.input';
+import { serializeTsvWithByteCap, ROWS_PAYLOAD_BYTE_CAP } from './tabular-serializer';
+import { toStructuredToolError } from '../mappers/mcp-error.mapper';
+
+@Injectable()
+export class QueryDataMartTool implements McpToolDefinition<QueryDataMartInput> {
+  readonly name = 'query_data_mart';
+  readonly description = `Query an OWOX data mart and return its data rows in a compact, header-once table, plus server-side totals computed over all matching rows (ignoring the row limit). Each call costs credits.
+
+Call get_data_mart_details_by_id first to get the data mart's exact field names, joinable/blended fields, and sample values, then copy field names verbatim into fields — unless you already have that schema in context. Field names must be exact; never guess or invent them.
```
romandubovyi: 💡 **Suggestion:** `get_data_mart_details_by_id` does not return sample values; its documentation explicitly says so. Please remove “and sample values” to keep the tool instructions consistent with the actual response contract.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
zapolsky: Added the documenting test in `33fb16770` — `blended-report-data.service.spec.ts`, in the `access denial` block right next to the existing pre-join-slice rejection: **"does NOT reject an EXCLUDED-but-accessible source referenced via a POST-join filter (documents current behavior)"**.

It asserts that `resolveBlendingDecision` builds the blend (`needsBlending: true`, `buildBlendedQuery` called) for a source that is `isIncluded: false` but `isAccessibleForReporting: true`, referenced via a post-join filter — the counterpart to the pre-join-slice test that rejects. Since both the report run path and the MCP tool go through this same method, it locks the shared behavior and makes the pre-join-vs-post-join asymmetry explicit. The role gate (`isAccessibleForReporting`) is still enforced.

Left the real exclusion-enforcement decision for a platform-wide follow-up (reports + MCP together) rather than adding an MCP-only guard here.

apps/backend/src/ee/mcp/tools/tabular-serializer.ts
romandubovyi: 💡 **Suggestion:** Since the newline character `\n` is always 1 byte in UTF-8, we can avoid string allocation on each iteration by using `Buffer.byteLength(line, 'utf8') + 1` instead of `Buffer.byteLength('\n' + line, 'utf8')`.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -0,0 +1,262 @@
...
+  ) {}
+
+  async run(command: QueryDataMartCommand): Promise<McpQueryDataMartResponse> {
+    const r = command.request;
+
+    // Defense-in-depth for direct facade callers: the MCP tool clamps limit to 1–1000, but the
+    // facade contract types it as a bare number. A limit < 1 would still read and then fail to
+    // persist the run (McpQueryRunMetadataSchema requires limit.positive()), silently dropping the
+    // audit row inside the best-effort catch. Reject it up front, before any read or billing.
+    if (!Number.isInteger(r.limit) || r.limit < 1) {
```
romandubovyi: 💡 **Suggestion:** While the input schema of the MCP tool enforces a maximum limit of 1000, it would be safer to add a corresponding upper bound validation check directly in the `QueryDataMartService.run` method to prevent potential memory overload or DB abuse from direct service or facade invocations.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
zapolsky: Fixed in `be83755d3`. Added the same PUBLISHED boundary `StreamHttpDataService` uses — the query path now throws right after loading the mart:

```ts
if (dataMart.status !== DataMartStatus.PUBLISHED) {
  throw new NotFoundException('Data Mart not found');
}
```

A visible DRAFT can no longer be executed by ID. It reuses the constant `Data Mart not found` message (same as the hidden/missing paths) so an unpublished mart stays indistinguishable from a missing one, and it's checked before the access gate and before any warehouse read. Test added (DRAFT → not-found, no access check, no read). Also had to add `status: PUBLISHED` to the service-spec's mock mart.

apps/backend/src/ee/mcp/tools/tabular-serializer.ts
zapolsky: Fixed in `be83755d3`. The cap is now a hard ceiling — dropped the "always keep the first row" exception, so a row that alone exceeds `maxBytes` is omitted (not emitted in full) and the result is flagged `capped` → the tool returns `truncated: true` with `returned_rows: 0`, which tells the model to narrow the query (fewer fields). Chose omit over mid-value truncation so the emitted TSV stays structurally valid. Updated the old "always keep first row" test to assert `rowCount: 0` + header-only, and added a wide-first-row test.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Good catch — fixed in `be83755d3`. Dropped "and sample values" from the description; it now reads "get the data mart's exact field names and joinable/blended fields". Matches `get_data_mart_details_by_id`'s own contract, which explicitly states it does not return sample values.

apps/backend/src/ee/mcp/tools/tabular-serializer.ts
zapolsky: Applied in `be83755d3` — now `Buffer.byteLength(line, 'utf8') + 1` (the `\n` separator is always 1 byte in UTF-8), so we no longer allocate a `'\n' + line` string per row.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
zapolsky: Added in `be83755d3`. Introduced `MAX_QUERY_LIMIT = 1000` in the service and folded an upper bound into the same boundary guard, so a direct service/facade caller can't request an unbounded row count:

```ts
if (!Number.isInteger(r.limit) || r.limit < 1 || r.limit > MAX_QUERY_LIMIT) {
  throw new BadRequestException(
    `query_data_mart: limit must be an integer between 1 and ${MAX_QUERY_LIMIT}`
  );
}
```

Pairs with the lower-bound (`>= 1`) guard from the previous round. Test added (limit 1001 → BadRequest, no read). Kept the constant local to the service rather than importing the tool-layer `MAX_LIMIT` to avoid a use-case → ee/mcp dependency; the two independently agree on 1000.

laskevych (approved):

PR #1385, opened 2026-07-03
feat(mcp): add add_report MCP tool

laskevych:
## Summary

Adds an `add_report` MCP tool (PRD §15.9, task #6700): an AI assistant connected over MCP can create a report that exports a data mart to a **Google Sheets** destination. A new Google Sheet is created automatically and the report is linked to it — the caller gets ready-to-open `report_url` and `sheet_url` back, without leaving the chat.

First **write** tool on the MCP surface (`mcp:write` scope, `readOnlyHint: false`). Free per the PRD billing model.

> **Stacked on #1377** (`get_data_mart_reports`) — it extends the same `McpReportsFacade`. GitHub will retarget this PR to `main` once #1377 merges.

## What changed

- **New MCP tool** `AddReportTool` (`ee/mcp/tools/add-report.tool.ts`) — input `{ data_mart_id, destination_id, fields, name }` (`fields: ['*']` = all columns; `name` is trimmed); output `{ report_id, report_url, sheet_url, owner, status: 'created', placed_in_root?, shared_with_requester? }`.
- **`McpReportsFacade.addReport`** orchestrates the existing domain services — nothing reimplemented:
  1. **Pre-flight validation** (see below),
  2. `CreateGoogleSheetDocumentService` mints the Sheet (synchronous; also rejects non-Google-Sheets destinations),
  3. `CreateReportService` creates the report pointing at it (`destinationConfig` built from the fresh `spreadsheetId`/`sheetId`).
- **Reports UI-path helper** `buildReportsUiPath` next to the existing data-mart path helper.

## Design notes

- **Validation runs BEFORE the external side effect.** Sheet creation is not transactional, so the facade pre-validates everything `CreateReportService` would reject — data mart `PUBLISHED`, `USE` access to the data mart **and** the destination, and the `fields` selection (`OutputControlsValidatorService`) — before any Google API call. Without this, a DRAFT data mart, a permissions miss, or a typo'd field name would orphan a real document in the destination owner's Drive (and MCP clients retry on error, multiplying orphans); an unauthorized caller could trigger file creation at all. This mirrors the REST path, where the controller enforces `Action.USE` before `createGoogleSheetDocument` ([data-destination.controller.ts:334](../blob/main/apps/backend/src/data-marts/controllers/data-destination.controller.ts)). The service still re-validates inside its transaction; the duplication is deliberate and commented.
- **`sheet_url` is synchronous** because the tool itself mints the sheet first — the `report.created` listener only stamps developer metadata and never produces the document.
- **Sharing/placement warnings are surfaced, not swallowed:** `placed_in_root` (configured Drive folder could not be used) and `shared_with_requester` (sheet could not be shared with the caller) pass through to the tool output with LLM-facing descriptions, so the assistant can warn the user instead of handing over a link that shows "Request access" — same signals the web UI warns about.
- **Google-Sheets-only v1** (product decision): other destination types are rejected with a clear error by the document-creation service. `slices`/`filters` and `description` from the PRD input are **deferred**: the PRD filter vocabulary doesn't map 1:1 onto the domain filter model yet, and half-mapping them would create reports that silently return wrong rows; `description` has no field on the Report entity.
- `owner` defaults to the requesting user (service default); `report_url` points at the data mart's reports tab (no per-report route exists in the web app).

## Review

A max-effort multi-agent code review ran on this branch (10 finder angles → adversarial verification → gap sweep). All confirmed findings are addressed in this PR: validation-before-side-effect ordering, destination `USE` parity with REST, surfaced sharing/placement flags, trimmed report name. One pre-existing perf item (eager `DataMart` hydration in `ScheduledTriggerService` queries) is tracked separately.

## Testing

- Facade: pre-validation rejection matrix (DRAFT mart / no mart access / no destination access / invalid fields — each asserting the sheet is **not** created), happy path incl. check→sheet→report ordering and command payloads, `['*']` → no column projection, flags passthrough, sheet-failure short-circuit, owner fallback.
- Tool: output shape incl. warning flags, strict input validation, name trimming and whitespace-only rejection, `mcp:write` scope + annotations, registry.
- `npx jest src/data-marts/facades src/ee/mcp/tools` — 51 tests, 11 suites green; `nest build` and eslint clean.
- Live-verified earlier in Claude Desktop via `mcp-remote` against a local `owox-better-auth` server (pre-fix version of the flow; the GS auto-creation path is the same one exercised by the web "Create document" button).

## Out of scope / follow-ups

- `update_report`, `delete_report` MCP tools (tracked separately).
- `slices`/`filters` input support — shared design with the future `query_data_mart` filter vocabulary.
- Lean trigger query (perf) in `ScheduledTriggerService`.


apps/backend/src/ee/mcp/tools/add-report.tool.ts
```diff
@@ -0,0 +1,108 @@
...
+      .optional()
+      .describe(
+        'False when the sheet could not be shared with you; opening the link may require requesting access.'
+      ),
+  };
+  readonly annotations = {
+    title: 'Add Report',
+    readOnlyHint: false,
+    destructiveHint: false,
+    openWorldHint: false,
```
kalnik-a-a: Blocking: this tool creates a Google Sheet in Drive/Sheets and may share it with the requester, so `openWorldHint: false` misdescribes the tool contract to MCP clients. Please set `openWorldHint: true` and update the spec before publishing this tool.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
```diff
@@ -0,0 +1,108 @@
...
+        buildReportsUiPath(context.projectId, data_mart_id)
+      ),
+      sheet_url: result.sheet_url,
+      owner: result.owner,
+      status: result.status,
+      placed_in_root: result.placed_in_root,
+      shared_with_requester: result.shared_with_requester,
+    };
+
+    return {
```
kalnik-a-a: Non-blocking: this response is the standard small JSON MCP result shape. Please use the shared `jsonToolResult(structuredContent)` helper instead of rebuilding the same `structuredContent`/`content` wrapper manually, so future formatting changes stay centralized.

kalnik-a-a (changes requested):
I found one blocking MCP annotation issue and one non-blocking formatter cleanup. Details inline.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
laskevych: Agreed — fixed in c9e659f40. `openWorldHint` is now `true` with a code comment explaining why: unlike the other tools (which stay inside the OWOX domain and correctly keep `false`), `add_report` creates a document in Google Drive and may share it with the requester. Spec assertion updated. Note for the stack: `update_report`/`delete_report` (#1382/#1383) stay `false` — they perform no external writes (deletion cleanup is an async internal event).

apps/backend/src/ee/mcp/tools/add-report.tool.ts
laskevych: Done in c9e659f40 — switched to the shared `jsonToolResult(structuredContent)` helper (it produces the identical shape, so the spec assertions held). I will apply the same cleanup to the stacked `update_report`/`delete_report` tools when they get rebased over this PR.

kalnik-a-a (approved):

PR #1390, opened 2026-07-05
feat(web): move report quick actions to the title column

artemfenoga:
# Improved the reports table UX by making frequently used actions easier to discover and access

- Moved **Open document** and **Preview SQL** actions next to the report title.
- Kept the **More actions** menu in the actions column.
- Added tooltips for truncated report titles.
- Extracted reusable report action components for better consistency across report tables.

This update improves discoverability and provides faster access to common report actions while keeping the table layout clean and consistent.

Recording 2026-07-05 204824

### DIFF
--- a/apps/web/src/features/data-marts/reports/list/components/EmailReportsTable/EmailReportTitleCell.tsx
+++ b/apps/web/src/features/data-marts/reports/list/components/EmailReportsTable/EmailReportTitleCell.tsx
@@ -1,0 +1,21 @@
+import type { DataMartReport } from '../../../shared/model/types/data-mart-report';
+import {
+  ReportGeneratedSqlAction,
+  ReportTitleCell,
+  reportTitleCellQuickActionClassName,
+} from '../../../shared/components';
+
+interface EmailReportTitleCellProps {
+  report: DataMartReport;
+}
+
+export function EmailReportTitleCell({ report }: EmailReportTitleCellProps) {
+  return (
+    <ReportTitleCell
+      title={report.title}
+      actions={
+        <ReportGeneratedSqlAction report={report} className={reportTitleCellQuickActionClassName} />
+      }
+    />
+  );
+}
--- a/apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
+++ b/apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
@@ -1,5 +1,5 @@
 import { useState, useCallback, useEffect } from 'react';
-import { MoreHorizontal, Pencil, FileText, Trash2, Play } from 'lucide-react';
+import { MoreHorizontal, Pencil, Trash2, Play } from 'lucide-react';
 import { Button } from '@owox/ui/components/button';
 import {
   DropdownMenu,
@@ -8,20 +8,9 @@ import {
   DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from '@owox/ui/components/dropdown-menu';
-import {
-  Tooltip,
-  TooltipContent,
-  TooltipProvider,
-  TooltipTrigger,
-} from '@owox/ui/components/tooltip';
 import { ConfirmationDialog } from '../../../../../../shared/components/ConfirmationDialog';
-import { getGoogleSheetTabUrl, isGeneratedSqlSupported } from '../../../shared';
-import type {
-  DataMartReport,
-  GoogleSheetsDestinationConfig,
-} from '../../../shared/model/types/data-mart-report';
+import type { DataMartReport } from '../../../shared/model/types/data-mart-report';
 import { useReport, ReportStatusEnum } from '../../../shared';
-import { GeneratedSqlViewer } from '../../../../edit/components/ReportColumnPicker/GeneratedSqlViewer';
 
 interface GoogleSheetsActionsCellProps {
   row: { original: DataMartReport };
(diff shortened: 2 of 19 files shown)


apps/web/src/features/data-marts/reports/list/components/EmailReportsTable/EmailReportTitleCell.tsx
```diff
@@ -0,0 +1,26 @@
+import type { DataMartReport } from '../../../shared/model/types/data-mart-report';
+import { ReportGeneratedSqlAction, ReportTitle } from '../../../shared/components';
+
+interface EmailReportTitleCellProps {
+  report: DataMartReport;
+}
+
+export function EmailReportTitleCell({ report }: EmailReportTitleCellProps) {
+  const quickActionClassName =
+    'opacity-0 transition-opacity hover:opacity-100 group-hover:opacity-100';
```
romandubovyi: 🟡 **Important:** `GoogleSheetsReportTitleCell` uses `pointer-events-none` + `group-hover:pointer-events-auto` so invisible quick actions don't steal row clicks (`onRowClick` opens edit). Please apply the same pattern here and in `ProjectReportTitleCell` (`DataMartReportsPage.tsx:129`) for consistent behavior across all clickable report tables.

apps/web/src/features/data-marts/reports/shared/components/index.ts
```diff
@@ -1,2 +1,5 @@
 export * from './ReportHoverCard';
 export * from './ReportQuickRunCell';
+export * from './ReportGeneratedSqlAction';
```
romandubovyi, line 3: 🟡 **Important:** These new shared components (`ReportTitle`, `ReportOpenDocumentAction`, `ReportGeneratedSqlAction`) don't have unit tests yet. Consider adding coverage for conditional SQL rendering, Google Sheets URL building, and tooltip/aria behavior — similar to `ReportQuickRunCell.test.tsx`.

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
```diff
@@ -120,6 +125,46 @@ function buildProjectReportFilters(
   ];
 }
 
+function ProjectReportTitleCell({ report }: { report: DataMartReport }) {
+  const quickActionClassName = 'opacity-0 transition-opacity group-hover:opacity-100';
```
romandubovyi: 🟡 **Important:** These hidden quick actions still receive pointer events while they are only `opacity-0`. The Google Sheets title cell already uses `pointer-events-none` until row hover; please apply the same pattern here so invisible actions cannot intercept clicks or focus.

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
```diff
@@ -120,6 +125,46 @@ function buildProjectReportFilters(
   ];
 }
 
+function ProjectReportTitleCell({ report }: { report: DataMartReport }) {
```
romandubovyi, line 129: 💡 **Suggestion:** `EmailReportTitleCell`, `GoogleSheetsReportTitleCell`, and this inline `ProjectReportTitleCell` share nearly identical layout/behavior. Consider extracting a single shared `ReportTitleWithActionsCell` that accepts destination-specific quick actions to reduce duplication and keep hover/pointer-events behavior consistent.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export function ReportOpenDocumentAction({ report, className }: ReportOpenDocumentActionProps) {
+  const destinationConfig = report.destinationConfig as GoogleSheetsDestinationConfig;
+
+  return (
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger asChild>
+          <a
```
romandubovyi, line 41: 💡 **Suggestion:** This link only has an icon with `aria-hidden`, so it has no accessible name for screen readers. Consider adding `aria-label={`Open document: ${report.title}`}` (matching the pattern used in `GoogleSheetsReportEditForm`).

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
```diff
@@ -102,124 +91,87 @@ export function GoogleSheetsActionsCell({
   }, [canEditConfig]);
 
   return (
-    <TooltipProvider>
-      <div
-        className='flex justify-end gap-1'
-        onClick={e => {
-          e.stopPropagation();
+    <div
+      className='flex justify-end gap-1'
```
romandubovyi: ⚪ **Nit:** `gap-1` is no longer needed now that this cell only renders the "More actions" dropdown — can be removed for consistency with `EmailActionsCell`.

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
```diff
@@ -308,7 +353,7 @@ export default function DataMartReportsPage() {
           if (!title) {
```
romandubovyi, line 346: 🟡 **Important:** Please add a focused test for the new title-column quick actions. This page has row-level click handling for editing, so we should verify that clicking `Open document` or `Preview SQL` does not also open the report edit sheet.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
```diff
@@ -0,0 +1,50 @@
...
+  GoogleSheetsDestinationConfig,
+} from '../model/types/data-mart-report';
+
+interface ReportOpenDocumentActionProps {
+  report: DataMartReport;
+  className?: string;
+}
+
+export function ReportOpenDocumentAction({ report, className }: ReportOpenDocumentActionProps) {
+  const destinationConfig = report.destinationConfig as GoogleSheetsDestinationConfig;
```
romandubovyi: 🔴 **Blocking:** This component accepts any `DataMartReport` but unsafely casts `destinationConfig` to `GoogleSheetsDestinationConfig`. If used with a non-Google Sheets report or if `destinationConfig` is missing/ malformed, it will throw at runtime. Guard against this by checking `report.dataDestination.type` and validating `spreadsheetId`/`sheetId` before building the URL.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
romandubovyi, line 41: 🟡 **Important:** This icon-only link lacks an accessible name. The icon is hidden from assistive technologies (`aria-hidden=true`), so screen reader users won't know the link opens the report document. Add an `aria-label` like `Open document: ${report.title}` to match the pattern used in `GeneratedSqlViewer`.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsReportTitleCell.tsx
```diff
@@ -0,0 +1,32 @@
...
+  ReportOpenDocumentAction,
+  ReportTitle,
+} from '../../../shared/components';
+
+interface GoogleSheetsReportTitleCellProps {
+  report: DataMartReport;
+}
+
+export function GoogleSheetsReportTitleCell({ report }: GoogleSheetsReportTitleCellProps) {
+  const quickActionClassName =
```
romandubovyi: 🟡 **Important:** The quick-action hover classes differ between `EmailReportTitleCell` and `GoogleSheetsReportTitleCell` (`pointer-events-none` + `group-hover:pointer-events-auto` here vs. only opacity in Email). This makes the same action pattern behave differently depending on the destination type. Unify the hover/interaction classes so shared actions behave consistently.

apps/web/src/features/data-marts/reports/list/components/EmailReportsTable/EmailReportTitleCell.tsx
```diff
@@ -0,0 +1,26 @@
+import type { DataMartReport } from '../../../shared/model/types/data-mart-report';
+import { ReportGeneratedSqlAction, ReportTitle } from '../../../shared/components';
+
+interface EmailReportTitleCellProps {
+  report: DataMartReport;
+}
+
+export function EmailReportTitleCell({ report }: EmailReportTitleCellProps) {
+  const quickActionClassName =
```
romandubovyi: 🟡 **Important:** The quick actions here lack `pointer-events-none` when they are hidden (`opacity-0`). As a result, these transparent buttons can still intercept click events, leading to accidental triggers when a user tries to interact with the report title or row. 

Please update `quickActionClassName` to match the correct implementation in `GoogleSheetsReportTitleCell.tsx`:
`pointer-events-none opacity-0 transition-opacity group-hover:pointer-events-auto group-hover:opacity-100`

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
```diff
@@ -120,6 +125,46 @@ function buildProjectReportFilters(
   ];
 }
 
+function ProjectReportTitleCell({ report }: { report: DataMartReport }) {
+  const quickActionClassName = 'opacity-0 transition-opacity group-hover:opacity-100';
+
+  let quickActions: ReactNode = null;
+
+  switch (report.dataDestination.type) {
```
romandubovyi, line 132: 💡 **Suggestion:** This `switch` duplicates the destination-type branching already used in `ProjectReportActionsCell` below. Consider extracting a shared helper or component that returns the quick actions for a report so the two places don't drift apart when new destination types are added.

apps/web/src/features/data-marts/reports/shared/components/ReportTitle.tsx
```diff
@@ -0,0 +1,24 @@
...
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from '@owox/ui/components/tooltip';
+
+interface ReportTitleProps {
+  title: string;
+}
+
+export function ReportTitle({ title }: ReportTitleProps) {
```
romandubovyi: 🔵 **Question:** The tooltip will show for every report title, even when the text isn't actually truncated. The PR description mentions tooltips for truncated titles — is this the intended behavior? If not, consider showing the tooltip only when the element's `scrollWidth` exceeds `clientWidth`.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger asChild>
+          <a
+            href={getGoogleSheetTabUrl(destinationConfig.spreadsheetId, destinationConfig.sheetId)}
+            className={cn(
+              'dm-card-table-body-row-actionbtn inline-flex h-6 w-6 items-center justify-center rounded-md',
```
romandubovyi: 💡 **Suggestion:** The class `.dm-card-table-body-row-actionbtn` in `globals.css` defines a default height of `h-8`. Since you are overriding this to `h-6 w-6` here, and adding `w-6` (but not `h-6`) in `GeneratedSqlViewer.tsx`, the action buttons might lose their square proportions (becoming 24x32px) depending on Tailwind's class resolution order.

To prevent styling issues, consider using important modifiers (e.g. `h-6! w-6!`) or explicitly ensuring that height and width values are resolved correctly across these action triggers.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
romandubovyi: 💡 **Suggestion:** Instead of casting `report.destinationConfig as GoogleSheetsDestinationConfig`, we can use the existing type-guard helper function `isGoogleSheetsDestinationConfig` to perform a safe check and avoid potential runtime `TypeError` issues if this component is ever rendered for other types of destinations.

```typescript
import { isGoogleSheetsDestinationConfig } from '../model/types/data-mart-report';

// ...

if (!isGoogleSheetsDestinationConfig(report.destinationConfig)) {
  return null;
}
const destinationConfig = report.destinationConfig;
```

apps/web/src/features/data-marts/reports/list/components/EmailReportsTable/EmailReportTitleCell.tsx
artemfenoga: Done. Applied the same quick action visibility and pointer-events pattern used in GoogleSheetsReportTitleCell to both components.

apps/web/src/features/data-marts/reports/shared/components/index.ts
artemfenoga: Done

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
artemfenoga: Done

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
artemfenoga: Done. Extracted a shared ReportTitleCell and centralized the shared quick action styling, so all report tables now use the same layout and hover/pointer-events behavior.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
artemfenoga: Done. Added an accessible label that includes the report title to match the existing pattern used in the report editor.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
artemfenoga: Done

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
artemfenoga: Added focused tests to verify that clicking Open document and Preview SQL does not open the report edit sheet.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
artemfenoga: Done. Removed the unsafe cast and added guards for the destination type and Google Sheets config before building the URL.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
artemfenoga: Done. Added an accessible label (Open document: ${report.title}) to the link, matching the pattern used by GeneratedSqlViewer.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsReportTitleCell.tsx
artemfenoga: Done. The quick action hover/interaction classes are now shared via reportTitleCellQuickActionClassName, so all report title cells use the same hover and pointer-events behavior.

apps/web/src/features/data-marts/reports/list/components/EmailReportsTable/EmailReportTitleCell.tsx
artemfenoga: Done. Updated the quick actions to use the same pointer-events/hover behavior as the Google Sheets implementation, and centralized the shared styling to keep all report title cells consistent.

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
artemfenoga: Good suggestion. I considered extracting the destination branching, but the title quick actions and the actions column have different responsibilities and are expected to evolve independently. I prefer to keep the branching local to each component to avoid introducing an abstraction that only wraps the switch.

apps/web/src/features/data-marts/reports/shared/components/ReportTitle.tsx
artemfenoga: Good question. At the moment the tooltip is shown consistently for all report titles. I considered rendering it only for truncated text, but that would require additional DOM measurement logic. I preferred to keep the implementation simple in this PR, and I'm happy to revisit conditional tooltips as a separate UX improvement if we decide it's worthwhile.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
artemfenoga: Good catch. The shared action button class was overriding the intended icon button size. I updated both quick action triggers to use !h-6 !w-6, ensuring consistent 24×24px dimensions regardless of CSS precedence.

apps/web/src/features/data-marts/reports/shared/components/ReportOpenDocumentAction.tsx
artemfenoga: Done. Replaced the unsafe cast with isGoogleSheetsDestinationConfig() and added runtime guards for the destination type and required Google Sheets configuration before building the URL.

apps/web/src/features/data-marts/reports/shared/components/ReportTitleCell.tsx
```diff
@@ -0,0 +1,42 @@
+import type { ReactNode } from 'react';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from '@owox/ui/components/tooltip';
+
+export const reportTitleCellQuickActionClassName =
+  'pointer-events-none opacity-0 transition-opacity group-hover:pointer-events-auto group-hover:opacity-100';
```
romandubovyi: 🟡 **Important:** These quick actions become visible only on row hover, so keyboard users can tab to an invisible button/link. Please also reveal and enable them on focus, e.g. with `focus-visible:opacity-100 focus-visible:pointer-events-auto` or a `group-focus-within:*` variant, so the focused control is visible without requiring a mouse.

apps/web/src/features/data-marts/reports/shared/components/__tests__/ReportOpenDocumentAction.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+      type: DestinationTypeConfigEnum.GOOGLE_SHEETS_CONFIG,
+      spreadsheetId: 'spreadsheet-123',
+      sheetId: '456',
+    } satisfies GoogleSheetsDestinationConfig,
+    ...overrides,
+  } as DataMartReport;
+}
+
+describe('ReportOpenDocumentAction', () => {
+  it('renders a link to the Google Sheets document', () => {
```
romandubovyi, line 30: 🟡 **Important:** This only covers the happy path, but the new behavior depends on guard branches. Please add focused coverage that `ReportOpenDocumentAction` renders nothing for non-Google Sheets reports and invalid/missing Google Sheets config, and add a small `ReportGeneratedSqlAction` test for the supported vs unsupported SQL cases.

romandubovyi (approved):
No blockers from my side. The previous review comments have been addressed and resolved, and the remaining notes are non-blocking follow-ups.

Approved — this can be merged once CI is green.

apps/web/src/features/data-marts/reports/shared/components/ReportTitleCell.tsx
artemfenoga: Good catch, thanks! Added group-focus-within so the quick actions are also revealed for keyboard users.

apps/web/src/features/data-marts/reports/shared/components/__tests__/ReportOpenDocumentAction.test.tsx
artemfenoga: Thanks! Added coverage for the guard branches in ReportOpenDocumentAction and the supported/unsupported cases for ReportGeneratedSqlAction.

PR #1391, opened 2026-07-06
feat(mcp): summarize data catalog

amarchenk0:


apps/backend/src/data-marts/use-cases/summarize-mcp-data-catalog.service.ts
```diff
@@ -0,0 +1,133 @@
...
+      if (!visiblePublishedIds.has(sourceId) || !visiblePublishedIds.has(targetId)) continue;
+      if (!relationship.joinConditions || relationship.joinConditions.length === 0) continue;
+      const list = bySource.get(sourceId);
+      if (list) list.push(relationship);
+      else bySource.set(sourceId, [relationship]);
+    }
+    return bySource;
+  }
+
+  private countReachableRelationships(
```
romandubovyi, line 83: 🟡 **Important:** This DFS runs once per data mart and allocates a new `Set` on every recursive step. For projects with many data marts and relationships this can become expensive. Consider caching reachable counts, reusing a single mutable path set with backtracking, or limiting the traversal to the top-N candidates earlier.

apps/backend/src/ee/mcp/tools/summarize-data-catalog.tool.ts
```diff
@@ -0,0 +1,95 @@
...
+    @Inject(MCP_DATA_MARTS_FACADE)
+    private readonly dataMarts: McpDataMartsFacade,
+    private readonly publicOriginService: PublicOriginService
+  ) {}
+
+  parseInput(input: unknown): SummarizeDataCatalogInput {
+    return inputSchema.parse(input);
+  }
+
+  async handler(input: SummarizeDataCatalogInput, context: McpAuthContext): Promise<McpToolResult> {
```
romandubovyi, line 60: 🟡 **Important:** `parseInput` is called both by the MCP framework and again inside `handler`. If the framework already validates before calling `handler`, this double-validation is redundant. Please align with other tools in the same directory and remove or keep it consistently.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
```diff
@@ -136,6 +140,14 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
     return this.queryDataMartService.run(new QueryDataMartCommand(request));
   }
 
+  async summarizeDataCatalog(
```
romandubovyi, line 144: 💡 **Suggestion:** `summarizeDataCatalog` reuses `McpListDataMartsRequest` even though it is a different operation. Consider introducing a dedicated `McpSummarizeDataCatalogRequest` type to keep the facade contract clear and avoid accidentally inheriting list-specific filters or pagination later.

apps/backend/src/data-marts/use-cases/summarize-mcp-data-catalog.service.ts
```diff
@@ -0,0 +1,133 @@
...
+  async run(command: SummarizeMcpDataCatalogCommand): Promise<McpDataCatalogSummaryDto> {
+    const roleScope = await this.getRoleScope(command);
+    const publishedDataMarts = await this.catalogSummaryService.findPublishedVisibleDataMarts({
+      projectId: command.projectId,
+      userId: command.userId,
+      roles: command.roles,
+      roleScope,
+    });
+    const publishedIds = publishedDataMarts.map(item => item.id);
+    const visiblePublishedIds = new Set(publishedIds);
```
romandubovyi, line 35: ⚪ **Nit:** Since `visiblePublishedIds` is passed down to other methods as a `ReadonlySet<string>` and is never modified inside `run`, you can declare it with an explicit type to prevent accidental mutations in the future:
```typescript
const visiblePublishedIds: ReadonlySet<string> = new Set(publishedIds);
```

apps/backend/src/ee/mcp/tools/summarize-data-catalog.tool.ts
```diff
@@ -0,0 +1,95 @@
...
+const inputSchema = z.object({}).strict();
+type SummarizeDataCatalogInput = z.infer<typeof inputSchema>;
+
+const INSTRUCTION =
+  'You have received a high-level summary of the published Data Mart catalog available to this MCP connection. Summarize the business areas covered by the listed Data Marts and suggest 4-6 concrete example prompts the user could ask. Do not claim access to data rows, sample values, row counts, or freshness details.';
+
+@Injectable()
+export class SummarizeDataCatalogTool implements McpToolDefinition<SummarizeDataCatalogInput> {
+  readonly name = 'summarize_data_catalog';
+  readonly description =
```
romandubovyi, line 22: ⚪ **Nit:** This `description` is quite long and mixes tool purpose with usage guidance. Consider keeping the description concise and moving usage hints (when to call, what it does not return) to a shorter summary or to `_instruction`.

apps/backend/src/data-marts/use-cases/summarize-mcp-data-catalog.service.ts
romandubovyi, line 83: 🟡 **Important:** `countReachableRelationships` performs a per-root DFS with no memoization across calls and copies the `path` set on every recursion frame (`new Set([...path, targetId])`). Cycle protection via `path.has(targetId)` prevents infinite loops, but not combinatorial blowup: in a densely interconnected relationship graph (multiple paths between the same nodes), the number of paths explored can grow exponentially with node count. This runs synchronously on every `summarize_data_catalog` MCP call with no depth/size guard or caching. Probably fine at current catalog sizes, but worth a safeguard (e.g. memoize reachable-set per node, or cap traversal) before this scales to larger, more interconnected projects.

apps/backend/src/data-marts/repositories/mcp-data-catalog-summary.repository.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+    expect(dataMartQb.andWhere).toHaveBeenCalledWith(
+      expect.stringContaining('data_mart_contexts'),
+      expect.objectContaining({
+        userId: 'user-1',
+        projectId: 'project-1',
+        roleScope: RoleScope.SELECTED_CONTEXTS,
+      })
+    );
+    expect(dataMartQb).not.toHaveProperty('take');
```
romandubovyi: ⚪ **Nit:** `expect(dataMartQb).not.toHaveProperty('take')` always passes since the mock `createQueryBuilder` never defines a `take` key in the first place — it doesn't actually verify that `.take()` is never called on the real query builder. If the intent is to assert "no pagination applied", consider mocking `take: jest.fn().mockReturnThis()` and asserting `expect(dataMartQb.take).not.toHaveBeenCalled()` instead.

romandubovyi (approved):
No blocking issues found in this review round. All open feedback is 🟡 Important / 💡 Suggestion / ⚪ Nit — worth addressing, but not merge blockers:

- 🟡 Two Important items: DFS traversal cost in `summarize-mcp-data-catalog.service.ts`, and a redundant `parseInput` call in `summarize-data-catalog.tool.ts`.
- 💡 One Suggestion: dedicated request type instead of reusing `McpListDataMartsRequest` for `summarizeDataCatalog`.
- ⚪ Two Nits: explicit type for `visiblePublishedIds`, and trimming the tool's `description`.

Approving — happy to have the Important items addressed in a follow-up if not resolved before merge.

PR #1392, opened 2026-07-06
feat(mcp): add report run tools

kalnik-a-a:
## Summary

This PR adds an MCP-safe start-and-poll flow for report runs. `run_report` starts an existing report run and returns immediately with `report_id` plus `run_id`; `get_report_run_status` lets MCP clients poll that run until it reaches a terminal state.

The new destination guard is intentionally scoped to the MCP `run_report` facade. Existing product REST APIs, report scheduling validation, and web quick-run behavior are left unchanged by this PR.

## What Changed

- Added the `run_report` MCP tool. It starts a manual report run by `report_id`, returns only `report_id` and `run_id`, and does not wait for execution to finish.
- Added the `get_report_run_status` MCP tool. It checks a specific `report_id` / `run_id` pair, returns status (`running`, `success`, `failed`, `cancelled`, `interrupted`, or `restricted`), and exposes `should_poll`, `stop_reason`, timestamps, raw backend status, and readable errors for failed runs.
- Wired both tools through `McpReportsFacade`, `McpReportsFacadeImpl`, provider registration, and the existing report-run service boundary.
- Kept the existing REST manual-run controller fire-and-forget while allowing the shared run-report use case to return the enqueued `dataMartRunId` for MCP.
- Added an MCP-only guard for destinations that should not be actively pushed by `run_report`. Data Studio-style consumers fetch data themselves, so they are rejected by `run_report` before enqueueing a run.
- Made the destination classification explicit so future unsupported destinations are not mislabeled as pull-based.
- Derived the status-tool run-type allowlist from the MCP `run_report` destination allowlist to avoid drift.
- Hardened `get_report_run_status` access checks so report operate access is checked before looking up the requested run or destination details.
- Shared report-run error message parsing between MCP status responses and failed-run notifications.
- Updated polling guidance to poll about every 15 seconds when waiting is possible. The tool no longer describes a 60-second cap or a contradictory exponential backoff.
- Added a longer-than-usual warning for runs claimed by a worker but missing `started_at` after 10 minutes, while keeping the 30-minute stuck threshold.

## Behavior Notes

- `run_report` is intentionally not a status endpoint. A successful response means the run was accepted; callers must use `get_report_run_status` to learn the terminal result.
- `get_report_run_status` keeps polling explicit through `should_poll`. Normal long-running executions continue to return `running`; terminal states return `success`, `failed`, `cancelled`, `interrupted`, or `restricted`.
- Queued runs may still execute later. If a run has been queued or running for an unusually long time, the status tool tells the client whether to keep polling or stop for now and check Run History before starting another run.
- Backend terminal statuses are preserved in `raw_status`; user-cancelled, interrupted, and restricted runs now surface as distinct MCP statuses instead of being flattened into `failed`.
- Existing product flows are not converted to MCP behavior here: REST manual runs, scheduled report-run validation, and web report quick-run UI keep their previous behavior.
- The MCP tool metadata keeps `destructiveHint: false`.

## Local Verification

- `git diff --check`
- `npx prettier --check .changeset/6702-mcp-run-report-tool.md apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts apps/backend/src/data-marts/facades/mcp-reports.facade.ts apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts apps/backend/src/ee/mcp/tools/get-report-run-status.tool.spec.ts apps/backend/src/ee/mcp/tools/run-report.tool.ts apps/backend/src/ee/mcp/tools/run-report.tool.spec.ts`
- `npm run lint -w @owox/backend` (0 errors; one pre-existing warning in `sql-clause-renderer.spec.ts`)
- `npm run test -w @owox/backend -- data-marts/facades/mcp-reports.facade.impl.spec.ts data-marts/services/data-mart-run.service.spec.ts data-marts/use-cases/run-report.service.spec.ts data-marts/utils/report-run-type.spec.ts data-marts/utils/run-error-message.spec.ts ee/mcp/tools/data-mart-catalog.tool.spec.ts ee/mcp/tools/get-report-run-status.tool.spec.ts ee/mcp/tools/mcp-tool.providers.spec.ts ee/mcp/tools/run-report.tool.spec.ts --runInBand`


apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  private elapsedMinutes(timestamp: string | null): number | null {
+    if (!timestamp) {
+      return null;
+    }
+    const parsed = Date.parse(timestamp);
+    if (Number.isNaN(parsed)) {
+      return null;
+    }
+    return Math.max(0, Math.round((Date.now() - parsed) / 60_000));
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in the latest push. `elapsedMinutes()` now uses `SystemTimeService` with `Math.floor`, with regression coverage for threshold behavior.

apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts
```diff
@@ -0,0 +1,163 @@
...
...an wait between tool calls, use exponential backoff — about 15 seconds before the first check, then double the interval up to 15 seconds; if your environment cannot wait between calls, call again immediately instead — an immediate re-check is always better than stopping, and never end your reply just to wait for the run. A "running" status is normal and expected even after many checks in a row: it is not an error and not a reason to stop polling or report a failure (stop early only if the user explicitly asks). If a status check itself errors (timeout, network error, 5xx), that says nothing about the run — retry the check. The message field tells you what to do next, including when to warn the user that a run is taking longer than usual and when to treat it as stuck and stop polling. stop_reason is set only when status is still "running" but should_poll is false, such as queued_too_long or running_too_long. queued_at is when the run was created; started_at is null until execution starts. raw_status exposes the backend terminal state (for example CANCELLED, INTERRUPTED, or RESTRICTED) even though status stays simplified for polling. error is populated only when status is "failed".';
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in the latest push. The tool now says to poll about every 15 seconds and no longer mentions exponential backoff or a 60-second cap; tests pin that wording.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -247,6 +274,59 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+    const report = await this.getReportService.run(
+      new GetReportCommand(request.reportId, request.projectId, request.userId, request.roles)
+    );
+
+    const destinationType = report.dataDestinationAccess.type;
+    if (!isPushBasedDataDestinationType(destinationType)) {
+      throw new BusinessViolationException(
+        `Reports with a ${toHumanReadable(destinationType)} destination are pull-based and cannot be run on demand`
+      );
+    }
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in the latest push: the facade pre-flight guard was removed, and the shared destination label now returns `Looker Studio`.

apps/backend/src/data-marts/services/report-run.service.ts
```diff
@@ -48,6 +53,13 @@ export class ReportRunService {
   async createPending(command: RunReportCommand): Promise<ReportRun | null> {
     const report = await this.reportService.getById(command.reportId);
 
+    const destinationType = report.dataDestination.type;
+    if (!isPushBasedDataDestinationType(destinationType)) {
+      throw new BusinessViolationException(
+        `Reports with a ${toHumanReadable(destinationType)} destination are pull-based and cannot be run on demand`
+      );
+    }
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in the latest push by changing the shared LOOKER_STUDIO label to `Looker Studio` and adding coverage for the user-facing string.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -247,6 +274,59 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
     };
   }
 
+  async runReport(request: McpRunReportRequest): Promise<McpRunReportResponse> {
+    const report = await this.getReportService.run(
+      new GetReportCommand(request.reportId, request.projectId, request.userId, request.roles)
+    );
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in the latest push. `runReport()` now delegates destination gating to the service path, avoiding the duplicate report fetch and duplicated validation.

apps/backend/src/ee/mcp/tools/run-report.tool.ts
```diff
@@ -0,0 +1,62 @@
...
+    report_id: z.string(),
+    run_id: z.string(),
+  };
+  readonly annotations = {
+    title: 'Run Report',
+    readOnlyHint: false,
+    destructiveHint: true,
+    idempotentHint: false,
+    openWorldHint: true,
+  };
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in the latest push: `run_report` now keeps `readOnlyHint: false` but sets `destructiveHint: false`. It remains non-idempotent and `mcp:write`-protected because each call starts a billable report run.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -247,6 +297,59 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+
+    const report = await this.getReportService.run(
+      new GetReportCommand(request.reportId, request.projectId, request.userId, request.roles)
+    );
+    const destinationType = report.dataDestinationAccess.type;
+    if (!MCP_RUN_REPORT_DESTINATION_TYPES.has(destinationType)) {
+      throw new BusinessViolationException(
+        `Reports with a ${toHumanReadable(destinationType)} destination are pull-based and cannot be run through run_report`
+      );
+    }
```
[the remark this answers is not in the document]
kalnik-a-a: Keeping this scoped to the MCP facade for now. The extra report read is acceptable here because `run_report` is not a hot path, and moving destination gating behind `RunReportService` would broaden the shared product service boundary and risk changing existing REST/scheduling behavior. This PR intentionally keeps that policy in the MCP-only layer.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -247,6 +297,59 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+      request.projectId
+    );
+
+    const report = await this.getReportService.run(
+      new GetReportCommand(request.reportId, request.projectId, request.userId, request.roles)
+    );
+    const destinationType = report.dataDestinationAccess.type;
+    if (!MCP_RUN_REPORT_DESTINATION_TYPE_SET.has(destinationType)) {
+      throw new BusinessViolationException(
+        `Reports with a ${toHumanReadable(destinationType)} destination are pull-based and cannot be run through run_report`
```
laskevych: 💡 **Suggestion:** The "pull-based" rejection is derived by *exclusion* — anything not in `MCP_RUN_REPORT_DESTINATION_TYPES` is reported as `"... destination are pull-based and cannot be run through run_report"`. Today that's accurate (only `LOOKER_STUDIO` is excluded and it genuinely is pull-based). But the moment a new **push** destination type is added to `DataDestinationType`, `run_report` will reject it with a misleading "pull-based" message instead of a "not yet supported" one. Consider driving this off an explicit pull/push classification (an `isPullBasedDestinationType` helper next to the existing `isEmailBasedDataDestinationType` in `data-destination-type.enum.ts`) so the allowlist and the message can't drift apart.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -247,6 +297,59 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
     };
   }
 
+  async runReport(request: McpRunReportRequest): Promise<McpRunReportResponse> {
+    await this.reportAccessService.checkOperateAccess(
```
laskevych, line 321: 🔵 **Question:** `runReport` calls `reportAccessService.checkOperateAccess(...)` here, and then `runReportService.run(..., RunType.manual)` runs `checkOperateAccess` again internally. Each evaluation reloads the report and issues up to three `accessDecisionService.canAccess` calls, so a single `run_report` performs the full operate-access evaluation twice. I think this pre-check is intentional — it normalizes the foreign-project "Report not found." error *before* the destination-type check can leak that the report exists (the `normalizes foreign-project report access` test seems to depend on this ordering). Can you confirm that's the reason? If so, a one-line comment on this pre-check explaining "must run before `getReport`/destination check to avoid leaking existence" would keep the intentional double-check from looking accidental.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -40,18 +63,45 @@ function buildGoogleSheetUrl(spreadsheetId: string, sheetId: number): string {
...
+const RUN_REPORT_RUN_TYPES: ReadonlySet<DataMartRunType> = new Set(
+  MCP_RUN_REPORT_DESTINATION_TYPES.map(toReportRunType)
+);
+
+const MCP_STATUS_BY_RAW_STATUS: Record<DataMartRunStatus, McpReportRunStatus> = {
+  [DataMartRunStatus.PENDING]: 'running',
+  [DataMartRunStatus.RUNNING]: 'running',
+  [DataMartRunStatus.SUCCESS]: 'success',
+  [DataMartRunStatus.FAILED]: 'failed',
+  [DataMartRunStatus.CANCELLED]: 'failed',
```
laskevych: 🔵 **Question:** `CANCELLED`, `INTERRUPTED`, and `RESTRICTED` all map to the assistant-facing `status: "failed"` with `error: "Report run ended with status CANCELLED"`. For a run the user *cancelled themselves*, surfacing it to the AI as a failure (which the AI may then report as "the run failed") seems slightly off. `raw_status` does preserve the real state, and the changeset documents the three-state simplification — so this may be deliberate. Just want to confirm you're comfortable with a user-initiated cancel being described to the assistant as `failed` rather than a distinct terminal signal.

apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts
```diff
@@ -0,0 +1,200 @@
...
+        shouldPoll: true,
+        stopReason: null,
+        message: `This run has been queued for about ${queuedMinutes} minutes and may still execute when workers catch up. Let the user know it is taking longer than normal, and keep polling. ${KEEP_POLLING_MESSAGE}`,
+      };
+    }
+
+    return this.keepPollingAdvice();
+  }
+
+  private claimedWithoutStartAdvice(queuedAt: string | null): PollingAdvice {
```
laskevych: 🔵 **Question:** There's an asymmetry in the polling advice: `pendingAdvice` and `startedAdvice` both emit a "taking longer than usual, keep polling" warning once elapsed ≥ `RUN_LONGER_THAN_USUAL_MINUTES` (10 min), but `claimedWithoutStartAdvice` jumps straight from the normal keep-polling message to the hard stop at 30 min — a run that's been `RUNNING` without a recorded `started_at` for 10–29 minutes reports as `"normal"`. Is skipping the intermediate "longer than usual" nudge for the claimed-without-start case intentional? A claimed-but-not-started run for ~20 minutes seems at least as worth warning the user about as a still-queued one.

apps/backend/src/ee/mcp/tools/run-report.tool.ts
```diff
@@ -0,0 +1,62 @@
...
+    report_id: z.string().trim().min(1),
+  })
+  .strict();
+
+type RunReportInput = z.infer<typeof inputSchema>;
+
+@Injectable()
+export class RunReportTool implements McpToolDefinition<RunReportInput> {
+  readonly name = 'run_report';
+  readonly description =
```
laskevych, line 22: 💡 **Suggestion (low priority):** The tool `description` strings (~1.5–1.6k chars each here and in `get-report-run-status.tool.ts`) plus the `KEEP_POLLING_MESSAGE` returned on every poll are quite large, and they're sent to the model on every listing/poll. I understand this is deliberate steering to stop the assistant from bailing out of the polling loop — so this is low-priority. But if you ever see context pressure, consolidating the repeated "a running status is normal, don't stop" guidance into a single canonical place (the `message` field only, or the description only) would cut a lot of per-call tokens.

laskevych (commented):
Automated code review — a few refinement suggestions and questions. No blocking correctness issues found; the PR is well-tested and cleanly structured. Comments left inline.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
kalnik-a-a: Addressed in cfaae3c5e. Added an explicit `isPullBasedDataDestinationType` helper and split the error path: actual pull-based destinations still get the pull-based message, while future unsupported destination types now get a neutral `not supported by run_report` message. Added facade coverage for that distinction.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
kalnik-a-a: Addressed in cfaae3c5e. The pre-check stays intentional for this ordering: it normalizes cross-project or inaccessible report ids before the facade loads destination policy. I added a short code comment before the check to make that explicit.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
kalnik-a-a: Addressed in cfaae3c5e. Expanded the MCP status contract so CANCELLED, INTERRUPTED, and RESTRICTED surface as separate terminal statuses instead of being flattened to failed. `raw_status` is still returned, failed runs still carry backend error details, and the changeset/tests were updated to cover the expanded terminal set.

apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts
kalnik-a-a: Addressed in cfaae3c5e. Added the same longer-than-usual warning path for RUNNING runs that have been claimed by a worker but have no started_at after 10 minutes. They still keep should_poll=true until the 30-minute stuck threshold, with test coverage for the 10-minute warning and 30-minute stop.

apps/backend/src/ee/mcp/tools/run-report.tool.ts
kalnik-a-a: Addressed in cfaae3c5e. Shortened the repeated polling guidance in the tool descriptions and kept the detailed "do not stop polling" steering in the per-call message where it is most useful. The description tests now pin that the duplicated wording is not repeated.

apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts
```diff
@@ -0,0 +1,211 @@
...
```
romandubovyi: 🟡 **Important:** `claimedWithoutStartAdvice` measures elapsed time from `queuedAt` (`createdAt`), which includes queue wait. Report runs are claimed as RUNNING in `claimRunSlotAtomically` before `startedAt` is set in `markAsStarted`, so a run that waited in PENDING for ~30 minutes can be flagged as stuck immediately after claim even though execution just began. Consider using a separate "claimed at" timestamp, or only applying the stuck thresholds once `startedAt` is available / after a grace period in the claimed-without-start state.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -247,6 +298,65 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+      const destinationLabel = toHumanReadable(destinationType);
+      const reason = isPullBasedDataDestinationType(destinationType)
+        ? 'are pull-based and cannot be run through run_report'
+        : 'are not supported by run_report';
+      throw new BusinessViolationException(
+        `Reports with a ${destinationLabel} destination ${reason}`
+      );
+    }
+
+    const enqueued = await this.runReportService.run({
```
romandubovyi, line 345: 💡 **Suggestion:** The MCP path calls `checkOperateAccess` here and again inside `RunReportService.run()` for manual runs. Consider skipping the service-level check when the facade has already authorized, or documenting that the duplicate is intentional for REST parity — it's harmless but adds an extra auth round-trip per MCP invocation.

apps/backend/src/ee/mcp/tools/run-report.tool.ts
```diff
@@ -0,0 +1,62 @@
...
+    'Start a report run by report_id, delivering fresh data to its push destination (Google Sheets, Email, Slack, Microsoft Teams, or Google Chat). Use this when the user wants up-to-date data in an existing report. Returns immediately with just report_id and run_id — it does not wait for the run to finish and does not report status. After starting one, poll get_report_run_status with the returned ids until should_poll is false; follow that tool message, waiting up to 15 seconds between checks when possible. Each run_report call starts a new, separately billed run, so never call run_report again for the same report while a run is in progress — keep polling the existing run_id instead. If the report already has a run in flight, this call fails with "Report is already running or pending"; that is not a report failure. Pull-based consumers (Data Studio, HTTP Data API) fetch data themselves and cannot be run with this tool. Billed as a standard Report Run for the destination, not as an MCP query.';
```
romandubovyi: 🔵 **Question:** The description lists "HTTP Data API" alongside Data Studio as pull-based consumers, but `isPullBasedDataDestinationType()` only covers `LOOKER_STUDIO` today. Is HTTP Data API a future destination type, or should the description stick to what's actually rejected by the facade guard?

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -247,6 +298,65 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
     };
   }
 
+  async runReport(request: McpRunReportRequest): Promise<McpRunReportResponse> {
+    // Must happen before loading the report/destination policy, so foreign-project ids
+    // normalize to "Report not found" instead of leaking destination details.
+    await this.reportAccessService.checkOperateAccess(
```
romandubovyi, line 321: 🟡 **Important:** `runReport` calls `reportAccessService.checkOperateAccess` before `runReportService.run`, but the service repeats the same check for manual runs. This causes the access check to run twice and couples the facade to auth details that the service already owns. Consider removing the facade-level check or adding an option to skip it when the service will enforce it.

apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts
```diff
@@ -0,0 +1,211 @@
...
+  private pollingAdvice(
+    startedAt: string | null,
+    queuedAt: string | null,
+    rawStatus: DataMartRunStatus
+  ): PollingAdvice {
+    if (rawStatus === DataMartRunStatus.PENDING) {
+      return this.pendingAdvice(queuedAt);
+    }
+
+    if (rawStatus === DataMartRunStatus.RUNNING && !startedAt) {
```
romandubovyi, line 109: 🟡 **Important:** The `claimedWithoutStartAdvice` branch handles `rawStatus === RUNNING` with `startedAt === null`. It is unclear whether this state is reachable in practice: runs are created either as `PENDING` or as `RUNNING` with `startedAt` already set. If this is a real edge case, please document how it can occur; otherwise consider removing this branch to simplify the polling logic.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -44,19 +68,46 @@ function buildGoogleSheetUrl(spreadsheetId: string, sheetId: number): string {
   return `https://docs.google.com/spreadsheets/d/${spreadsheetId}/edit#gid=${sheetId}`;
 }
 
+const MCP_RUN_REPORT_DESTINATION_TYPES = [
+  DataDestinationType.GOOGLE_SHEETS,
+  DataDestinationType.EMAIL,
+  DataDestinationType.SLACK,
```
romandubovyi, line 74: 💡 **Suggestion:** The `MCP_RUN_REPORT_DESTINATION_TYPES` allowlist is hardcoded next to `toReportRunType` and `isPullBasedDataDestinationType`. Consider deriving the supported set from those existing mappings so a new push destination does not need to be added in two places, or add a comment explaining why this exact list is maintained separately.

apps/backend/src/ee/mcp/tools/run-report.tool.ts
romandubovyi, line 22: 💡 **Suggestion:** The `run_report` description is very long (~700 chars). LLM clients parse tool descriptions better when they are concise. Consider moving the detailed polling/billing guidance into `get_report_run_status.description` or tool annotations, and keeping this one focused on what the tool does and its immediate return value.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -317,4 +427,27 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
       last_run_at: trigger.lastRunTimestamp?.toISOString() ?? null,
     };
   }
+
+  private toRunResponse(
+    run: DataMartRun
+  ): Pick<
+    McpGetReportRunStatusResponse,
```
romandubovyi, line 439: 💡 **Suggestion:** The fallback message in `toErrorMessage` interpolates `run.status`, but this method is only called for failed runs, so the status is always `FAILED`. Consider a more specific fallback like "Report run failed; no detailed error was recorded" and optionally include the run id for support/debugging.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -44,19 +68,46 @@ function buildGoogleSheetUrl(spreadsheetId: string, sheetId: number): string {
...
+
+const MCP_STATUS_BY_RAW_STATUS: Record<DataMartRunStatus, McpReportRunStatus> = {
+  [DataMartRunStatus.PENDING]: 'running',
+  [DataMartRunStatus.RUNNING]: 'running',
+  [DataMartRunStatus.SUCCESS]: 'success',
+  [DataMartRunStatus.FAILED]: 'failed',
+  [DataMartRunStatus.CANCELLED]: 'cancelled',
+  [DataMartRunStatus.INTERRUPTED]: 'interrupted',
+  [DataMartRunStatus.RESTRICTED]: 'restricted',
+};
```
romandubovyi, line 95: ⚪ **Nit:** The constructor now takes 12 dependencies. That is a lot for one facade; if more MCP report tools are added, consider splitting the facade by responsibility (e.g., read vs. write/run) to keep each class cohesive.

romandubovyi (approved):
No blocking issues. The `run_report` / `get_report_run_status` flow is cleanly separated from existing REST paths, access checks are tightened, and the polling logic is well covered by tests. I left a few non-blocking inline suggestions. Overall — approved, good to merge.

apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts
kalnik-a-a: Addressed in 19b5d4f84. claimed-without-start no longer applies the 10/30 minute thresholds from queued_at, because queued_at includes queue wait. That state now stays pollable and explains that the worker has claimed the run but started_at is not recorded yet. Added coverage for the long-queued-then-claimed case.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
kalnik-a-a: Kept the duplicate check intentionally and documented the boundary in 19b5d4f84. The facade pre-check preserves MCP-specific error ordering before destination policy is loaded, while RunReportService keeps its own manual-run check for non-MCP callers. I did not add a skip flag to avoid changing the shared service contract in this PR.

apps/backend/src/ee/mcp/tools/run-report.tool.ts
kalnik-a-a: Addressed in 19b5d4f84. Removed HTTP Data API from the run_report description so the text matches the destination guard that exists today; it now only describes currently supported push destinations and the concurrent-run behavior.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
kalnik-a-a: Documented in 19b5d4f84. I kept the facade-level pre-check because it normalizes inaccessible or cross-project ids before the MCP-only destination check can load destination details. The repeated service check remains as the shared RunReportService invariant for manual runs, so existing non-MCP flows do not depend on the MCP facade.

apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts
kalnik-a-a: Addressed in 19b5d4f84. This state is reachable between report-run trigger claim and markAsStarted: the worker first moves the run to RUNNING, then markAsStarted records started_at shortly after. I documented that worker-claimed state in the status-tool description and kept it pollable.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
kalnik-a-a: Addressed in 19b5d4f84. Added a comment clarifying that MCP_RUN_REPORT_DESTINATION_TYPES is the tool contract allowlist, not a global push/pull classifier. RUN_REPORT_RUN_TYPES remains derived from it; future unsupported destinations still get the neutral not-supported path.

apps/backend/src/ee/mcp/tools/run-report.tool.ts
kalnik-a-a: Addressed in 19b5d4f84. Shortened run_report.description and kept it focused on the immediate return/polling handoff, duplicate-run warning, billing note, and currently supported destinations. The detailed polling behavior remains in get_report_run_status and its message.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
kalnik-a-a: Addressed in 19b5d4f84. The fallback now says Report run failed; no detailed error was recorded instead of interpolating the raw FAILED status, with facade coverage for failed runs without recorded errors.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
kalnik-a-a: Left this as-is for now. The dependency count is high, but splitting the facade would be a broader structure change than this review pass needs. I would treat that as a follow-up if more MCP report tools are added.

PR #1393, opened 2026-07-06
feat(mcp): add add_destination MCP tool

romandubovyi:
## Summary

Adds an `add_destination` MCP tool (task #6707): an AI assistant connected to OWOX over MCP can now create a new report-delivery destination directly, instead of only picking from existing ones via `list_destinations`.

Write tool (`mcp:write` scope, `readOnlyHint: false`).

## What changed

- **New MCP tool** `AddDestinationTool` (`ee/mcp/tools/add-destination.tool.ts`) — input `{ destination_type, title?, emails? }`; behavior branches on `destination_type`:
  - **email-based** (`email`, `slack`, `teams`, `google_chat`) — created directly via `McpDataDestinationsFacade.createDestination`; requires `emails`; `destination_id` returned immediately.
  - **`looker_studio`** — created directly, no OAuth/emails needed; returns connector deployment URL + destination id/secret so the caller can wire up the connector right away.
  - **`google_sheets`** — no facade round-trip. Returns a link into a new, minimal, authenticated in-app "Connect Google Sheets" page (`ConnectGoogleSheetsPage`); the destination doesn't exist yet, so no `destination_id` is returned — the caller is told to poll `list_destinations` afterward once the user confirms they finished.
- **New frontend flow** for the `google_sheets` case:
  - `ConnectFlowLayout` + extracted `ProjectAuthGuards` (pulled out of `MainLayout`, no behavior change there) — a reusable, chrome-free, but still fully authenticated/project-scoped layout for future single-purpose "connect X" pages, not just this one.
  - `ConnectGoogleSheetsPage` — a simplified version of the existing "Configure destination" form: title input + `GoogleOAuthConnectButton`; creates the destination automatically as soon as Google access is granted, no separate save step.
  - `ConnectGoogleSheetsDonePage` — a dedicated confirmation route (not just local component state), so refreshing the tab after success can't accidentally re-show the connect form and risk a duplicate destination. Deliberately generic (no destination name from the URL — an unverified query param isn't proof of what was actually created).
- **`availableForUse` threaded through creation**: `CreateDataDestinationCommand`, the REST `CreateDataDestinationApiDto`, and the create use-case all gained an optional `availableForUse` (default `true`, preserving existing behavior). Every destination created via `add_destination` starts `availableForUse: false` — a human must review and share it (Configure destination → Sharing) before it can be used in reports.
- **`list_destinations` updated** to expose `availableForUse` and `connectedGoogleAccount` (for `google_sheets`, the Google account that completed OAuth) so an agent can find the destination it just created and confirm it's the right one before using it.
- **Cleanup while touching this area**: removed a duplicate `joinPublicOrigin` helper (was defined in both `data-mart-ui-path.ts` and the new `mcp-public-url.util.ts`), consolidated on one.

## Design notes

- **No parent-entity role gate.** Unlike report/schedule MCP tools (which check access against an existing data mart/report), destination creation has no parent entity to check access against — matching the REST `POST /data-destinations` endpoint, which already allows any project member (viewer included).
- **`google_sheets` never round-trips a title/name through the URL.** Earlier iterations passed `title` as a query param into the connect page and pre-filled it there; this was dropped entirely — an unverified value from a link isn't proof of what the user actually wants, so the user always sets the destination's name directly in the form.
- **The confirmation page carries no data from the URL either**, for the same reason — it's a static "success" screen, not a place to reflect back arbitrary content from a link.

## Testing

- Unit tests: `add-destination.tool.spec.ts` (all three destination-type branches, default-title generation, missing-emails rejection, metadata), `mcp-data-destinations.facade.impl.spec.ts`, `create-data-destination.service.spec.ts`, `ConnectGoogleSheetsPage.test.tsx`, `ConnectGoogleSheetsDonePage.test.tsx`, `ProjectAuthGuards.test.tsx`, `ConnectFlowLayout.test.tsx`.
- Full regression: backend `src/data-marts` + `src/ee/mcp` (336 suites / 4657 tests) and the touched frontend suites — all green. `tsc`, `eslint`, `prettier` clean on both apps.

### DIFF
--- a/apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
+++ b/apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
@@ -1,16 +1,42 @@
-import { Injectable } from '@nestjs/common';
+import { BadRequestException, Injectable, Logger } from '@nestjs/common';
+import { DataDestinationType } from '../data-destination-types/enums/data-destination-type.enum';
 import { ListDataDestinationsCommand } from '../dto/domain/list-data-destinations.command';
 import { ListDataDestinationsService } from '../use-cases/list-data-destinations.service';
-import { toMcpDestinationType } from './mcp-destination-type';
+import { CreateDataDestinationService } from '../use-cases/create-data-destination.service';
+import { CreateDataDestinationCommand } from '../dto/domain/create-data-destination.command';
+import { DataDestinationCredentialService } from '../services/data-destination-credential.service';
+import { PublicOriginService } from '../../common/config/public-origin.service';
+import { DataDestinationCredential } from '../entities/data-destination-credential.entity';
+import { DataDestinationCredentials } from '../data-destination-types/data-destination-credentials.type';
+import { LookerStudioConnectorCredentials } from '../data-destination-types/looker-studio-connector/schemas/looker-studio-connector-credentials.schema';
+import { toMcpDestinationType, type McpDestinationType } from './mcp-destination-type';
 import {
   McpDataDestinationsFacade,
   McpListDestinationsRequest,
   McpListDestinationsResponse,
+  McpCreateDestinationRequest,
+  McpCreateDestinationResponse,
 } from './mcp-data-destinations.facade';
+
+const REVERSE_DESTINATION_TYPE_MAP: Record<McpDestinationType, DataDestinationType> = {
+  google_sheets: DataDestinationType.GOOGLE_SHEETS,
+  looker_studio: DataDestinationType.LOOKER_STUDIO,
+  email: DataDestinationType.EMAIL,
+  slack: DataDestinationType.SLACK,
+  teams: DataDestinationType.MS_TEAMS,
+  google_chat: DataDestinationType.GOOGLE_CHAT,
+};
 
 @Injectable()
 export class McpDataDestinationsFacadeImpl implements McpDataDestinationsFacade {
-  constructor(private readonly listDataDestinationsService: ListDataDestinationsService) {}
+  private readonly logger = new Logger(McpDataDestinationsFacadeImpl.name);
+
+  constructor(
+    private readonly listDataDestinationsService: ListDataDestinationsService,
+    private readonly createDataDestinationService: CreateDataDestinationService,
+    private readonly dataDestinationCredentialService: DataDestinationCredentialService,
+    private readonly publicOriginService: PublicOriginService
+  ) {}
 
   async listDestinations(
     request: McpListDestinationsRequest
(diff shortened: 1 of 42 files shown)


apps/web/src/pages/connect/ConnectGoogleSheetsPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    const payload: CreateDataDestinationRequestDto = {
+      title: currentTitle,
+      type: DataDestinationType.GOOGLE_SHEETS,
+      credentials: { type: DataDestinationCredentialsType.GOOGLE_SHEETS_CREDENTIALS },
+      credentialId: connectedCredentialId,
+      // Created via an external MCP-driven flow — starts unshared so a human reviews it
+      // before the whole project can use it in reports.
+      availableForUse: false,
```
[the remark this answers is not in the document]
romandubovyi: Project membership and authentication are already enforced earlier by the project-scoped route guards before the user reaches this page. The MCP instructions now also explicitly tell the user to open the link with the same OWOX account that connected MCP and explain that otherwise the unshared destination will not be discoverable afterward. This keeps the account/project requirement at the flow-entry level instead of adding MCP-specific messaging to the generic setup UI.

romandubovyi:
@codex review

apps/backend/src/data-marts/mappers/data-destination.mapper.ts
```diff
@@ -61,6 +63,26 @@ export class DataDestinationMapper {
...
+      DataDestinationType.GOOGLE_SHEETS,
+      context.userId,
+      undefined,
+      dto.credentialId,
+      undefined,
+      undefined,
+      context.roles ?? [],
+      undefined,
+      // MCP-driven connect flow — starts unshared; not client-controlled.
+      false
```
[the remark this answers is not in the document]
romandubovyi: This is intentional, not a bypass.

`availableForUse: false` means the destination is **not shared with the rest of the project** — it does not block the creator/owner. Destination owners are granted `USE` for any sharing state by our access matrix (`OwnerStatus.OWNER` + `SharingState.NOT_SHARED`), and that is the desired behavior here: the MCP user who completes the connect flow should be able to call `add_report` with the destination they just created.

`list_destinations` still surfaces `availableForUse` so the agent knows the destination is not project-wide yet and should not suggest it to other team members until a human shares it (Configure destination → Sharing). The MCP tool descriptions already steer the agent accordingly.

We are **not** adding an extra MCP-only gate on top of ownership — that would contradict the product model where unshared only restricts non-owners.

apps/backend/src/data-marts/use-cases/create-data-destination.service.ts
```diff
@@ -160,7 +161,7 @@ export class CreateDataDestinationService {
         projectId: command.projectId,
         credentialId: command.credentialId,
         createdById: command.userId,
-        availableForUse: true,
+        availableForUse,
```
kalnik-a-a, line 166: This only checks that the standalone OAuth credential belongs to the project. The credential was created with `createdById`, but for an unlinked credential there is no check that it was created by `command.userId`; any same-project viewer who gets this UUID can bind another user's Google OAuth tokens to a destination they own. Please require `credential.createdById === command.userId` for unlinked credentials, while keeping the existing `COPY_CREDENTIALS` flow for already-linked credentials.

apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
```diff
@@ -19,18 +45,148 @@ export class McpDataDestinationsFacadeImpl implements McpDataDestinationsFacade
...
+      type: dataDestinationType,
+      userId: request.userId,
+      credentials,
+      roles: request.roles,
+      // MCP-created destinations start unshared — an agent-created destination shouldn't
+      // silently become available to the whole project's reports before a human reviews it.
+      availableForUse: false,
+    });
+
+    const result = await this.createDataDestinationService.run(command);
```
kalnik-a-a, line 150: This persists the Looker Studio destination before the later credential lookup can throw. If that post-create lookup/config resolution fails, MCP reports failure even though an unshared destination already exists, so retries can create duplicates. Please either avoid failing after persistence for metadata that is not returned through MCP, or move the validation into the create transaction so failure rolls back.

kalnik-a-a (changes requested):
Requesting changes for the Google Sheets OAuth credential ownership issue. I also left one non-blocking note inline.

apps/backend/src/data-marts/use-cases/create-data-destination.service.ts
romandubovyi: Fixed. For unlinked credentials, CreateDataDestinationService now requires `credential.createdById === command.userId`. The already-linked credential path still uses the existing `COPY_CREDENTIALS` check.

Added a regression test for rejecting an unlinked same-project credential created by another user.

apps/backend/src/data-marts/facades/mcp-data-destinations.facade.impl.ts
romandubovyi: Fixed by avoiding post-create failure in the MCP facade. After the destination is created, Looker Studio credential metadata resolution is best-effort: if the lookup/secret resolution fails, the facade returns the created destination without `lookerStudioCredentials` instead of throwing and causing retry-created duplicates.

Updated the facade test to cover this behavior.

kalnik-a-a (approved):

PR #1395, opened 2026-07-06
feat(mcp): make query_data_mart timeout-resilient

zapolsky:
## What

Makes the synchronous MCP tool `query_data_mart` resilient to slow/abandoned queries so a long or cancelled run ends cleanly and is **never billed**, instead of running to completion and charging for a result nobody receives.

## Design — three layers of defense

1. **App-side deadline** (`DEFAULT_QUERY_DEADLINE_MS = 3 min`). `QueryDataMartService.run` bounds the DWH work (compose + read + totals) with `Promise.race([produce, deadline, aborted])`. On expiry it throws `QueryTimeoutError` → the run is recorded `FAILED` and billing (success-path only) is skipped.
2. **Client abort** via the MCP request's `AbortSignal`, threaded `tool → sdk factory → facade → service`. On abort it throws `QueryAbortedError` → recorded `CANCELLED`, not billed. The read loop also bails cooperatively once the signal fires.
3. **Native warehouse cancellation** as a cost backstop. `signal` + `queryTimeoutMs` are threaded (optional) into the BigQuery/Snowflake adapters: BigQuery sets `jobTimeoutMs` and calls `job.cancel()` on abort; Snowflake sets `STATEMENT_TIMEOUT_IN_SECONDS` and calls `statement.cancel()` on abort. Both the rows read and the totals query are cancelled. Storages that don't honor it (`honorsQueryTimeout` unset) are logged so the gap is observable.

The `/mcp` controller raises the per-request socket timeout to 4 min so the global 3-min socket-idle timeout (`SERVER_TIMEOUT_MS`) doesn't blunt-reset a computing request before the app deadline returns a clean `query_timeout`.

**Billing invariant:** consumption is success-path only and additionally gated on the Run History (audit) write succeeding — no timeout/abort/failed run is ever billed, and no success is billed without an audit row.

## Backward compatibility

`signal` and `queryTimeoutMs` are **optional** params on the shared adapters/readers. Every other `executeQuery` caller (create-view, sql-run, schema-provider, report runs) passes neither → identical behavior. No migrations, no entity changes, no new env vars (the timeout is a code constant, deliberately not env-driven).

## Testing

- New/changed facade impl (`queryDataMart(signal)`) has unit tests (CI-blocking rule).
- Adversarial cases covered: reader-finalize leak on a lost race, true totals parallelism, `limit+1` over-read + `=== limit` boundary, resource cleanup (timer + abort listener), cooperative loop-bail, BigQuery/Snowflake native-cancel paths, and Snowflake sync-complete listener safety.
- Verified: `tsc -p tsconfig.build.json` clean, ESLint clean, full touched-area suite green.

### DIFF
--- a/apps/backend/src/data-marts/config/mcp-operation-timeout-exclusions.spec.ts
+++ b/apps/backend/src/data-marts/config/mcp-operation-timeout-exclusions.spec.ts
@@ -1,0 +1,35 @@
+import { pathToRegexp } from 'path-to-regexp';
+import { addLeadingSlash } from '@nestjs/common/utils/shared.utils';
+import { isRouteExcluded } from '@nestjs/core/router/utils';
+import { RequestMethod } from '@nestjs/common';
+import { MCP_OPERATION_TIMEOUT_EXCLUSIONS } from './mcp-operation-timeout-exclusions';
+
+/**
+ * Regression guard for the /mcp operation-timeout preemption. NestJS binds the module's global
+ * `{*path}` middleware to the global-prefix-excluded /mcp route, so without an explicit exclusion the
+ * 30s operation-timeout middleware 408s a long query_data_mart run (and bills a SUCCESS the client
+ * already saw fail). This asserts our exclusion entries actually match a real /mcp request, using
+ * NestJS's own `isRouteExcluded` + `pathToRegexp` — the same matching the middleware uses at runtime.
+ */
+describe('MCP operation-timeout exclusions', () => {
+  // Mirror NestJS's mapToExcludeRoute: each entry compiles to a pathRegex over its leading-slashed path.
+  const compiled = MCP_OPERATION_TIMEOUT_EXCLUSIONS.map(route => ({
+    path: route.path,
+    requestMethod: route.method,
+    pathRegex: pathToRegexp(addLeadingSlash(route.path)).regexp,
+  }));
+
+  const isExcluded = (url: string) =>
+    isRouteExcluded(compiled, url, RequestMethod[RequestMethod.POST]);
+
+  it('excludes the /mcp transport route and its subpaths', () => {
+    expect(isExcluded('/mcp')).toBe(true);
+    expect(isExcluded('/mcp/')).toBe(true);
+    expect(isExcluded('/mcp/sse')).toBe(true);
+  });
+
+  it('does not over-exclude unrelated routes', () => {
+    expect(isExcluded('/api/data-marts/abc')).toBe(false);
+    expect(isExcluded('/api/mcp-lookalike')).toBe(false);
+  });
+});
--- a/apps/backend/src/data-marts/config/mcp-operation-timeout-exclusions.spec.ts
+++ b/apps/backend/src/data-marts/config/mcp-operation-timeout-exclusions.spec.ts
@@ -19,8 +19,7 @@ describe('MCP operation-timeout exclusions', () => {
     pathRegex: pathToRegexp(addLeadingSlash(route.path)).regexp,
   }));
 
-  const isExcluded = (url: string) =>
-    isRouteExcluded(compiled, url, RequestMethod[RequestMethod.POST]);
+  const isExcluded = (url: string) => isRouteExcluded(compiled, url, RequestMethod.POST);
 
   it('excludes the /mcp transport route and its subpaths', () => {
     expect(isExcluded('/mcp')).toBe(true);
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
@@ -87,20 +87,15 @@ export class BigQueryApiAdapter {
     jobTimeoutMs?: number,
     signal?: AbortSignal
   ): Promise<{ jobId: string }> {
-    const queryConfig: Query =
-      params && params.length > 0
-        ? {
-            query,
-            params: Object.fromEntries(params.map(p => [p.name, p.value])),
-            parameterMode: 'NAMED',
-            ...this.getLocationOption(),
-            ...(jobTimeoutMs !== undefined ? { jobTimeoutMs } : {}),
-          }
-        : {
-            query,
-            ...this.getLocationOption(),
-            ...(jobTimeoutMs !== undefined ? { jobTimeoutMs } : {}),
-          };
+    const queryConfig: Query = {
+      query,
+      ...this.getLocationOption(),
+      ...(jobTimeoutMs !== undefined ? { jobTimeoutMs } : {}),
+    };
+    if (params && params.length > 0) {
+      queryConfig.params = Object.fromEntries(params.map(p => [p.name, p.value]));
+      queryConfig.parameterMode = 'NAMED';
+    }
 
     const [job] = await this.bigQuery.createQueryJob(queryConfig);
 
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
@@ -136,6 +136,7 @@ export class BigQueryApiAdapter {
    * {@link executeQuery}, matching the previous `bigQuery.query()` behaviour.
    */
   private async waitForJobToComplete(job: Job, signal?: AbortSignal): Promise<void> {
+    let repolledAfterAbort = false;
     while (true) {
       const [metadata] = await job.getMetadata();
       if (metadata?.status?.state === 'DONE') {
@@ -147,18 +148,17 @@ export class BigQueryApiAdapter {
         }
         return;
       }
-      // Interruptible: an abort wakes the sleep so we re-poll immediately after job.cancel() (fired
-      // by executeQuery's abort handler) instead of waiting out the full poll interval.
-      await this.interruptibleSleep(BigQueryApiAdapter.JOB_POLL_INTERVAL_MS, signal);
+      // One immediate re-poll after job.cancel(), then a bounded interval — never busy-poll jobs.get.
+      if (signal?.aborted && !repolledAfterAbort) {
+        repolledAfterAbort = true;
+        continue;
+      }
+      await this.edgeInterruptibleSleep(BigQueryApiAdapter.JOB_POLL_INTERVAL_MS, signal);
     }
   }
 
-  private interruptibleSleep(ms: number, signal?: AbortSignal): Promise<void> {
+  private edgeInterruptibleSleep(ms: number, signal?: AbortSignal): Promise<void> {
     return new Promise(resolve => {
-      if (signal?.aborted) {
-        resolve();
-        return;
-      }
       const onAbort = () => {
         clearTimeout(timer);
         resolve();
@@ -167,6 +167,7 @@ export class BigQueryApiAdapter {
         signal?.removeEventListener('abort', onAbort);
         resolve();
       }, ms);
+      // Fires only on the abort edge; if already aborted it never fires, so the full interval elapses.
       signal?.addEventListener('abort', onAbort, { once: true });
     });
   }
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
@@ -76,8 +76,17 @@ export class BigQueryApiAdapter {
    * When params are provided, BigQuery named parameter mode is used
    * (@paramName placeholders). The SDK infers types from JS values
    * (string, number, boolean, Date).
+   *
+   * `jobTimeoutMs` (already in ms) caps warehouse cost by aborting the job server-side; unset = no change.
+   * `signal` (client disconnect/cancel) cancels the running job immediately — the poll loop then
+   * surfaces the cancellation as a thrown error. Unset = no cancellation hook.
    */
-  public async executeQuery(query: string, params?: SqlParameter[]): Promise<{ jobId: string }> {
+  public async executeQuery(
+    query: string,
+    params?: SqlParameter[],
+    jobTimeoutMs?: number,
+    signal?: AbortSignal
+  ): Promise<{ jobId: string }> {
     const queryConfig: Query =
       params && params.length > 0
         ? {
@@ -85,14 +94,31 @@ export class BigQueryApiAdapter {
             params: Object.fromEntries(params.map(p => [p.name, p.value])),
             parameterMode: 'NAMED',
             ...this.getLocationOption(),
+            ...(jobTimeoutMs !== undefined ? { jobTimeoutMs } : {}),
           }
         : {
             query,
             ...this.getLocationOption(),
+            ...(jobTimeoutMs !== undefined ? { jobTimeoutMs } : {}),
           };
 
     const [job] = await this.bigQuery.createQueryJob(queryConfig);
-    await this.waitForJobToComplete(job);
+
+    // Best-effort warehouse cancel: on abort, ask BigQuery to cancel the job. The poll loop below
+    // then sees the job finish (cancelled) and throws, so we stop billing compute for a gone client.
+    const onAbort = () => {
+      job.cancel().catch(() => undefined);
+    };
+    if (signal) {
+      if (signal.aborted) onAbort();
+      else signal.addEventListener('abort', onAbort, { once: true });
+    }
+
+    try {
+      await this.waitForJobToComplete(job, signal);
+    } finally {
+      signal?.removeEventListener('abort', onAbort);
+    }
     this.setLocationFromJob(job);
 
     const jobId = job.id;
@@ -109,7 +135,7 @@ export class BigQueryApiAdapter {
    * (e.g. invalid SQL) so a bad query still surfaces as a thrown error from
    * {@link executeQuery}, matching the previous `bigQuery.query()` behaviour.
    */
-  private async waitForJobToComplete(job: Job): Promise<void> {
+  private async waitForJobToComplete(job: Job, signal?: AbortSignal): Promise<void> {
     while (true) {
       const [metadata] = await job.getMetadata();
       if (metadata?.status?.state === 'DONE') {
@@ -121,8 +147,28 @@ export class BigQueryApiAdapter {
         }
         return;
       }
-      await new Promise(resolve => setTimeout(resolve, BigQueryApiAdapter.JOB_POLL_INTERVAL_MS));
+      // Interruptible: an abort wakes the sleep so we re-poll immediately after job.cancel() (fired
+      // by executeQuery's abort handler) instead of waiting out the full poll interval.
+      await this.interruptibleSleep(BigQueryApiAdapter.JOB_POLL_INTERVAL_MS, signal);
     }
+  }
+
+  private interruptibleSleep(ms: number, signal?: AbortSignal): Promise<void> {
+    return new Promise(resolve => {
+      if (signal?.aborted) {
+        resolve();
+        return;
+      }
+      const onAbort = () => {
+        clearTimeout(timer);
+        resolve();
+      };
+      const timer = setTimeout(() => {
+        signal?.removeEventListener('abort', onAbort);
+        resolve();
+      }, ms);
+      signal?.addEventListener('abort', onAbort, { once: true });
+    });
   }
 
   /**
--- a/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-report-reader.service.ts
+++ b/apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-report-reader.service.ts
@@ -37,6 +37,7 @@ import type { SqlParameter } from '../../utils/sql-clause-renderer';
 export class BigQueryReportReader implements DataStorageReportReader {
   protected readonly logger = new Logger(BigQueryReportReader.name);
   readonly type: DataStorageType = DataStorageType.GOOGLE_BIGQUERY;
+  readonly honorsQueryTimeout = true;
 
   private adapter: BigQueryApiAdapter;
   private reportResultTable: Table;
@@ -52,6 +53,8 @@ export class BigQueryReportReader implements DataStorageReportReader {
   private sqlOverride?: string;
   private sqlOverrideParams?: SqlParameter[];
   private columnFilter?: string[];
+  private queryTimeoutMs?: number;
+  private signal?: AbortSignal;
 
   constructor(
     protected readonly adapterFactory: BigQueryApiAdapterFactory,
@@ -90,6 +93,8 @@ export class BigQueryReportReader implements DataStorageReportReader {
     this.sqlOverride = options?.sqlOverride;
     this.sqlOverrideParams = options?.sqlOverrideParams;
     this.columnFilter = options?.columnFilter;
+    this.queryTimeoutMs = options?.queryTimeoutMs;
+    this.signal = options?.signal;
 
     this.reportDataHeaders = resolveReportDataHeaders(
       this.headersGenerator.generateHeaders(schema),
@@ -181,7 +186,12 @@ export class BigQueryReportReader implements DataStorageReportReader {
   }
 
   private async prepareQueryData(query: string, params?: SqlParameter[]): Promise<void> {
-    const { jobId } = await this.adapter.executeQuery(query, params);
+    const { jobId } = await this.adapter.executeQuery(
+      query,
+      params,
+      this.queryTimeoutMs,
+      this.signal
+    );
     const jobResult = await this.adapter.getJob(jobId);
     const destinationTable = jobResult.metadata.configuration.query.destinationTable;
     this.defineReportResultTable(
--- a/apps/backend/src/data-marts/data-storage-types/snowflake/adapters/snowflake-api.adapter.ts
+++ b/apps/backend/src/data-marts/data-storage-types/snowflake/adapters/snowflake-api.adapter.ts
@@ -93,11 +93,16 @@ export class SnowflakeApiAdapter {
   }
 
   /**
-   * Executes a SQL query
+   * Executes a SQL query. `queryTimeoutMs` caps warehouse cost via STATEMENT_TIMEOUT_IN_SECONDS
+   * (Snowflake's unit is SECONDS, so the ms budget is rounded up to ≥1s); unset = no change.
+   * `signal` (client disconnect/cancel) cancels the running statement immediately so an abandoned
+   * query stops billing compute; the cancel surfaces as a rejected promise. Unset = no cancel hook.
    */
   public async executeQuery(
     query: string,
-    dryRun: boolean = false
+    dryRun: boolean = false,
+    queryTimeoutMs?: number,
+    signal?: AbortSignal
   ): Promise<{
     queryId: string;
     rows?: Record<string, unknown>[];
@@ -108,10 +113,24 @@ export class SnowflakeApiAdapter {
     }
 
     return new Promise((resolve, reject) => {
-      this.connection.execute({
+      let settled = false;
+      let onAbort: (() => void) | undefined;
+      const cleanup = () => {
+        settled = true;
+        if (signal && onAbort) signal.removeEventListener('abort', onAbort);
+      };
+      const statement = this.connection.execute({
         sqlText: query,
         describeOnly: dryRun,
+        ...(queryTimeoutMs !== undefined
+          ? {
+              parameters: {
+                STATEMENT_TIMEOUT_IN_SECONDS: Math.max(1, Math.ceil(queryTimeoutMs / 1000)),
+              },
+            }
+          : {}),
         complete: (err, stmt, rows) => {
+          cleanup();
           if (err) {
             reject(
               new Error(
@@ -138,6 +157,15 @@ export class SnowflakeApiAdapter {
           }
         },
       });
+
+      // Best-effort warehouse cancel: on abort, cancel the running statement so a gone client stops
+      // billing compute. The cancel triggers the complete callback with an error → the promise rejects.
+      // `!settled` guards a synchronous complete (would leave a listener attached past resolve).
+      if (signal && !settled) {
+        onAbort = () => statement.cancel(() => undefined);
+        if (signal.aborted) onAbort();
+        else signal.addEventListener('abort', onAbort, { once: true });
+      }
     });
   }
 
--- a/apps/backend/src/data-marts/data-storage-types/snowflake/adapters/snowflake-api.adapter.ts
+++ b/apps/backend/src/data-marts/data-storage-types/snowflake/adapters/snowflake-api.adapter.ts
@@ -162,7 +162,14 @@ export class SnowflakeApiAdapter {
       // billing compute. The cancel triggers the complete callback with an error → the promise rejects.
       // `!settled` guards a synchronous complete (would leave a listener attached past resolve).
       if (signal && !settled) {
-        onAbort = () => statement.cancel(() => undefined);
+        onAbort = () => {
+          // Guard a driver that throws synchronously on an already-gone statement/connection.
+          try {
+            statement.cancel(() => undefined);
+          } catch {
+            /* best-effort cancel */
+          }
+        };
         if (signal.aborted) onAbort();
         else signal.addEventListener('abort', onAbort, { once: true });
       }
--- a/apps/backend/src/data-marts/facades/mcp-data-marts.facade.ts
+++ b/apps/backend/src/data-marts/facades/mcp-data-marts.facade.ts
@@ -86,8 +86,33 @@ export interface McpDataCatalogSummaryResponse {
 export interface McpDataMartsFacade {
   listDataMarts(request: McpListDataMartsRequest): Promise<McpListDataMartsResponse>;
   getDataMartDetails(request: McpGetDataMartDetailsRequest): Promise<McpDataMartDetailsResponse>;
-  queryDataMart(request: McpQueryDataMartRequest): Promise<McpQueryDataMartResponse>;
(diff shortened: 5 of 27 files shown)


apps/backend/src/ee/mcp/controllers/mcp-transport.controller.ts
```diff
@@ -5,6 +5,10 @@ import { McpAuthExceptionFilter } from '../auth/mcp-auth.exception-filter';
 import { McpAuthGuard } from '../auth/mcp-auth.guard';
 import { McpStreamableHttpTransportHandler } from '../sdk/mcp-streamable-http-transport.handler';
 
+// Above query_data_mart's 3-min deadline so the global socket-idle timeout (SERVER_TIMEOUT_MS)
+// doesn't blunt-reset a computing MCP call before it can return a clean query_timeout. LB (1h) caps.
+export const MCP_REQUEST_SOCKET_TIMEOUT_MS = 4 * 60_000;
```
zapolsky, line 10: **[Blocked — verified] A pre-existing 30s `operation-timeout` middleware fires on `/mcp` and preempts this entire timeout feature — and a run that finishes in the 30s–3min window is still billed for a result the client already received as a 408.**

This comment reasons about `SERVER_TIMEOUT_MS`, but the binding limiter on `/mcp` is a 30-second middleware this PR doesn't touch.

**Evidence chain** (verified against the worktree + an empirical run of NestJS 11.1.27's own `RouteInfoPathExtractor`):

1. `data-marts/data-marts.module.ts:838-845` — `createOperationTimeoutMiddleware(30000)` is `.forRoutes({ path: '{*path}', method: ALL })`; its `.exclude(...)` lists only the two `data-marts/:id` PUTs + `external/{*path}` — **`/mcp` is not excluded**. (The sibling `180000` middleware is PUT-only on those two paths, so it never binds to POST `/mcp`.)
2. `bootstrap.ts:50` sets global prefix `api` with `exclude: PROTOCOL_ROUTE_EXCLUSIONS`, and `config/protocol-route-exclusions.config.ts` puts `mcp` + `mcp/(.*)` in that list. For a wildcard middleware route, `RouteInfoPathExtractor.extractPathsFrom` appends `addLeadingSlash(route.path)` for every prefix-excluded route → the `{*path}` middleware binds to `/mcp` and `/mcp/(.*)`. Running the real extractor with this exact config returns `[…,"/mcp","/mcp/(.*)"]`.
3. `ee/mcp/sdk/mcp-streamable-http-transport.handler.ts:48-50` sets `enableJsonResponse: true`; the SDK buffers the `tools/call` response as one JSON body and writes nothing to `res` until the tool finishes, so for a `query_data_mart` >30s, `res.headersSent` is still false at 30s.
4. `common/middleware/operation-timeout.middleware.ts:16-25` — at 30s, `!res.headersSent && !res.destroyed` is true → it sends `res.status(408).json({message:"Request timeout after 30 seconds"})`. The client gets a generic 408; the PR's 3-min `QueryTimeoutError` / `query_timeout` structured error and the new 4-min socket timeout never take effect over HTTP.
5. **Billing:** the 408 is a normally-finished response (`res.writableFinished === true`), so `@hono/node-server` does **not** abort the request (`node_modules/@hono/node-server/.../listener.js:621-628` aborts only on `incoming.errored` or `!writableFinished`). Independently, the tool's `extra.signal` is a per-request `AbortController` (`@modelcontextprotocol/sdk/.../shared/protocol.js:318-323`) aborted **only** by an MCP `CancelledNotification` or transport `close()` — never by a Node POST `res` close. So the signal stays live: `QueryDataMartService.run` (`use-cases/query-data-mart.service.ts:240`) keeps racing, and if `produce` wins in the 30s–3min window it records SUCCESS and calls `registerMcpQueryRunConsumption` (`:249-286`) → **billed**, while the client already saw a 408. That is exactly the "billed for a result nobody receives" case this PR set out to prevent.

**Trigger:** any `query_data_mart` call that runs >30s. Client sees `408 Request timeout after 30 seconds`; if it then finishes before 3min, the run is billed. (The careful service-layer billing invariant is intact *inside* `run()` — the divergence is purely at the HTTP edge, which `run()` can't see.)

**Fix:** exclude `mcp` + `mcp/(.*)` from the 30s `operation-timeout` middleware, or give the MCP route its own operation timeout aligned with `MCP_REQUEST_SOCKET_TIMEOUT_MS` (4min), so the app-side 3-min deadline returns first. Add an integration test: a `/mcp` `tools/call` whose handler exceeds 30s must return `query_timeout` (not 408) and must not bill.

**Related (please verify with the same harness):** since `extra.signal` is aborted only by a `CancelledNotification` / transport close, a *plain client disconnect* on the POST likely doesn't fire it either — so Layer 2 ("client abort → CANCELLED, not billed") may also not trigger on a hard disconnect in this JSON-POST transport, only on an explicit cancel.

zapolsky (commented):
🔴 **1 blocking issue** — see the inline comment. It should be addressed before merge, but this automated review does not formally block; a human owns the merge decision.

zapolsky:
Solid, carefully-built PR — the `Promise.race` / listener / timer hygiene is meticulous, the service-layer billing invariant (`success ∧ audit ⟹ bill`, never on timeout/abort) is correct, moving `reader.finalize()` into `produce`'s own `finally` genuinely closes the lost-race leak, transient-reader isolation makes the parallel totals safe, and the test coverage is unusually thorough (96/96 green across the touched suites; the facade-impl signal test is present). ESLint is clean on all touched files; `tsc` is clean on the changed `src` (the spec type-errors under a strict `tsc -b` are a pre-existing, repo-wide baseline — the backend isn't type-checked over specs in CI, ts-jest runs with `diagnostics:false`).

**One blocking issue is posted inline** (🔴): a pre-existing 30s `operation-timeout` middleware bound to `/mcp` returns a generic 408 at 30s — preempting the whole 3-min timeout feature over HTTP, and billing a SUCCESS for a query the client already saw fail. See the inline comment on `mcp-transport.controller.ts` for the full chain.

Primary (criticality ≥ 80) — none of these block merge:

- `data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts:156-161` — **BigQuery busy-polls `job.getMetadata()` with zero backoff after a client abort.** Once `signal.aborted` is true, `interruptibleSleep` resolves immediately on *every* call, so `waitForJobToComplete` (`:138-154`) re-issues `jobs.get` back-to-back (no 2s spacing) until BigQuery flips the cancelled job to DONE. The intended "one immediate re-poll after `job.cancel()`" becomes a continuous poll; under concurrent abandonment this can trip BigQuery `jobs.get` rate limits (bounded to 3min by `jobTimeoutMs` for today's only caller, but that's the only bound). Fix: keep the first post-cancel poll immediate, then resume a bounded interval (e.g. a plain `setTimeout(JOB_POLL_INTERVAL_MS)` once aborted).
- `data-marts/use-cases/query-data-mart.service.ts:150-154, 181-189` — **In-flight DWH work is actively cancelled only on client abort, not on the app deadline or a non-abort failure.** On a pure 3-min timeout (no client signal) nothing calls `job.cancel()` / `statement.cancel()`; and when the rows read rejects on a non-abort error, the parallel totals query (started at `:181`, awaited only at `:227`) keeps running. Either way the reader(s) — and, for Snowflake, up to two dedicated connections — stay live until the warehouse-side cap (~3min) elapses, ~3min *after* the client already got its response. Self-heals, but is real connection/cost pressure under load. Fix: have the service own an internal `AbortController` linked to the client signal, and abort it from `produce`'s `finally` and from the deadline timer so both the timeout and a rows failure promptly cancel the warehouse work.

<details>

- `data-marts/use-cases/query-data-mart.service.ts:195-198` — the "storage doesn't honor `queryTimeoutMs`" gap (Athena/Redshift/Databricks) is only `logger.debug`, i.e. invisible at prod log level; a newly added storage silently ships with no warehouse cost cap. Consider `logger.warn` or a one-time startup assertion enumerating non-honoring storages.
- `data-marts/use-cases/query-data-mart.service.ts:43` + `ee/mcp/controllers/mcp-transport.controller.ts:10` — the two coupled timeout constants are hardcoded and scattered, unlike the repo's centralized env-driven timeout trio (`config/constants.ts:4-6` + `ConfigService`). Consequence: an operator who raises `SERVER_TIMEOUT_MS` gets zero effect on the MCP deadline (still 3min) or socket (still 4min). The `socket(4min) > deadline(3min)` invariant *is* CI-locked (good) — but consider deriving the socket from the deadline + env-driving the deadline.
- `data-marts/data-storage-types/snowflake/adapters/snowflake-api.adapter.ts:276-298` — Snowflake result paging (`fetchResultsByQueryId`, `RESULT_SCAN`) is neither timeout-capped nor signal-cancellable; only the initial `executeQuery` is. The cooperative loop stops *further* paging on abort, but a page already in flight runs uncancelled. Bounded (≤ maxRows), low impact.
- **Question** — `data-marts/data-storage-types/snowflake/adapters/snowflake-api.adapter.ts:125-131`: the type allows `parameters: { STATEMENT_TIMEOUT_IN_SECONDS }`, but please confirm the installed snowflake-sdk actually applies a per-statement `parameters` object server-side; if it silently ignores an unknown key the warehouse cost cap no-ops (the `statement.cancel()` abort path is unaffected). BigQuery's `jobTimeoutMs` is confirmed present on the `Query` type.
- `.changeset/` — no changeset added in this PR. If the new `query_timeout` / `query_cancelled` tool error codes and the timeout/billing behavior count as user-facing, add one (`'owox': minor`).
- minor: `data-marts/use-cases/query-data-mart.service.ts:79-137` — the not-found / access / `verifyCanPerformOperations` work runs *before* the `if (signal?.aborted)` guard at `:137`, so an already-aborted request still does that DB/access work (not billed, just wasted).

</details>

<sub>Automated OWOX review · worktree multi-agent passes (reviewer, logic/billing, architect, tester, fresh-eyes) · the blocker was adversarially verified (independent skeptic + an empirical run of NestJS's `RouteInfoPathExtractor`). Static trace only — no live end-to-end HTTP run; an integration test on a >30s `/mcp` call is the recommended confirmation.</sub>

apps/backend/src/ee/mcp/controllers/mcp-transport.controller.ts
zapolsky: Confirmed — this is correct, and I reproduced it against NestJS 11.1.27's own internals rather than trusting my earlier (wrong) reasoning that the global-prefix exclusion kept the middleware off `/mcp`:

- `RoutesMapper.getRouteInfoFromObject` applies `addLeadingSlash` → the `{*path}` route reaches `RouteInfoPathExtractor` as `/{*path}`.
- For a wildcard route, `extractPathsFrom` appends every `excludedGlobalPrefixRoutes` entry, so the bound paths are `["/api$","/api/{*path}","/mcp","/mcp/(.*)"]` — i.e. the 30s middleware **does** bind `/mcp`.
- `pathToRegexp('/mcp')` matches the POST `/mcp` request URL, so at 30s the middleware sends the generic 408 while `run()` keeps racing and can still bill a SUCCESS the client already saw fail.

Fixed in `5a96594bc`: `/mcp` + `/mcp/{*path}` are now excluded from the 30s operation-timeout middleware (mirroring the `external/{*path}` exclusion), so the app-side 3-min deadline is the binding limiter and returns a clean `query_timeout`. Extracted the exclusion into `MCP_OPERATION_TIMEOUT_EXCLUSIONS` with a regression test that uses NestJS's own `isRouteExcluded` + `pathToRegexp` to assert `/mcp` (and subpaths) are excluded while `/api/...` routes are not.

On the related note (client disconnect not firing `extra.signal`): agreed — in the stateless JSON-POST transport `extra.signal` fires on an MCP `CancelledNotification`/transport `close()`, not on a raw socket drop, so Layer 2 (CANCELLED, not billed) triggers on explicit cancel, not a hard disconnect. Layer 1 (the deadline) is what governs the abandoned-request case, and after this fix it is now authoritative over HTTP. A hard-disconnect → abort bridge is a good follow-up but needs transport-level wiring.

zapolsky:
Thanks for the thorough pass. Addressed the blocker (see the inline thread) plus the two primary non-blocking findings, in `f5f35f025`:

**Warehouse work now cancelled on the deadline and on a rows failure, not only on client abort.** `run()` now owns an internal `AbortController` threaded into both the rows read and the parallel totals; it is aborted by the client signal, the deadline timer, **and** `produce`'s `finally` (so a rows failure cancels the still-in-flight totals). BigQuery `job.cancel()` / Snowflake `statement.cancel()` therefore fire on timeout and on failure, not just on explicit client abort — the readers/connections no longer stay live to the ~3-min warehouse cap. Added tests asserting the work signal is aborted on the deadline and that the totals signal is aborted when the rows read throws.

**BigQuery no longer busy-polls `jobs.get` after an abort.** Once aborted, the loop does exactly one immediate re-poll (to catch the cancelled DONE state fast), then falls back to a bounded, edge-interruptible interval — so it never re-issues `jobs.get` back-to-back. Added a fake-timer regression test asserting exactly two polls before the bounded wait.

**On the Snowflake `parameters` question:** confirmed the installed `snowflake-sdk` applies it server-side — `dist/lib/connection/statement.js:285-286` copies `statementOptions.parameters` into the statement context and `:1049-1051`/`:1103-1105` emit `json.parameters` in the request body, so `STATEMENT_TIMEOUT_IN_SECONDS` reaches the warehouse (not a silent no-op).

**Deliberately not changed (noted for the record):**
- Snowflake result paging (`RESULT_SCAN`) cancellation — bounded (≤ maxRows) and low-impact as you flagged; left as a follow-up.
- `honorsQueryTimeout` at `debug` — a per-request `warn` would be log spam for a normal storage capability gap; a one-time startup enumeration is the better shape and is a reasonable follow-up.
- Env-driving the timeout constants — the team's decision is to keep MCP timeouts as code constants (not env), and the `socket > deadline` invariant is CI-locked.
- No changeset — MCP-internal tool behavior isn't treated as customer-facing here.

Full touched-area suite green (94 suites / 1204 tests), `tsc -p tsconfig.build.json` + ESLint clean.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.ts
```diff
@@ -64,5 +64,30 @@ export interface McpQueryDataMartResponse {
...
+export class QueryTimeoutError extends Error {
+  readonly deadlineMs: number;
+  constructor(deadlineMs: number) {
+    super(`query_data_mart timed out after ${deadlineMs} ms`);
+    this.name = 'QueryTimeoutError';
+    this.deadlineMs = deadlineMs;
+  }
+}
+
+// Client aborted (disconnect / cancel). Recorded CANCELLED, never billed. Stops the server waiting;
```
romandubovyi: 🟡 **Important:** This comment is stale after Phase 3 — `QueryAbortedError` now triggers `workController.abort()`, which threads `signal` into BigQuery/Snowflake adapters and cancels the in-flight warehouse job/statement. Please update the doc to reflect that native cancellation is attempted (for storages that honor it).

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -126,59 +129,131 @@ export class QueryDataMartService {
...
+                );
+                return null;
+              });
+
+          reader = await this.readerResolver.resolve(dataMart.storage.type);
+          // Make the silent gap observable: a cap was requested but this storage drops it, so the
+          // query has no warehouse-side cost cap — only the app-side deadline. Adding a new storage
+          // without honorsQueryTimeout will surface here.
+          if (queryTimeoutMs !== undefined && !reader.honorsQueryTimeout) {
+            this.logger.debug(
```
romandubovyi: 🟡 **Important:** For storages that don't set `honorsQueryTimeout` (Athena, Redshift, Databricks, etc.), the warehouse query keeps running after timeout/abort — only the app-side deadline stops billing, not DWH compute cost. A `debug` log is easy to miss in production. Consider `warn` level or a metric so this gap is observable when MCP queries hit unsupported storages.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -50,17 +59,17 @@ export class QueryDataMartService {
     private readonly dataMartRunService: DataMartRunService,
     private readonly accessDecisionService: AccessDecisionService,
     private readonly projectBalanceService: ProjectBalanceService,
-    private readonly consumptionTrackingService: ConsumptionTrackingService
+    private readonly consumptionTrackingService: ConsumptionTrackingService,
+    @Optional() private readonly queryDeadlineMs: number = DEFAULT_QUERY_DEADLINE_MS
```
romandubovyi, line 63: 💡 **Suggestion:** `@Optional()` on a primitive constructor param is an unconventional NestJS pattern — it works here because tests pass the value directly and DI injects `undefined` (falling back to the default), but a future dev might try to provide it via a token and get confused. A dedicated config provider or injection token would make the test override intent clearer.

apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
```diff
@@ -121,10 +148,30 @@ export class BigQueryApiAdapter {
...
+    return new Promise(resolve => {
+      const onAbort = () => {
+        clearTimeout(timer);
+        resolve();
+      };
+      const timer = setTimeout(() => {
+        signal?.removeEventListener('abort', onAbort);
+        resolve();
+      }, ms);
+      // Fires only on the abort edge; if already aborted it never fires, so the full interval elapses.
```
romandubovyi, line 165: 💡 **Suggestion:** If `signal` is already `aborted` when entering `edgeInterruptibleSleep`, the abort listener won't fire (edge already passed) and the loop waits the full `JOB_POLL_INTERVAL_MS` (~2s) before surfacing the cancellation error. Consider an early `resolve()` when `signal?.aborted` is true on entry to shorten the abort feedback loop.

apps/backend/src/data-marts/data-storage-types/snowflake/adapters/snowflake-api.adapter.ts
```diff
@@ -138,6 +157,15 @@ export class SnowflakeApiAdapter {
           }
         },
       });
+
+      // Best-effort warehouse cancel: on abort, cancel the running statement so a gone client stops
+      // billing compute. The cancel triggers the complete callback with an error → the promise rejects.
+      // `!settled` guards a synchronous complete (would leave a listener attached past resolve).
+      if (signal && !settled) {
+        onAbort = () => statement.cancel(() => undefined);
```
romandubovyi: 💡 **Suggestion:** Consider wrapping `statement.cancel()` in a `try/catch` block. Even though it accepts an asynchronous callback, wrapping it guards against any unexpected synchronous exceptions thrown by the driver if the statement or connection is in an invalid state.

apps/backend/src/data-marts/config/mcp-operation-timeout-exclusions.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+describe('MCP operation-timeout exclusions', () => {
+  // Mirror NestJS's mapToExcludeRoute: each entry compiles to a pathRegex over its leading-slashed path.
+  const compiled = MCP_OPERATION_TIMEOUT_EXCLUSIONS.map(route => ({
+    path: route.path,
+    requestMethod: route.method,
+    pathRegex: pathToRegexp(addLeadingSlash(route.path)).regexp,
+  }));
+
+  const isExcluded = (url: string) =>
+    isRouteExcluded(compiled, url, RequestMethod[RequestMethod.POST]);
```
romandubovyi: 🟡 **Important:** `isRouteExcluded` expects the numeric `RequestMethod` enum (e.g. `RequestMethod.POST === 1`), not the string key. Because all exclusions use `RequestMethod.ALL`, the test happens to pass, but a future per-method exclusion could silently be missed. Pass `RequestMethod.POST` directly.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-report-reader.service.ts
```diff
@@ -181,7 +186,12 @@ export class BigQueryReportReader implements DataStorageReportReader {
   }
 
   private async prepareQueryData(query: string, params?: SqlParameter[]): Promise<void> {
```
romandubovyi, line 188: 🟡 **Important:** The `signal` and `queryTimeoutMs` only reach `executeQuery`, not the subsequent `getJob`/`getRows` calls. If the client aborts or the deadline fires after the job is DONE but while rows are still being fetched, the reader continues fetching. Consider threading the signal into `getRows` or cancelling the reader's work in `QueryDataMartService` too.

apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
```diff
@@ -76,23 +76,49 @@ export class BigQueryApiAdapter {
...
             ...this.getLocationOption(),
+            ...(jobTimeoutMs !== undefined ? { jobTimeoutMs } : {}),
           };
 
     const [job] = await this.bigQuery.createQueryJob(queryConfig);
-    await this.waitForJobToComplete(job);
+
+    // Best-effort warehouse cancel: on abort, ask BigQuery to cancel the job. The poll loop below
+    // then sees the job finish (cancelled) and throws, so we stop billing compute for a gone client.
+    const onAbort = () => {
```
romandubovyi, line 104: 💡 **Suggestion:** The abort handler calls `job.cancel()` unconditionally. If the signal fires after the job has already reached DONE, this still issues a cancel request to BigQuery. Consider guarding it with job-state knowledge (similar to Snowflake's `settled` flag) to avoid an unnecessary API call.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.ts
```diff
@@ -64,5 +64,30 @@ export interface McpQueryDataMartResponse {
...
   listDataMarts(request: McpListDataMartsRequest): Promise<McpListDataMartsResponse>;
   getDataMartDetails(request: McpGetDataMartDetailsRequest): Promise<McpDataMartDetailsResponse>;
-  queryDataMart(request: McpQueryDataMartRequest): Promise<McpQueryDataMartResponse>;
+  queryDataMart(
+    request: McpQueryDataMartRequest,
+    signal?: AbortSignal
+  ): Promise<McpQueryDataMartResponse>;
+}
+
+// Part of the queryDataMart contract: the tool catches these to emit query_timeout / query_cancelled.
```
romandubovyi, line 98: 🔵 **Question:** These errors live in the facade file but are imported by the use-case service, making the service depend on the facade. Would it be cleaner to extract them to a dedicated `mcp-query.errors.ts` module so both the facade and the service can depend on them without an inverted dependency?

apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
```diff
@@ -76,23 +76,49 @@ export class BigQueryApiAdapter {
...
+   * surfaces the cancellation as a thrown error. Unset = no cancellation hook.
    */
-  public async executeQuery(query: string, params?: SqlParameter[]): Promise<{ jobId: string }> {
+  public async executeQuery(
+    query: string,
+    params?: SqlParameter[],
+    jobTimeoutMs?: number,
+    signal?: AbortSignal
+  ): Promise<{ jobId: string }> {
     const queryConfig: Query =
```
romandubovyi: ⚪ **Nit:** The `jobTimeoutMs` spread is duplicated in both branches of the `queryConfig` construction. Consider building the common part once and adding `jobTimeoutMs` after the conditional to reduce duplication.

apps/backend/src/data-marts/services/report-totals.service.ts
```diff
@@ -58,6 +60,9 @@ export class ReportTotalsService {
         blendedDataHeaders: totals.blendedDataHeaders,
         // Row Count / Unique Count are not part of the totals summary.
         rowCount: false,
+        // Only when supplied, so non-MCP callers stay unchanged.
+        ...(queryTimeoutMs !== undefined ? { queryTimeoutMs } : {}),
```
romandubovyi, line 64: 💡 **Suggestion:** The main reader logs when a storage does not honor `queryTimeoutMs`, but `computeTotals` does not. For observability consistency, consider logging the same gap here when `queryTimeoutMs` is supplied and the resolved reader has `honorsQueryTimeout` unset.

romandubovyi:
No blockers. The three-layer defense (app deadline, client abort, warehouse cancel) plus the billing gate is a clean, well-tested design. Review complete.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.ts
zapolsky: Good catch — fixed in ec254c5b2. Updated the QueryAbortedError doc to reflect that abort now threads through `workController` into the BigQuery/Snowflake adapters and asks the warehouse to cancel the in-flight job/statement (for storages that honor it), replacing the stale "does not cancel the warehouse job".

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
zapolsky: Fixed in ec254c5b2 — raised to `warn` so a storage without `honorsQueryTimeout` (Athena/Redshift/Databricks) is visible at prod log level. A query that runs warehouse-side uncapped is a real cost signal worth surfacing per run, not just a debug trace.

apps/backend/src/data-marts/data-storage-types/snowflake/adapters/snowflake-api.adapter.ts
zapolsky: Fixed in ec254c5b2 — `statement.cancel()` is now wrapped in try/catch, so a driver that throws synchronously on an already-gone statement/connection cannot take down the abort listener. Still best-effort: the callback error stays ignored.

apps/backend/src/data-marts/config/mcp-operation-timeout-exclusions.spec.ts
zapolsky: Fixed in ec254c5b2 — the test now passes the numeric `RequestMethod.POST` directly. As you noted it was masked by every exclusion using `RequestMethod.ALL` (so `isRequestMethodAll` short-circuits), but a future per-method exclusion would have been silently mis-tested with the string key.

apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
zapolsky: Fixed in ec254c5b2 — the query config is now built once (query + location + jobTimeoutMs) and `params`/`parameterMode` are appended only when named params are present, so jobTimeoutMs and location are no longer duplicated across the two branches.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
zapolsky: Kept as-is, intentionally. `queryDeadlineMs` is a test-only override with a production default; the team decision is to keep the MCP timeouts as code constants (not env / DI-driven), and the socket > deadline invariant is CI-locked. A dedicated config provider/token would add DI indirection for a constant that is deliberately not runtime-configurable. Happy to revisit if these ever become env-driven.

apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
zapolsky: Deliberately not changed — an early resolve() here would reintroduce the busy-poll this file already guards against. `waitForJobToComplete` does exactly one immediate re-poll after `job.cancel()` (the `repolledAfterAbort` flag) to catch the cancelled→DONE state fast, then falls to `edgeInterruptibleSleep`, which by design waits the full `JOB_POLL_INTERVAL_MS` once the abort edge has passed — that is what bounds the `jobs.get` rate after cancel. Resolving immediately on an already-aborted signal would make the loop re-issue `jobs.get` back-to-back again (the exact regression fixed in an earlier commit). The one-poll fast path already covers the abort-latency case.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-report-reader.service.ts
zapolsky: Acknowledged as a bounded follow-up. The service loop already checks the client signal between pages (`query-data-mart.service.ts:226`) and stops paging on abort, and a single in-flight `getRows` page is bounded to ≤ limit+1 rows (one page). On the deadline path `workController` aborts the underlying `executeQuery`/job. Threading the signal deeper into `getJob`/`getRows` for the rare mid-page abort is real but low-impact (same class as the acknowledged Snowflake RESULT_SCAN paging follow-up), so it is left out of this PR.

apps/backend/src/data-marts/data-storage-types/bigquery/adapters/bigquery-api.adapter.ts
zapolsky: Minor, left as-is. The abort listener is detached synchronously in the `finally` of `executeQuery` immediately after `waitForJobToComplete` resolves (no await in between), so once the job is DONE `onAbort` can no longer fire. The only remaining window is an abort arriving during the very last `getMetadata()` poll that returns DONE, which issues one `jobs.cancel` on an already-done job that BigQuery no-ops. A settled-style flag would add cross-function state for that single harmless no-op, so not worth it here.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.ts
zapolsky: The placement is intentional and documented at `mcp-data-marts.facade.ts:73-74`: these errors are part of the `queryDataMart` facade contract, and the consumer that catches them — the tool at `ee/mcp/tools/query-data-mart.tool.ts:7-9` — imports them from the facade surface, not from use-case internals. The only inverted edge is the use-case importing them back, which is a mild smell. A neutral `mcp-query.errors.ts` that both the facade and use-case import (with the facade re-exporting to keep the tool import stable) is a clean further refactor — I can do that as a follow-up if you prefer it over the current facade-surface placement.

apps/backend/src/data-marts/services/report-totals.service.ts
zapolsky: For the MCP path the gap is already surfaced once per run by the main reader in `query-data-mart.service.ts` (the same `dataMart.storage.type` that `computeTotals` then resolves in the same run), so logging it again inside `computeTotals` would double-log the identical gap. The only other caller, `stream-http-data.service.ts:248`, calls `computeTotals` without `queryTimeoutMs`, so the `queryTimeoutMs !== undefined` guard means there is nothing to log there. Left the single run-level warn as the source of truth.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -18,24 +25,26 @@ import { ConsumptionTrackingService } from '../services/consumption-tracking.ser
...
 
-// Upper bound on rows read per query, independent of the MCP tool's own clamp (query-data-mart.
-// input.ts MAX_LIMIT) — this guards direct service/facade callers that bypass the tool schema.
+// Guards direct facade callers that bypass the tool schema's own clamp.
 const MAX_QUERY_LIMIT = 1000;
 
+// Server-side deadline for one run; on expiry the run fails query_timeout and is not billed. Matches
+// SERVER_TIMEOUT_MS (3 min), so the /mcp controller raises its socket timeout above this or the idle
+// timer would blunt-reset a computing request first. Overridable via constructor for tests.
+export const DEFAULT_QUERY_DEADLINE_MS = 3 * 60_000;
```
romandubovyi, line 43: 💡 **Suggestion:** `DEFAULT_QUERY_DEADLINE_MS` is hardcoded here while `SERVER_TIMEOUT_MS` is env-driven (`DEFAULT_SERVER_TIMEOUT_MS`). The comment says they "match", but if ops changes `SERVER_TIMEOUT_MS` the app deadline won't follow, breaking the documented ordering (`DWH ≈ app < socket < LB`). Consider deriving this from the same constant or injecting `SERVER_TIMEOUT_MS` so the two stay in sync.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -138,6 +143,21 @@ If truncated is true, not all matching rows were returned: narrow the query (few
       return toStructuredToolError('invalid_input', `Invalid query input — ${detail}`);
     }
 
+    if (err instanceof QueryTimeoutError) {
```
romandubovyi, line 146: 💡 **Suggestion:** `QueryTimeoutError` carries `deadlineMs`, but the tool message is static. Including the actual deadline (e.g. `stopped after ${err.deadlineMs} ms`) would give the MCP client a concrete bound to reason about when lightening the query.

romandubovyi (approved):
**Review: approved — no blockers**

Second pass confirms all blocking and important feedback from the first review is addressed: `/mcp` is excluded from the 30s operation-timeout middleware, `QueryAbortedError` docs are current, non-capping storages log at `warn`, and the Snowflake/BigQuery adapter fixes landed.

The three-layer timeout design (app deadline → `workController` → native warehouse cancel) is solid, billing stays success-path only with an audit gate, and the adversarial test coverage gives good confidence.

**Non-blocking notes for follow-up:**
- Athena / Redshift / Databricks still lack `honorsQueryTimeout` — warehouse compute can run uncapped after timeout/abort; the `warn` log makes this observable, but native cancellation for those storages would close the cost gap.
- Two minor suggestions left inline (deadline constant sync with `SERVER_TIMEOUT_MS`, surfacing `deadlineMs` in the `query_timeout` tool message).

LGTM — ready to merge.

PR #1403, opened 2026-07-09
fix(google-ads): share Login Customer ID across auth types

prokopetsv:
# Problem

The Google Ads connector defined **Login Customer ID** separately inside each authentication type (OAuth2 manual credentials and Service Account). Because the field lived under `AuthType.items`, its value was tied to the selected auth type: switching authentication methods dropped the entered value, and the OAuth-button flow had no way to set it at all. It was also marked required for Service Account, forcing a manager-account ID even when connecting directly to an ad account.

# Solution

Promote **Login Customer ID** to a single top-level, optional connector field shared by every authentication type, so its value survives auth-type switches and is reachable from all flows including the OAuth button.

- `packages/connectors/src/Sources/GoogleAds/Source.js` — moved `LoginCustomerId` out of both `AuthType.items` blocks to a top-level field; the request builder now reads `this.config.LoginCustomerId?.value` and falls back to the old nested location so existing connectors keep working. The `login-customer-id` header is still only sent when a value is present and differs from the Customer ID.
- `apps/web/.../ConfigurationStep.tsx` — pinned the GoogleAds `LoginCustomerId` field to priority 1 in the config form so it renders directly under Customer ID.
- Docs (`GETTING_STARTED.md`, `CREDENTIALS.md`) — updated to describe the field as a single optional input, removed the per-auth-type duplicates, and clarified the stats-endpoint note for the empty (direct-access) case.

# Changes

- Moved `LoginCustomerId` from the OAuth2 and Service Account `items` to a shared top-level connector field (now optional).
- Added a backward-compatible read fallback so connectors that stored the value under `AuthType.items` still resolve it.
- Added `placeholder` text to the Customer ID and Login Customer ID fields.
- Pinned GoogleAds `LoginCustomerId` under Customer ID in the web configuration form.
- Removed the duplicate Login Customer ID entries from the OAuth-manual and Service Account doc sections; marked it optional in CREDENTIALS.md and reworded the stats-endpoint notice.
- Added changeset `6740-google-ads-login-customer-id.md`.

Result of the changes


packages/connectors/src/Sources/GoogleAds/Source.js
```diff
@@ -443,7 +439,8 @@ var GoogleAdsSource = class GoogleAdsSource extends AbstractSource {
         requestBody.pageToken = nextPageToken;
       }
       
-      const loginCustomerIdRaw = this.config.AuthType.items?.LoginCustomerId?.value;
+      const loginCustomerIdRaw =
+        this.config.LoginCustomerId?.value || this.config.AuthType.items?.LoginCustomerId?.value;
```
romandubovyi: 🔴 **Blocking:** Using `||` here means an existing connector that previously stored a value under the old nested `AuthType.items.LoginCustomerId` path can never clear Login Customer ID again — setting the new top-level field to an empty string just falls through to the stale nested value. Since clearing this field to go "direct" is an explicitly documented use case, this should distinguish "top-level field explicitly set (even to empty)" from "top-level field never set" — e.g. check `this.config.LoginCustomerId?.value !== undefined` before falling back, or migrate/clear the nested value once the top-level field exists.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
```diff
@@ -230,14 +230,17 @@ export function ConfigurationStep({
...
-  // 1. Required fields without default value
+  // 1. Required fields without default value (and GoogleAds LoginCustomerId, pinned under Customer ID)
   // 2. Required fields with default value
   // 3. Non-required fields without default value
   // 4. All others (non-required with default value)
   const sortedSpecifications = [...connectorSpecification]
     .filter(spec => spec.name !== 'Fields' && !spec.attributes?.includes('HIDE_IN_CONFIG_FORM'))
     .sort((a, b) => {
       const getPriority = (spec: ConnectorSpecificationResponseApiDto) => {
+        if (connector.name === 'GoogleAds' && spec.name === 'LoginCustomerId') {
```
romandubovyi: 🟡 **Important:** Hardcoding `connector.name === 'GoogleAds'` inside a shared, connector-agnostic sorting function doesn't scale — every future connector that needs a field pinned out of its natural priority bucket will add another special case here. Consider making this priority declarative in the connector's own schema (e.g. an explicit `priority`/`PINNED` attribute on the spec returned by `Source.js`) so this component doesn't need to know about specific connectors.

packages/connectors/src/Sources/GoogleAds/GETTING_STARTED.md
```diff
@@ -81,7 +78,7 @@ Leave all other fields as default, then click **Next** to continue.
 4. Click **Finish**, then **Publish Data Mart**.
 
 > ⚠️ **Important Notice:**  
-> If you select any **stats endpoint** (e.g., *Campaign Stats*, *Ad Group Stats*, *Keyword Stats*, etc.), the **Customer ID** and **Login Customer ID** must be **different**.  
+> If you select any **stats endpoint** (e.g., *Campaign Stats*, *Ad Group Stats*, *Keyword Stats*, etc.), the **Customer ID** must be an **ad account**, not your MCC. When you set a **Login Customer ID**, it must be **different** from the Customer ID; when querying the ad account directly, leave **Login Customer ID** empty.
```
romandubovyi: 🟢 **Praise:** Nice rewrite of the stats-endpoint notice — clearly separating "Customer ID must be an ad account" from "Login Customer ID must differ when set" is much clearer than the old wording, and matches the new optional-field behavior well.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
romandubovyi: 🟡 **Important:** Existing connectors store `LoginCustomerId` under `AuthType.items`. When such a connector is edited, the new top-level `LoginCustomerId` field is initialized empty; switching the auth type then drops the old value entirely. Consider copying the legacy `AuthType.items.LoginCustomerId` value into the top-level `LoginCustomerId` during form initialization so the field is pre-filled and survives auth-type changes.

packages/connectors/src/Sources/GoogleAds/Source.js
romandubovyi: 💡 **Suggestion:** Consider using `this.config.AuthType?.items?.LoginCustomerId?.value` to guard against a missing `AuthType` object and avoid a potential runtime error.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
prokopetsv: Fixed. I moved the ordering rule into the connector schema by adding a generic `PINNED` config attribute. `LoginCustomerId` is now marked as pinned in the Google Ads source schema, and the shared configuration sorter only checks for that declarative attribute. This removes the Google Ads-specific condition from the shared UI component and keeps future connector ordering needs schema-driven.

packages/connectors/src/Sources/GoogleAds/Source.js
prokopetsv: Fixed. You were right: using `||` made an intentionally cleared top-level `LoginCustomerId` fall back to the stale nested value.

I changed it to fall back only when the top-level value is `undefined`, so an empty string now correctly means “do not send `login-customer-id`”:

```js
const topLevelLoginCustomerId = this.config.LoginCustomerId?.value;
const loginCustomerIdRaw =
  topLevelLoginCustomerId !== undefined
    ? topLevelLoginCustomerId
    : this.config.AuthType.items?.LoginCustomerId?.value;
```

This keeps legacy configs working while allowing users to clear the new top-level field. `npm run lint -w @owox/connectors` passes.

prokopetsv: Updated. The legacy fallback now uses `this.config.AuthType?.items?.LoginCustomerId?.value` so missing `AuthType` won’t throw while preserving the same fallback behavior when the legacy nested field exists.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
prokopetsv: Fixed. I added a generic form initialization migration that looks for top-level fields now present in the connector spec and copies matching legacy values from the selected `oneOf` config only when the top-level field is `undefined`. This preserves old `AuthType.<variant>.LoginCustomerId` values when editing existing Google Ads connectors and keeps them stable across auth-type changes, without adding a Google Ads-specific branch. Explicitly cleared top-level values are not overwritten.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
```diff
@@ -26,6 +27,38 @@ interface ConfigurationStepProps {
...
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+function migrateNestedConfigValuesToTopLevel(
+  config: Record<string, unknown>,
+  specs: ConnectorSpecificationResponseApiDto[]
+): Record<string, unknown> {
+  const nextConfig = { ...config };
+  const topLevelFields = specs.filter(spec => !spec.oneOf).map(spec => spec.name);
```
romandubovyi: 🟡 **Important:** This migration runs for every connector, not just GoogleAds. Several other connectors (FacebookMarketing, TikTokAds, MicrosoftAds, LinkedInAds, LinkedInPages) already have top-level fields with the same names as their `AuthType.items` fields (e.g. `AppId`, `ClientSecret`, `RefreshToken`) — but those are marked `DEPRECATED` + `HIDE_IN_CONFIG_FORM` legacy remnants. This function will silently copy the live nested value into those hidden deprecated fields on every edit-form load, and it will be included in the saved config on submit. Consider excluding specs with `DEPRECATED`/`HIDE_IN_CONFIG_FORM` attributes from `topLevelFields`, or scoping the migration to the specific field(s) this PR actually needs to migrate.

prokopetsv: Fixed. I kept the migration generic, but narrowed its target set so it only migrates visible, non-deprecated top-level fields. Specs marked DEPRECATED or HIDE_IN_CONFIG_FORM, plus Fields, are now excluded from topLevelFields, so connectors with legacy hidden top-level auth fields will not have nested live values copied into saved config.

romandubovyi (approved):
All review findings addressed: the `||` fallback bug is fixed, the hardcoded connector-name check was generalized into a `PINNED` attribute, and the migration helper is now correctly scoped away from deprecated/hidden legacy fields in other connectors. LGTM.

PR #1407, opened 2026-07-10
6693: add project-based MCP servers

amarchenk0:


apps/backend/src/idp/oauth/oauth-request.validator.ts
```diff
@@ -105,22 +133,29 @@ export class OAuthRequestValidator {
...
+    return {
+      resource,
+      resourceContext: this.validateResource(resource),
+    };
   }
 
-  private optionalResourceOrDefault(value: unknown): string {
-    if (value === undefined || value === null) {
-      return this.config.resource;
+  private rejectClientResource(value: unknown): void {
```
romandubovyi: 🔴 **Blocking:** This unconditionally rejects any authorization/token request where the client includes a `resource` parameter, even when the value matches the host-derived resource. RFC 8707 (which MCP Authorization is built on) requires clients to send `resource` — Claude Desktop, Claude web, and ChatGPT connectors all do. As written, spec-compliant clients sending a correct `resource` value will get a 400 `resource is not supported` instead of succeeding. Consider validating that a client-supplied `resource` matches the request-derived one (and only rejecting on mismatch), rather than rejecting outright.

apps/backend/src/mcp-resource/mcp-resource-resolver.service.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    return {
+      kind: 'project',
+      resource: parsedResource.toString(),
+      publicBaseUrl: `${parsedResource.protocol}//${parsedResource.host}`,
+      projectId,
+    };
+  }
+
+  resolveRequest(request: McpRequestLike): McpResourceContext {
```
romandubovyi, line 60: 🔵 **Question:** `resolveRequest` trusts `X-Forwarded-Host`/`X-Forwarded-Proto` unconditionally. The impact is bounded since the resolved host must still end in the configured `.mcp.owox.com` suffix, but if the origin is ever reachable without going through a proxy that overwrites these headers, a client could set an arbitrary valid `{projectId}.mcp.owox.com` value and get back issuer/metadata URLs or a `WWW-Authenticate` challenge for a project it does not control. Is there an infra-level guarantee (e.g. NestJS `trust proxy` config, or the proxy always stripping client-supplied `X-Forwarded-*`) that these headers cannot be spoofed by a direct client?

apps/backend/src/ee/mcp/controllers/mcp-metadata.controller.ts
```diff
@@ -12,18 +22,32 @@ const OPENAI_APPS_CHALLENGE_PATH = '/.well-known/openai-apps-challenge';
...
-      authorization_servers: [this.config.authorizationServer],
+      resource: resourceContext.resource,
+      authorization_servers: [this.authorizationServerFor(resourceContext)],
       scopes_supported: this.config.scopes,
       resource_documentation: this.config.resourceDocumentationUrl,
     });
   }
 
+  private authorizationServerFor(resourceContext: McpResourceContext) {
+    return resourceContext.kind === 'project'
```
romandubovyi: 🔴 **Blocking:** For shared resources this still advertises `OWOX_AUTH_PUBLIC_BASE_URL` as the authorization server, while authorize/token now derive `resource` only from the request Host (via `tryResolveRequest`). In production those hosts differ (`app…` vs `mcp…`), so a shared-flow client hits the AS host, Host is not a valid MCP resource host, and validation fails with `resource is required` (host-derived is `null`). Local setups where both env URLs are the same hide this. Either point shared `authorization_servers` at the MCP public base (so Host can derive the shared resource), or keep accepting/validating a client `resource` when the request host is the AS issuer rather than an MCP host.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
```diff
@@ -44,34 +65,85 @@ export class OAuthAuthorizationController {
...
+    projects: ReturnType<OAuthProjectSelectionService['filterSelectableProjects']>,
+    selectedProjectId: string | undefined,
+    authorizationRequest: Awaited<
+      ReturnType<OAuthRequestValidator['validateAuthorizationRequest']>
+    >['request'],
+    response: Response
+  ): Promise<McpOAuthProjectMemberContext | null> {
     if (!selectedProjectId && projects.length > 1) {
       response.type('html').send(
         this.projectSelectionService.renderSelectionPage({
```
romandubovyi, line 125: 🟡 **Important:** The project selection page still posts `resource` as a hidden field (`renderHiddenFields` → `resource: request.resource`), but `OAuthRequestValidator` now rejects any client-supplied `resource` with `resource is not supported`. Multi-project shared OAuth will fail on the second authorize round-trip after the user clicks Next. Align the two sides: either drop `resource` from `renderHiddenFields` (if resource is host-only), or stop hard-rejecting a matching client `resource` (if the shared AS host still needs it).

apps/backend/src/ee/mcp/controllers/mcp-metadata.controller.ts
```diff
@@ -12,18 +22,32 @@ const OPENAI_APPS_CHALLENGE_PATH = '/.well-known/openai-apps-challenge';
...
+      ? resourceContext.publicBaseUrl
```
romandubovyi: 🔵 **Question:** For project resources, `authorization_servers` becomes the project MCP host (`{projectId}.mcp…`). That means authorize, token, and the sign-in redirect in `OAuthAuthorizationController` all run on that subdomain (cookies included). Previously the AS was always `OWOX_AUTH_PUBLIC_BASE_URL`. Is the full auth UI (`/auth/sign-in`) and session/refresh-cookie domain model guaranteed to work on `*.mcp.owox.com`, or do we risk breaking the browser login step for project-specific connectors?

apps/backend/src/idp/oauth/oauth-request.validator.spec.ts
```diff
@@ -1,13 +1,24 @@
...
 import { OAuthClientRegistry } from './oauth-client.registry';
 import { OAuthConfigService } from './oauth-config.service';
 import { OAuthRequestValidator } from './oauth-request.validator';
 
-function makeConfig(): OAuthConfigService {
+function makeServices(): {
+  config: OAuthConfigService;
+  resourceResolver: McpResourceResolverService;
+} {
+  const configService = new ConfigService({
```
romandubovyi, line 12: 🟡 **Important:** Specs only cover cases where the request host is already an MCP host (`mcp.owox.com` / `localhost` with both env URLs equal). There is no test for the production split where `OWOX_AUTH_PUBLIC_BASE_URL` ≠ `MCP_PUBLIC_BASE_URL` and authorize/token hit the AS host. That gap is what lets the shared-flow host-derived resource failure slip through. Please add a case that exercises authorize/token validation when Host is the auth base URL and the shared MCP resource must still be accepted (or clearly documents the intended AS host strategy).

apps/backend/src/idp/oauth/oauth-project-selection.service.ts
```diff
@@ -72,6 +72,22 @@ export class OAuthProjectSelectionService {
...
+    provider: IdpProvider,
+    context: AuthorizationContext,
+    projects: Projects,
+    projectId: string
+  ): Promise<McpOAuthProjectMemberContext> {
+    const matches = this.filterSelectableProjects(projects).filter(
+      project => project.id === projectId
+    );
+    if (matches.length !== 1) {
+      throw new BadRequestException('selected project is not available for user');
```
romandubovyi: 💡 **Suggestion:** The error message `"selected project is not available for user"` is reused for both user-selected and host-derived projects. Use a distinct message for the project-host case to make debugging easier.

apps/backend/src/mcp-resource/mcp-resource-resolver.service.ts
```diff
@@ -0,0 +1,144 @@
...
+      parsedResource.pathname !== '/mcp' ||
+      parsedResource.search !== '' ||
+      parsedResource.hash !== '' ||
+      !parsedResource.hostname.endsWith(projectSuffix)
+    ) {
+      throw new McpInvalidResourceError();
+    }
+
+    const projectId = parsedResource.hostname.slice(0, -projectSuffix.length);
+    if (!isMcpProjectId(projectId) || projectId.includes('.')) {
```
romandubovyi: ⚪ **Nit:** `projectId.includes(\u0027.\u0027)` is redundant because `isMcpProjectId` already restricts the value to `/^[a-f0-9]{32}$/`.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
```diff
@@ -44,34 +65,85 @@ export class OAuthAuthorizationController {
...
+      authorizationRequest,
+      projectMember
+    );
+    const redirectUrl = new URL(authorizationRequest.redirectUri);
+    redirectUrl.searchParams.set('code', authorizationCode.code);
+    redirectUrl.searchParams.set('state', authorizationRequest.state);
+    response.redirect(redirectUrl.toString());
+  }
+
+  private async resolveProjectHostMember(
```
romandubovyi: 💡 **Suggestion:** `resolveProjectHostMember` is `async` but does not `await` the returned Promise. Remove `async` or add `await` for consistency.

romandubovyi:
Review complete. All relevant tests pass.

PR #1409, opened 2026-07-10
feat(mcp): observability instrumentation, conversation grouping, layer-B decoupling

zapolsky:
## What

Instruments every MCP tool call (one structured event → logs/BQ), groups a conversation's calls into a single OTLP trace (root + child) on GCP Cloud Trace, and refactors `internal-helpers` into an MCP-agnostic delivery layer.

## Highlights

- **Instrumentation** — one `mcp.tool_call` event per call at the SDK `registerTool` choke point; best-effort, never on the tool's response path (a logging failure can't fail a call).
- **PII-safe by default** — tool arguments/results/SQL stay out of general logs: dropped without a bucket, offloaded to GCS (ADC-gated) with one, inlined only via `MCP_LOG_INLINE_PAYLOADS=true`. Event keys are BQ-safe (underscore); GCS records are self-describing and `LOAD DATA`-able.
- **Conversation grouping (OTLP)** — deterministic root+child spans keyed by `conversation_id` (from `_meta`) or a daily pseudo-id; env-gated; GCP Cloud Trace dedups the re-emitted root.
- **Layer-B decoupling** — generic `PayloadOffloader(pathBuilder)` + `OtlpTransport(spanMapper)`; all MCP wiring injected at the composition root via an optional `BUS_EXTRAS` DI token (no `common → ee` import).
- Env surface: `MCP_OTEL_ENABLED`, `MCP_LOG_*`, `OTEL_EXPORTER_OTLP_*`. Laminar/gRPC removed (Cloud Trace dedups; Laminar does not). OTel/GCS are optional peer deps — community installs don't pull them.

## Testing

- `internal-helpers`: 71 tests, clean under `--detectOpenHandles`.
- `backend`: 5145 tests pass, `nest build` clean.

## Known follow-ups (not blockers)

- `executedSql` CLS race across JSON-RPC batches (single calls unaffected; batches deprecating in the MCP spec).
- Conversation grouping degrades to per-user/day for clients that send no `_meta` (Claude) — needs a product decision on session keying.
- Schema-invalid calls (rejected by the SDK before the handler) emit no event — documented known gap.
- `meta_*` client hints still inline; assorted low-severity items tracked separately.

No changeset — MCP-internal observability.


apps/backend/src/ee/mcp/controllers/mcp-transport.controller.ts
```diff
@@ -27,52 +33,94 @@ export class McpTransportController {
...
   private getRequestedSessionId(request: McpAuthenticatedRequest): string | undefined {
-    const header = request.headers?.['mcp-session-id'];
-    return Array.isArray(header) ? header[0] : header;
+    return this.firstHeader(request, 'mcp-session-id');
+  }
+
+  private extractRequestMeta(body: unknown): Record<string, unknown> | undefined {
+    if (Array.isArray(body)) {
+      for (const item of body) {
+        const meta = this.extractRequestMeta(item);
```
amarchenk0: A JSON-RPC batch does not have request-wide `_meta`: this picks the first message’s metadata and stores it once in CLS, so every tool call in a batch is attributed to that first conversation. Extract metadata per JSON-RPC message/callback (and add a mixed-conversation batch test) to preserve correct grouping.

amarchenk0 (changes requested):
Please address the blocking issue in the inline comment.

apps/backend/src/ee/mcp/controllers/mcp-transport.controller.ts
zapolsky: Fixed in a7bc91771. `_meta` is now read per call from the SDK `extra` — the protocol layer sets `extra._meta = request.params._meta` for each JSON-RPC message, so each call in a batch is attributed to its own conversation. Dropped the request-wide `MCP_REQUEST_META_KEY` CLS extraction that took the first message's `_meta`.

Added a mixed-conversation batch test in `mcp-call-instrumentation.spec.ts` (two calls with different `_meta` → different `owox_conversation_id`).

amarchenk0 (approved):

PR #1411, opened 2026-07-13
feat(tiktok-ads): let Data Level choose insights reporting grain

prokopetsv:
# Problem

TikTok Ads `ad_insights` and `ad_insights_by_country` always required `ad_id` + `stat_time_day` as unique keys, no matter which **Data Level** was selected. The field selector pinned `ad_id` and disabled its checkbox, so advertiser-, campaign-, and ad-group-level (date-only) reporting was impossible — even though the connector already sent the correct dimensions to TikTok for those levels. Any run at a non-ad level failed with `Missing required unique fields for endpoint 'ad_insights'. Missing fields: ad_id`. The root cause: the connector, the backend field-schema DTO, and the web field selector all read a single static `uniqueKeys` from the schema instead of one that varies by Data Level. Data Level also sat under Advanced settings, making the coupling easy to miss.

# Solution

The unique-key set for the insights nodes is now derived from the selected Data Level everywhere it matters, and that mapping is surfaced to the UI so it can pin the right fields.

- **Connector (`Source.js`, `Connector.js`)**: added `getUniqueKeysForNode(node, dataLevel)` as the single source of truth (advertiser→`stat_time_day`, campaign→`campaign_id,stat_time_day`, etc.), and overrode `getFieldsSchema()` to expose a `uniqueKeysByDataLevel` map. Storage creation and the fetch-time "missing required fields" guard now use these derived keys, so the destination key structure and the validation match the actual grain.
- **Backend**: threaded `uniqueKeysByDataLevel` through the zod field schema, the service mapper, the connector mapper, and the response DTO so the value survives every hop to the client.
- **Web**: the field selector pins the correct keys for the chosen Data Level (falling back to static keys for other connectors), shows an inline tip explaining the dependency, and unions the required keys into the saved field list on configuration-only edits. Data Level moved out of Advanced into the main settings as a dropdown, with a warning about changing grain on data that already exists.
- **Docs**: updated the TikTok README/GETTING_STARTED, the setup guide, and the connector-authoring guide.

Known trade-offs (documented, not enforced): changing Data Level on a destination that already has data can misalign merges because the table's primary key is fixed at creation while the merge key now varies — the field description and docs steer users to a new Data Mart instead. Config-only save is best-effort if the field schema has not finished loading.

# Changes

- Added `getUniqueKeysForNode` and a `getFieldsSchema` override exposing `uniqueKeysByDataLevel` in TikTok Ads `Source.js`
- Switched TikTok `Connector.js` storage creation and `Source.js` fetch validation to data-level-derived unique keys
- Made `DataLevel` a dropdown (via `options`), moved it out of Advanced settings, and rewrote its description to warn about changing grain on existing data
- Threaded `uniqueKeysByDataLevel` through the backend zod schema, `ConnectorService` mapper, `ConnectorMapper`, and web response DTO
- Updated `FieldsSelectionStep` to pin fields by selected Data Level and show an inline explanatory tip
- Unioned required fields into the saved field list on configuration-only edits in `ConnectorEditForm` and `ConnectorDefinitionField`
- Updated TikTok Ads `README.md` / `GETTING_STARTED.md`, the connector setup guide, and `CREATING_CONNECTOR.md`; fixed a broken self-link to point to `CREDENTIALS.md`

Drop down list

Note with current Data Level

### DIFF
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
@@ -341,6 +341,7 @@ export function ConnectorEditForm({
               connectorFields={connectorFields}
               selectedField={selectedNode}
               selectedFields={selectedFields}
+              configuration={connectorConfiguration}
               onFieldToggle={handleFieldToggle}
               onSelectAllFields={handleSelectAllFields}
             />
@@ -394,6 +395,7 @@ export function ConnectorEditForm({
             connectorFields={connectorFields}
             selectedField={selectedNode}
             selectedFields={selectedFields}
+            configuration={connectorConfiguration}
             onFieldToggle={handleFieldToggle}
             onSelectAllFields={handleSelectAllFields}
           />
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
@@ -14,7 +14,7 @@ import type { ConnectorConfig } from '../../../../data-marts/edit';
 import type { ConnectorFieldsResponseApiDto } from '../../../shared/api';
 import { AppWizard, AppWizardLayout, AppWizardActions } from '@owox/ui/components/common/wizard';
 import { trackEvent } from '../../../../../utils';
-import { DATA_LEVEL_CONFIG_KEY } from '../../../shared/constants/connector-config';
+import { resolveEffectiveDataLevel } from '../../../shared/constants/connector-config';
 
 interface ConnectorEditFormProps {
   onSubmit: (connector: ConnectorConfig) => void;
@@ -178,7 +178,12 @@ export function ConnectorEditForm({
     selectedConnector,
   ]);
 
-  // Union the persisted fields with whatever the chosen DataLevel requires (e.g. TikTok
+  const effectiveDataLevel = useMemo(
+    () => resolveEffectiveDataLevel(connectorConfiguration, connectorSpecification),
+    [connectorConfiguration, connectorSpecification]
+  );
+
+  // Union the persisted fields with whatever the effective DataLevel requires (e.g. TikTok
   // ad_insights needs ad_id at AUCTION_AD). No-op for nodes without uniqueKeysByDataLevel.
   // Falls back to the unchanged fields if connectorFields hasn't loaded yet, but that
   // window isn't user-reachable: loadFieldsSafely is always triggered for an existing
@@ -186,15 +191,14 @@ export function ConnectorEditForm({
   // the <StepNavigation> call site below until that fetch settles.
   const fieldsForSave = useMemo(() => {
     const fields = existingConnector?.source.fields ?? selectedFields;
-    const dataLevel = connectorConfiguration[DATA_LEVEL_CONFIG_KEY];
-    if (typeof dataLevel !== 'string') return fields;
+    if (!effectiveDataLevel) return fields;
 
     const node = existingConnector?.source.node ?? selectedNode;
     const required = connectorFields?.find(f => f.name === node)?.uniqueKeysByDataLevel?.[
-      dataLevel
+      effectiveDataLevel
     ];
     return required?.length ? Array.from(new Set([...fields, ...required])) : fields;
-  }, [existingConnector, selectedFields, selectedNode, connectorConfiguration, connectorFields]);
+  }, [existingConnector, selectedFields, selectedNode, effectiveDataLevel, connectorFields]);
 
   const handleConnectorSelect = (connector: ConnectorListItem) => {
     setSelectedConnector(connector);
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
@@ -14,6 +14,7 @@ import type { ConnectorConfig } from '../../../../data-marts/edit';
 import type { ConnectorFieldsResponseApiDto } from '../../../shared/api';
 import { AppWizard, AppWizardLayout, AppWizardActions } from '@owox/ui/components/common/wizard';
 import { trackEvent } from '../../../../../utils';
+import { DATA_LEVEL_CONFIG_KEY } from '../../../shared/constants/connector-config';
 
 interface ConnectorEditFormProps {
   onSubmit: (connector: ConnectorConfig) => void;
@@ -179,12 +180,13 @@ export function ConnectorEditForm({
 
   // Union the persisted fields with whatever the chosen DataLevel requires (e.g. TikTok
   // ad_insights needs ad_id at AUCTION_AD). No-op for nodes without uniqueKeysByDataLevel.
-  // ponytail: best-effort — if fields haven't loaded yet the run fails loudly with
-  // "Missing required unique fields"; acceptable since changing DataLevel on existing data
-  // is already discouraged in the field's description.
+  // Falls back to the unchanged fields if connectorFields hasn't loaded yet, but that
+  // window isn't user-reachable: loadFieldsSafely is always triggered for an existing
+  // connector, and the Save button is disabled via isLoading={... || loadingFields} at
+  // the <StepNavigation> call site below until that fetch settles.
   const fieldsForSave = useMemo(() => {
     const fields = existingConnector?.source.fields ?? selectedFields;
-    const dataLevel = connectorConfiguration.DataLevel;
+    const dataLevel = connectorConfiguration[DATA_LEVEL_CONFIG_KEY];
     if (typeof dataLevel !== 'string') return fields;
 
     const node = existingConnector?.source.node ?? selectedNode;
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
@@ -2,7 +2,6 @@ import { useConnector } from '../../../shared/model/hooks/useConnector';
 import type { ConnectorListItem } from '../../../shared/model/types/connector';
 import { useEffect, useState, useCallback, useMemo } from 'react';
 import { DataStorageType } from '../../../../data-storage';
-import { Alert, AlertDescription } from '@owox/ui/components/alert';
 import {
   ConnectorSelectionStep,
   ConfigurationStep,
@@ -15,9 +14,6 @@ import type { ConnectorConfig } from '../../../../data-marts/edit';
 import type { ConnectorFieldsResponseApiDto } from '../../../shared/api';
 import { AppWizard, AppWizardLayout, AppWizardActions } from '@owox/ui/components/common/wizard';
 import { trackEvent } from '../../../../../utils';
-
-const DATA_LEVEL_DEFAULT = 'AUCTION_AD';
-const DATA_LEVEL_DEPENDENT_TIKTOK_NODES = new Set(['ad_insights', 'ad_insights_by_country']);
 
 interface ConnectorEditFormProps {
   onSubmit: (connector: ConnectorConfig) => void;
@@ -87,19 +83,6 @@ export function ConnectorEditForm({
   );
 
   const totalSteps = steps.length;
-
-  const getDataLevel = (configuration?: Record<string, unknown>) => {
-    return typeof configuration?.DataLevel === 'string'
-      ? configuration.DataLevel
-      : DATA_LEVEL_DEFAULT;
-  };
-
-  const isDataLevelDependentTikTokNode = (connectorName?: string, nodeName?: string) => {
-    return (
-      connectorName === 'TikTokAds' &&
-      Boolean(nodeName && DATA_LEVEL_DEPENDENT_TIKTOK_NODES.has(nodeName))
-    );
-  };
 
   const loadSpecificationSafely = useCallback(
     async (connectorName: string) => {
@@ -171,19 +154,8 @@ export function ConnectorEditForm({
         setSelectedConnector(existingConnectorDef);
 
         void loadSpecificationSafely(existingConnectorDef.name);
-        if (!configurationOnly && mode !== 'fields-only') {
-          void loadFieldsSafely(existingConnectorDef.name);
-        }
-        if (mode === 'fields-only') {
-          void loadFieldsSafely(existingConnectorDef.name);
-        }
-        if (
-          configurationOnly &&
-          source.configuration.length > 0 &&
-          isDataLevelDependentTikTokNode(existingConnectorDef.name, source.node)
-        ) {
-          void loadFieldsSafely(existingConnectorDef.name);
-        }
+        // Fields power both the Fields step and the data-level reconciliation at save.
+        void loadFieldsSafely(existingConnectorDef.name);
       }
     }
 
@@ -205,42 +177,22 @@ export function ConnectorEditForm({
     selectedConnector,
   ]);
 
-  const selectedNodeForConfiguration = existingConnector?.source.node ?? selectedNode;
-  const currentDataLevel = getDataLevel(connectorConfiguration);
-  const originalDataLevel = getDataLevel(existingConnector?.source.configuration[0]);
-  const isDataLevelReconciliationRelevant =
-    configurationOnly &&
-    Boolean(existingConnector?.source.configuration.length) &&
-    isDataLevelDependentTikTokNode(existingConnector?.source.name, selectedNodeForConfiguration);
-  const hasDataLevelChanged =
-    isDataLevelReconciliationRelevant && currentDataLevel !== originalDataLevel;
-  const requiredFieldsForCurrentDataLevel = useMemo(() => {
-    if (!hasDataLevelChanged) return [];
-
-    const selectedNodeSchema = connectorFields?.find(
-      field => field.name === selectedNodeForConfiguration
-    );
-
-    return selectedNodeSchema?.uniqueKeysByDataLevel?.[currentDataLevel] ?? [];
-  }, [connectorFields, currentDataLevel, hasDataLevelChanged, selectedNodeForConfiguration]);
-  const fieldsAfterDataLevelReconciliation = useMemo(() => {
+  // Union the persisted fields with whatever the chosen DataLevel requires (e.g. TikTok
+  // ad_insights needs ad_id at AUCTION_AD). No-op for nodes without uniqueKeysByDataLevel.
+  // ponytail: best-effort — if fields haven't loaded yet the run fails loudly with
+  // "Missing required unique fields"; acceptable since changing DataLevel on existing data
+  // is already discouraged in the field's description.
+  const fieldsForSave = useMemo(() => {
     const fields = existingConnector?.source.fields ?? selectedFields;
-    if (!hasDataLevelChanged || requiredFieldsForCurrentDataLevel.length === 0) {
-      return fields;
-    }
+    const dataLevel = connectorConfiguration.DataLevel;
+    if (typeof dataLevel !== 'string') return fields;
 
-    return Array.from(new Set([...fields, ...requiredFieldsForCurrentDataLevel]));
-  }, [
-    existingConnector?.source.fields,
-    hasDataLevelChanged,
-    requiredFieldsForCurrentDataLevel,
-    selectedFields,
-  ]);
-  const fieldsAddedForDataLevel = useMemo(() => {
-    const existingFields = new Set(existingConnector?.source.fields ?? selectedFields);
-    return requiredFieldsForCurrentDataLevel.filter(fieldName => !existingFields.has(fieldName));
-  }, [existingConnector?.source.fields, requiredFieldsForCurrentDataLevel, selectedFields]);
-  const isDataLevelReconciliationPending = hasDataLevelChanged && connectorFields === null;
+    const node = existingConnector?.source.node ?? selectedNode;
+    const required = connectorFields?.find(f => f.name === node)?.uniqueKeysByDataLevel?.[
+      dataLevel
+    ];
+    return required?.length ? Array.from(new Set([...fields, ...required])) : fields;
+  }, [existingConnector, selectedFields, selectedNode, connectorConfiguration, connectorFields]);
 
   const handleConnectorSelect = (connector: ConnectorListItem) => {
     setSelectedConnector(connector);
@@ -336,9 +288,7 @@ export function ConnectorEditForm({
 
   const canGoNext = () => {
     if (configurationOnly) {
-      return (
-        selectedConnector !== null && configurationIsValid && !isDataLevelReconciliationPending
-      );
+      return selectedConnector !== null && configurationIsValid;
     }
 
     if (mode === 'fields-only') {
@@ -480,23 +430,7 @@ export function ConnectorEditForm({
 
   return (
     <AppWizard>
-      <AppWizardLayout>
-        {renderCurrentStep()}
-
-        {hasDataLevelChanged && (
-          <Alert className='bg-background'>
-            <AlertDescription>
-              Data Level changed from <span className='font-medium'>{originalDataLevel}</span> to{' '}
-              <span className='font-medium'>{currentDataLevel}</span>.
-              {isDataLevelReconciliationPending
-                ? ' Loading field requirements before saving.'
-                : fieldsAddedForDataLevel.length > 0
-                  ? ` OWOX will also keep ${fieldsAddedForDataLevel.join(', ')} selected so this configuration can run with the new Data Level.`
-                  : ' The selected fields already include the required fields for the new Data Level.'}
-            </AlertDescription>
-          </Alert>
-        )}
-      </AppWizardLayout>
+      <AppWizardLayout>{renderCurrentStep()}</AppWizardLayout>
 
       <AppWizardActions variant='horizontal'>
         <StepNavigation
@@ -514,7 +448,7 @@ export function ConnectorEditForm({
                   name: selectedConnector.name,
                   configuration: [connectorConfiguration],
                   node: existingConnector?.source.node ?? selectedNode,
-                  fields: fieldsAfterDataLevelReconciliation,
+                  fields: fieldsForSave,
                 },
                 storage: existingConnector?.storage ?? {
                   fullyQualifiedName: existingConnector?.storage.fullyQualifiedName ?? '',
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
@@ -2,6 +2,7 @@ import { useConnector } from '../../../shared/model/hooks/useConnector';
 import type { ConnectorListItem } from '../../../shared/model/types/connector';
 import { useEffect, useState, useCallback, useMemo } from 'react';
 import { DataStorageType } from '../../../../data-storage';
+import { Alert, AlertDescription } from '@owox/ui/components/alert';
 import {
   ConnectorSelectionStep,
   ConfigurationStep,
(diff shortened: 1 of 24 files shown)


packages/connectors/src/Sources/TikTokAds/Source.js
```diff
@@ -5,6 +5,8 @@
  * file that was distributed with this source code.
  */
 
+var TIKTOK_ADS_DATA_LEVELS = ["AUCTION_ADVERTISER", "AUCTION_CAMPAIGN", "AUCTION_ADGROUP", "AUCTION_AD"];
```
romandubovyi, line 8: 🟡 **Important:** This PR introduces significant data-level-dependent logic but doesn't include tests. Please add unit tests for `getDimensionsForDataLevel`, `getUniqueKeysForNode`, and `getFieldsSchema`, plus an integration test verifying that the field selector pins the correct unique keys per Data Level.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/FieldsSelectionStep.tsx
```diff
@@ -49,10 +51,18 @@ export function FieldsSelectionStep({
...
-  );
+  const dataLevel = configuration?.DataLevel;
+  const selectedDataLevel = typeof dataLevel === 'string' ? dataLevel : null;
+  const uniqueKeys = useMemo(() => {
+    if (selectedDataLevel && selectedFieldData?.uniqueKeysByDataLevel?.[selectedDataLevel]) {
+      return selectedFieldData.uniqueKeysByDataLevel[selectedDataLevel];
+    }
+    return selectedFieldData?.uniqueKeys ?? [];
+  }, [selectedFieldData, selectedDataLevel]);
+  const showDataLevelFieldsTip =
```
romandubovyi: 🔵 **Question:** Why hardcode `ad_insights` and `ad_insights_by_country` here? If the tip is gated only on the presence of `uniqueKeysByDataLevel[selectedDataLevel]`, the component stays reusable for any connector that provides per-data-level unique keys.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/form/ConnectorDefinitionField.tsx
```diff
@@ -107,6 +107,7 @@ export function ConnectorDefinitionField({
             source: {
               ...currentDefinition.connector.source,
               configuration: updatedConfigurations,
+              fields: connector.source.fields,
```
romandubovyi: 🟡 **Important:** Updating configuration now silently overwrites the user's field selection. This is needed for Data Level changes, but it will also reset any custom field choices when unrelated configuration parameters are edited. Consider gating the `fields` overwrite to cases where `DataLevel` actually changed, or warn the user that fields will be reconciled.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
```diff
@@ -181,6 +177,23 @@ export function ConnectorEditForm({
...
   ]);
 
+  // Union the persisted fields with whatever the chosen DataLevel requires (e.g. TikTok
+  // ad_insights needs ad_id at AUCTION_AD). No-op for nodes without uniqueKeysByDataLevel.
+  // ponytail: best-effort — if fields haven't loaded yet the run fails loudly with
+  // "Missing required unique fields"; acceptable since changing DataLevel on existing data
+  // is already discouraged in the field's description.
+  const fieldsForSave = useMemo(() => {
+    const fields = existingConnector?.source.fields ?? selectedFields;
+    const dataLevel = connectorConfiguration.DataLevel;
```
romandubovyi: 💡 **Suggestion:** `DataLevel` is referenced as a magic string in both `ConnectorEditForm` and `FieldsSelectionStep`. Consider exporting a constant from the TikTok connector types and using it everywhere to avoid typos and make refactors safer.

romandubovyi (commented):
Review complete. Verified the TikTok Ads Data Level flow end-to-end from connector source through backend to web field selector.

packages/connectors/src/Sources/TikTokAds/Source.js
```diff
@@ -360,9 +361,56 @@ var TikTokAdsSource = class TikTokAdsSource extends AbstractSource {
     return dimensions;
   }
 
+  /**
+   * Get unique key fields for a node, accounting for data-level dependent dimensions
```
romandubovyi: 🟢 **Praise:** The data-level dimension logic is cleanly factored — `getDimensionsForDataLevel`, `getUniqueKeysForNode`, and `getFieldsSchema` each have a single responsibility, and `populateDimensions` avoids duplicating the country-code handling.

packages/connectors/src/Sources/TikTokAds/README.md
```diff
@@ -6,6 +6,31 @@ The **TikTok Ads Source** allows you to transfer raw data from TikTok advertisin
...
+## Available Data
+
+| Node | Description | Destination table |
+| --- | --- | --- |
+| `advertiser` | Advertiser account details, including name, company, and currency. | `tiktok_ads_advertiser` |
+| `campaigns` | Campaign settings, status, budget, and schedule. | `tiktok_ads_campaigns` |
+| `ad_groups` | Ad group settings, bid strategy, optimization goal, placement, and targeting schedule. | `tiktok_ads_ad_groups` |
+| `ads` | Individual ad creatives, statuses, and campaign/ad group links. | `tiktok_ads_ads` |
+| `ad_insights` | Daily performance metrics at the selected TikTok Ads data level. | `tiktok_ads_ad_insights` |
+| `ad_insights_by_country` | Daily performance metrics at the selected TikTok Ads data level, broken down by country. | `tiktok_ads_ad_insights_by_country` |
```
romandubovyi, line 18: ⚪ **Nit:** This Data Level table duplicates the one in `GETTING_STARTED.md`. Consider keeping it in one place and linking to it to avoid drift.

romandubovyi (commented):
(Retracted — this was an accidental test submission while verifying comment anchor lines. See below for the actual review.)

packages/connectors/src/Sources/TikTokAds/Source.js
```diff
@@ -360,9 +361,56 @@ var TikTokAdsSource = class TikTokAdsSource extends AbstractSource {
...
+   *
+   * @param {string} nodeName - The node name (e.g. ad_insights, ad_insights_by_country)
+   * @param {string} dataLevel - The reporting data level (only relevant for insights nodes)
+   * @return {array} - Array of unique key fields
+   */
+  getUniqueKeysForNode(nodeName, dataLevel) {
```
romandubovyi, line 378: 🔴 **Blocking:** `getUniqueKeysForNode('ad_insights', 'AUCTION_ADVERTISER')` returns only `['stat_time_day']`. Since `AdvertiserIDs` supports multiple advertisers writing to the same destination table (see `Connector.js` `startImportProcessOfTimeSeriesData`, which loops advertisers and reuses one cached storage per node), and `GoogleBigQueryStorage.buildMergeQuery` builds its `MERGE ... ON` clause from exactly these unique keys, two advertisers' rows for the same date will collide and overwrite each other under `AUCTION_ADVERTISER`. `campaign_id`/`adgroup_id`/`ad_id` are platform-global so the other levels are fine, but the advertiser level has no natural unique dimension besides `advertiser_id` itself — it should be included in the key when multiple advertisers can be configured together.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/FieldsSelectionStep.tsx
romandubovyi: 🟡 **Important:** Hardcoding `selectedField === 'ad_insights' || selectedField === 'ad_insights_by_country'` couples this generic, connector-agnostic step to TikTok's specific node names. The check is also redundant — `Boolean(selectedFieldData?.uniqueKeysByDataLevel?.[selectedDataLevel])` alone already implies "this node varies by Data Level," regardless of which connector or node it is. Suggest dropping the name check so the component stays reusable for any future connector with the same schema shape.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
```diff
@@ -181,6 +177,23 @@ export function ConnectorEditForm({
     selectedConnector,
   ]);
 
+  // Union the persisted fields with whatever the chosen DataLevel requires (e.g. TikTok
+  // ad_insights needs ad_id at AUCTION_AD). No-op for nodes without uniqueKeysByDataLevel.
+  // ponytail: best-effort — if fields haven't loaded yet the run fails loudly with
```
romandubovyi: 🔵 **Question:** The `ponytail:` comment calls this best-effort if `connectorFields` hasn't loaded yet — the run then fails later with "Missing required unique fields" instead of failing at save time. Since `loadFieldsSafely` is now always triggered (good fix), would it be worth gating the "Finish" button on `!loadingFields` for configuration-only mode, so users get immediate feedback instead of a failed run later? Or is the current UX (fail fast on next run) intentional given how rare the race is?

packages/connectors/src/Sources/TikTokAds/Source.js
```diff
@@ -394,7 +442,10 @@ var TikTokAdsSource = class TikTokAdsSource extends AbstractSource {
 
     // Validate that required unique fields are included
     if (this.fieldsSchema[nodeName].uniqueKeys) {
-      const uniqueKeys = this.fieldsSchema[nodeName].uniqueKeys;
+      const nodeDataLevel = (nodeName === 'ad_insights' || nodeName === 'ad_insights_by_country')
+        ? this.getValidatedDataLevel()
+        : null;
+      const uniqueKeys = this.getUniqueKeysForNode(nodeName, nodeDataLevel);
```
romandubovyi, line 449: 💡 **Suggestion:** Nice fix for the fetch-time guard. One related loose end: `castFields()` further down (search for "First ensure uniqueKey fields are always included") still reads the static `schema[nodeName].uniqueKeys` rather than the Data Level-derived keys from `getUniqueKeysForNode`. It's harmless today (the next loop in `castFields` copies all schema-defined fields regardless), but it's now inconsistent with the rest of this file's data-level-aware unique key handling. Consider either removing that now-redundant block or threading the resolved keys through for clarity.

romandubovyi (commented):
Review complete. Solid fix for a real limitation (Data Level was always forcing ad_id regardless of grain), with thorough docs. Found one likely data-loss bug for the AUCTION_ADVERTISER grain when multiple Advertiser IDs are configured, plus a few smaller points.

packages/connectors/src/Sources/TikTokAds/README.md
```diff
@@ -6,6 +6,31 @@ The **TikTok Ads Source** allows you to transfer raw data from TikTok advertisin
...
+| `audiences` | Custom audiences, including type, size, validity status, and expiration. | `tiktok_ads_audiences` |
+
+## Data Level for Performance Nodes
```
romandubovyi, line 21: 🟢 **Praise:** Really solid documentation work here — the Data Level → required-keys tables for both endpoints, the inline UI tip in `FieldsSelectionStep`, and the explicit warning about changing Data Level on existing data (in the field description, docs, and UI) all reinforce each other. This is the kind of trade-off communication that's easy to skip on a backend-focused fix.

packages/connectors/src/Sources/TikTokAds/Source.js
prokopetsv: Good catch — added unit tests for all three (`getDimensionsForDataLevel`, `getUniqueKeysForNode`, `getFieldsSchema`) in `packages/connectors/test/Sources/TikTokAds/Source.test.js`, plus an integration test in `FieldsSelectionStep.test.tsx` verifying the field selector pins the right unique keys per Data Level (and falls back correctly when Data Level isn't set). Note: `packages/connectors` had no test runner at all before this, so I added `vitest` (already used elsewhere in the repo) along with a small loader for this package's GAS-style source files, since they use bare `class X extends Y` with no imports/exports.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/FieldsSelectionStep.tsx
prokopetsv: Right — that node-name check was redundant leftover from when I was building this against TikTok specifically. Removed it; `showDataLevelFieldsTip` is now gated only on `uniqueKeysByDataLevel` presence, so it's generic for any connector/node. Added a test with a synthetic non-TikTok node name confirming the tip and pinning both work without the hardcoded list.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/form/ConnectorDefinitionField.tsx
prokopetsv: Good catch on the local reasoning — traced it and today `connector.source.fields` arriving here is always the additive-only union from `ConnectorEditForm`'s `fieldsForSave` (never removes fields, regardless of which parameter triggered the save), so there's no live data loss. But you're right that this line shouldn't rely on that invisibly — updated it to union with `currentDefinition.connector.source.fields` locally, so this file guarantees the "never drops selected fields" property itself rather than trusting a sibling component to keep guaranteeing it. Didn't add a warning since nothing is destructive anymore either way.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
prokopetsv: Done, but I put the constant in `apps/web/.../connectors/shared/constants/connector-config.ts` instead of TikTok-specific types — `FieldsSelectionStep` and `ConnectorEditForm` are intentionally connector-agnostic (per the earlier "why hardcode ad_insights" comment), so importing from a TikTok types module would re-couple them to one connector. The constant documents `DataLevel` as a convention key any connector's schema can use, not a TikTok identifier.

packages/connectors/src/Sources/TikTokAds/README.md
prokopetsv: Good catch — moved the table to live only in README.md (matches how other connectors like XAds keep reference tables there), and GETTING_STARTED.md now links to `README.md#data-level-for-performance-nodes` instead of repeating it.

packages/connectors/src/Sources/TikTokAds/Source.js
prokopetsv: Confirmed — real gap, thanks for catching it before it shipped. Added `advertiser_id` to the merge/unique-key set for both insights nodes at every Data Level (not just `AUCTION_ADVERTISER`): it costs nothing at the other levels since it's redundant-but-harmless there, and it stops relying on "IDs are platform-global" being true forever. `advertiser_id` doesn't need to be a TikTok API dimension for this — it's already force-set on every row via `currentAdvertiserId` in `castFields`. Also refactored `getFieldsSchema` to reuse `getUniqueKeysForNode` directly so the UI pinning and the actual merge key can't drift apart the way the old duplicated derivation could have. Updated tests and docs to match.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/FieldsSelectionStep.tsx
prokopetsv: This one's already resolved — same feedback as your earlier comment on this line, fixed a few rounds back. `showDataLevelFieldsTip` is gated only on `Boolean(selectedFieldData?.uniqueKeysByDataLevel?.[selectedDataLevel])` now, no node-name check. There's a test (`FieldsSelectionStep.test.tsx` → "Data Level tip is node-name agnostic") proving it works for a non-TikTok node name too. Might be looking at a stale diff view — let me know if you're seeing something different in the current code.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
prokopetsv: Good question — traced it and it's already covered, I just didn't call it out in the comment. `isLoading={loadingSpecification || loadingFields}` is already passed to `StepNavigation`, and its single-step layout (which `configurationOnly` uses) disables Save while `isLoading` is true. Since `loadFieldsSafely` always fires now, that window is closed structurally, not by luck — no explicit `!loadingFields` gate needed beyond what's already there. Updated the comment to explain that instead of hedging like the race was still open.

packages/connectors/src/Sources/TikTokAds/Source.js
prokopetsv: Went with removing it. Checked it's redundant for every node, not just the two insights ones — each node's `uniqueKeys` field(s) are always defined in that node's own `fields` schema (verified with grep across all 7 field-definition files), so the very next loop already copies them whenever present. The only thing this block uniquely contributed was the `advertiser_id` fallback for `ad_insights`, which duplicates the `currentAdvertiserId` fallback a few lines above (that one already covers `ad_insights_by_country` too, which this block's special-case never did). Ran both an `AUCTION_ADVERTISER`-level `ad_insights` record and an `ads` record through `castFields` by hand against the rebuilt package to confirm output is unchanged.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/FieldsSelectionStep.tsx
```diff
@@ -49,9 +52,16 @@ export function FieldsSelectionStep({
     () => selectedFieldData?.fields ?? [],
     [selectedFieldData?.fields]
   );
-  const uniqueKeys = useMemo(
-    () => selectedFieldData?.uniqueKeys ?? [],
-    [selectedFieldData?.uniqueKeys]
+  const dataLevel = configuration?.[DATA_LEVEL_CONFIG_KEY];
+  const selectedDataLevel = typeof dataLevel === 'string' ? dataLevel : null;
+  const uniqueKeys = useMemo(() => {
```
romandubovyi, line 57: 💡 **Suggestion:** The `uniqueKeys` useMemo depends on the whole `selectedFieldData` object. Consider narrowing the dependency array to `selectedFieldData?.uniqueKeys`, `selectedFieldData?.uniqueKeysByDataLevel`, and `selectedDataLevel` to avoid recomputing/rerendering when unrelated field metadata changes.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/form/ConnectorDefinitionField.tsx
```diff
@@ -101,12 +101,21 @@ export function ConnectorDefinitionField({
         const updatedConfigurations = [...currentDefinition.connector.source.configuration];
         updatedConfigurations[configIndex] = connector.source.configuration[0] || {};
 
+        // connector.source.fields may add fields required by a changed config parameter
+        // (e.g. TikTok Ads Data Level). Union rather than overwrite so editing
+        // configuration can never drop previously selected fields, regardless of what
+        // triggered the update.
+        const updatedFields = Array.from(
```
romandubovyi, line 108: ⚪ **Nit:** Consider adding `?? []` fallbacks on both `currentDefinition.connector.source.fields` and `connector.source.fields` so the union stays safe even if a caller passes an incomplete config object.

packages/connectors/test/Sources/TikTokAds/Source.test.js
```diff
@@ -0,0 +1,195 @@
...
+    });
+  });
+
+  it('does not mutate the source fieldsSchema it read from', () => {
+    schema();
+    expect(fakeSource.fieldsSchema.ad_insights.uniqueKeysByDataLevel).toBeUndefined();
+  });
+});
+
+describe('fetchData unique-key validation', () => {
```
romandubovyi, line 137: 💡 **Suggestion:** The `fetchData` test fake only defines `ad_insights` in `fieldsSchema`. Since `getUniqueKeysForNode` falls back to `this.fieldsSchema[nodeName].uniqueKeys` for non-insights nodes, consider adding a minimal non-insights entry (e.g. `ads: { uniqueKeys: ['ad_id'] }`) so the fake source doesn't silently break if the test ever touches another node.

romandubovyi (commented):
Re-review complete. All previous blocking/important comments have been addressed; tests pass. A few minor suggestions remain.

romandubovyi (commented):
Review complete. One blocking issue found.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
```diff
@@ -433,7 +450,7 @@ export function ConnectorEditForm({
                   name: selectedConnector.name,
                   configuration: [connectorConfiguration],
                   node: existingConnector?.source.node ?? selectedNode,
-                  fields: existingConnector?.source.fields ?? selectedFields,
+                  fields: fieldsForSave,
```
romandubovyi, line 457: 🔴 **Blocking:** `fieldsForSave` is lost when this form is used by `AddConfigurationButton`: that caller only reads `connector.source.configuration[0]` and ignores `connector.source.fields`. Adding a new TikTok config with a different Data Level can therefore save the new `DataLevel` without the newly required unique fields, so the next run still fails validation. Please propagate the updated fields when adding a configuration, or merge them in `addConfiguration` like configuration edits do.

packages/connectors/src/Sources/TikTokAds/Source.js
```diff
@@ -360,9 +361,58 @@ var TikTokAdsSource = class TikTokAdsSource extends AbstractSource {
...
+   */
+  getUniqueKeysForNode(nodeName, dataLevel) {
+    if (nodeName === 'ad_insights') {
+      return this.populateDimensions(this.getDimensionsForDataLevel(dataLevel), 'advertiser_id');
+    }
+    if (nodeName === 'ad_insights_by_country') {
+      const dimensions = this.populateDimensions(this.getDimensionsForDataLevel(dataLevel), 'country_code');
+      return this.populateDimensions(dimensions, 'advertiser_id');
+    }
+    return this.fieldsSchema[nodeName].uniqueKeys;
```
romandubovyi: 💡 **Suggestion:** To prevent potential `TypeError` crashes if this method is called for a missing or improperly configured node, it's safer to use optional chaining here.

```javascript
return (this.fieldsSchema[nodeName] && this.fieldsSchema[nodeName].uniqueKeys) || [];
```

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/FieldsSelectionStep.tsx
prokopetsv: Applied — narrowed to `[selectedFieldData?.uniqueKeys, selectedFieldData?.uniqueKeysByDataLevel, selectedDataLevel]`. Good catch: since `selectedFieldData` comes from a `.find()` that runs every render, the old whole-object dep meant the memo never actually held, so this makes the memoization real rather than just tidier.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/form/ConnectorDefinitionField.tsx
prokopetsv: Tried this, but it trips the repo's `@typescript-eslint/no-unnecessary-condition` rule: `ConnectorSourceConfig.fields` is typed `string[]` (required, non-nullable), so eslint flags `?? []` as a provably-unnecessary condition and fails the build. The type contract already precludes an "incomplete config object" reaching here — a caller can't pass one without a cast that violates the types — and the project's lint config enforces trusting that over defensive fallbacks. So I've left it as the plain union. If we did want runtime tolerance for malformed configs, the right place would be loosening the type to `fields?: string[]` at the source, but that's a broader change than this line and outside this PR's scope.

packages/connectors/test/Sources/TikTokAds/Source.test.js
prokopetsv: Done — added a minimal `ads` node to the fake `fieldsSchema` so `getUniqueKeysForNode`'s static-fallback branch stays exercisable if a future test in this block targets a non-insights node.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
prokopetsv: Confirmed and fixed — good catch, this was a real gap. `ConnectorEditForm` was computing `fieldsForSave` correctly, but `AddConfigurationButton.onSubmit` only forwarded `configuration[0]` and dropped `source.fields`, so `addConfiguration` never saw the new Data Level's required keys. Now the whole connector is passed through and `addConfiguration` unions fields the same way `updateConnectorConfiguration` does, so a new config with a different Data Level contributes its required unique keys to the shared field set without dropping any existing selection.

packages/connectors/src/Sources/TikTokAds/Source.js
prokopetsv: Done — `return this.fieldsSchema[nodeName]?.uniqueKeys ?? [];`. Worth noting it's not reachable today (both callers, `fetchData` and `Connector.js` `getStorageByNode`, throw on an unknown node before this line runs), but agreed it's cheap hardening that makes the method safe to call standalone. Used `?? []` rather than `|| []` since `uniqueKeys` is only ever an array or `undefined`.

romandubovyi (commented):
Review complete. Found one blocking issue around DataLevel default reconciliation.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
```diff
@@ -181,6 +178,24 @@ export function ConnectorEditForm({
...
+  // Union the persisted fields with whatever the chosen DataLevel requires (e.g. TikTok
+  // ad_insights needs ad_id at AUCTION_AD). No-op for nodes without uniqueKeysByDataLevel.
+  // Falls back to the unchanged fields if connectorFields hasn't loaded yet, but that
+  // window isn't user-reachable: loadFieldsSafely is always triggered for an existing
+  // connector, and the Save button is disabled via isLoading={... || loadingFields} at
+  // the <StepNavigation> call site below until that fetch settles.
+  const fieldsForSave = useMemo(() => {
+    const fields = existingConnector?.source.fields ?? selectedFields;
+    const dataLevel = connectorConfiguration[DATA_LEVEL_CONFIG_KEY];
+    if (typeof dataLevel !== 'string') return fields;
```
romandubovyi: 🔴 **Blocking:** This skips reconciliation when `DataLevel` is only present as the spec default. `ConfigurationStep` can display/default `AUCTION_AD` without pushing it into `connectorConfiguration`, so saving an added/edited configuration may keep the old field list without `ad_id` while runtime defaults to `AUCTION_AD` and requires it. Please resolve the effective data level from the spec default too, or ensure defaults are propagated before `fieldsForSave` runs.

prokopetsv: Confirmed and fixed. Root cause: `ConfigurationStep` guards `onConfigurationChange` behind `updatingFromParentRef`, which its default-seeding sets true and only clears in a `setTimeout(0)` — so an untouched spec default never reaches `connectorConfiguration`, and `fieldsForSave` saw `undefined`. Went with your first option: added `resolveEffectiveDataLevel(configuration, specification)` that falls back to the spec `default`, so field reconciliation targets the same level the run defaults to. Extracted it as a pure helper with unit tests (config value wins; spec default fallback; non-string ignored; undefined when neither applies). Preferred this over propagating defaults out of ConfigurationStep, since that seeding intentionally keeps user-set vs default distinct across all params.

romandubovyi (commented):
Review complete. Found one blocking issue: package-lock is stale after adding the connectors test dependency.

packages/connectors/package.json
```diff
@@ -57,6 +58,7 @@
     "fs-extra": "^11",
     "glob": "^13",
     "globals": "^16",
-    "vite": "^6.3.5"
+    "vite": "^6.3.5",
+    "vitest": "^4.1.8"
```
romandubovyi, line 62: 🔴 **Blocking:** This adds a new workspace devDependency, but `package-lock.json` was not updated. The lockfile entry for `packages/connectors` still does not include `vitest`, so a clean `npm ci` can fail because the workspace manifest and lockfile are out of sync. Please run `npm install --package-lock-only` or the project’s normal install flow and commit the lockfile update.

prokopetsv: Fixed — ran `npm install --package-lock-only` and the lockfile now records `vitest ^4.1.8` for `packages/connectors`, so `npm ci` is back in sync. The only other change in the lock is npm dropping `peer: true` from 12 `@esbuild/*` binaries, which is expected now that vitest depends on esbuild directly. (Heads-up: I'd reconciled this earlier but the unstaged lockfile change got lost between revisions — committing it explicitly this time.)

romandubovyi (approved):

PR #1412, opened 2026-07-14
6712: add project-aware MCP instructions

amarchenk0:


romandubovyi (commented):
Review complete. One important finding around a localStorage key collision when the CollapsibleCard was renamed — see inline comment.

apps/web/src/pages/project-settings/OverviewTab.tsx
```diff
@@ -162,10 +174,10 @@ export function OverviewTab() {
 
   return (
     <div className='flex flex-col gap-4'>
-      <CollapsibleCard collapsible name='project-description'>
+      <CollapsibleCard collapsible name='project-overview'>
```
romandubovyi, line 177: 🟡 **Important:** The first card's `name` changed from `'project-description'` to `'project-overview'`, and the new business-context card below now uses `name='project-description'` (line ~373). Since `CollapsibleCard` persists collapsed state in `localStorage` keyed by `name`, existing users who had collapsed the old "Description" (now "Overview") card will find that stale `true` value applied to the brand-new Description card instead — it'll render collapsed on first load even though they've never interacted with it, while their actual Overview collapse preference is silently lost. Consider giving the new card a fresh, non-colliding `name` (e.g. `'project-business-context'`).

romandubovyi (commented):
Review of project-aware MCP instructions. Main themes: keep MCP initialize resilient when optional project settings fail to load, and tighten a couple of edge cases around description limits and first-write concurrency.

apps/backend/src/ee/mcp/instructions/mcp-instructions.service.ts
```diff
@@ -0,0 +1,19 @@
...
+import { composeMcpInstructions } from './mcp-system-instructions';
+
+@Injectable()
+export class McpInstructionsService {
+  constructor(
+    @Inject(PROJECT_SETTINGS_FACADE)
+    private readonly projectSettings: ProjectSettingsFacade
+  ) {}
+
+  async getInstructions(projectId: string): Promise<string> {
```
romandubovyi, line 17: 🟡 **Important:** Project description is optional context, but a failure in `getDescription()` currently fails the entire MCP `initialize` handshake. If the DB/settings lookup errors, clients lose tool access even though system instructions alone would still be useful. Consider catching lookup errors, logging a warning, and falling back to `composeMcpInstructions(null)` so initialize remains resilient.

apps/web/src/pages/project-settings/OverviewTab.tsx
```diff
@@ -258,6 +270,42 @@ export function OverviewTab() {
...
+          <div className='flex flex-col gap-2'>
+            <p className='text-muted-foreground text-xs'>
+              Describe the project&apos;s business context, goals, terminology, and conventions.
+              This description is shared with connected AI assistants through MCP. Do not include
+              secrets.
+            </p>
+            {isProjectSettingsLoading ? (
+              <Skeleton className='h-64 w-full' />
+            ) : (
+              <InlineEditDescription
```
romandubovyi, line 292: 💡 **Suggestion:** Backend enforces `PROJECT_DESCRIPTION_MAX_LENGTH` (10_000), but the UI has no matching client-side limit. Admins can paste a long description and only learn about the constraint after a failed save. Consider enforcing the same max length in the editor (or at least surfacing remaining characters) so validation matches the API.

apps/backend/src/project-settings/services/project-settings.service.ts
```diff
@@ -0,0 +1,23 @@
...
+  constructor(
+    @InjectRepository(ProjectSettings)
+    private readonly repository: Repository<ProjectSettings>
+  ) {}
+
+  async findByProjectId(projectId: string): Promise<ProjectSettings | null> {
+    return this.repository.findOne({ where: { projectId } });
+  }
+
+  async saveDescription(projectId: string, description: string | null): Promise<ProjectSettings> {
```
romandubovyi, line 17: 💡 **Suggestion:** `find` + `create` + `save` can race on the first write for a project: two concurrent first saves may both observe `null` and the second insert then fails on the `projectId` primary key. Consider an upsert (`ON CONFLICT (projectId) DO UPDATE`) so the first description write is idempotent under concurrency.

romandubovyi (approved):
Approved. No blocking issues.

Open non-blocking notes from review (important/suggestions): resilient MCP initialize if project settings lookup fails, client-side max length for description, and upsert for first-write concurrency. Previously noted localStorage key collision for CollapsibleCard `name` is also non-blocking but worth addressing if easy.

PR #1414, opened 2026-07-14
feat(backend): introduce buildBlendedFieldUnifiedName for stable field aliasing

romandubovyi:
## Summary

- Fixes blended field unified-name collisions for nested (struct) fields in joinable data marts.
- Flat fields keep their existing names byte-for-byte (`{sqlPrefix}__{field}`); nested fields always get a stable hash suffix so paths like `campaign_id` and `campaign.id` no longer collapse to the same alias.
- Centralizes naming in `buildBlendedFieldUnifiedName` so the identity depends only on `(aliasPath, originalFieldName)`, not on sibling fields or reactive collision detection.

## Problem

`BlendableSchemaService` used to build report column names as:

```
${sqlPrefix}__${field.name.replace(/\./g, '_')}
```

That made a flat `campaign_id` and a nested `campaign.id` both become `…__campaign_id`. The column picker, filters, and field index then treated two different fields as one.

## Solution

| Field type | Unified name |
|---|---|
| Flat | `{sqlPrefix}__{name}` (unchanged) |
| Nested | `{sqlPrefix}__{dots→_}__{sha1(aliasPath\|originalFieldName)[0:8]}` |

Example: `ads__campaign_id` vs `ads__campaign_id__a8702665`.

## Scope

- **In scope:** nested blended fields from joined data marts (names used in blendable schema / reports).
- **Out of scope in this PR:** data migration for already-saved report `columnConfig` / `filterConfig` / `sortConfig` that still store legacy nested names (can be a follow-up if needed).

## Test plan

- [x] Unit tests for `buildBlendedFieldUnifiedName` pass (flat unchanged, nested hashed, collision case distinct, multi-level paths distinct, stability independent of other fields).
- [x] Open a data mart with a join where both `campaign_id` and `campaign.id` exist — both appear as separate selectable columns.
- [x] Selecting one does not toggle the other in the report column picker.
- [x] Existing reports that only use **flat** blended fields still resolve and run without config changes.
- [x] Nested field names in a newly saved report include the hash suffix and round-trip correctly on reload/run.

### DIFF
--- a/apps/backend/src/data-marts/services/blendable-schema.service.ts
+++ b/apps/backend/src/data-marts/services/blendable-schema.service.ts
@@ -223,11 +223,10 @@ export class BlendableSchemaService {
       );
       const flatTargetFields = flattenSchemaFields(targetSchemaFields);
 
-      // `sqlPrefix` is SQL‑safe because each `targetAlias` segment in
-      // `currentPath` is validated against `^[a-z0-9_]+$` in the Join
-      // Settings form. `displayPrefix` is free‑form and must never flow
-      // into SQL identifiers.
-      const sqlPrefix = currentPath.replace(/\./g, '_');
+      // Each `targetAlias` segment in `currentPath` is validated against
+      // `^[a-z0-9_]+$` in the Join Settings form, so the SQL-safe prefix
+      // derived inside `buildBlendedFieldUnifiedName` is safe. `displayPrefix`
+      // is free-form and must never flow into SQL identifiers.
       const displayPrefix = sourceConfig?.alias ?? rel.targetDataMart.title;
 
       const availableSource = new AvailableSourceDto();
@@ -246,7 +245,7 @@ export class BlendableSchemaService {
         const fieldOverride = sourceConfig?.fields?.[field.name];
 
         const dto = new BlendedFieldDto();
-        dto.name = buildBlendedFieldUnifiedName(currentPath, sqlPrefix, field.name);
+        dto.name = buildBlendedFieldUnifiedName(currentPath, field.name);
         dto.aliasPath = currentPath;
         dto.outputPrefix = displayPrefix;
         dto.sourceRelationshipId = rel.id;
--- a/apps/backend/src/data-marts/services/blendable-schema.service.ts
+++ b/apps/backend/src/data-marts/services/blendable-schema.service.ts
@@ -20,6 +20,7 @@ import { BusinessViolationException } from '../../common/exceptions/business-vio
 import { DataMartRelationship } from '../entities/data-mart-relationship.entity';
 import { DataMartStatus } from '../enums/data-mart-status.enum';
 import { IdpProjectionsFacade } from '../../idp/facades/idp-projections.facade';
+import { buildBlendedFieldUnifiedName } from './blended-field-name';
 
 export interface BlendableSchemaAccessor {
   userId: string;
@@ -245,9 +246,7 @@ export class BlendableSchemaService {
         const fieldOverride = sourceConfig?.fields?.[field.name];
 
         const dto = new BlendedFieldDto();
-        // Replace dots in nested struct field paths (e.g. `struct.field`) with
-        // underscores so the resulting alias is a valid SQL identifier.
-        dto.name = `${sqlPrefix}__${field.name.replace(/\./g, '_')}`;
+        dto.name = buildBlendedFieldUnifiedName(currentPath, sqlPrefix, field.name);
         dto.aliasPath = currentPath;
         dto.outputPrefix = displayPrefix;
         dto.sourceRelationshipId = rel.id;
--- a/apps/backend/src/data-marts/services/blended-field-name.ts
+++ b/apps/backend/src/data-marts/services/blended-field-name.ts
@@ -1,0 +1,20 @@
+import { createHash } from 'node:crypto';
+
+/** First 8 hex chars of sha1(aliasPath + '|' + originalFieldName) — a stable
+ * per-field identity, independent of any other field that does or doesn't exist. */
+function shortHash(aliasPath: string, originalFieldName: string): string {
+  return createHash('sha1').update(`${aliasPath}|${originalFieldName}`).digest('hex').slice(0, 8);
+}
+
+/** Unified name used in blendable-schema payloads and report saves. */
+export function buildBlendedFieldUnifiedName(
+  aliasPath: string,
+  sqlPrefix: string,
+  originalFieldName: string
+): string {
+  if (!originalFieldName.includes('.')) {
+    return `${sqlPrefix}__${originalFieldName}`; // flat — byte-identical to today, always
+  }
+  const readable = originalFieldName.replace(/\./g, '_');
+  return `${sqlPrefix}__${readable}__${shortHash(aliasPath, originalFieldName)}`; // nested — always hashed
+}
--- a/apps/backend/src/data-marts/services/blended-field-name.ts
+++ b/apps/backend/src/data-marts/services/blended-field-name.ts
@@ -6,14 +6,16 @@ function shortHash(aliasPath: string, originalFieldName: string): string {
   return createHash('sha1').update(`${aliasPath}|${originalFieldName}`).digest('hex').slice(0, 8);
 }
 
-/** Unified name used in blendable-schema payloads and report saves. */
-export function buildBlendedFieldUnifiedName(
-  aliasPath: string,
-  sqlPrefix: string,
-  originalFieldName: string
-): string {
+/**
+ * Unified name used in blendable-schema payloads and report saves.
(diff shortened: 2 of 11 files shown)


apps/backend/src/data-marts/services/blendable-schema.service.ts
```diff
@@ -245,9 +246,7 @@ export class BlendableSchemaService {
         const fieldOverride = sourceConfig?.fields?.[field.name];
 
         const dto = new BlendedFieldDto();
-        // Replace dots in nested struct field paths (e.g. `struct.field`) with
-        // underscores so the resulting alias is a valid SQL identifier.
-        dto.name = `${sqlPrefix}__${field.name.replace(/\./g, '_')}`;
+        dto.name = buildBlendedFieldUnifiedName(currentPath, sqlPrefix, field.name);
```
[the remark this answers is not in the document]
romandubovyi: Thanks for flagging this — we considered dual-key / migration support for legacy nested unified names.

**Decision: intentional no-compat for this case.**

- Nested blended fields in saved report `columnConfig` / filters / sorts essentially do not exist in production today (this area was not used that way yet).
- Flat blended names are unchanged byte-for-byte, so the common production path stays backward-compatible.
- Adding runtime dual-key resolution or a data migration would be pure speculative complexity for configs we don't have.

If such reports appear before this ships, we'd migrate or add a legacy alias map — but for the current rollout, breaking only pre-existing nested blended references is an accepted non-issue.

Not planning to change the implementation for this comment.

apps/backend/src/data-marts/services/blended-field-name.ts
```diff
@@ -0,0 +1,20 @@
...
+export function buildBlendedFieldUnifiedName(
+  aliasPath: string,
+  sqlPrefix: string,
+  originalFieldName: string
+): string {
+  if (!originalFieldName.includes('.')) {
+    return `${sqlPrefix}__${originalFieldName}`; // flat — byte-identical to today, always
+  }
+  const readable = originalFieldName.replace(/\./g, '_');
+  return `${sqlPrefix}__${readable}__${shortHash(aliasPath, originalFieldName)}`; // nested — always hashed
```
amarchenk0, line 21: Hashing only nested paths does not guarantee a unique output alias because flat names stay unchanged. Under `customers`, a valid flat field named `campaign_id__b996a659` and the nested field `campaign.id` both become `customers__campaign_id__b996a659`. The duplicate then makes `buildBlendedFieldIndex` throw, so the report cannot be built. Please disambiguate generated nested aliases against the complete set of flat aliases, or make the shared naming scheme unambiguous.

amarchenk0 (changes requested):
Please address the inline comment before approval.

apps/backend/src/data-marts/services/blended-field-name.ts
romandubovyi: Thanks — the edge case is real.

A flat field literally named `{readable}__{sameHash}` can collide with a hashed nested name. We accept this as a residual risk for now:

- flat names stay byte-identical (backward compat);
- nested names are intentionally non-reactive (`aliasPath` + `originalFieldName` only);
- duplicates already fail loud in `buildBlendedFieldIndex`.

No code change planned for this PR. We can revisit a stronger marker later if it ever shows up in practice.

zapolsky:
✅ **No new blocking issues found in this automated pass.** Flat blended names are byte-for-byte unchanged (verified), and new nested-field reports are fully self-consistent (produce → store → resolve all key off the same `dto.name`). A human `CHANGES_REQUESTED` is already open and the author has documented two accepted tradeoffs — a maintainer owns the merge decision.

Deep pass: worktree checkout at `2439495`, 5 subagents (reviewer · logic-hunter · architect · tester · fresh-eyes) + one adversarial skeptic. `tsc`/`eslint` on the touched files are clean; the 60 unit tests pass and the hardcoded sha1 constants were recomputed independently and match.

The two already-flagged concerns are respected as **deliberately accepted tradeoffs** and not re-litigated: the Codex `P1` (legacy nested names in saved configs break) and @amarchenk0's flat-vs-nested hash collision (fails loud in `buildBlendedFieldIndex`). Both mechanisms were independently reproduced and confirmed real. The notes below are **new angles** on that same accepted scope — worth an explicit conscious confirmation, not a code change:

- **The `#1329` pre-join-filter migration is a second, un-updated producer of the nested name.** `apps/backend/src/migrations/1779200000000-migrate-prejoin-filter-to-unified-column.ts:28` folds a pre-join slice into `${aliasPath→_}__${column→_}` — the **pre-hash** format. For a nested (dotted) column it now mints `customers__campaign_id`, which the post-PR hashed index (`blended-field-index.ts:44`) can never resolve. The skeptic **refuted this as a blocker**: `migrateRules` only reshapes *pre-existing* nested pre-join rules (`…migrate-prejoin…ts:24-33`, `up` reads persisted `filterConfig`), so its trigger population is a strict subset of "nested blended refs in saved configs" — the set you've declared empty. So it's covered by your tradeoff, **but** it's a live forward-writing path that will silently keep emitting the legacy format on first run. Worth a one-line comment at line 28 noting it deliberately emits the pre-hash nested name (so a future dev doesn't assume it tracks `buildBlendedFieldUnifiedName`), and a mention in the follow-up-migration note.
- **Question — one sub-case of the accepted break is *silent wrong data*, not a loud failure.** Sequence: (1) a report is saved (pre-PR) selecting a nested `campaign.id` → stored as `customers__campaign_id`; (2) a flat sibling `campaign_id` is later added to that joined mart; (3) this PR deploys. The stored `customers__campaign_id` now resolves to the **flat** field (its unchanged name), so `assertNoOrphanedColumnReferences` (`blended-report-data.service.ts:390`) passes and the report returns the flat column's data in place of the nested field the user picked. Pre-PR this exact schema was a loud `Ambiguous blended column name` throw (`blended-field-index.ts:28`); the PR turns it into a silent mis-resolution for this narrow window. Same empty population, but a materially worse failure mode than the loud break you accepted — worth confirming the tradeoff is meant to include it.

<details>

- **Rolling-deploy skew (nested fields only, transient).** During a rolling deploy, old pods emit `ads__campaign_id` and new pods emit `ads__campaign_id__a8702665` for the same `campaign.id`; a nested-field report saved/run mid-rollout can transiently orphan until the fleet fully rolls. Self-heals for new-pod-minted names; worth a line in the release note alongside the accepted legacy break.
- `apps/backend/src/migrations/1779200000000-...ts` / `blended-report-data.service.ts` — no changeset ships with this branch. The column-picker collision fix is user-facing, so a `'owox'` changeset is likely expected (~convention).
- `blended-field-name.ts:10-13` — the `sqlPrefix` parameter is redundant with `aliasPath` (`sqlPrefix === aliasPath.replace(/\./g,'_')` at the sole caller, `blendable-schema.service.ts:230`). The flat branch uses only `sqlPrefix`, the hash uses only `aliasPath`; a future caller passing a `sqlPrefix` not derived from `aliasPath` would silently desync the readable prefix from the hash. Deriving it internally would make the identity depend only on `(aliasPath, originalFieldName)` as the doc claims.
- minor (doc drift + low-risk edges): (1) two comments still describe the old flat-only format — `blended-report-data.service.ts:374` ("Blended column refs are flat `<aliasPath>__<field>` strings") and `apps/web/src/features/data-marts/shared/types/output-config.ts:107` (example `category_details__item_event_count` is itself a nested field, now hash-suffixed); (2) nested aliases grow +10 chars — only a concern on identifier-truncating warehouses (none of the current storages truncate that low, so effectively inert); (3) `.includes('.')` would hash a *flat* field whose literal name contains a dot (possible only on non-BigQuery storages) — internally consistent, only matters for a pre-existing saved config; (4) `buildBlendedFieldUnifiedName` throws `TypeError` on a null `originalFieldName` — pre-existing behavior (old inline code had the same), the new helper is just the natural place to harden it.

</details>

**Not attributable to this PR:** a full `tsc -b` in the review worktree reports errors only in untouched files (`@owox/internal-helpers` re-exports in `producer.module.ts` / `ee/mcp/observability/*`, and several `test/*.e2e-spec.ts` type mismatches) — a pre-existing/infra baseline from the linked `node_modules`, none in the PR-touched files.

<sub>Automated OWOX review · worktree multi-agent pass · the one MEDIUM/HIGH candidate was adversarially refuted to a non-blocker.</sub>

romandubovyi:
@zapolsky Thanks for the pass.

Confirmed both notes are covered by the already accepted tradeoff (no nested blended refs in saved prod configs). No migration logic change.

Follow-up commit only:
- comment on the pre-join migration that it keeps the pre-hash form
- changeset release note (legacy nested names + rolling-deploy skew)
- docs + derive `sqlPrefix` from `aliasPath` inside `buildBlendedFieldUnifiedName`

PR #1415, opened 2026-07-14
refactor(web): move refresh and AI buttons to TableToolbar

artemfenoga:
# Refactor Output Schema toolbar actions

Frame 23137619

Frame 23137620

Frame 23137621

This PR improves the discoverability and organization of Output Schema actions. The **Refresh schema** and global **Generate field aliases & descriptions** actions are now located in the table toolbar, keeping all primary actions in one place. The global AI action has also been simplified by removing the dropdown menu, allowing users to generate both field aliases and descriptions with a single click.

To make AI features easier to discover, AI action buttons have been added to the **Alias** and **Description** column headers, allowing users to generate values for an entire column directly where the action applies.

To support these changes, all business logic remains in `DataMartSchemaSettings`, while `TableToolbar` stays a presentational component. A new `SchemaToolbar` interface is used to pass toolbar state and callbacks through the component hierarchy.

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
@@ -1,12 +1,4 @@
 import { Button } from '@owox/ui/components/button';
-import {
-  DropdownMenu,
-  DropdownMenuContent,
-  DropdownMenuItem,
-  DropdownMenuTrigger,
-} from '@owox/ui/components/dropdown-menu';
-import { Tooltip, TooltipContent, TooltipTrigger } from '@owox/ui/components/tooltip';
-import { Loader2, Sparkles } from 'lucide-react';
 import { useCallback, useEffect, useRef } from 'react';
 import toast from 'react-hot-toast';
 import { useOutletContext } from 'react-router-dom';
@@ -23,6 +15,7 @@ import { SchemaContent } from './SchemaContent';
 import { DataMartDefinitionType, DataMartMetadataScope } from '../../../shared/index.ts';
 import { useAiHelper, useAiHelperAvailability } from '../../model/hooks';
 import type { ResolvedSchema } from '../../model/hooks';
+import type { SchemaToolbar } from './types/schema-toolbar';
 
 interface DataMartSchemaSettingsProps {
   definitionType: DataMartDefinitionType | null;
@@ -182,7 +175,7 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
   // Backend rejects metadata generation for CONNECTOR data marts; hide the buttons
   // so the user is never offered an action that's guaranteed to 422.
   const isConnector = definitionType === DataMartDefinitionType.CONNECTOR;
-  const showAiHelper = isAiHelperEnabled && !isConnector;
+  const showAiHelper = Boolean(isAiHelperEnabled) && !isConnector;
 
   // Reset schema when operation is successful
   useEffect(() => {
@@ -377,13 +370,50 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
   );
 
   // Disable buttons during schema operations (save or actualization)
-  const isSchemaOperationInProgress = isLoading || isSchemaActualizationLoading;
+  const isSchemaOperationInProgress = isLoading || Boolean(isSchemaActualizationLoading);
   const isAiBusy = aiPendingScope !== null;
-  const isBulkAiBusy =
-    aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_DESCRIPTIONS ||
-    aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_ALIASES ||
-    aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_METADATA;
   const hasFields = !!schema && schema.fields.length > 0;
+
+  const schemaToolbar: SchemaToolbar = {
+    showAiHelper,
+
+    refresh: {
+      disabled: !definitionType || isSchemaOperationInProgress,
+      onClick: handleActualize,
+    },
+
+    ai: {
+      disabled: !hasFields || isSchemaOperationInProgress || isAiBusy,
+      loading: {
+        metadata: aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_METADATA,
+
+        aliases: aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_ALIASES,
+
+        descriptions: aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_DESCRIPTIONS,
+      },
+
+      onGenerateMetadata: () => {
+        runGuarded?.(resolved => runBulkAi(DataMartMetadataScope.ALL_FIELD_METADATA, resolved), {
+          intent: 'ai',
+        });
+      },
+
+      onGenerateDescriptions: () => {
+        runGuarded?.(
+          resolved => runBulkAi(DataMartMetadataScope.ALL_FIELD_DESCRIPTIONS, resolved),
+          {
+            intent: 'ai',
+          }
+        );
+      },
+
+      onGenerateAliases: () => {
+        runGuarded?.(resolved => runBulkAi(DataMartMetadataScope.ALL_FIELD_ALIASES, resolved), {
+          intent: 'ai',
+        });
+      },
+    },
+  };
 
   if (!dataMart) {
     return <div>Error: Data mart not found</div>;
@@ -404,6 +434,7 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
               }
             : undefined
         }
+        schemaToolbar={schemaToolbar}
       />
       <div className='align-items-center mt-4 flex justify-between'>
         <div className='flex items-center gap-2'>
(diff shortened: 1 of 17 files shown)


apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
```diff
@@ -244,14 +237,51 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
...
+      loading: {
+        metadata: aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_METADATA,
+
+        aliases: aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_ALIASES,
+
+        descriptions: aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_DESCRIPTIONS,
+      },
+
+      onGenerateMetadata: () => {
+        runGuarded?.(resolved => runBulkAi(DataMartMetadataScope.ALL_FIELD_METADATA, resolved), {
```
romandubovyi, line 396: 🔴 **Blocking:** This toolbar wires into `runBulkAi` → `handleGenerateAllField*`, but this branch predates #1418 (`stop schema output edits from being silently reverted`), which rewrote those exact handlers on `main`. Here they're still the older `targetSchema.fields.map(...)` form — no `mergeGeneratedMetadata`, no `activeDataMartIdRef` guard, no duplicate-name handling — i.e. the very code #1418 fixed. Those handlers don't show in this PR's diff (the merge-base also lacks #1418) and git auto-merges cleanly, so #1418 can be silently lost on merge/rebase. Please rebase onto latest `main` and re-apply the toolbar refactor on top of #1418's logic, then re-verify bulk generation preserves in-flight edits.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/BaseSchemaTable.tsx
```diff
@@ -276,10 +282,27 @@ export function BaseSchemaTable<T extends BaseSchemaField>({
...
+
+              <TooltipContent>Alternative name for the field</TooltipContent>
+            </Tooltip>
+            <div
+              className={cn(
+                'opacity-0 transition-opacity group-hover:opacity-100',
+                schemaToolbar.ai.loading.aliases && 'opacity-100'
+              )}
+            >
+              <SchemaHeaderAiButton
```
romandubovyi: 🟡 **Important:** The Alias/Description column-header AI buttons render unconditionally — `showAiHelper` is never referenced in this file. The global toolbar button is gated with `{schemaToolbar.showAiHelper && ...}`, but these header buttons are not, so on CONNECTOR data marts (or deployments with the AI helper disabled) they still appear and clicking them triggers a generation call the backend rejects. That defeats the stated invariant in `DataMartSchemaSettings` ("hide the buttons so the user is never offered an action that's guaranteed to 422"). Gate both header buttons on `schemaToolbar.showAiHelper`.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/BaseSchemaTable.tsx
```diff
@@ -302,10 +325,27 @@ export function BaseSchemaTable<T extends BaseSchemaField>({
...
+          <div className='group flex items-center gap-1'>
+            <Tooltip>
+              <TooltipTrigger className='cursor-default'>Description</TooltipTrigger>
+
+              <TooltipContent>Detailed information about the field</TooltipContent>
+            </Tooltip>
+            <div
+              className={cn(
+                'opacity-0 transition-opacity group-hover:opacity-100',
+                schemaToolbar.ai.loading.aliases && 'opacity-100'
```
romandubovyi: 🟡 **Important:** Copy-paste from the Alias header — the Description header wrapper keeps itself visible via `schemaToolbar.ai.loading.aliases`, but the button spins on `loading.descriptions` (line 344). As a result, while descriptions are generating the button fades out on mouse-leave, and while aliases are generating the Description button's wrapper stays visible for no reason. Should be `schemaToolbar.ai.loading.descriptions`.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/BaseSchemaTable.test.tsx
```diff
@@ -16,6 +17,27 @@ vi.mock('react-router-dom', async importOriginal => {
   };
 });
 
+const mockSchemaToolbar: SchemaToolbar = {
```
romandubovyi, line 20: 🟡 **Important:** The test only threads `mockSchemaToolbar` into the existing aggregation tests — nothing covers the actual new behavior: header AI buttons rendering, clicking them invoking `onGenerateAliases`/`onGenerateDescriptions`, the loading state, or `showAiHelper` gating. The mock also sets `showAiHelper: false`, so the headline feature is never exercised. Please add at least one test asserting a header button click calls the right callback, and one asserting the buttons are hidden when `showAiHelper` is false (see point on gating).

romandubovyi (commented):
Review complete. Structurally a clean refactor (nice SchemaToolbar separation), but one blocker: this branch predates #1418 and reverts its bulk-AI merge fix, plus a few issues on the new column-header AI buttons.

romandubovyi (commented):
Review complete. Two issues found in the new schema toolbar AI actions.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/BaseSchemaTable.tsx
romandubovyi: 🔴 **Blocking:** These header AI buttons are rendered regardless of `schemaToolbar.showAiHelper`, while `ai.disabled` does not include that guard. This exposes generation actions for connector Data Marts, where the backend rejects them with 422, and when AI is disabled. Render the header controls only when `showAiHelper` is true (or include it in `ai.disabled`).

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/BaseSchemaTable.tsx
romandubovyi: 💡 **Suggestion:** This visibility condition checks `loading.aliases`, so the Description button becomes transparent while description generation is in progress unless the header is hovered. Use `loading.descriptions` here.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
```diff
@@ -377,14 +370,51 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
...
-  const isSchemaOperationInProgress = isLoading || isSchemaActualizationLoading;
+  const isSchemaOperationInProgress = isLoading || Boolean(isSchemaActualizationLoading);
   const isAiBusy = aiPendingScope !== null;
-  const isBulkAiBusy =
-    aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_DESCRIPTIONS ||
-    aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_ALIASES ||
-    aiPendingScope?.scope === DataMartMetadataScope.ALL_FIELD_METADATA;
   const hasFields = !!schema && schema.fields.length > 0;
 
+  const schemaToolbar: SchemaToolbar = {
```
romandubovyi, line 377: 🟡 **Important:** `schemaToolbar` is built inline, so `ai.onGenerateAliases` / `onGenerateDescriptions` get a new identity on every render. `BaseSchemaTable` lists exactly those in the `columns` `useMemo` deps (lines 413-418), so the memo is invalidated on every render and TanStack Table receives a fresh column array each time. Previously these handlers lived directly in JSX, so identity did not matter. Suggest wrapping the three handlers in `useCallback` and the object in `useMemo` — consistent with `handleActualize`, `handleGenerateAll*` and `runBulkAi` in this same file.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/BaseSchemaTable.tsx
```diff
@@ -276,10 +282,29 @@ export function BaseSchemaTable<T extends BaseSchemaField>({
...
+          <div className='group flex items-center gap-1'>
+            <Tooltip>
+              <TooltipTrigger className='cursor-default'>Alias</TooltipTrigger>
+
+              <TooltipContent>Alternative name for the field</TooltipContent>
+            </Tooltip>
+            {schemaToolbar.showAiHelper && (
+              <div
+                className={cn(
+                  'opacity-0 transition-opacity group-hover:opacity-100',
```
romandubovyi, line 294: 🟡 **Important:** The button stays in the DOM and focusable while `opacity-0`, so a keyboard user tabbing through the header lands on a completely invisible control with no focus indicator. Adding `group-focus-within:opacity-100` alongside `group-hover:opacity-100` fixes it. Same applies to the Description column header below.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
```diff
@@ -424,88 +455,7 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
...
-          <Button
-            type='button'
-            variant='outline'
-            onClick={handleActualize}
-            disabled={!definitionType || isSchemaOperationInProgress}
-          >
-            Refresh schema
-          </Button>
-        </div>
+        <div className='flex items-center gap-2'></div>
```
romandubovyi, line 458: 💡 **Suggestion:** Leftover empty div from the refactor. The parent is `justify-between`, but with a single child the remaining group already sits on the left, so removing this does not change the layout.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/BaseSchemaTable.test.tsx
```diff
@@ -135,3 +205,72 @@ describe('BaseSchemaTable — Aggregations column', () => {
     expect(updatedFields[0].allowedAggregations).toContain('COUNT');
   });
 });
+
+it('renders AI header buttons when AI helper is enabled', () => {
```
romandubovyi, line 209: ⚪ **Nit:** These three tests sit at the top level, after the `Aggregations column` describe closes, so they do not get its `beforeEach(vi.clearAllMocks)` — and `mockSchemaToolbar` holds shared `vi.fn()` instances that carry over between them. Harmless today since none of them assert on those shared mocks, but it becomes an order-dependent trap the moment someone does. Wrapping them in their own `describe(...)` with a `beforeEach` would keep it safe.

romandubovyi (commented):
Re-review after the latest push.

**Previous round — all four points resolved:**
- 🔴 #1418 rebase — resolved. `b1fc8ff` is now an ancestor of this branch and `mergeGeneratedMetadata` / `activeDataMartIdRef` / `showBulkMergeFeedback` are present in `DataMartSchemaSettings.tsx`. No silent loss on merge.
- 🟡 Header AI buttons not gated — resolved. Both Alias and Description headers are now wrapped in `{schemaToolbar.showAiHelper && ...}`.
- 🟡 Description wrapper checked `loading.aliases` — resolved, now `loading.descriptions`.
- 🟡 Missing tests for the new behaviour — resolved. Three tests added covering render, gating, and callback wiring.

Four new points below, none blocking.

romandubovyi (approved):

PR #1416, opened 2026-07-15
feat(data-marts): dedup-aware joined-field aggregations and missing-PK warning (#6733)

zapolsky:
Fibery **#6733** — items 1 (PK highlight) + 2 (integer aggregations for joined data marts). Downstream of the aggregations engine (#1373).

## What & why

**Item 2 — dedup-aware joined-field aggregations.** A joined data mart field is pre-aggregated ("dedup", the join-level `aggregateFunction`) to one row per join key before the blend, so its value in the blended result is the dedup function's **output**, not the raw source value. This PR makes the blended field's type and its report-level available aggregations follow that dedup-output ("effective") type:

- `COUNT` / `COUNT_DISTINCT` dedup → integer → arithmetic functions (`SUM`/`AVG`/`MIN`/`MAX`) become available, `SUM` default-active.
- A fan-out funnel that sums per-key distinct counts now adds up correctly instead of collapsing joined events into a `STRING_AGG` string and under-counting them.
- The **raw** type is preserved as `sourceFieldType` and used for **pre-join slices** (they run on the raw column in the `*_raw` CTE, before dedup), output-control validation, and the authoring display — so those stay correct. Default dedups (numeric→SUM, date→MAX, other→STRING_AGG) give effective == raw, so there is no behavior change unless a type-changing dedup is chosen.

**Item 1 — missing primary key warning.** A joined data mart with no primary key now shows a **non-blocking** "No primary key" warning in both the relationship list and the graph, since without a key the join cannot deduplicate rows reliably and its metrics can be double-counted (fan-out). Recomputed per fetch, so removing a PK later surfaces it. A PK hidden from reporting still counts as a usable dedup key. This is a highlight, not a hard block (Looker-style symmetric aggregates are explicitly out of scope).

## Scope across the triad
- **Backend**: blended field effective type + governance, `sourceFieldType` for the pre-join path, `hasPrimaryKey` on the relationship response.
- **Web** (app.owox.com): authoring UI (available aggregations follow the dedup, prune only on category change), the "No primary key" warning in list + graph.
- **Extension**: no code change — the report picker consumes backend-resolved governance directly.

## Testing
- Unit coverage across backend (governance, output-control validator, blended query builder, field index, mappers) and web (authoring table, governance, relationship warnings).
- A **real-BigQuery funnel integration test** reproducing the funnel and asserting the corrected counts (e.g. `addToCarts = 2` where the old path returned 1).
- `nest build` + web `tsc -b` clean.

## Follow-ups (out of scope, tracked separately)
- Drop the web `effectiveAggregationType` mirror in favor of server-authoritative governance (coupled to extension parity).
- Product decision: empty post-join numeric group currently returns SQL `NULL` (no `COALESCE`); the manual ground truth showed `0`.
- Items 3–5 of #6733 (UI parity fixes; aggregations on HTTP Data + CLI) ship on separate branches.


apps/backend/src/data-marts/services/blendable-schema.service.ts
```diff
@@ -255,16 +260,33 @@ export class BlendableSchemaService {
...
+        // recategorizes to string, since the STRING_AGG output genuinely IS a string.
+        const effectiveType = computeEffectiveType(
+          field.type as StorageFieldType,
+          dedupFunction,
+          ctx.storageType
+        );
+        // Carry the RAW type BEFORE overwriting `type` with the effective type, so the web can
+        // recompute effective types for type-preserving dedups off the true base (#6733).
+        dto.sourceFieldType = field.type;
+        dto.type = effectiveType;
```
kalnik-a-a, line 278: **Criticality: HIGH / BLOCKING**

Could we please carry the raw source type through every pre-join slice authoring surface before merging? This line correctly changes `BlendedField.type` to the dedup-output (effective) type and preserves the raw value in `sourceFieldType`, but the consumers that author slices still read only `type`.

In the row entry point, `ReportColumnPicker.tsx:660-679` builds a single map from `f.type`; that value reaches both the post-join Filter tab and the pre-join Slice tab through `RowFilterIcon` and `FilterOrSliceEditorPopover.tsx:224-240`. The Output settings → Slices path has the same issue: `ReportColumnPicker.tsx:717-749` stores `type: field.type` in `joinedSources`, and `OutputSettingsDropdown.tsx:267-336` uses it to choose operators and parse values. MCP metadata also exposes only the effective type in `mcp-data-marts.facade.impl.ts:121-130`, although `get_data_mart_details_by_id` explicitly advertises joined fields for slices.

For example, a raw `STRING` field deduped with `COUNT_DISTINCT` becomes effective `INTEGER`. The UI then offers numeric slice operators such as `gt` and hides valid string operators such as `contains`. The backend correctly validates pre-join slices against `sourceFieldType ?? type` in `output-controls-validator.service.ts:186-204`, so a numeric rule authored by the UI is rejected, while a valid string slice cannot be authored.

Please keep the effective type for post-join filters/aggregations, introduce a separate `sliceFieldType` based on `sourceFieldType ?? type` for the Slice tab and `joinedSources`, and expose the raw/slice type in MCP details. Regression coverage should include raw `STRING` and raw `DATE` fields with an effective `INTEGER` type across both slice entry points, existing-slice editing, and MCP metadata.

apps/backend/src/data-marts/services/blendable-schema.service.ts
```diff
@@ -255,16 +260,33 @@ export class BlendableSchemaService {
...
-        dto.aggregateFunction =
-          fieldOverride?.aggregateFunction ?? getDefaultAggregateFunction(field.type);
-        // No override → type-derived allowed set (governance default); explicit `[]` = none allowed.
+        dto.aggregateFunction = dedupFunction;
+        // No override → effective-type governance default; explicit `[]` = none allowed. Existing
+        // explicit overrides are kept verbatim — we only widen the offered set, never rewrite it.
         dto.postJoinAggregations =
           fieldOverride?.postJoinAggregations ??
-          resolveFieldGovernance(field.type).allowedAggregations;
+          resolveFieldGovernance(effectiveType).allowedAggregations;
```
kalnik-a-a, line 287: **Criticality: HIGH / BLOCKING**

Could we please add an explicit backward-compatibility strategy for already persisted report configurations before changing this default? Previously, a raw `STRING` joined field without an explicit `postJoinAggregations` override defaulted to `COUNT` / `COUNT_DISTINCT`. With a `COUNT_DISTINCT` dedup, this fallback now follows the effective `INTEGER` type and becomes `SUM` / `AVG` / `MIN` / `MAX`.

A report saved before this change may therefore still contain an outer `COUNT` or `COUNT_DISTINCT` aggregation that was valid when it was created. After deployment, `output-controls-validator.service.ts:642-655` and `:731-757` rebuild governance from the new default and reject that persisted rule with `AGGREGATION_FUNCTION_NOT_ALLOWED_FOR_FIELD`. This is an immediate runtime regression rather than only an edit/save incompatibility: `blended-report-data.service.ts:49-71` revalidates the configuration on both generated-SQL and report-run paths.

Preserving explicit relationship overrides does not cover reports whose previous type-derived default was never materialized. Please migrate the affected persisted configurations/overrides, grandfather previously saved SQL-valid functions, or provide another controlled normalization path with a visible user-facing outcome. A regression test should load a pre-change report with raw `STRING` + dedup `COUNT_DISTINCT` + outer `COUNT`/`COUNT_DISTINCT` and verify both generated SQL and the normal run path after applying the new governance.

kalnik-a-a (changes requested):
**Overall criticality: HIGH / BLOCKING**

Thank you for the thorough implementation and the extensive coverage around the backend aggregation semantics and missing-primary-key warning. I found two blocking regressions, described inline, that can prevent valid slice authoring and break existing saved reports at runtime. Could we please address both and add the corresponding regression coverage before merging?

apps/backend/src/data-marts/services/blendable-schema.service.ts
zapolsky: We checked production before deciding on a fix. The only persisted report matching this pattern — an outer `COUNT`/`COUNT_DISTINCT` on a joined field whose pre-join dedup is `COUNT`/`COUNT_DISTINCT` (effective `INTEGER`) — is a test fixture: data mart **"Test Sessions"**, a hard-coded funnel (`WITH session AS (SELECT 's1'…'s8' …)`) used while developing this change, not a customer data source. **Zero customer reports are affected.**

The set is also closed going forward: after deploy, governance no longer offers `COUNT`/`COUNT_DISTINCT` for an effective-`INTEGER` field, so the picker will not offer it and save-time validation rejects it — a new report cannot enter this state. Holding such a config requires having saved it before the deploy, and the one pre-existing instance is our own test fixture (which we will remove).

Given zero customer impact and a closed class, we propose not adding a data migration or grandfather path here. If you would still prefer a belt-and-suspenders guard, we are glad to add validator grandfathering (accept a function valid under the field's raw `sourceFieldType`, which we already carry) plus a regression test — just say the word.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/SourceFieldsTable.tsx
```diff
@@ -291,15 +296,37 @@ export function SourceFieldsTable({
...
+                        const override: Partial<BlendedFieldOverride> = { aggregateFunction };
+                        // Reset the analyst-allowed set to the new effective type's default ONLY
+                        // when the dedup change crosses aggregation categories (e.g. string→number):
+                        // that surfaces SUM for an integer-producing dedup instead of pruning to
+                        // "none". A same-category change (SUM→MIN) keeps the current selection; and
+                        // an explicit empty set (`[]` = analyst turned all off) is preserved, never
+                        // silently re-enabled — omit postJoinAggregations in both cases.
+                        const rawType = field.sourceFieldType ?? field.type;
+                        if (
+                          field.postJoinAggregations !== undefined &&
```
romandubovyi: 🟡 **Important:** `field.postJoinAggregations.length` assumes the value is never `null`. If the API ever returns `null`, this will throw. Use `field.postJoinAggregations?.length > 0` (or an explicit null guard) for safety.

apps/backend/test/integration/bigquery.integration.ts
```diff
@@ -1532,3 +1532,252 @@ describeIfCredentials(
...
+      const byKey = new Map(rows.map(r => [`${dateKey(r)}|${r.country}|${r.dataSource}`, r]));
+
+      // THE headline case (was under-counted pre-#6733): s1 logged 2 distinct
+      // hits (h1, h2) under the SAME sessionId; s2 has none. STRING_AGG dedup
+      // used to collapse same-session hits — COUNT_DISTINCT inside the CTE +
+      // outer SUM correctly yields 2.
+      const uaWeb = byKey.get('2026-01-01|UA|WEB')!;
+      expect(uaWeb).toBeDefined();
+      expect(Number(uaWeb['sessionId | COUNTUNIQUE'])).toBe(2);
+      expect(Number(uaWeb['addToCarts | SUM'])).toBe(2);
```
romandubovyi, line 1762: 🟡 **Important:** The assertions depend on the exact `"outputAlias | FUNCTION"` column naming convention produced by the query builder. If that formatting changes, the integration test will break even though the SQL is correct. Consider looking up columns via a normalized key helper or deriving the expected names from the builder.

romandubovyi (commented):
Review complete. Minor defensive-coding and test-brittleness points noted.

apps/backend/src/data-marts/services/blendable-schema.service.ts
zapolsky: Done — post-join filters/aggregations keep the effective (dedup-output) type, and the raw pre-join type is now routed to every slice surface.

**Web** (a3c76e4ed): `FilterOrSliceEditorPopover` gains an optional `sliceFieldType`; the Slice tab and the joined-sources slice list use `sourceFieldType ?? type`, while the Filter tab keeps the effective type. The raw type is read inline from the field (which already carries `sourceFieldType`), so no parallel type map was introduced. Tests assert the Slice tab offers the raw type's operators (STRING `contains`) while the Filter tab keeps the effective INTEGER operators.

**MCP** (18743f258): `get_data_mart_details` now returns a `sliceType` (the raw pre-join type) for a joined field whose dedup changed its type, and documents that slices use it — so the model has the type up front. `query_data_mart` also maps an `INVALID_OPERATOR_FOR_TYPE` on a slice to a clear `invalid_operator` error that points the model at `sliceType`, so it gets both the metadata and an actionable error if it still picks the wrong operator. Tests cover both.

Separately, on the backward-compatibility point in the other thread: we checked production and the only persisted report matching that pattern is our own test fixture, so we did not add a migration there — details in that thread.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/SourceFieldsTable.tsx
zapolsky: Fixed in `ebed4612c`. Collapsed the guard to `(field.postJoinAggregations?.length ?? 0) > 0` — optional chaining treats `null` and `undefined` identically (skip the reset), and `.length === 0` still preserves the explicit empty-set (`[]`) case. Added a regression test that drives a category-crossing dedup change with `postJoinAggregations: null` and asserts the handler no longer throws.

apps/backend/test/integration/bigquery.integration.ts
zapolsky: The literal `"<column> | <TOKEN>"` names are intentional here. This is a credential-gated, real-BigQuery test whose purpose is to pin the exact end-to-end output-header contract — the strings that land as spreadsheet column headers. Deriving the expected names from the builder would make the assertion self-referential: it would pass even if the builder emitted wrong headers, which is exactly the regression this test exists to catch.

The token half (`COUNTUNIQUE`, `SUM`, …) comes from the shared `REPORT_AGGREGATE_FUNCTION_TOKENS` map, which is unit-tested separately, so an intentional rename is caught there and updating this test in lockstep is the intended signal rather than spurious breakage.

If localizing an intentional rename to one place is preferred, I can source the token from `REPORT_AGGREGATE_FUNCTION_TOKENS` while still asserting the `| ` format literally — glad to do that if you think it's worth it.

kalnik-a-a (approved):

PR #1418, opened 2026-07-15
fix(web): stop schema output edits from being silently reverted

prokopetsv:
# Problem

In the Output Schema table, several kinds of edits could get silently undone:

1. Unchecking a column (e.g. Alias, Description) in the column visibility menu would revert to visible again a few seconds later, or whenever the user switched tabs and came back. `useColumnVisibility` re-applied the default visibility set inside a `useEffect` keyed on a freshly-computed `defaultHiddenColumns` object, which got a new object identity on almost every re-render, overwriting the manual toggle each time. Since the schema settings page also lives at a real route, switching tabs unmounted and remounted the component with no persistence layer to restore the choice.
2. Toggling "Hidden from reports" on a field — or clearing an alias/description — while a bulk AI generation was still running would get reverted once generation finished. The AI handlers captured a schema snapshot at click time and, after the slow (polling) generation call resolved, rebuilt the field list from that stale snapshot, discarding any edits made in the meantime.
3. Running a bulk AI generation that filled in nothing new still marked the schema as unsaved, needlessly enabling Save and arming the unsaved-changes guard.

# Solution

- Fixed `useColumnVisibility`'s reset effect to key off a stable, content-derived signature of the hidden-column set instead of the unstable object reference, and to merge rather than overwrite visibility state.
- Added optional `localStorage` persistence to `useColumnVisibility`, scoped per data mart id, so the choice survives route unmounts/remounts and page reloads. Handled the case where the storage key changes on an already-mounted table (navigating between data marts without an unmount) by re-reading the new key's saved value during render, before the persist effect can write stale data into it.
- Extracted a single `mergeGeneratedMetadata` helper used by all three bulk AI handlers. It merges generated values onto the live schema (`schemaRef.current`) instead of the pre-generation snapshot, applies a generated value only when the field's property is both empty and unchanged since the request started (so mid-flight edits and deliberate clears are preserved), and reports whether anything changed so the handlers can skip a no-op schema update.

# Changes

- Reworked `useColumnVisibility` to key its reset effect on a stable hidden-columns signature instead of an unstable object reference, and to merge rather than overwrite visibility state.
- Added optional `storageKey` param to `useColumnVisibility` with `localStorage` read/write for persisting column visibility per data mart.
- Added a render-time re-read of saved visibility when `storageKey` changes without a remount, preventing cross-data-mart state leakage.
- Updated `SchemaTable` to pass a data-mart-scoped storage key (`data-mart-schema-column-visibility-{id}`) into `useColumnVisibility`.
- Extracted `mergeGeneratedMetadata` (plus module-scope `isFilled`) and routed all three bulk AI handlers through it, so generated metadata is applied onto the live schema, mid-generation edits and clears are preserved, and no-op generations no longer mark the schema dirty.

# Rationale

- `SchemaTable` uses a visibility-only hook because `useTableStorage` also owns sorting, pagination, and sizing and assumes a stable storage prefix. The schema table remains mounted when the data mart ID changes, so its hook explicitly rehydrates visibility when the storage key changes.
- Column visibility is intentionally scoped per data mart so changing one schema’s layout does not affect other data marts. The small per-mart localStorage entries are an accepted tradeoff.

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
@@ -33,6 +33,67 @@ type BulkAiScope =
   | DataMartMetadataScope.ALL_FIELD_DESCRIPTIONS
   | DataMartMetadataScope.ALL_FIELD_ALIASES;
 
+const isFilled = (value: string | undefined): boolean => !!value && value.trim() !== '';
+
+interface EditableMetadataField {
+  name: string;
+  alias?: string;
+  description?: string;
+}
+
+type GeneratedMetadata = Pick<EditableMetadataField, 'alias' | 'description'>;
+
+function countByName(fields: readonly EditableMetadataField[]): Map<string, number> {
+  const counts = new Map<string, number>();
+  for (const field of fields) {
+    counts.set(field.name, (counts.get(field.name) ?? 0) + 1);
+  }
+  return counts;
+}
+
+/**
+ * Merges AI-generated alias/description values onto the live schema fields. A value
+ * is applied only when the field's property is empty AND unchanged since the request
+ * started (compared against `originalFields`), so edits or deliberate clears made
+ * while the slow generation call was in flight are never overwritten. Unchanged
+ * fields keep their original reference and `changed` reports whether anything was
+ * applied, letting callers skip a no-op schema update.
+ */
+function mergeGeneratedMetadata(
+  currentFields: readonly EditableMetadataField[],
+  originalFields: readonly EditableMetadataField[],
+  generatedByName: ReadonlyMap<string, GeneratedMetadata>,
+  props: readonly ('alias' | 'description')[]
+): { fields: EditableMetadataField[]; changed: boolean } {
+  const originalByName = new Map(originalFields.map(field => [field.name, field]));
+  // Fields, generated results, and the pre-generation snapshot are all matched by
+  // name (there is no stable field id). Names are user-editable and can briefly
+  // collide, so a duplicated name can't be attributed to a specific row - skip
+  // those to avoid applying AI output to the wrong field.
+  const currentNameCounts = countByName(currentFields);
+  const originalNameCounts = countByName(originalFields);
+  let changed = false;
+  const fields = currentFields.map(field => {
+    const gen = generatedByName.get(field.name);
+    if (!gen) return field;
+    if (currentNameCounts.get(field.name) !== 1 || originalNameCounts.get(field.name) !== 1) {
+      return field;
+    }
+    const original = originalByName.get(field.name);
+    let next = field;
+    for (const prop of props) {
+      const value = gen[prop];
+      const unchanged = !original || field[prop] === original[prop];
+      if (value && !isFilled(field[prop]) && unchanged) {
+        next = { ...next, [prop]: value };
+        changed = true;
+      }
+    }
+    return next;
+  });
+  return { fields, changed };
+}
+
 /**
  * Main component for editing data mart schema settings
  * Uses custom hooks for state management and the SchemaContent component for rendering
@@ -166,20 +227,23 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
     [dataMartId, generateFieldDescription]
   );
 
-  const isFilled = (value: string | undefined): boolean => !!value && value.trim() !== '';
-
+  // The three bulk handlers all merge generated metadata onto the current live
+  // schema (schemaRef), preserving edits made during the slow generation call, and
+  // skip the schema update entirely when nothing was applied. See mergeGeneratedMetadata.
   const handleGenerateAllFieldDescriptions = useCallback(
     async (targetSchema: ResolvedSchema) => {
       if (!dataMartId || !targetSchema) return;
       const generated = await generateAllFieldDescriptions(dataMartId);
       if (!generated) return;
-      const byName = new Map(generated.map(field => [field.name, field.description]));
-      const updated = targetSchema.fields.map(field => {
-        if (isFilled(field.description)) return field;
-        const description = byName.get(field.name);
-        return description ? { ...field, description } : field;
-      });
-      updateSchema(updated as typeof targetSchema.fields);
+      const byName = new Map(generated.map(f => [f.name, { description: f.description }]));
+      const currentFields = schemaRef.current?.fields ?? targetSchema.fields;
+      const { fields, changed } = mergeGeneratedMetadata(
+        currentFields,
+        targetSchema.fields,
+        byName,
+        ['description']
+      );
+      if (changed) updateSchema(fields as typeof targetSchema.fields);
     },
     [dataMartId, generateAllFieldDescriptions, updateSchema]
   );
@@ -189,13 +253,15 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
       if (!dataMartId || !targetSchema) return;
       const generated = await generateAllFieldAliases(dataMartId);
       if (!generated) return;
-      const byName = new Map(generated.map(field => [field.name, field.alias]));
-      const updated = targetSchema.fields.map(field => {
-        if (isFilled(field.alias)) return field;
-        const alias = byName.get(field.name);
-        return alias ? { ...field, alias } : field;
-      });
-      updateSchema(updated as typeof targetSchema.fields);
+      const byName = new Map(generated.map(f => [f.name, { alias: f.alias }]));
+      const currentFields = schemaRef.current?.fields ?? targetSchema.fields;
+      const { fields, changed } = mergeGeneratedMetadata(
+        currentFields,
+        targetSchema.fields,
+        byName,
+        ['alias']
+      );
+      if (changed) updateSchema(fields as typeof targetSchema.fields);
     },
     [dataMartId, generateAllFieldAliases, updateSchema]
   );
@@ -205,16 +271,17 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
       if (!dataMartId || !targetSchema) return;
       const generated = await generateAllFieldMetadata(dataMartId);
       if (!generated) return;
-      const byName = new Map(generated.map(field => [field.name, field]));
-      const updated = targetSchema.fields.map(field => {
-        const gen = byName.get(field.name);
-        if (!gen) return field;
-        const next = { ...field };
-        if (!isFilled(field.alias) && gen.alias) next.alias = gen.alias;
-        if (!isFilled(field.description) && gen.description) next.description = gen.description;
-        return next;
-      });
-      updateSchema(updated as typeof targetSchema.fields);
+      const byName = new Map(
+        generated.map(f => [f.name, { alias: f.alias, description: f.description }])
+      );
+      const currentFields = schemaRef.current?.fields ?? targetSchema.fields;
+      const { fields, changed } = mergeGeneratedMetadata(
+        currentFields,
+        targetSchema.fields,
+        byName,
+        ['alias', 'description']
+      );
+      if (changed) updateSchema(fields as typeof targetSchema.fields);
     },
     [dataMartId, generateAllFieldMetadata, updateSchema]
   );
--- a/apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
@@ -8,6 +8,7 @@ import {
 import { Tooltip, TooltipContent, TooltipTrigger } from '@owox/ui/components/tooltip';
 import { Loader2, Sparkles } from 'lucide-react';
 import { useCallback, useEffect, useRef } from 'react';
+import toast from 'react-hot-toast';
 import { useOutletContext } from 'react-router-dom';
 import type {
   AthenaSchemaField,
@@ -33,7 +34,9 @@ type BulkAiScope =
   | DataMartMetadataScope.ALL_FIELD_DESCRIPTIONS
   | DataMartMetadataScope.ALL_FIELD_ALIASES;
 
-const isFilled = (value: string | undefined): boolean => !!value && value.trim() !== '';
+const normalizeMetadataValue = (value: string | undefined): string => value?.trim() ?? '';
+
+const isFilled = (value: string | undefined): boolean => normalizeMetadataValue(value) !== '';
 
 interface EditableMetadataField {
   name: string;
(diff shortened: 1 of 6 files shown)


apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
```diff
@@ -166,20 +227,23 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
...
       if (!generated) return;
-      const byName = new Map(generated.map(field => [field.name, field.description]));
-      const updated = targetSchema.fields.map(field => {
-        if (isFilled(field.description)) return field;
-        const description = byName.get(field.name);
-        return description ? { ...field, description } : field;
-      });
-      updateSchema(updated as typeof targetSchema.fields);
+      const byName = new Map(generated.map(f => [f.name, { description: f.description }]));
+      const currentFields = schemaRef.current?.fields ?? targetSchema.fields;
```
laskevych, line 290: 🟡 **Important:** `schemaRef.current` isn't guaranteed to belong to `dataMartId` by the time the generation resolves.

`data-marts/:id` reuses the same route element across ids (`routes/index.tsx:92`), so navigating from data mart A to B does **not** unmount this component — the same no-remount scenario the `storageKey` handling in `useColumnVisibility` was added for. `useSchemaState`'s `initialSchema` effect swaps `schema` over to B, so `schemaRef.current.fields` is B's fields while the in-flight `generateAll*` call is still for A's `dataMartId`.

The merge is by name, and names like `id`, `date`, `created_at` collide across data marts, so A's generated aliases/descriptions get written onto B's schema and mark B dirty. Switching from `targetSchema` to `schemaRef.current` actually shrinks the blast radius here (the old code would have replaced B's field list with A's outright), so this isn't a regression — but it's worth closing while we're in this code:

```ts
const generated = await generateAllFieldDescriptions(dataMartId);
if (!generated || dataMart?.id !== dataMartId) return;
```

Same applies to the other two bulk handlers below.

laskevych (commented):
Reviewed the three fixes. The core mechanics look sound: the stable `defaultHiddenColumnsKey` genuinely fixes the clobbering effect, the render-time storage-key re-read is the right React pattern for the no-remount case, and merging onto `schemaRef.current` is a real improvement over the stale snapshot.

Seven comments below: one real cross-data-mart bug that the `schemaRef` switch narrows but doesn't close, an unvalidated `localStorage` cast, and a note that `apps/web/src/hooks/useTableStorage.ts` already implements the persistence this PR hand-rolls.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/hooks/useColumnVisibility.ts
```diff
@@ -1,18 +1,34 @@
...
+/**
+ * Loads the initial visibility for a storage key, layering any saved value over the
+ * default-hidden set so unsaved columns fall back to their default state.
+ */
+function loadColumnVisibility(
+  storageKey: string | undefined,
+  defaultHiddenColumns: Record<string, boolean>
+): Record<string, boolean> {
+  const saved = storageKey
+    ? (storageService.get(storageKey, 'json') as Record<string, boolean> | null)
```
laskevych: 🟡 **Important:** the saved value is cast, not validated, and then spread straight into table state.

`storageService.get(key, 'json')` returns whatever `JSON.parse` produced as long as `typeof parsed === 'object'` — which includes arrays. So a stored `["a"]` spreads as `{0: 'a'}` into `columnVisibility`, and non-boolean values reach TanStack's `VisibilityState` unchecked. Only a corrupted or hand-edited entry triggers it, but the entry is long-lived and there's no self-heal path.

`useTableStorage` guards this exact case with `savedRaw && typeof savedRaw === 'object' && !Array.isArray(savedRaw)` and filters to current column ids — worth mirroring at minimum.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/SchemaTable.tsx
```diff
@@ -91,8 +91,14 @@ export function SchemaTable<T extends BaseSchemaField>({
   const DragContext = dragContext;
   const RowComponent = rowComponent;
 
-  // Column visibility state
-  const { columnVisibility, setColumnVisibility } = useColumnVisibility(columns);
+  // Column visibility state, persisted per data mart so it survives tab switches and reloads
+  const columnVisibilityStorageKey = dataMart?.id
+    ? `data-mart-schema-column-visibility-${dataMart.id}`
+    : undefined;
+  const { columnVisibility, setColumnVisibility } = useColumnVisibility(
```
laskevych, line 98: 💡 **Suggestion:** `apps/web/src/hooks/useTableStorage.ts` already implements column-visibility persistence for this codebase — keyed by a `storageKeyPrefix` with a `-column-visibility` suffix, validated reads, filtering to current column ids, and `excludedColumnsFromVisibility`. It even carries this note:

> *"Effects must NOT depend on configuration to avoid render loops when multiple tables mount or columns are unstable."*

which is precisely the bug class this PR is fixing in the second hook.

This isn't a cosmetic dedupe request — it means the schema table's storage key won't follow the `${prefix}-column-visibility` convention every other table uses, and the validation/filtering fixes have to be re-derived here.

If `useTableStorage` doesn't fit (it also owns sorting/paging/sizing, which this table doesn't need), that's a fine answer — but worth saying so in the PR description, since it'll be the first question a future reader has.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
```diff
@@ -33,6 +33,67 @@ type BulkAiScope =
...
+    const gen = generatedByName.get(field.name);
+    if (!gen) return field;
+    if (currentNameCounts.get(field.name) !== 1 || originalNameCounts.get(field.name) !== 1) {
+      return field;
+    }
+    const original = originalByName.get(field.name);
+    let next = field;
+    for (const prop of props) {
+      const value = gen[prop];
+      const unchanged = !original || field[prop] === original[prop];
```
laskevych: 💡 **Suggestion:** `unchanged` uses strict `===` while `isFilled` normalizes, and the two disagree on `undefined` vs `''`.

`EditableText.handleSave` fires `onValueChange(editedValue.trim())` whenever `editedValue !== value` (`packages/ui/src/components/common/editable-text.tsx:80`). So: a field starts with `alias: undefined`; while generation is in flight the user opens the alias editor, types, deletes it, and applies → `alias` becomes `''`. Now `isFilled('')` is `false` (the field reads as empty) but `'' === undefined` is `false` → `unchanged` is `false` → the generated alias is silently dropped for a field the user never actually filled, and there was nothing there to "deliberately clear."

Comparing normalized values would make the two checks agree:

```ts
const norm = (v: string | undefined) => v?.trim() ?? '';
const unchanged = !original || norm(field[prop]) === norm(original[prop]);
```

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
```diff
@@ -33,6 +33,67 @@ type BulkAiScope =
...
+
+/**
+ * Merges AI-generated alias/description values onto the live schema fields. A value
+ * is applied only when the field's property is empty AND unchanged since the request
+ * started (compared against `originalFields`), so edits or deliberate clears made
+ * while the slow generation call was in flight are never overwritten. Unchanged
+ * fields keep their original reference and `changed` reports whether anything was
+ * applied, letting callers skip a no-op schema update.
+ */
+function mergeGeneratedMetadata(
```
laskevych, line 74: 💡 **Suggestion:** consider making the helper generic so the three `as typeof targetSchema.fields` casts can go:

```ts
function mergeGeneratedMetadata<T extends EditableMetadataField>(
  currentFields: readonly T[],
  originalFields: readonly T[],
  generatedByName: ReadonlyMap<string, GeneratedMetadata>,
  props: readonly ('alias' | 'description')[]
): { fields: T[]; changed: boolean }
```

Right now the return type widens to `EditableMetadataField[]`, which drops `type`, `isPrimaryKey`, `status` and the storage-specific members from the type (they survive at runtime via the spread). That makes each cast load-bearing rather than incidental — it would happily paper over a genuine shape mismatch reaching `updateSchema`, which does runtime field-type discrimination. The casts existed before, but they were previously covering an array that really did have the right type.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
```diff
@@ -33,6 +33,67 @@ type BulkAiScope =
...
+  // name (there is no stable field id). Names are user-editable and can briefly
+  // collide, so a duplicated name can't be attributed to a specific row - skip
+  // those to avoid applying AI output to the wrong field.
+  const currentNameCounts = countByName(currentFields);
+  const originalNameCounts = countByName(originalFields);
+  let changed = false;
+  const fields = currentFields.map(field => {
+    const gen = generatedByName.get(field.name);
+    if (!gen) return field;
+    if (currentNameCounts.get(field.name) !== 1 || originalNameCounts.get(field.name) !== 1) {
```
laskevych, line 92: 🔵 **Question:** both new skip paths are completely silent — is that the intended UX?

If every field is already filled, or a duplicated name makes rows unattributable, `changed` stays `false`, `updateSchema` is skipped, and nothing at all happens: no toast, no dirty state, no dropdown feedback. The user clicks "Generate field aliases", waits out the spinner, and sees an unchanged table with no way to tell whether the request succeeded, was a genuine no-op, or was silently discarded.

Skipping the dirty flag is right (that was bug #3), but a "Nothing to generate — all fields already have aliases" toast would make the difference legible. The duplicate-name skip especially: from the user's side that's indistinguishable from a broken button.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/SchemaTable.tsx
```diff
@@ -91,8 +91,14 @@ export function SchemaTable<T extends BaseSchemaField>({
   const DragContext = dragContext;
   const RowComponent = rowComponent;
 
-  // Column visibility state
-  const { columnVisibility, setColumnVisibility } = useColumnVisibility(columns);
+  // Column visibility state, persisted per data mart so it survives tab switches and reloads
+  const columnVisibilityStorageKey = dataMart?.id
+    ? `data-mart-schema-column-visibility-${dataMart.id}`
```
laskevych, line 96: 🔵 **Question:** was per-data-mart scoping a deliberate product call over a single shared preference?

Two consequences worth confirming: `data-mart-schema-column-visibility-{id}` keys are never cleaned up when a data mart is deleted, so they accumulate for the lifetime of the browser profile; and because the persist effect in `useColumnVisibility` has no dirty condition, it writes an entry on first mount for every data mart whose schema tab is ever opened — even for users who never touch the visibility menu. (The reset effect also always returns a fresh object, so mount costs an extra render plus a second write.)

Neither will exhaust localStorage in practice, so this is a design question rather than a defect. But "which columns I want to see in a schema table" reads to me like a per-user preference rather than a per-data-mart one — a single key would sidestep both the growth and the whole `storageKey`-changed-without-remount branch. If per-mart is intentional, a line in the PR description explaining why would help.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
prokopetsv: Good catch. I added a ref tracking the latest active data-mart ID and now discard stale responses before reading or updating the schema. The guard covers all three bulk generation handlers, with regression tests for switching data marts while each request is pending.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/SchemaTable.tsx
prokopetsv: Thanks for pointing this out. I checked `useTableStorage`, but it does not fit this table: it also initializes and persists sorting, pagination, and sizing, and it does not rehydrate state when `storageKeyPrefix` changes on an existing mount. `SchemaTable` remains mounted when navigating between data marts, so using it would risk carrying the previous mart’s state into the new storage key. I’ll document why this table uses the focused visibility hook in the PR description.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
prokopetsv: Good catch. I’ll normalize the current and original values before comparing them, using the same semantics as `isFilled`. This allows equivalent empty states such as `undefined` and `''` while still protecting a deliberate clear of a previously populated value. I’ll add coverage for both cases.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/hooks/useColumnVisibility.ts
prokopetsv: Good catch. I replaced the unchecked cast with runtime validation and filtering to current column IDs. Invalid and stale entries are discarded, and the sanitized state is persisted back to storage.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
prokopetsv: Agreed. I’ll update the helper to preserve the input field-array type and remove the three call-site assertions. Since `ResolvedSchema.fields` is a union of storage-specific arrays, I’ll preserve that array union in the return type so `updateSchema` continues receiving a correctly typed homogeneous field array. Runtime behavior remains unchanged.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
prokopetsv: Agreed. Keeping the schema clean for a no-op is intentional, but leaving a successful request without feedback is not. I’ll add neutral no-op feedback for each bulk scope and report duplicate-name skips, while continuing to ignore stale results after navigation and relying on `useAiHelper` for request failures.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/SchemaTable.tsx
prokopetsv: Yes, per-data-mart scoping is deliberate. We want each schema table to retain its own layout, rather than a visibility change in one data mart affecting all others. The entries are small and only created for schema tabs that are opened, so the storage growth is an accepted tradeoff. I’ll document this rationale in the PR description.

laskevych (approved):

PR #1419, opened 2026-07-15
fix(reports): report column picker UX fixes (#6733)

zapolsky:
## What & why

A set of small UX fixes for the report column picker (Fibery #6733). All front-end, scoped to
`apps/web/src/features/data-marts/edit/components/ReportColumnPicker/`.

- **Aggregations editor opens on the first invoke.** Picking a column from the "Add aggregation"
  picker set the pending column but left the editor popover closed, so it looked like nothing
  happened until a second click. The editor now opens immediately, and its **Apply button stays
  disabled until a function or bucket is chosen** — a brand-new column can no longer be silently
  discarded by an empty Apply (editing an existing selection keeps Apply enabled so clearing it
  removes the rule).
- **"Unique count" row font.** The auto-generated "Unique count" virtual row rendered in the
  default font while every native field row uses `font-mono`, so it was the lone sans-serif entry.
  It now matches the surrounding rows.
- **Clear the relative-date "Last N days/months" filter input with Backspace.** The value was stored
  as a number and empty input coerced to `0`, so Backspace could never empty the field. It now
  stores the raw text and validates on apply; an empty value is treated as invalid.
- **Delete a filter/slice from the editor popover.** Editing a single existing filter (or slice)
  offered only Cancel/Apply, so the only way to remove it was from the top Filters section. A trash
  button in the editor header (edit mode only) now deletes it directly.

## Testing

- `apps/web` ReportColumnPicker suite: 192/192 (Vitest), `tsc -b` clean.
- New tests: pending-column auto-open flow, Apply-gating, filter/slice delete wiring + list-mode
  absence guards, Last-N clear-to-empty.

## Notes

- The Google Sheets extension mirrors these changes in a separate PR (`OWOX/google-sheets-extension`).
- Known non-blocking follow-up (pre-existing, shared with `OutputSettingsDropdown`): keyboard focus
  is not restored when the auto-opened aggregation editor is dismissed — to be addressed together
  with that sibling.

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/AggregationEditorPopover.tsx
+++ b/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/AggregationEditorPopover.tsx
@@ -136,6 +136,15 @@ export function AggregationEditorPopover(props: AggregationEditorPopoverProps) {
   }
 
   const bucketActive = isDate && bucket !== null;
+  // A brand-new column (nothing pre-selected) must choose a function or bucket — an
+  // empty Apply would silently discard it. Editing an existing selection keeps Apply
+  // enabled so clearing it and applying removes the rule.
+  const hadInitialSelection =
+    (props.initialFunctions?.length ?? 0) > 0 || (props.initialBucket ?? null) !== null;
+  const canApply =
+    props.allowedAggregations.some(fn => functions.includes(fn)) ||
+    bucketActive ||
+    hadInitialSelection;
 
   return (
     <Popover open={props.open} onOpenChange={props.onOpenChange}>
@@ -235,7 +244,7 @@ export function AggregationEditorPopover(props: AggregationEditorPopoverProps) {
           <Button variant='outline' size='sm' onClick={handleCancel}>
             Cancel
           </Button>
-          <Button size='sm' onClick={handleApply}>
+          <Button size='sm' onClick={handleApply} disabled={!canApply}>
             Apply
           </Button>
         </div>
--- a/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/AggregationSettingsDropdown.tsx
+++ b/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/AggregationSettingsDropdown.tsx
@@ -201,6 +201,7 @@ function AggregationSection({
       <div className='mt-2'>
         {pendingColumn ? (
           <RowAggregationIcon
+            key={pendingColumn.name}
             column={pendingColumn.name}
             fieldType={pendingColumn.type}
             displayLabel={pendingColumn.label}
@@ -209,6 +210,10 @@ function AggregationSection({
             activeFunctions={functionsForColumn(pendingColumn.name, aggregations)}
             activeBucket={bucketForColumn(pendingColumn.name, dateTrunc)}
             alwaysVisible
+            autoOpen
+            onClose={() => {
+              setPendingColumn(null);
+            }}
             onApplyDraft={draft => {
               const next = applyAggregationDraft(
                 pendingColumn.name,
--- a/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterOrSliceEditorPopover.tsx
+++ b/apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterOrSliceEditorPopover.tsx
@@ -2,7 +2,7 @@ import { type ReactNode, useCallback, useEffect, useRef, useState } from 'react'
 import { Popover, PopoverContent, PopoverTrigger } from '@owox/ui/components/popover';
 import { Button } from '@owox/ui/components/button';
 import { Label } from '@owox/ui/components/label';
-import { X, Layers } from 'lucide-react';
+import { X, Layers, Trash2 } from 'lucide-react';
 import type { FilterRule } from '../../../shared/types/output-config';
 import { operatorLabelFor } from './output-controls-operators';
 import { summarizeFilterRule } from './filter-rule-summary';
@@ -20,6 +20,8 @@ interface SliceOnlyProps {
   existingSlicesForColumn?: readonly FilterRule[];
   onRemoveExistingAt?: (index: number) => void;
   initialRule?: FilterRule;
+  /** Delete the single slice being edited; shown as a header trash action (edit mode only). */
+  onDelete?: () => void;
 }
 
 export interface FilterOrSliceEditorPopoverProps {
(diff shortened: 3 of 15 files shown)


apps/web/src/features/data-marts/edit/components/ReportColumnPicker/RowFilterIcon.tsx
```diff
@@ -100,6 +100,23 @@ export function RowFilterIcon({
       }
     : undefined;
 
+  // Delete affordances for single-item edit mode (no "Active …" list is shown there).
+  // Unlike apply, delete reads the live index: editFilter/editSlice is the original rule
+  // (still a member of the array), not a post-edit draft, so indexOf stays valid — no snapshot ref needed.
+  const deleteEditedFilter = editFilter
+    ? () => {
+        const idx = activeRules.indexOf(editFilter);
```
romandubovyi, line 108: 💡 **Suggestion:** Apply already snapshots the index on open (`editingFilterIndexRef` / `editingSliceIndexRef`) because rule object identity can flip on parent re-renders. Delete re-resolves via live `indexOf` instead. It works today in single-item mode (`editFilter === activeRules[0]`), but the dual approach is easy to trip over later. Consider reusing the same snapshot refs for delete so both paths share one index contract.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/AggregationSettingsDropdown.tsx
```diff
@@ -209,6 +210,10 @@ function AggregationSection({
             activeFunctions={functionsForColumn(pendingColumn.name, aggregations)}
             activeBucket={bucketForColumn(pendingColumn.name, dateTrunc)}
             alwaysVisible
+            autoOpen
+            onClose={() => {
+              setPendingColumn(null);
```
romandubovyi, line 215: ⚪ **Nit:** `setPendingColumn(null)` in `onApplyDraft` is redundant with this `onClose` path — Apply always closes the popover (`onOpenChange(false)` → `handleOpenChange` → `onClose`). Harmless, but clearing pending only in `onClose` would keep a single source of truth for “editor dismissed → pending reset”.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterValueEditor.tsx
```diff
@@ -263,7 +265,7 @@ export function FilterValueEditor({
               max={3650}
               value={state.relativeN}
               onChange={e => {
-                setState(s => ({ ...s, relativeN: Number(e.target.value) || 0 }));
+                setState(s => ({ ...s, relativeN: e.target.value }));
```
romandubovyi, line 269: 💡 **Suggestion:** Storing `relativeN` as raw text and validating in `buildRule` is the right fix for the Backspace→0 snap. Optional follow-up: the surrounding `type="number"` input still varies across browsers for empty/intermediate values (spinners, `e`/scientific notation). If that UX ever feels sticky again, `type="text"` + `inputMode="numeric"` tends to give a more predictable controlled empty state while keeping the same parse rules.

romandubovyi (commented):
Review complete. No blocking issues — solid UX fixes with good test coverage for auto-open, Apply-gating, filter delete mutual exclusion, and Last-N clear. A few non-blocking suggestions only.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/RowAggregationIcon.autoOpen.test.tsx
```diff
@@ -0,0 +1,109 @@
+import { describe, it, expect, vi } from 'vitest';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { RowAggregationIcon } from './RowAggregationIcon';
+import type { ReportAggregateFunction } from '../../../shared/types/relationship.types';
+
+// Unlike the sibling spec, this mock EXPOSES the controlled `open` prop (as data-open)
+// so we can assert the auto-open behaviour of the dropdown's pending "add" entry —
+// the whole point of item #6733.3 ("the Aggregation block does not show on first invoke").
+vi.mock('@owox/ui/components/popover', () => ({
```
romandubovyi, line 9: 💡 **Suggestion:** The `vi.mock` blocks for `@owox/ui/components/popover` and `@owox/ui/components/select` are duplicated in `AggregationSettingsDropdown.autoOpen.test.tsx`. Consider extracting them to a shared test helper so future mock changes stay in one place.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterValueEditor.tsx
```diff
@@ -110,13 +111,15 @@ function buildRule(args: { column: string; fieldType: string; state: EditorState
 
   if (op === 'relative_date') {
     if (state.relativeKind === 'last_n_days' || state.relativeKind === 'last_n_months') {
-      if (!Number.isInteger(state.relativeN) || state.relativeN <= 0 || state.relativeN > 3650) {
+      const trimmed = state.relativeN.trim();
+      const n = Number(trimmed);
+      if (trimmed === '' || !Number.isInteger(n) || n <= 0 || n > 3650) {
         throw new Error('N must be between 1 and 3650');
```
romandubovyi, line 117: 💡 **Suggestion:** This specific validation message is swallowed by the `useMemo` catch block in `FilterValueEditor` and surfaces only as a generic "Value is required" error in `FilterEditorPopover`. Consider propagating the message to the UI so users understand why Apply is disabled.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterOrSliceEditorPopover.tsx
```diff
@@ -133,17 +135,34 @@ function TabbedPopover(props: TabbedPopoverProps) {
   const hasExistingFilters = !!props.filterProps.existingRules?.length;
   const hasExistingSlices = !!props.sliceProps.existingSlicesForColumn?.length;
   const applyLabel = (tab === 'filter' ? hasExistingFilters : hasExistingSlices) ? 'Add' : 'Apply';
+  const activeDelete = tab === 'filter' ? props.filterProps.onDelete : props.sliceProps.onDelete;
```
romandubovyi, line 138: 🔵 **Question:** When editing a single existing filter and the user switches to the Slice tab, the header delete button disappears because `sliceProps.onDelete` is undefined. Is this intentional? It may feel confusing since the popover is still editing the filter.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.test.tsx
```diff
@@ -881,6 +881,18 @@ describe('ReportColumnPicker Unique count virtual row', () => {
...
+  it('renders the Unique count label in the same monospace font as the native field rows', () => {
+    // The native field rows (e.g. `field.name`) render in font-mono; the virtual
+    // Unique count row must not visually diverge from them (#6733.4 font fix).
+    renderPicker(pkSchema(), ['id', 'name'], {
+      storageType: DataStorageType.GOOGLE_BIGQUERY,
+      outputConfig: { ...baseOutputConfig },
+      onOutputConfigChange: vi.fn(),
+    });
+
+    expect(screen.getByText('Unique count').className).toMatch(/font-mono/);
```
romandubovyi, line 893: ⚪ **Nit:** Consider using `expect(...).toHaveClass('font-mono')` instead of matching `className` with a regex; it's more idiomatic and clearer about the intent.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/AggregationEditorPopover.tsx
```diff
@@ -136,6 +136,15 @@ export function AggregationEditorPopover(props: AggregationEditorPopoverProps) {
   }
 
   const bucketActive = isDate && bucket !== null;
+  // A brand-new column (nothing pre-selected) must choose a function or bucket — an
+  // empty Apply would silently discard it. Editing an existing selection keeps Apply
+  // enabled so clearing it and applying removes the rule.
+  const hadInitialSelection =
+    (props.initialFunctions?.length ?? 0) > 0 || (props.initialBucket ?? null) !== null;
```
romandubovyi, line 143: ⚪ **Nit:** `(props.initialBucket ?? null) !== null` can be simplified to `props.initialBucket != null` for the same result and slightly better readability.

romandubovyi (commented):
Review complete. All relevant tests pass.

romandubovyi (approved):

PR #1422, opened 2026-07-16
feat(mcp): sort query_data_mart results

zapolsky:
## Summary

Adds an optional `sort` parameter to the `query_data_mart` MCP tool — an ordered list of `{ field, direction }` rules (`asc`/`desc`, first rule primary) — so an AI assistant can order query results (for example "top 10 campaigns by spend, descending") without post-processing the rows itself.

Sorting was already available in the Reports, HTTP Data streaming, and CLI query paths; the MCP tool was the only query channel missing it. This PR mirrors the existing `aggregations` / `date_buckets` MCP mapping pattern exactly rather than forking any logic — the underlying `SortConfig` → SQL composer path is reused unchanged.

## Changes

- **`query-data-mart.input.ts`**: `McpSortSchema` (`{ field, direction: 'asc' | 'desc' }`), a `sort` field on `queryDataMartInputSchema`, and `mapMcpSort(sort) → SortConfig` (`field` → `column`, `null` on empty).
- **`query-data-mart.tool.ts`**: maps `parsed.sort` and passes `sortConfig` to the facade; tool description and the `fields` guidance now mention sort.
- **`mcp-data-marts.facade.ts`**: `sortConfig?: SortConfig` on `McpQueryDataMartRequest`.
- **`query-data-mart.service.ts`**: threads `sortConfig` into the ad-hoc read plan and the persisted run metadata.
- **`mcp-system-instructions.ts`** + **`docs/.../mcp.md`**: enumerate sorting.

## Contract

A sorted field must also be listed in `fields` — enforced by the shared output-controls validator (`SORT_COLUMN_NOT_SELECTED`), which the tool's error mapper already surfaces as `field_not_selected`. The OpenAI JSON-Schema `$ref`-free contract is preserved (`McpSortSchema` is a single-use const). Grand totals are unaffected (the totals query is composed with `sortConfig: null`).

## Testing

- 125 MCP unit tests green across input / tool / service / facade / instructions, including: `mapMcpSort` mapping, schema validation, the tool → facade wiring, `sortConfig` threading into the composed read plan, and the persisted run-metadata `query.sort` contract.
- `nest build` and `tsc` clean.

Includes a changeset (`owox: minor`).


apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -124,6 +125,7 @@ export class QueryDataMartService {
     const queryMetadata = {
       fields: r.fields,
       ...(r.filterConfig ? { filters: r.filterConfig } : {}),
+      ...(r.sortConfig ? { sort: r.sortConfig } : {}),
```
kalnik-a-a, line 128: Could we please extend the runtime metadata schema before merging this?

This line adds `sort` to `queryMetadata`, but the production persistence path later runs `record.metadata` through `McpQueryRunMetadataSchema.parse()` in `data-mart-run.service.ts`. The nested `query` schema currently declares `fields`, `filters`, `aggregations`, `dateBuckets`, and `limit`, but not `sort`. Because Zod object schemas strip unknown keys by default, every sorted MCP query is saved without its sort configuration. I reproduced this on the current head: parsing metadata with `query.sort = [{ column: "revenue", direction: "desc" }]` returns `query` containing only `fields` and `limit`.

The new test does not catch this because it asserts the argument passed to a mocked `DataMartRunService`, before the real schema parse. As a result, Run History cannot explain the `ORDER BY`/top-N result, and the PR's persisted-metadata contract is not actually met.

Please add `sort` to `McpQueryRunMetadataSchema.query` (ideally using the core sort rule/schema) and add a `DataMartRunService.recordMcpQueryRun` test that asserts `additionalParams.mcpQuery.query.sort` survives parsing and is persisted. Thank you.

kalnik-a-a (changes requested):
Thank you for the clear end-to-end sort wiring. I found one persistence gap that needs to be addressed before this can be merged; the details are in the inline comment.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
zapolsky: Fixed in eadea9075. You are right — `McpQueryRunMetadataSchema.query` did not declare `sort`, so the `recordMcpQueryRun` parse stripped it and the sort configuration was never persisted. My earlier test asserted the pre-parse argument handed to a mocked `DataMartRunService`, so it ran before the real schema parse and missed this.

Changes:
- Added `sort` to `McpQueryRunMetadataSchema.query` as `z.array(z.unknown()).optional()`, matching the existing `filters` / `aggregations` / `dateBuckets` fields — they stay loose because the values are already validated upstream by the input schema before they reach persistence.
- Added a `DataMartRunService.recordMcpQueryRun` test that passes `query.sort` and asserts it survives the schema parse into `additionalParams.mcpQuery.query.sort` (it fails on the pre-fix schema).

Happy to switch `sort` to the core `SortRule` schema for stricter validation if you would prefer that over matching the sibling fields.

kalnik-a-a (approved):

PR #1425, opened 2026-07-16
fix(web): clearer permission errors that persist until dismissed

prokopetsv:
# Problem

When a user lacked permission to manage owners or configure sharing on a Storage, Destination, or Data Mart, the backend returned vague messages such as "You cannot manage owners of this Storage" — with no indication of which role was required or how to obtain access. On the frontend, these 403 errors surfaced as a toast that auto-dismissed after ~6 seconds, so a user who looked away could miss the explanation entirely. The role names in the messages also didn't match what users actually see in the Members UI (Technical User, Business User, Project Admin).

# Solution

- Rewrote the `ForbiddenException` messages across the affected use-cases to state the reason and the minimum role required, using the UI-facing role names. Wording is entity-specific because the access matrix genuinely differs: Storage and Data Mart require an owner with the Technical User role (Business-User owners are denied access), whereas any Destination owner qualifies.
- Made 403 error toasts persist until the user dismisses them, instead of auto-hiding. The persistent toast is rendered as a real `<button>` so it is keyboard-accessible, carries a stable id so identical repeated errors don't stack, and falls back safely when the server sends no error body.

# Changes

- Updated permission-denied messages in `update-data-storage`, `update-data-destination`, `update-data-mart-owners`, and `update-availability` use-cases to name the required role.
- Added a `persistent` option to `showApiErrorToast` that keeps the toast open (`duration: Infinity`), dedupes by message id, and is dismissable via a focusable button.
- Wired the 403 branch of the API client interceptor to request a persistent toast.
- Guarded `showApiErrorToast` against an undefined extracted API error so it falls back to the default message instead of throwing.

Example of the error

### DIFF
--- a/apps/backend/src/data-marts/use-cases/update-availability.service.ts
+++ b/apps/backend/src/data-marts/use-cases/update-availability.service.ts
@@ -42,7 +42,9 @@ export class UpdateAvailabilityService {
       projectId
     );
     if (!canConfigure)
-      throw new ForbiddenException('You cannot configure sharing for this DataMart');
+      throw new ForbiddenException(
+        'You do not have permission to configure sharing for this Data Mart. You must be the Technical Owner with the Technical User role, or a Project Admin.'
+      );
 
     dm.availableForReporting = availableForReporting;
     dm.availableForMaintenance = availableForMaintenance;
@@ -71,7 +73,9 @@ export class UpdateAvailabilityService {
       projectId
     );
     if (!canConfigure)
-      throw new ForbiddenException('You cannot configure sharing for this Storage');
+      throw new ForbiddenException(
+        'You do not have permission to configure sharing for this Storage. You must be an owner with the Technical User role, or a Project Admin.'
+      );
 
     storage.availableForUse = availableForUse;
     storage.availableForMaintenance = availableForMaintenance;
@@ -100,7 +104,9 @@ export class UpdateAvailabilityService {
       projectId
     );
     if (!canConfigure)
-      throw new ForbiddenException('You cannot configure sharing for this Destination');
+      throw new ForbiddenException(
+        'You do not have permission to configure sharing for this Destination. You must be an owner of this Destination, or a Project Admin.'
+      );
 
     dest.availableForUse = availableForUse;
     dest.availableForMaintenance = availableForMaintenance;
--- a/apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
+++ b/apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
@@ -67,7 +67,9 @@ export class UpdateDataDestinationService {
         command.projectId
       );
       if (!canManage) {
-        throw new ForbiddenException('You cannot manage owners of this Destination');
+        throw new ForbiddenException(
+          'You do not have permission to manage owners of this Destination. You must be an owner of this Destination, or a Project Admin.'
+        );
       }
     }
 
@@ -85,7 +87,9 @@ export class UpdateDataDestinationService {
         command.projectId
       );
       if (!canConfigure) {
-        throw new ForbiddenException('You cannot configure sharing for this Destination');
+        throw new ForbiddenException(
+          'You do not have permission to configure sharing for this Destination. You must be an owner of this Destination, or a Project Admin.'
+        );
       }
     }
 
--- a/apps/web/src/shared/utils/showApiErrorToast.ts
+++ b/apps/web/src/shared/utils/showApiErrorToast.ts
@@ -1,4 +1,5 @@
 import toast from 'react-hot-toast';
+import { createElement } from 'react';
 import type { ApiError } from '../../app/api/api-error.interface';
 import { extractApiError } from '../../app/api/extract-api-error.util';
 
@@ -7,17 +8,52 @@ import { extractApiError } from '../../app/api/extract-api-error.util';
  * - Always shows a fallback message if server message is missing.
  * - Appends details (if provided) in a readable way.
  */
-export function showApiErrorToast(error: unknown, fallbackMessage = 'Something went wrong') {
-  const apiError: ApiError = extractApiError(error);
+export function showApiErrorToast(
+  error: unknown,
+  fallbackMessage = 'Something went wrong',
+  options?: { persistent?: boolean }
+) {
+  const apiError = extractApiError(error) as ApiError | undefined;
 
-  // Ensure we always have a base message
-  let message = apiError.message?.trim() ?? fallbackMessage;
+  // Ensure we always have a base message (empty server messages fall back too)
+  let message = apiError?.message?.trim() ?? fallbackMessage;
+  if (message.length === 0) message = fallbackMessage;
 
   // Append details if present
-  if (apiError.errorDetails?.error?.trim()) {
-    message = `${message}. ${apiError.errorDetails.error}`;
+  const errorDetails = apiError?.errorDetails?.error?.trim();
+  if (errorDetails) {
+    message = `${message}. ${errorDetails}`;
   }
 
-  // Show toast
+  // Persistent toasts stay until the user dismisses them. A stable id keeps
+  // repeated identical errors from stacking; a button makes it keyboard-accessible.
+  if (options?.persistent) {
+    toast.error(
+      t =>
+        createElement(
+          'button',
+          {
+            type: 'button',
+            onClick: () => {
+              toast.dismiss(t.id);
+            },
+            style: {
+              cursor: 'pointer',
+              background: 'none',
+              border: 'none',
+              padding: 0,
+              margin: 0,
+              font: 'inherit',
+              color: 'inherit',
+              textAlign: 'left',
+            },
+          },
+          message
+        ),
+      { duration: Infinity, id: `persistent-error:${message}` }
+    );
+    return;
+  }
+
   toast.error(message);
 }
--- a/apps/web/src/shared/utils/showApiErrorToast.ts
+++ b/apps/web/src/shared/utils/showApiErrorToast.ts
@@ -40,7 +40,7 @@ export function showApiErrorToast(
             'button',
             {
               type: 'button',
-              'aria-label': 'Dismiss',
+              'aria-label': 'Dismiss error',
               onClick: () => {
                 toast.dismiss(t.id);
               },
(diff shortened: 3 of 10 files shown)


apps/web/src/shared/utils/showApiErrorToast.ts
```diff
@@ -7,17 +8,52 @@ import { extractApiError } from '../../app/api/extract-api-error.util';
...
+    message = `${message}. ${errorDetails}`;
+  }
+
+  // Persistent toasts stay until the user dismisses them. A stable id keeps
+  // repeated identical errors from stacking; a button makes it keyboard-accessible.
+  if (options?.persistent) {
+    toast.error(
+      t =>
+        createElement(
+          'button',
```
romandubovyi: 🟡 **Important:** Wrapping the entire message in a `<button>` means clicking anywhere on the text dismisses the toast — including a click that ends a text-selection drag. Since the whole point of making this persistent is to let users read and copy the required-role text, this can dismiss it before they finish selecting/copying. Consider a small dedicated close control instead of making the whole message clickable.

apps/web/src/shared/utils/showApiErrorToast.ts
```diff
@@ -7,17 +8,52 @@ import { extractApiError } from '../../app/api/extract-api-error.util';
...
+              padding: 0,
+              margin: 0,
+              font: 'inherit',
+              color: 'inherit',
+              textAlign: 'left',
+            },
+          },
+          message
+        ),
+      { duration: Infinity, id: `persistent-error:${message}` }
```
romandubovyi, line 61: 🔵 **Question:** Deduping by `message` text means two distinct 403s that happen to produce the same wording (e.g., denied on two different Storages, since the message only names the entity type, not its id) will collapse into a single toast — the second failure is silently swallowed. Is that the intended tradeoff, or should the id include something request-specific?

apps/web/src/shared/utils/showApiErrorToast.ts
```diff
@@ -7,17 +8,52 @@ import { extractApiError } from '../../app/api/extract-api-error.util';
...
  * - Appends details (if provided) in a readable way.
  */
-export function showApiErrorToast(error: unknown, fallbackMessage = 'Something went wrong') {
-  const apiError: ApiError = extractApiError(error);
+export function showApiErrorToast(
+  error: unknown,
+  fallbackMessage = 'Something went wrong',
+  options?: { persistent?: boolean }
+) {
+  const apiError = extractApiError(error) as ApiError | undefined;
```
romandubovyi, line 16: 🟡 **Important:** No test file exists for `showApiErrorToast` (before or after this PR), but this change adds several branches worth locking down: the guard against `extractApiError` returning `undefined` (the bug this PR fixes), the empty-message fallback, and the persistent/dedup behavior. A small unit test would prevent this guard from silently regressing later.

romandubovyi (commented):
Review complete. Backend permission messages checked against the access matrix and existing wording conventions — accurate and consistent. A few notes on the new persistent-toast logic in showApiErrorToast.ts.

apps/web/src/shared/utils/showApiErrorToast.ts
prokopetsv: Good catch — agreed. Clicking to finish a text selection would have dismissed the toast, which defeats the purpose of keeping it readable/copyable. Changed the persistent toast to render the message as plain selectable text with a separate `✕` close button (`aria-label="Dismiss"`, still keyboard-focusable). Text selection and copying now work; dismissal only happens via the explicit close control.

apps/web/src/shared/utils/showApiErrorToast.ts
prokopetsv: Intended tradeoff. Since these toasts never auto-dismiss, the first toast is still visible when the second identical 403 arrives — and with a matching id react-hot-toast updates the existing toast rather than dropping the call, so the failure isn't silent. Because the message only names the entity type, a second toast would be pixel-identical and add no information, just an extra dismiss. Keying the id per-request would bring back the unbounded stacking of permanent toasts this dedup was added to prevent. If per-entity distinction becomes useful, the cleaner path is including the entity name in the backend message — dedup-by-message then separates them automatically.

apps/web/src/shared/utils/showApiErrorToast.ts
prokopetsv: Agreed — added `showApiErrorToast.test.ts` covering the branches this PR introduced: the undefined-body guard (the original bug — asserts no throw + fallback message), the empty/whitespace-message fallback, errorDetails appending, and the persistent path (Infinity duration, message-keyed dedup id, selectable text with a dedicated dismiss button wired to `toast.dismiss`). Only `react-hot-toast` is mocked; the toast body is asserted structurally without a DOM render.

apps/backend/src/data-marts/use-cases/update-availability.service.ts
```diff
@@ -42,7 +42,9 @@ export class UpdateAvailabilityService {
       projectId
     );
     if (!canConfigure)
-      throw new ForbiddenException('You cannot configure sharing for this DataMart');
+      throw new ForbiddenException(
```
romandubovyi, line 45: 🔵 **Question:** The Data Mart message says "Technical Owner with the Technical User role" while the Storage message below says "an owner with the Technical User role". Since the PR description states both require an owner with the Technical User role, is the "Technical Owner" phrasing intentional? It may confuse users if the UI does not expose a "Technical Owner" concept.

apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
```diff
@@ -67,7 +67,9 @@ export class UpdateDataDestinationService {
         command.projectId
       );
       if (!canManage) {
-        throw new ForbiddenException('You cannot manage owners of this Destination');
+        throw new ForbiddenException(
```
romandubovyi, line 70: 🟡 **Important:** The new permission message here is not asserted in `update-data-destination.service.spec.ts`. Please add a test that verifies the exact `ForbiddenException` text for managing owners, similar to the test added in `update-data-mart-owners.service.spec.ts`. This prevents silent regressions if the message changes back.

apps/web/src/shared/utils/showApiErrorToast.ts
```diff
@@ -7,17 +8,60 @@ import { extractApiError } from '../../app/api/extract-api-error.util';
  * - Always shows a fallback message if server message is missing.
  * - Appends details (if provided) in a readable way.
  */
-export function showApiErrorToast(error: unknown, fallbackMessage = 'Something went wrong') {
-  const apiError: ApiError = extractApiError(error);
+export function showApiErrorToast(
```
romandubovyi, line 11: 🟡 **Important:** The new `persistent` branch and the `extractApiError` undefined fallback are not covered by tests. Please add unit tests for `showApiErrorToast` that verify: (1) persistent toasts use `duration: Infinity` and a stable id, (2) the dismiss button renders, and (3) empty/undefined server messages fall back to the provided message.

apps/web/src/shared/utils/showApiErrorToast.ts
```diff
@@ -7,17 +8,60 @@ import { extractApiError } from '../../app/api/extract-api-error.util';
...
+
+  // Persistent toasts stay until the user dismisses them. A stable id keeps
+  // repeated identical errors from stacking. The message text stays selectable
+  // (users may want to copy the required role), so dismissal is a dedicated,
+  // keyboard-accessible close button instead of the whole message.
+  if (options?.persistent) {
+    toast.error(
+      t =>
+        createElement(
+          'span',
```
romandubovyi, line 36: 💡 **Suggestion:** Add an `aria-label` (e.g., `aria-label="Dismiss error"`) to the persistent toast button so screen reader users understand the button dismisses the notification, not just re-reads the error text.

romandubovyi, line 36: 🟢 **Praise:** Nice UX improvement — making 403 toasts persistent and keyboard-accessible via a real `<button>` with stable deduplication helps users actually read and act on permission errors.

romandubovyi (approved):
LGTM. No blockers. Left a few inline comments with suggestions and questions.

apps/backend/src/data-marts/use-cases/update-availability.service.ts
prokopetsv: Intentional. The Data Mart UI exposes two distinct ownership roles — Technical Owner and Business Owner (see the Data Mart overview page and the owners columns) — and only the Technical Owner can configure sharing, and only when their role is Technical User. The in-app help even states that assigning Technical Owner to a Business User grants nothing. Storage has no such split (a single "Storage Owner" that is the technical owner), which is why its message just says "an owner." So the wording deliberately mirrors each entity's UI vocabulary. Good catch on the PR description, though — it over-simplified by saying "owner" for both; I've updated it to reflect the Technical Owner distinction for Data Marts.

apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
prokopetsv: Added — `update-data-destination.service.spec.ts` now has a `permission checks` block asserting the exact ForbiddenException text for managing owners (matching the mart-owners precedent), and I included the configure-sharing message too since it was likewise unasserted. Both force `canAccess` to return false and verify the thrown message, so a regression in either string fails the suite.

apps/web/src/shared/utils/showApiErrorToast.ts
prokopetsv: These tests already exist in `apps/web/src/shared/utils/showApiErrorToast.test.ts` (commit ef8772a8) — you may be looking at an earlier revision. It covers all three: `duration: Infinity` + the stable `persistent-error:<message>` id, the dismiss button rendering (`aria-label="Dismiss"`) and its click calling `toast.dismiss`, and the empty/undefined-message fallback (both a missing response body and a whitespace-only message). Happy to add more cases if any specific path is still uncovered.

apps/web/src/shared/utils/showApiErrorToast.ts
prokopetsv: Good call — the button already had an `aria-label`, but I've adopted your clearer wording (`"Dismiss error"` instead of just `"Dismiss"`) so screen-reader users know it dismisses the notification. Updated the unit test assertion to match.

PR #1428, opened 2026-07-17
feat(auth): enforce view-only session restrictions

romandubovyi:
## Summary

- Treat view-only sessions as read-only at the API layer: allow `GET` / `HEAD` / `OPTIONS`, reject `POST` / `PUT` / `PATCH` / `DELETE` with **403** and code `ACTION_NOT_ALLOWED_IN_VIEW_ONLY_MODE`.
- Add optional `Payload.viewOnly` to the IDP protocol, claim resolution via `resolveViewOnlyFromClaims` (canonical `viewOnly` only — no aliases such as `readOnly`), and normalization in `idp-owox-better-auth` `toPayload`.
- Surface `viewOnly` on auth context (`GET /api/auth/context`), request/CLS context, and the web user model (`User.viewOnly` + `isViewOnlySession`).
- Reject MCP OAuth authorization (`GET /oauth/authorize`) for view-only sessions so they cannot mint MCP tokens that would bypass REST restrictions via write tools.
- Suppress client analytics for view-only sessions: skip GTM bootstrap and no-op dataLayer events (including identify/logout).

## Behavior notes

- API enforcement lives in `IdpGuard` for **required** authenticated Nest REST routes. IDP session endpoints such as `/auth/access-token` stay outside this guard so clients can still refresh tokens.
- **`Role.none()` / optional routes intentionally skip IDP auth and view-only checks.** Those endpoints authenticate outside the user IDP token path (for example `InternalApiGuard` service-account tokens on legacy data-marts sync, and `GoogleJwtBody` connector JWTs on Looker Studio). A user view-only access token is not accepted as credentials there.
- **MCP:** view-only sessions cannot complete OAuth authorize, so they cannot obtain new MCP tokens. MCP write tools remain gated by `mcp:write` scopes on tokens that were already issued outside view-only mode.
- Until identity tokens actually include a view-only signal, normal sessions are unchanged (dormant until the claim is present).
- Analytics suppression is driven by `user.viewOnly` on the web client; GTM is not loaded for those sessions, so tags that would load PostHog / session recording do not start.

## Test plan

- [x] Unit: `@owox/idp-protocol` claim resolution + safe methods
- [x] Unit: `idp-owox-better-auth` payload mapper
- [x] Unit: `IdpGuard` method matrix, error code/status, projections skip
- [x] Unit: `IdpGuard` documents Role.none() skip of view-only (no token parse)
- [x] Unit: OAuth authorize rejects view-only sessions (header token + refresh cookie)
- [x] Unit: auth context controller/decorator + exception filter body
- [x] Unit: web `currentUserResponseToUser` / `isViewOnlySession`
- [x] Unit: web dataLayer analytics gate
- [x] Unit: web GTM skipped for view-only sessions
- [ ] Manual: token with `viewOnly` → mutating API returns 403
- [ ] Manual: same session → GET endpoints and `/auth/access-token` still succeed
- [ ] Manual: view-only session cannot complete MCP OAuth authorize
- [ ] Manual: `/api/auth/context` and `/auth/api/user` expose `viewOnly: true` when applicable
- [ ] Manual: view-only session does not load GTM / does not emit dataLayer events

### DIFF
--- a/apps/backend/src/idp/guards/idp.guard.ts
+++ b/apps/backend/src/idp/guards/idp.guard.ts
@@ -3,8 +3,11 @@ import { Request } from 'express';
 import {
   AuthenticationError,
   AuthorizationError,
+  isSafeHttpMethodForViewOnly,
+  isViewOnlyPayload,
   Payload,
   Role as RoleType,
+  ViewOnlyModeError,
 } from '@owox/idp-protocol';
 import { IdpProjectionsService } from '../services/idp-projections.service';
 import { Strategy } from '../types';
@@ -28,6 +31,8 @@ export interface AuthenticatedRequest extends Request {
     projectTitle?: string;
     authFlow?: string;
     apiKeyId?: string;
+    /** True when the session is in view-only mode. */
+    viewOnly?: boolean;
   };
 }
 
@@ -60,6 +65,10 @@ export class IdpGuard implements CanActivate {
 
     const request = context.switchToHttp().getRequest<AuthenticatedRequest>();
 
+    // Role.none() / optional auth: IdpGuard does not parse user tokens and does not
+    // apply view-only restrictions. Callers of these routes authenticate outside IDP
+    // (e.g. InternalApiGuard service-account tokens, GoogleJwtBody connector JWTs).
+    // A user view-only access token is not accepted as credentials on those paths.
     if (roleConfig.optional) {
       return true;
     }
@@ -68,6 +77,10 @@ export class IdpGuard implements CanActivate {
       const tokenPayload = await this.authenticateUser(request, roleConfig.strategy);
       this.checkApiKeyUsageRestrictions(tokenPayload, Boolean(rejectApiKeyAuth));
       this.checkApiKeyHeaderBinding(request, tokenPayload);
+      this.checkViewOnlyRestrictions(request, tokenPayload);
+
+      // Propagate only when true so normal sessions stay free of the flag.
+      const viewOnly = isViewOnlyPayload(tokenPayload) || undefined;
 
       request.idpContext = {
         userId: tokenPayload.userId,
@@ -79,6 +92,7 @@ export class IdpGuard implements CanActivate {
         projectTitle: tokenPayload.projectTitle,
         authFlow: tokenPayload.authFlow,
         apiKeyId: tokenPayload.apiKeyId,
+        viewOnly,
       };
 
       this.cls.set(AUTH_CONTEXT, {
@@ -87,6 +101,7 @@ export class IdpGuard implements CanActivate {
         roles: tokenPayload.roles,
         authFlow: tokenPayload.authFlow,
         apiKeyId: tokenPayload.apiKeyId,
+        viewOnly,
       });
 
       if (request && STATE_CHANGING_METHODS.includes(request.method)) {
@@ -94,6 +109,7 @@ export class IdpGuard implements CanActivate {
         void this.idpProjectionsService.updateProjectionsFromIdpPayload(tokenPayload);
       }
     } catch (error) {
+      // AuthorizationError covers ViewOnlyModeError (subclass) and role/API-key denials.
       if (error instanceof UnauthorizedException || error instanceof AuthorizationError) {
         throw error;
       }
@@ -124,6 +140,33 @@ export class IdpGuard implements CanActivate {
     }
 
     return tokenPayload;
+  }
+
+  /**
+   * Blocks state-changing requests when the session is in view-only mode.
+   * Safe methods (GET/HEAD/OPTIONS) remain allowed.
+   *
+   * Scope: only routes that go through authenticateUser (required @Auth roles).
+   * Intentionally NOT applied to Role.none() / optional routes (service or
+   * connector auth, not user IDP).
+   *
+   * MCP is a separate auth path; view-only sessions are blocked from minting
+   * MCP tokens in OAuthAuthorizationController so write tools cannot bypass
+   * this guard.
+   *
+   * View-only detection is delegated to idp-protocol so claim resolution can
+   * evolve independently of this guard.
+   */
+  private checkViewOnlyRestrictions(request: AuthenticatedRequest, tokenPayload: Payload): void {
+    if (!isViewOnlyPayload(tokenPayload)) {
+      return;
+    }
+
+    if (isSafeHttpMethodForViewOnly(request.method)) {
+      return;
+    }
+
+    throw new ViewOnlyModeError();
   }
 
   private checkApiKeyHeaderBinding(request: AuthenticatedRequest, tokenPayload: Payload): void {
--- a/apps/backend/src/idp/guards/idp.guard.ts
+++ b/apps/backend/src/idp/guards/idp.guard.ts
@@ -15,6 +15,7 @@ import { Reflector } from '@nestjs/core';
 import { IdpProviderService } from '../services/idp-provider.service';
 import { ClsService } from 'nestjs-cls';
 import { REJECT_API_KEY_AUTH_METADATA } from '../decorators/reject-api-key-auth.decorator';
+import { VIEW_ONLY_SAFE_METADATA } from '../decorators/view-only-safe.decorator';
 
 export interface AuthenticatedRequest extends Request {
   idpContext: {
@@ -57,6 +58,9 @@ export class IdpGuard implements CanActivate {
       REJECT_API_KEY_AUTH_METADATA,
       [context.getHandler(), context.getClass()]
     );
+    const viewOnlySafe = this.reflector.getAllAndOverride<boolean>(VIEW_ONLY_SAFE_METADATA, [
+      context.getHandler(),
+    ]);
 
     if (!roleConfig) {
       throw new AuthenticationError('No role configuration found');
@@ -76,7 +80,7 @@ export class IdpGuard implements CanActivate {
       const tokenPayload = await this.authenticateUser(request, roleConfig.strategy);
       this.checkApiKeyUsageRestrictions(tokenPayload, Boolean(rejectApiKeyAuth));
       this.checkApiKeyHeaderBinding(request, tokenPayload);
-      this.checkViewOnlyRestrictions(request, tokenPayload);
+      this.checkViewOnlyRestrictions(request, tokenPayload, Boolean(viewOnlySafe));
 
       // Propagate only when true so normal sessions stay free of the flag.
       const viewOnly = isViewOnlyPayload(tokenPayload) || undefined;
@@ -146,8 +150,9 @@ export class IdpGuard implements CanActivate {
 
   /**
    * Blocks POST/PUT/PATCH/DELETE when the session is in view-only mode.
-   * GET/HEAD/OPTIONS remain allowed. Some POSTs are read-semantics in practice;
-   * that trade-off is accepted for a simple method-based policy.
+   * GET/HEAD/OPTIONS remain allowed. A POST endpoint may opt in with
+   * @ViewOnlySafe only when it has read semantics and does not mutate project
+   * or external data. The escape hatch never applies to PUT/PATCH/DELETE.
    *
    * Scope: only routes that go through authenticateUser (required @Auth roles).
    * Intentionally NOT applied to Role.none() / optional routes (service or
@@ -159,12 +164,20 @@ export class IdpGuard implements CanActivate {
    * are not re-checked against live session viewOnly here — that is Identity /
    * MCP token lifecycle responsibility.
    */
-  private checkViewOnlyRestrictions(request: AuthenticatedRequest, tokenPayload: Payload): void {
+  private checkViewOnlyRestrictions(
+    request: AuthenticatedRequest,
+    tokenPayload: Payload,
+    viewOnlySafe: boolean
+  ): void {
     if (!isViewOnlyPayload(tokenPayload)) {
       return;
     }
 
     if (!STATE_CHANGING_METHODS.includes(request.method)) {
+      return;
+    }
+
+    if (request.method === 'POST' && viewOnlySafe) {
       return;
     }
 
--- a/apps/backend/src/idp/guards/idp.guard.ts
+++ b/apps/backend/src/idp/guards/idp.guard.ts
@@ -3,7 +3,6 @@ import { Request } from 'express';
 import {
   AuthenticationError,
   AuthorizationError,
-  isSafeHttpMethodForViewOnly,
   isViewOnlyPayload,
   Payload,
   Role as RoleType,
@@ -104,6 +103,9 @@ export class IdpGuard implements CanActivate {
         viewOnly,
       });
 
+      // Strategy.PARSE reads viewOnly only from the locally verified JWT. If
+      // Identity flips a session to view-only without revoking/re-issuing access
+      // tokens, write access remains until the token expires.
       if (request && STATE_CHANGING_METHODS.includes(request.method)) {
         // Update IDP projections in the background
         void this.idpProjectionsService.updateProjectionsFromIdpPayload(tokenPayload);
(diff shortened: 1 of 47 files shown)


apps/backend/src/idp/guards/idp.guard.ts
```diff
@@ -126,6 +142,33 @@ export class IdpGuard implements CanActivate {
...
+   * connector auth, not user IDP).
+   *
+   * MCP is a separate auth path; view-only sessions are blocked from minting
+   * MCP tokens in OAuthAuthorizationController so write tools cannot bypass
+   * this guard.
+   *
+   * View-only detection is delegated to idp-protocol so claim resolution can
+   * evolve independently of this guard.
+   */
+  private checkViewOnlyRestrictions(request: AuthenticatedRequest, tokenPayload: Payload): void {
```
laskevych: 🟡 **Important (blocking):** Pure method-based blocking 403s several read-semantics POST endpoints, breaking the read experience view-only mode must preserve. Confirmed cases, all `@Auth(Role.viewer(...))`:

- `POST /data-marts/health-status` — batch status read used by the data-mart list; health badges will fail for view-only users.
- `POST /markdown/parse-to-html` — pure transform used to render markdown content.
- `POST /intercom/jwt` — called on every app load by `IntercomChat`; since the apiClient interceptor shows a **persistent** "Access forbidden" toast on any 403 without `skipErrorToast`, every view-only page load will display a spurious error toast.
- `POST /data-storage/:id/validate-access` and the insight SQL-preview trigger — read/derive operations.

Consider a per-route escape hatch consumed by this guard (e.g. a `@ViewOnlySafe()` decorator), keeping method-based blocking as the default.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
```diff
@@ -51,6 +53,12 @@ export class OAuthAuthorizationController {
       return;
     }
 
+    // View-only sessions must not mint MCP tokens: write tools would otherwise
+    // bypass IdpGuard and mutate project state under the impersonated user.
+    if (authorization.context.viewOnly === true) {
```
laskevych, line 58: 🟡 **Important / 🔵 Question (blocking unless covered by Identity):** View-only is enforced only at authorization-code mint time. `POST /oauth/token` supports `grant_type=refresh_token` and delegates to `exchangeToken` with no `viewOnly` check; `McpAuthGuard.assertPayload` validates authFlow/resource/scopes/roles but not `viewOnly`; and `McpTokenPayloadSchema` has no `viewOnly` field at all. So an MCP token minted before a session became view-only keeps producing write-capable access tokens via the refresh grant — the exact bypass this comment says the check prevents.

Does the Identity service revoke/re-validate MCP refresh tokens when a user becomes view-only? If not, `viewOnly` should be carried into `McpTokenPayload` and checked in `McpAuthGuard` / token exchange.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
```diff
@@ -51,6 +53,12 @@ export class OAuthAuthorizationController {
       return;
     }
 
+    // View-only sessions must not mint MCP tokens: write tools would otherwise
```
laskevych, line 56: 💡 **Suggestion (non-blocking):** Two ordering/shape notes on this rejection:

1. In the refresh-cookie path, `resolveAuthorizationContext` rotates and persists the refresh token (`persistRefreshTokenIfReturned`) *before* this throw — a completed side effect on a denied request. Checking `viewOnly` before rotation would avoid that.
2. The rejection renders as raw JSON 403 in the browser mid-OAuth-flow, so the waiting client never receives an `error=access_denied` callback. This matches the controller's existing convention for other denials, so not a regression — but since `redirect_uri` is validated by this point, a proper OAuth error redirect would give MCP clients a clean failure instead of a hang.

apps/web/src/app/gtm/GoogleTagManager.tsx
```diff
@@ -68,17 +69,24 @@ function extractContainerId(flags: Record<string, unknown>): string {
...
+ * Skipped entirely in view-only sessions so PostHog / GA4 tags and
+ * session recordings never start for these read-only sessions.
+ */
 export function GoogleTagManager(): null {
   const { flags, callState } = useFlags();
+  const { user } = useAuth();
 
   useEffect(() => {
     if (callState !== RequestStatus.LOADED || !flags) return;
+    if (isViewOnlySession(user)) return;
```
laskevych, line 89: 🟡 **Important (non-blocking):** There's no teardown for the normal → view-only transition mid-SPA. `AuthContext.refresh()` re-fetches the user on every 401-triggered token renewal, so `viewOnly` can flip mid-session; this effect then only early-returns, and `suppressClientAnalytics()` gates only the app's own dataLayer pushes — GA4/PostHog tags and session recordings inside the already-installed GTM container keep running autonomously. Notably the reverse transition *is* handled (`setAnalyticsDisabled(false)`), so mid-session flips are an anticipated state. If a full teardown is impractical, a forced reload on the flip — or a comment documenting the one-way limitation — would help.

apps/web/src/app/gtm/GoogleTagManager.tsx
```diff
@@ -68,17 +69,24 @@ function extractContainerId(flags: Record<string, unknown>): string {
...
+ * Loads GTM when configured via app flags.
+ * Skipped entirely in view-only sessions so PostHog / GA4 tags and
+ * session recordings never start for these read-only sessions.
+ */
 export function GoogleTagManager(): null {
   const { flags, callState } = useFlags();
+  const { user } = useAuth();
 
   useEffect(() => {
     if (callState !== RequestStatus.LOADED || !flags) return;
```
laskevych: 🟢 **Nit (non-blocking):** `isViewOnlySession(null)` is `false`, so this gate defaults *open* if the effect ever runs with flags loaded but `user` still `null`. Today that can't happen — but only because `AppBootstrap` withholds children until authentication resolves, an invariant this file has no visibility into. A one-line `if (!user) return;` makes the gate locally safe against future eager/cached flags loading. Optionally, depending on a derived `isViewOnlySession(user)` boolean instead of the `user` object would also stop the effect re-running on every token refresh.

apps/backend/src/idp/guards/idp.guard.ts
```diff
@@ -68,6 +77,10 @@ export class IdpGuard implements CanActivate {
       const tokenPayload = await this.authenticateUser(request, roleConfig.strategy);
       this.checkApiKeyUsageRestrictions(tokenPayload, Boolean(rejectApiKeyAuth));
       this.checkApiKeyHeaderBinding(request, tokenPayload);
+      this.checkViewOnlyRestrictions(request, tokenPayload);
```
laskevych: 🔵 **Question (non-blocking):** With `Strategy.PARSE` (~77 `@Auth` routes, including state-changing ones like the insight run triggers), `viewOnly` comes solely from the locally-verified JWT — a session flipped to view-only server-side keeps write access until its access token expires. Is that window acceptable given the token TTL, i.e. does view-only activation always accompany fresh token issuance / revocation? Worth documenting the assumption either way; nothing currently enforces or tests it.

apps/backend/src/idp/guards/idp.guard.ts
```diff
@@ -126,6 +142,33 @@ export class IdpGuard implements CanActivate {
...
+   *
+   * View-only detection is delegated to idp-protocol so claim resolution can
+   * evolve independently of this guard.
+   */
+  private checkViewOnlyRestrictions(request: AuthenticatedRequest, tokenPayload: Payload): void {
+    if (!isViewOnlyPayload(tokenPayload)) {
+      return;
+    }
+
+    if (isSafeHttpMethodForViewOnly(request.method)) {
```
laskevych: 💡 **Suggestion (non-blocking):** The guard now classifies methods against two independently maintained, mutually-inverse lists: local `STATE_CHANGING_METHODS` (case-sensitive `.includes`) and `VIEW_ONLY_SAFE_HTTP_METHODS` here (case-insensitive, fail-closed). They already disagree on edge inputs — for a lowercase `post`, the projections update is silently skipped while the view-only check still blocks. The projections check could become `!isSafeHttpMethodForViewOnly(request.method)`, deleting the local const so there's a single definition of "mutating".

packages/idp-protocol/src/utils/view-only.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Resolves whether token claims represent a view-only session.
+ *
+ * Only the protocol contract field `viewOnly` is considered. Provider-specific
+ * flags (for example API-key `readOnly`) must be normalized into `viewOnly` by
+ * the identity provider / token issuer — they are not interpreted here.
+ */
+export function resolveViewOnlyFromClaims(
+  claims: Record<string, unknown> | null | undefined
```
laskevych: 💡 **Suggestion (non-blocking):** `resolveViewOnlyFromClaims` and `isViewOnlyPayload` have identical bodies (`x?.viewOnly === true`) and differ only in parameter type — and the former's single caller has to cast `src as Record<string, unknown>` anyway. Since the guard's JSDoc says claim resolution "can evolve independently", having two copies means an evolution can silently apply to the issuer layer but not the enforcement layer (or vice versa). Suggest keeping one helper with a widened signature.

laskevych (commented):
## View-only session restrictions — review

The core mechanism (guard check, dedicated error type, claim mapping, analytics gate) is solid and well tested. The findings below are at the edges: two enforcement gaps outside the guard's reach, a UX break in the read-only experience itself, and a few consolidation opportunities. Most are attached as line comments; two findings concern files outside this diff, so they live here.

---

### 🔴 Critical (blocking): `POST /auth/onboarding` escapes view-only enforcement entirely

`packages/idp-owox-better-auth/src/controllers/onboarding-controller.ts` registers `app.post('/auth/onboarding', ...)` directly on Express via `provider.registerRoutes()` (before Nest mounts), so it never passes `IdpGuard`. It authenticates from the `refreshToken` cookie in `resolveUserContext` and persists answers via `onboardingService.saveAnswers` with no `viewOnly` check anywhere in the path. Since view-only sessions hold a valid refresh cookie (the PR's own test "rejects view-only sessions resolved via refresh cookie" proves it), this write path stays open — contradicting the changeset claim that state-changing methods return 403. Suggest checking `payload.viewOnly` in `resolveUserContext` (the parsed payload already carries it after this PR).

### 🔵 Question (non-blocking): is web UI gating an intentional scope split?

`ACTION_NOT_ALLOWED_IN_VIEW_ONLY_MODE` isn't referenced anywhere in `apps/web`, and `isViewOnlySession` is wired only into GTM/AuthContext — so mutation buttons stay enabled and a blocked action surfaces as the same generic persistent "Access forbidden - insufficient permissions" toast (`apps/web/src/app/api/apiClient.ts:113`) as any RBAC denial. The PR's own docs promise "Clients should disable analytics and mutating UI actions". If UI gating is deferred to a follow-up, mapping the new error code to a view-only-specific message in the apiClient interceptor would be a cheap interim win.

romandubovyi:
## Review follow-up

Thanks for the thorough review. Summary of how we handled the findings:

### Fixed
- **`POST /auth/onboarding` write bypass** — Express path now reads `viewOnly` from the refresh-token payload, rejects mutating submit with `403` + `ACTION_NOT_ALLOWED_IN_VIEW_ONLY_MODE`, and redirects the onboarding page home. Covered by unit tests.
- **Claim helpers** — `isViewOnlyPayload` delegates to `resolveViewOnlyFromClaims` (single source of truth).
- **Method lists** — dropped `isSafeHttpMethodForViewOnly` from `@owox/idp-protocol`; `IdpGuard` uses one local `STATE_CHANGING_METHODS` list.
- **GTM** — fail-closed when `user` is null; documented mid-session normal→view-only limitation (no full GTM teardown without reload).
- **apiClient** — maps `ACTION_NOT_ALLOWED_IN_VIEW_ONLY_MODE` to a clearer toast (still fail-after-attempt, not proactive UI disable).

### Intentional scope / accepted trade-offs
- **No `@ViewOnlySafe` / per-route escape hatch.** View-only is a **session safety** control (block mutating API traffic), not a polished product read-only UI. Some POSTs are read-semantics in practice; we accept that they may 403 under this simple method-based policy rather than obscure the rule with a second decorator.
- **UI mutation buttons may stay enabled.** Blocked actions fail after the request — that is acceptable for this use case. Client work here is analytics suppression, not full UI gating.
- **MCP refresh / existing tokens** — this PR blocks **new** MCP OAuth authorization for view-only sessions. Pre-existing MCP tokens and `refresh_token` grants are Identity lifecycle (revoke/re-issue when view-only flips, or carry `viewOnly` into MCP tokens later). Documented on the guard.
- **`Strategy.PARSE` JWT window** — enforcement uses the access-token claim until expiry/re-issue; documented. Token TTL / revocation when flipping view-only is an Identity concern.
- **OAuth cookie rotation before deny / OAuth error redirect shape** — acknowledged, not changed in this PR (see thread replies).

Pushed in `0faf21ed3`.

apps/backend/src/idp/guards/idp.guard.ts
romandubovyi: **Won’t add a per-route escape hatch in this PR.**

We keep a single, transparent rule: block `POST` / `PUT` / `PATCH` / `DELETE` for view-only sessions.

Some POSTs are read-semantics in practice (health-status, markdown parse, intercom JWT, validate-access, SQL preview). We accept that trade-off: view-only may break those paths until/unless we revisit specific endpoints later in a dedicated change — not via a second decorator that obscures the policy.

Rationale: this mode is about **safety of mutating requests**, not preserving a perfect browse UX.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
romandubovyi: **Partly out of scope / needs Identity.**

This PR stops **new** MCP OAuth authorization for view-only sessions. Pre-existing MCP tokens (and `refresh_token` grants for them) are a separate lifecycle: Identity should revoke/re-issue when the session becomes view-only, or we carry `viewOnly` into MCP tokens and enforce in `McpAuthGuard` / token exchange.

Documented on the guard. Happy to track a follow-up with Identity if current revoke semantics are not enough.

apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts
romandubovyi: **Acknowledged, not changing in this PR.**

1. Refresh rotation happens inside `refreshToken()` before we can read claims; skipping cookie persist on deny can leave the SPA cookie stale if Identity already rotated. Side effect on deny is imperfect but preferable to logging the user out of the app.
2. Raw JSON 403 matches existing denial style in this controller (not a regression). OAuth `error=access_denied` redirect is a nice follow-up, not required for view-only safety.

apps/web/src/app/gtm/GoogleTagManager.tsx
romandubovyi: **Documented the one-way limitation.**

Full GTM teardown / forced reload is out of scope. After bootstrap we only stop our own dataLayer pushes via `suppressClientAnalytics`. Mid-session flip without reload is an edge case; if we need a hard stop of tags later, forced reload is the simple option.

apps/web/src/app/gtm/GoogleTagManager.tsx
romandubovyi: **Done.** Gate is now fail-closed: `if (!user) return` before installing GTM, with a unit test.

apps/backend/src/idp/guards/idp.guard.ts
romandubovyi: **Accepted / documented.**

With `PARSE`, enforcement uses the access-token claim until expiry (or re-issue). We assume Identity issues/revokes tokens when view-only is flipped; ODM does not re-introspect on every `PARSE` route. Comment added next to the guard path. If TTL is too long for the threat model, that is an Identity token-lifetime / revocation concern.

apps/backend/src/idp/guards/idp.guard.ts
romandubovyi: **Simplified further.**

Dropped `isSafeHttpMethodForViewOnly` / `VIEW_ONLY_SAFE_HTTP_METHODS` from `@owox/idp-protocol` (HTTP policy is not a protocol concern).

`IdpGuard` uses a single local `STATE_CHANGING_METHODS` list for both view-only blocking and projection updates.

packages/idp-protocol/src/utils/view-only.ts
romandubovyi: **Done.**

`isViewOnlyPayload` delegates to `resolveViewOnlyFromClaims` with a widened claims type; the mapper no longer needs a cast.

laskevych (approved):

amarchenk0 (approved):

PR #1430, opened 2026-07-20
feat(mcp): support every destination type in add_report and update_report

laskevych:
## Summary

Extends the MCP report tools from Google-Sheets-only to **every destination type** the product supports:

- **`add_report`** now creates reports for **Looker Studio** and the **email family** (Email, Slack, Microsoft Teams, Google Chat) in addition to Google Sheets.
- **`update_report`** can now edit the **message** (subject and/or body) of an existing email-family report.
- `add_report` responses include the `destination_type`, and the Looker Studio path returns `instructions` + a `setup_guide_url` the assistant relays to the user (the same guide link was added to the `add_destination` response for `looker_studio`).

Closes the main gap left by the report-tools track (#1376, #1377, #1385, #1382, #1383).

## What changed

Four commits, reviewable in order:

1. **Looker Studio support in `add_report`** — synthesizes the default config (`cacheLifetime: 300`, the web-form default); no external side effect, so `CreateReportService`'s transactional validation suffices (no pre-flight needed). The response guides the user through connecting Looker Studio, since dashboard data only flows after they paste the destination's JSON Config themselves (it contains a secret, so it is never sent through the assistant).
2. **Email-family support in `add_report`** — new `message { subject?, body }` input group: `body` is the CUSTOM_MESSAGE template (supports `{{table}}`), `subject` defaults to the report name, send condition gets the product default (`ALWAYS`). Recipients/channels stay on the destination — they were never part of the report config.
3. **Message editing in `update_report`** — the same `message` group as a PATCH: subject-only changes keep the stored template source (including legacy-shaped configs) untouched; a new body rebuilds a clean CUSTOM_MESSAGE config. Everything else (destination, filters, sorting, owners, send condition) is preserved, as before.
4. **Review hardening** (see below).

## Design notes

- **Type branching lives in the facade**: the destination's actual type is resolved first; `message` is rejected up front for any non-email type (one guard, so future types reject rather than silently drop it), and unsupported types get a clear error naming the supported set.
- **Google Sheets keeps its pre-flight** (validate → mint sheet → create report): it is the only path with a non-transactional external side effect. Looker/email paths rely on `CreateReportService`'s own transactional checks — same authorization, no orphan risk.
- **`name` is per-type** (review outcome): optional at the tool layer; required by the facade for Google Sheets and email family (it is the sheet title / default subject); **rejected for Looker Studio** — Looker reports carry no name (the domain clears the title on insert), so accepting one would make the agent confirm a name that never persists.
- **One Looker report per data mart + destination** (review outcome): the domain gives Looker reports a deterministic id per pair by design; the facade now pre-checks and returns a clean business error ("already exists — use the existing report or delete it first") instead of letting the insert die on the primary key with a raw DB message. Mirrors the web UI, which prevents duplicates with an existence-driven toggle.
- **Single email-config construction site**: `buildCustomMessageConfig` is shared by the add and update paths so their shapes cannot drift.
- Insight-template message sources and the `slices`/`filters` input remain deferred (insight templates would need insight MCP tools first; the PRD filter vocabulary doesn't map 1:1 onto the domain filter model).

## Code review

A high-effort multi-angle review ran on this branch before the PR. Two confirmed Looker-path bugs were fixed in the hardening commit (duplicate-pair raw PK error; silently discarded `name`); the removed-behavior and line-by-line angles came back clean (GS pre-flight ordering, merge preservation, legacy email-config migration all verified intact). Deferred non-blocking findings, tracked for follow-up: sharing the Looker default-cache constant with the web form via the schema package, deriving guide URLs from the configurable docs base, collapsing the extra destination fetch on the GS path, and (longer-term) moving per-type config synthesis behind a domain-side factory next to the existing per-type strategy resolvers.

## Testing

- Facade: per-type branching matrix (GS / Looker / each email type), config synthesis against the real zod schemas, `name` per-type rules, Looker duplicate guard, message merge semantics incl. legacy-shaped stored configs, `['*']` projection, error propagation.
- Tools: input schemas (strict, per-field rules, optional `name`), output schemas incl. per-type optional fields, annotations/scopes, registry.
- `npx jest src/data-marts/facades src/ee/mcp` — **371 tests, 39 suites green**; `nest build`, eslint, markdownlint clean.
- Docs (`docs/getting-started/setup-guide/mcp.md`) updated: per-destination behavior, `name`/`message` semantics, Looker one-report-per-pair rule.


amarchenk0 (approved):

romandubovyi (commented):
Review complete. No blocking issues found; a few non-blocking observations are included inline.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -162,7 +194,243 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+   * destination pair (the report id is deterministic), so a supplied name is
+   * rejected and duplicates are caught here with a clean error instead of a
+   * raw primary-key violation.
+   */
+  private async addLookerStudioReport(request: McpAddReportRequest): Promise<McpAddReportResult> {
+    if (request.name !== undefined) {
+      throw new BadRequestException(
+        'The name parameter is not applicable to Looker Studio destinations: ' +
+          'Looker Studio reports carry no name, and each data mart + destination ' +
+          'pair has exactly one report.'
```
romandubovyi, line 357: 🟡 **Important:** Loading all reports for the data mart to check for an existing Looker Studio report is inefficient when the list is large and still leaves a race window. Consider adding a dedicated existence query (by `dataMartId` + `destinationId`) or translating the domain's unique-constraint error instead.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -162,7 +194,243 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+        'A Looker Studio report already exists for this data mart and destination — ' +
+          'each pair has exactly one report. Use the existing report, or delete it first.'
+      );
+    }
+
+    // The title is discarded by the domain for Looker Studio reports; pass the
+    // empty string it would end up as anyway.
+    return this.createReportWithConfig(
+      request,
+      DataDestinationType.LOOKER_STUDIO,
```
romandubovyi: 🔵 **Question:** Passing `title: ''` relies on `CreateReportService` accepting an empty title for Looker Studio reports. Has this been verified against the service's validation? If the domain ever enforces non-empty titles, this path will break.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -162,7 +194,243 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+      // Subject-only change: keep the rest of the stored config exactly as-is
+      // (whatever template source — or legacy shape — it has).
+      return { ...current, subject };
+    }
+
+    // A new body always means a CUSTOM_MESSAGE source (replacing an insight
+    // template if one was set). Rebuild the config from scratch: stored
+    // configs may still be in the legacy messageTemplate shape, and carrying
+    // legacy remnants forward alongside templateSource would be invalid.
+    return this.buildCustomMessageConfig(subject || current.subject, body, current.reportCondition);
```
romandubovyi, line 250: 🟡 **Important:** The `!isEmailConfig(current)` guard may reject subject-only updates for legacy-shaped email configs, despite the comment saying such configs should be preserved. Please verify that `isEmailConfig` returns `true` for all supported legacy shapes, or loosen the guard for subject-only changes.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Fixed in b6fda3c78 — both halves of your suggestion: the guard now uses a dedicated existence query (`ReportService.existsByDataMartIdAndDestinationIdAndProjectId`, a targeted `exists` on the pair) instead of loading the full report list, and the race window is closed by translating the constraint failure: when the INSERT throws `QueryFailedError`, the facade re-checks existence and converts a confirmed duplicate into the same clean business error — genuine DB failures keep their original error (covered by two new specs: lost-race translation and non-duplicate passthrough).

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Verified: on this path `title` is never validated — `CreateReportService` passes it straight through (`title: command.title`, create-report.service.ts:119; the class-validator rules live only on the REST DTO layer, which MCP does not traverse), and `Report.generateId` (@BeforeInsert) overwrites the title to an empty string for LOOKER_STUDIO regardless of the caller. If the domain ever enforces non-empty titles, it would conflict with that entity hook for every Looker path (web included), so it would surface immediately — and our spec pinning `title: ""` in the command would catch it on this path first.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Verified safe on two independent levels: (1) `isEmailConfig` only checks `definition.type === EmailConfigType` (data-destination-config.guards.ts:18-20), and `LegacyEmailConfigSchema` declares the very same `type: z.literal(EmailConfigType)` (email-config.schema.ts:13) — so even a raw legacy shape passes the guard; (2) in practice `current` can never be legacy-shaped anyway, because the Report entity loads `destinationConfig` through the zod transformer (`report.entity.ts:61`) whose parse runs `EmailConfigWithMigrationSchema` and migrates legacy → templateSource shape on read. Subject-only updates on legacy-origin configs are covered by the existing specs around the legacy `messageTemplate` fixtures.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -145,7 +184,7 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
         request.roles,
         request.name ?? current.title,
```
prokopetsv, line 198: 🟡 **Important** — `update_report` accepts and persists a `name` for Looker Studio reports, breaking the invariant the add path enforces.

`addLookerStudioReport` rejects `name` because "Looker Studio reports carry no name" — the `@BeforeInsert` hook in `report.entity.ts` clears the title. But that hook only fires on insert. Here, `request.name ?? current.title` flows into `UpdateReportCommand` with no destination-type check, and `UpdateReportService.run` assigns `report.title = command.title` unconditionally. So `update_report` with a `name` on a Looker report persists a title the product otherwise guarantees is empty, and the agent confirms a rename the UI never shows — the exact "silently discarded name" bug class the hardening commit fixed on the add path, resurfacing one tool over.

The destination type is already in hand (`current.dataDestinationAccess.type`), so the guard is one condition plus a test, mirroring the `message` rejection right next to it.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -162,7 +201,248 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+  private async addLookerStudioReport(request: McpAddReportRequest): Promise<McpAddReportResult> {
+    if (request.name !== undefined) {
+      throw new BadRequestException(
+        'The name parameter is not applicable to Looker Studio destinations: ' +
+          'Looker Studio reports carry no name, and each data mart + destination ' +
+          'pair has exactly one report.'
+      );
+    }
+
+    if (await this.lookerStudioReportExists(request)) {
```
prokopetsv, line 361: 🔵 **Question** — the existence pre-check (and the name/message guards) run before `CreateReportService` performs its `USE` access checks on the data mart and destination. A caller holding only `mcp:write` but lacking `USE` on those entities can learn the destination's type and whether a report exists for a pair from the error they get. Everything is project-scoped (`getByIdAndProjectId`), so exposure is limited to same-project metadata that read tools mostly surface anyway — but the Google Sheets path deliberately kept authorization first (`assertCanCreateReport` before the sheet mint), so the asymmetry is worth a conscious decision, even if the answer is "acceptable."

docs/getting-started/setup-guide/mcp.md
```diff
@@ -438,34 +438,42 @@ Removes a single schedule identified by `trigger_id`. This is destructive and ca
...
+- **Looker Studio**: the report is created with default settings (data cache lifetime of 5 minutes) — the tool accepts no Looker-Studio-specific parameters, and no `name` (Looker Studio reports carry no name). Each data mart + destination pair can have exactly one Looker Studio report; creating a second one returns an error instead of a duplicate. Creating the report only makes the data mart available to the destination: data appears in a dashboard after you connect Looker Studio to OWOX with the destination's JSON Config (it contains a secret key, so it is never sent through the assistant). The response includes `instructions` and a `setup_guide_url` so the assistant can walk you through it; the full guide is [Data Studio destination](../../destinations/supported-destinations/data-studio.md). If the project has no Looker Studio destination yet, the assistant can create one with `add_destination` first.
```
prokopetsv, line 444: 💡 **Suggestion** — the house docs style (`.claude/CLAUDE.md`) requires max 20 words per sentence and no passive voice in `docs/`. This bullet contains 40+ word sentences and passives ("the report is created", "a new Google Sheet is created automatically" above). The surrounding page already reads this way, so this isn't a regression — but since the page is being substantially rewritten anyway, this is the natural moment to bring the touched sections up to the standard rather than growing the debt.

prokopetsv (commented):
## Review

The overall shape is good: branching in the facade (not the tool layer), dual-layer validation (zod schema + facade guards, correctly treating the facade as a public interface), a single construction site for the email config, and the Looker duplicate race handled with a re-check rather than string-matching the DB error. Test coverage of the new branching matrix is thorough, including legacy-shaped email configs and the lost-race path.

Findings are inline: one 🟡 Important (Looker rename gap in `update_report` — worth fixing before merge since it undoes an invariant this same PR introduces), one 🔵 Question (authorization ordering on the Looker add path), one 💡 Suggestion (docs style).

### Verified clean

- **Legacy email-config handling**: subject-only updates spread the stored config (`{ ...current, subject }`), which keeps the legacy `messageTemplate` shape intact; `EmailConfigInputSchema` is a union that still accepts that shape, so `UpdateReportService`'s `checkAccess` validation passes. A new body correctly rebuilds a clean `CUSTOM_MESSAGE` config instead of carrying legacy remnants forward. Both paths are tested.
- **Duplicate-race translation**: the `QueryFailedError` + re-check pattern keeps genuine DB failures on their original error, and the deterministic UUID v5 id in the entity hook confirms the losing insert really does fail on the primary key.
- **Merge preservation in `updateReport`**: destination, filters, sorting, owners, and send condition all flow through from the loaded report; send condition is preserved in both the subject-only and body-rebuild paths.
- **Architecture**: the facade orchestrates use cases and entity services consistently with the existing MCP facade pattern; no repositories leak upward, entities stay below the facade boundary, and the new `ReportService.existsBy…` method sits at the right layer.
- **Changeset** present with `'owox': minor` and the task-number prefix, per project convention.

The deferred items the PR description already tracks (shared cache-lifetime constant, configurable docs base URL, collapsing the extra destination fetch on the Google Sheets path) are reasonable to defer.

docs/getting-started/setup-guide/mcp.md
prokopetsv: UPD: Will be addressed in the following documentation changes.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Great catch — exactly the "silently misleading" class we fixed on the add path, and you are right that @BeforeInsert cannot save the update path. Fixed in 63ab38002: the facade now rejects a supplied `name` when the loaded report's destination is Looker Studio (same wording family as the add-path guard), while column-selection updates stay allowed. Covered by two new specs: rename rejected without touching UpdateReportService, and a fields-only update on a Looker report still passing `title: ""` through.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Conscious decision, now recorded in the code (63ab38002 adds it to the addReport dispatcher comment): the cheap input-shape guards and the Looker existence check reveal only same-project metadata — the destination's type and pair-existence — which the mcp:read surface (`list_destinations`, `get_data_mart_reports`) already exposes to the same caller, and in practice every `mcp:write` token also carries `mcp:read`. Duplicating the USE pre-flight on the side-effect-free paths would buy no confidentiality and re-introduce the drift-prone duplication we keep only where a side effect demands it (the Google Sheets mint). So: acceptable, and deliberate.

docs/getting-started/setup-guide/mcp.md
laskevych: Acknowledged — and noted your UPD that the touched sections will be brought up to the docs standard in the follow-up documentation changes. Happy to apply the 20-words/active-voice pass there. (Small heads-up: `.claude/CLAUDE.md` does not exist in the repo at the moment, so the standard currently lives only in review lore — worth committing it so tooling and future sessions pick it up.)

PR #1431, opened 2026-07-20
fix(web): keep storage config sheet open when dismissing an error toast

prokopetsv:
# Problem

When editing a Storage in the "Configure Storage Provider" sheet and clicking Save, a permission error surfaced as a persistent toast. Clicking the toast's ✕ to dismiss it unexpectedly also triggered the "You have unsaved changes. Exit without saving?" prompt. The root cause: the sheet is a Radix dialog whose default behavior closes on any pointer-down outside its DOM subtree, and the error toast (react-hot-toast) renders in its own top-level portal outside that subtree. So dismissing the toast registered as an outside click, fired the sheet's close handler, and hit the unsaved-changes guard.

# Solution

Handle the toast interaction at the sheet's call site, matching the existing pattern in `DataDestinationConfigSheet` (which does the same for its folder picker):

- Added an `onInteractOutside` handler to the storage config sheet's `SheetContent` that calls `event.preventDefault()` when the interaction originates inside the toast portal, suppressing the close for that one case while leaving all other outside clicks, Escape, and the ✕ button flowing through the normal close + unsaved-changes path.
- Verified against Radix's `DismissableLayer` source that `onInteractOutside` fires for both pointer and focus paths and that `preventDefault()` blocks dismissal in both, so a single handler is sufficient (no need for separate pointer/focus handlers).
- Kept the shared `packages/ui` Sheet untouched — the toast-specific logic lives in the app layer, not the design system.
- Extracted the toast-container class and a reusable `isEventTargetInsideToaster` predicate into a dedicated module so the class string is a single source of truth (shared by the `<Toaster>` and the sheet) and the logic is unit-testable.

# Changes

- Added `onInteractOutside` handler on the storage config sheet to ignore clicks originating inside the toast portal.
- Added `apps/web/src/shared/components/Toaster/toaster-container.ts` exporting `TOASTER_CONTAINER_CLASS` and `isEventTargetInsideToaster()`.
- Updated `Toaster` to apply the shared `containerClassName` constant instead of an inline string.
- Added unit tests for `isEventTargetInsideToaster` covering targets inside, on, and outside the container, plus null/non-Element targets.


apps/web/src/features/data-storage/edit/components/DataStorageEditSheet/DataStorageConfigSheet.tsx
```diff
@@ -85,7 +86,17 @@ export function DataStorageConfigSheet({
         }
       }}
     >
-      <SheetContent data-testid='storageConfigSheet'>
+      <SheetContent
+        data-testid='storageConfigSheet'
+        onInteractOutside={event => {
```
romandubovyi: 🟡 **Important:** This guard fixes one sheet, but the bug is global: `apiClient.ts:114` shows a `persistent: true` toast for *every* 403, and `useUnsavedGuard` is used by 9 sheets (destination, connectors, email/gsheets/looker reports, scheduled triggers, insights). `EmailReportEditSheet.tsx:72-78` even raises its own `duration: Infinity` toast, so it reproduces today. Consider applying the guard once instead of per call site — e.g. a default `onInteractOutside` in the shared `SheetContent` keyed off a `data-*` attribute on the toast container (so `packages/ui` stays unaware of react-hot-toast), composed with any caller-supplied handler. Otherwise this handler has to be copy-pasted into 8 more sheets, and every new sheet reintroduces the bug.

apps/web/src/shared/components/Toaster/toaster-container.ts
```diff
@@ -0,0 +1,15 @@
...
+ * must not treat a toast dismiss as an outside click) match against this.
+ */
+export const TOASTER_CONTAINER_CLASS = 'app-toaster';
+
+/**
+ * True when a DOM event target sits inside the toast portal — e.g. a click on a
+ * toast's dismiss button. Sheets/dialogs use this to avoid treating a toast
+ * interaction as an outside click that would close them.
+ */
+export function isEventTargetInsideToaster(target: EventTarget | null): boolean {
```
romandubovyi: 🟡 **Important:** `MainLayout.tsx:111-113` mounts two toasters — sonner (`closeButton` enabled) and react-hot-toast. This predicate only matches the react-hot-toast container, so clicking sonner's ✕ still registers as an outside click and closes the sheet with the unsaved-changes prompt. Sonner marks its container with `[data-sonner-toaster]`; matching both selectors in the same `closest()` call keeps this a one-line change and covers the whole toast surface.

apps/web/src/features/data-storage/edit/components/DataStorageEditSheet/DataStorageConfigSheet.tsx
```diff
@@ -85,7 +86,17 @@ export function DataStorageConfigSheet({
...
+          // Error toasts render in react-hot-toast's own portal, outside this
+          // sheet's DOM subtree. Dismissing a toast must not count as an outside
+          // click that closes the sheet and triggers the unsaved-changes guard.
+          if (isEventTargetInsideToaster(event.detail.originalEvent.target)) {
```
romandubovyi: 🔵 **Question:** `DataDestinationConfigSheet.tsx:170-184` solves the same mechanic with three handlers (`onPointerDownOutside` + `onFocusOutside` + `onInteractOutside`) where — as this PR correctly establishes — a single `onInteractOutside` suffices. Should the destination sheet be collapsed to the same single-handler shape here? It is still exposed to the toast case anyway (its predicate only covers the picker), so unifying both predicates behind one handler would leave one pattern instead of two divergent ones.

apps/web/src/shared/components/Toaster/toaster-container.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Class applied to the react-hot-toast portal container. Consumers that need to
+ * detect whether a DOM event originated inside a toast (e.g. Radix sheets that
+ * must not treat a toast dismiss as an outside click) match against this.
+ */
+export const TOASTER_CONTAINER_CLASS = 'app-toaster';
```
romandubovyi: 💡 **Suggestion:** This class styles nothing — it is purely a DOM marker, but in a Tailwind codebase a bare `className` reads as styling and is easy to drop during a style cleanup. A `data-*` marker (e.g. `[data-toast-container]`) states the contract more clearly, and it is also what would let `packages/ui` host the shared guard without knowing which toast library the app uses.

apps/web/src/shared/components/Toaster/toaster-container.test.ts
```diff
@@ -0,0 +1,44 @@
+import { afterEach, describe, it, expect } from 'vitest';
+import { isEventTargetInsideToaster, TOASTER_CONTAINER_CLASS } from './toaster-container';
+
+/**
+ * Guards the toast-vs-sheet interaction fix: a click inside the toast portal
+ * must be recognised so sheets don't treat a toast dismiss as an outside click
+ * (which would close the sheet and trigger the unsaved-changes prompt).
+ */
+describe('isEventTargetInsideToaster', () => {
```
romandubovyi: 💡 **Suggestion:** These four cases effectively assert `Element.closest()` and `instanceof Element` — the whole body of the function — so they cannot catch the regression this fix is about. What will actually break is the wiring: someone drops `containerClassName` from `<Toaster>`, or a new sheet forgets the handler. A single sheet-level RTL test (open sheet → make form dirty → click the toast's ✕ → sheet still open, no unsaved prompt) would guard the real behaviour; `MembershipRequestSheet.test.tsx` and `EditApiKeySheet.test.tsx` already set that precedent.

romandubovyi (approved):
Review complete — no blockers, approving. The fix itself is correctly scoped: a single `onInteractOutside` guard that leaves Escape, the sheet ✕ and ordinary outside clicks on the normal close + unsaved-changes path, with `packages/ui` untouched.

Main follow-up concern: the bug is global, not storage-specific. `apiClient.ts:114` raises a `persistent: true` toast for every 403 and `useUnsavedGuard` is used by 9 sheets, so the same defect still reproduces elsewhere (e.g. `EmailReportEditSheet` with its own `duration: Infinity` toast), and sonner — also mounted, with `closeButton` — is not covered by the predicate. Worth deciding whether to hoist the guard to a single shared place rather than repeating it per call site.

amarchenk0 (approved):

apps/web/src/features/data-storage/edit/components/DataStorageEditSheet/DataStorageConfigSheet.tsx
prokopetsv: You're right on both points — verified: 9 sheets use `useUnsavedGuard`, and since `apiClient.ts:114` raises the persistent toast for every 403 globally, all of them had the same bug; my call-site fix only covered one. Moved the guard into `SheetContent` so it applies everywhere, matching on the data attributes the portals already render (`data-rht-toaster`, `data-sonner-toaster`, plus a generic `data-ignore-outside-dismiss` opt-in) — so `packages/ui` needs no toast-library dependency and covers both. The caller's own `onInteractOutside` is still forwarded, so the destination sheet's picker handling is unaffected. `apps/web` is back to its original state apart from the added test.

PR #1432, opened 2026-07-20
fix(backend): report totals cover all selected metrics, not just numeric fields

zapolsky:
## Problem

Report totals — returned by `query_data_mart` and retrievable for a Data Mart run via its `x-owox-run-id` → run details — were computed only for **numeric** fields. A scorecard that aggregates a non-numeric column (e.g. `COUNT_DISTINCT` over a text field) returned `totals: null` from `getRunById` / `query_data_mart` even after the run succeeded and the aggregate was present in the streamed rows.

## Fix

Totals are now driven by field governance instead of a hard numeric-type gate. A field contributes to totals when it is either:

- **numeric** — an auto per-column summary (unchanged), or
- **aggregated by the report as a metric** — the only metric signal, since there is no schema-level dimension/metric role.

Each field is summarized by its governance-allowed functions, so a text metric yields `COUNT` / `COUNT_DISTINCT` rather than a `SUM`/`AVG` it can't run. This is **symmetric across native and joined** Data Mart fields. `ANY_VALUE` and `STRING_AGG` are excluded from totals — neither reduces to a meaningful grand total.

`query_data_mart` (MCP) and the HTTP Data run / `getRunById` path share the same `ReportTotalsService.computeTotals`, so they behave identically.

## Testing

- **Unit:** new coverage for a `COUNT_DISTINCT`-over-text scorecard, `ANY_VALUE`/`STRING_AGG` exclusion, and joined-field symmetry (positive + negative), alongside the existing totals suites.
- **End-to-end on BigQuery:** verified totals for a text `COUNT_DISTINCT`, numeric `SUM/AVG/MIN/MAX` + percentiles, exclusion of `ANY_VALUE`/`STRING_AGG`, and a joined text `COUNT` — against the previous `null` behavior.
- **Docs:** `report-aggregations.md` Totals section updated.

### DIFF
--- a/apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
+++ b/apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
@@ -488,7 +488,10 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
     const blendedField = (
       name: string,
       type: string,
-      postJoinAggregations?: ReportAggregateFunction[]
+      postJoinAggregations?: ReportAggregateFunction[],
+      // The PRE-join roll-up run inside the bottom-up CTE. Must be valid for the field type —
+      // e.g. a STRING joined field cannot roll up with SUM, so pass ANY_VALUE there.
+      preJoinAggregation: ReportAggregateFunction = 'SUM'
     ): BlendedFieldDto => {
       const f = new BlendedFieldDto();
       f.name = name;
@@ -501,7 +504,7 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       f.alias = '';
       f.description = '';
       f.isHidden = false;
-      f.aggregateFunction = 'SUM';
+      f.aggregateFunction = preJoinAggregation;
       f.postJoinAggregations = postJoinAggregations;
       f.transitiveDepth = 1;
       f.aliasPath = 'partner';
@@ -600,14 +603,15 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
     // metric (e.g. COUNT_DISTINCT over a joined text column) must appear in totals too, by its
     // post-join allowed functions — minus ANY_VALUE / STRING_AGG.
     it('includes a JOINED non-numeric field in totals when the report aggregates it', async () => {
+      // A STRING joined field must roll up pre-join with a STRING-valid function (ANY_VALUE),
+      // not SUM — so the generated CTE + outer SQL is executable BigQuery.
       const fields = [
-        blendedField('partner__country', 'STRING', [
-          'MIN',
-          'MAX',
-          'COUNT',
-          'COUNT_DISTINCT',
-          'STRING_AGG',
-        ]),
+        blendedField(
+          'partner__country',
+          'STRING',
+          ['MIN', 'MAX', 'COUNT', 'COUNT_DISTINCT', 'STRING_AGG'],
+          'ANY_VALUE'
+        ),
       ];
       const { service } = makeBlendedTotalsComposer(fields);
       const report = buildTotalsReport({
@@ -625,6 +629,24 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       expect(fns).toEqual(expect.arrayContaining(['COUNT', 'COUNT_DISTINCT', 'MIN', 'MAX']));
       // STRING_AGG (and ANY_VALUE) are excluded from totals on the joined path too.
       expect(fns).not.toContain('STRING_AGG');
+
+      // The generated SQL must be executable: the bottom-up CTE rolls the STRING column up with
+      // ANY_VALUE (never SUM over text), and the ungrouped outer SELECT applies the post-join
+      // totals functions — COUNT / COUNT_DISTINCT / MIN / MAX — over the joined column.
+      const splitAt = result!.sql.lastIndexOf('\n\nSELECT');
+      const cte = result!.sql.slice(0, splitAt);
+      const finalSelect = result!.sql.slice(splitAt);
+      expect(cte).toContain('ANY_VALUE(country) AS partner__country');
+      expect(result!.sql).not.toMatch(/SUM\(/); // no SUM over a text column anywhere
+      expect(finalSelect).toContain('MIN(partner.partner__country)');
+      expect(finalSelect).toContain('MAX(partner.partner__country)');
+      expect(finalSelect).toContain('COUNT(partner.partner__country)');
+      expect(finalSelect).toContain('COUNT(DISTINCT partner.partner__country)');
+      // Excluded functions never reach the totals SELECT.
+      expect(finalSelect).not.toContain('ANY_VALUE');
+      expect(finalSelect).not.toMatch(/STRING_?AGG/i);
+      // Single grand-total row — no outer GROUP BY.
+      expect(finalSelect).not.toMatch(/GROUP BY/);
     });
 
     // A JOINED non-numeric field the report does NOT aggregate stays a plain dimension — not in
@@ -632,7 +654,7 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
     it('excludes a JOINED non-numeric field from totals when the report does not aggregate it', async () => {
       const fields = [
         blendedField('partner__cost', 'FLOAT', ['SUM']),
-        blendedField('partner__country', 'STRING', ['COUNT', 'COUNT_DISTINCT']),
+        blendedField('partner__country', 'STRING', ['COUNT', 'COUNT_DISTINCT'], 'ANY_VALUE'),
       ];
       const { service } = makeBlendedTotalsComposer(fields);
       const report = buildTotalsReport({
--- a/apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
+++ b/apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
@@ -9,7 +9,10 @@ import {
   countPositionalPlaceholders,
 } from '../data-storage-types/athena/services/athena-clause-renderer';
 import { BlendedFieldDto } from '../dto/domain/blendable-schema.dto';
-import { ReportAggregateFunction } from '../dto/schemas/aggregate-function.schema';
+import {
+  AggregateFunction,
+  ReportAggregateFunction,
+} from '../dto/schemas/aggregate-function.schema';
 
 describe('ReportSqlComposerService — aggregations wiring', () => {
   const buildReport = (overrides: Partial<Report> = {}): Report =>
@@ -369,16 +372,9 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       expect(blendedReportDataService.resolveBlendingDecision).not.toHaveBeenCalled();
     });
 
-    // REPRODUCTION — Vlad's dashboard scorecard (chat thread YjeZ1yvBxSg / repro-totals.mjs):
-    // a COUNT_DISTINCT metric over a STRING field, NO grouping. Per WI #6680 §D totals are "по
-    // обраних метриках" (over the SELECTED metrics), and §C: "Unique-by-PK — звичайна метрика з
-    // COUNT_DISTINCT". Governance already permits COUNT/COUNT_DISTINCT on STRING fields
-    // (field-aggregation-governance.ts: string → ['COUNT','COUNT_DISTINCT']). But
-    // resolveTotalsAllowedForColumn hard-gates on `categorizeFieldType(type) !== 'number'` and
-    // returns [] BEFORE consulting governance — so the COUNT_DISTINCT metric is dropped,
-    // composeTotals returns null, and getRunById.totals is null even though the run SUCCEEDED and
-    // the aggregate streamed ({"country | COUNTUNIQUE":18}). Totals must follow governance, not a
-    // numeric-type gate. RED today.
+    // Requirement (WI #6680 §D): a COUNT_DISTINCT metric over a STRING field with no grouping
+    // must appear in totals — totals follow the field's governance-allowed functions, not a
+    // numeric-type gate (governance permits COUNT/COUNT_DISTINCT on STRING).
     it('scorecard: COUNT_DISTINCT over a STRING metric (NO grouping) is included in totals (WI #6680 §D)', async () => {
       const { service } = makeBqTotalsComposer(['country']);
       const report = buildTotalsReport(
@@ -490,8 +486,9 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       type: string,
       postJoinAggregations?: ReportAggregateFunction[],
       // The PRE-join roll-up run inside the bottom-up CTE. Must be valid for the field type —
-      // e.g. a STRING joined field cannot roll up with SUM, so pass ANY_VALUE there.
-      preJoinAggregation: ReportAggregateFunction = 'SUM'
+      // e.g. a STRING joined field cannot roll up with SUM, so pass ANY_VALUE there. Pre-join
+      // aggregates are the base function set (no percentiles).
+      preJoinAggregation: AggregateFunction = 'SUM'
     ): BlendedFieldDto => {
       const f = new BlendedFieldDto();
       f.name = name;
@@ -666,6 +663,39 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       expect(result).not.toBeNull();
       expect(result!.columns).toContain('partner__cost');
       expect(result!.columns).not.toContain('partner__country');
+    });
+
+    // A stale post-join override can name functions the current type can't run (e.g. SUM saved
+    // before a field became STRING). collectBlendedAllowedSets clamps the override to the type's
+    // supported set — as the native path does — so the totals plan never carries SUM(string),
+    // which would make the validator throw and silently null the whole totals block.
+    it('clamps a stale numeric override on a joined STRING field to type-supported functions', async () => {
+      const fields = [
+        blendedField(
+          'partner__country',
+          'STRING',
+          ['SUM', 'AVG', 'COUNT', 'COUNT_DISTINCT'],
+          'ANY_VALUE'
+        ),
+      ];
+      const { service } = makeBlendedTotalsComposer(fields);
+      const report = buildTotalsReport({
+        columnConfig: ['partner__country'],
+        aggregationConfig: [{ column: 'partner__country', function: 'COUNT_DISTINCT' }],
+      } as Partial<Report>);
+
+      const result = await service.composeTotals(report, {} as never);
+
+      expect(result).not.toBeNull();
+      const fns = result!.aggregations
+        .filter(a => a.column === 'partner__country')
+        .map(a => a.function);
+      // SUM / AVG are not supported for STRING → clamped away; the valid ones survive.
+      expect(fns).not.toContain('SUM');
+      expect(fns).not.toContain('AVG');
+      expect(fns).toEqual(expect.arrayContaining(['COUNT', 'COUNT_DISTINCT']));
+      expect(result!.sql).not.toMatch(/SUM\(/);
+      expect(result!.sql).not.toMatch(/AVG\(/);
     });
 
     it('resolves the blendable schema once and reuses it for the blended decision (no recompute)', async () => {
--- a/apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
+++ b/apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
@@ -369,6 +369,71 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       expect(blendedReportDataService.resolveBlendingDecision).not.toHaveBeenCalled();
     });
 
+    // REPRODUCTION — Vlad's dashboard scorecard (chat thread YjeZ1yvBxSg / repro-totals.mjs):
+    // a COUNT_DISTINCT metric over a STRING field, NO grouping. Per WI #6680 §D totals are "по
+    // обраних метриках" (over the SELECTED metrics), and §C: "Unique-by-PK — звичайна метрика з
+    // COUNT_DISTINCT". Governance already permits COUNT/COUNT_DISTINCT on STRING fields
+    // (field-aggregation-governance.ts: string → ['COUNT','COUNT_DISTINCT']). But
+    // resolveTotalsAllowedForColumn hard-gates on `categorizeFieldType(type) !== 'number'` and
+    // returns [] BEFORE consulting governance — so the COUNT_DISTINCT metric is dropped,
+    // composeTotals returns null, and getRunById.totals is null even though the run SUCCEEDED and
+    // the aggregate streamed ({"country | COUNTUNIQUE":18}). Totals must follow governance, not a
+    // numeric-type gate. RED today.
+    it('scorecard: COUNT_DISTINCT over a STRING metric (NO grouping) is included in totals (WI #6680 §D)', async () => {
+      const { service } = makeBqTotalsComposer(['country']);
+      const report = buildTotalsReport(
+        {
+          columnConfig: ['country'],
+          aggregationConfig: [{ column: 'country', function: 'COUNT_DISTINCT' }],
+        } as Partial<Report>,
+        // No schema-level metric/dimension role in practice — STRING defaults to 'dimension'.
+        // The ONLY signal that `country` is a metric here is that the report aggregates it.
+        [field('country', 'STRING')]
+      );
+
+      const result = await service.composeTotals(report, {} as never);
+
+      expect(result).not.toBeNull();
+      expect(result!.columns).toContain('country');
+      expect(result!.aggregations).toEqual(
+        expect.arrayContaining([{ column: 'country', function: 'COUNT_DISTINCT' }])
+      );
+    });
+
+    // ANY_VALUE (an arbitrary row's value) and STRING_AGG (whole-column concat) are meaningless
+    // as a single grand-total number, so they are excluded from the totals summary for EVERY
+    // field even when the field's allowed set (per-field override or type-default) permits them.
+    it('excludes ANY_VALUE and STRING_AGG from totals even when the field allows them', async () => {
+      const { service } = makeBqTotalsComposer(['country']);
+      const report = buildTotalsReport(
+        {
+          columnConfig: ['country'],
+          aggregationConfig: [{ column: 'country', function: 'COUNT_DISTINCT' }],
+        } as Partial<Report>,
+        [
+          field('country', 'STRING', {
+            allowedAggregations: [
+              'MIN',
+              'MAX',
+              'ANY_VALUE',
+              'COUNT',
+              'COUNT_DISTINCT',
+              'STRING_AGG',
+            ],
+          }),
+        ]
+      );
+
+      const result = await service.composeTotals(report, {} as never);
+
+      expect(result).not.toBeNull();
+      const fns = result!.aggregations.map(a => a.function);
+      expect(fns).not.toContain('ANY_VALUE');
+      expect(fns).not.toContain('STRING_AGG');
+      // The meaningful ones survive.
+      expect(fns).toEqual(expect.arrayContaining(['COUNT', 'COUNT_DISTINCT', 'MIN', 'MAX']));
+    });
+
     it('strips HAVING (function-carrying) filters from the totals query, keeping WHERE filters', async () => {
       const { service } = makeBqTotalsComposer(['revenue']);
       const report = buildTotalsReport({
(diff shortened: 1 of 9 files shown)


apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
```diff
@@ -531,6 +596,56 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       expect(finalSelect).not.toMatch(/GROUP BY/);
     });
 
+    // Symmetry with the main-mart rule: a JOINED non-numeric field the report aggregates as a
+    // metric (e.g. COUNT_DISTINCT over a joined text column) must appear in totals too, by its
+    // post-join allowed functions — minus ANY_VALUE / STRING_AGG.
+    it('includes a JOINED non-numeric field in totals when the report aggregates it', async () => {
+      const fields = [
+        blendedField('partner__country', 'STRING', [
```
max-voloshyn: 🟡 Non-blocking: This joined-string test does not validate executable SQL. `blendedField()` hard-codes `aggregateFunction = 'SUM'`, so this case builds the inner roll-up as `SUM(country)` even though `country` is a `STRING`; BigQuery would reject that query. The assertions only inspect the derived function list, allowing invalid SQL to pass. Please parameterize the helper with a valid pre-join aggregation such as `ANY_VALUE` and assert the generated inner and outer SQL.

🤖 Reviewed by Codex (GPT-5.6 Sol High)

docs/getting-started/setup-guide/report-aggregations.md
```diff
@@ -109,9 +109,9 @@ Whenever a report is aggregated, OWOX automatically adds a **`Row Count`** colum
 
 ## Totals
 
-**Totals** are a per-column summary over the full filtered dataset, with no grouping. Every selected **numeric** field is aggregated by **all of its allowed functions** — for example `Sum`, `Average`, `Min`, and `Max` of `revenue`. Totals are computed **in the warehouse** by a separate query and returned as a **separate block**, so they stay accurate and are never recomputed from the displayed rows.
+**Totals** are a per-column summary over the full filtered dataset, with no grouping. Totals cover every selected **numeric** field — aggregated by **all of its allowed functions** (for example `Sum`, `Average`, `Min`, and `Max` of `revenue`) — plus any **non-numeric field the report aggregates as a metric** (for example `Count Unique` on a text `country` column, giving its distinct count). `Sample` (`ANY_VALUE`) and `Combined` (`STRING_AGG`) are **never** part of Totals: a single representative value or a full-column concatenation is not a meaningful grand total. Totals are computed **in the warehouse** by a separate query and returned as a **separate block**, so they stay accurate and are never recomputed from the displayed rows.
```
max-voloshyn, line 112: 🔴 Blocking: The public run-details Swagger schema remains stale: `data-mart-run-response-api.dto.ts` still describes totals as numeric-only, and its example contains only numeric totals. Please update that `ApiProperty` description and example to include report-aggregated non-numeric metrics and the `ANY_VALUE`/`STRING_AGG` exclusion so the generated API documentation matches this behavior.

🤖 Reviewed by Codex (GPT-5.6 Sol High)

max-voloshyn (dismissed):

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
...
     for (const blendedField of blendableSchema.blendedFields) {
-      if (blendedField.isHidden || categorizeFieldType(blendedField.type) !== 'number') {
+      if (blendedField.isHidden) {
+        continue;
+      }
+      const isNumeric = categorizeFieldType(blendedField.type) === 'number';
+      if (!isNumeric && !aggregatedColumns.has(blendedField.name)) {
         continue;
       }
       const allowed =
```
laskevych: 🟡 Important: `postJoinAggregations` is taken verbatim here, with no intersection against the effective type's supported set — unlike the native path, where `resolveFieldGovernance` clamps overrides via `intersectWithSupported`. A stale override can legally contain `SUM`/`AVG` while the field's effective type is STRING (the override is saved once, but the effective type follows the dedup function — `blendable-schema.service.ts:283-287` keeps overrides verbatim, and `BlendedFieldOverrideSchema` only checks enum membership). Before this PR such a field never entered totals; now, once the report aggregates it with any surviving function (e.g. `MIN`), the derived totals plan carries `SUM(string_col)`, the validator's type floor (`output-controls-validator.service.ts:334-346`) throws independently of the allowed set, and **both** run paths swallow the error into `totals: null` — silently dropping even the native numeric totals that worked before.

Suggest clamping here the same way the native path does: `resolveFieldGovernance(blendedField.type, { allowedAggregations: blendedField.postJoinAggregations }).allowedAggregations`.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
...
   ): ReportAggregateFunction[] {
     const descriptor = mainByName.get(name);
+    let allowed: ReportAggregateFunction[];
     if (descriptor) {
-      if (categorizeFieldType(descriptor.type) !== 'number') {
+      // Numerics get an auto summary (type-default 'metric'); non-numerics only count as
+      // metrics when the report itself aggregates them — the sole metric signal without a
+      // schema-level role. Either way governance decides which functions are valid for the
+      // type, so a STRING metric yields COUNT/COUNT_DISTINCT rather than SUM/AVG it can't run.
+      const isNumeric = categorizeFieldType(descriptor.type) === 'number';
```
laskevych: 🟡 Important: The comments here (and at lines 254, 276, 354) say "no schema-level dimension/metric role exists today", but `aggregationRole: 'dimension' | 'metric'` is a persisted schema field (`data-mart-schema.utils.ts:101-104`), accepted by the schema-update API and preserved by all three schema mergers — and this very call site passes it into `resolveFieldGovernance` at line 336, then discards the resolved `.role` in favor of `categorizeFieldType(...) === 'number'`.

Since the default role is `'metric'` exactly for the number category, gating on `governance.role === 'metric' || aggregatedColumns.has(name)` would be byte-identical for every field without an explicit role, while fixing both inversions: a STRING field explicitly marked `metric` currently gets no auto totals, and a numeric ID marked `dimension` still gets meaningless `SUM/AVG` grand totals. (The blended path would additionally need `aggregationRole` plumbed into `BlendedFieldDto`, which today only carries `postJoinAggregations` — fine to leave as a TODO, but the comments should stop denying the field exists.)

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
...
+    blendableSchema: BlendableSchemaDto,
+    aggregatedColumns: ReadonlySet<string>
   ): Map<string, ReportAggregateFunction[]> {
     const result = new Map<string, ReportAggregateFunction[]>();
     for (const blendedField of blendableSchema.blendedFields) {
-      if (blendedField.isHidden || categorizeFieldType(blendedField.type) !== 'number') {
+      if (blendedField.isHidden) {
+        continue;
+      }
+      const isNumeric = categorizeFieldType(blendedField.type) === 'number';
```
laskevych: 🟡 Important: For joined text fields the default dedup is `STRING_AGG` (`blendable-schema.service.ts`, `getDefaultAggregateFunction`), so the per-join CTE emits concatenated lists and the outer totals then compute `COUNT(DISTINCT ...)` over those concatenations — `partner__country | COUNTUNIQUE` counts distinct `"DE, FR"` strings, not distinct countries. The changeset presents native and joined fields as equivalent ("across both native and joined Data Mart fields") and the doc example promises "its distinct count", but a joined text metric's grand total is dedup-function-dependent.

The existing comment already documents the analogous unweighted-AVG limitation for numerics — this deserves the same explicit "accepted limitation" note (comment + docs), or exclusion of dedup-rolled-up text fields from totals until the semantics are right.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -244,15 +244,21 @@ export class ReportSqlComposerService {
...
+   *     when the report itself does not aggregate it; or
+   *   - explicitly AGGREGATED by the report (its name appears in `aggregationConfig`) — this
+   *     is the only signal that a NON-numeric field (e.g. a STRING with COUNT_DISTINCT) is a
+   *     metric, because there is no schema-level dimension/metric designation today.
+   * Either way the function set comes from the field's governance (per-field allowed set else
+   * the type-default), so a STRING metric contributes COUNT/COUNT_DISTINCT and a numeric one
+   * SUM/AVG/MIN/MAX — never a function the type cannot run. Non-numeric fields the report does
+   * NOT aggregate (plain dimensions) and unresolved columns are skipped. Selected JOINED
+   * (blended) numeric fields are included too, governed by their DM-level `postJoinAggregations`
+   * (else the type-default); any blended numeric column drives `compose` onto the blended path,
```
laskevych: 🟡 Important: The private helper's doc was updated, but every outward-facing contract still states the old numeric-only rule: `composeTotals`'s own JSDoc (lines 155-171 — "For every NUMERIC field…", "Returns null … when no selected numeric field…"), the inline comment at line 205 ("Totals are a numeric-field summary only"), `ReportTotalsService`'s class doc (`report-totals.service.ts:14-24` and the comment at 54), the published Swagger description of `totals` (`data-mart-run-response-api.dto.ts:157` — "every selected numeric field aggregated by its allowed functions"), and both web-side comments (`data-mart-run.ts:30`, `data-mart-run.response.dto.ts:33`). The method name `deriveNumericTotalsAggregations` and this line's claim ("any blended **numeric** column drives `compose` onto the blended path") understate the new behavior too.

Since totals values can now legitimately be strings (`MIN`/`MAX` of a text metric), a consumer coding against the documented contract would mis-handle them — these should all be updated in this PR, as they describe exactly the invariant it removes.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -265,6 +271,10 @@ export class ReportSqlComposerService {
   }> {
     const descriptors = collectSchemaFieldPathDescriptors(report.dataMart.schema?.fields ?? []);
     const byName = new Map(descriptors.map(d => [d.name, d]));
+    // The fields the report explicitly aggregates — the only marker that a non-numeric column
+    // is a metric (WI #6680 §D: totals are over the SELECTED metrics; §C: Unique-by-PK is a
+    // normal COUNT_DISTINCT metric). No schema-level dimension/metric role exists yet.
+    const aggregatedColumns = new Set((report.aggregationConfig ?? []).map(rule => rule.column));
```
laskevych, line 281: 🔵 Question: The comment above this line cites WI #6680 §C ("Unique-by-PK is a normal COUNT_DISTINCT metric") as justification, yet a report whose only metric is the built-in **Unique count** toggle (`uniqueCountConfig`, no `aggregationConfig` rules) still gets `totals: null` — the toggle is not treated as a metric signal. So the semantically identical COUNT_DISTINCT yields totals when written as a manual rule but null when expressed via the toggle.

The docs do say "Unique count is not part of Totals", so this may be intended — but the changeset claims totals "now summarize every metric a report aggregates". Is this asymmetry deliberate? If yes, worth an explicit note; if not, `uniqueCountConfig` should feed eligibility too.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
...
     const descriptor = mainByName.get(name);
+    let allowed: ReportAggregateFunction[];
     if (descriptor) {
-      if (categorizeFieldType(descriptor.type) !== 'number') {
+      // Numerics get an auto summary (type-default 'metric'); non-numerics only count as
+      // metrics when the report itself aggregates them — the sole metric signal without a
+      // schema-level role. Either way governance decides which functions are valid for the
+      // type, so a STRING metric yields COUNT/COUNT_DISTINCT rather than SUM/AVG it can't run.
+      const isNumeric = categorizeFieldType(descriptor.type) === 'number';
+      if (!isNumeric && !aggregatedColumns.has(name)) {
```
laskevych: 🔵 Question: Eligibility counts *any* `aggregationConfig` rule as metric evidence, including rules whose only function is `ANY_VALUE` or `STRING_AGG` — the two functions this same method then strips as "not a meaningful grand total". A report that uses `Sample` on a text column purely as a per-group passthrough (reachable with a per-field `allowedAggregations` override that also permits COUNT/COUNT_DISTINCT) flips `composeTotals` from null to a totals block computing `COUNT` + exact `COUNT_DISTINCT` nobody asked for.

Should eligibility require at least one summarizable (non-stripped) function among the field's rules? Same applies to the blended check at line 369.

laskevych (commented):
Deep review of the totals-eligibility change. Posted inline: 4 🟡 Important, 5 🔵 Question, 3 💡 Suggestion, 1 🟢 Nitpick.

Also checked and **not** confirmed (no action needed): case-mismatched `aggregationConfig` columns on the MCP path (`validateForReport` is a mandatory chokepoint inside `compose()` → `resolveBlendingDecision`); JS `Date` leaking into totals from Snowflake/Databricks `MIN`/`MAX` over dates (it silently coerces to an ISO string — no runtime zod parse of totals exists); blended non-numeric metrics failing to route onto the blended path (routing is name-based, works); aggregating a blended column absent from `columnConfig` (rejected by the validator); and the `'owox': minor` changeset bump (matches the repo's documented release strategy).

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
...
       }).allowedAggregations;
+    } else {
+      // Joined (blended) field: eligibility already applied in collectBlendedAllowedSets
+      // (numeric → auto summary; non-numeric → only when the report aggregates it).
+      allowed = blendedByName.get(name) ?? [];
     }
-    return blendedNumericByName.get(name) ?? [];
+    // ANY_VALUE (an arbitrary row's value) and STRING_AGG (whole-column concat) do not summarize
+    // to a meaningful grand total, so the totals summary omits them for every field/type.
+    return allowed.filter(fn => fn !== 'ANY_VALUE' && fn !== 'STRING_AGG');
```
laskevych: 🔵 Question: This strip also changes behavior for *numeric* fields: one whose per-field override is exactly `['ANY_VALUE']` produced a `"field | ANYVALUE"` totals entry before this PR and now — being the only eligible column — makes `composeTotals` return `null` entirely. The changeset does say Sample/Combined are no longer included, but "totals disappear completely for existing reports with such overrides" is a sharper migration edge than "two functions dropped from the block". Intentional and acceptable?

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
...
+      // Numerics get an auto summary (type-default 'metric'); non-numerics only count as
+      // metrics when the report itself aggregates them — the sole metric signal without a
+      // schema-level role. Either way governance decides which functions are valid for the
+      // type, so a STRING metric yields COUNT/COUNT_DISTINCT rather than SUM/AVG it can't run.
+      const isNumeric = categorizeFieldType(descriptor.type) === 'number';
+      if (!isNumeric && !aggregatedColumns.has(name)) {
         return [];
       }
-      return resolveFieldGovernance(descriptor.type, {
+      allowed = resolveFieldGovernance(descriptor.type, {
```
laskevych, line 349: 🔵 Question: A text metric aggregated with plain `COUNT` gets the full governed set in totals, which with the string default (`['COUNT','COUNT_DISTINCT']`) silently adds an **exact** `COUNT(DISTINCT col)` over the full filtered dataset on every run — a cost class (shuffle-heavy on BigQuery, OOM-prone on Athena) that pre-PR totals never had, since numeric all-functions summaries are all single-pass. The requested functions are right there in `aggregationConfig` (this method's `aggregatedColumns` is built from it, discarding the function).

Was "all allowed functions" (vs. intersecting with the requested ones) a deliberate choice for non-numeric metrics? The docs do document it — flagging the warehouse-cost consequence in case it wasn't weighed.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
   private resolveTotalsAllowedForColumn(
     name: string,
     mainByName: ReadonlyMap<string, SchemaFieldDescriptor>,
-    blendedNumericByName: ReadonlyMap<string, ReportAggregateFunction[]>
+    blendedByName: ReadonlyMap<string, ReportAggregateFunction[]>,
+    aggregatedColumns: ReadonlySet<string>
   ): ReportAggregateFunction[] {
     const descriptor = mainByName.get(name);
```
laskevych, line 341: 🔵 Question: `resolveTotalsAllowedForColumn` resolves names main-schema-first, while the validator, the blending decision, and the blended SQL all resolve blended-first (blended entries overwrite native in `homeFieldTypes`/`buildAggregationGovernance`). A native column literally named `<joinAlias>__<field>` (e.g. `partner__country` — exactly the naming convention of materialized blended-report output, so round-tripping such a table into a mart with the same join recreates it) therefore gets totals derived from the wrong field: depending on the type orientation this yields a silently-nulled totals block or totals computed over the joined value with the native field's function set.

This divergence predates the PR, but since this function was rewritten here: should totals adopt the same blended-wins precedence, or should native-vs-blended unified-name collisions be rejected the way blended-vs-blended ones already are (`buildBlendedFieldIndex`)?

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
...
+    aggregatedColumns: ReadonlySet<string>
   ): Map<string, ReportAggregateFunction[]> {
     const result = new Map<string, ReportAggregateFunction[]>();
     for (const blendedField of blendableSchema.blendedFields) {
-      if (blendedField.isHidden || categorizeFieldType(blendedField.type) !== 'number') {
+      if (blendedField.isHidden) {
+        continue;
+      }
+      const isNumeric = categorizeFieldType(blendedField.type) === 'number';
+      if (!isNumeric && !aggregatedColumns.has(blendedField.name)) {
```
laskevych: 💡 Suggestion: The totals-eligibility predicate (`isNumeric || aggregatedColumns.has(name)`) is hand-inlined in both the native path (lines 331-332) and here. Since this is the load-bearing rule of the PR — and the likely landing spot for role-based gating later — extracting one shared helper would keep native and joined totals from silently diverging on the next tweak (the exact symmetry the new tests guard).

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -303,34 +318,55 @@ export class ReportSqlComposerService {
...
           | ReportAggregateFunction[]
           | undefined,
       }).allowedAggregations;
+    } else {
+      // Joined (blended) field: eligibility already applied in collectBlendedAllowedSets
+      // (numeric → auto summary; non-numeric → only when the report aggregates it).
+      allowed = blendedByName.get(name) ?? [];
     }
-    return blendedNumericByName.get(name) ?? [];
+    // ANY_VALUE (an arbitrary row's value) and STRING_AGG (whole-column concat) do not summarize
```
laskevych: 💡 Suggestion: The `fn !== 'ANY_VALUE' && fn !== 'STRING_AGG'` literal is the second independent encoding of "these functions don't summarize" — `LookerStudioAggregationMapperService.mapAggregateFunctionToLookerType` already maps the same pair to `DIMENSION` (lines 72-74). An exported `NON_SUMMARIZABLE_AGGREGATIONS` set in `field-aggregation-governance.ts` (the module that already owns per-function semantics) would force the next added function (`ARRAY_AGG`, `MODE`, …) to get one deliberate decision instead of silently leaking into totals.

apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
```diff
@@ -531,6 +596,56 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       expect(finalSelect).not.toMatch(/GROUP BY/);
     });
 
+    // Symmetry with the main-mart rule: a JOINED non-numeric field the report aggregates as a
+    // metric (e.g. COUNT_DISTINCT over a joined text column) must appear in totals too, by its
+    // post-join allowed functions — minus ANY_VALUE / STRING_AGG.
+    it('includes a JOINED non-numeric field in totals when the report aggregates it', async () => {
```
laskevych, line 602: 💡 Suggestion: The `blendedField` helper hardcodes `aggregateFunction: 'SUM'` (line 504) even for STRING fields, so this new blended non-numeric test drives the rollup through `SUM(country)` — SQL production never generates (the real default for text is `STRING_AGG`), through a mocked `resolveBlendingDecision` that also bypasses the run-time validator. The two new blended tests additionally assert only `result.columns`/`result.aggregations` metadata, never the composed SQL, so a renderer/alias regression on exactly the new non-numeric blended path wouldn't fail anything.

Suggest deriving `aggregateFunction` from the type (mirroring `getDefaultAggregateFunction`) and asserting the outer `COUNT(DISTINCT …)` + alias in at least one test; a case for the Unique-count-only and Sample-only shapes would close the remaining eligibility gaps too.

apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
```diff
@@ -369,6 +369,71 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
       expect(blendedReportDataService.resolveBlendingDecision).not.toHaveBeenCalled();
     });
 
+    // REPRODUCTION — Vlad's dashboard scorecard (chat thread YjeZ1yvBxSg / repro-totals.mjs):
```
laskevych: 🟢 Nitpick: This 10-line repro comment narrates the pre-fix code in present tense ("resolveTotalsAllowedForColumn hard-gates … RED today") and cites artifacts outside the repo (chat thread ID, `repro-totals.mjs`). At PR head those statements are already false or unresolvable. Two lines stating the requirement (COUNT_DISTINCT over a STRING metric with no grouping must appear in totals, per WI #6680 §D) preserve the intent without sending a future reader hunting for a gate that no longer exists.

zapolsky:
Thanks for the deep review. Addressed the blocking + important items and the suggestions; the open questions are deliberate decisions or scoped follow-ups.

**Fixed in code** (commits `4139d21`, `6b7aece`):

- **Blended override clamped** (L1): `postJoinAggregations` now goes through `resolveFieldGovernance` like the native path, so a stale numeric override on a STRING field can no longer emit `SUM(text)` and silently null the whole totals block. Added a clamp test.
- **Shared eligibility rule** (L10): extracted `isTotalsEligible`, used by both the native and joined paths so they can't diverge.
- **`NON_SUMMARIZABLE_AGGREGATIONS`** (L11): `ANY_VALUE`/`STRING_AGG` moved into `field-aggregation-governance.ts`, next to the Looker mapper's equivalent.
- **Contracts refreshed** (L4): `composeTotals` JSDoc, `ReportTotalsService` doc, the web run type/response comments, and the Swagger `totals` description + example; renamed `deriveNumericTotalsAggregations` → `deriveTotalsAggregations`. Comments no longer deny `aggregationRole` exists (L2) — it's persisted, just type-derived in practice.
- **Joined non-numeric limitation documented** (L3): the dedup-dependent `COUNT_DISTINCT` (computed over per-join roll-ups, default `STRING_AGG` for text) is now an explicit accepted-limitation note in code and docs, alongside the existing unweighted-AVG note.
- **Tests** (C2 / L12 / L13): the joined test now parameterizes the pre-join aggregate (STRING-valid) and asserts the generated inner CTE + ungrouped outer SQL; the repro comment was de-narrated.

**Deliberate — documented, no behavior change:**

- **L5** — the Unique-count toggle is intentionally not a totals metric (a separate PK-based metric; the docs already state "Unique count is not part of Totals").
- **L7** — a numeric field whose override is exactly `['ANY_VALUE']` now yields no totals; accepted edge, covered by the "`ANY_VALUE`/`STRING_AGG` excluded" contract.
- **L8** — totals use the field's full allowed set (not only the requested functions), symmetric with the numeric auto-summary; added the warehouse-cost note to the docs.

**Scoped as follow-ups — behavior-changing, beyond this bugfix:**

- **L2 (full)** — role-based eligibility gating (needs `aggregationRole` plumbed into `BlendedFieldDto`).
- **L6 / L8** — intersecting non-numeric totals with the requested functions.
- **L9** — native-vs-blended unified-name precedence (pre-existing divergence).

Happy to split any of the follow-ups into tickets.

apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
zapolsky: Done in `6b7aece` — the joined test now takes a STRING-valid pre-join (`ANY_VALUE`) and asserts the generated inner CTE plus the ungrouped outer SELECT, not just the derived function list.

docs/getting-started/setup-guide/report-aggregations.md
zapolsky: Done in `4139d21` — the `totals` ApiProperty description and example now cover report-aggregated non-numeric metrics and the `ANY_VALUE`/`STRING_AGG` exclusion.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Fixed in `6b7aece` — `postJoinAggregations` now goes through `resolveFieldGovernance` (the same clamp as the native path), so a stale numeric override on a STRING field can no longer emit `SUM(text)` and null the whole block. Added a clamp test.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Fixed the comments in `6b7aece` — they no longer deny `aggregationRole` exists (it is persisted, just type-derived in practice). Full role-based gating is a scoped follow-up, since it also needs `aggregationRole` plumbed into `BlendedFieldDto`.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Documented in `6b7aece` as an accepted limitation (code + docs), next to the existing unweighted-AVG note: joined non-numeric totals re-aggregate per-join roll-ups (default `STRING_AGG` for text), so the grand total is approximate. The clamp keeps the SQL valid.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Fixed in `6b7aece` — updated `composeTotals` JSDoc, `ReportTotalsService` doc, the web run type/response comments, and the Swagger `totals` description; renamed `deriveNumericTotalsAggregations` to `deriveTotalsAggregations`.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Deliberate — the Unique-count toggle is a separate PK-based metric and the docs already state it is not part of Totals. Left as-is.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Acknowledged — kept as-is for now; requiring at least one summarizable requested function (and intersecting totals with the requested set) is a scoped follow-up.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Accepted edge — a numeric field allowing only `ANY_VALUE` now contributes nothing, consistent with the `ANY_VALUE`/`STRING_AGG` exclusion. Documented.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Deliberate — totals use the field's full allowed set, symmetric with the numeric auto-summary; added the warehouse-cost note to the docs. Intersecting with the requested functions is a scoped follow-up.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Pre-existing divergence — left as-is here; adopting blended-wins precedence or rejecting native-vs-blended unified-name collisions is a scoped follow-up.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Done in `6b7aece` — extracted `isTotalsEligible`, shared by the native and joined paths.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Done in `6b7aece` — added `NON_SUMMARIZABLE_AGGREGATIONS` to `field-aggregation-governance.ts` and used it here.

apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
zapolsky: Done in `6b7aece` — the pre-join aggregate is parameterized and the test asserts the outer `COUNT(DISTINCT ...)` + alias; also added a clamp test. Unique-count-only and Sample-only shapes are noted as a follow-up.

apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
zapolsky: Done in `6b7aece` — replaced the repro narration with a two-line requirement statement.

max-voloshyn (approved):

laskevych (approved):

PR #1435, opened 2026-07-21
feat(data-marts): field-type matrix for AI queries + native IN + week/quarter relative-date presets

laskevych:
Closes the `query_data_mart` feedback loop ("AI fails on invalid operator sets") in three steps, one commit each.

## 1. Field type → operators/aggregations matrix (`8db8c99`)

- `get_data_mart_details_by_id` now annotates every native (incl. nested) and joined field with its type `category` and the effective `allowedAggregations`, resolved with the SAME governance function the validator enforces, and returns an `operators_by_category` matrix. The tool description of `query_data_mart` embeds a field-type matrix generated from the validator's own constants (spec-locked to the filter mapper, so the advertised contract cannot drift from the enforced one).
- `INVALID_OPERATOR_FOR_TYPE` now maps to a targeted `invalid_operator_for_type` error naming the field, its type, and the operators that DO fit — previously it fell through to a generic `query_failed` telling the model to re-fetch the schema, which caused retry loops. `DATE_TRUNC_*` misuse maps to `invalid_date_bucket` with per-variant fixes.
- Boolean fields become filterable: `eq`/`neq` with a boolean value translate to internal `is_true`/`is_false`.
- Removed the wrong `unsupported_operator` advice to emulate `in` with several `eq` filters (filters combine with AND — that could never match).

## 2. Native `in`/`not_in` on every storage + UI (`642f9af`)

- New `FilterRule` branch: array of scalars, 1..500 values (`IN_LIST_MAX_VALUES`), mirrored in the web zod schema.
- All 5 output-controls storages: BigQuery (one named param per value, CAST for date types), Athena (positional `?` per value, order-checked), Snowflake/Redshift/Databricks (inline escaped literals).
- Report UI: "is any of" / "is none of" for string/number/date/time columns with a comma-separated values input.
- MCP: `in`/`not_in` enabled, `is_empty`/`is_not_empty` exposed, and a new `invalid_filter_value` error surfaces malformed operands (empty list, bad `between` shape) that were previously swallowed.

## 3. Week/quarter/next-n-days relative-date presets (`c9439cd`)

- `this_week`/`last_week`, `this_quarter`/`last_quarter`, `next_n_days` across schema, web mirror, all 5 renderers, the report UI preset picker, and MCP.
- Week boundaries are ISO (Monday) on every storage — BigQuery truncates with `ISOWEEK` (its plain `WEEK` starts Sunday); Snowflake computes the Monday via `DAYOFWEEKISO` arithmetic because `DATE_TRUNC('week')` follows the session-level `WEEK_START`. `next_n_days` includes today, mirroring `last_n_days`.
- With these, `UNSUPPORTED_MCP_OPERATORS` is empty — every operator `query_data_mart` advertises is executable.

## Verification

- Full backend suite green: 412 suites / 5309 tests
- Web: 207 picker/schema tests green
- `tsc` backend + web clean, ESLint + markdownlint clean
- Docs updated: `docs/getting-started/setup-guide/mcp.md`, `output-controls.md`; changeset included

CleanShot 2026-07-23 at 10 20 31


apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -243,6 +271,54 @@ If truncated is true, not all matching rows were returned: narrow the query (few
...
+          .map(e => {
+            const category = categorizeFieldType(e.type ?? '');
+            // eq/neq on a boolean column ARE legal via translation — reaching here means the
+            // value was not a real boolean (e.g. the string "true"), so say that instead of
+            // listing eq/neq as both invalid and valid.
+            if (category === 'boolean' && (e.operator === 'eq' || e.operator === 'neq')) {
+              return `field '${e.column}' is boolean — use '${e.operator}' with a boolean true or false as the value (not a string)`;
+            }
+            const allowed = mcpOperatorsForCategory(category).join(', ');
+            return `operator '${e.operator}' cannot apply to field '${e.column}' (type ${e.type}); operators valid for this field: ${allowed}`;
```
prokopetsv: 🔵 **Question** — `e.operator` here is the *internal* operator from the validator error, which for translated operators differs from what the caller sent: `before` on a STRING column surfaces as `operator 'lt' cannot apply…`, and `eq true` on a STRING column surfaces as `'is_true'` — operators the model never used and cannot find in `operators_by_category`. The allowed-list guidance still lets it recover, so this is cosmetic, but a reverse lookup (internal → MCP names, which `MCP_TO_INTERNAL_OPERATOR` already gives you by inversion) would keep the error in the same vocabulary the tool speaks.

💡 **Suggestion** — same block: when a response mixes `INVALID_OPERATOR_FOR_TYPE` with date-bucket errors, the early return here reports only the operator issues; the model fixes those, retries, and only then sees `invalid_date_bucket` — one extra round trip for exactly the retry-loop this PR set out to eliminate. Concatenating both detail lists into one response would close it.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -151,19 +155,31 @@ const TIME_OPS = new Set([
...
 ]);
 const BOOL_OPS = new Set(['is_true', 'is_false', 'is_null', 'is_not_null']);
 
+/**
+ * Internal filter operators legal per field-type category — the authority
+ * `operatorAllowed` enforces. Exported so other layers (e.g. the MCP tools) can
+ * derive "which operators fit this field" guidance from the same source instead
+ * of keeping a copy that drifts.
+ */
+export const INTERNAL_OPERATORS_BY_CATEGORY: Record<FieldTypeCategory, ReadonlySet<string>> = {
```
prokopetsv: 💡 **Suggestion** — exporting this from the validator *service* makes `ee/mcp/tools/field-type-matrix.ts` import from the service layer for what is pure schema-level data. The per-category operator sets would sit more naturally in `dto/schemas/` next to `field-type-category.ts` (which already defines the categories they key on), with the validator importing them like it imports the category sets today. Same single-source guarantee, one layer lower — and the MCP tool no longer depends on a NestJS service module for a constant. Fine to defer; worth doing before a third consumer appears.

docs/getting-started/setup-guide/mcp.md
```diff
@@ -216,15 +216,16 @@ Returns field-level metadata for one data mart visible to you in the current pro
...
+| `operators_by_category` | For each field-type category present in the data mart (`number`/`string`/`date`/`time`/`boolean`/`other`), the `query_data_mart` filter/slice operators its fields accept                                                          |
 
-Use this tool when you need to understand the fields available in a specific data mart — both its native fields and any joined fields you can then query with `query_data_mart`. It does not return sample values, data freshness, owners, or actual data rows. To learn how joined/blended fields are set up, see [Joinable Data Marts](./joinable-data-marts.md).
+Use this tool when you need to understand the fields available in a specific data mart — both its native fields and any joined fields you can then query with `query_data_mart`. A field's `allowedAggregations` and its category's entry in `operators_by_category` tell the assistant exactly which aggregations and filter operators that field supports, so queries can be built without trial and error. It does not return sample values, data freshness, owners, or actual data rows. To learn how joined/blended fields are set up, see [Joinable Data Marts](./joinable-data-marts.md).
```
prokopetsv: 💡 **Suggestion** — house docs style (20-word sentences, no passive voice): the new sentence here is 30+ words with a passive tail ("so queries can be built without trial and error"). E.g.: "A field's `allowedAggregations` and its category's entry in `operators_by_category` tell the assistant which aggregations and operators the field supports. The assistant can build queries without trial and error." Same note as on #1430 — the page already reads long, but new sentences are the cheap place to hold the line.

prokopetsv (commented):
## Review

No High or Medium findings — this PR is in very good shape. The drift-protection design is the standout: the advertised operator matrix is derived from the validator's own `INTERNAL_OPERATORS_BY_CATEGORY`, and a spec locks `MCP_TO_INTERNAL_OPERATOR` to what `mapMcpFiltersToRules` actually produces, so the contract the tool description promises cannot silently diverge from what the validator enforces.

### Verified clean

- **Week/quarter arithmetic on all five dialects**: Snowflake's `DAYOFWEEKISO` Monday computation checks out (`1-DOW` → this Monday, `-6-DOW` → last Monday, session `WEEK_START`-independent); Spark `trunc(…, 'WEEK')`, Trino/Redshift `DATE_TRUNC('week')`, and BigQuery `ISOWEEK` all produce the same Monday boundary. Half-open ranges throughout; BigQuery's inclusive `<=` is correct because `lhs` wraps sub-day types in `DATE(col)`.
- **`in`/`not_in` safety**: bound parameters on BigQuery (with `CAST` for date types, param index correctly advanced) and Athena (positional, order-checked); inline dialects reuse the existing `lit` escaping, with `O'Brien`-style tests on each. List size capped at 500 (`IN_LIST_MAX_VALUES`) with the rationale documented.
- **Boolean translation**: `eq`/`neq` with a real boolean → `is_true`/`is_false`, string `"true"` deliberately left untranslated so type validation catches it — and the error mapper special-cases exactly that to give value guidance instead of a contradictory operator list.
- **Coverage**: renderer specs per dialect, mapper matrix spec, validator matrix, tool error-mapping paths, web picker + schema mirror. The removed `unsupported_operator` advice (emulating `in` with multiple `eq` filters — impossible under AND) was a genuine correctness fix.
- Changeset present with `'owox': minor`; web `TIME_OPERATORS` derives from `DATE_OPERATORS`, so time columns get "is any of / is none of" as the docs table claims.

Inline: two 💡 and one 🔵, all minor.

romandubovyi (commented):
Review complete. Focused on the new field-type matrix, native `in`/`not_in`, and the week/quarter presets — test coverage across all five storage renderers is thorough. Two non-blocking notes below.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
```diff
@@ -16,12 +23,33 @@ const inputSchema = z
...
 
+// Factories, not shared instances — a shared zod instance surfaces as a JSON-Schema
+// $ref that OpenAI's tool verifier cannot resolve (same hazard as makeMcpFilterSchema
+// in query-data-mart.input.ts). One factory keeps a single copy of each describe text.
+const makeCategoryField = () =>
+  z
+    .string()
+    .optional()
+    .describe(
+      'Field-type category (number/string/date/time/boolean/other) — the key into operators_by_category for the filter/slice operators this field accepts.'
```
romandubovyi: ⚪ **Non-blocking:** This describe text says `category` is the key for the filter/**slice** operators, but `enrichField` computes it from the post-join `type` only. For a joined field with a type-changing dedup aggregation (where `sliceType` is present and differs), slices run on the pre-join type — so following this guidance picks the wrong operator set for slices. Consider emitting a `sliceCategory` (from `sliceType`) when the two differ, or dropping "slice" from this description and pointing slice operator choice at `sliceType`, mirroring what the `sliceType` describe already warns about.

apps/backend/src/data-marts/dto/schemas/filter-config.schema.ts
```diff
@@ -51,6 +66,11 @@ const FilterRuleBaseSchema = z.discriminatedUnion('operator', [
     column: z.string().min(1),
     operator: NoValueOperatorEnum,
   }),
+  z.object({
+    column: z.string().min(1),
+    operator: z.enum(['in', 'not_in']),
+    value: z.array(ScalarValueSchema).min(1).max(IN_LIST_MAX_VALUES),
```
romandubovyi: ⚪ **Non-blocking:** The MCP path rejects mixed-type `in` lists in `mapOne` (its comment notes BigQuery types each bound param from its JS value and fails with "No matching signature for operator IN"), but this schema — the report/API path — accepts `['a', 5, true]`. A rule created via the API validates as clean and only fails at query time with a raw warehouse error on param-binding storages. A `.refine(list => list.every(v => typeof v === typeof list[0]))` here would enforce it once for both paths, and the MCP-side check could then just relay it.

romandubovyi (commented):
Review complete — two non-blocking notes on error-mapping cleanup and scope documentation.

🤖 Generated with Kimi Code

amarchenk0 (dismissed):
[P1] Do not split comma-containing list values

A list value containing a comma, such as `Acme, Inc.`, is accepted by the backend schema and preserved only while the rule remains untouched. On the first edit, `.split(/[,\n]/)` turns it into `['Acme', 'Inc.']`; users cannot create or correctly edit such values because there is no tag editor, escaping, or quoting grammar. This saves a different filter and returns incorrect report rows. Use an array/tag editor or define a supported escaping grammar, with regression coverage for creating and editing comma-containing values.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
laskevych: Both points addressed (and they survived the merge into the shared mapper, `output-controls-error.mapper.ts`, so add_report/update_report get them too):

- **Vocabulary**: `mcpOperatorNamesForInternal()` (reverse lookup over `MCP_TO_INTERNAL_OPERATOR`, spec-locked) translates the internal operator back — `relative_date` renders as the preset names that produce it, `lt` as `lt`/`before`. The translated `is_true`/`is_false` on a non-boolean column gets its own branch: the message says the *value* was a boolean and how to fix it, instead of naming an operator the caller never sent.
- **One response per retry**: the mapper now collects every recognized family into sections and returns one combined message (`error_code` = first family), so mixed operator + date-bucket + aggregation errors surface in a single round-trip. Covered by `combines every recognized family into one message` in the mapper spec.

Landed in 3abbe833f / a20f3b743, reintegrated post-merge in 287d17a83.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
laskevych: Done in 951dbfc84 — `INTERNAL_OPERATORS_BY_CATEGORY` and the per-category operator sets now live in `dto/schemas/field-type-category.ts`, right next to the category sets they key on. The validator imports them like it imports the type sets, and `field-type-matrix.ts` (plus the shared error mapper, which became the third consumer you predicted) no longer touches the service layer.

docs/getting-started/setup-guide/mcp.md
laskevych: Done in 951dbfc84 — applied your wording verbatim: two sentences, active voice, under 20 words each.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
laskevych: Done in 951dbfc84 — went with the `sliceCategory` option: `enrichField` now emits it (computed from `sliceType`) whenever `sliceType` is present, and registers that category in `operators_by_category` so the lookup key always resolves. The `category` describe no longer claims slices — it points at `sliceCategory` when present. Spec covers an INTEGER/STRING dedup field asserting both categories land in the matrix.

apps/backend/src/data-marts/dto/schemas/filter-config.schema.ts
laskevych: Done in 951dbfc84 — added the `.refine` (same-type check) to the backend `in`/`not_in` branch and mirrored it in the web schema, so API-created rules fail at save time with a named message instead of a raw warehouse error at query time. Kept the MCP-side check in `mapOne` rather than relaying: it fires before the facade call, so the caller gets the precise `invalid_filter_value` without a billable round-trip. Schema specs cover mixed-rejected / homogeneous-accepted on both sides.

laskevych:
@amarchenk0 P1 addressed in 951dbfc84 — went with the escaping-grammar option:

- **Create/edit comma-containing values**: the value input now speaks a CSV-style grammar — wrap a value in double quotes to keep commas (and newlines) literal, `""` escapes a literal quote. `alpha, "Acme, Inc.", beta` → `['alpha', 'Acme, Inc.', 'beta']`.
- **Round-trip**: the display serializer quotes any stored value containing a comma/quote/newline when the editor opens, so re-parsing an edited rule can no longer split it. On top of that, an *untouched* rule short-circuits to its pristine value array (added earlier in 3abbe833f), which also preserves value types exactly.
- **Regression coverage**: FilterValueEditor specs for creating a quoted comma value, editing an existing comma-containing rule (quoted display → comma survives the save), and `""` quote-escaping; plus the earlier untouched-round-trip specs.
- The field label now documents the grammar: *Values (comma-separated; wrap a value in "double quotes" if it contains a comma)*.

A chip/tag editor would still be the nicer long-term UX — happy to file it as a follow-up — but the grammar closes the correctness hole: comma values are now creatable, editable, and never silently split.

prokopetsv (commented):
## Re-review at 951dbfc84 (review rounds 1–3 + follow-up fixes)

No blockers. I re-verified every thread from the three prior review rounds against the current head, and each is genuinely addressed — not just claimed:

- **Comma-containing in-list values** (the CHANGES_REQUESTED item): fixed with a two-layer design — a pristine-array bypass (an untouched rule re-emits its original value array verbatim, covering non-string types too) plus a CSV-style quote grammar (`"Acme, Inc."`, `""` escaping) for created/edited text. Tests cover create, edit round-trip (display shows the quoted form), the in↔not_in operator flip keeping pristine values, and literal-quote escaping.
- **`sliceCategory`**: emitted whenever `sliceType` is present, registered in `operators_by_category` so the lookup key always resolves, and the `category` describe no longer claims slices; tested with an INTEGER/STRING dedup field.
- **Mixed-type `in` lists**: same-type `.refine` on the backend schema (save-time rejection with a named message instead of a query-time warehouse error), mirrored in the web schema; the MCP mapper additionally rejects non-finite numbers and reports the precise `invalid_filter_value`.
- **Operator sets** moved to `dto/schemas/field-type-category.ts`; the validator and the MCP matrix now import the same schema-layer constant, and the shared error mapper became the predicted third consumer.
- **Error vocabulary**: `mcpOperatorNamesForInternal` translates internal operators back to MCP names (`relative_date` → the preset list), with dedicated handling for `is_true`/`is_false` on a non-boolean column (points at the boolean *value*, the actual mistake).
- **Combined error families**: the shared mapper now surfaces every recognized family in one message (first family sets the code), so mixed validation failures cost one retry, not one per class — and the fix landing in the shared mapper means add_report/update_report filters get it too.
- Also caught along the way (good find): explicit `allowedAggregations: []` ("nothing allowed") is now forwarded instead of dropped, so the tool no longer advertises type-default aggregations the validator would reject; and `in_last_n_days`/`in_next_n_days` no longer coerce `true`→1 or `[7]`→7.

Two 🟡 non-blockers inline. Note @amarchenk0's CHANGES_REQUESTED is still standing and needs their re-review or dismissal.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterValueEditor.tsx
```diff
@@ -87,6 +121,56 @@ function parseScalar(raw: string, fieldType: string): string | number | boolean
...
+/**
+ * Splits the comma-separated list text into values. Double-quoted sections keep
+ * commas/newlines literal and use `""` for a literal quote; unquoted content is
+ * split on commas and newlines. Assembled tokens are trimmed; empties dropped.
+ */
+function parseListText(text: string): string[] {
+  const values: string[] = [];
+  let current = '';
+  let inQuotes = false;
+  const push = () => {
```
prokopetsv, line 180: 🟡 **Non-Blocker** — `push()` trims every assembled token, including quoted ones, so the quote grammar cannot round-trip a value with meaningful leading/trailing whitespace: `" x "` parses to `x`, even though quoting is exactly the CSV mechanism that should preserve it (and `formatListValue` will happily emit `" x, y "` for such a value on edit — which then comes back as `x, y`). The pristine-array bypass protects untouched rules, so the exposure is edits only, and whitespace-padded IN values are rare — hence non-blocking. Fix is small: track whether the current token contained a quoted section and skip the trim for those.

apps/backend/src/ee/mcp/tools/output-controls-error.mapper.ts
```diff
@@ -82,70 +103,126 @@ export function translateOutputControlsError(
...
+    sections.push({
+      code: 'invalid_date_bucket',
+      message: `Invalid date_buckets — ${details}. The field name(s) are correct, so do not re-fetch the schema.`,
+    });
+  }
+
+  // Codes no branch above recognizes still get named — no current or future
+  // validator code is ever fully opaque to the MCP client. Appended as the last
+  // section so recognized families keep their targeted guidance first.
+  const unrecognized =
```
prokopetsv: 🟡 **Non-Blocker** — this `unrecognized` filter is a hand-maintained negative list re-enumerating every code the sections above already claim. The next error family added to this mapper must also be added here, in a second place ~150 lines away; forgetting it means the error renders twice (once with targeted guidance, once as a raw code in the fallback) — benign, but confusing to the model the message is written for. Building one `recognizedCodes` set as each section claims its codes (or hoisting the per-family code sets into a single registry the fallback subtracts) makes it self-maintaining.

prokopetsv:
@laskevych I expect @amarchenk0 will approve it once the requested changes are applied, so let me know if I should mark the pull request as approved, since I don't see any blockers here.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterValueEditor.tsx
laskevych: Fixed in 7f8c6b2a8 — went with your suggestion plus the serializer half: `parseListText` tracks whether the token contained a quoted section and keeps quoted content verbatim (`" x "` → `' x '`), while stray whitespace *around* a quoted token is still dropped (leading skipped before the token starts, an unquoted tail run sliced off). `formatListValue` got its edge-whitespace trigger back, so `' x '` displays as `" x "` on edit and re-parses losslessly. Specs cover both directions: create (`" x " , "a" , plain`) and the edit round-trip.

apps/backend/src/ee/mcp/tools/output-controls-error.mapper.ts
laskevych: Fixed in 7f8c6b2a8 — the fallback now subtracts a single `RECOGNIZED_CODES` registry composed from the per-family code sets it sits next to at the top of the file (spread of `NOT_SELECTED_CODES`/`AGG_NOT_ALLOWED_CODES`/`DATE_BUCKET_ERROR_CODES` plus the single-code families). A new family's codes get added once, next to the other sets — the negative list at the bottom is gone.

apps/backend/src/data-marts/dto/schemas/filter-config.schema.ts
```diff
@@ -51,6 +66,21 @@ const FilterRuleBaseSchema = z.discriminatedUnion('operator', [
...
+      .min(1)
+      .max(IN_LIST_MAX_VALUES)
+      .refine(list => list.every(v => typeof v === typeof list[0]), {
+        message:
+          'in/not_in values must all be the same type (all strings, all numbers, or all booleans)',
+      }),
+  }),
   z.object({
     column: z.string().min(1),
     operator: z.literal('between'),
```
romandubovyi, line 107: ⚪ **Non-blocking:** The `in`/`not_in` branch above got a same-type `.refine` (a mixed list fails only at query time — BigQuery types each bound param individually and raises "No matching signature for operator IN"). The `between` branch here accepts `{ from: Scalar, to: Scalar }` with no such check, and the MCP `mapOne` path only verifies the `from`/`to` keys exist. So `between` with `from: "2026-01-01"` and `to: 100` passes both layers and fails only at query time — the same save-time-clean/query-time-failure class the list refine was added to close. Consider a same-type `.refine` on the `between` value (and mirroring it in `mapOne`).

apps/backend/src/data-marts/dto/schemas/filter-config.schema.ts
```diff
@@ -51,6 +66,21 @@ const FilterRuleBaseSchema = z.discriminatedUnion('operator', [
...
+    column: z.string().min(1),
+    operator: z.enum(['in', 'not_in']),
+    // Same-type only: param-binding storages type each bound value individually
+    // (BigQuery raises "No matching signature for operator IN" on a mixed list),
+    // so a mixed list saved via the API would fail only at query time otherwise.
+    value: z
+      .array(ScalarValueSchema)
+      .min(1)
+      .max(IN_LIST_MAX_VALUES)
+      .refine(list => list.every(v => typeof v === typeof list[0]), {
```
romandubovyi: ⚪ **Non-blocking:** This refine accepts an all-boolean list (all same type), and the message advertises "all booleans" as valid — but no field category permits `in`/`not_in` on boolean columns (`BOOL_OPS` has no `in`), and on a STRING/NUMBER/DATE column an all-boolean list passes the validator (`operatorAllowed` checks only the operator name, not element types) and dies at query time with a raw warehouse error (BigQuery types each bound param as BOOL → "No matching signature for operator IN"). Reachable via MCP `mapOne` with real JSON booleans. Same class as the mixed-list gap this refine closes. Consider rejecting boolean elements for `in`/`not_in` (steer callers to `is_true`/`is_false`) and dropping "all booleans" from the message.

apps/backend/src/ee/mcp/tools/query-data-mart.input.ts
```diff
@@ -220,39 +252,101 @@ function mapOne(
...
+      // Only a number or a numeric string counts — Number() alone would coerce
+      // true→1 and [7]→7, silently running a query the caller didn't ask for.
+      const raw = f.value;
+      const n =
+        typeof raw === 'number'
+          ? raw
+          : typeof raw === 'string' && raw.trim() !== ''
+            ? Number(raw)
+            : NaN;
+      if (!Number.isInteger(n) || n <= 0) {
```
romandubovyi: ⚪ **Non-blocking:** This branch requires a positive integer but not the upper bound that `RelativeDatePresetSchema` (and the UI) enforce — `n` is capped at `.max(3650)`. An MCP call with e.g. `in_next_n_days: 9999` therefore passes `mapOne`, then fails later in `FilterConfigSchema.safeParse` as a generic "invalid shape" instead of the precise `InvalidFilterValueError` this PR introduces for other bad operands. Reject `n > 3650` here (same bound as the schema) so the model gets a fixable value error.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterValueEditor.tsx
```diff
@@ -109,20 +206,36 @@ function buildRule(args: { column: string; fieldType: string; state: EditorState
...
+    // unquoted entries are trimmed and empties dropped. Mirrors the backend
+    // schema bounds (1..IN_LIST_MAX_VALUES values).
+    const entries = parseListText(state.list);
+    if (entries.length === 0) {
+      throw new Error('At least one value is required');
+    }
+    if (entries.length > IN_LIST_MAX_VALUES) {
+      throw new Error(`At most ${IN_LIST_MAX_VALUES} values are allowed`);
+    }
+    return { column, operator: op, value: entries.map(e => parseScalar(e, fieldType)) };
```
romandubovyi, line 254: ⚪ **Non-blocking:** Number columns validate every list entry (`parseScalar` throws on a non-numeric value, so the rule stays null), but date/time list entries get no per-entry validation here — the list is a `type='text'` input and `parseScalar` returns the raw string for non-number types, while the single-value date path relies on the native `type=date` input to enforce the format. An `in` list with a typo like `2026-13-45` on a DATE column therefore saves as a clean rule and only fails at query time with a raw warehouse error — the save-time-clean/query-time-failure pattern the rest of this PR eliminates. Consider validating date-category entries (e.g. an ISO-date shape check), mirroring the number handling.

docs/getting-started/setup-guide/mcp.md
```diff
@@ -216,15 +216,16 @@ Returns field-level metadata for one data mart visible to you in the current pro
...
+| `name`                  | Data mart name                                                                                                                                                                                                                    |
+| `description`           | Data mart description                                                                                                                                                                                                             |
+| `fields`                | The data mart's own (native) output fields with names, types, descriptions, and business names when available. Each field also carries its type `category` and the effective `allowedAggregations` `query_data_mart` may apply to it |
+| `joined_fields`         | Fields contributed by blended/joined data marts (empty when the data mart has no joins), each with its qualified `<alias>__<field>` name, source data mart, type, `category`, and effective `allowedAggregations`                   |
```
romandubovyi: ⚪ **Non-blocking:** `joined_fields` here lists `category` and `allowedAggregations`, but the tool now also emits `sliceType` / `sliceCategory` when a type-changing dedup makes the pre-join type differ from the blended type (and the field descriptions point assistants at those keys for slice operators). Please document those optional fields here so the public MCP contract matches what `get_data_mart_details_by_id` returns.

docs/getting-started/setup-guide/output-controls.md
```diff
@@ -36,21 +36,24 @@ A filter runs against the final `SELECT`, after all joins complete. Use filters
...
-| Number | =, ≠, >, <, ≥, ≤, between, is null, is not null |
-| Date / DateTime / Timestamp | on, not on, after, before, on or after, on or before, between, relative, is null, is not null |
-| Time | at, not at, after, before, at or after, at or before, between, is null, is not null |
+| String | is, is not, is any of, is none of, contains, does not contain, starts with, ends with, is empty, is not empty, is null, is not null, matches regex, does not match regex |
+| Number | =, ≠, is any of, is none of, >, <, ≥, ≤, between, is null, is not null |
+| Date / DateTime / Timestamp | on, not on, is any of, is none of, after, before, on or after, on or before, between, relative, is null, is not null |
+| Time | at, not at, is any of, is none of, after, before, at or after, at or before, between, is null, is not null |
 | Boolean | is true, is false, is null, is not null |
 
+**Is any of / is none of** match a column against a list of values (SQL `IN` / `NOT IN`). Enter the values comma-separated — up to 500 per rule.
```
romandubovyi: ⚪ **Non-blocking:** The editor supports (and, for comma-containing values, requires) a CSV-style quoting grammar — wrap a value in double quotes, escape a literal quote as `""` — but this line only says "Enter the values comma-separated". A user with a value like `Acme, Inc.` who follows the docs gets it split into two values. Add a sentence documenting the grammar, e.g.: `Wrap a value in double quotes if it contains a comma — for example "Acme, Inc.". Use "" for a literal quote.`

romandubovyi (approved):
Multi-agent review (Antigravity CLI, Grok CLI, Claude Code). No blockers — approving. Left 6 non-blocking inline notes: `between`/all-boolean/day-count validation gaps of the same save-time-clean/query-time-failure class this PR closes elsewhere, one web-side date-list validation gap, and two docs completeness notes.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Grok CLI (Grok 4.5) + Claude Code (Opus 4.8)

apps/backend/src/data-marts/dto/schemas/filter-config.schema.ts
laskevych: Fixed in 9aea6e91e — same-type `.refine` on the `between` value (backend + web mirror) and a mirrored precise `InvalidFilterValueError` in `mapOne`, so the MCP caller gets the operand error before the billable call and the API path fails at save time. Schema specs cover both directions.

apps/backend/src/data-marts/dto/schemas/filter-config.schema.ts
laskevych: Fixed in 9aea6e91e — boolean elements are now rejected for `in`/`not_in` at every layer: schema refines (backend + web mirror) plus a dedicated `mapOne` check whose message steers to `eq`/`neq` with a single true|false (the MCP boolean vocabulary); "all booleans" is gone from the same-type message. Specs cover all-boolean and mixed lists on both sides.

apps/backend/src/ee/mcp/tools/query-data-mart.input.ts
laskevych: Fixed in 9aea6e91e — the cap is now a shared `RELATIVE_DATE_MAX_N` constant exported from filter-config.schema.ts (the preset schema's `.max()` reads it too), and `mapOne` rejects `n > 3650` with the precise operand error ("positive integer up to 3650") instead of letting it fall through to the generic shape failure.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/FilterValueEditor.tsx
laskevych: Fixed in 9aea6e91e — date-category list entries are validated per entry (YYYY-MM-DD shape + a component round-trip through Date.UTC, since V8 rolls `2026-02-30` over to Mar 2 instead of rejecting it), and TIME columns get an HH:MM(:SS) shape check — mirroring what the native single-value inputs enforce. Specs cover the typo, the rollover date, and the valid-list path.

docs/getting-started/setup-guide/mcp.md
laskevych: Fixed in 9aea6e91e — the `joined_fields` row now documents `sliceType`/`sliceCategory` (present when a type-changing dedup makes the pre-join type differ; use them to pick slice operators).

docs/getting-started/setup-guide/output-controls.md
laskevych: Fixed in 9aea6e91e — added your suggested wording to the output-controls page: wrap a comma-containing value in double quotes (`"Acme, Inc."`), `""` for a literal quote.

PR #1440, opened 2026-07-21
test(backend): anchor relative_date integration seed to the calendar year

zapolsky:
## Problem

The scheduled **Integration Tests (Real DB)** workflow started failing deterministically for **BigQuery** and **Athena** from **2026-07-20** onward (green through Jul 19, red every run since). Snowflake / Redshift / Databricks stayed green.

Both failures were the same test on both engines:

> `Output controls — operator matrix & dates › relative_date this_year on created_at → rows 1,2,4,5 (excludes future row 7)`

It is **not flaky** — it is a calendar time-bomb. The matrix table seeded its "last year" row (row 6) as `CURRENT_DATE − 200 days`. `this_year` renders as the whole calendar year (`>= DATE_TRUNC(CURRENT_DATE, YEAR) AND < …+1 YEAR`). In a non-leap year day 200 is Jul 19, so once the run date passes day 201 (Jul 20), `CURRENT_DATE − 200 days` lands back in the current calendar year and row 6 unexpectedly enters `this_year`, breaking `toEqual([1,2,4,5])`.

The over-specified `toEqual` also hid a second, mirror-image window: recent rows 2 (`−40d`) and 4 (`−5d`) legitimately fall *out* of `this_year` in early January, so the same test would also fail for the first weeks of the year.

## Fix

Bring BigQuery and Athena in line with the robust pattern already used by the Snowflake / Redshift / Databricks suites:

- **Anchor row 6 to the calendar year** — `DATE_TRUNC(YEAR) − 6 months` (mid last year) instead of a sliding `−200 days`, so it stays firmly in last year all year round.
- **Relax the `this_year` assertion to membership checks** — today rows in (`toContain`), other-year rows out (`not.toContain` for rows 3/6/7). Recent rows 2/4 are no longer asserted for `this_year`; their bucket membership is covered by the `last_n_days` / `last_n_months` tests.

## Regression guard (TDD)

Added `relative-date-seed-invariants.spec.ts` — a **date-independent** unit test that mirrors the `renderRelativeDate` boundaries and asserts each seed role lands in its intended calendar bucket for **every day of several representative years** (incl. leap 2028 and non-leap century 2100). It fails with the old `−200 days` formula (from day 201) and passes with the anchored one — proving the fix holds regardless of run date, not just today. An `anchoring` block ties the mirrored semantics back to the real renderer, so the guard fails loudly if the preset boundaries ever change.

## Verification

Ran the full matrix suites locally against the real databases:

| Suite | Before | After |
|---|---|---|
| BigQuery | 67 ✅ / 1 ❌ | **68 / 68 ✅** |
| Athena | 57 ✅ / 1 ❌ | **58 / 58 ✅** |

Unit guard: 9/9 across all sampled days/years.

> Note: these integration jobs run on a schedule (not on PRs), so they won't run automatically here — the results above are from local runs against the real BigQuery/Athena.

## Notes

- Test-only change; the production `relative_date` renderer is unchanged. No changeset (per repo convention, test-only changes don't get one).

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/utils/relative-date-seed-invariants.spec.ts
+++ b/apps/backend/src/data-marts/data-storage-types/utils/relative-date-seed-invariants.spec.ts
@@ -1,0 +1,183 @@
+/**
+ * Calendar-bucket invariants for the relative_date integration-test seed.
+ *
+ * The real-DB integration suites (bigquery/athena/snowflake/redshift/databricks
+ * `.integration.ts`) seed a fixed set of rows whose dates are expressed relative
+ * to the DB clock, then assert which rows each `relative_date` preset returns.
+ * Because a preset compares against the *live* DB date, any seed row whose date
+ * is a sliding day-offset can drift across a calendar boundary and silently
+ * change which bucket it lands in — turning a green suite red on a specific day
+ * of the year. (Concretely: a "-200 days" row enters `this_year` once the run
+ * date passes day 201 of the year, which is exactly what broke the scheduled
+ * BigQuery + Athena suites on 2026-07-20.)
+ *
+ * This spec pins the seed *design* — not the DB — by mirroring the preset
+ * boundaries as pure date arithmetic and asserting each seed role lands in its
+ * intended calendar bucket for EVERY day of several representative years. It is
+ * the date-independent guard the integration suites themselves cannot be (they
+ * only ever run on a single day).
+ *
+ * The `anchoring` block ties the mirrored semantics back to production, so this
+ * spec fails loudly if `renderRelativeDate` ever moves the bucket boundaries and
+ * the mirror below (and the integration seed) must be revisited.
+ */
+import { BigQueryClauseRenderer } from '../bigquery/services/bigquery-clause-renderer';
+
+// --- UTC date arithmetic mirroring the SQL the renderer emits ----------------
+
+const DAY_MS = 86_400_000;
+
+/** DATE_ADD/DATE_SUB(..., INTERVAL n DAY) — exact day shift. */
+const shiftDays = (d: Date, n: number): Date => new Date(d.getTime() + n * DAY_MS);
+
+/** DATE_ADD/DATE_SUB(..., INTERVAL n MONTH) — month shift with end-of-month clamp. */
+function shiftMonths(d: Date, n: number): Date {
+  const total = d.getUTCFullYear() * 12 + d.getUTCMonth() + n;
+  const year = Math.floor(total / 12);
+  const month = ((total % 12) + 12) % 12;
+  const lastDay = new Date(Date.UTC(year, month + 1, 0)).getUTCDate();
+  return new Date(Date.UTC(year, month, Math.min(d.getUTCDate(), lastDay)));
+}
+
+/** DATE_TRUNC(..., YEAR) — first day of the row's calendar year. */
+const truncYear = (d: Date): Date => new Date(Date.UTC(d.getUTCFullYear(), 0, 1));
+
+const iso = (d: Date): string => d.toISOString().slice(0, 10);
+
+// --- preset predicates, mirroring renderRelativeDate -------------------------
+
+const isToday = (row: Date, today: Date): boolean => row.getTime() === today.getTime();
+
+const inLastNDays = (row: Date, today: Date, n: number): boolean =>
+  row.getTime() >= shiftDays(today, -n).getTime() && row.getTime() <= today.getTime();
+
+const inLastNMonths = (row: Date, today: Date, n: number): boolean =>
+  row.getTime() >= shiftMonths(today, -n).getTime() && row.getTime() <= today.getTime();
+
+const inThisYear = (row: Date, today: Date): boolean =>
+  row.getTime() >= truncYear(today).getTime() &&
+  row.getTime() < Date.UTC(today.getUTCFullYear() + 1, 0, 1);
+
+// --- seed roles from the bigquery/athena matrix table ------------------------
+// The date each role holds, as a function of the DB "today".
+
+const seed = {
+  today: (t: Date): Date => t, // rows 1 & 5
+  recent5d: (t: Date): Date => shiftDays(t, -5), // row 4
+  mid40d: (t: Date): Date => shiftDays(t, -40), // row 2
+  oldLastYear400d: (t: Date): Date => shiftDays(t, -400), // row 3
+  // row 6 — the "last year" bucket, anchored to the calendar year (mid last year)
+  // so it never drifts into this_year or the sliding windows regardless of run date.
+  // SQL: DATE_SUB(DATE_TRUNC(CURRENT_DATE(), YEAR), INTERVAL 6 MONTH) → Jul 1 of last year.
+  lastYearBucket: (t: Date): Date => shiftMonths(truncYear(t), -6),
+  future13m: (t: Date): Date => shiftMonths(t, 13), // row 7
+};
+
+// --- every day of a spread of representative years ---------------------------
+
+function eachDayUTC(years: number[]): Date[] {
+  const out: Date[] = [];
+  for (const y of years) {
+    for (let m = 0; m < 12; m++) {
+      const daysInMonth = new Date(Date.UTC(y, m + 1, 0)).getUTCDate();
+      for (let day = 1; day <= daysInMonth; day++) out.push(new Date(Date.UTC(y, m, day)));
+    }
+  }
+  return out;
+}
+
+// 2028 is a leap year; 2100 is a non-leap century year — both stress boundaries.
+const YEARS = [2025, 2026, 2027, 2028, 2100];
+const DAYS = eachDayUTC(YEARS);
+
+describe('relative_date seed calendar-bucket invariants', () => {
+  describe('anchoring: production preset boundaries the mirror depends on', () => {
+    const bq = new BigQueryClauseRenderer();
+
+    it('this_year spans the whole calendar year (start of year .. start of next year)', () => {
+      const sql = bq.renderWhere([
+        { column: 'd', operator: 'relative_date', value: { kind: 'this_year' } },
+      ]).sql;
+      expect(sql).toContain('DATE_TRUNC(CURRENT_DATE(), YEAR)');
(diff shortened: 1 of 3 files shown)


apps/backend/test/integration/bigquery.integration.ts
```diff
@@ -795,7 +798,7 @@ describeIfCredentials('Output controls — operator matrix & dates (real BigQuer
         (3, 'gamma',    'c',    30,  true,  DATE_SUB(CURRENT_DATE(), INTERVAL 400 DAY),    TIMESTAMP(DATE_SUB(CURRENT_DATE(), INTERVAL 400 DAY))),
         (4, 'alphabet', 'a%b',  40,  true,  DATE_SUB(CURRENT_DATE(), INTERVAL 5 DAY),      TIMESTAMP(DATE_SUB(CURRENT_DATE(), INTERVAL 5 DAY))),
         (5, 'ALPHA',    'a_b',  50,  false, CURRENT_DATE(),                                TIMESTAMP_ADD(TIMESTAMP(CURRENT_DATE()), INTERVAL 13 HOUR)),
-        (6, '',         'x',     0,  true,  DATE_SUB(CURRENT_DATE(), INTERVAL 200 DAY),    TIMESTAMP(DATE_SUB(CURRENT_DATE(), INTERVAL 200 DAY))),
+        (6, '',         'x',     0,  true,  DATE_SUB(DATE_TRUNC(CURRENT_DATE(), YEAR), INTERVAL 6 MONTH),  TIMESTAMP(DATE_SUB(DATE_TRUNC(CURRENT_DATE(), YEAR), INTERVAL 6 MONTH))),
```
romandubovyi, line 801: 💡 **Suggestion:** BigQuery and Athena matrix suites are updated in this PR. It would be good to double-check if Snowflake, Redshift, and Databricks integration suites already use calendar anchoring for row 6 or if they should be aligned as well to keep test seeds consistent.

apps/backend/test/integration/athena.integration.ts
```diff
@@ -710,7 +713,7 @@ AS SELECT * FROM (VALUES
   (3, 'gamma',    'c',    30,  true,  date_add('day', -400, current_date),   cast(date_add('day', -400, current_date) AS timestamp)),
   (4, 'alphabet', 'a%b',  40,  true,  date_add('day', -5, current_date),     cast(date_add('day', -5, current_date) AS timestamp)),
   (5, 'ALPHA',    'a_b',  50,  false, current_date,                          date_add('hour', 13, cast(current_date AS timestamp))),
-  (6, '',         'x',     0,  true,  date_add('day', -200, current_date),   cast(date_add('day', -200, current_date) AS timestamp)),
+  (6, '',         'x',     0,  true,  date_add('month', -6, date_trunc('year', current_date)),  cast(date_add('month', -6, date_trunc('year', current_date)) AS timestamp)),
```
romandubovyi, line 716: 💡 **Suggestion:** Since the operator matrix table schema, row semantics, and comments are duplicated across multiple integration test suites (`athena.integration.ts`, `bigquery.integration.ts`, etc.), consider extracting a shared helper or builder function for seed values in a future refactoring to simplify maintenance when updating test dates.

apps/backend/src/data-marts/data-storage-types/utils/relative-date-seed-invariants.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+ * change which bucket it lands in — turning a green suite red on a specific day
+ * of the year. (Concretely: a "-200 days" row enters `this_year` once the run
+ * date passes day 201 of the year, which is exactly what broke the scheduled
+ * BigQuery + Athena suites on 2026-07-20.)
+ *
+ * This spec pins the seed *design* — not the DB — by mirroring the preset
+ * boundaries as pure date arithmetic and asserting each seed role lands in its
+ * intended calendar bucket for EVERY day of several representative years. It is
+ * the date-independent guard the integration suites themselves cannot be (they
+ * only ever run on a single day).
```
romandubovyi, line 18: 💡 **Suggestion:** Since this spec lives in a generic `utils` directory but anchors on `BigQueryClauseRenderer`, consider adding a one-line note explaining that BigQuery's renderer is the canonical reference for the cross-engine seed boundaries. This saves future maintainers from wondering why only BigQuery is checked.

apps/backend/src/data-marts/data-storage-types/utils/relative-date-seed-invariants.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+ * spec fails loudly if `renderRelativeDate` ever moves the bucket boundaries and
+ * the mirror below (and the integration seed) must be revisited.
+ */
+import { BigQueryClauseRenderer } from '../bigquery/services/bigquery-clause-renderer';
+
+// --- UTC date arithmetic mirroring the SQL the renderer emits ----------------
+
+const DAY_MS = 86_400_000;
+
+/** DATE_ADD/DATE_SUB(..., INTERVAL n DAY) — exact day shift. */
```
romandubovyi, line 30: 🔵 **Question:** Have you verified that this JS month-shift logic exactly matches BigQuery / Athena semantics for end-of-month clamping (e.g., May 31 minus 3 months)? The current seed roles don't hit those edges, but future extensions might.

apps/backend/src/data-marts/data-storage-types/utils/relative-date-seed-invariants.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    for (let m = 0; m < 12; m++) {
+      const daysInMonth = new Date(Date.UTC(y, m + 1, 0)).getUTCDate();
+      for (let day = 1; day <= daysInMonth; day++) out.push(new Date(Date.UTC(y, m, day)));
+    }
+  }
+  return out;
+}
+
+// 2028 is a leap year; 2100 is a non-leap century year — both stress boundaries.
+const YEARS = [2025, 2026, 2027, 2028, 2100];
```
romandubovyi, line 90: 💡 **Suggestion:** Consider adding a focused test case that uses Feb 29 as `today` (or a role derived from it) to ensure leap-day handling is covered, or add a comment explaining why the current year spread already provides enough boundary coverage.

romandubovyi (commented):
Review complete. The fix correctly anchors the last-year bucket and the new regression guard prevents future calendar time-bombs.

apps/backend/test/integration/bigquery.integration.ts
```diff
@@ -1013,13 +1019,19 @@ describeIfCredentials('Output controls — operator matrix & dates (real BigQuer
...
+  it('relative_date this_year on created_at → current-year rows in, other-year rows out', async () => {
     const rows = await runMatrix({
       filters: [{ column: 'created_at', operator: 'relative_date', value: { kind: 'this_year' } }],
     });
-    // Row 7 is ~13 months in the future (next calendar year) and must NOT appear.
-    expect(ids(rows)).not.toContain(7);
-    expect(ids(rows)).toEqual([1, 2, 4, 5]);
+    const got = ids(rows);
+    // today rows are always in the current calendar year
+    expect(got).toContain(1);
```
romandubovyi, line 1028: 💡 **Suggestion:** Membership checks fix the January brittleness, but the positive side now only asserts *today* rows (1, 5). A regression that narrows `this_year` to "today only" would still pass here. If you want a date-stable positive check for non-today current-year membership without bringing back the early-January flake, consider seeding a row at `DATE_TRUNC(CURRENT_DATE(), YEAR)` (always in-year, never "today" except Jan 1 — or use `+ 1 DAY` with care) and `toContain` that id. Optional; Snowflake's suite is similarly soft today.

apps/backend/src/data-marts/data-storage-types/utils/relative-date-seed-invariants.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+const DAYS = eachDayUTC(YEARS);
+
+describe('relative_date seed calendar-bucket invariants', () => {
+  describe('anchoring: production preset boundaries the mirror depends on', () => {
+    const bq = new BigQueryClauseRenderer();
+
```
romandubovyi, line 96: 🔵 **Question:** The `anchoring` block only exercises `BigQueryClauseRenderer`. Athena has its own clause renderer/dialect (`date_trunc('year', current_date)`, etc.). Is BQ intentionally the single representative for shared calendar semantics, or should Athena get a parallel string check so a dialect-only boundary change can't silently drift from this mirror?

apps/backend/src/data-marts/data-storage-types/utils/relative-date-seed-invariants.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+const inThisYear = (row: Date, today: Date): boolean =>
+  row.getTime() >= truncYear(today).getTime() &&
+  row.getTime() < Date.UTC(today.getUTCFullYear() + 1, 0, 1);
+
+// --- seed roles from the bigquery/athena matrix table ------------------------
+// The date each role holds, as a function of the DB "today".
+
+const seed = {
+  today: (t: Date): Date => t, // rows 1 & 5
+  recent5d: (t: Date): Date => shiftDays(t, -5), // row 4
```
romandubovyi, line 66: 💡 **Suggestion:** These `seed.*` formulas are the design under test, but nothing asserts they still match the INSERT expressions in `bigquery.integration.ts` / `athena.integration.ts`. If someone reintroduces `−200 days` only in the SQL seed, this suite stays green while the calendar bomb comes back. A shared constant/helper would be ideal longer-term; even a loud dual-update comment (or a tiny snapshot of the SQL fragments) would tighten the loop.

romandubovyi (commented):
Review complete. Solid calendar time-bomb fix for the BQ/Athena relative_date seed; no blocking issues. A few non-blocking notes on assertion strength, Athena anchoring coverage, and keeping the unit seed formulas in sync with the integration INSERT SQL.

romandubovyi (approved):
Approved. No blocking issues — the calendar anchor for row 6 and the date-independent seed invariants address the Jul 20 time-bomb cleanly. Non-blocking notes on the review stand as optional follow-ups.

PR #1442, opened 2026-07-21
fix(backend): keep NULL rows for negative report filters

romandubovyi:
## Summary

- Negative filter/slice operators (`neq`, `not_in`, `not_contains`, `not_regex`) no longer drop rows where the column is `NULL` (SQL three-valued logic).
- All storages (BigQuery, Athena, Redshift, Snowflake, Databricks) use one **portable** null-inclusive shape: `(col IS NULL OR …)` — no `IS DISTINCT FROM` (unsupported on Redshift). `not_in` shares the same shape via the base `inListSql` helper: `(col IS NULL OR col NOT IN (…))`.
- Same logic applies to metric filters (`renderHaving` shares `renderFilterFragment`).
- Docs cover all UI labels (`is not`, `≠` / `not equals`, `not on`, `not at`, `is none of`, `does not contain`, `does not match regex`), opt-out via **filter** `is not null` / `is not empty`, and the filter-vs-slice caveat for joined columns.
- Changeset documents the post-join `LEFT JOIN` blast radius and HAVING behavior.
- Integration operator matrices seed an all-NULL row and assert `neq` / `not_in` / `not_contains` / `not_regex` keep it.

## Review feedback addressed

| Comment | Resolution |
|---|---|
| Redshift `IS DISTINCT FROM` unsupported | Portable `(IS NULL OR …)` on all five engines |
| Two mechanisms for same intent | Single shape for all negative operators |
| Docs: slice cannot remove join-NULL | Docs say **filter** only; explain pre-join |
| Docs: only String labels | All type labels listed |
| No NULL in live seeds | All-NULL row + assertions per storage |
| HAVING silent change | Specs + changeset note |
| post-join LEFT JOIN blast radius | Documented in changeset + docs |
| `not_in` should agree with `is not` on NULL rows | Folded into this PR — `not_in` is now null-inclusive too, with unit + integration coverage |

## Test plan

- [x] Unit specs for all five clause renderers (including BigQuery DATE `neq` + CAST)
- [x] `not_in` null-inclusive specs (all five renderers) + integration NULL-row assertions
- [x] HAVING null-inclusive specs for metric filters
- [x] Blended builder specs updated for the new SQL shapes
- [x] Integration seeds + assertions for NULL row (cron / workflow_dispatch)
- [x] Prettier + eslint on changed files; clause-renderer suites green
- [x] Optional manual: report filter "country is not Canada" keeps customers with NULL country; add `is not null` to exclude them


laskevych (commented):
Reviewed the five clause renderers, the shared `SqlClauseRenderer` (WHERE + HAVING paths), the blended builders, the integration suites and the docs change. Four inline comments below; three more that have no line in the diff to hang on, plus what went well.

---

## 🟡 Important — a NULL-behavior fix with no NULL anywhere in the live test seeds

📍 `apps/backend/test/integration/{athena,bigquery,databricks,redshift,snowflake}.integration.ts`

The repo already has the right mechanism: each storage has a live operator matrix that runs `neq`, `not_contains` and `not_regex` against the real engine. But every seed is NULL-free — the suites literally assert it:

```ts
it('is_null on name → 0 rows (no NULLs in seed)', …)
```

So the unit specs prove *what string we render*, and nothing proves *that the engine accepts it* or *that NULL rows actually come back*. For a change whose entire point is NULL handling, that's the gap that matters — and it's exactly where the Redshift risk flagged inline would surface.

Suggest adding one NULL-valued row to each matrix seed plus three assertions per storage (`neq`, `not_contains`, `not_regex` all include it).

Worth noting for scheduling: `.github/workflows/test-integration.yml` runs on `workflow_dispatch` + cron `0 5,13 * * *`, not on PRs. A Redshift syntax failure would surface up to 12 h after merge, not in this PR's checks.

---

## 🔵 Question — HAVING / metric filters pick this up silently

📍 `apps/backend/src/data-marts/data-storage-types/utils/sql-clause-renderer.ts:140`

`renderHaving` reuses `renderFilterFragment`, so metric filters change too:

- `SUM(amount) ≠ 0` → `SUM(amount) IS DISTINCT FROM 0` — groups whose aggregate is `NULL` (all-NULL input) now pass where they were previously dropped.
- `not_contains` on a `STRING_AGG` metric emits the aggregate twice: `(array_join(array_agg(…)) IS NULL OR strpos(array_join(array_agg(…)), ?) = 0)`.

Both are reachable — `computeEffectiveType` maps `STRING_AGG` → STRING and `COUNT` → NUMBER, so the validator allows these operator/function pairs. Neither the changeset nor the docs mention metric filters, and there's no spec on the HAVING path for the changed operators. Deliberate, or collateral?

---

## 🔵 Question — `not_in` in #1435 has the identical defect

Open PR #1435 adds native `in` / `not_in`, rendered as a bare `NOT IN (…)`:

```ts
case 'not_in':
  return { sql: `${col} NOT IN (${rule.value.map(lit).join(', ')})`, params: [] };
```

`col NOT IN ('x','y')` is `UNKNOWN` when `col` is `NULL` — the same three-valued-logic drop this PR fixes for `neq`. Whichever of the two lands second should extend the fix to `not_in`, otherwise `is not Canada` and `is not one of [Canada]` will disagree on NULL rows. Worth an explicit note so it isn't lost between the branches.

---

## 👍 What went well

The parenthesization is right everywhere — each `(col IS NULL OR …)` is a self-contained group, so the `\n  AND ` join in `renderWhere` stays safe. Athena's positional-placeholder invariant still holds: one `?` per param in every changed fragment, and `countPositionalPlaceholders` correctly ignores the `', '` inside `array_join`. The BigQuery DATE `neq` + `CAST` spec is a nice catch.

I ran the affected suites on this branch — **24 suites / 663 tests pass** — and grepped the rest of the repo: no spec elsewhere still asserts the old `!=` / `<>` / `NOT CONTAINS` shapes.

apps/backend/src/data-marts/data-storage-types/redshift/services/redshift-clause-renderer.ts
```diff
@@ -106,8 +106,10 @@ export class RedshiftClauseRenderer extends SqlClauseRenderer {
     switch (rule.operator) {
       case 'eq':
         return { sql: `${col} = ${lit(rule.value)}`, params: [] };
+      // Null-inclusive: SQL `<>` drops NULLs (UNKNOWN). BI expectation is that
+      // "is not X" keeps rows where the column is missing — use IS DISTINCT FROM.
       case 'neq':
-        return { sql: `${col} <> ${lit(rule.value)}`, params: [] };
+        return { sql: `${col} IS DISTINCT FROM ${lit(rule.value)}`, params: [] };
```
laskevych: ## 🟡 Important — `IS DISTINCT FROM` is officially unsupported on Amazon Redshift

AWS lists `IS DISTINCT FROM` explicitly under [Unsupported PostgreSQL functions](https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-functions.html) for Redshift. That page also warns up front:

> some unsupported functions will not return an error when run on the leader node. The fact that these functions do not return an error in some cases should not be taken to indicate that the function is supported by Amazon Redshift.

It does parse on current Redshift builds, but it is undocumented and carries no compatibility guarantee — and after this PR **every** `neq` filter, slice and metric filter on Redshift depends on it.

The portable form is already used by this same PR for the other three operators:

```ts
case 'neq':
  return { sql: `(${col} IS NULL OR ${col} <> ${lit(rule.value)})`, params: [] };
```

Same result set, no reliance on an unsupported construct. BigQuery, Athena/Trino, Snowflake and Databricks all document `IS DISTINCT FROM` — Redshift is the only outlier of the five.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-clause-renderer.ts
```diff
@@ -77,8 +77,13 @@ export class BigQueryClauseRenderer extends SqlClauseRenderer {
     switch (rule.operator) {
       case 'eq':
         return { sql: `${col} = ${ph}`, params: [{ name: paramName, value: rule.value }] };
+      // Null-inclusive: SQL `!=` drops NULLs (UNKNOWN). BI expectation is that
+      // "is not X" keeps rows where the column is missing — use IS DISTINCT FROM.
       case 'neq':
-        return { sql: `${col} != ${ph}`, params: [{ name: paramName, value: rule.value }] };
+        return {
+          sql: `${col} IS DISTINCT FROM ${ph}`,
```
laskevych: ## 💡 Suggestion — one mechanism instead of two for the same intent

The PR solves the same problem two ways: `IS DISTINCT FROM` for `neq`, `(col IS NULL OR …)` for `not_contains` / `not_regex`.

Standardizing on the `(col IS NULL OR col <> v)` form would give all five renderers a single shape for "negative operator", make the intent legible at a glance, and remove the Redshift dialect dependency flagged on `redshift-clause-renderer.ts` in one move.

If `IS DISTINCT FROM` stays, these five renderer comments are a good place to record that it was verified against each engine — the surrounding comments in these files already follow that convention ("live-verified", "verified live: 3-arg CONCAT rejected, `||` chain accepted").

docs/getting-started/setup-guide/output-controls.md
```diff
@@ -64,6 +64,8 @@ Use it so rolling reports stay current without touching filter values manually.
 
 Multiple filters use `AND` logic — every condition must match for a row to appear. OR logic between filters is not supported.
 
+**NULL values and negative operators.** Operators that mean "is not / does not..." (`is not`, `does not contain`, `does not match regex`) **include** rows where the column is `NULL`. A missing value is treated as "not equal to X" rather than being dropped. The same applies to slices. To also remove missing values, add a separate filter or slice on the same column: `is not null` or `is not empty`.
```
laskevych: ## 🟡 Important — the NULL workaround does not work as a slice

> To also remove missing values, add a separate filter **or slice** on the same column: `is not null` or `is not empty`.

For a joined column that's incorrect. A slice runs *inside the joined subquery, before the `LEFT JOIN`*, so unmatched source rows still arrive with `NULL` on the joined columns — an `is not null` slice cannot remove them.

The Slices section further down this same page already states the correct rule:

> To also drop source rows with no match, add a **filter** on the same column (`is not null`).

Suggest narrowing the new sentence to "add a separate **filter**", with a note that for joined columns it must be a filter, not a slice.

laskevych: ## 📝 Docs — the note only names the String operator labels

`neq` reaches users under four different labels (`output-controls-operators.ts:62-89`): `is not` (String), `≠` / `not equals` (Number), `not on` (Date), `not at` (Time).

This paragraph lists only `is not`, `does not contain`, `does not match regex` — and it sits right under a table whose Number row reads `≠` and whose Date row reads `not on`. A reader filtering a date column won't connect the note to what they're looking at.

apps/backend/src/data-marts/data-storage-types/athena/services/athena-blended-query-builder.spec.ts
```diff
@@ -209,7 +209,7 @@ describe('AthenaBlendedQueryBuilder — output controls', () => {
         ],
       })
     );
-    expect(sql).toContain('WHERE main.a = ?\n  AND main.a != ?');
+    expect(sql).toContain('WHERE main.a = ?\n  AND main.a IS DISTINCT FROM ?');
```
laskevych: ## 🔵 Question — post-join negative filters on `LEFT JOIN`ed columns: intended blast radius?

This test is the post-join `neq` case, which is where the change bites hardest.

Blended reports join with `LEFT JOIN`, so a joined column carries two different kinds of `NULL`: "the row matched but the value is missing" and "the row never matched at all".

After this change a post-join `crm.status is not Churned` returns every source row with **no CRM match whatsoever**, not just rows with a genuinely empty status. Previously such a filter acted as an implicit inner join. The integration suites exercise this exact shape already (`SLICE + post-join role IS NOT NULL → order 2 eliminated`), so it's a live pattern, not a hypothetical.

Was the join-produced-NULL case considered separately from the missing-value case, or is "keep both" the deliberate call? Either way it seems worth a sentence in the changeset — it's the change users are most likely to notice.

apps/backend/src/data-marts/data-storage-types/redshift/services/redshift-clause-renderer.ts
romandubovyi: Good catch — agreed.

Switched all five renderers (including Redshift) off `IS DISTINCT FROM` to the portable form:

```sql
(col IS NULL OR col <> v)
```

Same result set, no dependency on an unsupported Redshift construct. `not_contains` / `not_regex` already used this pattern; `neq` now matches them.

apps/backend/src/data-marts/data-storage-types/bigquery/services/bigquery-clause-renderer.ts
romandubovyi: Took the suggestion: one shape everywhere.

All five engines now render negative operators as `(col IS NULL OR …)` — `neq`, `not_contains`, and `not_regex`. No `IS DISTINCT FROM` left in production SQL; comments on the `neq` cases note that the portable form is shared because Redshift has no `IS DISTINCT FROM`.

docs/getting-started/setup-guide/output-controls.md
romandubovyi: Fixed.

The note now says to add a separate **filter** (`is not null` / `is not empty`), and explicitly that for a joined column it must be a filter, not a slice — because a slice runs before the join and cannot drop rows that arrive with `NULL` from an unmatched join. Aligns with the Slices section further down the page.

romandubovyi: Updated.

The paragraph now names the “not equal” operator under every type label (`is not`, `≠` / `not equals`, `not on`, `not at`) plus `does not contain` and `does not match regex`, so it maps to the operator table above it (including Number / Date / Time rows).

apps/backend/src/data-marts/data-storage-types/athena/services/athena-blended-query-builder.spec.ts
romandubovyi: Deliberate: keep both kinds of NULL (data-missing and join-unmatched).

Rationale: we cannot distinguish them in SQL after a `LEFT JOIN`, and the product expectation for “is not X” is BI-style “missing ≠ X”. That does change post-join negative filters from an implicit inner join to keeping unmatched source rows.

Documented in:
- the changeset (post-join `LEFT JOIN` blast radius + opt-out)
- docs: add an `is not null` **filter** on the joined column to drop unmatched rows

Happy to revisit if product wants null-inclusive only for non-joined / pre-join paths — that would need a separate placement-aware rule.

romandubovyi:
### Replies to the review items without inline anchors

Thanks for the thorough pass — addressed in `91da397f7` (rewritten vs the earlier `IS DISTINCT FROM` approach).

#### 🟡 NULL-behavior fix with no NULL in live seeds
Done for all five storages.

Each operator-matrix seed now includes an all-NULL row, and the suites assert that `neq`, `not_contains`, and `not_regex` **keep** it (plus `is_null` / `is_empty` updates). Aggregation expectations that share the same seed (group-by / totals row counts) were updated where the extra NULL group or Row Count would change (Databricks / Snowflake / Redshift).

Noted on scheduling: integration still runs via cron / `workflow_dispatch`, not PR checks — portable `(IS NULL OR …)` also removes the Redshift-syntax risk that would only have shown up post-merge.

#### 🔵 HAVING / metric filters pick this up silently
Deliberate (shared `renderFilterFragment`).

- Documented in the changeset.
- Added unit specs on the HAVING path for null-inclusive `neq` (BigQuery) and `not_contains` (Athena), including the double-emitted aggregate expression form.

#### 🔵 `not_in` in #1435 has the identical defect
Agreed — out of scope for this PR.

Whichever lands second should apply the same `(col IS NULL OR col NOT IN (…))` (or equivalent) so `is not X` and `is not one of [X]` agree on NULL rows. Called out in the PR summary as a follow-up on #1435.

laskevych (approved):

PR #1443, opened 2026-07-21
feat(api): add project insight template client

max-voloshyn:
## Scope

- Add `client.insights.getTemplates({ limit, offset })` and exported project insight-template response types with runtime response validation.
- Add focused OpenAPI contract evidence for `GET /api/data-marts/insight-templates`.
- Document client usage and add the Project insight templates capability to API coverage.
- Append a consumer-focused Project insight templates section to the canonical API surface maintenance changeset.

## Coverage matrix

| Surface | Before | After |
| --- | --- | --- |
| Backend/OpenAPI | Operation, optional pagination, and typed `200` response existed without focused controller evidence | Focused controller OpenAPI contract test covers operation, queries, response container, optional nullable creator, Data Mart reference, and `canDelete` |
| `@owox/api-client` | No project insight-template listing abstraction or exported response types | `insights.getTemplates({ limit, offset })` with exported types and runtime response validation |
| Documentation | No project insight-template client usage | Consumer usage and bounded offset-pagination guidance for automation and agent discovery |
| Support matrix | Merged matrix listed the endpoint as unassessed, with 6 of 155 endpoints fully covered | Endpoint is dated `Covered · 2026-07-21`; 7 of 155 endpoints are fully covered and 148 remain unassessed |
| Changeset | Canonical file contained four prior capability sections | Prior sections preserved; distinct consumer-focused Project insight templates section appended |

## Audited commits

- Original audited base: `6e1a7dd110b18975f1c48835c616a08a18bf3631`
- Reconciled `main`: `724f62ccdf5b9ec46c742bc98c4c03685a3977f5`

## Verification

- Docs support-matrix validator and tests: 8 passed; 155 endpoints, 7 fully covered
- Backend focused OpenAPI contract test: 1 passed
- API-client focused tests: 5 passed
- API-client typecheck and focused backend/client ESLint passed
- Affected-file Prettier, affected Markdown lint, and `git diff --check` passed
- `changeset status --since origin/main --verbose` passed
- Merged support matrix and canonical workflow changeset were re-audited after reconciling `main`
- `docs/api/openapi.md` matches current `main`

## Exemptions

None.

## Blocker register

- `Audit all` failed on head `06e29779e269a896ceabe74800711ac09423172b` because the unchanged production dependency graph contains 12 advisories (6 moderate and 6 high), including fixes requiring breaking or out-of-range upgrades and a `sharp` advisory with no available fix ([run 29900123358](https://github.com/OWOX/owox-data-marts/actions/runs/29900123358)). This PR changes no dependency metadata or lockfile; remediation requires a separate repository dependency decision.

🤖 Codex (GPT-5.6 Sol High)


.changeset/api-surface-maintenance.md
```diff
@@ -32,3 +32,13 @@ project-wide Data Mart execution monitoring. It exports `OWOXProjectDataMartRuns
 `OWOXProjectDataMartRunStatus`, `OWOXProjectDataMartRunType`,
 `OWOXProjectDataMartRunTriggerType`, and `OWOXProjectRunHistoryOptions`. Existing viewer access
 and HTTP behavior are unchanged, and consumers can adopt the client method without a migration.
+
+## Add project insight-template discovery API client support
+
+`@owox/api-client` adds `insightTemplates.list({ limit, offset })` for authenticated,
```
max-voloshyn: `insightTemplates.list` -> `insights.getTemplates()` ?

packages/api-client/src/insight-templates.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    const client = new OWOXApiClient({ apiKey, fetchImpl });
+
+    const result = client.insightTemplates.list();
+    await expect(result).rejects.toMatchObject({
+      name: 'OWOXApiError',
+      message: 'OWOX Project Insight Templates API returned an unexpected response shape',
+      details: response,
+    });
+    await expect(result).rejects.toBeInstanceOf(OWOXApiError);
+  });
```
romandubovyi, line 127: 🟡 **Important:** Please add a test that rejects a malformed `createdByUser` (e.g. `{ userId: 42 }`), matching `project-run-history.test.ts`. The validator covers this path via `isOptionalNullableUser`, but nothing currently locks that behavior in.

apps/backend/src/data-marts/controllers/project-insight-templates.controller.openapi.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        'createdById',
+        'createdAt',
+        'modifiedAt',
+        'dataMart',
+        'canDelete',
+      ]),
+      properties: {
+        dataMart: {
+          $ref: '#/components/schemas/ProjectInsightTemplateDataMartRefResponseApiDto',
+        },
```
romandubovyi, line 124: 💡 **Suggestion:** Consider asserting the optional/nullable `createdByUser` shape (and the underlying user projection schema) the way `project-data-mart-runs.controller.openapi.spec.ts` does. The client documents and validates creator metadata, so locking that part of the OpenAPI contract would improve parity with the run-history evidence.

packages/api-client/src/insight-templates.ts
```diff
@@ -0,0 +1,117 @@
...
+    return false;
+  }
+
+  return (
+    typeof value.id === 'string' &&
+    typeof value.title === 'string' &&
+    typeof value.sourcesCount === 'number' &&
+    Number.isInteger(value.sourcesCount) &&
+    value.sourcesCount >= 0 &&
+    isOptionalNullableString(value.lastRenderedTemplateUpdatedAt) &&
```
romandubovyi: ⚪ **Nit:** `isOptionalNullableString(...) && value.lastRenderedTemplateUpdatedAt !== undefined` works, but a direct `(value === null || typeof value === 'string')` check would match the `string | null` type more clearly than “optional, then forbid undefined”.

romandubovyi (commented):
## Review summary

Solid api-surface maintenance PR that follows the established project run-history client pattern (typed client method, runtime validation, OpenAPI evidence, docs, coverage, changeset).

A few non-blocking notes on test/OpenAPI parity and a small readability nit in the validator.

max-voloshyn:
Engineer-review readiness is no longer effective, so I returned this PR to draft. The current Factory workflow now requires the repository-wide validated API support matrix, while target `main` still has the prior audit-table format and no matrix validator. That migration is outside this capability case and is now recorded in the blocker register; this case will re-audit its coverage entry after the sequencing is resolved.

All non-audit CI checks are green. `Audit all` remains red only on the previously recorded unchanged dependency advisories. Four inline review threads are open; the concrete test/schema feedback is understood, while the public client namespace suggestion needs confirmation before changing the API surface.

🤖 Codex (GPT-5.6 Sol High)

.changeset/api-surface-maintenance.md
max-voloshyn: `insightTemplates.list()` follows the existing resource namespaces (`dataMarts.list()`, `storages.list()`, and `destinations.list()`) and keeps templates distinct from rendered insights. Should this case intentionally adopt `insights.getTemplates()` as the public API despite that pattern, or keep the resource-oriented namespace?

🤖 Codex (GPT-5.6 Sol High)

romandubovyi (commented):
Review complete. The client mirrors the existing `runs.getHistory()` shape closely and the test coverage is solid. Three non-blocking notes: one on documenting the hard pagination cap, one on the growing duplication between `runs.ts` and `insight-templates.ts`, and one question about pagination metadata.

docs/api/api-client.md
```diff
@@ -105,6 +105,25 @@ Each run includes its Data Mart ID and title, creator metadata when available, e
 types, status, timestamps, and available logs, errors, metadata, and totals. This makes the method
 suitable for monitoring and automation without calling the HTTP endpoint directly.
 
+## List project insight templates
+
+Use `insightTemplates.list()` to discover reusable insight definitions across the Data Marts
+visible to the current project member. Pass optional `limit` and `offset` values to page through
+the project-wide list; the API defaults to at most 100 templates.
```
romandubovyi: 🟡 **Important:** This says the API "defaults to at most 100 templates", but `normalizeProjectListPagination` also *caps* `limit` at 100 and silently falls back to the defaults for invalid or non-positive values (`offset` → 0). A consumer calling `list({ limit: 500 })` gets 100 rows with no indication of truncation. Worth stating the hard cap explicitly here so automation callers know they must page rather than raise the limit.

packages/api-client/src/insight-templates.ts
```diff
@@ -0,0 +1,117 @@
...
+
+type InsightTemplatesRequester = {
+  getJson<T>(path: string, query?: Record<string, string>): Promise<T>;
+};
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+function isOptionalNullableString(value: unknown): boolean {
```
romandubovyi, line 45: 💡 **Suggestion:** `isOptionalNullableString`, the nullable-user predicate, the user type, the pagination options type, and the query-building block are verbatim copies of `runs.ts`. This is the second copy in a row, and given this is part of a series of API-surface PRs there will be a third. Consider extracting the shared pieces (`isOptionalNullableString`, a shared user ref type, and a `buildPaginationQuery(options)` helper) into a small internal validation module. Non-blocking — `isRecord` is already duplicated across eight files and is not worth touching here.

docs/api/api-client.md
```diff
@@ -105,6 +105,25 @@ Each run includes its Data Mart ID and title, creator metadata when available, e
...
+
+```ts
+const templates = await client.insightTemplates.list({ limit: 50, offset: 0 });
+
+for (const template of templates.insights) {
+  console.log(template.dataMart.title, template.title, template.canDelete);
```
romandubovyi, line 119: 🔵 **Question:** The response exposes only `insights`, with no `total` or `hasMore`, so callers have no reliable stop condition when paging — only the "fewer rows than `limit` means last page" heuristic. Was omitting pagination metadata a deliberate contract decision? Not asking to change the endpoint in this PR, but documenting the paging idiom here would help the automation/agent consumers this section targets.

.changeset/api-surface-maintenance.md
max-voloshyn: The main point is to have `.insights.` as entry point.
I agree that `list(...)` practice is important to follow when namespace is direct map to entity: dataMarts, storages, destinations.
BTW with that logic, may you change `.runs.getHistory` to `.runs.list` ?

max-voloshyn: Implemented as `client.insights.getTemplates()` in commit `741e7931`, with docs, coverage, and the changeset updated. I left the existing `runs.getHistory()` API unchanged: renaming a released public method is a separate compatibility and migration decision outside this endpoint case; it can be handled in a dedicated API-client case if desired.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/insight-templates.test.ts
max-voloshyn: Added a malformed `createdByUser` fixture with a numeric `userId`. The focused client suite now verifies that `getTemplates()` rejects it; all five client tests pass.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/controllers/project-insight-templates.controller.openapi.spec.ts
max-voloshyn: The contract test now asserts an optional and nullable `createdByUser` reference and verifies the complete `UserProjectionDto` schema. I added explicit OpenAPI decorators to the shared DTO so the referenced fields are represented accurately.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/insight-templates.ts
max-voloshyn: Simplified the validator to a direct required-nullable check: each timestamp must be either `null` or a string. The focused client suite passes with the clearer condition.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
max-voloshyn: Documents the endpoint defaults and hard bounds now: `limit` defaults to 10 and is capped at 100, while `offset` defaults to 0 and is capped at 100,000.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/insight-templates.ts
max-voloshyn: I kept the validation local to this module. Extracting a shared helper would require a cross-module refactor of existing public-client internals and is not needed to correct this endpoint behavior; that cleanup is better handled separately if more consumers accrue.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
max-voloshyn: Added a bounded offset-pagination idiom to the client documentation: choose a limit from 1 to 100, advance by the number returned, stop on a short page, and stop before the next offset would exceed 100,000. This documents the existing response contract without adding `total` or `hasMore`.

🤖 Codex (GPT-5.6 Sol High)

PR #1445, opened 2026-07-22
feat(mcp): query-parity output controls (filters, slices, aggregations, date buckets, sort, limit) in add_report/update_report

laskevych:
## Problem

PM report (Ruslan): a business user explores data with `query_data_mart` using filters, then asks to export it — and the created report silently contains the **whole raw data mart**. `add_report` only supported column selection (`fields`), so the last step of the core business flow dropped exactly the filters (and, per PM follow-up, the slices and aggregations) whose numbers the user had just seen.

Root cause: the domain supported every output control all along (`CreateReportCommand` carries `filterConfig`/`aggregationConfig`/`dateTruncConfig`/`sortConfig`/`limitConfig`; the create/update services thread and validate them; the UI report builder uses them) — only the MCP surface never exposed them, and the facade hard-coded them to `null`.

## Changes

- **`add_report`**: new optional `filters`, `slices`, `aggregations`, `date_buckets`, `sort`, and `limit` parameters — same shape and vocabulary as `query_data_mart`, so the agent copies them verbatim from the query the user was looking at. The tool description instructs agents to do exactly that. `limit` is deliberately NOT auto-copied from the interactive query (a silent 20-row export cap would be a trap); the description warns about it.
- **`update_report`**: the same parameters as replacements — `[]` removes a control, `null` removes the row limit, omitting keeps the current value; any of them counts as a valid "at least one change". `filters` + `slices` map into the one domain `FilterConfig` and therefore replace it as a unit.
- **Shared vocabulary**: `makeMcpFilterSchema` / `makeMcpAggregationSchema` / `makeMcpDateBucketSchema` / `makeMcpSortSchema` and the `mapMcp*` mappers are reused from `query-data-mart.input.ts` via the new `report-output-controls-input.ts`, so query and report vocabularies can never drift. Unsupported operators/functions/units and malformed operands become `BadRequestException` with a caller-vocabulary message, before the facade is touched.
- **Validation before side effects**: for Google Sheets every control goes through the existing pre-flight `validateForReport` **before** the sheet document is created (same as columns), so invalid input can't leave an orphaned Drive document. All controls work for every destination type.
- Docs (`docs/mcp.md`) and changeset updated.

## Notes for review

- Based on `main`, which already includes the multi-destination report tools (#1430). Task: #6767.
- The current operator vocabulary is main's. When #1435 (`in`/`not_in`, week/quarter presets) merges, report filters pick the new operators up automatically through the shared schema — no follow-up code needed.
- Commit 1 = row filters; commit 3 = the rest of the output controls (PM follow-up). Reviewing the combined diff is fine — commit 3 generalizes commit 1's helper (`report-filter-input.ts` → `report-output-controls-input.ts`).

## Testing

- Unit specs for every `mapReport*` mapper; cases in `add-report.tool.spec`, `update-report.tool.spec`, `mcp-reports.facade.impl.spec` (mapping incl. pre-join slices, [] → null clearing, null limit, unsupported operator/function/unit rejection before the facade, threading of every config into pre-flight + created/updated report, untouched controls preserved on update).
- `npx jest src/ee/mcp src/data-marts/facades` → 400 passed; eslint clean; no new typecheck errors.


amarchenk0 (commented):
[P1] Do not suggest repeated `eq` filters for an OR match

Report filters are combined with `AND`, so following this guidance for `ads` and `organic` produces `channel = 'ads' AND channel = 'organic'` and an empty export. This error is shown precisely when the client tries an unsupported `in`, so it steers MCP clients into a successfully created but incorrectly filtered report. Remove this fallback or state that the current vocabulary cannot express an OR list, and add a regression assertion for the error message.

apps/backend/src/ee/mcp/tools/report-output-controls-input.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+  throw err;
+}
+
+/**
+ * `slices` (pre-join) and `filters` (post-join) form ONE domain FilterConfig, so
+ * they replace it as a unit: providing either replaces every stored rule of both
+ * placements. Slices only apply to blended data marts, mirroring query_data_mart.
+ */
+export function mapReportFilters(
```
prokopetsv, line 54: 🟡 **Important** — replace-as-a-unit makes `update_report(filters=[...])` silently wipe stored slices (and vice versa).

Realistic flow: a blended report was created with `slices` + `filters`; later the user says "change the filter to purchases > 10". The agent calls `update_report` with only `filters`, gets a clean `status: "updated"`, and every pre-join slice is gone — the exported numbers silently diverge from what the user expects, which is exactly the failure mode this PR exists to close. The param descriptions do document it, but recovery depends on the agent re-sending slices it may have never seen: `get_data_mart_reports` doesn't return output controls, so within a fresh session the current slices are unknowable and unrestorable.

Since every stored `FilterRule` carries `placement`, the facade could partition `current.filterConfig` and replace per placement — `filters` omitted → keep current post-join rules, `slices` omitted → keep current pre-join rules. That matches the "untouched controls keep their current values" semantics the other four controls already have, and `[]` still works as an explicit clear per kind.

If replace-as-a-unit is intentional, a cheaper mitigation: have `update_report` return the resulting output controls so the agent (and the user) can see what was dropped.

prokopetsv (commented):
Reviewed the full diff plus the surrounding code (`OutputControlsValidatorService`, `CreateReportService`/`UpdateReportService`, the MCP error path, and the Looker/Sheets/email facade branches). No blockers.

**What holds up well:**

- The tool → facade → use-case flow is intact; no repositories or entities leak into the MCP layer, and the mappers keep an explicit transformation boundary.
- Validation-before-side-effect is real: the Sheets path pre-flights every new config via `validateForReport` before creating the Drive document, and `UpdateReportService` re-validates the fully merged state inside its transaction — so cross-control invariants (e.g. an aggregation referencing a column removed in the same call) are checked, not just per-parameter shapes.
- Reusing `makeMcp*Schema` factories from `query-data-mart.input.ts` genuinely prevents vocabulary drift, and the fresh-instance-per-use convention (the OpenAI `$ref` issue) was preserved.
- Test coverage is thorough: the `undefined` / `[]`→`null` conventions, pre-flight threading, untouched-control preservation on update, and pre-facade rejection of unsupported operators are all pinned.

**Findings** (inline, none blocking):

- 🟡 `update_report` with `filters` alone silently wipes stored slices (and vice versa) — a silent-divergence trap the agent cannot recover from, since no read tool exposes a report's current output controls.
- 🟡 Schema-level validation failures reach the agent as an opaque `Output controls validation failed` with the structured details dropped — `query_data_mart` translates these same errors, the report tools don't.
- 💡 The unsupported-operator message is now duplicated verbatim in two files.
- 💡 The `add_report` description contradicts itself about Looker Studio accepting output controls.

The changeset and docs updates look accurate to the implemented behavior (including the deliberate non-copying of `limit`).

apps/backend/src/ee/mcp/tools/add-report.tool.ts
```diff
@@ -137,12 +192,29 @@ export class AddReportTool implements McpToolDefinition<AddReportInput> {
...
+      slices,
+      aggregations,
+      date_buckets,
+      sort,
+      limit,
+      name,
+      message,
+    } = this.parseInput(input);
 
     const result = await this.reports.addReport({
```
prokopetsv: 🟡 **Important** — schema-level validation failures surface to the MCP client as an opaque string, with the actionable details dropped.

`OutputControlsValidatorService.validateForReport` throws `BadRequestException({ message: 'Output controls validation failed', details: { errors: [...] } })`. Nothing on the add/update tool path unwraps it, and a thrown error reaches the client as just `error.message` — so the codes and column names in `details.errors` are lost. Meanwhile `query_data_mart.mapError` translates these same errors into agent-actionable guidance (`FILTER_COLUMN_UNKNOWN` → "call get_data_mart_details_by_id…", `PRE_JOIN_FILTERS_REQUIRE_JOINED_DATA_MART` → "move these predicates to filters", …).

This was pre-existing for `fields`, but this PR multiplies the surface where it bites, and likely agent mistakes now dead-end without a self-correction path: `fields: ['*']` + `aggregations` (→ `AGGREGATION_REQUIRES_COLUMN_CONFIG` — an agent asked to "export everything, summed" will plausibly do this), slices on a non-blended mart, or a filter on a misspelled field. The PR's own goal is caller-vocabulary errors before/at the boundary — that currently only covers unsupported operators/operands, not schema-mismatch errors.

Suggest extracting the relevant `BadRequestException`-unwrapping branch of `query_data_mart.tool.ts#mapError` and applying it in both report tools (fine as a fast follow-up).

apps/backend/src/ee/mcp/tools/report-output-controls-input.ts
```diff
@@ -0,0 +1,91 @@
...
+ * - An empty array maps to `null` (control removed), which the update path uses
+ *   to clear a control.
+ * - Mapping failures — an operator/function/unit the internal engine does not
+ *   support, or a malformed operand — become BadRequestException so the MCP
+ *   client gets the caller-vocabulary message instead of an opaque server error.
+ */
+function wrapMappingError(err: unknown): never {
+  if (err instanceof UnsupportedOperatorError) {
+    throw new BadRequestException(
+      `Filter operator '${err.operator}' is not supported yet. Supported operators: ` +
```
prokopetsv: 💡 **Suggestion** — this message is duplicated verbatim from `query-data-mart.tool.ts#mapError` (the `UnsupportedOperatorError` branch). When #1435 lands and `in`/`not_in` become supported, both copies need the same edit; the hard-coded tail `(there is no 'in'/'not_in')` will go stale in whichever copy is missed — and note it renders for *every* unsupported operator (e.g. `in_next_n_days`), not just `in`/`not_in`. Worth extracting a single `unsupportedOperatorMessage(op)` helper next to `UnsupportedOperatorError` in `query-data-mart.input.ts` and using it in both places.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
```diff
@@ -75,7 +130,7 @@ type AddReportInput = z.infer<typeof inputSchema>;
...
+    'Create a report that exports a data mart to an existing destination (create one with add_destination if the project has none — check list_destinations first). Every destination type accepts the same optional output controls as query_data_mart — filters, slices, aggregations, date_buckets, sort — applied on each run: when the user asks to export numbers they saw in a query, copy those parameters verbatim from that query so the report matches what they saw. Google Sheets: a new Google Sheet is created automatically (an external Google Drive side effect) and the report is linked to it; returns the report and sheet links. Looker Studio: the report is created with default settings and accepts no extra parameters — not even name (Looker Studio reports carry no name), and each data mart + destination pair can have only one Looker report; the result includes instructions and a setup_guide_url to relay to the user, because dashboard data only flows after they connect Looker Studio to OWOX themselves. Email, Slack, Microsoft Teams, Google Chat: requires the message parameter; each report run sends the rendered message to the recipients or channels configured on the destination.';
```
prokopetsv: 💡 **Suggestion** — the description now contradicts itself for Looker Studio: it opens with "Every destination type accepts the same optional output controls…" but the Looker sentence still says "accepts no extra parameters — not even name". The code does thread and validate all output controls for the Looker path (`createReportWithConfig`), so the first claim is the true one. An agent reading the second sentence may wrongly refuse to pass filters when creating a Looker report — the exact hesitation this PR wants to remove. Reword to something like "accepts no Looker-specific parameters and no name".

apps/backend/src/ee/mcp/tools/report-output-controls-input.ts
laskevych: Agreed — implemented the partition option in e4e9265db. The facade now merges per placement (`mergeFilterConfig`): `filters` replaces only the stored post-join rules, `slices` only the pre-join ones, and `[]` clears just that kind — matching the "untouched controls keep their current values" semantics of the other controls. Rules without a `placement` (UI-created) count as post-join, matching how the query engine applies them. Param descriptions and docs updated; tests cover the mixed-stored-rules cases including clearing one kind while the other survives. Returning the resulting output controls from `update_report` (your cheaper mitigation) is still worth doing as a separate read-back improvement — noted for a follow-up.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
laskevych: Done in this PR rather than as a follow-up (e4e9265db): the `BadRequestException`-unwrapping branch of `query_data_mart.tool.ts#mapError` is extracted into `output-controls-error.mapper.ts` (`translateOutputControlsError`), the query tool now delegates to it, and both report tools wrap their facade calls with `rethrowTranslatedOutputControlsError` — so `FILTER_COLUMN_UNKNOWN`, `PRE_JOIN_FILTERS_REQUIRE_JOINED_DATA_MART`, the `*_NOT_SELECTED` family, `AGGREGATION_FUNCTION_NOT_ALLOWED_*`, and `INVALID_OPERATOR_FOR_TYPE` all surface the same agent-actionable guidance in add/update_report. Also added a new translation for the exact case you named: `AGGREGATION_REQUIRES_COLUMN_CONFIG` (fields ['*'] + aggregations) now tells the agent to replace ['*'] with the explicit field list. Covered by a dedicated mapper spec plus a tool-level test.

apps/backend/src/ee/mcp/tools/report-output-controls-input.ts
laskevych: Extracted in e4e9265db: `unsupportedOperatorMessage(op)` lives next to `UnsupportedOperatorError` in `query-data-mart.input.ts` and both call sites use it. The stale tail got a bigger fix than deduplication — per @amarchenk0's P1 the repeated-`eq` advice was actively wrong (filters AND together), so the single message now states that an OR match cannot be expressed and warns against emulating it, with regression assertions on both the query and report paths.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
laskevych: Fixed in e4e9265db — the Looker sentence now reads "accepts no Looker-specific parameters and no name (Looker Studio reports carry no name) — the shared output controls above DO apply", so the two claims no longer contradict and the agent won't refuse to pass filters on the Looker path.

laskevych:
@amarchenk0 [P1] fixed in e4e9265db. You're right that the fallback advice was actively harmful — filters combine with AND, so repeated `eq` on one field matches nothing, and the error fired exactly when a client attempted `in`. The message is now built in one place (`unsupportedOperatorMessage()` in `query-data-mart.input.ts`, shared by query_data_mart and both report tools — the query tool's copy had the same flaw) and states plainly that an OR match across several values cannot be expressed in the current vocabulary, with an explicit "do not emulate it with repeated 'eq' filters" warning. Regression assertions added on both paths: `report-output-controls-input.spec.ts` and `update-report.tool.spec.ts` assert the new wording and the absence of the old advice. (Once #1435 lands with native `in`/`not_in`, the single construction site is the only place to update.)

@prokopetsv thanks for the deep pass — all four findings addressed in e4e9265db, replied inline. Summary: per-placement filter replacement in `update_report` (slices survive a filters-only update and vice versa), validator errors translated for the report tools via the extracted `output-controls-error.mapper.ts` (including a new `AGGREGATION_REQUIRES_COLUMN_CONFIG` translation), the unsupported-operator message deduplicated into a single helper, and the Looker description contradiction reworded.

apps/backend/src/ee/mcp/tools/output-controls-error.mapper.ts
```diff
@@ -0,0 +1,134 @@
...
+      code: 'invalid_operator',
+      message: `Operator not valid for the field's type: ${detail}. ${
+        hasSlice
+          ? 'A slice filters a joined field on its PRE-JOIN values, so use an operator valid for that field’s "sliceType" from get_data_mart_details, not its blended-result "type". '
+          : ''
+      }Choose an operator that matches the field type and retry.`,
+    };
+  }
+
+  return null;
```
prokopetsv, line 148: 🟡 **Non-blocker** — this `return null` fallback still yields the opaque `Output controls validation failed` for validator codes the report tools can realistically hit but the query tool historically couldn't, so they were never in the copied branch set:

- `PRE_JOIN_FILTERS_REQUIRE_COLUMN_CONFIG` — reachable via `fields: ['*']` + `slices`, the exact sibling of the `AGGREGATION_REQUIRES_COLUMN_CONFIG` case this round *did* add (query_data_mart always has an explicit field list, so only reports can trigger it). Same agent-visible fix: use an explicit field list.
- `DATE_TRUNC_REQUIRES_DATE_COLUMN` / `DATE_TRUNC_COLUMN_IS_AGGREGATED` / `DATE_TRUNC_TIMEZONE_REQUIRES_TIMESTAMP` — a `date_bucket` on a non-date field is a plausible agent mistake.
- `HAVING_FILTER_NOT_AGGREGATED` — reachable when `update_report` replaces `aggregations` on a UI-created report carrying a HAVING rule; the agent cannot see that rule, so this one dead-ends hard (see the related comment on `mergeFilterConfig`).
- `OUTPUT_COLUMN_NAME_COLLISION`.

Rather than adding branches one code at a time, the robust minimal fix is to make the *fallback* informative: when `details.errors` exists but no specific branch matched, return a generic translation that appends the codes and columns (e.g. `Output controls validation failed: DATE_TRUNC_REQUIRES_DATE_COLUMN (channel)`). Then no current or future validator code is ever opaque, and specific branches stay reserved for cases needing real guidance.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -202,18 +211,55 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+    }
+    const currentRules = current ?? [];
+    const nextPreJoin =
+      preJoinReplacement !== undefined
+        ? (preJoinReplacement ?? [])
+        : currentRules.filter(rule => rule.placement === 'pre-join');
+    const nextPostJoin =
+      postJoinReplacement !== undefined
+        ? (postJoinReplacement ?? [])
+        : currentRules.filter(rule => rule.placement !== 'pre-join');
```
prokopetsv, line 258: 🟡 **Non-blocker** — `rule.placement !== 'pre-join'` puts UI-created HAVING rules (rules carrying `function`) into the replaceable post-join set, so `update_report(filters: [...])` or `filters: []` silently drops a report's post-aggregation constraints. This is the same cannot-see-cannot-restore trap this commit just fixed for slices, but stricter: the MCP filter vocabulary cannot even *express* a HAVING rule (no `function` key in `makeMcpFilterSchema`), so once dropped it is unrestorable through MCP by construction.

That said, blanket-preserving HAVING rules has its own trap: a preserved rule whose aggregation the agent removes would fail `HAVING_FILTER_NOT_AGGREGATED` with no MCP way to clear the rule — a stuck state. So I'd *not* simply mirror the slices fix. The lazy honest option: keep the current clearing semantics, but (1) say in the `filters` description that replacement also replaces UI-created post-aggregation (HAVING) constraints, and (2) translate `HAVING_FILTER_NOT_AGGREGATED` in the error mapper with "pass `filters: []` (or re-add the aggregation) to clear the stored HAVING constraint" — that turns both edges of this tradeoff into recoverable states instead of silent or opaque ones.

apps/backend/src/ee/mcp/tools/report-output-controls-input.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+  throw err;
+}
+
+/**
+ * `slices` (pre-join) and `filters` (post-join) form ONE domain FilterConfig, so
+ * they replace it as a unit: providing either replaces every stored rule of both
+ * placements. Slices only apply to blended data marts, mirroring query_data_mart.
+ */
+export function mapReportFilters(
```
prokopetsv, line 54: 🟡 **Non-blocker** — this docblock is now stale: "they replace it as a unit: providing either replaces every stored rule of both placements" was true in round 1 but contradicts the semantics this commit shipped. On update, `filters` and `slices` are mapped separately and replace only their own kind (`update-report.tool.ts` calls `mapReportFilters(undefined, filters)` / `mapReportFilters(slices, undefined)`; the facade merges per placement). Only `add_report` still uses the combined form — where "as a unit" is vacuous since there's no stored state.

A maintainer reading this comment as the contract of the shared mapper is one refactor away from re-introducing the cross-kind wipe the rest of this commit exists to prevent. Reword to describe what it actually does (combines the given kinds into domain rules; per-kind vs. combined replacement is the caller's choice), or split it into two single-kind helpers so the signature can't tell the old story.

prokopetsv (commented):
Re-reviewed at e4e9265 (round 2). All four round-1 findings are properly addressed and verified against the code, not just the descriptions:

- Per-kind filter replacement is implemented in the facade (`mergeFilterConfig`), with tests covering the cross-kind preservation and the placement-less (UI-created) rules-count-as-post-join case.
- The error translation was extracted into `output-controls-error.mapper.ts` and is now genuinely shared: `query_data_mart.mapError` delegates to it, and both report tools wrap their facade calls with `rethrowTranslatedOutputControlsError` — verified the mapper fires on the real exception shape `validateForReport` throws.
- `unsupportedOperatorMessage` is a single construction site now — and the author caught something I missed in round 1: the old guidance ("use multiple filters with 'eq'") was actively wrong, since filters combine with AND. The new message is correct and regression-tested.
- The `add_report` description no longer contradicts itself about Looker Studio.

Round 2 findings: **no blockers**, three 🟡 non-blockers, all in the newly added code (inline). The largest one: the new error mapper covers the codes the query tool historically hit, but several validator codes that only the *report* tools can reach still fall through to the opaque `Output controls validation failed` — the exact failure mode this round set out to close.

prokopetsv (approved):

apps/backend/src/ee/mcp/tools/output-controls-error.mapper.ts
laskevych: Implemented your "informative fallback" option in eb83b261e: when `details.errors` exists but no specific branch matched, the mapper now returns `output_controls_invalid` with the codes and columns appended (e.g. `Output controls validation failed: DATE_TRUNC_REQUIRES_DATE_COLUMN (channel), PRE_JOIN_FILTERS_REQUIRE_COLUMN_CONFIG`) plus a short retry hint — so no current or future validator code stays opaque, and dedicated branches remain reserved for cases needing real recovery guidance. Of the codes you listed, `HAVING_FILTER_NOT_AGGREGATED` did get a dedicated branch (it needs recovery guidance, per your other comment); the rest flow through the fallback. Covered by mapper spec cases for both the fallback and the empty-`errors` null path.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Took exactly your "lazy honest option" in eb83b261e — agreed that blanket-preserving invisible HAVING rules trades a silent drop for an uncleachable stuck state, which is worse. (1) The `filters` description now states that replacement also removes UI-created post-aggregation (HAVING) constraints, which cannot be expressed over MCP. (2) `HAVING_FILTER_NOT_AGGREGATED` gets a dedicated translation naming the orphaned rule (`SUM(revenue)`) and both ways out: re-add the matching aggregation, or pass `filters: []` to clear the stored constraint and then re-apply the wanted filters. Both edges of the tradeoff are now recoverable instead of silent or opaque. Clearing semantics themselves are unchanged.

apps/backend/src/ee/mcp/tools/report-output-controls-input.ts
laskevych: Fixed in eb83b261e — the docblock now states that the mapper only combines the given kinds into domain rules and that replacement scope is the caller's contract (add_report: combined, no stored state; update_report: one kind per call, facade merges per placement). Kept the single helper rather than splitting: the two call sites read clearly at the call site (`mapReportFilters(undefined, filters)` / `mapReportFilters(slices, undefined)`), and a split pair would duplicate the operand-error wrapping.

PR #1449, opened 2026-07-22
fix(web): preserve typed input and focus in connector settings

prokopetsv:
# Problem

Typing into a field on the connector "Configure Settings" step silently dropped characters, so text often appeared only on the second attempt, and pasted values could revert to their previous contents. Credential fields had a second, separate defect: they lost focus after the very first character, forcing the user to click back into the field to continue.

The dropped characters came from a state echo loop. `ConfigurationStep` keeps the configuration in local state and reports every change up to `ConnectorEditForm`, which stores that object as-is and passes it straight back down as the `initialConfiguration` prop. Both hydration effects list that prop as a dependency and react by calling `setConfiguration` with a full replacement built from it — but that copy is always at least one keystroke behind local state, so each keystroke raced and overwrote the one before it.

The lost focus was unrelated: in both `oneOf` renderers, `isSecret` was derived partly from the field's current value. The first keystroke flipped it from `false` to `true`, which swapped `ConfigurationStringField` for `ConfigurationSecretField` at the same position in the tree. React unmounts the input on a component-type change, so focus jumped to the surrounding tabs panel.

# Solution

For the echo loop, `ConfigurationStep` now records the exact object it last reported upward in `lastEchoedConfigRef`, and both hydration effects skip when `initialConfiguration` is that same reference. Reference equality is reliable here because the wizard stores the object without copying it, so identity cleanly separates "my own echo" from a genuine external change. Configurations arriving from outside still seed the form as before.

Deleting the second hydration effect was tried first and rejected. That effect is load-bearing: it restores the nested OAuth configuration at mount, and removing it wiped stored credentials when opening an existing connector and marked the form dirty on mount, producing a false unsaved-changes prompt. Guarding the effect keeps that behaviour intact while closing the race.

For the focus loss, `isSecret` is now derived from the specification alone, so the rendered component type is stable while typing. A new `hasStoredSecret` check (value equals `SECRET_MASK`, which is how the backend returns saved secrets) takes over the masked/readonly decision and the "Edit" button condition, so existing connectors still show stored secrets masked and unchanged.

Verified in the browser against the running app: typing the full alphabet at both 40ms and 0ms per character, and pasting, now yield the exact string in top-level and nested fields — previously these returned truncated values such as `ACEGKLPSTVXZ1348`. Regression checks confirm stored secrets remain masked and readonly, no unsaved-changes prompt appears on open, and focus survives the first keystroke.

# Changes

- Added `lastEchoedConfigRef` to `ConfigurationStep` to track the configuration object last reported to the wizard
- Guarded both hydration effects in `ConfigurationStep` against re-seeding from the component's own echo, fixing dropped characters and reverted pastes
- Changed `isSecret` in `ConfigurationOneOfRender` and `OauthRenderFactory` to be derived from the field specification only, so the rendered field component no longer changes identity mid-typing
- Added a `hasStoredSecret` check in both renderers to drive the masked/readonly state and the "Edit" button, preserving existing-connector behaviour
- Removed the now-unused `currentObjectValue` computation from `ConfigurationOneOfRender`

### DIFF
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
@@ -82,6 +82,10 @@ export function ConfigurationStep({
   const [configuration, setConfiguration] = useState<Record<string, unknown>>({});
   const initializedRef = useRef(false);
   const updatingFromParentRef = useRef(false);
+  // The exact object last reported upwards. The wizard stores it as-is and hands it
+  // straight back as `initialConfiguration`, so reference equality identifies our
+  // own echo and distinguishes it from a genuine outside change.
+  const lastEchoedConfigRef = useRef<Record<string, unknown> | null>(null);
   const [secretEditing, setSecretEditing] = useState<Record<string, boolean>>({});
 
   useEffect(() => {
@@ -94,7 +98,10 @@ export function ConfigurationStep({
   }, [connector]);
 
   useEffect(() => {
-    if (connectorSpecification) {
+    // `initialConfiguration` is a dependency so a genuinely new configuration seeds
+    // the form, but our own echo must not: re-seeding from it discarded characters
+    // as they were typed, which is why input appeared to lag a keystroke behind.
+    if (connectorSpecification && initialConfiguration !== lastEchoedConfigRef.current) {
       updatingFromParentRef.current = true;
 
       const config = migrateNestedConfigValuesToTopLevel(
@@ -129,7 +136,10 @@ export function ConfigurationStep({
       initializedRef.current &&
       initialConfiguration &&
       Object.keys(initialConfiguration).length > 0 &&
-      connectorSpecification
+      connectorSpecification &&
+      // Ignore our own echo. It is always at least one keystroke behind local
+      // state, so re-applying it here overwrote characters as they were typed.
+      initialConfiguration !== lastEchoedConfigRef.current
     ) {
       updatingFromParentRef.current = true;
       setConfiguration(
@@ -147,6 +157,7 @@ export function ConfigurationStep({
       !updatingFromParentRef.current &&
       Object.keys(configuration).length > 0
     ) {
+      lastEchoedConfigRef.current = configuration;
       onConfigurationChange?.(configuration);
     }
   }, [configuration, onConfigurationChange]);
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/ConfigurationOneOfRender.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/ConfigurationOneOfRender.tsx
@@ -1,10 +1,9 @@
 import { useState, useEffect, useMemo } from 'react';
 import { type ConnectorSpecificationResponseApiDto } from '../../../../../shared/api/types';
-import { AppWizardStepItemOneOf, AppWizardStepLabel } from '@owox/ui/components/common/wizard';
-import { configurationFieldRender } from './ConfigurationFieldRender';
+import { AppWizardStepItemOneOf } from '@owox/ui/components/common/wizard';
 import { TabsContent } from '@owox/ui/components/tabs';
-import { Button } from '@owox/ui/components/button';
 import { OauthRenderFactory } from './Oauth/OauthRenderFactory';
+import { NestedConfigurationField } from './NestedConfigurationField';
 import { SECRET_MASK } from '../../../../../../../shared/constants/secrets';
 
 interface ConfigurationOneOfRenderProps {
@@ -112,7 +111,6 @@ export function ConfigurationOneOfRender({
               />
             ) : (
               Object.entries(option.items).map(([itemName, itemSpec]) => {
-                const isFieldSecretEditing = fieldSecretEditing[itemName] ?? false;
                 const optionConfiguration =
                   typeof configuration[specification.name] === 'object' &&
                   configuration[specification.name] !== null
@@ -123,55 +121,22 @@ export function ConfigurationOneOfRender({
                   optionConfiguration[option.value] !== null
                     ? (optionConfiguration[option.value] as Record<string, unknown>)
                     : {};
-                // Spec-derived only: keying this off the config made it flip on the
-                // first keystroke, remounting the input and dropping focus.
-                const isSecret = itemSpec.attributes?.includes('SECRET') ?? false;
-                const hasStoredSecret =
-                  isSecret && isEditingExisting && nestedConfiguration[itemName] === SECRET_MASK;
                 return (
-                  <div key={itemName} className='mb-4'>
-                    <div className='flex items-center justify-between'>
-                      <AppWizardStepLabel
-                        htmlFor={itemName}
-                        required={itemSpec.required}
-                        tooltip={itemSpec.description}
-                        className='mb-2 justify-start'
-                      >
-                        {itemName}
-                      </AppWizardStepLabel>
-                      {/* Editing clears the mask, so `hasStoredSecret` alone would
-                          drop the button and leave no way to cancel. */}
-                      {(hasStoredSecret || isFieldSecretEditing) && (
-                        <Button
-                          variant='ghost'
-                          size='sm'
-                          type='button'
-                          onClick={() => {
-                            handleFieldSecretEditToggle(
-                              option.value,
-                              itemName,
-                              !isFieldSecretEditing
-                            );
-                          }}
-                        >
-                          {isFieldSecretEditing ? 'Cancel' : 'Edit'}
-                        </Button>
-                      )}
-                    </div>
-                    {configurationFieldRender({
-                      specification: { ...itemSpec, name: itemName },
-                      configuration: nestedConfiguration,
-                      onValueChange: (name, value) => {
-                        handleNestedValueChange(option.value, name, value);
-                      },
-                      flags: {
-                        isEditingExisting: hasStoredSecret,
-                        isSecret: isSecret,
-                        isSecretEditing: isFieldSecretEditing,
-                      },
-                      connectorName: connectorName,
-                    })}
-                  </div>
+                  <NestedConfigurationField
+                    key={itemName}
+                    itemName={itemName}
+                    itemSpec={itemSpec}
+                    nestedConfiguration={nestedConfiguration}
+                    isEditingExisting={isEditingExisting}
+                    isSecretEditing={fieldSecretEditing[itemName] ?? false}
+                    onSecretEditToggle={(name, enable) => {
+                      handleFieldSecretEditToggle(option.value, name, enable);
+                    }}
+                    onValueChange={(name, value) => {
+                      handleNestedValueChange(option.value, name, value);
+                    }}
+                    connectorName={connectorName}
+                  />
                 );
               })
             )}
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/ConfigurationOneOfRender.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/ConfigurationOneOfRender.tsx
@@ -118,16 +118,6 @@ export function ConfigurationOneOfRender({
               />
             ) : (
               Object.entries(option.items).map(([itemName, itemSpec]) => {
-                const currentObjectValue = (
-                  configuration[specification.name] &&
-                  typeof configuration[specification.name] === 'object'
-                    ? configuration[specification.name]
-                    : {}
-                ) as Record<string, unknown>;
-                const isSecret =
-                  Array.isArray(itemSpec.attributes) &&
-                  itemSpec.attributes.includes('SECRET') &&
-                  Object.keys(currentObjectValue)[0] === option.value;
                 const isFieldSecretEditing = fieldSecretEditing[itemName] ?? false;
                 const optionConfiguration =
                   typeof configuration[specification.name] === 'object' &&
@@ -139,6 +129,11 @@ export function ConfigurationOneOfRender({
                   optionConfiguration[option.value] !== null
                     ? (optionConfiguration[option.value] as Record<string, unknown>)
                     : {};
+                // Spec-derived only: keying this off the config made it flip on the
+                // first keystroke, remounting the input and dropping focus.
+                const isSecret = itemSpec.attributes?.includes('SECRET') ?? false;
+                const hasStoredSecret =
+                  isSecret && isEditingExisting && nestedConfiguration[itemName] === SECRET_MASK;
                 return (
                   <div key={itemName} className='mb-4'>
                     <div className='flex items-center justify-between'>
@@ -150,7 +145,7 @@ export function ConfigurationOneOfRender({
                       >
                         {itemName}
                       </AppWizardStepLabel>
-                      {isSecret && isEditingExisting && (
+                      {hasStoredSecret && (
                         <Button
                           variant='ghost'
                           size='sm'
@@ -174,7 +169,7 @@ export function ConfigurationOneOfRender({
                         handleNestedValueChange(option.value, name, value);
                       },
                       flags: {
-                        isEditingExisting: isEditingExisting,
+                        isEditingExisting: hasStoredSecret,
                         isSecret: isSecret,
                         isSecretEditing: isFieldSecretEditing,
                       },
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/ConfigurationOneOfRender.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/ConfigurationOneOfRender.tsx
@@ -12,8 +12,6 @@ interface ConfigurationOneOfRenderProps {
   configuration: Record<string, unknown>;
   onValueChange: (name: string, value: unknown) => void;
   isEditingExisting: boolean;
-  isSecretEditing: boolean;
-  onSecretEditToggle: (name: string, enable: boolean) => void;
   connectorName: string;
 }
 
@@ -22,8 +20,6 @@ export function ConfigurationOneOfRender({
   configuration,
   onValueChange,
   isEditingExisting,
-  isSecretEditing,
-  onSecretEditToggle,
   connectorName,
 }: ConfigurationOneOfRenderProps) {
   const detectSelectedOption = useMemo(() => {
@@ -112,8 +108,6 @@ export function ConfigurationOneOfRender({
                 configuration={configuration}
                 onValueChange={onValueChange}
                 connectorName={connectorName}
-                onSecretEditToggle={onSecretEditToggle}
-                secretEditing={{ [specification.name]: isSecretEditing }}
                 isEditingExisting={isEditingExisting}
               />
             ) : (
@@ -145,7 +139,9 @@ export function ConfigurationOneOfRender({
                       >
                         {itemName}
                       </AppWizardStepLabel>
-                      {hasStoredSecret && (
+                      {/* Editing clears the mask, so `hasStoredSecret` alone would
+                          drop the button and leave no way to cancel. */}
+                      {(hasStoredSecret || isFieldSecretEditing) && (
                         <Button
                           variant='ghost'
                           size='sm'
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/OauthRenderFactory.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/OauthRenderFactory.tsx
@@ -26,8 +26,6 @@ interface OauthRenderFactoryProps {
   configuration: Record<string, unknown>;
   onValueChange: (name: string, value: unknown) => void;
   connectorName: string;
-  onSecretEditToggle?: (name: string, enable: boolean) => void;
-  secretEditing?: Record<string, boolean>;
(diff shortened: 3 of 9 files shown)


apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
```diff
@@ -82,6 +82,10 @@ export function ConfigurationStep({
   const [configuration, setConfiguration] = useState<Record<string, unknown>>({});
   const initializedRef = useRef(false);
   const updatingFromParentRef = useRef(false);
+  // The exact object last reported upwards. The wizard stores it as-is and hands it
+  // straight back as `initialConfiguration`, so reference equality identifies our
+  // own echo and distinguishes it from a genuine outside change.
+  const lastEchoedConfigRef = useRef<Record<string, unknown> | null>(null);
```
laskevych, line 88: 🟡 **Important (non-blocking)** — implicit cross-component invariant

The dropped-character fix works only because the parent stores the echoed object *by reference* and hands it straight back: `ConnectorEditForm.handleConfigurationChange` (`ConnectorEditForm.tsx:252`) does `setConnectorConfiguration(configuration)` and passes it back as `initialConfiguration`, so `initialConfiguration === lastEchoedConfigRef.current` recognizes the echo.

The risk: if anyone later clones/normalizes it there — e.g. `setConnectorConfiguration({ ...configuration })` — every echo becomes a new reference, both hydration effects stop skipping, they re-seed on each keystroke, and this exact bug returns silently with no test to catch it.

This component has a clear comment; the parent has none. Suggest a short warning at `handleConfigurationChange` / the `initialConfiguration` prop ("stored & passed back by reference on purpose — see the echo guard in ConfigurationStep; do not clone"), and ideally the regression test noted below.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
```diff
@@ -147,6 +157,7 @@ export function ConfigurationStep({
       !updatingFromParentRef.current &&
       Object.keys(configuration).length > 0
     ) {
+      lastEchoedConfigRef.current = configuration;
```
laskevych, line 160: 💡 **Suggestion (non-blocking)** — add a regression test

Both bugs are subtle and easy to reintroduce (this hydration/echo race + the component-identity swap on `isSecret`). A small RTL test would lock them in: type a multi-character string and assert the full value survives (echo fix), and assert a secret field keeps the same input element / focus across the first keystroke (spec-derived `isSecret`). The PR description even has a ready assertion — the truncated `ACEGKLPSTVXZ1348`.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/OauthRenderFactory.tsx
```diff
@@ -235,10 +241,14 @@ export function OauthRenderFactory({
     return (
       <div className='space-y-4'>
         {Object.entries(option.items).map(([itemName, itemSpec]) => {
-          const isSecret = Array.isArray(itemSpec.attributes)
-            ? itemSpec.attributes.includes('SECRET') && nestedConfiguration[itemName] !== undefined
-            : false;
-          const isSecretEditing = secretEditing[specification.name] ?? false;
+          // Spec-derived only: keying this off the value made it flip on the first
+          // keystroke, swapping the rendered field component and remounting the input.
+          const isSecret = itemSpec.attributes?.includes('SECRET') ?? false;
```
laskevych: 💡 **Suggestion (non-blocking)** — duplicated nested secret-field logic

This block and the oneOf non-OAuth branch in `ConfigurationOneOfRender.tsx` (~128–172) are now near-identical: same spec-derived `isSecret`, same `hasStoredSecret` (`=== SECRET_MASK`) check, same `(hasStoredSecret || isSecretEditing)` button gate, and the same toggle that clears to `''` / restores `SECRET_MASK`. This PR extends that duplication.

Not blocking, but a shared `NestedSecretField` component (or a small hook returning `{ isSecret, hasStoredSecret, onToggle }`) would stop the two paths from drifting the next time secret/edit behavior changes.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/ConfigurationOneOfRender.tsx
```diff
@@ -174,7 +165,7 @@ export function ConfigurationOneOfRender({
                         handleNestedValueChange(option.value, name, value);
                       },
                       flags: {
-                        isEditingExisting: isEditingExisting,
+                        isEditingExisting: hasStoredSecret,
```
laskevych: 🔵 **Question (non-blocking)** — `isEditingExisting: hasStoredSecret`

Passing `hasStoredSecret` into the field's `isEditingExisting` flag reads odd next to the untouched top-level `ConfigurationItemRender`, which passes the *real* `isEditingExisting` into the same flag (same at `OauthRenderFactory.tsx:286`).

I believe it's correct — nested secrets aren't force-masked on mount the way top-level ones are (the masking loop in `ConfigurationStep` only walks top-level specs), so nested fields must key readonly/masked off an actually-stored value, and the backend returns stored secrets as exactly `SECRET_MASK` (`connector-secret.service.ts` → `maskRecursively`). Could you add a one-line comment (or rename to something like `isMaskedStoredSecret`) so the next reader doesn't "fix" the asymmetry and reintroduce a readonly-empty field? Just confirming the top-level/nested divergence is intentional.

laskevych (commented):
## Code review — no blocking issues

The fix looks correct. The character-dropping echo race and the focus loss are both addressed at the right level — the nested renderers are aligned with the already-correct top-level pattern where `isSecret` is spec-derived, so the rendered component type no longer changes mid-typing. I also checked the secret-masking assumption against the backend contract in `connector-secret.service.ts` (`maskRecursively` returns stored secrets as exactly `SECRET_MASK`), so `hasStoredSecret` is sound.

All comments below are **non-blocking** (robustness / suggestions / one question).

Neutral notes (no action needed):
- Changeset `'owox': minor` for a fix is consistent with this repo's convention (e.g. `6685-bq-table-column-name-collision.md` is a fix with `minor`).
- Moving OAuth secret-edit state from the container-keyed `secretEditing` to per-field `fieldSecretEditing` also quietly fixes the old behavior where the Edit button toggled *all* secret fields in a oneOf at once. Nice side effect.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
prokopetsv: Good catch — agreed, and thanks for spelling out the failure mode. The invariant
was documented on the consuming side only, which is exactly backwards: the
constraint lives where the tempting edit would be made.

Both suggestions applied.

**1. Warning at the parent** (`ConnectorEditForm.tsx:251`, directly above
`handleConfigurationChange`):

```js
// Stores the object by reference on purpose, and `initialConfiguration` hands that
// same reference back to ConfigurationStep. That step compares it against the last
// value it reported (see its `lastEchoedConfigRef`) to tell its own echo apart from
// a genuine outside change, and skips re-seeding the form on an echo.
// Do not clone or normalise here (`{ ...configuration }`, structuredClone, etc.):
// every echo would become a new reference, the step would re-seed on each keystroke,
// and typed characters would be dropped again.
```

Named the concrete clone forms rather than a general "don't change this", since
`{ ...configuration }` is the specific edit that looks harmless in review.

**2. Regression test** — new `ConnectorEditForm.test.tsx`. It renders the real
`ConnectorEditForm` with `ConfigurationStep` stubbed as a prop recorder, then
asserts `initialConfiguration` **is** (`Object.is`) the object just handed to
`onConfigurationChange`. Two cases: a single report, and repeated reports.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/OauthRenderFactory.tsx
prokopetsv: Agreed — and they'd already drifted, which settles it.

The two blocks were identical except for the label: the OAuth branch rendered
`{itemSpec.title ?? itemName}` while the oneOf branch rendered bare `{itemName}`.
That's visible today: on Google Ads the OAuth tab shows "Refresh Token" / "Client
ID", while the Service Account tab shows "ServiceAccountKey" / "DeveloperToken".
Exactly the drift you predicted, already shipped.

**Extracted** `NestedConfigurationField.tsx` (sibling of `ConfigurationOneOfRender`),
owning the whole nested-item render: `isSecret`, `hasStoredSecret`, the
`(hasStoredSecret || isSecretEditing)` gate, the Edit/Cancel button and the
`configurationFieldRender` call.

Went with a component over the hook option: a hook returning
`{ isSecret, hasStoredSecret, onToggle }` would have deduplicated the three
derived values but left the label, button and field-render JSX — most of the
duplication — copied in both places.

Callers now differ only in how they bind their option, which stays in their own
closures:

```jsx
// OauthRenderFactory — already scoped to one option
onSecretEditToggle={handleFieldSecretEditToggle}
onValueChange={handleNestedValueChange}

// ConfigurationOneOfRender — binds option.value
onSecretEditToggle={(name, enable) => handleFieldSecretEditToggle(option.value, name, enable)}
onValueChange={(name, value) => handleNestedValueChange(option.value, name, value)}
```

Named it `NestedConfigurationField` rather than `NestedSecretField`, since
non-secret items render through it too and the secret handling is one branch
inside it. Happy to rename if you prefer yours.

**Two things worth knowing:**

1. Merging the variants meant choosing one label. I took `itemSpec.title ?? itemName`,
   so the Service Account tab now shows "Service Account Key (JSON)" instead of
   "ServiceAccountKey". That's a user-visible improvement but a real behaviour
   change beyond the original scope — added a line to the changeset for it. Say
   the word if you'd rather keep it out of this PR.

2. Not a line-count win: −106 in the call sites, +38 back, +82 for the new
   component, so roughly +14 net. The value is the single source of truth for
   secret masking, not less code.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/ConfigurationOneOfRender.tsx
prokopetsv: Confirmed intentional, and your reasoning is exactly right. I re-derived it from
the code rather than just agreeing, and all three parts hold:

1. `ConfigurationStep`'s masking loop is a flat `forEach` over top-level
   `connectorSpecification` — it never descends into `spec.oneOf[].items`, so
   nested secrets are never pre-filled with `SECRET_MASK` on mount.
2. The flag's only effect is `isReadonly = isEditingExisting && !isSecretEditing`
   in `ConfigurationSecretField`, which drives `readOnly`, `disabled`, `type` and
   a hardcoded `SECRET_MASK` value.
3. `maskRecursively` does recurse into nested objects
   (`connector-secret.service.ts:441`), so a stored nested secret arrives as
   exactly `SECRET_MASK`. Confirmed against a live payload:
   `"AuthType":{"oauth2":{"AccessToken":"**********"}}`.

So passing the raw `isEditingExisting` here would render a never-set nested
secret `readOnly disabled` behind a fake mask with no way to type into it. That
is not hypothetical — it's a regression I introduced earlier in this PR and
caught in browser testing, which is what led to `hasStoredSecret`.

Added the comment at the flags site, where the "fix" would be attempted:

```js
flags: {
  // Intentionally `hasStoredSecret`, not `isEditingExisting` as the top-level
  // ConfigurationItemRender passes. This flag only drives
  // `isReadonly = isEditingExisting && !isSecretEditing` in ConfigurationSecretField,
  // and ConfigurationStep's masking loop walks top-level specs only, so nested
  // secrets are never pre-filled with SECRET_MASK on mount. Passing the raw prop
  // here would render a never-set nested secret readonly and disabled behind a
  // fake mask, with no way to enter it.
  isEditingExisting: hasStoredSecret,
  ...
```

Kept the name `hasStoredSecret` over `isMaskedStoredSecret`: the local is already
accurate, and the confusion isn't about what it holds but about why it feeds a
prop named `isEditingExisting` — which only prose can answer.

Worth noting the real root cause is that flag's name. It means "start masked and
readonly", not "we are editing an existing connector"; the top-level path only
gets away with the raw value because its secrets are force-masked at hydration.
Renaming it (`startsMasked`?) across `ConfigurationFieldRender`,
`ConfigurationSecretField`, `ConfigurationItemRender` and `ConfigurationListRender`
would remove the asymmetry rather than document it, but it touches the top-level
path this PR otherwise leaves alone. Happy to do it here or file it as a
follow-up — your call.

ESLint, `tsc --noEmit`, Prettier and the connector unit tests (12) all pass.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
prokopetsv: Added `ConfigurationStep.test.tsx`. I checked each test by reverting the fix it
targets rather than assuming it works — which turned out to matter, because the
two suggestions came out differently.

**The `isSecret` identity test works.** Reverting `isSecret` to value-derived:

```
× renders a nested secret field as the same element across the first keystroke
AssertionError: expected <input data-slot="input" …(10)></input> to be
                        <input data-slot="input" …(6)></input> // Object.is equality
```

It asserts both `getByLabelText(...)` returns the same node across the first
keystroke and that `document.activeElement` still points at it — the remount and
the focus loss, caught directly. This is the valuable one.

**The multi-character typing test does not catch the echo race.** I reverted the
echo guard from both hydration effects in `ConfigurationStep` and reran: **3/3
still passed.** So it would have shipped as coverage that doesn't exist.

The reason is structural, not fixable by typing more characters. `fireEvent`
wraps each event in `act()`, which flushes effects to completion between
keystrokes, so the echo round-trip always settles before the next keystroke and
the prop is never staler than local state. The `ACEGKLPSTVXZ1348` truncation
needs real event timing against React's passive-effect scheduling — I got it from
Playwright against the dev server, where it reproduces at 40ms/char and on paste.

So I kept the typing tests as controlled-input smoke coverage but labelled them
honestly, so nobody reads them as protecting the race:

```js
// The two typing tests below are smoke coverage for controlled-input wiring, not
// regression cover for the parent/child echo race. Verified by reverting the echo
// guard in ConfigurationStep: they still pass. `fireEvent` wraps every event in
// `act()`, which flushes effects to completion between keystrokes, so the prop can
// never be staler than local state and the race cannot occur here. The echo contract
// is pinned by reference in ConnectorEditForm.test.tsx instead.
```

The echo bug is instead covered from the angle that *is* deterministic — the
invariant it depends on. `ConnectorEditForm.test.tsx` (added for your other
comment) asserts the wizard hands back the exact object reference it received,
and fails on `setConnectorConfiguration({ ...configuration })`. That's the most
likely reintroduction path; the in-component guard itself is browser-covered.

Test files now: 131 passed, 901 tests. ESLint, `tsc --noEmit` and Prettier clean.

If you'd still like a unit-level test that reproduces the truncation, I think it
would need fake timers plus manual control of effect flushing to force the
interleaving — happy to attempt it, but it would be asserting against a
hand-built schedule rather than real behaviour, so I'd rather not pretend it
proves much.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
```diff
@@ -82,6 +82,10 @@ export function ConfigurationStep({
   const [configuration, setConfiguration] = useState<Record<string, unknown>>({});
```
romandubovyi, line 82: ⚪ **Non-blocking:** This component fixes a state-echo regression that is easy to reintroduce. Per `docs/contributing/testing.md`, frontend tests should be co-located with source files, but `ConfigurationStep.tsx` has no corresponding `ConfigurationStep.test.tsx`. Please add tests that simulate typing and assert that (a) `setConfiguration` is not overwritten by the parent echo and (b) secret inputs do not remount and lose focus after the first keystroke.

🤖 Generated with Kimi Code (Kimi K2.5)

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
```diff
@@ -94,7 +98,10 @@ export function ConfigurationStep({
   }, [connector]);
 
   useEffect(() => {
-    if (connectorSpecification) {
+    // `initialConfiguration` is a dependency so a genuinely new configuration seeds
```
romandubovyi, line 101: ⚪ **Non-blocking:** The echo guard relies on reference identity (`initialConfiguration !== lastEchoedConfigRef.current`). This is correct today because the parent stores the exact object received in `onConfigurationChange`, but it is brittle: if `ConnectorEditForm` ever clones the configuration before passing it back, the guard will silently fail and the dropped-character bug will regress. Consider adding a unit test that asserts the form does not re-seed when the parent echoes the same object back, or document this reference-equality contract explicitly.

🤖 Generated with Kimi Code (Kimi K2.5)

romandubovyi (approved):
Review complete. No blocking issues found.

- Verified `npm run type-check -w @owox/web` passes.
- Verified `npm test -w @owox/web` passes (903 tests).
- The fix correctly addresses the state-echo and focus-loss issues.
- Left two non-blocking suggestions: add regression tests and consider hardening the reference-equality echo guard.

🤖 Generated with Kimi Code (Kimi K2.5)

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
prokopetsv: `ConfigurationStep.test.tsx` is co-located next to `ConfigurationStep.tsx` — it
landed in `5bed3d61c`, probably just after you commented.

**(b) done and verified.** The test asserts the input is the same DOM node and
still holds focus across the first keystroke. Confirmed it fails when `isSecret`
is made value-derived again:

```
× renders a nested secret field as the same element across the first keystroke
AssertionError: expected <input …(10)> to be <input …(6)> // Object.is equality
```

**(a) can't be caught at this level.** I reverted the echo guard from both
hydration effects and reran: all tests still passed. `fireEvent` wraps each event
in `act()`, which flushes effects to completion between keystrokes, so the prop
is never staler than local state and the race cannot occur under RTL. The typing
tests are labelled in-file as controlled-input smoke coverage so nobody mistakes
them for echo-race cover.

The echo contract is pinned instead in `ConnectorEditForm.test.tsx`: it asserts
the wizard hands back the exact object reference, and fails on
`setConnectorConfiguration({ ...configuration })` — the realistic reintroduction
path. Reproducing the truncation itself needs real event timing; that part is
covered by Playwright against the dev server.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx
prokopetsv: Both already landed in `5bed3d61c`.

**Documented** at `ConnectorEditForm.tsx:251`, where the clone would be written —
an explicit "do not clone or normalise here (`{ ...configuration }`,
structuredClone, etc.)" with the consequence spelled out.

**Tested** in `ConnectorEditForm.test.tsx`, pinning exactly your scenario.
Verified it fails on `setConnectorConfiguration({ ...configuration })`:

```
AssertionError: expected { CustomerId: 'ABC' } to be { CustomerId: 'ABC' } // Object.is equality
```

One note: I pinned the contract from the parent side ("the wizard hands back the
same reference") rather than the child side ("the form does not re-seed"). The
child-side assertion isn't reachable under RTL — `fireEvent` wraps each event in
`act()`, which flushes effects between keystrokes, so re-seeding is unobservable;
confirmed by reverting the guard and watching the `ConfigurationStep` tests still
pass. Cloning in the parent is the realistic regression path anyway.

laskevych (approved):
## Re-review — all comments resolved, approving

Re-reviewed `5bed3d61c` and `a74eabd35`. Every comment from my first pass is addressed, and each fix was verified rather than just acknowledged:

1. **🟡 Reference invariant** — resolved the best possible way: the warning comment now sits exactly where the tempting clone would be written (`handleConfigurationChange`), and `ConnectorEditForm.test.tsx` pins the contract with `toBe` on the exact reported object, including across repeated reports.
2. **💡 Regression tests** — `ConfigurationStep.test.tsx` added. The focus/identity test is real regression cover (confirmed to fail when `isSecret` is reverted to value-derived). The typing tests are honestly documented as smoke-only — `act()` flushes effects between keystrokes, so the async race genuinely can't be reproduced in RTL; pinning the echo contract by reference at the parent instead is the right engineering compromise.
3. **💡 Duplication** — extracted shared `NestedConfigurationField`, used by both branches. Bonus: the extraction surfaced a drift the duplication had already caused (oneOf tabs showed raw field names like "ServiceAccountKey" while OAuth tabs showed titles); both now render `itemSpec.title ?? itemName`, documented in the changeset. I diffed the extraction line-by-line — that label change is the only intentional behavioral delta, and no e2e selectors rely on the raw names.
4. **🔵 `isEditingExisting: hasStoredSecret`** — confirmed intentional, now explained in place with the full three-part reasoning (the masking loop never descends into `oneOf` items, so the raw prop would render a never-set nested secret readonly behind a fake mask).

Verified in CI: both new test files ran and passed (3 + 2 tests, 132/132 web test files green), ESLint/Prettier/E2E green. The `Audit all` failure is pre-existing docs-site dependency vulnerabilities (astro/sharp/hono), unrelated to this PR.

LGTM — ready to merge. ✅

PR #1450, opened 2026-07-22
feat(data-marts): HTTP Data streaming on report level (#6759)

zapolsky:
## Summary

Adds `GET /api/external/http-data/reports/{reportId}.ndjson` — stream a saved report's data as NDJSON, applying the report's stored output controls (columns, filters, aggregations, date buckets, unique count, sort) without reconstructing query parameters by hand. The only accepted query parameter is an optional `limit` override; any other key returns `400`. Mirrored into `@owox/api-client` as `client.reports.traverseData(reportId, { limit })`.

Fibery: [HTTP Data on Report Level #6759](https://owox.fibery.io/Sprint_Work/Work_Item/HTTP-Data-on-Report-Level-6759)

## How it works

- Reuses `StreamHttpDataService` by extracting a shared `executeStream` tail driven by a discriminated-union stream plan (`data-mart | report`, guarded by `assertNever`). The report path builds the read plan from a loaded `Report`, which already satisfies `ReportLike`.
- Auth: `@Auth(Role.viewer())` + `Action.USE` on the report's Data Mart. Tenant isolation is enforced at the query level — `getByIdAndProjectId` inner-joins on `dataMart.projectId`, so a report from another project returns `404` before any reader work.
- The response carries an `x-owox-run-id` header; the run is a `HTTP_DATA` run tagged with the reportId, recording the inlined executed SQL under `additionalParams.httpData.executionSqlQuery` and grand totals in the run history.
- No migration — reuses the existing `DataMartRun.reportId` column.

## Design decisions & tradeoffs

1. **Access = `Action.USE` on the Data Mart, no per-report ACL.** A report is a saved query over its Data Mart, so anyone who can stream the Data Mart can stream any of its reports. This never grants wider data access than the Data Mart endpoint already does, and destination access is deliberately NOT required (the stream never touches the destination). *Tradeoff:* no per-report sharing granularity — a product boundary, not a code limitation.

2. **Failed-run recording aligned across BOTH HTTP Data endpoints.** A failure during query execution (blending/compose/reader, or a report's saved-column config drift) now records a `FAILED` `HTTP_DATA` run + `x-owox-run-id` for the report endpoint **and** the existing Data Mart endpoint. Previously the Data Mart endpoint recorded runs only for post-compose failures; this widens it for consistency with the DoD ("record a FAILED run on failure"). *Tradeoff:* the shipped Data Mart endpoint now logs slightly more failures in run history (additive; failed runs never bill). Pre-execution gates (report lookup, access, balance) still record no run — covered by boundary tests.

3. **MCP `query_data_mart` no longer records failed/cancelled runs** (a separate concern folded into this task). *Tradeoff:* this reduces observability of MCP query timeouts/aborts in run history — noted for the MCP timeout-resilience follow-up. It does not affect either HTTP Data endpoint (both still record `FAILED` runs).

4. **`?limit` is the only query override (v1).** Every other output control comes from the saved report config. *Tradeoff:* no ad-hoc filter/sort/aggregation override on the report route — this matches the request's intent ("without assembling parameters by hand") and is deferred.

5. **Executed SQL is captured only when the report applies output controls or blending.** A plain column-only report streams the Data Mart's own query and records no separate `executionSqlQuery` (`hasOutputControls` intentionally excludes `columnConfig`). The base query still survives in `definitionRun` plus the recorded columns.

6. **Grand totals are computed for any report with an eligible metric** (per the totals product decision), not only explicitly aggregated ones. *Tradeoff:* a plain report stream with a numeric field runs a separate best-effort totals query (extra cost/latency) — this is inherited from the Data Mart endpoint's existing behavior and is best-effort, so it never breaks the stream.

## API client

- New `client.reports.traverseData(reportId, { limit })` returning the NDJSON traversal.
- The shared NDJSON primitive was moved to a neutral `HttpNdjsonTraversal`, with `DataMartDataTraversal` kept as a `@deprecated` back-compat alias — report-stream errors now carry `reportId` (not `dataMartId`).

## Testing

- Backend unit: 7 suites / 145 (stream & query-data-mart services, mappers, schemas, validator, OpenAPI spec).
- Backend e2e: `http-data.e2e-spec` + new `http-data-report-cross-project.e2e-spec` (tenant isolation via `MultiTenantIdpProvider`).
- `@owox/api-client`: 44 · `@owox/ctl`: 40 · `nest build` + `tsc -b` clean.
- Explicit coverage: config-drift FAILED run tagged reportId, balance-gate no-run boundary, `?limit` override/fallback/null precedence, uniqueCount threading, best-effort SQL-inline throw, cross-project `404`, aggregated totals in run history.

## Deployment

No migration, no new env vars. The endpoint is additive; rollback simply drops the route.

### DIFF
--- a/.changeset/6759-http-data-report-level.md
+++ b/.changeset/6759-http-data-report-level.md
@@ -1,0 +1,11 @@
+---
+'owox': minor
+---
+
+# HTTP Data on report level
+
+Adds `GET /api/external/http-data/reports/{reportId}.ndjson` — stream an existing report's data as NDJSON, applying the report's saved filters, aggregations, date buckets, unique count, and sort, without reconstructing query parameters by hand. An optional `limit` query parameter overrides the report's saved limit. The [`@owox/api-client`](../docs/api/api-client.md) `reports.traverseData()` method exposes the same endpoint.
+
+The response carries an `x-owox-run-id` header (consistent with the Data Mart HTTP Data endpoint); on a successful stream the referenced run is a `HTTP_DATA` run tagged with the reportId. When the report applies a filter, aggregation, sort, date bucket, row limit, or unique count — or blends fields across Data Marts — the run also records the exact executed SQL via `additionalParams.httpData.executionSqlQuery`, discoverable through the run history endpoint. (A plain column selection with none of those controls streams the Data Mart's own query and records no separate executed SQL.) Grand totals — one per selected metric (every numeric field, plus non-numeric fields eligible for Count / Count Unique) aggregated over the full result — are recorded in the run history whenever the report selects an eligible metric, computed as a separate best-effort query.
+
+Failed MCP `query_data_mart` calls create no run-history entry; both HTTP Data endpoints (Data Mart and report level) record a `FAILED` run for any failure encountered during query execution.
--- a/apps/backend/src/data-marts/controllers/external/http-data.controller.ts
+++ b/apps/backend/src/data-marts/controllers/external/http-data.controller.ts
@@ -4,7 +4,7 @@ import type { Response } from 'express';
 import { Auth, AuthContext, AuthorizationContext, Role } from '../../../idp';
 import { HttpDataMapper } from '../../mappers/http-data.mapper';
 import { StreamHttpDataService } from '../../use-cases/stream-http-data.service';
-import { StreamHttpDataSpec } from '../spec/external/http-data.api';
+import { StreamHttpDataSpec, StreamHttpReportDataSpec } from '../spec/external/http-data.api';
 
 @Controller('external/http-data')
 @ApiTags('HTTP Data')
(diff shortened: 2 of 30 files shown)


amarchenk0 (changes requested):
### P1

`apps/backend/src/data-marts/use-cases/query-data-mart.service.ts:309`

The deleted catch means every failed or cancelled `query_data_mart` call—DWH read errors, composition errors, timeouts, or aborts—now skips `recordMcpQueryRun`. These outcomes were previously persisted with their query metadata and `FAILED`/`CANCELLED` state, and Run History no longer exposes them. This is unrelated to the report stream and removes the audit trail; retain the failure-recording path while continuing to skip consumption billing.

zapolsky:
@amarchenk0 Re: P1 on `query-data-mart.service.ts:309` — this is intentional per the task's DoD, not incidental to the report stream. The #6759 requirements explicitly call for **not** recording run-history entries for failed/cancelled MCP `query_data_mart` calls.

Both HTTP Data endpoints (Data Mart and report level) still record a `FAILED` run on failure — only the MCP `query_data_mart` path omits them, by design. Consumption billing is independent of this (registered only on the success path), so nothing billing-related is coupled to the change.

Your point about losing the audit trail for MCP timeouts/aborts is fair. I'd rather close that observability gap with the MCP timeout-resilience follow-up (where timeout/abort handling is being reworked end-to-end) than reinstate failure-recording here, to stay aligned with the DoD.

Keeping the current behavior for this PR.

laskevych (approved):
Everything's great! I suggest you take a look at the **non-blocked** comments 👍

PR #1451, opened 2026-07-22
test(api): cover auth context surface

max-voloshyn:
## Scope

- Add a focused generated OpenAPI contract test for `GET /api/auth/context`, including its exact operation identity, Authentication tag, authorization headers, and success schema.
- Register and validate the exact Swagger operation and public API-client guide targets.
- Mark the endpoint covered in both dimensions, recalculate the support-matrix totals, and append the consumer-facing Auth context release note to the canonical workflow changeset.

## Coverage matrix

| Dimension | Before | After |
| --- | --- | --- |
| OpenAPI | Unassessed | [Covered](https://app.owox.com/api/swagger-ui#/Authentication/AuthContextController_getContext) · 2026-07-22 |
| API client | Unassessed | [Covered](https://github.com/OWOX/owox-data-marts/blob/api-surface-maintenance/auth-context/docs/api/api-client.md#get-auth-context) · 2026-07-22 |

Audited base: `b097a3decfacebccfece7f9061a886850c4694f1`

## Verification

- Backend focused controller and generated OpenAPI specs: 2 suites, 6 tests passed.
- API-client focused contract tests: 19 tests passed, covering API-key-derived auth, response-shape validation, and secret non-exposure.
- Coverage validator: 18 tests passed; 155 endpoints, 9 fully covered.
- Docs build: 145 pages built; all internal links valid, including `#get-auth-context`.
- Backend/docs ESLint, Prettier, Markdownlint, `git diff --check`, and the new spec's file-scoped TypeScript check passed.
- Changesets validation against `origin/main` passed and includes the fixed release group.
- Independent full-diff review found no Critical, Important, or Minor issues.
- `docs/api/openapi.md` is identical to current `origin/main`.

## Exemptions

None.

## Blocker register

- Required `Audit all` check: `npm audit --omit=dev` reports 12 advisories from the unchanged dependency lockfile, including one with no fix. Dependency remediation is outside this API-surface case, so the PR remains draft while the heartbeat monitors repository-level resolution.
- Non-blocking local note: a broader test-inclusive backend TypeScript probe reports existing errors only in unchanged files; the changed spec passes its scoped TypeScript check.

🤖 Codex (GPT-5.6 Sol High)

PR #1452, opened 2026-07-22
fix(api): centralize OpenAPI auth contract

max-voloshyn:
## Scope

- Model `X-OWOX-Authorization` as the shared OpenAPI security scheme for protected backend operations.
- Derive conditional `X-OWOX-Api-Key-Id` metadata from the existing `@RejectApiKeyAuth()` eligibility boundary, while preserving the required unauthenticated exchange header.
- Replace the inline `GET /api/auth/context` response schema with `AuthContextResponseApiDto` and a generated component reference.
- Keep the Support Matrix operation identity and coverage dates unchanged; `docs/api/openapi.md` remains byte-for-byte identical to `main`.

## Coverage matrix

| Surface | Before | After |
| --- | --- | --- |
| Protected authentication metadata | Standard bearer metadata plus endpoint-local X-OWOX header copies | One X-OWOX security scheme with generated conditional key-ID metadata |
| API-key-rejected operations | Runtime rejection could diverge from generated metadata | Existing rejection decorator owns runtime and OpenAPI eligibility |
| `GET /api/auth/context` response | Complete inline object schema | Named `AuthContextResponseApiDto` component reference |
| Support Matrix | Covered since 2026-07-22 | Unchanged operation link and covered-since date |

## Audited base

`a64abfd6fc4a4d8803280d95b0e442b82605a659`, the merge commit of PR #1451.

## Verification

- 9 focused backend suites, 33 tests, sequential with `--runInBand`
- changed-file TypeScript check using the case-local runtime configuration
- affected-file ESLint and Prettier checks
- Support Matrix validator: 17 tests; 155 endpoints and 9 fully covered
- canonical Changesets validation and Markdown lint
- independent full-diff review plus delta review: no remaining findings
- `docs/api/openapi.md` equality check against current `origin/main`
- review follow-up: 2 focused suites, 8 tests, changed-file typecheck, affected-file lint and formatting
- final-head CI: 17 successful checks; the known `Audit all` baseline failure is recorded below

## Exemptions

None.

## Blocker register

- Review blockers: none.
- Merge blocker: engineer reapproval remains pending for final head `18d762d06`.
- Merge blocker: `Audit all` reports the 12 dependency findings already present in the audited `package-lock.json`; this change does not modify dependency manifests or lockfiles.
- Merge blocker: repository-wide backend `tsc --noEmit` currently reports pre-existing errors in unrelated test files; the changed-file typecheck is clean.

🤖 Codex (GPT-5.6 Sol High)

### DIFF
--- a/apps/backend/src/config/swagger.config.ts
+++ b/apps/backend/src/config/swagger.config.ts
@@ -1,18 +1,99 @@
 import { INestApplication } from '@nestjs/common';
-import { SwaggerModule, DocumentBuilder } from '@nestjs/swagger';
+import { DocumentBuilder, OpenAPIObject, SwaggerModule } from '@nestjs/swagger';
+import type {
+  OperationObject,
+  ParameterObject,
+  ReferenceObject,
+} from '@nestjs/swagger/dist/interfaces/open-api-spec.interface';
+import {
+  OWOX_API_KEY_AUTH_EXTENSION,
+  OWOX_API_KEY_ID_DESCRIPTION,
+  OWOX_API_KEY_ID_HEADER,
+  OWOX_AUTHORIZATION_DESCRIPTION,
+  OWOX_AUTHORIZATION_HEADER,
+  OWOX_AUTHORIZATION_SECURITY_SCHEME,
+} from '../idp/openapi/authentication.openapi';
 
-export function setupSwagger(app: INestApplication, path: string): void {
-  const config = new DocumentBuilder()
+const HTTP_METHODS = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'] as const;
+
+export function buildSwaggerConfig() {
+  return new DocumentBuilder()
     .setTitle('OWOX Data Marts API')
     .setDescription('REST API used by frontend clients and service integrations.')
     .setVersion('1.0')
+    .addApiKey(
+      {
+        type: 'apiKey',
+        in: 'header',
+        name: OWOX_AUTHORIZATION_HEADER,
+        description: OWOX_AUTHORIZATION_DESCRIPTION,
+      },
+      OWOX_AUTHORIZATION_SECURITY_SCHEME
+    )
     .build();
+}
 
-  const document = SwaggerModule.createDocument(app, config);
+export function setupSwagger(app: INestApplication, path: string): void {
+  const document = createSwaggerDocument(app);
   SwaggerModule.setup(path, app, document, {
     useGlobalPrefix: true,
     jsonDocumentUrl: 'openapi.json',
     raw: ['json', 'yaml'],
     yamlDocumentUrl: 'openapi.yaml',
   });
+}
+
+export function createSwaggerDocument(app: INestApplication): OpenAPIObject {
+  const document = SwaggerModule.createDocument(app, buildSwaggerConfig());
+
+  for (const path of Object.values(document.paths)) {
+    for (const method of HTTP_METHODS) {
+      const operation = path[method];
+      if (operation && usesOwoxAuthorization(operation)) {
+        applyOwoxAuthenticationContract(operation);
+      }
+    }
+  }
+
+  return document;
+}
+
+function usesOwoxAuthorization(operation: OperationObject): boolean {
+  return Boolean(
+    operation.security?.some(requirement => OWOX_AUTHORIZATION_SECURITY_SCHEME in requirement)
+  );
+}
+
+function applyOwoxAuthenticationContract(operation: OperationObject): void {
+  const extensionOperation = operation as OperationObject & Record<string, unknown>;
+  const rejectsApiKeyAuth = extensionOperation[OWOX_API_KEY_AUTH_EXTENSION] === false;
+  const parameters = (operation.parameters ?? []).filter(
+    parameter =>
+      !isHeaderParameter(parameter, OWOX_AUTHORIZATION_HEADER) &&
+      !isHeaderParameter(parameter, OWOX_API_KEY_ID_HEADER)
+  );
+
+  if (!rejectsApiKeyAuth) {
+    parameters.push({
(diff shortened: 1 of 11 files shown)


apps/backend/src/idp/decorators/reject-api-key-auth.decorator.ts
```diff
@@ -1,5 +1,11 @@
...
+import { ApiExtension } from '@nestjs/swagger';
+import { OWOX_API_KEY_AUTH_EXTENSION } from '../openapi/authentication.openapi';
 
 export const REJECT_API_KEY_AUTH_METADATA = 'rejectApiKeyAuth';
 
-export const RejectApiKeyAuth = () => SetMetadata(REJECT_API_KEY_AUTH_METADATA, true);
+export const RejectApiKeyAuth = () =>
+  applyDecorators(
+    SetMetadata(REJECT_API_KEY_AUTH_METADATA, true),
+    ApiExtension(OWOX_API_KEY_AUTH_EXTENSION, false)
```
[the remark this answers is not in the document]
max-voloshyn: Validated against the installed `@nestjs/swagger` 11.2.0 implementation. `ApiExtension` does not leave this metadata only on the controller: when applied as a class decorator, it enumerates the controller's route handlers and calls `applyExtension` on each method. `SwaggerExplorer` then reads that method metadata.

The focused generated-document test uses `@RejectApiKeyAuth()` on the controller class and independently verifies that `X-OWOX-Api-Key-Id` is absent from the resulting operation. It passes against the installed implementation, so no handler-level duplication or document-pass controller lookup is needed here.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/idp/controllers/auth-context.controller.openapi.spec.ts
```diff
@@ -33,54 +33,21 @@ describe('AuthContextController OpenAPI', () => {
     await app.close();
   });
 
-  it('documents the auth context operation, headers, and response schema', () => {
+  it('keeps the operation identity used by the Support Matrix', () => {
```
romandubovyi, line 36: ⚪ **Non-blocking:** This test builds the document with `SwaggerModule.createDocument`, so it never exercises `createSwaggerDocument()` and cannot assert that the canonical `X-OWOX-Authorization` / `X-OWOX-Api-Key-Id` contract is applied to `GET /api/auth/context`. Consider using `createSwaggerDocument(app)` here or adding a dedicated assertion in `auth.decorator.openapi.spec.ts` for this operation.

🤖 Generated with Kimi Code

apps/backend/src/config/swagger.config.ts
```diff
@@ -1,18 +1,99 @@
...
+      description: OWOX_API_KEY_ID_DESCRIPTION,
+      schema: { type: 'string' },
+    });
+  }
+
+  operation.parameters = parameters.length > 0 ? parameters : undefined;
+  delete extensionOperation[OWOX_API_KEY_AUTH_EXTENSION];
+}
+
+function isHeaderParameter(
```
romandubovyi, line 90: ⚪ **Non-blocking:** `isHeaderParameter` only handles inline `ParameterObject`; a $ref parameter that resolves to an `X-OWOX-Authorization` header would not be filtered out, producing a duplicate next to the global security scheme. Consider documenting this assumption or handling `ReferenceObject` if the project ever uses referenced header parameters.

🤖 Generated with Kimi Code

apps/backend/src/idp/dto/presentation/auth-context-response-api.dto.ts
```diff
@@ -0,0 +1,31 @@
...
+  @ApiPropertyOptional({ nullable: true })
+  email?: string;
+
+  @ApiPropertyOptional({ nullable: true })
+  fullName?: string;
+
+  @ApiPropertyOptional({ nullable: true })
+  avatar?: string;
+
+  @ApiPropertyOptional({ enum: ['admin', 'editor', 'viewer'], isArray: true })
```
romandubovyi: ⚪ **Non-blocking:** The `roles` field hardcodes `[admin, editor, viewer]` while the source of truth is `RoleEnum` from `@owox/idp-protocol`. If the enum values diverge in the future, this DTO will publish stale metadata. Consider using `RoleEnum` directly in the `@ApiPropertyOptional` decorator.

🤖 Generated with Kimi Code

romandubovyi (approved):

apps/backend/src/config/swagger.config.ts
max-voloshyn: Confirmed that the repository currently emits these authentication inputs only as inline header parameters; there are no referenced header parameters to dereference here. The filter deliberately leaves arbitrary `$ref` parameters untouched because safely removing one would require resolving its component and ownership semantics. If referenced authentication headers are introduced later, that new representation should add component-aware normalization and its own generated-document test.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/idp/dto/presentation/auth-context-response-api.dto.ts
max-voloshyn: Fixed in `18d762d06`: the presentation DTO now derives its Swagger enum values from `RoleEnum.options` while retaining the `Role` property type. Focused OpenAPI tests, changed-file typechecking, lint, formatting, and an independent delta review are green.

🤖 Codex (GPT-5.6 Sol High)

max-voloshyn:
The review follow-up is now on head `18d762d06`. It removes the duplicated role-value list by deriving the auth-context presentation enum from the shared IDP protocol source of truth.

Because this commit was added after the approval on `4ada5017ee`, the effective ready-to-merge state no longer applies until the final head is reapproved and its checks settle. The pull request remains ready for review and has not been converted back to draft.

🤖 Codex (GPT-5.6 Sol High)

PR #1455, opened 2026-07-23
Google Chat Destination integration 

AnastasiaKhukhra:
Google Chat Destinations now offer Google Chat API and Channel Email delivery methods. Existing
email-based destinations keep their current method, while new destinations default to direct API
delivery through a space-specific incoming webhook. API messages are formatted cards containing the
subject, Data Mart, complete rendered Insight, and a link back to OWOX.

### DIFF
--- a/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-access-validator.ts
+++ b/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-access-validator.ts
@@ -1,0 +1,51 @@
+import { Injectable, Logger } from '@nestjs/common';
+import { DataDestination } from '../../../../entities/data-destination.entity';
+import { DataDestinationConfig } from '../../../data-destination-config.type';
+import { DataDestinationCredentialsResolver } from '../../../data-destination-credentials-resolver.service';
+import { DataDestinationType } from '../../../enums/data-destination-type.enum';
+import {
+  DataDestinationAccessValidator,
+  ValidationResult,
+} from '../../../interfaces/data-destination-access-validator.interface';
+import { EmailConfigInputSchema } from '../../email/schemas/email-config.schema';
+import { EmailCredentialsSchema } from '../../email/schemas/email-credentials.schema';
+import { GoogleChatCredentialsSchema } from '../schemas/google-chat-credentials.schema';
+
+@Injectable()
+export class GoogleChatAccessValidator implements DataDestinationAccessValidator {
+  private readonly logger = new Logger(GoogleChatAccessValidator.name);
+  readonly type = DataDestinationType.GOOGLE_CHAT;
+
+  constructor(private readonly credentialsResolver: DataDestinationCredentialsResolver) {}
+
+  async validate(
+    destinationConfig: DataDestinationConfig,
+    dataDestination: DataDestination
+  ): Promise<ValidationResult> {
+    let resolvedCredentials;
+    try {
+      resolvedCredentials = await this.credentialsResolver.resolve(dataDestination);
+    } catch {
+      return new ValidationResult(false, 'Credentials are not configured');
+    }
+
+    const webhookCredentials = GoogleChatCredentialsSchema.safeParse(resolvedCredentials);
+    const emailCredentials = EmailCredentialsSchema.safeParse(resolvedCredentials);
+    if (!webhookCredentials.success && !emailCredentials.success) {
+      this.logger.warn('Invalid Google Chat credentials format', webhookCredentials.error);
+      return new ValidationResult(false, 'Invalid Google Chat credentials', {
+        errors: webhookCredentials.error.errors,
+      });
+    }
+
+    const config = EmailConfigInputSchema.safeParse(destinationConfig);
+    if (!config.success) {
+      this.logger.warn('Invalid Google Chat report configuration', config.error);
+      return new ValidationResult(false, 'Invalid configuration', {
+        errors: config.error.errors,
+      });
+    }
+
+    return new ValidationResult(true);
+  }
+}
--- a/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-access-validator.ts
+++ b/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-access-validator.ts
@@ -8,7 +8,10 @@ import {
   ValidationResult,
 } from '../../../interfaces/data-destination-access-validator.interface';
 import { EmailConfigInputSchema } from '../../email/schemas/email-config.schema';
-import { EmailCredentialsSchema } from '../../email/schemas/email-credentials.schema';
+import {
+  EmailCredentialsSchema,
+  EmailCredentialsType,
+} from '../../email/schemas/email-credentials.schema';
 import { GoogleChatCredentialsSchema } from '../schemas/google-chat-credentials.schema';
 
 @Injectable()
@@ -29,12 +32,14 @@ export class GoogleChatAccessValidator implements DataDestinationAccessValidator
       return new ValidationResult(false, 'Credentials are not configured');
     }
 
-    const webhookCredentials = GoogleChatCredentialsSchema.safeParse(resolvedCredentials);
-    const emailCredentials = EmailCredentialsSchema.safeParse(resolvedCredentials);
-    if (!webhookCredentials.success && !emailCredentials.success) {
-      this.logger.warn('Invalid Google Chat credentials format', webhookCredentials.error);
+    const credentialsValidation =
+      resolvedCredentials.type === EmailCredentialsType
+        ? EmailCredentialsSchema.safeParse(resolvedCredentials)
+        : GoogleChatCredentialsSchema.safeParse(resolvedCredentials);
+    if (!credentialsValidation.success) {
+      this.logger.warn('Invalid Google Chat credentials format', credentialsValidation.error);
       return new ValidationResult(false, 'Invalid Google Chat credentials', {
-        errors: webhookCredentials.error.errors,
+        errors: credentialsValidation.error.errors,
       });
     }
 
--- a/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-credentials-validator.ts
+++ b/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-credentials-validator.ts
@@ -1,0 +1,28 @@
+import { Injectable, Logger } from '@nestjs/common';
+import { DataDestinationCredentials } from '../../../data-destination-credentials.type';
+import { DataDestinationType } from '../../../enums/data-destination-type.enum';
+import {
+  DataDestinationCredentialsValidator,
+  ValidationResult,
+} from '../../../interfaces/data-destination-credentials-validator.interface';
+import { EmailCredentialsSchema } from '../../email/schemas/email-credentials.schema';
+import { GoogleChatCredentialsSchema } from '../schemas/google-chat-credentials.schema';
+
+@Injectable()
+export class GoogleChatCredentialsValidator implements DataDestinationCredentialsValidator {
+  private readonly logger = new Logger(GoogleChatCredentialsValidator.name);
+  readonly type = DataDestinationType.GOOGLE_CHAT;
+
+  async validate(credentials: DataDestinationCredentials): Promise<ValidationResult> {
+    const webhookCredentials = GoogleChatCredentialsSchema.safeParse(credentials);
+    const emailCredentials = EmailCredentialsSchema.safeParse(credentials);
+    if (!webhookCredentials.success && !emailCredentials.success) {
+      this.logger.warn('Invalid Google Chat credentials format', webhookCredentials.error);
+      return new ValidationResult(false, 'Invalid Google Chat credentials', {
+        errors: webhookCredentials.error.errors,
+      });
+    }
+
+    return new ValidationResult(true);
+  }
+}
--- a/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-credentials-validator.ts
+++ b/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-credentials-validator.ts
@@ -5,7 +5,10 @@ import {
   DataDestinationCredentialsValidator,
   ValidationResult,
 } from '../../../interfaces/data-destination-credentials-validator.interface';
-import { EmailCredentialsSchema } from '../../email/schemas/email-credentials.schema';
+import {
+  EmailCredentialsSchema,
+  EmailCredentialsType,
+} from '../../email/schemas/email-credentials.schema';
 import { GoogleChatCredentialsSchema } from '../schemas/google-chat-credentials.schema';
 
 @Injectable()
@@ -14,12 +17,14 @@ export class GoogleChatCredentialsValidator implements DataDestinationCredential
   readonly type = DataDestinationType.GOOGLE_CHAT;
 
   async validate(credentials: DataDestinationCredentials): Promise<ValidationResult> {
-    const webhookCredentials = GoogleChatCredentialsSchema.safeParse(credentials);
-    const emailCredentials = EmailCredentialsSchema.safeParse(credentials);
-    if (!webhookCredentials.success && !emailCredentials.success) {
-      this.logger.warn('Invalid Google Chat credentials format', webhookCredentials.error);
+    const credentialsValidation =
+      credentials.type === EmailCredentialsType
+        ? EmailCredentialsSchema.safeParse(credentials)
+        : GoogleChatCredentialsSchema.safeParse(credentials);
+    if (!credentialsValidation.success) {
+      this.logger.warn('Invalid Google Chat credentials format', credentialsValidation.error);
       return new ValidationResult(false, 'Invalid Google Chat credentials', {
-        errors: webhookCredentials.error.errors,
+        errors: credentialsValidation.error.errors,
       });
     }
 
--- a/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
+++ b/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
@@ -1,0 +1,278 @@
+import { Inject, Injectable, Logger, Scope } from '@nestjs/common';
+import { PublicOriginService } from '../../../../../common/config/public-origin.service';
+import {
+  EMAIL_PROVIDER_FACADE,
+  EmailProviderFacade,
+} from '../../../../../common/email/shared/email-provider.facade';
+import { OwoxEventDispatcher } from '../../../../../common/event-dispatcher/owox-event-dispatcher';
+import { MarkdownParser } from '../../../../../common/markdown/markdown-parser.service';
+import { buildDataMartUrl } from '../../../../../common/helpers/data-mart-url.helper';
+import { DataMartInsightTemplateFacadeImpl } from '../../../../ai-insights/data-mart-insight-template.facade';
+import { Report } from '../../../../entities/report.entity';
+import { InsightTemplateSourceDataService } from '../../../../services/insight-template-source-data.service';
+import { InsightTemplateService } from '../../../../services/insight-template.service';
+import { InsightTemplateSourceUsageService } from '../../../../services/insight-template-source-usage.service';
+import { DataDestinationCredentialsResolver } from '../../../data-destination-credentials-resolver.service';
+import { DataDestinationType } from '../../../enums/data-destination-type.enum';
+import { BaseEmailReportWriter } from '../../email/services/email-report-writer';
+import {
+  GoogleChatCredentials,
+  GoogleChatCredentialsSchema,
+} from '../schemas/google-chat-credentials.schema';
+import { GoogleChatMessagePayload, GoogleChatWebhookClient } from './google-chat-webhook.client';
+
+const MAX_GOOGLE_CHAT_PAYLOAD_BYTES = 30_000;
+const MAX_GOOGLE_CHAT_MESSAGE_PARTS = 20;
+const MAX_HEADER_FIELD_BYTES = 512;
+const GOOGLE_CHAT_WRITE_INTERVAL_MS = 1_100;
+
+function truncateUtf8(value: string, maxBytes: number): string {
+  if (Buffer.byteLength(value, 'utf8') <= maxBytes) return value;
+
+  const suffix = '…';
+  const availableBytes = maxBytes - Buffer.byteLength(suffix, 'utf8');
+  let result = '';
+  let usedBytes = 0;
+
+  for (const character of value) {
+    const characterBytes = Buffer.byteLength(character, 'utf8');
+    if (usedBytes + characterBytes > availableBytes) break;
+    result += character;
+    usedBytes += characterBytes;
+  }
+
+  return `${result}${suffix}`;
+}
+
+function buildGoogleChatMessage(input: {
+  subject: string;
+  dataMartTitle: string;
+  reportUrl: string;
+  chunk: string;
+  index: number;
+  total: number;
+}): GoogleChatMessagePayload {
+  const partLabel = input.total > 1 ? ` · Part ${input.index + 1} of ${input.total}` : '';
+  const subtitle = `Data Mart: ${input.dataMartTitle}${partLabel}`;
+
+  return {
+    fallbackText: `${input.subject} — ${subtitle}`,
+    cardsV2: [
+      {
+        cardId: `owox-insight-${input.index + 1}`,
+        card: {
+          header: {
+            title: input.subject,
+            subtitle,
+          },
+          sections: [
+            {
+              widgets: [
+                {
+                  textParagraph: {
+                    text: input.chunk,
+                    textSyntax: 'MARKDOWN',
+                  },
+                },
+                {
+                  buttonList: {
+                    buttons: [
+                      {
+                        text: 'Open report in OWOX',
+                        onClick: { openLink: { url: input.reportUrl } },
+                      },
+                    ],
+                  },
+                },
+              ],
+            },
+          ],
+        },
+      },
+    ],
+  };
+}
+
+function serializedPayloadBytes(payload: GoogleChatMessagePayload): number {
+  return Buffer.byteLength(JSON.stringify(payload), 'utf8');
+}
+
+function splitMarkdownForGoogleChat(input: {
+  markdown: string;
+  subject: string;
+  dataMartTitle: string;
+  reportUrl: string;
+}): string[] {
+  const characters = Array.from(input.markdown);
+  const chunks: string[] = [];
+  let offset = 0;
+
+  const fits = (chunk: string): boolean =>
+    serializedPayloadBytes(
+      buildGoogleChatMessage({
+        ...input,
+        chunk,
+        index: MAX_GOOGLE_CHAT_MESSAGE_PARTS - 1,
+        total: MAX_GOOGLE_CHAT_MESSAGE_PARTS,
+      })
+    ) <= MAX_GOOGLE_CHAT_PAYLOAD_BYTES;
+
+  if (!fits('')) {
+    throw new Error('Google Chat message headers exceed the API size limit');
+  }
+
+  while (offset < characters.length) {
+    if (chunks.length >= MAX_GOOGLE_CHAT_MESSAGE_PARTS) {
+      throw new Error(
+        `Google Chat Insight exceeds the ${MAX_GOOGLE_CHAT_MESSAGE_PARTS}-message delivery limit`
+      );
+    }
+
+    let low = 1;
+    let high = characters.length - offset;
+    let bestLength = 0;
+
+    while (low <= high) {
+      const candidateLength = Math.floor((low + high) / 2);
+      const candidate = characters.slice(offset, offset + candidateLength).join('');
+      if (fits(candidate)) {
+        bestLength = candidateLength;
+        low = candidateLength + 1;
+      } else {
+        high = candidateLength - 1;
+      }
+    }
+
+    if (bestLength === 0) {
+      throw new Error('Unable to split Google Chat message within the API size limit');
+    }
+
+    const candidateCharacters = characters.slice(offset, offset + bestLength);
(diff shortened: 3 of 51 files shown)


apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-access-validator.ts
```diff
@@ -0,0 +1,51 @@
...
+    let resolvedCredentials;
+    try {
+      resolvedCredentials = await this.credentialsResolver.resolve(dataDestination);
+    } catch {
+      return new ValidationResult(false, 'Credentials are not configured');
+    }
+
+    const webhookCredentials = GoogleChatCredentialsSchema.safeParse(resolvedCredentials);
+    const emailCredentials = EmailCredentialsSchema.safeParse(resolvedCredentials);
+    if (!webhookCredentials.success && !emailCredentials.success) {
```
romandubovyi: 🔴 **Blocking:** If both validations fail, we return the error details from `webhookCredentials.error.errors`. If the user provided invalid `email-credentials` (e.g. an empty list or invalid email format), they will be shown confusing error messages complaining about a missing/invalid Google Chat webhook URL. We should inspect the credential `type` property first, and then parse and return the errors for that specific schema.

🤖 Generated with Antigravity (Gemini 3.5 Flash)

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-credentials-validator.ts
```diff
@@ -0,0 +1,28 @@
...
+
+@Injectable()
+export class GoogleChatCredentialsValidator implements DataDestinationCredentialsValidator {
+  private readonly logger = new Logger(GoogleChatCredentialsValidator.name);
+  readonly type = DataDestinationType.GOOGLE_CHAT;
+
+  async validate(credentials: DataDestinationCredentials): Promise<ValidationResult> {
+    const webhookCredentials = GoogleChatCredentialsSchema.safeParse(credentials);
+    const emailCredentials = EmailCredentialsSchema.safeParse(credentials);
+    if (!webhookCredentials.success && !emailCredentials.success) {
```
romandubovyi: 🔴 **Blocking:** If both validations fail, we return the error details from `webhookCredentials.error.errors`. If the user provided invalid `email-credentials` (e.g. an empty list or invalid email format), they will be shown confusing error messages complaining about a missing/invalid Google Chat webhook URL. We should inspect the credential `type` property first, and then parse and return the errors for that specific schema.

🤖 Generated with Antigravity (Gemini 3.5 Flash)

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-webhook.client.ts
```diff
@@ -0,0 +1,139 @@
...
+
+      if (
+        this.isTransientStatus(response.status) &&
+        attempt < GoogleChatWebhookClient.MAX_ATTEMPTS
+      ) {
+        await this.waitBeforeRetry(response, attempt);
+        continue;
+      }
+
+      throw new Error(`Google Chat API returned HTTP ${response.status}`);
```
romandubovyi, line 69: ⚪ **Non-blocking:** When the Google Chat API returns an error status (e.g., 400 or 403), we currently throw a generic error containing only the HTTP status code, as `response.body` is immediately canceled in the `post()` method. Reading a small portion of the response text (e.g., using `await response.text()`) on failure and including it in the error message would greatly improve observability and ease troubleshooting for users.

🤖 Generated with Antigravity (Gemini 3.5 Flash)

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
```diff
@@ -0,0 +1,278 @@
...
+      } else {
+        high = candidateLength - 1;
+      }
+    }
+
+    if (bestLength === 0) {
+      throw new Error('Unable to split Google Chat message within the API size limit');
+    }
+
+    const candidateCharacters = characters.slice(offset, offset + bestLength);
```
romandubovyi, line 220: ⚪ **Non-blocking:** The character-based splitting logic (even when aligning with newlines) does not verify if the split occurs inside a Markdown block element, such as code blocks (``` ```) or tables. This might lead to broken or malformed formatting on the cards in Google Chat. Consider adding a small warning in code comments regarding this limitation, or a basic check to prevent breaking block structures if a cleaner parser-based split is too complex.

🤖 Generated with Antigravity (Gemini 3.5 Flash)

romandubovyi (commented):
Review complete. Thank you for the contribution! The changes look very solid and well-tested. Please address the comments regarding credentials validation error handling, Google Chat API error reporting, and markdown splitting safety.

🤖 Generated with Antigravity (Gemini 3.5 Flash)

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
```diff
@@ -0,0 +1,278 @@
...
+      '/reports'
+    );
+    const messages = buildGoogleChatMessages({
+      subject: this.emailConfig.subject,
+      markdown: markdownContent,
+      dataMartTitle: this.report.dataMart.title,
+      reportUrl,
+    });
+
+    for (let index = 0; index < messages.length; index += 1) {
```
romandubovyi, line 346: ⚪ **Non-blocking:** When a multi-part message is delivered, parts 1..N-1 may already be in Google Chat if part N fails. Consider logging each part's success/failure and documenting this at-least-once partial-delivery risk so users understand the behavior when a large Insight does not fully arrive.

🤖 Generated with Kimi Code

apps/web/src/features/data-destination/shared/model/mappers/google-chat.mapper.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    const credentials =
+      formData.credentials.deliveryMethod === 'email'
+        ? {
+            type: DataDestinationCredentialsType.EMAIL_CREDENTIALS as const,
+            to: formData.credentials.to,
+          }
+        : {
+            type: DataDestinationCredentialsType.GOOGLE_CHAT_CREDENTIALS as const,
```
romandubovyi: ⚪ **Non-blocking:** When `deliveryMethod` is `webhook` but `webhookUrl` is missing, this sends an empty string. Although backend validation rejects it, avoid emitting incomplete payloads by only including `webhookUrl` when it is non-empty or by throwing a client-side error for invalid form data.

🤖 Generated with Kimi Code

romandubovyi:
⚪ **Non-blocking:** `EMAIL_BASED_TYPES` still includes `google_chat`, and the tool schema enforces the `emails` parameter for it. The PR introduces a webhook delivery method, but exposing the secret webhook URL through MCP may be a security concern. Decide whether MCP should support webhook-based Google Chat destinations; if yes, extend the schema with an optional `webhook_url` and route it to `google-chat-credentials`. If not, update the tool description to clarify that only channel-email delivery is available via MCP.

🤖 Generated with Kimi Code

romandubovyi:
⚪ **Non-blocking:** The tool description in `apps/backend/src/ee/mcp/tools/add-destination.tool.ts` (around line 82) groups `google_chat` with email-based types. Update the wording to reflect that Google Chat now supports both direct API (webhook) and channel-email delivery in the UI/API, and clarify which method is available through MCP so agents do not mislead users.

🤖 Generated with Kimi Code

apps/web/src/features/data-destination/shared/model/types/data-destination.ts
```diff
@@ -56,9 +57,9 @@ export interface MsTeamsDataDestination extends BaseDataDestination<EmailCredent
   credentials: EmailCredentials;
 }
 
-export interface GoogleChatDataDestination extends BaseDataDestination<EmailCredentials> {
+export interface GoogleChatDataDestination extends BaseDataDestination<GoogleChatCredentials> {
   type: DataDestinationType.GOOGLE_CHAT;
-  credentials: EmailCredentials;
+  credentials: GoogleChatCredentials;
```
romandubovyi, line 62: ⚪ **Non-blocking:** Changing `GoogleChatDataDestination.credentials` to the webhook union leaves a stale call site outside this diff: `RecipientsDisplay` (rendered in `EmailReportEditForm` for all message destination types, including GOOGLE_CHAT) checks `'to' in credentials` and will show "Recipients of this report: No recipients found" for webhook-based Google Chat destinations, which reads like a misconfiguration. Consider hiding the recipients block for webhook delivery (or showing "Delivered via Google Chat webhook" instead).

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    if (parsed.success) {
+      this.googleChatCredentials = parsed.data;
+      this.usesEmailDelivery = false;
+      return;
+    }
+
+    // Channel email remains a supported delivery method for existing and new destinations.
+    this.usesEmailDelivery = true;
+    await super.prepareDeliveryCredentials(report);
```
romandubovyi, line 318: ⚪ **Non-blocking:** If the resolved credentials match neither the webhook nor the email schema, this falls through to the base class, which throws `Invalid Email credentials provided`. For a GOOGLE_CHAT destination that message is misleading when troubleshooting failed runs. Consider catching/wrapping the fallback failure with a Google Chat-specific message (e.g. "Google Chat destination has neither valid webhook nor channel-email credentials").

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
```diff
@@ -0,0 +1,278 @@
...
+      markdown: markdownContent,
+      dataMartTitle: this.report.dataMart.title,
+      reportUrl,
+    });
+
+    for (let index = 0; index < messages.length; index += 1) {
+      if (index > 0) {
+        await new Promise(resolve => setTimeout(resolve, GOOGLE_CHAT_WRITE_INTERVAL_MS));
+      }
+      await this.webhookClient.send(this.googleChatCredentials.webhookUrl, messages[index]);
```
romandubovyi: ⚪ **Non-blocking:** Multi-part Insights are posted as separate top-level messages, so a 20-part report scatters across the space's timeline. Google Chat incoming webhooks support the `threadKey` and `messageReplyOption=REPLY_MESSAGE_FALLBACK_TO_NEW_THREAD` query parameters — appending them (e.g. with the report run id as the thread key) would group all parts into a single thread under the first message.

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
```diff
@@ -0,0 +1,278 @@
...
+            title: input.subject,
+            subtitle,
+          },
+          sections: [
+            {
+              widgets: [
+                {
+                  textParagraph: {
+                    text: input.chunk,
+                    textSyntax: 'MARKDOWN',
```
prokopetsv, line 144: 🟡 **Non-Blocker** — the payload contract with the real Google Chat API is unverified: every test in this PR mocks `fetch`, so nothing proves an incoming webhook accepts this message shape. Two specific doubts:

1. Google's reference documents `textSyntax` as "Available for Google Chat apps and unavailable for Google Workspace add-ons" — incoming webhooks are not explicitly covered. If the webhook surface rejects or ignores the field, every API-delivery send 400s (feature dead on arrival) or silently renders Markdown source as HTML-escaped text.
2. Even where MARKDOWN is honored, Chat's card-markdown subset (bold/italic/strike/lists/links/code) does not include `#` headings or tables — which rendered Insights routinely contain — so "complete, formatted messages" may render with literal `#` and pipe characters.

Since the whole value of the webhook path is the formatted card, one manual smoke test against a real space webhook with a representative Insight (heading + table + bold + a multi-part-sized body) before merge is cheap insurance, and worth recording in the PR description. If headings/tables render literally, a small Markdown-downlevel step (headings → bold lines, tables → code block) in this writer would close the gap.

prokopetsv (commented):
## Review

I read the PR description and both existing review threads first; nothing below duplicates them. No blockers found beyond what is already flagged there — this is a carefully hardened change, and several areas deserve explicit credit:

### Verified solid

- **Secret handling is done right, end to end.** The webhook URL is allowlist-validated twice (schema on save, again at the network boundary in `GoogleChatWebhookClient.send` before any `fetch`), `redirect: 'error'` prevents token leakage via redirects, low-level network errors are sanitized because fetch implementations attach the request URL, API responses redact to `{type, configured: true}` with a mapper test asserting the serialized response never contains the URL or token, and the web form uses a `password` input with an explicit "paste a new URL to replace it" placeholder. The `DataDestinationCredentialsWithoutGoogleChat` exclusion type makes returning raw Google Chat credentials a compile error, not a code-review catch.
- **SSRF is closed**: `isGoogleChatWebhookUrl` pins protocol, hostname, port, path shape, and requires both `key` and `token`, with negative tests for host/protocol/port/path variants.
- **The masked-secret round-trip trap is avoided**: an update that omits credentials (or submits the blank hidden field of a configured webhook) sends no credential change, verified at the mapper, the update service, and e2e (legacy email preserved through a title edit). The stale-`configured` edge (flip to email, back to webhook, save) resolves to a correct no-op because `configured: true` only ever exists when the stored credential *is* the webhook.
- **Delivery-method switching** replaces the credential in place in both directions with the credential-type row updated (`EMAIL` ↔ `GOOGLE_CHAT_WEBHOOK`), tested both ways.
- **The size-splitting logic sizes the actual serialized payload** (binary search over the full JSON, worst-case part label), so multibyte content and JSON escape inflation can't overshoot the 30 KB limit — both covered by tests.

One 🟡 inline — the single thing the test suite structurally cannot cover.

zapolsky:

Reviewed against a worktree checkout of `53fee6c7` with five parallel analysis passes; every claim below was re-verified against the checked-out source, and several were verified by executing the code. Two candidate blockers were **refuted** during verification and are not posted (details at the bottom).

Credit where it's due: the secret-handling design is the right one — a dedicated `GOOGLE_CHAT_WEBHOOK` credential type with a non-round-trippable public shape rather than a mask-the-string hack, `redirect: 'error'`, re-validating the URL at the network boundary independently of what was persisted, deliberately sanitized fetch errors, and a redaction test that asserts on `JSON.stringify(response)` rather than one field. I traced every response, log, search-index, MCP and `identity` path and could not get the webhook URL out of the server.

- `apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-webhook.client.ts:86` — **a client-side timeout retries a non-idempotent POST, so one card can be posted up to 3 times.** Chain: `:86-89` aborts after 10 s → `:104-110` converts the `AbortError` into a generic `Error` → `:51-57` catches it in the *generic* retry arm and re-issues the same POST at `:92`. A 10 s abort is a client deadline, not a server rollback — Chat may have already created the message. Same for any `>= 500` (`:73-75`). Amplified per part by `google-chat-report-writer.ts:346-367`, so a duplicate can land *in the middle* of an ordered multi-part insight. `spaces.messages.create` accepts a `requestId` idempotency key and the pinned path (`google-chat-credentials.schema.ts:5`) is exactly that endpoint. **Fix:** append a stable `requestId` per part (run id + part index, constant across retries of that part), or treat a timeout as terminal.
- `apps/backend/src/data-marts/data-destination-types/ee/google-chat/schemas/google-chat-credentials.schema.spec.ts:28` — **the SSRF allowlist has no test that can fail.** The six malicious/malformed URL cases assert against the bare `isGoogleChatWebhookUrl` helper, never through `GoogleChatCredentialsSchema`; the only schema call is the positive case at `:11`. The validator specs' single negative case is `{type:'email-credentials', to: []}` (`google-chat-credentials-validator.spec.ts:26-29`, `google-chat-access-validator.spec.ts:34`), which routes to the *email* branch. Proven by mutation: deleting `.refine(isGoogleChatWebhookUrl, …)` from `google-chat-credentials.schema.ts:31` leaves 9 backend suites / 70 tests green. That guard is load-bearing at save time (`google-chat-credentials-validator.ts:22`), at access validation (`google-chat-access-validator.ts:38`), at send time (`google-chat-report-writer.ts:302`) and in the create/update API union (`data-destination-credentials.type.ts`). The web counterpart gets this right (`apps/web/.../google-chat-credentials.schema.test.ts:30-36`). **Fix:** assert the same six cases through `GoogleChatCredentialsSchema.safeParse(...).success === false` — one line, and the mutant dies.
- `apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts:51` — **valid GFM delimiter rows with 1–2 dashes (or a single column) skip the table downlevel entirely.** `/^:?-{3,}:?$/` with `cells.length >= 2` requires ≥3 hyphens and ≥2 columns; GFM requires one hyphen and allows one column. Verified by executing the function: `| Metric | Value |\n| :-: | -: |\n| Revenue | $10 |` comes back **unchanged** (control with `| --- | ---: |` is correctly fenced). Unchanged text flows to `textParagraph.text` at `:142-145`, and Chat cards have no table support — so the reader sees raw pipes. This matters more than it looks: the *email* path renders those same tables properly via `MarkdownParser.parseToHtml`, so flipping an existing destination from Channel Email to Incoming Webhook silently degrades content that used to render, and `| :-: |` is common in LLM-generated insight templates — exactly what this feature ships. **Fix:** `/^:?-+:?$/` and `cells.length >= 1`.
- `apps/backend/src/data-marts/dto/presentation/data-destination-response-api.dto.ts:29` — **the `Exclude<>` boundary is compile-time only; the sibling EMAIL branch still forwards a raw stored blob.** The runtime gate is `data-destination.mapper.ts:365-371`, keyed on the credential *row* type — which is derived from the `type` string *inside* the blob (`credential-type-resolver.ts:34-45`), and the blob is persisted exactly as the client sent it (`update-data-destination.service.ts:276-296` / `create-data-destination.service.ts:183-202` → `data-destination-credentials-processor.facade.ts:36-42` returns input unchanged, since no processor is registered for GOOGLE_CHAT at `data-destination-providers.ts:85`). Meanwhile `EmailCredentialsSchema` ends in `.passthrough()` (`ee/email/schemas/email-credentials.schema.ts:5-10`), so `{type:'email-credentials', to:[…], webhookUrl:'https://chat.googleapis.com/…?key=…&token=…'}` validates, persists, and is echoed verbatim by `data-destination.mapper.ts:359-363`. No product path produces this today — the web mapper always uses the webhook discriminator and MCP's schema is `.strict()` with no webhook field — which is why this is a hardening note and not a blocker. **Fix:** project explicitly in the EMAIL branch (`return { type: 'email-credentials', to: creds.to }`), matching what the `GOOGLE_CHAT_WEBHOOK` (`:368-371`) and `GOOGLE_OAUTH` (`:330-334`) branches already do; that makes the response independent of the storage schema.
- `apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts:202` — **the chunk binary search is bounded by the whole remaining insight instead of the 30 000-byte ceiling, so it does work proportional to total input size on the main event loop.** `let high = characters.length - offset` at `:202`, and every probe slices + `JSON.stringify`s up to that length (`:207`, `:180-188`, `:166-168`). Since each character contributes ≥1 byte, no chunk can ever exceed `MAX_GOOGLE_CHAT_PAYLOAD_BYTES` (`:25`), so every probe above that bound is provably wasted; and the 20-part rejection is only checked at the top of the loop (`:194-199`), so the full probe dance runs to completion *before* the rejection. Report runs execute in-process (`run-report.service.ts:202`; no worker threads in `apps/backend/src`). Measured on a faithful port: 0.5 MB → 122 ms / 12.6 M chars stringified; 2 MB → 405 ms / 45.2 M; 8 MB → **1 556 ms / 171.8 M**, all synchronous. **Fix:** `let high = Math.min(characters.length - offset, MAX_GOOGLE_CHAT_PAYLOAD_BYTES);` — output verified byte-identical across five cases including the multibyte and JSON-escape ones from the spec, and cost stops tracking input size.
- `apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts:346` — **worst-case delivery can outlive the graceful-shutdown budget, and nothing can interrupt it.** Per part: 3 attempts × 10 s timeout + 2 backoffs capped at 30 s (`google-chat-webhook.client.ts:37-40`, `:116-138`), plus the 1.1 s inter-part sleep (`:28`, `:348`) → ~30 min for 20 parts. Report runs register as active processes (`run-report.service.ts:385`) but never set a pid, so `sendSignalToProcess` (`graceful-shutdown.service.ts:95-114`) is a no-op for them; after `SCHEDULER_GRACEFUL_SHUTDOWN_TIMEOUT_MINUTES` (default **15**, `:39-42`) the shutdown force-resolves and an in-flight delivery dies mid-report. The run's `AbortSignal` never reaches the writer either — `ReportRunExecutionContext` carries only `{runId, logger}`. This is a tail risk (the 100-row cap at `report-execution-policy.resolver.ts:25` makes 1–3 parts typical), but the ceiling is real and it interacts with the duplicate-on-retry note above. **Fix:** thread an `AbortSignal` into `send()` and the inter-part sleep, and add a total-delivery budget well inside the shutdown window.

<details>

- `apps/backend/src/data-marts/data-destination-types/data-destination-credentials.guards.ts:20` — the repo already hosts `isEmailCredentials` / `isGoogleSheetsCredentials`; this PR added `GoogleChatCredentialsSchema` to the union (`data-destination-credentials.type.ts:5,9`) without adding `isGoogleChatCredentials` alongside, so "is this Google Chat destination email-delivered?" is now re-derived four different ways — `google-chat-credentials-validator.ts:20-23`, `google-chat-access-validator.ts:35-38` (copy), `google-chat-report-writer.ts:300-319` (order-dependent trial parse), and web `google-chat.mapper.ts:13-22` (extra `to.length > 0` the backend lacks). No live divergence today; pure future-drift cost.
- `apps/backend/src/data-marts/use-cases/run-report.service.ts:347` — `await reportWriter.finalize(...)` is unguarded inside `finally`, so a throwing writer skips `reportReader.finalize()` at `:353` (which closes cursors / destroys adapters, e.g. `databricks-report-reader.service.ts:117-141`). Pre-existing, but this PR adds a delivery path that fails far more often (remote HTTP, hard-fail on 4xx, up to 20 parts).
- `apps/backend/src/data-marts/data-destination-types/ee/email/services/email-report-writer.ts:397` — `produceRunEvent`'s `switch` has no `default:` arm; it is the only exhaustive switch in this feature that fails *silently* (all others throw). Also, the base class must still own the `GOOGLE_CHAT` case at `:413-417` even though `ee/email/README.md` now says the shared writer covers three types — consider a `protected abstract buildRunEvent(...)` hook, mirroring the two hooks this PR already introduced.
- `google-chat-report-writer.ts:195` — an oversized insight is rejected before any part is sent (docs are accurate), but only *after* `renderMessageTemplate()` has run the LLM prompts and consumption was registered (`email-report-writer.ts:143-144`), so switching a large existing report from Channel Email to Incoming Webhook can turn a working report into a hard, billed failure. The newline preference at `:225-228` can also roughly halve effective capacity for content with regular line breaks.
- `google-chat-report-writer.ts:56` — `boldHeading` treats "starts with `**` and ends with `**`" as already-bold, which does not imply the whole line is one bold span. Executed: `## **Q3** revenue vs **Q4**` → `**Q3** revenue vs **Q4**` (heading marker gone entirely); `# *Q3* revenue` → `***Q3* revenue**` (malformed nesting). Control `# Revenue` → `**Revenue**` is correct.
- `google-chat-webhook.client.ts:118` — `Retry-After: 0` returns 0, and `:79-81` skips the sleep entirely, so three POSTs fire back-to-back at an endpoint that just rate-limited us; the exponential fallback at `:133-137` is unreachable because `:121` already returned. Note the two client specs mock the header as `'0'` precisely so they need no fake timers (`google-chat-webhook.client.spec.ts:48`, `:64`) — they would still pass with a floor applied. Suggest `Math.min(Math.max(seconds * 1_000, RETRY_BASE_DELAY_MS), MAX_RETRY_DELAY_MS)`, and the same for the HTTP-date branch at `:126-129`.
- `google-chat-report-writer.ts:359` — `google_chat_part_failed` is written via `executionLogger.log`, not `.error`; the run-history viewer resolves severity from the entry `type` (`apps/web/.../DataMartRunHistoryView/utils.ts:55,95`), so the partial-delivery record renders as an informational line. The thrown error still surfaces, but "parts 1..N-1 were delivered" is the part the user most needs and it is the quietest line on the page.
- `google-chat-credentials-validator.ts:21-31` / `google-chat-access-validator.ts:25-52` validate URL *shape* only, with no liveness probe — compare Google Sheets, which does a live Drive round-trip inside the update transaction (`update-data-destination.service.ts:366-368`). A revoked or wrong-space webhook saves green and only fails on the next scheduled run. Possibly intentional (Chat offers no cheap probe short of posting), but it leaves a gap in the "does this destination work?" story.
- `google-chat-report-writer.ts:331` — on a self-hosted install with no `PUBLIC_ORIGIN`, `getPublicOrigin()` falls back to `http://localhost:PORT` (`public-origin.service.ts:93-103`), which becomes the card's `onClick.openLink.url`. Parity with the existing email template, but unlike email the whole message fails if Chat rejects the URL (400 is terminal at `google-chat-webhook.client.ts:61-69`). Consider omitting the button widget on the localhost fallback.
- `email-report-writer.spec.ts:148` — new `TS2345`: `it.each` widens the type to `DataDestinationType` while `createWriter` takes the narrowed `EMAIL | SLACK | MS_TEAMS` union (`:17-21`). **This does not fail CI** — `tsconfig.build.json` excludes `**/*spec.ts` and jest passes — but it is the one error under the full `tsconfig.json` attributable to this PR. A `as const` on the table fixes it.
- `.changeset/google-chat-direct-webhook.md` — filename lacks the Fibery task-number prefix used by 22 of 24 siblings (`6771-canvas-export.md` etc.). Bump level, structure and content all match, and there are no relative doc links, so the `apps/owox/CHANGELOG.md` link-resolution trap does not apply.
- minor: the post-build size guard at `google-chat-report-writer.ts:260-262` is unreachable by construction (the probe at `:184-187` already uses the worst-case `Part 20 of 20` label and longest `cardId`, and `splitLength <= bestLength`) — worth a comment saying it is defence-in-depth, or deleting; `google-chat-webhook.client.spec.ts:116-118` (`rejects.not.toThrow(webhookUrl)`) cannot fail given the exact-message assertion two lines above, and `google-chat-report-writer.spec.ts:188-190` / `:234-236` cannot fail because `buildGoogleChatMessages` already throws on oversize — the real guarantees are covered elsewhere, these three lines just carry no signal; `google-chat-report-writer.ts:278` holds `PublicOriginService` under a second name while also passing it to `super`, where widening `publicOriginService` to `protected` at `email-report-writer.ts:84` would match the four fields this PR already widened.

</details>

### Status of the existing review threads at `53fee6c7`

Not re-posting any of these — just closing the loop, since several were fixed after they were written.

| Thread | Status at HEAD |
|---|---|
| 🔴 Validator error-branch selection (both files) | **Fixed** — both now dispatch on `credentials.type` first (`google-chat-credentials-validator.ts:21-23`, `google-chat-access-validator.ts:35-38`), and `google-chat-credentials-validator.spec.ts:34` asserts the email branch is used |
| Misleading "Invalid Email credentials" fallback | **Fixed** — `google-chat-report-writer.ts:310-314` |
| Web mapper emitting an empty-string `webhookUrl` | **Fixed** — `google-chat.mapper.ts:49` guards with `webhookUrl?.trim()` |
| `RecipientsDisplay` "No recipients found" | **Fixed** — `RecipientsDisplay.tsx:12-17` returns `null` for webhook delivery |
| MCP `EMAIL_BASED_TYPES` wording | **Fixed** — renamed and reworded (`add-destination.tool.ts:17,84-89,130,148`) |
| Partial-delivery visibility | **Addressed** — per-part run logs (`:353-364`) plus the docs paragraph |
| Split cutting inside a Markdown block | **Addressed as documented intent** (`:223-224`); the limitation itself remains |
| Webhook client not reading the error body | **Still open** — body is cancelled at `:102` before the throw at `:69` |
| `threadKey` / `messageReplyOption` to group parts | **Not implemented** |
| Unverified `textSyntax: 'MARKDOWN'` on incoming webhooks | **Still open, and worth raising the stakes:** 400 is classified non-transient (`google-chat-webhook.client.ts:61-69`), so if the webhook surface rejects the field this is not a rendering nit — *every* delivery hard-fails on part 1. The manual smoke test suggested in that thread is the right gate before merge. |

### Verification notes

Two candidate blockers were **refuted** and are deliberately not posted:

- *"`isGoogleChatDataDestination` is not a type predicate, so `RecipientsDisplay.tsx:14` cannot compile."* Refuted by controlled experiment: with TypeScript 5.9.3, the un-annotated single-return guard at `data-destination.ts:93-95` gets an **inferred** type predicate (TS 5.5+), and narrowing works. Adding an explicit `: boolean` to the same function *does* produce `TS2339` — confirming inference is what makes it compile. `npx tsc --noEmit` and `tsc -b --force --listFiles` in `apps/web` are both clean with the file confirmed in the program.
- *"The schema-spec seam is a merge blocker."* Downgraded to a note: the shipped `.refine` is present and correct, CI is green, and no facade impl is missing a test — so it does not meet the blocking bar, even though the mutation result above makes it the highest-value note here.

CI parity, run inside the worktree: `tsc -p apps/backend/tsconfig.build.json --noEmit` clean · `apps/web` `tsc --noEmit` clean · ESLint clean on all changed backend + web files · Prettier clean · markdownlint clean · backend `jest` 10 suites / 85 tests pass (plus 4 further Google-Chat-touching suites / 54 tests) · web `vitest` full suite 197 files / 1455 tests pass. Playwright e2e was **not** run (needs a live backend), so the new `preserves a legacy Google Chat channel email when editing the title` case is unverified here.

**Not attributable to this PR:** `tsc -p tsconfig.json` (specs included) reports 141 errors; 140 are pre-existing in untouched files (`project-member-api-keys/services/project-member-api-key.service.spec.ts`, `test/*.e2e-spec.ts`, `test/utils/api-key-app-smoke.ts`). CI does not type-check specs, so none of these block anything.

zapolsky (approved):

PR #1459, opened 2026-07-23
fix(deps): resolve all npm audit advisories to zero

zapolsky:
## Summary

Drives `npm audit` and `npm audit --omit=dev` to **0 vulnerabilities**, across seven atomic, individually-validated commits.

| Fix | Advisory |
|-----|----------|
| `typeorm` 0.3.30 → 0.3.31 (backend + test-utils) | GHSA-2rp8-mm9q-fp49 |
| `adm-zip` 0.5.16 → 0.6.0 (@owox/connectors) | GHSA-xcpc-8h2w-3j85 |
| docs → astro 7 + starlight 0.41 + sharp 0.35 + links-validator 0.25 | GHSA-4g3v-8h47-v7g6, GHSA-f48w-9m4c-m7f5, GHSA-7pw4-f3q4-r2p2 |
| root overrides: `sharp` ^0.35.0, MCP's `@hono/node-server` ^2.0.5 | GHSA-f88m-g3jw-g9cj, GHSA-frvp-7c67-39w9 |
| `better-auth` 1.6.13 → 1.6.25 (both idp packages) | GHSA-qq9h-g4jm-xgf3 |
| `apps/web` → `react-router` 8.3.0, dropping `react-router-dom` | GHSA-qwww-vcr4-c8h2 |

The clean reinstall also floated in-range transitive advisories to patched releases: fast-uri 3.1.4 (GHSA-v2hh-gcrm-f6hx), fast-xml-parser 5.10.1 (GHSA-8r6m-32jq-jx6q) and brace-expansion (GHSA-mh99-v99m-4gvg).

## ⚠️ This raises the Node floor to 22.22.0 — everywhere, including the published CLI

`react-router` 8 declares `engines.node >= 22.22.0`, and this repo's `.npmrc` sets `engine-strict=true` — so `npm install` hard-fails on anything below. **Every workspace now declares `>=22.22.0`**, and so does `owox`.

Raising the published CLI is not cosmetic. `owox` keeps `@owox/web` in `dependencies`, and `@owox/web` is published with `access: public`, so a user installing `owox` also installs a package that now requires 22.22.0. Leaving `owox` at `>=22.16.0` would have shipped metadata that claims support for a range the install cannot actually satisfy under `engine-strict` — the CLI would fail on its own transitive dependency while its manifest promised otherwise. The declared floor has to match what gets installed.

- **Production is unaffected**: the Dockerfile already pins `node:22.22.3-slim`.
- **CI is unaffected**: workflows use `node-version: '22'`, which resolves to the latest 22.x.
- **Local development is affected**: anyone on 22.16–22.21 needs `nvm install 22` before their next install.
- **CLI users are affected**: installing or upgrading `owox` now requires Node 22.22.0.

The install instructions follow the code: `README.md`, `docs/editions/agency.md`, `docs/getting-started/deployment-guide/local-deployment.md` and `packages/connectors/CONTRIBUTING.md` all move to 22.22.0. The `apps/docs` copies under `src/content/docs/` are generated by `scripts/sync-docs.js` and regenerate from those sources.

## Why react-router had to be migrated rather than pinned

The obvious cheap fix is to pin below the vulnerable range, and it is a trap worth documenting.

`react-router-dom` has no 8.x line, and the advisory is patched only in `react-router` 8.3.0, so `^7.6.2` could never reach a patched release. The last version before the vulnerable range is 7.11.0 — but 7.11.0 sits inside the `6.0.0 - 7.17.0` range of **fourteen** further advisories, including an unauthenticated RCE (GHSA-49rj-9fvp-4h2h), stored XSS via unescaped Location headers (GHSA-f22v-gfqf-p8f3), and several open-redirect and DoS issues — all of which 7.18.x already fixed. Pinning back would have traded one advisory for fourteen worse ones.

For completeness on reachability: the advisory concerns RSC mode, which this app does not use — `apps/web` is a Vite SPA on `createBrowserRouter`, with no `@react-router/dev` and no RSC entry points. The migration is about clearing the advisory and staying on a supported line, not about a reachable exploit.

### What the migration touched

v8 removes the `react-router-dom` re-export package: DOM-specific APIs now come from `react-router/dom`, everything else from `react-router`.

- 127 files in `apps/web` move their imports to `react-router`.
- `App.tsx` and `use-schema-unsaved-guard.test.tsx` take `RouterProvider` from `react-router/dom`. This matters: the same-named export on the main entry is a **different component** (`lib/components` vs `lib/dom-export/dom-router-provider`), and only the latter is correct for a DOM app.
- `packages/ui` follows via its peer range (`react-router-dom: ^7` → `react-router: ^8`) and its one `Link` import.
- `apps/web` declares react-router 8's peer floor of `react`/`react-dom` `>=19.2.7`, replacing `^19.1.0` (19.2.8 was already installed).

No other API changed: all twenty router symbols this app uses are still exported from `react-router` in v8.

### Merging `main` produced a semantic conflict

The plugin gallery (#1493) landed on `main` with 15 new files importing from `react-router-dom` — a package this branch removes. Git merged them cleanly because nothing textually overlapped, but `apps/web` would not have built. Those imports move to `react-router` as part of the merge commit; every symbol involved (`Link`, `MemoryRouter`, `useNavigate`, `useLocation`, `useParams`, `RouteObject`) is exported from `react-router` in v8, and none of them is the DOM-specific `RouterProvider`.

## Notes on the out-of-range overrides

- **sharp**: `@huggingface/transformers` pins `^0.34.1`, so no in-range version carries the libvips CVE fix; the override forces 0.35.3. Verified sharp encodes and transformers loads with it.
- **@hono/node-server**: `@modelcontextprotocol/sdk` pins `^1.19.9` with no patched 1.x. MCP uses only `getRequestListener` (not the vulnerable `serve-static`), which is unchanged in 2.x; verified end-to-end (request → 200) and that the transport loads.

## Note on better-auth

Unlike most entries here, this one is directly reachable: the advisory covers account takeover via pre-account hijacking on magic-link and email-OTP sign-in, and the `magicLink` plugin is enabled in `packages/idp-better-auth/src/auth/auth-config.ts`. First patched in 1.6.22; both idp packages move to 1.6.25, the latest patch in the same minor.

## Validation

Re-run on Node 22.23.2 after merging `main` and raising the floor:

- `npm audit` and `npm audit --omit=dev`: **0 vulnerabilities**
- `apps/web`: builds (`tsc -b && vite build`), 1370 tests across 181 files pass, ESLint and Prettier clean
- Builds: `@owox/backend`, `@owox/connectors`, `@owox/plugin-sdk`, `@owox/api-client`, `@owox/idp-protocol`
- `apps/docs`: 150 pages built, all internal links valid
- `format:root:check` clean; markdownlint clean on every changed document
- `@owox/idp-better-auth` 156 tests, `@owox/idp-owox-better-auth` 262 tests pass


laskevych (approved):

apps/backend/package.json
```diff
@@ -113,7 +113,7 @@
     "rxjs": "^7.8.1",
     "snowflake-sdk": "^2.3.3",
     "swagger-ui-express": "^5.0.1",
-    "typeorm": "0.3.30",
+    "typeorm": "^0.3.31",
```
laskevych, line 117: 🔵 Question (non-blocking): `typeorm` moves from the exact pin `0.3.30` to `^0.3.31` (here and in `packages/test-utils`). Was the original exact pin deliberate? TypeORM has a track record of behavioral regressions inside 0.3.x patch releases, and the pin arrived together with the search-indexing work (#1357). The lockfile keeps installs reproducible either way, so this is only a question of whether the loosening is intentional rather than incidental — if the pin was on purpose, `0.3.31` exact would preserve that policy while still clearing GHSA-2rp8-mm9q-fp49.

package.json
```diff
@@ -115,6 +115,10 @@
     },
     "googleapis-common": {
       "google-auth-library": "^10.1.0"
+    },
+    "sharp": "^0.35.0",
```
laskevych, line 119: 🔵 Question (non-blocking): any reason the `sharp` override is global while the `@hono/node-server` override right below it is scoped under `@modelcontextprotocol/sdk`? Scoping it the same way — `"@huggingface/transformers": { "sharp": "^0.35.0" }` — would confine the forced upgrade to the one dependency that pins `^0.34.1`, so a future workspace adding a direct `sharp` dependency isn't silently overridden. Works fine as-is today since `apps/docs` already declares `^0.35.0`, so purely a targeting/consistency question.

apps/docs/package.json
```diff
@@ -20,21 +20,22 @@
...
+    "astro": "^7.1.3",
+    "sharp": "^0.35.0",
     "starlight-auto-sidebar": "^0.4.0"
   },
   "devDependencies": {
     "@owox/eslint-config": "*",
     "@owox/prettier-config": "*",
     "@types/js-yaml": "^4",
     "@types/papaparse": "^5",
+    "cookie": "^2.0.1",
```
laskevych, line 33: 💡 Suggestion (non-blocking): this `cookie` pin is a resolution workaround for the astro 7 build (per f2c2f0c). Workaround pins tend to outlive their reason — consider leaving a breadcrumb (a follow-up issue, or a line in the docs README) so it gets dropped once astro/starlight resolve the right copy on their own.

laskevych (approved):
✅ Approving — no blocking findings.

Verified independently: no `react-router-dom` references remain anywhere outside lockfile history; `RouterProvider` correctly comes from `react-router/dom` in both `App.tsx` and the unsaved-guard test (the same-named export on the main entry is indeed a different component); `packages/ui` is private, so its peer bump to `react-router ^8` breaks no external consumers; the three docs pages that still say Node 22.16.0 all describe installing the published `owox` CLI, whose floor deliberately stays at 22.16 — so they remain correct; Dockerfile (`node:22.22.3-slim`) and CI (`node-version: '22'`) clear the new floor. CI is fully green.

👏 Praise: the PR description is exemplary — the react-router pin-vs-migrate trap analysis and the reachability notes on each advisory made this review far cheaper than the 2.9k-line diff suggests.

The inline comments below are all non-blocking (2 questions, 1 suggestion).

apps/web/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "type": "module",
   "engines": {
-    "node": ">=22.16.0"
+    "node": ">=22.22.0"
```
amarchenk0, line 12: This raises `@owox/web` to Node `>=22.22.0`, but the published `owox` package still declares support for Node `>=22.16.0` and installs `@owox/web` as a dependency. With `engine-strict`, installing or updating `owox` on Node 22.16–22.21 is rejected for that transitive dependency even though the CLI metadata declares the range supported. Raise `owox`’s `engines.node` to `>=22.22.0` and update the documentation/release notes, or demonstrate a distribution in which `@owox/web` is not installed as a runtime dependency. Add an installation check at the declared minimum Node version.

amarchenk0 (changes requested):
Changes requested.

apps/web/package.json
zapolsky: Confirmed and fixed — the metadata was inconsistent exactly as you describe. `owox` keeps `@owox/web` in `dependencies` (not devDependencies), and `@owox/web` publishes with `access: public`, so the CLI declared a support range its own install tree could not satisfy.

**Every** workspace now declares `engines.node >= 22.22.0` — the root, `owox`, `@owox/backend`, `@owox/ctl`, `@owox/web`, `@owox/api-client`, `@owox/plugin-sdk`, `@owox/internal-helpers`, `@owox/idp-protocol`, `@owox/idp-better-auth`, `@owox/idp-owox-better-auth`, `@owox/test-utils` and `@owox/ui`. There is no longer a version at which one package accepts an install that another rejects.

Documentation follows the code: `README.md`, `docs/editions/agency.md`, `docs/getting-started/deployment-guide/local-deployment.md` and `packages/connectors/CONTRIBUTING.md` all state 22.22.0. The `apps/docs` copies under `src/content/docs/` are generated by `scripts/sync-docs.js` and regenerate from those sources. The PR description now states the CLI runtime requirement moves too, instead of claiming only building moves.

On the installation check — verified by hand against the declared minimum:

- `npm ci --dry-run` on Node **22.22.0** (the exact floor) resolves the full tree under `engine-strict=true`: exit 0, no `EBADENGINE`.
- The same command on Node 22.17.0 fails with `npm error code EBADENGINE` — the failure mode you predicted for the 22.16–22.21 window.

CI is unaffected (`node-version: '22'` resolves to the latest 22.x) and production already pins `node:22.22.3-slim`.

I left a dedicated install-at-minimum CI job out of this PR to keep it scoped to the dependency work. Happy to open a follow-up so the floor is guarded automatically rather than by hand — say the word.

This branch also merges current `main`, which required one extra fix: the plugin gallery (#1493) added 15 files importing from `react-router-dom`, a package this branch removes. Git merged them without a textual conflict, so `apps/web` would have failed to build; those imports now come from `react-router`. Full re-validation on the merged tree: `apps/web` builds with 1370 tests across 181 files passing, backend/connectors/plugin-sdk/api-client/idp-protocol build, docs build 150 pages with all internal links valid, and both `npm audit` and `npm audit --omit=dev` still report 0 vulnerabilities.

romandubovyi (approved):

laskevych (approved):

PR #1466, opened 2026-07-23
Google Sheets source connector

AnastasiaKhukhra:
Add a Google Sheets source connector that imports one sheet into a table in the user's Storage and makes it available as a Data Mart. The connector supports Service Account authentication, dynamic header-based field inference, and full refreshes across supported storages so the materialized table reflects the current sheet contents, including schema changes.


packages/connectors/src/Sources/GoogleSheets/Source.js
```diff
@@ -0,0 +1,903 @@
...
+  }
+
+  _buildA1Range({ preview = false } = {}) {
+    const configuredRange = String(this.config.Range?.value || '').trim();
+    const gridRange = this._getConfiguredGridRange(configuredRange);
+    const sheetPrefix = `${this._quoteSheetName(this.config.SheetName.value)}!`;
+
+    const parsedBounds = this._parseA1GridRange(gridRange);
+    if (gridRange && !parsedBounds) {
+      return `${sheetPrefix}${gridRange}`;
```
prokopetsv: 🟡 **Non-Blocker** — an unparseable-but-API-valid `Range` silently misaligns the import. When `_parseA1GridRange` returns null, the raw range is passed through to the API here, but `_getConfiguredRangeBounds` (used for the header math in `_buildSheetSnapshot`) falls back to `startRow: 1`. R1C1 notation is the concrete case: `R5C2:R10C4` fails the A1 parser, yet the Sheets API accepts it and returns values starting at row 5 — so `headerIndex = headerRow - 1` reads the wrong row as headers and the run imports mislabeled data with no error. Safest fix is to reject a non-empty range the parser cannot understand (the passthrough currently buys nothing that works correctly end-to-end); alternatively, derive the true start row from the `range` field the API echoes back in its response.

Related hardening in the same request path: `_fetchSheetValues` interpolates `_extractSpreadsheetId`'s fallback (the raw configured value) into the URL path without `encodeURIComponent`, so a value containing `/` or `?` reshapes the request path. Impact is minimal — the host is pinned and the bearer token is the caller's own — but encoding the segment is one line.

apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts
```diff
@@ -161,4 +174,85 @@ export class UpdateDataMartDefinitionService {
...
   }
+
+  private validateConnectorConfigurationCount(command: UpdateDataMartDefinitionCommand): void {
+    if (command.definitionType !== DataMartDefinitionType.CONNECTOR) {
+      return;
+    }
+
+    const definition = command.definition as ConnectorDefinition;
+    const source = definition?.connector?.source;
+    if (source?.name === 'GoogleSheets' && source.configuration?.length !== 1) {
```
prokopetsv: 🟡 **Non-Blocker** — the `'GoogleSheets'` literal now gates behavior in at least four shared services: this use case (single-configuration rule, credential-copy access check, secret-reference validation), `connector-secret.service.ts` (dropping `_secrets_id` on copy), `google-sheets-preview-credentials.service.ts`, and the web mirrors the same constant. Each future connector with dynamic fields or copy-by-value secrets (the PR's own docs anticipate more) grows every one of these name checks — the central-orchestration `if/switch` growth the project's review rules call out. A capability flag on the connector definition (e.g. `singleConfiguration`, `copySecretsByValue`, `dynamicFields` in the manifest/spec) would let these services branch on the capability instead of the name, and the next such connector becomes a manifest change instead of a four-service sweep. Fine as a follow-up; worth doing before connector #2 of this kind.

prokopetsv (commented):
## Review

No existing reviews or comments were present, so nothing here duplicates prior feedback. No blockers. This is an unusually well-engineered PR for its size, and the highest-risk surfaces hold up under scrutiny:

### Verified solid

- **Full-refresh safety across all five storages**: every `replaceData` builds a validated staging table first and only then publishes; the schema-unchanged path preserves the live table object (Redshift `DELETE`+`INSERT` in one Data API transaction, Snowflake `INSERT OVERWRITE`, Databricks `INSERT OVERWRITE`, BigQuery `TRUNCATE`+`INSERT` in a transaction), and the schema-changed path handles the things table replacement destroys — Redshift re-applies grants read from `svv_relation_privileges` before the rename, Snowflake clones with `COPY GRANTS COPY TAGS`, Athena keeps a backup table with crash recovery (`recoverSnapshotBackupIfNeeded`) plus S3 cleanup that never deletes the just-published snapshot. I verified the two vendor-syntax claims I doubted against current docs — Snowflake `COPY TAGS` on `CREATE OR REPLACE … CLONE` and BigQuery `TRUNCATE TABLE` inside multi-statement transactions are both valid.
- **`storage-snapshot-safety.test.js` is the right kind of test suite**: per-storage publish-order assertions, empty-snapshot publication, staging-failure leaves-live-intact, row-count mismatch rejection, identifier-length limits, and DDL-vs-DML quoting differences (Athena backticks vs double quotes).
- **Credential handling**: the preview endpoint re-validates every `_secrets_id` reference (project scope, connector pinning, config binding, and `EDIT` on the owning Data Mart — checked on every path, so a forged `_id` cannot bypass it); definition updates reject foreign secret references; Google Sheets secret copies drop the shared `_secrets_id` so two data marts never alias one secret record; the in-process preview is bounded by a 15s abort-signal timeout honored through fetch, retries, and backoff sleeps, plus a 50 MB response cap.
- **Field-sync race handling**: `updateConnectorSourceFields` performs a conditional update guarded by `modifiedAt` (with a SQLite millisecond-precision workaround tested against a real in-memory DB), so a user edit during a run wins over the run's field reconciliation; sync failure degrades to a warning on a still-successful run.
- **Core changes are sibling-safe**: `AbstractStorage` gains only additive methods (`replaceData` throws by default, `hasSameSchema` is order-insensitive and used only by the new paths); the Source's `isValidToRetry` override is a legitimate provider-specific narrowing per the connector rules; `uniqueKeys` (`_owox_row_number`) is always present in `fields`.

Two 🟡 non-blockers inline.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -235,6 +254,116 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+      let published = false;
+
+      try {
+        this.config.DestinationTableName.value = stagingTableName;
+        this.existingColumns = {};
+        this.updatedRecordsBuffer = {};
+        this.totalRecordsProcessed = 0;
+        this.quoteFieldIdentifiers = true;
+        stagingTableCreated = true;
+        const stagedColumns = await this.createTableIfItDoesntExist(true);
```
romandubovyi, line 287: 🔴 **Blocking:** After `createTableIfItDoesntExist(true)`, assign `this.existingColumns = stagedColumns` before `saveData`. Right now `this.existingColumns` stays `{}`, so the first row treats every field as new and runs `addNewColumns`. That path emits unquoted `ADD COLUMN IF NOT EXISTS` identifiers, while the staging table was created with quoted identifiers (`quoteColumnNames = true`). Sheet headers that normalize to reserved words (`order`, `group`, `select`, `date`, …) will fail the snapshot load with a SQL syntax error. The snapshot-safety tests mock `saveData`, so this path is untested. Redshift/Databricks already assign `this.existingColumns` in their create path — match that pattern here.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Grok CLI (grok-4.5)

packages/connectors/src/Storages/Snowflake/SnowflakeStorage.js
```diff
@@ -328,6 +351,98 @@ var SnowflakeStorage = class SnowflakeStorage extends AbstractStorage {
...
+      const stagingTable = `${database}.${schema}.${quoteIdentifier(stagingTableName)}`;
+      const originalExistingColumns = this.existingColumns;
+      const liveColumns = await this.getAListOfExistingColumns();
+      let published = false;
+
+      try {
+        this.config.DestinationTableName.value = stagingTableName;
+        this.existingColumns = {};
+        this.updatedRecordsBuffer = {};
+        const stagedColumns = await this.createTableIfItDoesntExist();
```
romandubovyi, line 380: ⚪ **Non-blocking:** Same gap as in the BigQuery `replaceData`: `createTableIfItDoesntExist()` returns `stagedColumns`, but `this.existingColumns` stays `{}` until after publication. `saveData` then issues an unnecessary `ADD COLUMN IF NOT EXISTS` for every column on each snapshot. Snowflake quotes identifiers here, so this is harmless DDL noise rather than a failure — but assigning `this.existingColumns = stagedColumns` before `saveData` avoids the extra ALTER and matches the Redshift/Databricks pattern.

🤖 Generated with Grok CLI (grok-4.5)

packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
```diff
@@ -430,6 +560,173 @@ var AwsRedshiftStorage = class AwsRedshiftStorage extends AbstractStorage {
...
+    return actualRowCount;
+  }
+  //----------------------------------------------------------------
+
+  //---- getTableGrantStatements -------------------------------------
+  async getTableGrantStatements(sourceTableName, targetTableName) {
+    const schemaName = stripQuotes(this.config.Schema.value);
+    const rows = await this.executeQueryWithResults(`
+      SELECT identity_name, identity_type, privilege_type, admin_option
+      FROM svv_relation_privileges
```
romandubovyi: ⚪ **Non-blocking:** This query matches `namespace_name` / `relation_name` with exact case, but Redshift folds unquoted identifiers to lowercase. With a mixed-case `Schema` or table config (e.g. `Analytics`), no privileges are returned, and the schema-change rename path silently publishes the new table without carrying grants over. `getAListOfExistingColumns` in this same file already documents this pitfall and falls back to `LOWER()` matching when the exact query returns nothing — apply the same fallback here.

apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx
```diff
@@ -287,7 +350,119 @@ export function ConnectorEditForm({
...
+        return true;
+      } catch (error) {
+        const apiError = extractApiError(error) as { message?: string } | undefined;
+        const message =
+          apiError?.message ??
+          (error instanceof Error ? error.message : 'Failed to load Google Sheets columns');
+        if (mode === 'fields-only') {
+          setFieldsOnlyPreviewError(message);
+        }
+        if (!apiError) {
```
romandubovyi: ⚪ **Non-blocking:** The global axios interceptor only toasts 400/403/404, but this endpoint is specified to also return 500/502/504 (`PreviewGoogleSheetsFieldsSpec`). In wizard/configuration-only mode (`mode !== 'fields-only'`), `extractApiError` returns truthy for any error response with a body, so this `if (!apiError)` branch skips the toast — clicking Next on a 502/504 silently does nothing. Consider toasting API errors here too when not in fields-only mode (or extending the interceptor to cover 5xx).

apps/backend/src/data-marts/services/connector/connector-fields-preview.service.ts
```diff
@@ -0,0 +1,236 @@
...
+      'access token',
+      'authentication failed',
+      'failed to get access token',
+      'invalid credential',
+      'invalid_grant',
+      'token error',
+    ].some(fragment => message.includes(fragment));
+  }
+
+  private looksLikeConfigurationFailure(message: string): boolean {
```
romandubovyi, line 222: ⚪ **Non-blocking:** `looksLikeConfigurationFailure` only matches a narrow set of message fragments. Configuration errors thrown inside `_fetchSheetValues` (e.g. `Range must use the selected sheet '…'`) get wrapped into `HttpRequestException` with no status code, so `mapPreviewError` classifies them via `isProviderRequestError` as 502 "provider is temporarily unavailable" — a misleading response for a user misconfiguration. Consider expanding the matcher or, better, using structured error codes from `GoogleSheetsSource` so validation failures consistently map to 400.

🤖 Generated with Grok CLI (grok-4.5)

apps/backend/src/data-marts/use-cases/connector/google-sheets-fields-preview.service.ts
```diff
@@ -0,0 +1,236 @@
...
+    const SourceClass = Connectors[GOOGLE_SHEETS_SOURCE_NAME].GoogleSheetsSource;
+    const sourceConfig = new Core.SourceConfigDto({
+      name: GOOGLE_SHEETS_SOURCE_NAME,
+      config: configuration,
+    });
+
+    return new SourceClass(new GoogleSheetsFieldsPreviewConfig(sourceConfig.config, this.logger));
+  }
+
+  private mapFieldsSchemaToDto(sourceFieldsSchema: SourceFieldsSchema) {
```
romandubovyi: ⚪ **Non-blocking:** `SourceFieldsSchema` and `mapFieldsSchemaToDto` are near-verbatim copies of the ones in `connector.service.ts` (lines 57 and 505), except this copy drops the `uniqueKeysByDataLevel` mapping — the first divergence is already here. Two copies of the same mapping will drift on the next fields-schema change. Consider exposing the existing mapper from `ConnectorService` (already available in this module) or extracting it into a shared helper.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Claude Code (Opus 4.8)

packages/connectors/src/Sources/GoogleSheets/Source.js
```diff
@@ -0,0 +1,903 @@
...
+
+    return this._buildFieldsSchema(schema);
+  }
+
+  async _fetchSheetValues({ preview = false, signal } = {}) {
+    const spreadsheetId = this._extractSpreadsheetId(this.config.SpreadsheetId.value);
+    const range = this._buildA1Range({ preview });
+    const encodedRange = encodeURIComponent(range);
+    const url =
+      `https://sheets.googleapis.com/v4/spreadsheets/${spreadsheetId}/values/${encodedRange}` +
```
romandubovyi: ⚪ **Non-blocking:** `_extractSpreadsheetId` returns the raw trimmed value when it doesn't match the URL regex, and it is interpolated into the request URL unencoded. A value containing `/`, `?`, or `#` changes the request path/query sent to sheets.googleapis.com, producing confusing API errors instead of a clear validation failure. Wrap the id with `encodeURIComponent(...)` (or validate it against the same `[a-zA-Z0-9-_]+` charset the extraction regex already uses).

apps/backend/src/data-marts/controllers/connector.controller.ts
```diff
@@ -62,6 +66,20 @@ export class ConnectorController {
     return this.mapper.toFieldsResponse(fields);
   }
 
+  @Auth(Role.editor())
+  @Post('GoogleSheets/fields/preview')
```
romandubovyi: ⚪ **Non-blocking:** Every sibling endpoint in this controller is parameterized (`:connectorName/fields`, `:connectorName/specification`, `:connectorName/oauth/...`), while this one hardcodes `GoogleSheets` into both the route and the service. The next connector with dynamic fields will require either a duplicate endpoint or a breaking route rename. A generic `POST :connectorName/fields/preview` that checks whether the source implements `fetchFieldsSchema` (a capability check that already exists in `GoogleSheetsFieldsPreviewService.run`) would fit the established pattern.

romandubovyi (commented):
Multi-agent review of the Google Sheets source connector. Findings were collected by three independent CLI agents, cross-validated against the actual PR code, and curated manually: 1 blocking issue (BigQuery snapshot path) and 9 non-blocking suggestions across storages, backend, and the wizard flow. 4 false positives were rejected during validation.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Grok CLI (grok-4.5) + Claude Code (Opus 4.8)

apps/backend/src/data-marts/services/connector/connector-secret.service.ts
```diff
@@ -815,6 +815,9 @@ export class ConnectorSecretService {
       ) as Record<string, unknown>;
 
       delete mergedItem._copiedFrom;
+      if (incoming.connector.source.name === 'GoogleSheets') {
```
romandubovyi: ⚪ **Non-blocking:** This is one of several independent declarations of the `'GoogleSheets'` literal across the backend (local constants in `google-sheets-preview-credentials.service.ts` and `google-sheets-fields-preview.service.ts`, three bare literals in `update-data-mart-definition.service.ts`). A connector-specific branch inside the generic secret-merge flow also doesn't explain why `_secrets_id` must be dropped for this connector only. At minimum share one constant; ideally add a short comment on the why, or model it as a connector capability instead of a name check.

packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js
```diff
@@ -9,6 +9,118 @@ function stripQuotes(value) {
...
+    result += character;
+    bytes += characterBytes;
+  }
+
+  return result;
+}
+
+const REDSHIFT_SNAPSHOT_MAX_QUERY_BYTES = 90 * 1024;
+
+function utf8ByteLength(value) {
```
romandubovyi, line 51: ⚪ **Non-blocking:** `utf8ByteLength` and `splitSnapshotRowsByQuerySize` are byte-for-byte duplicated here and in `DatabricksStorage.js`. This PR already moved the shared snapshot helper `hasSameSchema` into `AbstractStorage` — these two belong there as well; only the byte-limit constants differ per storage.

packages/connectors/src/Sources/GoogleSheets/Source.js
```diff
@@ -0,0 +1,903 @@
...
+      '_owox_imported_at',
+      'owox_row_number',
+      'owox_imported_at',
+    ]);
+
+    return Array.from({ length: columnCount }, (_, index) => {
+      const header = headerRow[index];
+      const originalName = header === undefined || header === null ? '' : String(header).trim();
+      const fallbackName = `column_${index + 1}`;
+      let normalizedName = this._normalizeColumnName(originalName || fallbackName);
```
zapolsky, line 639: **[Bug — verified]** Any header with no ASCII alphanumerics (Cyrillic, CJK, Greek, Hebrew, Arabic, Thai, emoji) collapses to the constant `column`, and duplicates are numbered in *encounter order* — so any structural edit to the sheet silently re-maps which sheet column lands in which warehouse column.

**Evidence chain:** `Source.js:624-626` — `originalName` is the raw trimmed cell, and the positional fallback is applied as `this._normalizeColumnName(originalName || fallbackName)`; the `||` is evaluated on the **raw** cell, so a non-empty non-Latin header is truthy and `column_${index+1}` never fires. It is not re-checked after normalisation. → `Source.js:740-757` `_normalizeColumnName` is ASCII-only (`.replace(/[^a-z0-9_]+/g,'_')`, no `u` flag) and ends `if (!name) name = 'column'`. → `Source.js:759-776` `_deduplicateColumnName` appends `_2`, `_3`… positionally. → `Source.js:18` + `:82-88` `ImportAllColumns` defaults to `true`; `Connector.js:37` full-refreshes via `replaceData`. → `Connector.js:28` persists the normalised names into `connector.source.fields`.

Reproduced against both the vm-loaded `Source.js` and the shipped `dist/index.cjs`:

```
[Дата, Выручка, Клики]     -> [column, column_2, column_3]
[Новый, ID, Дата, Выручка] -> [column, id, column_2, column_3]
[日付, 売上]                -> [column, column_2]
['', Дата]                 -> [column_1, column]
['', Неделя 1, Неделя 2]   -> [column_1, column_1_2, column_2]
```

End-to-end through `fetchData()`: run 1 `[Дата,Выручка,Клики]` puts `2026-01-01` in `column`; run 2 after inserting one header puts `NEW` in `column`.

**Worse than a plain re-map:** a *reorder* keeps both the column count and the inferred types identical, so the real `AbstractStorage.hasSameSchema` returns `true` and publication takes the **preserve** branch — no DDL, no table recreate, no log line. The run is indistinguishable from a normal refresh while every value has moved. Subset mode does not help: `Connector.js:24-28` persists the normalised names, so `Fields = 'sheet column, sheet column_3'` is equally positional.

Checked and ruled out: no transliteration, no `\p{L}`, no NFKD anywhere on the path; the backend preview does no normalisation of its own (it delegates to this same code, so the wizard shows the same names); all five storages quote identifiers, so nothing fails loudly. `GoogleSheets` is the only connector in `packages/connectors/src/Sources/*` that derives column names from user data, so this is not a pre-existing repo convention. `packages/connectors/test/GoogleSheets.test.js` has zero non-Latin header coverage.

**Fix:** re-check after normalisation, not before — if `_normalizeColumnName(originalName)` yields empty or the bare `column` sentinel, fall back to the positional `column_${index + 1}` (stable w.r.t. sheet position), or transliterate.

packages/connectors/src/Sources/GoogleSheets/Source.js
```diff
@@ -0,0 +1,903 @@
...
+    }
+
+    const originalName = column.originalName || `Blank header at column ${column.index + 1}`;
+    return `Google Sheets column: ${String(originalName)
+      .replace(/"/g, "'")
+      .replace(/[\r\n]+/g, ' ')}`;
+  }
+
+  _inferType(values) {
+    if (!values.length) {
```
zapolsky, line 885: **[Bug — verified]** `_inferType([])` returns `STRING` with no anchor to the types already published, so a *data* state silently becomes a *schema* change and forces the destructive publish branch.

**Evidence chain:** `Source.js:868-871` — no observed values ⇒ `STRING`. → `Source.js:826-835` builds `values` per column and `_normalizeCellValue` (`:804-810`) maps `''`/missing to `null`, which is then filtered out — so a column blank in *every* row yields `[]` even when other columns are populated. → `Source.js:308-333` — with the header row present and no data rows, `detectedColumns` is non-empty, `_assertImportSize` only checks the upper bound, and neither the `No columns found` nor the `No columns selected` guard fires; nothing throws. → `Connector.js:37` calls `replaceData(data)` **unconditionally** (`data.length` only picks a log string at `:30-34`), unlike `AbstractConnector.js:163` which guards with `if (data.length || CreateEmptyTables)`. → `AbstractStorage.js:149-170` sees identical names and count but a different type ⇒ `hasSameSchema` false. → `GoogleBigQueryStorage.js:342-357` emits `CREATE OR REPLACE TABLE live COPY staging` instead of the `TRUNCATE + INSERT` preserve path; the other four storages take the analogous destructive path with 0 rows (no storage has a 0-row guard).

Executed end-to-end against the real `Source.js` / `Connector.js` / `AbstractStorage.js` / `GoogleBigQueryStorage.js` (only HTTP + BigQuery I/O stubbed):

```
RUN 1 populated              -> revenue INT64
RUN 2 same sheet, rows cleared -> revenue STRING, publish: CREATE OR REPLACE TABLE ... COPY   <- run SUCCEEDS
RUN 3 rows present, one numeric column blank -> orders INT64, revenue STRING  <- narrower variant is real
RUN 5 data returns next day  -> revenue INT64 again
```

So the table thrashes between types across runs, each flip recreating the table object. Verified there is genuinely no anchor: `connector-source-config.service.ts:29-31,47` builds `Fields` as names only, the `FIELDS_UPDATE` message is `fields: string[]`, and `createTableIfItDoesntExist` builds staged columns purely from this run's inference.

Ruled out: a sheet whose header row is *also* wiped throws `No columns found` and leaves the table intact — so the trigger is specifically "data rows cleared, headers kept", which is the normal script-rewrites-the-body case.

**Fix:** when a column has no observed values, reuse the type already published for it (available from the storage's `existingColumns`) instead of defaulting to `STRING` — or skip schema replacement entirely when `rows.length === 0`.

packages/connectors/src/Sources/GoogleSheets/Source.js
```diff
@@ -0,0 +1,903 @@
...
+  }
+
+  _buildFieldDescription(column) {
+    if (column.generated) {
+      return `Google Sheets generated column ${column.index + 1}`;
+    }
+
+    const originalName = column.originalName || `Blank header at column ${column.index + 1}`;
+    return `Google Sheets column: ${String(originalName)
+      .replace(/"/g, "'")
```
zapolsky, line 880: **[Bug — verified]** A backslash in a sheet header reaches BigQuery DDL unescaped and breaks **every** run until the sheet is edited. `_buildFieldDescription` sanitises `"` and CR/LF but not `\`.

**Evidence chain:** `Source.js:624` `originalName` is the raw trimmed header (only `_normalizeColumnName` is applied to `column.name`, never to `originalName`) → `Source.js:857-865` `_buildFieldDescription` rewrites `"`→`'` and CR/LF→space, nothing else (grep for backslash handling in `Source.js` returns zero hits) → `Source.js:833` → `_setDynamicFieldsSchema:434` → `Connector.js:57` passes `nodeSchema.fields` as the storage `schema` → `AbstractStorage.js:26` `this.schema = schema` → **`GoogleBigQueryStorage.js:217`** raw interpolation into a GoogleSQL literal: `` ` OPTIONS(description="${this.schema[columnName]["description"]}")` `` (duplicated at `:403` in `addNewColumns`) → `GoogleBigQueryStorage.js:278` `replaceData` calls `createTableIfItDoesntExist(true)` against a per-run-unique staging name, so this DDL runs on every run.

Executed against the real `Source.js` + real `GoogleBigQueryStorage.js`:

```
CREATE TABLE IF NOT EXISTS `project.dataset.live__owox_staging_abc` (
`c_reports` STRING OPTIONS(description="Google Sheets column: C:\Reports"),
`path` STRING OPTIONS(description="Google Sheets column: Path\"),
...)
```

ZetaSQL's lexical spec lists the valid escapes and states "Any sequence not in this table produces an error" — `\R` is not among them; and a trailing `\` escapes the closing quote while quoted strings cannot span newlines, so that variant is an unclosed-literal error. `executeQuery` (`:690-694`) has no retry and no description-free fallback, so the run fails identically every time with a raw BigQuery syntax error the user cannot act on. The `finally` drops the staging table, so the live table is not corrupted — the impact is a permanently failing import, not data loss.

**BigQuery is the only storage that does this.** `SnowflakeStorage.js:322`, `DatabricksStorage.js:339` and `AwsAthenaStorage.js:1052` all wrap descriptions in `obfuscateSpecialCharacters`; `AwsRedshiftStorage.js:778` uses `escapeDescription`. BigQuery has the identical helper at `GoogleBigQueryStorage.js:698-702` and already uses it for cell *values* at `:603`.

Two things I checked and am **not** claiming: this is not an injection primitive — literal `"` is rewritten to `'`, and `\u0022`/`\x22` are consumed as escapes producing a `"` inside the string rather than terminating it, so the worst outcome is a syntax error. And `2024\2025` is a poor example: `\202` is a valid 3-digit octal escape. The sound triggers are `C:\Reports` and any header ending in `\`.

**Fix:** `GoogleBigQueryStorage.js:217` and `:403` → wrap with the existing helper, `this.obfuscateSpecialCharacters(this.schema[columnName]["description"])`, matching the other four storages.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
```diff
@@ -111,10 +111,10 @@ export function DataMartDefinitionSettings({
   const sqlCode = getSqlQueryFromDefinition(definitionType, currentDefinition);
 
   useEffect(() => {
-    if (definitionType) {
+    if (definitionType && !isDirty) {
```
zapolsky: **[Regression — introduced by this PR]** Adding `&& !isDirty` to this guard means that once the form is dirty, switching the definition type no longer writes the new type into form state, while the resolver *is* rebuilt for the new type. Every schema branch pins `definitionType` with `z.literal`, so submit fails validation on a field that renders no message.

**Evidence chain:** the diff changes `if (definitionType)` → `if (definitionType && !isDirty)` and adds `isDirty` to the deps (verified against the merge-base, `d3fca8c69`) → this `reset(getInitialFormValues())` is the **only** writer of `definitionType` into RHF state on a type change; the other `reset` at `:120-129` is gated on `!definitionType && !initialDefinitionType && preset?.definitionType`, i.e. it only fires for the initial preset when no type is selected yet, and `handleTypeSelect` (`:148-153`) updates only the parent `useState` → `:92-95` rebuilds the resolver from `createDataMartDefinitionSchema(definitionType, storageType)` on every type change → `model/schema/data-mart-definition.schema.ts:7,16,24,32,38` pin `definitionType` with `z.literal` on all five branches → `:197` `void handleSubmit(guardedSubmit)()` is called with **no `onInvalid` handler**, and this file renders no `<FormMessage>` at all, so a failure on `definitionType` produces no visible error.

**Trigger:** new Data Mart → Input Source → **SQL** → type anything (form becomes dirty) → switch to **Connector** → complete the wizard. `ConnectorDefinitionField.tsx:62-68` does a field-scoped `trigger('definition')` which passes, then the outer submit fails on the stale `definitionType`.

**Scope of what was executed:** the guard change, the only-writer property, the `z.literal` pins and the absent `onInvalid`/`<FormMessage>` are all verified directly in the checked-out source. An independent skeptic's first test confirmed the stale value at submit time before its run was cut short by an infrastructure limit, so the *staleness* is corroborated but the full click-through was not replayed end to end — please sanity-check the exact UX symptom. That same run **refuted** a claimed milder symptom: on Table → type → View, Save was observed **enabled**, not disabled.

**Fix:** keep the type-change reset unconditional and scope the `isDirty` guard to what it was actually meant to protect (re-initialisation from `initialDefinition` while the user is editing), e.g. split into two effects keyed on `definitionType` and on `initialDefinition` respectively.

packages/connectors/src/Sources/GoogleSheets/Source.js
```diff
@@ -0,0 +1,903 @@
...
+          normalizedRow._owox_imported_at = importedAt;
+        }
+
+        columns.forEach(column => {
+          normalizedRow[column.name] = this._normalizeCellValue(row[column.index]);
+        });
+
+        return normalizedRow;
+      })
+      .filter(row => columns.some(column => row[column.name] !== null));
```
zapolsky: **[Bug — verified]** In subset mode the blank-row filter is evaluated over **only the selected columns**, so narrowing the column selection silently deletes real sheet rows — including their `_owox_row_number`, the declared unique key other Data Marts join on.

**Evidence chain:** `Source.js:321-323` — `columns` is the *selected subset* when `applySelectedFields` is true → `:330` that subset is passed to `_buildRows` → `:797` `.filter(row => columns.some(column => row[column.name] !== null))` treats a row as blank if it is empty in the selected columns, regardless of what it holds elsewhere.

Reproduced against the real `_buildRows` (vm-loaded `Source.js`), sheet `Name | Email | Notes` where only the first data row has a value in `Notes`:

```
IMPORT-ALL    _owox_row_number: [ 2, 3, 4 ]
SUBSET(notes) _owox_row_number: [ 2 ]
rows lost by narrowing the column selection: 2
```

Rows 3 and 4 exist and have data in `Name`/`Email`; they vanish from the destination table purely because the user unticked columns. So the row count in the warehouse depends on which checkboxes are ticked, which contradicts the PR's own promise that the table reflects the current sheet contents. The existing test only covers a *fully* empty sheet row, so the intended "skip blank rows" behaviour is tested but this variant is not.

**Fix:** evaluate blankness over `detectedColumns` (all columns present at the header row), not over the selected subset — keep the projection for the output row, but decide row existence from the full row.

apps/web/src/features/connectors/shared/utils/google-sheets-fields.utils.ts
```diff
@@ -0,0 +1,166 @@
...
+  selectedFields: string[],
+  availableFields: string[],
+  defaultFields: string[],
+  importAllColumns = false
+): string[] {
+  const availableFieldSet = new Set(availableFields);
+  if (importAllColumns) {
+    const selectedFieldSet = new Set(selectedFields);
+    const defaultFieldSet = new Set(defaultFields);
+    return availableFields.filter(
```
zapolsky, line 40: **[Bug — verified]** Any re-preview in the wizard silently re-checks every column the user deselected and saves `ImportAllColumns: true`, discarding their explicit column subset.

**Evidence chain:** `Source.js:17,82-89` — `ImportAllColumns` defaults to `true` for a fresh spec → `connector.service.ts:461-474` and `connector.mapper.ts:43-47,84-96` copy `default` **and** `attributes` to the wire with no hidden-spec filter → `ConfigurationStep.tsx:118-130` seeds *all* specs into wizard state (the `HIDE_IN_CONFIG_FORM` filter exists only at `:314-319` for render ordering) → `ConnectorEditForm.tsx:280-287` `handleFieldToggle` mutates only `selectedFields`, and `withGoogleSheetsImportAllColumns` is called **only** inside `onFinish` (`:767`, `:798`, `:822`), never written back into `connectorConfiguration` → `ConnectorEditForm.tsx:455-469` every forward transition out of the configuration step re-previews (the `previewConfigurationKey` memo is consulted only by `canGoNext`, and `:366` clears it before each call) → `:394-402` → **`google-sheets-fields.utils.ts:37-46`** the import-all branch returns every non-system column unconditionally; `selectedFields` is consulted only to keep *system* fields.

Reproduced at component level with the real `ConnectorEditForm` / `ConfigurationStep` / `FieldsSelectionStep`, plus a control:

```
[COMPONENT] checked BEFORE Back:      ["_owox_row_number","Campaign"]
[COMPONENT] checked AFTER Back->Next: ["_owox_row_number","Campaign","Clicks","Spend"]
[COMPONENT][BUG]     saved fields / ImportAllColumns: [.. all 4 ..] / true
[COMPONENT][CONTROL] saved fields / ImportAllColumns: [.. user's 3 ..] / false   (no Back->Next)
[EXISTING false]     AFTER Back->Next: ["_owox_row_number","Campaign"]          (already-saved subset is safe)
```

Three things worth knowing: the trigger is **any** re-preview, not just Back→Next — editing anything on the configuration step and pressing Next does it too; an *absent* `ImportAllColumns` behaves identically to `true` (`:81` tests `value !== false`), so filtering hidden specs out of the seeding loop would **not** fix this; and connectors already persisted with `ImportAllColumns:false` are protected by the `preserveSubsetMode` guard at `:110-114`, so the exposure is every **new** Google Sheets setup plus every existing one saved in all-columns mode. No existing test encodes this as intended — `google-sheets-fields.utils.test.ts:27-34` exercises this branch only where the dropped field is the *system* `_owox_imported_at`.

**Fix:** recompute `ImportAllColumns` into `connectorConfiguration` whenever the user toggles a column, so the re-preview sees subset mode — or have `resolveGoogleSheetsPreviewSelection` intersect with `selectedFields` when a previous selection exists.

apps/web/src/features/data-marts/edit/model/hooks/use-refresh-data-mart-after-connector-run.ts
```diff
@@ -0,0 +1,49 @@
...
+      }
+      return;
+    }
+
+    if (!latestSuccessfulRun || lastObservedSuccessfulRunIdRef.current === latestSuccessfulRun.id) {
+      return;
+    }
+
+    lastObservedSuccessfulRunIdRef.current = latestSuccessfulRun.id;
+    void refreshDataMart(dataMartId);
```
zapolsky, line 49: **[Bug — verified]** This hook is the first *automatic, timer-driven* caller of `refreshDataMart`, and a completing connector run now silently wipes the user's unsaved Output Schema edits **and** clears `isDirty`, so the unsaved-changes guard never warns.

**Evidence chain:** `DataMartContext.tsx:672-680` polls runs every 5 s (active) / 30 s while the tab is visible, i.e. exactly while the user is editing → `DataMartDetails.tsx:110-116` calls this hook → **`use-refresh-data-mart-after-connector-run.ts:42-47`** — the steady-state branch has **no manual-run gate** (`isManualRunTriggered` gates only the first-observation branch at `:31-40`), so a scheduled run, an API-triggered run, or a run started in another tab all fire `refreshDataMart` → `DataMartContext.tsx:83-100` → `mapDataMartFromDto` builds a fresh object and `reducer.ts:56-66` replaces `state.dataMart` wholesale, so `dataMart.schema` is a new reference every time → `DataMartDataSetupContent.tsx:88` renders the schema editor on the same page, mounted throughout → `DataMartSchemaSettings.tsx:155,158` → `useSchemaState.ts:49-57` the effect is keyed on `[initialSchema]` by **reference identity** with no value compare, and unconditionally does `setSchema(deepCloneSchema(initialSchema)); setIsDirty(false)` → the guard registered at `DataMartSchemaSettings.tsx:234-242` (`isDirty: () => isDirty`) is therefore silenced.

Captured from a probe wiring the real reducer, the real hook and the real `useSchemaState` together:

```
PROBE1 after user edit          : alias="MY UNSAVED ALIAS" isDirty=true
PROBE1 after NEW-but-EQUAL      : alias="" isDirty=false
PROBE2 scheduled run SUCCESS    : refresh calls = 1   (isManualRunTriggered:false throughout)
PROBE3 after auto-refresh       : alias="" isDirty=false
PROBE3 guard would warn?        : NO
```

Note PROBE1 line 2: the loss happens even when the refetched payload is deeply **equal** — reference identity alone is enough. And the backend really does change the schema here, since `connector-executor.service.ts:181-183` runs `actualizeSchema` right after a successful field update.

**Introduced here:** `useSchemaState`'s identity-keyed reset and the poll are both pre-existing and untouched by this PR, but the baseline has exactly one `refreshDataMart` caller — `DataMartRelationshipsContent.tsx:288`, on a user action. This PR supplies the first automatic trigger, so the exposure is new. `useSchemaState.test.tsx` asserts the opposite intent (edits must survive a new `initialSchema` after `markSchemaSaved`), so this is not intended behaviour.

**Fix:** make the reset dirty-aware — skip it while `isDirty` (or content-compare before resetting) — and/or have this hook skip `refreshDataMart` while the registered schema guard reports dirty. Worth also considering whether the Output Schema editor should be read-only during an active run; nothing gates it on `hasActiveRuns` today.

zapolsky (commented):

zapolsky:

Nothing here re-posts the findings already raised by @prokopetsv and @romandubovyi — those threads still stand and I did not re-litigate them. Seven blockers are inline; everything else is below.

**On the shape of this review:** the first pass confirmed four blockers, which triggered one extra finder wave. That wave confirmed three more — all in new code, all silent, and none covered by the existing suite. That pattern usually means the remaining unknowns are better addressed by another pass over the feature than by another round of review. Worth weighing before the next push.

Primary notes (criticality ≥ 80), highest first — none of these block merge.

- **Pre-existing, not a regression — but found while reviewing this PR and worth its own issue:** `apps/backend/src/data-marts/services/connector/connector-secret.service.ts:336` — the `_secrets_id` → Data-Mart binding is enforced only in the use-case layer and only for `'GoogleSheets'` (`update-data-mart-definition.service.ts:222`); the write path checks only `projectId`, as does `injectSecrets` at run time (`connector-credential-injector.service.ts:220`). Every file in this chain is **byte-identical to the merge-base** — this PR actually *narrows* the hole by adding the binding check that did not exist before. Verified with probes over the real `ConnectorService` (Shopify spec), `ConnectorSecretService`, `DataMartMapper`, `UpdateDataMartDefinitionService` and `ACCESS_MATRIX`: a GET returns `_secrets_id` in the clear while masking the token; `EDITOR`+`NON_OWNER`+`SHARED_FOR_REPORTING` grants `SEE`+`USE` without `EDIT`; PUTting another Data Mart's `_secrets_id` with `AccessToken: "**********"` saves without error; and the run config then resolves to the other Data Mart's decrypted token. Three details beyond the original chain: `data-mart.entity.ts:69-70` defaults `availableForReporting` to **`true`**, so nearly every Data Mart sits in the vulnerable state; `GET /data-marts/by-connector/{name}` (`data-mart.controller.ts:143-152`) is guarded only by `Role.viewer` and performs **no per-Data-Mart access check**, so a plain viewer can dump every `_secrets_id` in the project in one request; and sending a *real* token overwrites the victim's stored credential in place, breaking their runs. The plaintext token never leaks and `projectId` holds, so this is intra-project escalation across the sharing boundary, not cross-tenant. 12 connectors besides GoogleSheets are exposed. The natural fix is to generalize this PR's own `validateGoogleSheetsSecretReferences` to all connectors and add the `dataMartId`/`configId` check at the write path — both columns already exist on the credentials row and `getSecretsByDataMartAndConfig` already queries by them.
- `packages/connectors/src/Storages/AwsAthena/AwsAthenaStorage.js:341` — **if the `staging → live` rename commits server-side but the client throws, cleanup deletes the S3 objects the live table now points at.** The open question from the first pass is settled: the tables are **Iceberg** (`:241-251` `TBLPROPERTIES ('table_type'='ICEBERG')`), and AWS documents `ALTER TABLE … RENAME TO` for Iceberg as catalog-only — *"you can update the database and table name of an Iceberg managed table without affecting underlying table information"* — so the renamed live table keeps the staging `LOCATION`. Then `:342` clears `stagingTableCreated` only on the success path, the `:344-356` rollback fails because live already exists, `:483` `DROP TABLE IF EXISTS` on the vanished staging name is a no-op *success*, and `:487` `deleteS3TempFolder(stagingPrefix, true)` wipes the prefix. Reproduced on the real `replaceData`/`cleanupSnapshotResources`/`renameTable`/`dropTargetTable` with a simulated Glue/S3 encoding those two AWS rules: `live table still in catalog: true`, `S3 objects under live LOCATION: 0`, `iceberg metadata files present: 0 => table UNREADABLE`, `recoverSnapshotBackupIfNeeded(): false`. **Not permanent data loss** — `:368` deliberately preserves the backup table, so an operator can `DROP live; ALTER TABLE …__owox_backup_<runId> RENAME TO live`, and the next healthy run self-heals; the impact is a bounded outage where the table exists but every query fails. The telling asymmetry: the mirror case (`live → backup` committing without an ack) *is* handled — `backupTableCreated` stays false and the next run's recovery restores it. Only `staging → live` lacks the guard. Minimal fix: clear `stagingTableCreated` *before* issuing the rename, or re-check `tableExists(stagingTableName)` in the catch. Note `storage-snapshot-safety.test.js:293-326` explicitly locks in the destructive delete for the *server-side failure* case (where it is correct), and nothing exercises "server committed, client threw".
- `packages/connectors/src/Sources/GoogleSheets/Source.js:686` — **in subset mode a plain header rename permanently drops the column from the user's selection**, and renaming it back does not restore it. Verified end to end (vm-loaded connector + a real `ConnectorExecutorService`/`updateConnectorSourceFields` probe): the missing column is reported via `logMessage`, not `addWarningToCurrentStatus`, so `success` stays true and `IMPORT_DONE` is emitted; `Connector.js:22` then emits *matched-only* fields; `updateConnectorSourceFields` accepts the write (its `data-mart.service.ts:370-378` guard only bails when the **user** edited the definition mid-run) and preserves `ImportAllColumns:false`, so the narrowing is sticky. The connector cannot distinguish *deleted* from *renamed*, and one code path serves both. Consider treating the connector's list as authoritative only when it is a superset, or gating the write on `ImportAllColumns`.
- `apps/backend/src/data-marts/use-cases/connector/google-sheets-fields-preview.service.ts:95` — **the 50 MB preview cap is applied after the whole body is already in the API process heap, and the pre-buffer guard is dead code against Google.** I probed the real `sheets.googleapis.com`: responses come back `content-encoding: gzip` + `transfer-encoding: chunked` with **no `content-length`**, so `Number(undefined)` → `NaN` → `Number.isFinite` false → the check at `Source.js:175-180` is always skipped; when a length *is* present it is the compressed size. The real check at `:186` runs only after `HttpUtils.js:74-79` `response.text()` has materialized everything, and `JSON.parse` at `:191` roughly doubles peak memory. Unlike an import, this runs **in-process** (`connector-executor.service.ts:399` spawns a child; the preview does not). Measured: one worst-case in-cap preview took RSS from 346 MB → 614 MB and is fatal below a ~256–384 MB heap; 8 concurrent previews peaked at 1159 MB, and there is no `ThrottlerModule` or semaphore anywhere in `apps/backend/src`. Note `packages/connectors/test/GoogleSheets.test.js:397` is named *"rejects oversized responses before reading their body"* but mocks a `content-length` Google never sends. Streaming with an abort at the threshold would make the cap real.
- `apps/backend/src/data-marts/services/connector/connector-executor.service.ts:181` — **`updateRunStatus` is no longer the first statement in `finally`, so a hung schema call leaves the run `RUNNING` forever.** Confirmed by an A/B runtime test with `actualizeSchema` mocked as a never-settling promise: PR head wrote `["RUNNING"]`, merge-base wrote `["RUNNING","SUCCESS"]`. `getActualDataMartSchema` has no `AbortSignal` and no deadline, and the underlying clients are unbounded for BigQuery (`teeny-request` sets no default timeout), Athena (`athena-api.adapter.ts:81-109` polls with no attempt cap) and Snowflake (`retryTimeout` bounds connect only) — **Redshift is the exception** (`redshift-api.adapter.ts:69-97` caps at 300 s). Nothing reaps stale `RUNNING` rows: `cleanupOrphanedRuns` filters `PENDING`, the retry processor sweeps `INTERRUPTED`, and `INTERRUPTED` is set inside the very `updateRunStatus` that never runs — so the row holds one of `MAX_CONNECTOR_RUNS_PER_PROJECT` (default 3) for the whole project until a human cancels it. Blast radius is small and new: only configs that both succeed and emit `FIELDS_UPDATE`, i.e. Google Sheets only. The two call sites at `:181` and `:220` can collapse into one below `updateRunStatus`.
- `packages/connectors/src/Storages/AwsAthena/AwsAthenaStorage.js:363` — **cancel/SIGTERM orphans a full staging copy of the dataset in all five storages.** All cleanup lives in in-process `finally` blocks, but `connector-process-spawner.service.ts:65,113` spawns `detached: true` and kills the group with SIGTERM, and `packages/connectors` has **zero** signal handlers (grepped all 197 non-`node_modules` JS files; measured `SIGTERM listeners before: 0 after: 0` after loading the runner and every storage SDK). Reproduced by running the real `AwsAthenaStorage` in a detached child and killing the group mid-publish: the log printed the CREATE and the `live → backup` rename, then ended with **zero** cleanup lines. There is no sweep for staging (`listSnapshotTables` is only ever called with `'backup'`), no table expiry, and no S3 lifecycle rule, so it is one orphan per interrupted run, forever. In fairness the leak *class* is not novel — pre-PR `cleanupTempResources` was already outside any try/finally — but the leaked object is now a full copy of the dataset.
- `packages/connectors/src/Storages/Databricks/DatabricksStorage.js:624` — **Databricks and Redshift escape `'` but not `\` in cell values, while Snowflake and BigQuery double the backslash via `obfuscateSpecialCharacters`.** Read directly: `DatabricksStorage.js:624-628` does `value.replace(/'/g, "''")` then `'${value}'` → `:660` `CAST(${value} AS ${type})`; `AwsRedshiftStorage.js:932-933` is the same shape. Both files pre-date this PR and neither escaping function is in the diff — but `replaceData` is new and makes **arbitrary spreadsheet cell text** flow into them for the first time, where previously only API-shaped data did. A cell like `C:\` emits `CAST('C:\' AS STRING)`. I could not execute against either warehouse, so I am not asserting the outcome — but the asymmetry across the five storages is a fact, and escaping the quote but not the backslash is the classic injection primitive. Worth closing regardless of which warehouse actually consumes `\`.

<details>

- `packages/connectors/src/Storages/Snowflake/SnowflakeStorage.js:652` — `_owox_imported_at` is shifted by the connector host's UTC offset on Snowflake only. `Source.js:800-802` emits `"YYYY-MM-DD HH:MM:SS.sss"` with the `Z` deliberately stripped (asserted by `GoogleSheets.test.js:358`), and `:649-657` then does `new Date(<that string>)` — V8 parses the space-separated form as **local** time, so a UTC wall clock is re-interpreted as local and re-encoded. The other four storages consume the same string correctly (BigQuery `SAFE_CAST` is UTC, Athena's SerDe is pinned to that exact format, Databricks/Redshift cast in a UTC-default session). Sub-second precision is also dropped by `substring(0,19)`. Static analysis only — not executed against Snowflake.
- `packages/connectors/src/Storages/Databricks/DatabricksStorage.js:447` — snapshot batching estimates each row by building a **complete** MERGE statement, whose preamble repeats the full column list four times (`:672-703`). On a wide sheet (up to `GOOGLE_SHEETS_MAX_IMPORT_COLUMNS = 1598` with 127-byte names) one "row" is scored at ~1 MB regardless of its data, so the 14 MB budget yields ~14 rows per batch instead of 250 — a 100 000-row sheet becomes ~7 000 round-trips. `:66-71` can also throw *"A single snapshot row exceeds the … limit"* purely on preamble size, blaming the row. `AwsRedshiftStorage.js:652-664` has a milder form of the same over-count against a 90 KB budget.
- `apps/backend/src/data-marts/connector-types/connector-message/schemas/types/message-fields-update.schema.ts:7` — `fields: z.array(z.string())` has no element cap, no per-string length cap and no cross-check against the connector's declared schema, and `normalizeFieldsUpdate` only trims/dedupes before the list is written verbatim into the `definition` JSON column. Its sibling in the same file is hardened exactly as the checklist asks (`connector-executor.service.ts:637-649` allowlist + 4096-char cap + dropped-key warning + CAS). No exploit chain found — the shipped connector bounds names at 127 chars and 1598 columns — but that bound lives in the sender. `z.array(z.string().min(1).max(255)).max(2048)` would close it.
- `packages/connectors/src/Sources/GoogleSheets/Source.js:171` — the 15 s preview abort never reaches the OAuth token request: `signal` is in scope at `:160` but is not passed to `getAccessToken`, and `OAuthUtils.getServiceAccountToken` has no `signal` parameter, so `HttpUtils.js:54-56` never sets it. A stalled `oauth2.googleapis.com` can outlive the 504 by up to undici's 300 s default, pinning a socket and the decoded key per impatient retry.
- `apps/backend/src/data-marts/use-cases/connector/google-sheets-fields-preview.service.ts:85` — `createSource()` sits outside every try/catch, so malformed client JSON becomes a 500 rather than a 400. Verified by running the real `SourceConfigDto` in a vm sandbox: `{"configuration":{"X":null}}` and `{"configuration":{"X":{}}}` both throw `TypeError: Cannot convert undefined or null to object` from `recursiveValueTransform`, and `{"configuration":{"Fields":["a","b"]}}` is silently mangled into `{value:"0",items:{...}}`. `@IsObject()` alone does not constrain the contents.
- `packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js:351` — the schema-change branch is `CREATE OR REPLACE TABLE … COPY`, which BigQuery documents as dropping all row access policies, while the `preserveTable` branch (TRUNCATE + INSERT) keeps them. So RLS survives an unchanged sheet and vanishes the first time a column is added — the exact scenario this PR is built for. Snowflake does `COPY GRANTS COPY TAGS` (`:397`) and Redshift replays grants (`:607-612`).
- `packages/connectors/src/Core/AbstractStorage.js:149` — `hasSameSchema` is name+type only, so the four preserve branches never refresh column descriptions. Since `_normalizeColumnName` maps `Cost` and `Cost, $` to the same identifier, relabelling a header leaves the warehouse (and, via `bigquery-data-mart-schema.provider.ts:128`, the OWOX catalog and the MCP/AI surface) describing the old header indefinitely.
- `packages/connectors/src/Storages/Databricks/DatabricksStorage.js:406` — the only `hasSameSchema` caller that passes **no** normalizer, so it keeps type parameters that the other three strip (`GoogleBigQueryStorage.js:12-20`, `SnowflakeStorage.js:36-48`, `AwsRedshiftStorage.js:107-122`). A destination table created outside OWOX with `VARCHAR(255)`/`DECIMAL(…)` would never match `STRING`/`DOUBLE` from `DESCRIBE TABLE`, forcing the `DEEP CLONE` replace path on every refresh and discarding grants/comments/tags each run. Unreachable today because `getColumnType` never emits a parameterised type — the defect is the asymmetry.
- `packages/connectors/test/storage-snapshot-safety.test.js:104` — the test named *"BigQuery preserves the live table object when the schema is unchanged"* skips `replaceData` and calls `publishSnapshotTable` with `preserveTable` hardcoded to `true`, so the only place the decision is made (`GoogleBigQueryStorage.js:290`) is untested. Mutation-proven: forcing that argument to `false` **and** to `true` both leave 65/65 passing, while the same mutation on the Databricks equivalent (`DatabricksStorage.js:406`) correctly fails one test. BigQuery is the default storage for this connector and the sole outlier — the Databricks/Snowflake (`:484`) and Redshift (`:649`) analogues do drive `replaceData`.
- `apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx:134` — `loadSpecificationSafely` closes over the pre-delete `loadedSpecifications`, so the forced refetch in `handleConnectorSelect` (`:263-268`) is skipped and nothing resets `state.connectorSpecification` (`reducer.ts:36-46` has no reset action). Selecting A → Back → B → Back → A appears to render B's fields at step 2. Pre-existing, but this PR rewrote the handler, and `isGoogleSheetsConnector` (`:93`) stays true while a foreign spec renders. Not independently verified — worth a look.
- `apps/web/src/features/connectors/edit/components/ConnectorEditForm/ConnectorEditForm.tsx:184` — the connectors-fetch effect depends on `loading`, and `reducer.ts:33` `FETCH_CONNECTORS_ERROR` sets `loading:false` with `connectors:[]`, which looks like an unbounded refetch loop with no backoff for as long as the sheet is open when `GET /connectors` fails. Not independently verified.
- `apps/backend/src/data-marts/use-cases/connector/google-sheets-fields-preview.service.ts:34` — this use-case owns connector-runtime concerns that live in `services/connector/` everywhere else: it is the only file outside `services/`/`connector-types/` importing `@owox/connectors`, it declares the repo's first backend-side `AbstractConfig` subclass, and its `run()` takes `(context, configuration)` rather than a mapper-built command like every sibling (`fields-connector.service.ts:9-11` and the checklist reference `exchange-project-member-api-key.service.ts:18`). That layering is also why the fields-schema mapper had to be duplicated (already raised by @romandubovyi).
- `apps/backend/src/data-marts/services/connector/connector-secret.service.ts:104` — no server-side guard stops a required secret from being blanked: `''` passes both `:104-105` and `:480`. Unreachable through the UI (`ConfigurationStep.tsx:249-253` folds `hasValidGoogleSheetsServiceAccountConfiguration` into step validity and `google-sheets-fields.utils.ts:130` rejects `''`), so this is residual risk for a direct API caller only.
- `packages/connectors/src/Storages/AwsRedshift/AwsRedshiftStorage.js:839` — `createSnapshotTableNames` deliberately fills the whole 127-byte budget, then `replaceData:583` sets `DestinationTableName = stagingTableName` and `saveData:839` builds `temp_${name}_${Date.now()}` — measured 145 bytes for a 126-byte staging name. Triggers for destination tables ≳ 97 chars, and the truncated tail is exactly the disambiguator.
- `apps/backend/src/data-marts/controllers/connector.controller.ts:70` — the only credential-touching route in this controller without `@RejectApiKeyAuth()`; the three OAuth siblings at `:84`, `:96`, `:116` all carry it. **Three** independent passes agreed this is not an auth bypass (`idp.guard.ts` populates `userId`/`roles` for API-key flows so `canAccess` still enforces the per-Data-Mart ACL, and both `POST :id/manual-run` and `POST :id/access-validation` also omit it), so this is a consistency question — either add it or add a line saying API-key callers are intentionally allowed.
- `packages/connectors/src/Sources/GoogleSheets/README.md:34` — the date contract is documented as "text remains text" but not that a *native* date cell's stored value tracks its **display** format: `Source.js:164-166` requests `dateTimeRenderOption=FORMATTED_STRING` and `_inferType` has no date branch, so date columns are `STRING`. Changing the sheet locale therefore keeps `hasSameSchema` true and silently rewrites every row from `7/24/2026` to `24.07.2026` via the preserve path. Design as stated — worth one README sentence.
- `.changeset/google-sheets-source-connector.md` — filename has no Fibery task number. Roughly half of the last 30 changesets on `main` also omit it, so this is a nudge rather than a rule; content and format match the siblings exactly.
- `README.md:135` — the user-facing "Data Sources" table still lists 14 connectors with no Google Sheets row, even though `packages/test-utils/src/constants.ts:48` and `packages/test-utils/README.md:103` were both updated, and the connector ships a full `GETTING_STARTED.md` / `CREDENTIALS.md` / `logo.svg` / `manifest.json` set matching `Sources/Shopify/`.
- minor: `Source.js:271-289` `_getRetryAfterMs` returns as soon as the header *key* exists, so an empty `Retry-After` yields `0`, and `:241` uses `??`, which does not fall back on `0` — every remaining retry then fires with no delay against an API that just asked us to slow down · `AbstractConfig` has no `updateFields` declaration next to its explicit `updateCredentials` no-op (`:277`), which is why `Connector.js:28` needs `?.` where `Sources/MicrosoftAds/Source.js:310` does not — and this PR's own `GoogleSheetsFieldsPreviewConfig` overrides six base methods but not this one · snapshot table names are truncated by **character** against byte limits (`GoogleBigQueryStorage.js:319-325`, Athena `:383`, Snowflake `:31`, Databricks `:27`; Redshift already has `truncateUtf8` with a test) · Athena's `listSnapshotTables:424-425` builds its prefix from the **untruncated** name, so crash recovery silently misses tables over ~221 chars · `useConnector.ts:12-22` resets `loadingFields` on unmount only when a *preview* AbortController exists, so an in-flight static `fetchConnectorFields` leaves the provider stuck · `use-refresh-data-mart-after-connector-run.ts:20-48` never resets its refs on `dataMartId` change, firing one spurious refresh when navigating between two Sheets data marts · `replaceData` validates row counts against a deduped key set on BigQuery but raw `data.length` on the other four · the endpoint documents `ApiOkResponse` (200) but `@Post` with no `@HttpCode(200)` returns 201 (three pre-existing instances elsewhere, so it joins a pattern rather than inventing one) · `ConfigurationOneOfRender.tsx:143` hardcodes `'GoogleSheets'` instead of the exported constant used at every other new call site · `Storages/AwsRedshift/manifest.json` still declares four globals after `connector-runner.js` added `BatchExecuteStatementCommand` · `connector-fields-response-api.dto.ts:14-32` omits `uniqueKeys`/`uniqueKeysByDataLevel`/`destinationName` that the mapper returns and the FE consumes · `Source.js:213` `'authorization retry loop ended unexpectedly'` is unreachable · the GoogleSheets README claims identifiers "start with a letter or underscore" while `_normalizeColumnName` strips leading underscores.

</details>

**Not attributable to this PR:** `npx tsc -b` in `apps/backend` reports 127 errors, but the merge-base (`d3fca8c69`) reports the **same 127 errors across the same 38 files** — including the two PR-touched specs (`data-mart.service.spec.ts`, `update-data-mart-definition.service.spec.ts`), where only the line numbers shift. This is a local dependency-float artifact, not a PR regression. `eslint` is clean on all changed backend and web files; `packages/connectors/src` is ignored by that package's own config.

**Investigated and dropped as refuted:** a claimed cross-package version-skew hazard for the new `updateFields` message — `.changeset/config.json` keeps `@owox/connectors`, `@owox/backend` and `owox` in one `fixed` group and internal specs are `"*"`, so a fresh install always resolves them together (all three are 0.30.1 on npm). A claimed MySQL precision gap in the new `modifiedAt` compare-and-set — the `configuration`/`fields` comparison at `data-mart.service.ts:370-378` already covers the same-second case, and a real TypeORM + better-sqlite3 harness showed the `Raw(strftime…)` criterion behaving correctly (matching value `affected: 1`, stale value `affected: 0`), including under `TZ=Europe/Kyiv`. A concern that Athena might not support `ALTER TABLE … RENAME TO` at all — AWS documents it for Iceberg tables, which is what `AwsAthenaStorage.js:241-251` creates.

**Tests:** all green in the worktree — backend 196 tests / 18 suites, `packages/connectors` 65 / 3 files, `apps/web` 40 / 9 files, and the three touched e2e specs 48 / 3 suites. The e2e specs initially failed with 500/404 and a connector count of 14; that was an environment artifact (the worktree's symlinked `node_modules` resolved `@owox/connectors` to a main-branch `dist`), not a PR defect — after building the PR's connectors they pass 48/48. No facade impl is added or changed, so the CI facade-test gate is not engaged. Notably, none of the seven blockers is covered by the existing suite.

<sub>Automated OWOX review · worktree multi-agent passes · two finder waves · all blockers adversarially verified.</sub>

packages/connectors/src/Sources/GoogleSheets/Source.js
zapolsky: skip

packages/connectors/src/Sources/GoogleSheets/Source.js
zapolsky: skip

zapolsky (approved):

PR #1467, opened 2026-07-23
fix(api): audit HTTP Data streaming contract

max-voloshyn:
## Scope

Audits and repairs the public contract for:

- `GET /api/external/http-data/data-marts/{dataMartId}.ndjson`

Audited base: `d3fca8c69869b7ef3dae4851a40653c5971e558e`

## Coverage result

| Surface | Before | After |
| --- | --- | --- |
| OpenAPI | Unassessed | [Covered](https://app.owox.com/api/swagger-ui#/HTTP%20Data/HttpDataController_stream) · 2026-07-23 |
| API client | Unassessed | [Covered](./docs/api/api-client.md#stream-data-mart-rows) · 2026-07-23 |

The support matrix now reports 139 API-key endpoints, 11 fully covered, and 128 unassessed.

## Contract checklist

- Authentication and visibility: retains the shared `@Auth(Role.viewer())` contract, generated `X-OWOX-Authorization` security metadata, shared 401 response, Business User role requirement, and endpoint-specific `Action.USE` authorization.
- Request: documents the exact path, wildcard and exact-column projection, bounded base64url filter/sort/aggregation/dateTrunc parameters, and positive-integer limit. The generated contract test pins the public `8192` bound independently of the runtime constant.
- Response: documents `application/x-ndjson`, the `x-owox-run-id` header, and 400/401/403/404/424/503 outcomes, including project blocking, provider context, and shutdown behavior.
- API client: consolidates filter, sort, aggregation, and date-bucket rule shapes into `TraverseDataOptions`, exports the constituent rule types, verifies URL/query/auth/API-key behavior, preserves 424 provider context, and rejects successful non-NDJSON responses before traversal.
- CLI: enforces the contract's positive `--limit` minimum and keeps JSON flag parsing as an explicit untyped transport boundary while the backend remains the canonical runtime validator; the source-level migration for previously untyped client variables is recorded in the changeset.
- Documentation: repairs the streaming example, documents the media-type guard, typed shape validation, and HAVING-style aggregate filters, and validates that the support-matrix client target resolves to the checked-in guide heading.
- Ownership: shared authentication owns security and 401 metadata; the endpoint specification owns its `Action.USE` extension and HTTP Data-specific errors.

Sources of truth and independent oracles:

- Runtime query schemas and stream behavior are checked against generated OpenAPI output and focused contract tests.
- Shared auth decorators are exercised with their real metadata while only the guard execution is overridden.
- Client behavior is checked through recorded HTTP requests, response media types, stream parsing, cancellation, and error context.
- Coverage links are checked by the matrix validator plus the generated operation identity and guide-heading tests.

No exclusions or exemptions.

## Verification

- Backend OpenAPI contract: 6/6 passing
- API-client traversal tests: 10/10 passing
- API-client package typecheck: passing
- Public typed-options compile contract, including invalid-operator rejection: passing
- CLI stream tests: 6/6 passing
- CLI package typecheck: passing
- Targeted backend, API-client, and CLI lint/format checks: passing
- Documentation coverage validator: 21/21 passing; 139 endpoints / 11 fully covered
- Documentation lint, format, and Markdown lint: passing
- Changeset status and `git diff --check`: passing
- Independent full-diff and review-feedback delta reviews: no Critical or Important blockers

## Blocker register

- Review blocker: none.
- Merge blockers: checks are running on follow-up head `764b7d059`, and a fresh engineer approval is required because the existing approval targets the preceding head. The non-required `Audit all` failure reports dependency advisories already present in the unchanged lockfile, and the repository-wide backend `tsc --noEmit` command has unrelated baseline test-fixture errors at the audited base; neither references this slice.

Assignee: @max-voloshyn

🤖 Codex (GPT-5.6 Sol High)


docs/api/api-client.md
```diff
@@ -248,6 +247,11 @@ for await (const rows of data.rowChunks()) {
 ```
 
 Call `await data.cancel()` if you open a traversal and decide not to iterate `rowChunks()`.
+The client rejects a successful response whose media type is not `application/x-ndjson` instead of
+attempting to interpret it as row data.
+Use the exported `TypedTraverseDataOptions` and traversal rule types when you want compile-time
+validation of these controls. The existing `TraverseDataOptions` remains available for callers that
+need a permissive compatibility surface.
```
max-voloshyn: Split of TypedTraverseDataOptions and TraverseDataOptions seems too complex for api-client consumers. Is there a way to simplify-consolidate it? For now it's probably consumed only by owox-ctl command that may be adjusted if needed

.changeset/api-surface-maintenance.md
```diff
@@ -24,3 +24,12 @@ unavailable.
...
 `runs.list(...)`; the response and option type exports remain available.
+
+## Strengthen HTTP Data streaming contracts
+
+`GET /api/external/http-data/data-marts/{dataMartId}.ndjson` now publishes its exact-column
+projection, bounded base64url controls, positive-integer limit, NDJSON response, run identifier,
+and failure contract in OpenAPI. `@owox/api-client` rejects successful
+`dataMarts.traverseData(...)` responses that are not NDJSON before they can be consumed as rows and
+exports opt-in typed traversal controls while preserving the existing permissive options type.
+Existing valid traversal calls require no migration.
```
max-voloshyn: Information about rejection feels too detailed here

docs/api/api-client.md
max-voloshyn: Consolidated this into one public `TraverseDataOptions` type backed directly by the exported rule shapes, removed the secondary type and overloads, and adapted the `owox-ctl` JSON boundary. The changeset now calls out the source-level migration for untyped or widened variables.

🤖 Codex (GPT-5.6 Sol High)

.changeset/api-surface-maintenance.md
max-voloshyn: Shortened the release note to the public contract: typed traversal controls plus concise NDJSON media-type validation. It now also records the required migration for previously untyped control variables.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/controllers/spec/external/http-data.api.ts
```diff
@@ -143,12 +155,14 @@ export function StreamHttpDataSpec() {
...
         'filter/sort/aggregation/dateTrunc/limit, aggregation or dateTrunc without an explicit ' +
-        '`column` projection, or unsupported storage type.',
+        '`column` projection, unsupported storage type, or a project blocked because it is ' +
+        'inactive or has exceeded its credit limit.',
     }),
-    ApiResponse({ status: 401, description: 'Missing or invalid `x-owox-authorization` token.' }),
     ApiResponse({
       status: 403,
-      description: 'Caller is authenticated but lacks `Action.USE` on the requested Data Mart.',
+      description:
```
romandubovyi: ⚪ **Non-blocking:** `@Auth(Role.viewer())` already publishes a 403 description ("Forbidden. Required role: Business User") via `ApiForbiddenResponse`, and NestJS Swagger merges same-status descriptions by concatenation (`mergeResponseEntry`), so the published operation ends up with two overlapping paragraphs both stating the role. Consider keeping this Spec description focused on the endpoint-specific `Action.USE` requirement (as on `main`) and letting the shared `@Auth` decorator own the role wording.

packages/api-client/src/data-marts.ts
```diff
@@ -9,13 +9,93 @@ export type OWOXDataMart = Record<string, unknown> & {
 
 export type OWOXDataMartRow = Record<string, unknown>;
 
+export type TraverseDataAggregateFunction =
```
romandubovyi, line 14: ⚪ **Non-blocking:** These rule types (`TraverseDataAggregateFunction`, filter operators, `dateTrunc` units, relative-date presets) hand-duplicate the backend zod schemas in `apps/backend/src/data-marts/dto/schemas/` (`filter-config.schema.ts`, `aggregate-function.schema.ts`, `date-trunc-config.schema.ts`). If the backend adds an operator or function, this type will lag and reject valid calls at compile time. The repo already uses keep-in-sync markers for exactly this situation (`aggregate-function.schema.ts` opens with a "Keep this list in sync with…" note) — consider adding a cross-reference comment on both sides so drift is caught during review.

packages/api-client/src/data-marts.ts
```diff
@@ -9,13 +9,93 @@ export type OWOXDataMart = Record<string, unknown> & {
...
+  | { kind: 'this_month' }
+  | { kind: 'last_month' }
+  | { kind: 'this_year' }
+  | { kind: 'last_n_days'; n: number }
+  | { kind: 'last_n_months'; n: number };
+
+type TraverseDataScalarValue = string | number | boolean;
+type TraverseDataFilterPlacement = {
+  placement?: 'pre-join' | 'post-join';
+  function?: TraverseDataAggregateFunction;
```
romandubovyi, line 40: ⚪ **Non-blocking:** The public `TraverseDataFilterRule` exposes an optional `function` for HAVING-style filters on aggregated values (per the backend schema, the `(column, function)` pair must match a configured aggregation), but `docs/api/api-client.md` only documents `placement` for filter rules. Please document when and how `function` should be used, or exclude it from the public type if HTTP Data consumers are not expected to filter aggregated values.

🤖 Generated with Kimi CLI

apps/ctl/src/commands/data-marts/stream.ts
```diff
@@ -44,17 +48,22 @@ function parseJsonArrayFlag(
...
-    aggregation: parseJsonArrayFlag(flags.aggregation, '--aggregation'),
-    dateTrunc: parseJsonArrayFlag(flags['date-bucket'], '--date-bucket'),
+    filter: parseJsonArrayFlag<TraverseDataFilterRule>(flags.filter, '--filter'),
+    sort: parseJsonArrayFlag<TraverseDataSortRule>(flags.sort, '--sort'),
+    aggregation: parseJsonArrayFlag<TraverseDataAggregationRule>(
+      flags.aggregation,
+      '--aggregation'
+    ),
+    dateTrunc: parseJsonArrayFlag<TraverseDataDateTruncRule>(flags['date-bucket'], '--date-bucket'),
     limit: flags.limit,
```
romandubovyi, line 67: ⚪ **Non-blocking:** (Re: the `--limit` flag definition at line 103, outside this diff.) The flag uses `Flags.integer()` without a `min` constraint, so `owox-ctl` accepts `0` or negative values and relies on the backend to reject them with a 400. Since this PR now publishes `limit` as `minimum: 1` in OpenAPI, consider adding `min: 1` to the flag definition to fail fast and stay aligned with the contract.

🤖 Generated with Kimi CLI

packages/api-client/src/data-marts.ts
```diff
@@ -281,6 +361,18 @@ export class DataMartsApi {
       });
     }
 
+    const contentType = response.headers.get('content-type');
+    const mediaType = contentType?.split(';', 1)[0]?.trim().toLowerCase();
+    if (mediaType !== 'application/x-ndjson') {
+      await response.body?.cancel().catch(() => undefined);
+      throw new OWOXApiError('OWOX Data Mart data stream returned an unexpected content type', {
```
romandubovyi, line 370: ⚪ **Non-blocking:** This path throws `OWOXApiError` with only `dataMartId` and `contentType` in `details`, and never sets `status`, while every other HTTP error goes through `createHttpError` which always sets `status: response.status`. Callers that branch on `error.status` (as they can for 401/424) would see `undefined` here. Since `response` is available at this point, consider passing `status: response.status` for consistency.

romandubovyi (approved):
Review complete. 5 non-blocking comments, no blockers — approving.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Kimi CLI + Grok CLI (Grok 4.5) + Claude Code (Opus 4.8)

apps/backend/src/data-marts/controllers/spec/external/http-data.api.ts
max-voloshyn: Fixed in `764b7d059`: the endpoint-specific 403 description now covers only `Action.USE`, while `@Auth(Role.viewer())` owns the Business User wording. The OpenAPI test now verifies the composed description contains the role wording exactly once.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/data-marts.ts
max-voloshyn: Fixed in `764b7d059`: the API client now points to the canonical aggregate, filter, sort, and date-trunc schemas, and each backend schema has a reciprocal pointer to the public traversal literals.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/data-marts.ts
max-voloshyn: Documented in `764b7d059`: `function` is described as a HAVING-style filter, with a concrete example, the required matching `(column, function)` aggregation pair, and the prohibition on `placement: 'pre-join'`.

🤖 Codex (GPT-5.6 Sol High)

apps/ctl/src/commands/data-marts/stream.ts
max-voloshyn: Fixed in `764b7d059`: the CLI `--limit` flag now declares `min: 1`, with a focused regression test for alignment with the HTTP Data contract.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/data-marts.ts
max-voloshyn: Fixed in `764b7d059`: unexpected successful non-NDJSON responses now preserve `response.status` on `OWOXApiError`; the focused traversal test asserts status `200`.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/controllers/external/http-data.controller.ts
```diff
@@ -14,9 +14,9 @@ export class HttpDataController {
     private readonly streamHttpDataService: StreamHttpDataService
   ) {}
 
+  @StreamHttpDataSpec()
   @Auth(Role.viewer())
```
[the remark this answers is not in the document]
max-voloshyn: Verified on final head `764b7d059`: `@nestjs/swagger` v11's `ApiResponse` uses `mergeResponseEntry` for an existing status and concatenates both descriptions rather than replacing the first one. `SwaggerModule.createDocument` therefore emits both `Forbidden. Required role: Business User` and the endpoint-specific `Action.USE` text. The focused generated-OpenAPI test asserts both fragments, asserts the role wording appears exactly once, and passes 6/6.

🤖 Codex (GPT-5.6 Sol High)

PR #1468, opened 2026-07-23
fix(api): actualize Data Mart list contract

max-voloshyn:
## Scope

Actualize the public contract for `GET /api/data-marts` across the backend OpenAPI surface, the typed API client, focused contract tests, API documentation, the support matrix, and the canonical API-surface changeset.

Audited base: `3e71d55c13968ef945e667ec0c61f8ba31ba4af9`

## Grouping rationale

At the audited base, `DataMartsApi` exposed only `list()` for `GET /api/data-marts` plus
`traverseData()` for the separately covered HTTP Data endpoint. No other unassessed Data Marts
sibling had an existing typed API-client abstraction, so none was eligible under the controller
rule requiring both an OpenAPI specification and an API-client surface to appear to exist. The
project run-history and project insight-template endpoints already had separate client
abstractions, were already covered, and belonged to separate capability families. Adding another
Data Marts sibling would therefore broaden this implemented-but-unassessed actualization slice
into new client-surface design and make it incoherent.

Final slice size: 1 endpoint.

## Coverage result

| Endpoint | OpenAPI | API client |
| --- | --- | --- |
| `GET /api/data-marts` | Unassessed → Covered (2026-07-23), newly covered | Unassessed → Covered (2026-07-23), newly covered |

The support matrix now reports 12 of 139 public endpoints fully covered (9%), with 127 unassessed.

## Contract checklist

- Authentication and visibility: keeps the shared viewer authorization primitive and documents that the operation lists Data Marts visible to the authenticated viewer; the independent OpenAPI test asserts the shared auth contract.
- Request validation: adds a presentation DTO for an optional non-negative decimal-integer `offset` (default `0`) and optional `ownerFilter` enum (`has_owners` or `no_owners`), with focused rejection tests for negative, fractional, blank, whitespace, exponent, hexadecimal, and unknown values.
- Pagination: documents the service's shared 1,000-item page size; the client follows `nextOffset` automatically, preserves `ownerFilter` across pages, and rejects repeated offsets.
- Response contract: publishes required fields, nullability, integer constraints, enums, nested storage/user/context objects, availability flags, and RFC 3339 date-time formats. Draft `definitionType` values may be null, and optional user metadata follows the backend's nullable string contract; focused OpenAPI, mapper, and client tests use independent expectations and malformed-response cases.
- Generated operation identity: asserts `DataMartController_list`, the `DataMarts` tag, query constraints, response refs, and component requirements.
- Typed client: exports list options, owner/status/definition/storage enums and nested object types; validates options before authentication or network I/O while retaining the existing extensible `OWOXDataMart` shape.
- Documentation: expands the list guide with exact options, automatic paging, validation, and response typing; coverage links resolve to the specific Swagger operation and checked-in guide heading.
- Release note: reuses `.changeset/api-surface-maintenance.md`; existing `dataMarts.list()` calls remain compatible and require no migration.

## Verification

- API client focused suites: 5 suites, 102 tests passed.
- API client: typecheck, ESLint, and Prettier checks passed.
- Backend focused suites: 4 suites, 22 tests passed, including SQLite/MySQL TypeORM metadata generation.
- Backend: Nest source build, ESLint, and Prettier checks passed. An explicit `varchar` column type preserves nullable definition-type metadata across supported test dialects. ESLint reports one unrelated pre-existing warning in `sql-clause-renderer.spec.ts`.
- API coverage validator: 23 tests passed; checked-in matrix validated at 139 endpoints and 12 fully covered.
- Changed Markdown lint, Changesets status, `git diff --check`, and the excluded `docs/api/openapi.md` check passed.
- Independent full-diff and review-repair delta reviews: no remaining actionable blocking findings.

## Blocker register

- Review blockers: none.
- Merge blockers: only the confirmed baseline `Audit all` failure remains; all case-affected GitHub checks passed and final-head approval is present. `Audit all` fails on 10 existing production-dependency advisories; the audited `main` commit's immediately preceding run fails the same workflow, and this case changes no dependency manifest or lockfile. A broad backend `tsc --noEmit` invocation also reports pre-existing spec diagnostics outside this case; the backend source build for the changed production surface passes.
- Exemptions: none.

🤖 Codex (GPT-5.6 Sol High)


packages/api-client/src/validation.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function isOptionalNullableEmail(value: unknown): boolean {
+  return (
+    value === undefined ||
+    value === null ||
+    (typeof value === 'string' && /^[^\s@]+@[^\s@]+$/.test(value))
+  );
+}
+
+function isOptionalNullableUri(value: unknown): boolean {
```
romandubovyi: 🔴 **Blocking:** `isOptionalNullableUri` (avatar) requires `new URL()` to parse and `isOptionalNullableEmail` (line 19) enforces an email regex, but the backend contract only guarantees `z.string().optional().nullable()` (`apps/backend/src/data-marts/dto/schemas/user-projection.schema.ts:7`). A relative/proxied avatar path, an empty-string email, or any non-standard identifier is valid per the server contract, yet the client will throw `OWOXApiError` and abort the entire `dataMarts.list()` call. The client validator should match the contract it validates (accept any string), or the backend schema should be tightened to match — not diverge silently.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash (Medium)) + Claude Code (Opus 4.8)

packages/api-client/src/validation.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  try {
+    return new URL(value).protocol.length > 1;
+  } catch {
+    return false;
+  }
+}
+
+export function isUserProjection(value: unknown): value is UserProjectionShape {
```
romandubovyi, line 19: ⚪ **Non-blocking:** `isUserProjection` is shared into two already-shipped endpoints — `runs.ts` (`isNullableProjectDataMartRunUser`, line 170) and `insight-templates.ts` (`isOptionalNullableUser`, line 49) — which previously validated `email`/`avatar` as any string via `isOptionalNullableString`. Routing them through the stricter regex + `new URL()` check silently changes behavior for endpoints outside this PR's stated Data Mart-list scope: a previously-accepted `runs.list()` response can now throw `OWOXApiError`. Please confirm the IdP always emits absolute avatar URLs and regex-valid emails, or keep the looser `isOptionalNullableString` check on these two endpoints and apply the strict form only to the new list contract.

romandubovyi (commented):
Multi-agent review (agy + claude + grok). Both findings converge on the new client-side validation being stricter than the backend contract for user `email`/`avatar`.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash (Medium)) + Claude Code (Opus 4.8) + Grok CLI (grok-4.5)

packages/api-client/src/data-marts.ts
```diff
@@ -112,16 +175,79 @@ export type JsonRequester = {
...
+    typeof value.title === 'string' &&
+    typeof value.status === 'string' &&
+    DATA_MART_STATUSES.has(value.status) &&
+    typeof value.storage.type === 'string' &&
+    DATA_STORAGE_TYPES.has(value.storage.type) &&
+    typeof value.storage.title === 'string' &&
+    (typeof value.description === 'string' || value.description === null) &&
+    (value.definitionType === undefined ||
+      (typeof value.definitionType === 'string' &&
+        DATA_MART_DEFINITION_TYPES.has(value.definitionType))) &&
```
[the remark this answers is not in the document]
max-voloshyn: Fixed in `58c1594bd`. The persisted entity, list DTO, OpenAPI component, exported client type, and runtime guard now represent `definitionType` as optional-nullable. A mapper regression proves a persisted draft `null` reaches the list response, while the separate non-list DTO path normalizes it to `undefined`.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/validation.ts
max-voloshyn: Fixed in `58c1594bd`. The shared client validator now accepts any optional-nullable string for `email` and `avatar`, matching `UserProjectionSchema`. The shared OpenAPI DTO no longer claims unsupported email/URI formats, and regression cases cover empty email and relative avatar strings.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/validation.ts
max-voloshyn: Fixed in `58c1594bd`. The shared validator preserves the previously shipped string-only behavior, and focused regression tests now prove opaque creator metadata remains accepted by Data Mart lists, project run history, and insight templates.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/data-marts.ts
```diff
@@ -112,16 +175,80 @@ export type JsonRequester = {
...
 };
 
-function isRecord(value: unknown): value is Record<string, unknown> {
-  return typeof value === 'object' && value !== null && !Array.isArray(value);
+const DATA_MART_STATUSES = new Set<string>(DATA_MART_STATUS_VALUES);
+const DATA_MART_DEFINITION_TYPES = new Set<string>(DATA_MART_DEFINITION_TYPE_VALUES);
+const DATA_STORAGE_TYPES = new Set<string>(DATA_STORAGE_TYPE_VALUES);
+const DATA_MART_OWNER_FILTERS = new Set<string>(DATA_MART_OWNER_FILTER_VALUES);
+
+function isUser(value: unknown): value is OWOXDataMartUser {
```
romandubovyi, line 183: ⚪ **Non-blocking:** `isUser` is a one-line wrapper around the shared `isUserProjection`, and `OWOXDataMartUser` duplicates `UserProjectionShape` from `validation.ts`. Since the projection validator is now shared, consider calling `isUserProjection` directly and aliasing `OWOXDataMartUser = UserProjectionShape`, so the two identical shapes can't drift apart.

romandubovyi (approved):
Review complete — no blocking issues. The substantive concerns from earlier rounds (nullable `definitionType` for drafts, over-strict avatar/email validation) are already resolved in `58c1594bd`, and the client's strict list validation is consistent with the backend mapper output (all newly-required fields are guaranteed by entity defaults). One non-blocking cleanup noted inline.

packages/api-client/src/data-marts.ts
max-voloshyn: Thanks. I’m keeping the exported `OWOXDataMartUser` name and the narrow wrapper in this contract-focused slice so the public Data Mart surface remains self-describing without broadening the approved change into a shared-type refactor. Runtime validation already delegates to `isUserProjection`; the alias cleanup can be considered separately.

🤖 Codex (GPT-5.6 Sol High)

PR #1470, opened 2026-07-27
fix(connectors): classify customer-actionable failures as warnings

prokopetsv:
# Problem

Every connector failure was recorded as an ERROR, regardless of cause. That
made the production alert on `severity="ERROR"` +
`jsonPayload.context="ConnectorExecutorService"` page on-call for problems no
engineer can act on — a customer's Facebook session expiring, a TikTok
advertiser they lost access to, a dead refresh token, or a run the customer
cancelled themselves.

Run logs were also close to unreadable. The connector runner passed raw errors
to `console.error`, and the backend's capture splits stdout and stderr on every
newline, so a single crash exploded into one log entry per stack-trace line —
19 entries for one failure. Analysis of a 3.5-day production export showed
roughly 85% of alert volume was these fragments, and TikTok logged every error
twice on top of that.

# Solution

Failures are classified at the point they are thrown, using signals the
provider actually gives us, and that classification is carried end to end.

- Connector sources mark customer-actionable errors with `isWarning`.
  `AbstractSource` sets it from a new overridable `_isAuthError`, defaulting to
  HTTP 401/403; Facebook overrides it because its credential errors arrive as
  HTTP 400 with `error.type: 'OAuthException'`.
- The Facebook override deliberately excludes codes already listed in
  `FB_RETRYABLE_ERROR_CODES`. Facebook reuses `OAuthException` for throttling
  (80004) and outages (code 2), and exhausting retries on those is a real
  failure worth alerting on — not a credentials problem.
- `connector-runner` emits one structured JSON line on crash, typed as either
  `error` or `addWarningToCurrentStatus`. Because the payload is JSON-encoded,
  the stack trace's newlines are escaped and the whole thing stays one entry.
- The backend stops force-wrapping every stderr line as ERROR, handles the
  WARNING message type, and downgrades to `logger.warn` for cancelled runs and
  expired storage credentials. Warnings still land in `configErrors`, so the
  "finished without terminal success status" fallback does not double-report.
- Diagnostics move from raw `console.log`/`console.error` to
  `config.logMessage`, which JSON-encodes them. Retry decisions that previously
  printed seven separate lines now report the same values in one entry, and
  TikTok stack traces are retained for genuine errors while omitted for
  warnings, where the message is the whole story.

The `isWarning` flag is a plain property, so any `throw new Error(...)` on the
path silently drops it. Google Ads and the OAuth service-account helper both
did exactly that, which is why they now hand the flag across explicitly, with
tests pinning that behaviour.

Note that this changes run outcomes: TikTok runs where every storage write failed
previously reported success with the data silently missing, and now fail. The
changeset calls this out so the resulting failure notifications aren't a surprise.

# Changes

- Added `AbstractSource._isAuthError` and set `error.isWarning` on the
  non-retryable throw path
- Added a Facebook `_isAuthError` override matching `OAuthException` while
  excluding retryable codes and subcodes
- Flagged `invalid_grant` as a warning in `OAuthUtils` and Microsoft Ads, and
  preserved the flag through the Google Ads and service-account rewraps
- Flagged TikTok advertiser permission and deleted-advertiser errors, and
  classified the all-advertisers-failed aggregate from its constituent errors
- Replaced the runner's `main().catch(console.error)` with a single structured
  JSON line typed by classification
- Stopped `ConnectorOutputCaptureService` overwriting the type of structured
  stderr messages
- Handled `ConnectorMessageType.WARNING` in `ConnectorExecutorService` and
  downgraded cancelled runs, expired storage credentials, missing config ids
  and the redundant status flag to `logger.warn`
- Rendered warnings as `WARNING` with a friendly `Warning` label in run history
- Consolidated Facebook, Reddit and X Ads retry diagnostics into one entry each
- Restored TikTok failure stack traces via `_logFailure`, kept for real errors
  and omitted for warnings
- Removed duplicate TikTok error logging and the Facebook field-list array dump
- Moved recoverable notices off stderr so they are no longer recorded as run
  errors
- Added connector tests for auth classification, retryable-code exclusion,
  warning propagation through rewraps, and `_logFailure`
- Fixed TikTok marking an advertiser successful after its storage write failed, which
  let a run that lost all its data report success (pre-existing on `main`)


apps/backend/src/data-marts/services/connector/connector-executor.service.ts
```diff
@@ -286,6 +297,18 @@ export class ConnectorExecutorService {
                 configId,
               });
               break;
+            case ConnectorMessageType.WARNING:
```
laskevych, line 321: 🔴 **Blocking:** this branch is what puts the warning envelope into `run.errors`, and `extractRunErrorMessage` (`apps/backend/src/data-marts/utils/run-error-message.ts:4`) can't read it — it only looks at `error` / `message` / `msg`, so a warning falls through to the raw-string fallback.

Concretely, for the scenario this PR is built around: an expired Facebook session is persisted as `{"type":"addWarningToCurrentStatus","at":"…","warning":"HttpRequestException: Session has expired.\n    at …"}` and the run ends FAILED. `FailedRunsAllDmNotification.collectQueueItems` picks up FAILED runs, `buildQueuePayload` copies `item.errors` verbatim, and `getEmailContent` maps each entry through `extractRunErrorMessage` — which returns the whole JSON string and then truncates it at `MAX_ERROR_LENGTH = 300`. The customer's failure email now shows an escaped JSON blob with a stack trace in it instead of "Session has expired".

Before this PR the same failure was stored as `{"type":"error","error":"…"}` (wrapped by `captureError`) and extracted cleanly, so this is a regression on exactly the path the PR is trying to make more readable.

`mcp-reports.facade.impl.ts:812` (`toErrorMessage`) has the same gap. Adding `?? parsed['warning']` to the lookup chain covers both — worth a test that feeds it a real persisted warning entry.

apps/backend/src/data-marts/services/connector/connector-executor.service.ts
```diff
@@ -260,7 +271,7 @@ export class ConnectorExecutorService {
       const configId = (config as Record<string, unknown>)._id as string;
 
       if (!configId) {
-        this.logger.error(
+        this.logger.warn(
```
laskevych: 🟡 **Important:** a configuration missing `_id` isn't customer-actionable — it's a data-integrity defect in the stored data mart definition, so it belongs with the other ERROR-severity logs rather than with cancelled runs and expired credentials.

It also silently loses the run: the loop `continue`s without pushing to `configurationResults`, so if that's the only configuration, `successCount` is 0, `capturedErrors` is empty, and `updateRunStatus` writes `status = FAILED` with `errors = null`. Nothing in run history, nothing in the failure email — this `logger.error` was the last remaining signal.

I'd keep it at `error` here, and ideally also push a message into `configErrors` so the run can explain itself.

packages/connectors/src/connector-runner.js
```diff
@@ -146,4 +146,10 @@ async function main() {
 }
 
 // Execute main and handle errors
-main().catch(console.error);
+main().catch(error => {
+  console.error(JSON.stringify({
+    type: error?.isWarning ? 'addWarningToCurrentStatus' : 'error',
+    at: new Date().toISOString(),
+    [error?.isWarning ? 'warning' : 'error']: error?.stack ?? String(error),
```
laskevych: 💡 **Suggestion:** the crash handler puts `error.stack` into the `warning` field, which contradicts the rule `_logFailure` establishes in `TikTokAds/Connector.js` — customer-actionable failures are "fully described by their message, and a stack there is just noise".

The fixture pinned in `utils.test.ts` shows the result: the run-history warning a customer reads is `HttpRequestException: Error validating access token: Session has expired.\n    at FacebookMarketingSource._validateResponse (/app/index.cjs:436:11)`. Combined with the 300-char truncation in failure emails, the readable part can get cut off entirely.

`error?.message ?? String(error)` on the warning branch and `error?.stack` on the error branch would line up with `_logFailure`.

packages/connectors/src/Sources/TikTokAds/Connector.js
```diff
@@ -127,23 +124,26 @@ var TikTokAdsConnector = class TikTokAdsConnector extends AbstractConnector {
...
+            this._logFailure('Error saving data to storage', storageError);
             this._trackAdvertiserError(advertiserId, storageError);
           }
         }
 
-        // Mark this advertiser as having at least one successful operation
-        this.advertiserSuccesses.set(advertiserId, true);
+        // Only a completed save counts: marking success after a failed write would let a
+        // run where every write failed finish as successful, silently losing the data
+        if (saved) {
```
laskevych, line 142: 🟡 **Important:** gating `advertiserSuccesses.set()` on a completed save is a genuine fix and I'd keep it — but it changes run outcomes for existing users and isn't mentioned in the changeset or the "Changes" list.

A TikTok run where every storage write fails used to finish SUCCESS (each advertiser was marked successful right after the swallowed storage error) with the data silently lost. It now throws "All advertisers failed to import data" and lands as FAILED. Users whose runs were quietly "succeeding" will start receiving failure notifications on release with nothing explaining why.

Worth a line in the changeset.

packages/connectors/src/Core/AbstractSource.js
```diff
@@ -87,6 +87,7 @@ var AbstractSource = class AbstractSource {
         }
         catch (error) {
           if (!this._shouldRetry(error, attempt)) {
+            error.isWarning = this._isAuthError(error);
```
laskevych: 🔵 **Question:** this assigns unconditionally, so it overwrites a flag a deeper layer already set — the same fragility the PR description calls out for the `throw new Error(...)` rewraps.

Nothing breaks today: TikTok's provider errors don't pass through `urlFetchWithRetry`. But the moment any source with below-fetch classification routes through it, `_isAuthError` returns false for a payload-level error with no `statusCode`, the flag is cleared, and a permission failure starts paging on-call again — with no test covering it.

Was the overwrite deliberate, or would `error.isWarning = error.isWarning || this._isAuthError(error)` be safer?

laskevych (commented):
Reviewed the classification path end to end: `isWarning` → the runner's envelope → `captureError` preserving the type → the `WARNING` branch in `ConnectorExecutorService` → `configErrors` (so the "finished without terminal success status" fallback doesn't double-report) → the UI split into Warnings / Error Output. That chain holds together, and `splitJsonObjects` is strictly no worse than the old `}{` split — escaped quotes, braces inside string values and back-to-back envelopes all behave, and the spawner already delimits by newline before any of it runs.

One blocking issue: the new warning envelope isn't understood by `extractRunErrorMessage`, so the customer-facing failure email for the PR's headline scenario regresses to a raw JSON blob. Details inline. The rest are non-blocking.

Also verified and deliberately **not** flagged, in case it saves a round trip:
- `WARNING` not setting `success = false` is correct — run status comes from `successCount`, so TikTok partial successes stay SUCCESS.
- `logError` only needs `NodeJsConfig`; there is no other `AbstractConfig` subclass in the repo.
- The `'owox': minor` changeset is sufficient — every package sits in the `fixed` group in `.changeset/config.json`.

packages/connectors/src/Sources/TikTokAds/TiktokMarketingApiProvider.js
```diff
@@ -61,12 +61,18 @@ class TiktokMarketingApiProvider {
...
+            // stdout, not stderr: a retry notice is not a run error. This class has no
+            // config reference, so it cannot use logMessage.
+            console.log("TikTok Marketing API rate limit exceeded. Retrying...");
             await AsyncUtils.delay(backoff);
             backoff *= 2;
             continue;
           }
-          throw new Error(`TikTok API error: ${jsonData.message}`);
+          const error = new Error(`TikTok API error: ${jsonData.message}`);
+          // ponytail: message match — TikTok's Business API docs don't publish
```
laskevych: 🧹 **Nit:** stray `ponytail:` at the start of the comment — looks like an editing artifact. The caveat underneath is worth keeping, just without the prefix.

packages/connectors/src/Sources/TikTokAds/Connector.js
```diff
@@ -292,6 +294,26 @@ var TikTokAdsConnector = class TikTokAdsConnector extends AbstractConnector {
    * @param {Error} error - The error that occurred
    * @private
    */
+  _logFailure(context, error) {
```
laskevych, line 301: 📝 **Docs:** `_logFailure` was inserted directly beneath the existing "Track an error for a specific advertiser" JSDoc block, so that block now documents this method (the duplicate added below documents `_trackAdvertiserError` correctly).

Editor tooltips will show `_logFailure(context, error)` described with `@param {string} advertiserId`, which makes it easy to pass an advertiser ID as the first argument and end up with entries like `6812345: TikTok API error: …`. Move the new method above the block, or give it its own.

packages/connectors/src/Sources/FacebookMarketing/Source.js
```diff
@@ -255,35 +255,82 @@ var FacebookMarketingSource = class FacebookMarketingSource extends AbstractSour
...
+   *
+   * Facebook also reuses that same type for throttling, outages and anything it
+   * marks is_transient, which are not credential problems. Rather than restate
+   * which of those are transient, defer to the retry logic: if it would have
+   * retried, exhausting the attempts is a real failure worth alerting on.
+   *
+   * @param {HttpRequestException} error - The error to check
+   * @return {boolean} True if this is an authentication/authorization failure
+   */
+  _isAuthError(error) {
```
laskevych, line 329: ✅ **Praise:** the `_evaluateRetry` / `_isAuthError` split is the right call. Deferring "is this a credentials problem?" to the retry logic instead of restating which Facebook codes are transient means the two can't drift, and the 80004 / code-2 test cases pin exactly the case that would otherwise have silenced a real alert.

apps/backend/src/data-marts/services/connector/connector-executor.service.ts
prokopetsv: Confirmed — good catch, and you're right that it's a regression on exactly the path this PR is meant to improve.

Reproduced with the persisted envelope from a real run:

```
{"type":"addWarningToCurrentStatus","at":"2026-07-27T14:16:21.365Z","warning":"HttpRequestException: Error validating access token: Session has expired.\n    at FacebookMarketingSource._validateResponse (/app/index.cjs:436:11)..."}
```

`extractRunErrorMessage` fails all three lookups, returns the raw string, and the email renders that blob truncated at 300 chars instead of "Error validating access token: Session has expired."

Two changes in this PR put it there: `captureError` no longer force-wraps stderr as `{type:'error', error:…}`, and the executor now persists WARNING-typed entries for cancellations and expired storage credentials. Both land in `run.errors`, which is by design — a warning still counts as a run failure so the "finished without terminal success status" fallback doesn't double-report — and that's precisely what puts them on the notification path.

**Fixed** by adding `?? parsed['warning']` to the chain in `run-error-message.ts`, with a comment explaining why the field belongs there. Both call sites you named (`FailedRunsAllDmNotification` and `toErrorMessage` in `mcp-reports.facade.impl.ts`) go through this one helper, so the single change covers both.

**Tests** added to `run-error-message.spec.ts`: the real persisted Facebook envelope, asserting the extracted text contains the message and no longer leaks `addWarningToCurrentStatus` or a `{"type"` prefix into the truncated window, plus the cancelled-configuration warning. 7 pass.

I also swept for any other consumer of persisted run errors. The only other reader is `data-mart.mapper.ts`, which passes `run.errors` through verbatim to the web UI — and that side already handles the warning type via `parseLogEntry`/`isPersistedWarning`. The remaining hits (`looker-studio-report-run.service.ts`, `base-run-trigger-handler.service.ts`) write errors rather than interpret them, so there's no third gap.

One observation worth carrying forward: the existing spec already pinned this fallthrough as intended behaviour (`{detail: 'not supported'}` → returns the raw JSON), which is likely why nobody noticed a new field had joined the envelope vocabulary. That's the same contract-drift shape as the other findings in this review — a message type added at one end, with the consumers never told. If you think it's worth it, I can add a shared constant or type for the envelope field names so the next addition fails loudly instead of silently degrading.

apps/backend/src/data-marts/services/connector/connector-executor.service.ts
prokopetsv: Agreed on both counts, and the second half is the more serious one.

You're right that I misclassified this. A stored configuration without an `_id` is our data being wrong — a migration gap or a bad write — not something the customer can act on. It belongs with the other ERROR-severity logs. That downgrade came from my original audit of the nine `logger.error` sites, and this one didn't meet the "customer-actionable" bar I was applying.

The silent-loss part is exact. `continue` sits above the `configurationResults.push`, so a lone bad configuration leaves the results array empty: `successCount` is 0, `capturedErrors` is empty, and `mergeWithPersistedOutput` ends with `errors: mergedErrors.length > 0 ? mergedErrors : null` — so the run persists as FAILED with `errors = null`. Run history shows nothing and the failure email has no errors section, which made that `logger.error` the only remaining trace.

Fixed both ways:

- Restored `this.logger.error` with the same metadata.
- The guard now pushes a `ConfigurationExecutionResult` with `success: false` and a single `ERROR`-typed message before continuing, so the run explains itself in run history and in the email. `successCount` stays 0, so the run status is unchanged.

Added a spec case that runs a data mart whose only configuration has no `_id`, asserting the persisted `errors` is not null and names the defect.

packages/connectors/src/connector-runner.js
prokopetsv: Good suggestion — taken, and it turns out the stack was pure duplication.

`AbstractConnector` already writes the full stack via `config.logMessage` at both of its catch sites (`Core/AbstractConnector.js:33` and `:130`) before rethrowing, so it lands in `logs[]` as its own structured entry. The crash handler then wrote it a second time into the warning envelope. A real run JSON shows both copies. Dropping it from the warning branch therefore loses no diagnostic information — it just stops duplicating it into the customer-facing field.

Changed the handler to split the two branches as you described: `error?.message ?? String(error)` for warnings, `error?.stack ?? String(error)` for errors. That now matches the rule `_logFailure` follows, so there's one rule for the concept instead of two.

I also refreshed the fixtures in `utils.test.ts`, since they'd otherwise pin a shape the runner no longer emits. The warning fixture is message-only, the error fixture carries the stack, and the "multi-line content survives as one entry" assertion moved to the error envelope where it still reflects reality. Added a case asserting a warning starts with the readable message and contains no stack frame.

One correction on the truncation argument, since it affects how you'd weigh similar cases later: V8 formats `Error.stack` as `Name: message\n    at …`, so the message is always at the front and the 300-char cut keeps it — the run JSON you'd have seen confirms this ordering. The real cost was that the remaining ~230 characters were spent on frame paths rather than useful context. A message can still be truncated, but only when the message itself is long (the TikTok "All advertisers failed…" aggregate is the realistic case), and that's independent of the stack.

packages/connectors/src/Sources/TikTokAds/Connector.js
prokopetsv: You're right, and thanks for catching the release-communication gap rather than just the code.

Traced it to be sure: previously the storage error was swallowed, `advertiserSuccesses.set(advertiserId, true)` ran anyway, so `_checkAndReportErrors` saw `hasSuccess === true`, put every advertiser in `successfulAdvertisers`, and threw nothing. No failure, no warning either — the partial-failure branch needs `failedAdvertisers.length > 0`, which was also zero. The run then took the connector's `IMPORT_DONE` and landed SUCCESS with the data missing. With the gate in place those runs throw "All advertisers failed to import data" and land FAILED.

Added a paragraph to the changeset:

> TikTok runs that lose data now fail instead of reporting success. Previously a run could finish successfully even when every write to your storage failed. You lost the data with no warning. Such runs now fail and notify you, so you can find and fix the cause. If TikTok runs start failing after this release, check the run history for the storage error behind them.

I'll add the matching line to the PR "Changes" list too.

One extra consequence worth flagging, since it lands on us rather than on users: a storage write failure carries no `isWarning`, so the aggregate is ERROR-typed. These newly-failing runs will page on-call as well as notify the customer. I think that's correct — a BigQuery write failure is genuinely worth investigating, and "Could not serialize access to table … due to concurrent update" was one we explicitly kept at ERROR earlier in this review. But it does mean the alert volume drop this PR predicts will be partly offset on release, and it's better to expect that than to discover it.

packages/connectors/src/Core/AbstractSource.js
prokopetsv: Not deliberate — I wrote the assignment without thinking about a deeper layer having already classified. Your read is right, and `||` is the safer form. Changed to:

```js
error.isWarning = error.isWarning || this._isAuthError(error);
```

Your risk assessment matches what I found tracing it: nothing sets the flag below `urlFetchWithRetry` today, because every current classification site (`OAuthUtils.getAccessToken`, `MicrosoftAds.getAccessToken`, TikTok's provider) calls `HttpUtils.fetch` directly and bypasses it. The one source that does route auth through it is CriteoAds, and it relies entirely on `_isAuthError`. So it's latent rather than live — but it's the same failure mode the PR description calls out for the rewraps, and it would fail silently and in the paging direction.

On `||` versus `??`: `||` also lets a `false` from below be promoted, and I think that's correct rather than sloppy. A deeper `false` generally means "not one of the cases I recognise" — TikTok's provider, for instance, matches two message patterns and returns `false` for everything else — so a genuine 401/403 should still be able to classify it. A deeper `true` is a positive assertion made with more context, and that's the direction that must not be lost. Noted that reasoning in a comment at the assignment.

Added three cases to `AbstractSource.test.js` driving the real `urlFetchWithRetry` with a stubbed `HttpUtils.fetch`: a pre-flagged error keeps its flag, an unflagged 401 gets promoted, and an ordinary 500 stays unflagged. That covers the gap you identified.

packages/connectors/src/Sources/TikTokAds/TiktokMarketingApiProvider.js
prokopetsv: Fixed — you're right that it's an artifact. `ponytail:` is a marker from my own editing convention for flagging a deliberate shortcut; it carries no meaning in this repo and shouldn't have shipped. Rewrote it as:

```js
// Matched on message text because TikTok's Business API docs don't publish
// reliable numeric codes for these; switch to jsonData.code if they ever do
```

Grepped the rest of the branch for the same prefix — that was the only one.

packages/connectors/src/Sources/TikTokAds/Connector.js
prokopetsv: Confirmed and fixed — I inserted the method between an existing JSDoc block and the method it belonged to, then added a fresh block for `_trackAdvertiserError` below without noticing the original was now orphaned onto `_logFailure`.

Gave `_logFailure` its own block rather than moving it, since it needed real documentation anyway:

```js
/**
 * Record a failure that did not stop the import
 *
 * Routes by classification: customer-actionable failures become warnings, everything
 * else stays an error so it is still alerted on.
 *
 * @param {string} context - What was being attempted, e.g. "Error saving data to storage"
 * @param {Error} error - The failure, carrying isWarning when it has been classified
 * @private
 */
```

`_trackAdvertiserError` keeps the correct duplicate below it, so there's no longer a stray block.

Your point about the failure mode is the part I'd have missed: a tooltip advertising `@param advertiserId` for a parameter that's actually a context string is worse than no docs, because it invites exactly the call that produces `6812345: TikTok API error: …` — and nothing would catch it, since both parameters are strings and the output still looks plausible.

laskevych (approved):
All review findings addressed in b0d527eb — the warning envelope is now readable in failure emails and MCP status, the missing-_id run explains itself, the TikTok behavior change is in the changeset, and the isWarning flag survives every layer with tests pinning it. The failing Audit check is a pre-existing npm-audit failure on main, unrelated to this PR.

PR #1473, opened 2026-07-28
feat(data-marts): allow sorting reports by Unique Count

prokopetsv:
# Problem

Reports can add a Unique Count column — a synthetic `COUNT(DISTINCT <primary key>)` metric — but there was no way to sort by it. The Sort section of Output settings only lists schema-backed columns, so Unique Count never appeared in the "Add sort by" picker. Even a hand-crafted config was rejected at save time with `SORT_COLUMN_NOT_SELECTED`: `validateSort` checks each sort rule against the projected column set (`columnConfig` or the native field names), which never contains the synthetic `Unique Count` label. Users who wanted rows ranked by distinct-key volume had to export the report and sort it elsewhere.

# Solution

Sorting by the metric is now permitted end to end, across both the flat and the blended (joined-data-mart) query paths.

**Query builders.** The five flat builders already handled this: each renders the aggregated `ORDER BY` through `buildAggregatedAliasResolver`, which falls back to `quoteIdentifier(col)` for any column absent from the alias map — producing exactly the `"Unique Count"` alias that `renderAggregatedSelect` emits. There is a **sixth** user of that resolver, `abstract-blended-query-builder.ts`, and it needed work. Its outer `ORDER BY` was already correct, but upstream of it `referencedColumns` unions the sort columns verbatim, so the label flowed through `collectMainReferences` into the `main` raw CTE and emitted `SELECT "Unique Count" FROM <main table>` — a column that does not exist, failing every run and every Generated SQL preview. Synthetic labels are now dropped from that union (`Row Count` too, which is the same trap one validator line away), with a `columnSet` guard so a real column legitimately owning the name still reaches the CTE.

**Backend validation.** `Unique Count` is registered as an always-*known* output column, while its *selected* status is gated on `uniqueCountConfig`. Both additions use `Set.add`, so a real schema field owning that name is a harmless union rather than an override. This classifies a stale sort left after disabling the toggle as `SORT_COLUMN_NOT_SELECTED` rather than routing it to the harsher disconnected-columns path — a code-level distinction only, since the web renders `message` and not `details.errors[].code`.

**Frontend.** The synthetic column is kept in a dedicated sort-only list rather than the shared `dropdownColumns`. That shared list also feeds Filters (via `allColumns`) and Aggregations, and because the pseudo-column types as `INTEGER` it would otherwise pass `isFilterableType` and resolve a non-empty aggregation set — offering filter and aggregation options the backend then rejects. The `OutputSettingsDropdown` prop was renamed `selectedColumns` → `sortColumns` (it was already consumed only by `SortSection`); behaviour-preserving, but it makes the sort-only contract explicit so the list cannot silently be wired into the filter surfaces again. A single `hasUniqueCountMetric` gate is shared by the sort list and the disconnected-controls badge so the two cannot drift.

**Name collisions.** A real schema field may legitimately be named `Unique Count` (quoted identifiers on Snowflake/Redshift/Athena permit spaces, and the backend has a dedicated `OUTPUT_COLUMN_NAME_COLLISION` error for it). The real field always wins: the disconnected check resolves a selected real field before considering the synthetic case, and the sort list suppresses the synthetic whenever any known field owns the name — when selected it would duplicate the picker entry and collide on `FieldSearchPicker`'s `key={item.value}`; when merely present it would make `ORDER BY "Unique Count"` ambiguous between the outer alias and the base column, whose resolution precedence is unspecified across dialects.

**Stranded state.** Removing the mart's primary key hides the Unique count toggle, so the existing effect auto-clears a stored `uniqueCountConfig: true` that the user can no longer reach. That effect now prunes a stranded `Unique Count` sort rule in the same update — without it, `validateSort` rejects the leftover rule and every save and scheduled run fails. Skipped when a real field owns the name.

Row Count has the same synthetic-column shape and remains unsortable; only its blended-CTE trap is closed here.

# Changes

- Excluded `Unique Count` / `Row Count` from the blended builder's `referencedColumns`, so the synthetic labels are never projected into the `main` raw CTE while the outer `ORDER BY` still references them.
- Added `Unique Count` to the known-output-column set in `validateForReport`, and to the selected-column set passed to `validateSort` when `uniqueCountConfig` is enabled.
- Added `UNIQUE_COUNT_LABEL` / `ROW_COUNT_LABEL` to the existing web mirror `shared/utils/aggregation-labels.ts` and extended its CI drift-guard test.
- Added a sort-only `sortColumns` list carrying the synthetic column, leaving `dropdownColumns` and `selectedDropdownColumns` schema-pure so Filters and Aggregations are unaffected.
- Renamed the `OutputSettingsDropdown` prop `selectedColumns` → `sortColumns` and documented it as sort-only.
- Extracted a shared `hasUniqueCountMetric` gate used by both the sort list and the disconnected-controls badge.
- Replaced the disconnected-controls sort predicate so a selected real field resolves the rule before the synthetic special case, and suppressed the synthetic whenever any known field owns the name.
- Pruned stranded `Unique Count` sort rules in the primary-key-loss auto-heal effect.
- Added query-builder tests asserting the `main` CTE excludes the synthetic label while the outer `ORDER BY` keeps it, plus a real-column carve-out test.
- Added backend tests covering accept-when-enabled and reject-when-disabled for sorting by `Unique Count`.
- Added frontend tests covering: connected/disconnected rendering of a `Unique Count` sort rule, presence in "Add sort by", absence from "Add filter" and "Add aggregation", four real-field name-collision regressions, and the PK-loss prune.

### DIFF
--- a/apps/backend/src/data-marts/services/output-controls-validator.service.ts
+++ b/apps/backend/src/data-marts/services/output-controls-validator.service.ts
@@ -557,8 +557,12 @@ export class OutputControlsValidatorService {
         }
         // Unique Count is always a KNOWN sort target (like a schema field) — whether it's
         // currently SELECTED (projected) depends on uniqueCountConfig, checked below by
-        // validateSort. Keeps a stale sort-by-Unique-Count after disabling the toggle a
-        // SORT_COLUMN_NOT_SELECTED error, not a misleading "disconnected from schema" one.
+        // validateSort. So a stale sort-by-Unique-Count left after disabling the toggle is
+        // classified SORT_COLUMN_NOT_SELECTED (a 400) rather than routed to the harsher
+        // DISCONNECTED_REPORT_COLUMNS path, which is reserved for names absent from the
+        // schema entirely. NOTE: this is a code-level classification only — the web renders
+        // just `message`, not `details.errors[].code`, so the two read the same to the user.
+        // The web prunes this rule when the PK disappears, keeping the 400 largely unreachable.
         knownOutputColumns.add(UNIQUE_COUNT_LABEL);
 
         if (parsedFilters.length > 0) {
--- a/apps/backend/src/data-marts/services/output-controls-validator.service.ts
+++ b/apps/backend/src/data-marts/services/output-controls-validator.service.ts
@@ -555,6 +555,11 @@ export class OutputControlsValidatorService {
           homeFieldTypes.set(blended.name, blended.type);
           knownOutputColumns.add(blended.name);
         }
+        // Unique Count is always a KNOWN sort target (like a schema field) — whether it's
+        // currently SELECTED (projected) depends on uniqueCountConfig, checked below by
+        // validateSort. Keeps a stale sort-by-Unique-Count after disabling the toggle a
+        // SORT_COLUMN_NOT_SELECTED error, not a misleading "disconnected from schema" one.
+        knownOutputColumns.add(UNIQUE_COUNT_LABEL);
 
         if (parsedFilters.length > 0) {
           const fieldIndex = buildBlendedFieldIndex(blendableSchema);
@@ -577,6 +582,9 @@ export class OutputControlsValidatorService {
           // selection. Validate sort against that same native-only set so a sort on a
           // blended column is caught here at save time instead of failing at run time.
           const selectedSet = new Set(args.columnConfig ?? connectedNativeNames);
+          // Unique Count is a synthetic metric column (COUNT(DISTINCT <pk>)), not a
+          // projected field — allow sorting by it whenever it's enabled.
+          if (args.uniqueCountConfig === true) selectedSet.add(UNIQUE_COUNT_LABEL);
           errors.push(...this.validateSort(parsedSort, selectedSet));
         }
         if (parsedAggregations.length > 0) {
--- a/.changeset/6764-sort-by-unique-count.md
+++ b/.changeset/6764-sort-by-unique-count.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Sort by Unique Count
+
+Reports can now sort by the Unique Count metric. Previously, enabling Unique Count added the column to a report, but it couldn't be used as a sort column — the Sort section had no way to reference it. Now, once Unique Count is enabled, it shows up as an option in the Sort section like any other column, in ascending or descending order.
--- a/apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.spec.ts
+++ b/apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.spec.ts
@@ -2155,6 +2155,40 @@ describe('AbstractBlendedQueryBuilder — post-join aggregation', () => {
     expect(sql).toContain('COUNT(DISTINCT main.user_id) AS `Unique Count`');
     expect(sql).toContain('GROUP BY');
   });
+
+  // Regression: `Unique Count` is an OUTER-SELECT alias, not a column of any CTE. The
+  // sort-derived refs feed the main raw CTE projection, so an unfiltered sort column would
+  // emit `SELECT \`Unique Count\` FROM main_table` — a column that does not exist.
+  it('does NOT project the synthetic Unique Count label into the main raw CTE when sorted by it', () => {
+    const { sql } = builder.buildBlendedQuery({
+      ...buildContext([spendChain()], ['channel', 'spend__cost']),
+      aggregations: [{ column: 'spend__cost', function: 'SUM' }],
+      uniqueCount: true,
+      primaryKeyColumns: ['user_id'],
+      sort: [{ column: 'Unique Count', direction: 'desc' }],
+    });
+
+    const mainCte = /main AS \(([\s\S]+?)\n {2}\)/m.exec(sql);
+    expect(mainCte).not.toBeNull();
+    expect(mainCte![1]).not.toContain('Unique Count');
+    // The PK is still projected (Unique Count aggregates it) and the outer ORDER BY still works.
+    expect(mainCte![1]).toContain('user_id');
+    expect(sql).toContain('COUNT(DISTINCT main.user_id) AS `Unique Count`');
+    expect(sql).toContain('ORDER BY\n  `Unique Count` DESC');
+  });
+
+  // A real main-mart column legitimately named `Unique Count` must still reach the CTE —
+  // it arrives via `columns`, so the synthetic exclusion must not strip it.
+  it('still projects a REAL main column named "Unique Count" into the main raw CTE', () => {
+    const { sql } = builder.buildBlendedQuery({
+      ...buildContext([spendChain()], ['Unique Count', 'spend__cost']),
+      sort: [{ column: 'Unique Count', direction: 'desc' }],
+    });
+
+    const mainCte = /main AS \(([\s\S]+?)\n {2}\)/m.exec(sql);
+    expect(mainCte).not.toBeNull();
+    expect(mainCte![1]).toContain('Unique Count');
+  });
 });
 
 describe('AbstractBlendedQueryBuilder — regression: ambiguous column in WHERE/ORDER BY', () => {
--- a/apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
+++ b/apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
@@ -13,6 +13,7 @@ import {
   SqlParameter,
 } from '../utils/sql-clause-renderer';
 import { FilterRule } from '../../dto/schemas/filter-config.schema';
+import { ROW_COUNT_LABEL, UNIQUE_COUNT_LABEL } from '../../dto/schemas/aggregation-labels';
 import { DateTruncUnit } from '../../dto/schemas/date-trunc-config.schema';
 import { buildTimeZoneMap } from '../utils/date-trunc-maps.utils';
 import { effectiveComparisonType } from '../field-aggregation';
@@ -164,6 +165,15 @@ export abstract class AbstractBlendedQueryBuilder implements BlendedQueryBuilder
       // must project the PK columns even when they aren't a selected/filtered/sorted column.
       ...(context.uniqueCount === true ? (context.primaryKeyColumns ?? []) : []),
     ]);
+    // Row Count / Unique Count are OUTER-SELECT aliases, not columns of any CTE. A sort (or
+    // HAVING) on one would otherwise flow through collectMainReferences into the main raw CTE
+    // and emit `SELECT "Unique Count" FROM <main table>` — a column that does not exist, so
+    // every run and Generated SQL preview fails in the warehouse. Dropped here rather than
+    // per-source so filters and any future ref source are covered too. A real column that
+    // legitimately owns the name arrives via `columns`, so keep it when it is selected.
+    for (const label of [UNIQUE_COUNT_LABEL, ROW_COUNT_LABEL]) {
+      if (!columnSet.has(label)) referencedColumns.delete(label);
+    }
 
     const roots = this.buildTree(chains);
 
--- a/apps/backend/src/data-marts/services/output-controls-validator.service.spec.ts
+++ b/apps/backend/src/data-marts/services/output-controls-validator.service.spec.ts
@@ -3489,6 +3489,10 @@ describe('OutputControlsValidatorService', () => {
       }
 
       expect(caught).toBeDefined();
+      // Assert the class before getResponse(): if a regression routes this to the
+      // BusinessViolationException (disconnected-columns) path instead, that surfaces as a
+      // readable assertion diff rather than `TypeError: caught.getResponse is not a function`.
+      expect(caught).toBeInstanceOf(BadRequestException);
       const response = caught!.getResponse() as { details: { errors: { code: string }[] } };
       expect(response.details.errors.some(e => e.code === 'SORT_COLUMN_NOT_SELECTED')).toBe(true);
     });
--- a/apps/backend/src/data-marts/services/output-controls-validator.service.spec.ts
+++ b/apps/backend/src/data-marts/services/output-controls-validator.service.spec.ts
@@ -3433,6 +3433,65 @@ describe('OutputControlsValidatorService', () => {
         })
       ).resolves.toBeUndefined();
     });
+
+    it('accepts sorting by "Unique Count" when uniqueCountConfig is true', async () => {
+      const capabilitySvc = makeCapabilityService(true);
+      const schemaSvc = makeBlendableSchemaService([
+        { name: 'id', type: 'INTEGER', isPrimaryKey: true },
+        { name: 'name', type: 'STRING' },
+      ]);
+      const validator = new OutputControlsValidatorService(
+        capabilitySvc as never,
+        schemaSvc as never
+      );
+
+      await expect(
+        validator.validateForReport({
+          storageType: supportedStorageType,
+          dataMartId: 'dm-1',
+          projectId: 'proj-1',
+          columnConfig: null,
+          filterConfig: null,
+          sortConfig: [{ column: 'Unique Count', direction: 'desc' }],
+          limitConfig: null,
+          uniqueCountConfig: true,
+          accessor: { userId: 'user-1', roles: ['admin'] },
+        })
+      ).resolves.toBeUndefined();
+    });
+
+    it('rejects sorting by "Unique Count" when uniqueCountConfig is not enabled → SORT_COLUMN_NOT_SELECTED', async () => {
+      const capabilitySvc = makeCapabilityService(true);
+      const schemaSvc = makeBlendableSchemaService([
+        { name: 'id', type: 'INTEGER', isPrimaryKey: true },
+        { name: 'name', type: 'STRING' },
+      ]);
+      const validator = new OutputControlsValidatorService(
+        capabilitySvc as never,
+        schemaSvc as never
+      );
+
+      let caught: BadRequestException | undefined;
+      try {
+        await validator.validateForReport({
+          storageType: supportedStorageType,
+          dataMartId: 'dm-1',
+          projectId: 'proj-1',
+          columnConfig: null,
+          filterConfig: null,
+          sortConfig: [{ column: 'Unique Count', direction: 'desc' }],
+          limitConfig: null,
+          uniqueCountConfig: false,
+          accessor: { userId: 'user-1', roles: ['admin'] },
+        });
+      } catch (e) {
+        caught = e as BadRequestException;
+      }
+
+      expect(caught).toBeDefined();
+      const response = caught!.getResponse() as { details: { errors: { code: string }[] } };
+      expect(response.details.errors.some(e => e.code === 'SORT_COLUMN_NOT_SELECTED')).toBe(true);
+    });
   });
 
   describe('validateHavingFilters (post-aggregation)', () => {
--- a/apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
+++ b/apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts
@@ -159,6 +159,66 @@ describe('ReportSqlComposerService — aggregations wiring', () => {
     );
   });
 
+  // Regression: the renderer omits the Unique Count metric when there is no PK, so a stored sort
+  // on that label must be dropped too — otherwise ORDER BY references a column the SELECT lacks.
+  it('drops a "Unique Count" sort rule when uniqueCountConfig is true but the PK was removed', async () => {
+    const { service, queryBuilderFacade } = createService();
+    const report = buildReport({
+      uniqueCountConfig: true,
+      sortConfig: [
+        { column: 'channel', direction: 'asc' },
+        { column: 'Unique Count', direction: 'desc' },
+      ],
+      dataMart: {
+        id: 'dm-1',
+        projectId: 'proj-1',
+        definition: { type: 'table', fullyQualifiedName: 'p.d.t' },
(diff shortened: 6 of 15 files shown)


apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -577,6 +582,9 @@ export class OutputControlsValidatorService {
           // selection. Validate sort against that same native-only set so a sort on a
           // blended column is caught here at save time instead of failing at run time.
           const selectedSet = new Set(args.columnConfig ?? connectedNativeNames);
+          // Unique Count is a synthetic metric column (COUNT(DISTINCT <pk>)), not a
+          // projected field — allow sorting by it whenever it's enabled.
+          if (args.uniqueCountConfig === true) selectedSet.add(UNIQUE_COUNT_LABEL);
```
zapolsky, line 591: **[Bug — verified]** This line makes a `Unique Count` sort persistable, and on the **blended (joined-data-mart) path** that sort rule is projected into the `main` raw CTE as a physically non-existent column — so every run and every "Generated SQL" preview of such a report fails in the warehouse.

The PR body's premise — *"The SQL layer already supported this, so no query-builder changes were needed. All five builders render the aggregated `ORDER BY` through `buildAggregatedAliasResolver`"* — holds for the five **flat** builders. There are **six** non-test users of that resolver; the sixth is `abstract-blended-query-builder.ts:232`, which was not audited. Its `ORDER BY` is indeed correct; the breakage is upstream of it, in the CTE it builds first.

**Evidence chain:**
1. `apps/web/.../ReportColumnPicker/ReportColumnPicker.tsx:798-799` — `hasUniqueCountMetric = hasPrimaryKey && outputControlsAvailable && uniqueCountConfig`. `hasPrimaryKey` (`:480-483`) reads only the **home** mart's native fields, so it is fully independent of joined fields, while `dropdownColumns` (`:774-787`) deliberately includes `includedBlendedFields`. There is no blended carve-out → `:805-817` appends the synthetic and `:1179` hands it to the sort picker, which persists the literal string `'Unique Count'` (`OutputSettingsDropdown.tsx:507-514`).
2. **This line** (`output-controls-validator.service.ts:587`) puts `UNIQUE_COUNT_LABEL` into `selectedSet`, so `validateSort` (`:239-247`) accepts it; `:562` puts it into `knownOutputColumns`, so the disconnected check (`:730-734`) also passes. Both added lines are load-bearing and there is no blended-specific guard.
3. `use-cases/run-report.service.ts:243` calls `resolveBlendingDecision` unconditionally → `services/blended-report-data.service.ts:59` re-validates (passes, same reason) → `:138-142` `hasBlendedColumns` is true → `:204`/`:209` pass `sort: report.sortConfig` and `uniqueCount: true` into `buildBlendedQuery`. Nothing strips synthetic labels; `assertNoOrphanedColumnReferences` (`:382-400`) inspects only `columnConfig`, never `sortConfig`.
4. `data-storage-types/interfaces/abstract-blended-query-builder.ts:162` — `referencedColumns` unions `...(context.sort ?? []).map(s => s.column)` verbatim. (`:163-165` already special-cases `uniqueCount` to add PK columns, but never excludes the label itself.)
5. `:179` → `collectMainReferences` (`:492-505`) keeps every referenced column that is not a blended output alias; `'Unique Count'` is not one (`mapOutputAliasesToRoot:582-595` maps only `chain.blendedFields`) → it enters `mainColumns`.
6. `:180-187` → `buildRawCte('main', …)` → `:457-458` `safeForExplicitProjection` is true (a space, but no dot) → `:485-486` emits it as a quoted projection against the physical main table. **Note `:180` runs before the `if (aggregated && renderer)` branch at `:207`** — so the correct outer `ORDER BY` is not a mitigation.

**Captured SQL** (real `AbstractBlendedQueryBuilder` + `BigQueryClauseRenderer`, `uniqueCount: true`, `primaryKeyColumns: ['id']`, `sort: [{column:'Unique Count',direction:'desc'}]`):

```sql
main AS (
    SELECT
      `Unique Count`,     -- does not exist in main_table
      customer_name,
      id
    FROM main_table
),
...
ORDER BY
  `Unique Count` DESC
```

The identical report **without** the Unique Count sort correctly projects only `customer_name, id`, so it is unambiguously the sort entry that poisons the CTE. BigQuery: `Unrecognized name: Unique Count`. Athena/Trino: `Column 'Unique Count' cannot be resolved`. Snowflake/Redshift: `invalid identifier`. Databricks: `UNRESOLVED_COLUMN`.

**Newly reachable — verified against the merge base.** Restoring the merge-base (`2680781ee`) validator and replaying the same blended config gives `REJECTED` (`Cannot build report SQL. Disconnected columns: "Unique Count".`); on this head it is `ACCEPTED`. So this is a new regression, not pre-existing behavior.

**Trigger:** a joined data mart whose home schema has a primary key → report with an explicit `columnConfig` containing ≥1 joined column (needed for `hasBlendedColumns` at `:143`; with `columnConfig: null` it falls back to the harmless native path) → tick **Unique count** → add **Unique Count** to Sort → Save succeeds (200) → Run or open Generated SQL fails. *Masked* only when some `mainColumn` contains a dot (nested BigQuery RECORD path), since `:457-463` then widens the CTE to `SELECT *`. Flat schemas break — i.e. always on Snowflake/Redshift/Athena/Databricks.

**Fix** — exclude synthetic output labels from the sort-derived refs in `abstract-blended-query-builder.ts:159-166`. A real column legitimately owning the name still arrives via `...columns`:

```ts
const syntheticLabels = new Set(
  [UNIQUE_COUNT_LABEL, ROW_COUNT_LABEL].filter(l => !columnSet.has(l))
);
const referencedColumns = new Set<string>([
  ...columns,
  ...postJoinFilters.map(f => f.column),
  // Synthetic metrics are outer-SELECT aliases, not columns of any CTE — projecting
  // one into `main` emits a column that does not exist.
  ...(context.sort ?? []).map(s => s.column).filter(c => !syntheticLabels.has(c)),
  ...(context.uniqueCount === true ? (context.primaryKeyColumns ?? []) : []),
]);
```

Including `ROW_COUNT_LABEL` is defensive: Row Count is not sortable today, but it is the same trap one validator line away.

**Test gap that let this through:** both new backend tests use `columnConfig: null`, which routes to the native path (`blended-report-data.service.ts:107`), so the blended case has zero coverage — and there is no query-builder-layer assertion anywhere in the PR. Please add a spec beside `abstract-blended-query-builder.spec.ts:1155` (which already documents the general mechanism: *"projects native main columns referenced only by filter/sort into the main raw CTE"*) asserting the `main` CTE does **not** contain `` `Unique Count` `` while the outer `ORDER BY` still does.

zapolsky (commented):
🔴 **Blocked issue** — see the inline comment. It should be addressed before merge, but this review does not formally block; a human owns the merge decision. Posted by automated OWOX review.

zapolsky:

One blocking issue is posted inline. Notes below do not block merge.

- `apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx:491-496` — **the PK-loss auto-heal no longer fully heals.** That effect exists specifically to escape "a trap the user cannot escape through the UI": when the mart's primary key is removed, the Unique count toggle stops rendering (`:1329`) so the user cannot clear a stored `uniqueCountConfig: true` by hand, and the effect clears it for them. This PR adds a second stranded artifact the effect does not prune — the `Unique Count` sort rule — and `validateSort` now rejects it (`output-controls-validator.service.ts:584-588`, the `uniqueCountConfig === true` branch is no longer taken), so every save and every scheduled run keeps failing with `SORT_COLUMN_NOT_SELECTED`. Recoverable (the row renders struck-through with a working "Remove sort", `SortRow.tsx:84-92`), hence non-blocking. Consider pruning `sortConfig` entries equal to `UNIQUE_COUNT_LABEL` in the same `onOutputConfigChange` call that clears the flag — guarded on `!knownFieldNames.has(UNIQUE_COUNT_LABEL)` so a real field owning the name is left alone.
- `apps/backend/src/data-marts/services/output-controls-validator.service.ts:558-562` — **the stated rationale for the unconditional `knownOutputColumns.add` is not observable by the user.** The comment justifies it as yielding "a precise `SORT_COLUMN_NOT_SELECTED` rather than a misleading 'disconnected from schema' error". Code-level that is accurate, but nothing in `apps/web` reads `details.errors[].code` — `showApiErrorToast.ts:19-26` renders only `message` and `errorDetails.error`, and grepping `apps/web/src` for `details.errors` / `SORT_COLUMN_NOT_SELECTED` returns zero hits. So the user moves from the actionable sentence in `disconnected-report-columns.error.ts:11-14` ("Disconnected columns: `"Unique Count"` … Uncheck them to remove them from the report") to a bare `"Output controls validation failed"`, and a failed scheduled run records only that string. The FE also still labels this exact state `aria-label='Column not found in schema'`, so FE wording and BE classification now disagree. Either map the codes to text in `showApiErrorToast`, or prune locally (previous note) so the 400 is unreachable — and drop the "better error" claim from the comment either way.
- `apps/web/.../ReportColumnPicker/output-controls-display.ts:1-2` — **the new constant bypasses the repo's designated mirror and its CI drift guard.** `apps/web/src/features/data-marts/shared/utils/aggregation-labels.ts:3-9` already exists as *the* web mirror of `apps/backend/src/data-marts/dto/schemas/aggregation-labels.ts` (its docblock names that file and already mentions `Row Count` / `Unique Count`), and it carries a drift-guard test that pins the values so a silent edit fails CI (`aggregation-labels.test.ts:5-11`). `ReportColumnPicker/` already imports from it (`AggregationRow.tsx:7`, `AggregationEditorPopover.tsx:15`). Putting `UNIQUE_COUNT_LABEL` in a component-local display helper with only a prose comment leaves two files claiming to mirror the same backend module and no guard on the new copy. Moving it there and adding two lines to the existing guard test costs nothing.

<details>

- `apps/backend/src/data-marts/services/output-controls-validator.service.ts:562` — `Row Count` asymmetry. It is the other synthetic label emitted by the same renderer (`sql-clause-renderer.ts:266-268`, automatic for every aggregated report), is not registered here, and a sort on it therefore still routes to the *harsher* `DISCONNECTED_REPORT_COLUMNS` — not even the `SORT_COLUMN_NOT_SELECTED` this change argues for. Out of scope per the PR body, but the registry that would cover both already exists 50 lines below: `validateOutputColumnNames` (`:389-405`) computes exactly "dimensions + aggregated labels + Row Count + Unique Count", and its docblock claims it "can never drift from the SELECT". Extracting that name list and feeding it to the sort `selectedSet` would replace both ad-hoc `Set.add` calls. Please also see the inline blocker — `ROW_COUNT_LABEL` is the same blended-CTE trap one validator line away.
- **Question (unproven):** with a *real, unselected* main-table column named `Unique Count` and the toggle on, the FE appends the synthetic anyway (`ReportColumnPicker.tsx:810` suppresses it only when the real field is **selected**) and `validateOutputColumnNames` inspects only *projected* names, so no `OUTPUT_COLUMN_NAME_COLLISION` fires. The emitted `ORDER BY "Unique Count"` is then ambiguous between the outer SELECT alias and the base column of the FROM. We could not establish per-dialect resolution precedence from the code, so we are not claiming a failure — but the semantics are unspecified and the three new collision tests all use a *selected* real field. Is that combination intentionally left undefined?
- `apps/backend/src/data-marts/services/output-controls-validator.service.ts:584` — **pre-existing, not introduced here**, but on the line this PR edits: for a metrics-only report (`uniqueCountConfig: true`, `columnConfig: null`) the sort is validated against `connectedNativeNames`, while the *same function* 55 lines later knows the projection is empty (`:639-644`, `isMetricsOnly → projectedColumns = []`). So a sort on a native column saves clean and then emits `ORDER BY` against a query with no such SELECT item and no GROUP BY. Mirroring `:639-644` in the sort branch would close it.
- `apps/web/.../ReportColumnPicker/ReportColumnPicker.tsx:815` — the synthetic column's `type: 'INTEGER'` is both dead and dialect-inaccurate: nothing on the sort path reads `.type` (`columnToPickerItem` projects only `value/label/dataMartName/path`), and the real header type is `integerTypeFor(storageType)` — `INT` on Databricks (`field-aggregation.ts:63-83`). Since `OutputSettingsDropdownColumn.type` is plain `string`, the "orderable but not filterable/aggregatable" invariant rests on a JSDoc comment. A dedicated `Pick<OutputSettingsDropdownColumn, 'name' | 'label' | 'dataMartName' | 'path'>` for the sort surface would drop the fabricated value and turn `allColumns={sortColumns}` into a compile error instead of a comment violation.
- PR description — "All five builders render the aggregated `ORDER BY` through `buildAggregatedAliasResolver`". There are **six** non-test users of it; the sixth is `abstract-blended-query-builder.ts:232`, which is exactly the one the inline blocker is about. The five flat builders were audited and the blended one was skipped.
- Tests — a few of the new cases do not gate the change. `OutputSettingsDropdown.test.tsx:184` and `:204` pass verbatim on `main` (that file's diff is a pure prop rename with zero logic, and both tests hand-pass `sortColumns`); `ReportColumnPicker.test.tsx:1147` and `:1161` also pass against the reverted production logic, because old and new predicates agree whenever `hasUniqueCountMetric` is false. Relatedly, the description's "reordered … so a real field is never hijacked" is untestable: the new predicate is a pure disjunction (`disconnected ⇔ !(selected && known) && !(label && hasUC)`), so evaluation order cannot change the result. The two tests that genuinely gate the diff are `ReportColumnPicker.test.tsx:1041` and `:1091`; the two backend tests are real regression tests (each dies under mutation of `:562` / `:587`).
- minor: `SortSection`'s prop is still named `selectedColumns` (`OutputSettingsDropdown.tsx:417,421`) while its local `sortColumns` (`:422`) means the *already-sorted* columns — so `:426` reads `selectedColumns.filter(c => !sortColumns.has(c.name))` with both names inverted from the caller's vocabulary; the empty state still says "All selected columns are already sorted" (`:503-505`) although Unique Count is not a selected column · the comment at `ReportColumnPicker.tsx:796-799` says it is the "Same gate as the toggle row below, so the two can never disagree", but `:1329` re-derives `hasPrimaryKey && outputControlsAvailable` independently — extract one shared const if the invariant is meant to hold · `hasUniqueCountMetric` is typed `boolean | null | undefined` despite the `has*` name (`uniqueCountConfig` is `boolean | null`), and its `outputControlsAvailable` term is dead weight since both consumers already sit behind that gate (`:1138`, `:1174`) · the reject test (`output-controls-validator.service.spec.ts:3480-3492`) casts the caught value to `BadRequestException` and calls `getResponse()` without asserting the class, so a regression surfaces as `TypeError: caught.getResponse is not a function` rather than an assertion diff · the two new backend tests omit `aggregationConfig`, a *required* property of the `validateForReport` args type (`:428`) — invisible only because `apps/backend/tsconfig.build.json` excludes `**/*spec.ts` and `apps/backend` has no `type-check` script (matches 53 sibling calls in that file, so convention, not regression).

</details>

**Not attributable to this PR:** the `Audit all` CI job fails on this head and also on `main`. A local `tsc -b` in `apps/backend` reports ~133 errors, but `origin/main` reports ~140 with the same shape in the same files — a stale-workspace-build artifact of the review environment, not a signal about this PR. Backend specs are excluded from `tsconfig.build.json` anyway; the authoritative signal is CI on this head, where ESLint, Prettier, all unit suites, API E2E (3 shards) and Browser E2E are green.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
prokopetsv: Confirmed and fixed. Your chain reproduces end to end, including the part the PR body got wrong: there are **six** non-test users of `buildAggregatedAliasResolver`, and I audited five. `abstract-blended-query-builder.ts:232` was the gap.

**Reproduced.** I added a spec beside the one you pointed at and captured the real SQL from `AbstractBlendedQueryBuilder` plus the test renderer, with `uniqueCount: true`, `primaryKeyColumns: ['user_id']`, and `sort: [{ column: 'Unique Count', direction: 'desc' }]`. Before the fix:

```sql
main AS (
    SELECT
      `Unique Count`,   -- does not exist in main_table
      channel,
      date,
      user_id
    FROM main_table
),
```

After the fix, the same config emits:

```sql
main AS (
    SELECT
      channel,
      date,
      user_id
    FROM main_table
),
...
SELECT
  main.channel AS `channel`,
  SUM(spend.spend__cost) AS `spend__cost | SUM`,
  COUNT(DISTINCT main.user_id) AS `Unique Count`
FROM main
LEFT JOIN spend ON main.date = spend.date
GROUP BY
  main.channel
ORDER BY
  `Unique Count` DESC
```

The PK is still projected for the `COUNT(DISTINCT ...)`, and the outer ORDER BY still resolves to the aggregate alias. Your reading of line 179 running before the `if (aggregated && renderer)` branch at line 207 is correct — the outer ORDER BY was never a mitigation.

**Fix applied** in `abstract-blended-query-builder.ts`, with one deviation from your patch. I drop the labels from the assembled union rather than filtering the sort source, so filters/HAVING and any future ref source are covered by the same three lines instead of one call site:

```ts
// Row Count / Unique Count are OUTER-SELECT aliases, not columns of any CTE. A sort (or
// HAVING) on one would otherwise flow through collectMainReferences into the main raw CTE
// and emit `SELECT "Unique Count" FROM <main table>` — a column that does not exist, so
// every run and Generated SQL preview fails in the warehouse. Dropped here rather than
// per-source so filters and any future ref source are covered too. A real column that
// legitimately owns the name arrives via `columns`, so keep it when it is selected.
for (const label of [UNIQUE_COUNT_LABEL, ROW_COUNT_LABEL]) {
  if (!columnSet.has(label)) referencedColumns.delete(label);
}
```

`ROW_COUNT_LABEL` is included per your point that it is the same trap one validator line away. The `!columnSet.has(label)` guard preserves a real column that legitimately owns the name, exactly as in your version.

Note that this makes the feature **work** on the blended path rather than blocking it — the generated SQL is now valid there — so I did not add a blended-specific rejection in the validator.

**Tests** — two new cases in `abstract-blended-query-builder.spec.ts`, both at the query-builder layer you asked for:

- `does NOT project the synthetic Unique Count label into the main raw CTE when sorted by it` — asserts the main CTE lacks the label while the outer ORDER BY keeps it. Fails on the unfixed builder.
- `still projects a REAL main column named "Unique Count" into the main raw CTE` — pins the `columnSet` carve-out.

79/79 passing in that suite.

prokopetsv:
Reply to the non-blocking notes: thanks — three of these were right and are fixed; two details I disagree with.

**Fixed**

1. **Constant bypassed the designated mirror + its drift guard.** Correct, and I had not found `shared/utils/aggregation-labels.ts`. Moved `UNIQUE_COUNT_LABEL` *and* `ROW_COUNT_LABEL` there, repointed the import, deleted the component-local copy, and extended the existing guard test with a `pins the synthetic output-column labels` case. One mirror, one guard.

2. **PK-loss auto-heal no longer fully heals.** Correct, and worse than cosmetic — with the flag cleared, `validateSort` rejects the leftover rule, so every save and scheduled run fails. The effect now prunes `sortConfig` entries equal to the label in the same `onOutputConfigChange`, guarded on `!knownFieldNames.has(UNIQUE_COUNT_LABEL)` as you suggested. (The effect moved below `knownFieldNames` — referencing it from the old position put it in the dep array before initialisation, a TDZ error.) Two tests: the prune, and the real-field-owns-the-name carve-out.

3. **The "better error" claim is not user-observable.** Correct — `grep` for `details.errors` / `SORT_COLUMN_NOT_SELECTED` in `apps/web/src` returns nothing, and `showApiErrorToast` renders only `message` + `errorDetails.error`. Comment reworded to state the code-level classification and explicitly note it is a user-facing no-op. With the prune above, the 400 is largely unreachable anyway.

4. **Your open question — unselected real field, toggle on, ambiguous `ORDER BY`.** Good catch; it was genuinely undefined. Closed rather than documented: the synthetic is now suppressed whenever **any** known field owns the name, not only a selected one, so the web can never emit that ambiguous sort. Test added.

**Two I disagree with**

5. `hasUniqueCountMetric` is not `boolean | null | undefined`. The web `OutputConfig.uniqueCountConfig` is a non-nullable `boolean` (`output-config.ts:171`) — that is why eslint's `no-unnecessary-boolean-literal-compare` flagged my earlier `=== true` and I removed it. The expression types as `boolean`. (You may be reading the backend `UniqueCountConfig`, which *is* `boolean | null`.)

6. "Reordered … is untestable: pure disjunction, so evaluation order cannot change the result." The **predicate changed**, not just its order: the old code had an early `return` on the label branch that skipped the real-field check entirely, so a selected real field named `Unique Count` was reported disconnected. That is why the collision test fails on the old logic and passes on the new one — and your own note grants that two of the tests genuinely gate the diff. The imprecision was my PR wording ("reordered"); fixed in the updated description.

**Acknowledged, not acted on**

- `Row Count` sort asymmetry and the `validateOutputColumnNames` extraction — agreed that is the better registry, but it changes behaviour for a column that is not sortable today, so I am leaving it out of this PR. The blended-CTE half of that trap **is** now closed (`ROW_COUNT_LABEL` is in the builder fix).
- `type: 'INTEGER'` being dead and dialect-inaccurate, and the `Pick<…>` that would make `allColumns={sortColumns}` a compile error — fair, and the stronger version of my JSDoc-only invariant. Worth its own change.
- Metrics-only `columnConfig: null` sort mismatch (`:584` vs `:639-644`) — agreed, pre-existing, and out of scope here.
- Tests that pass on `main`: accurate. They are regression guards for the new contract rather than gates on this diff; I would rather keep them.
- `SortSection`'s inverted `selectedColumns`/`sortColumns` vocabulary and the "All selected columns are already sorted" empty state — both real readability issues introduced by the rename. Noted for follow-up.

**Verification:** `abstract-blended-query-builder.spec.ts` 79/79 · full web suite 943/943 after all edits · eslint + prettier + tsc clean. The only backend file touched after its last green run (`output-controls-validator.service.ts`, 177/177) is a comment-only change, so CI on this head is the authoritative signal for the backend suites.

zapolsky (approved):

PR #1474, opened 2026-07-28
feat(data-quality): add configurable Data Mart quality checks

kalnik-a-a:
## Summary

This PR introduces the first version of Data Quality checks for Data Marts. Users can configure checks, run them manually or on a schedule, and inspect the results through the same Run History used by connector and report runs.

The feature is available for BigQuery, Legacy BigQuery, Athena, Snowflake, Redshift, and Databricks Data Marts.

### Data Quality engine and storage support

- Added 11 configurable checks:
  - **Table:** empty table, primary key uniqueness, duplicate rows
  - **Field:** null rate, column uniqueness, constant column, type mismatch, data freshness, negative values
  - **Relationship:** relationship integrity and reverse relationship
- Added a storage-neutral compiler with provider-specific dialect support for quoting, null semantics, complex-value canonicalization, time expressions, type introspection, and query parameter handling.
- Each executed check now uses one readable, provider-native SQL query that returns the violation count, applicability, and up to three examples. The same query is stored and can be copied for reproduction without an artificial result limit.
- SQL-based Data Marts are checked through the existing managed service-view mechanism instead of embedding their raw source SQL into every generated query.
- Data freshness is exposed only for supported timestamp field types. The removed timezone configuration and `future_values` check are not part of this version.
- Checks within one Data Mart execute sequentially and persist after each result. A failed check does not prevent the remaining checks from running, and partial results are retained on execution failure or cancellation.

### Run lifecycle, persistence, and API

- Added `DATA_QUALITY` as a regular `DataMartRun` type.
- Stores the configuration snapshot, schema context, results, generated SQL, examples, and summary directly on `DataMartRun`; no separate run-result entity is introduced.
- Added APIs for reading and replacing the effective configuration, starting individual or atomic Save & Run executions, starting batch runs, and reading the latest or a specific run.
- Added lightweight quality summaries to Data Mart details and the Models canvas while keeping full results lazy-loaded.
- Reused the existing run-trigger and cancellation lifecycle, including best-effort cancellation between queries and active-run protection.
- Added Data Quality-specific consumption tracking with an idempotent run-based command and a dedicated Pub/Sub topic.
- Added scheduled Data Quality runs through the existing Data Mart scheduled-trigger framework.
- Preserves permission boundaries for related Data Marts: relationship summaries remain visible while target SQL and examples are redacted when the viewer cannot access the target.
- Added the SQLite/MySQL-compatible persistence migration and repository coverage.

### Web experience

- Added a **Data Quality** tab with configuration, latest-run summary, result filtering, examples, and copyable SQL.
- Field checks are added through a field-and-check picker, so the page only lists fields that actually have configured checks.
- Added manual Run and Save & Run flows, dirty-state protection, polling, cancellation states, and reusable active-run indicators.
- Added Data Quality runs to both Data Mart-level and project-level Run History without changing the existing collapsed-row interaction model.
- Added severity-aware status icons and finding counters to Models nodes, with details and the run action in the tooltip.
- Added shared bulk actions for selected Data Marts and for the currently filtered Models canvas.
- Added compact Data Quality status navigation to Output Schema while intentionally keeping configuration in the dedicated tab.

## Run semantics

- Findings complete the underlying run successfully while the Data Quality summary reports issues by configured severity.
- Warehouse or infrastructure failures mark the run as failed while preserving completed check results.
- Summary counters represent failed rules by severity; `violationCount` represents the affected rows, groups, or schema-level violations for that rule.
- Disabled checks are not stored in new run snapshots, avoiding unnecessary noise in Run History.
- Consumption is registered once after the run reaches durable `RUNNING`; runs that never start are not charged.

## Validation

- Backend unit and contract tests: **459 suites / 5,946 tests / 6 snapshots passed**
- Web unit and component tests: **160 files / 1,189 tests passed**
- Backend lint: passed
- Web lint and type-check: passed
- Backend production build: passed
- Web production build: passed
- Provider integration runner: completed successfully; real warehouse suites were skipped locally because no `.env.tests` credentials were configured

### Model Canvas
CleanShot 2026-07-29 at 11 38 18
CleanShot 2026-07-29 at 11 39 01

### Output Schema Badges
CleanShot 2026-07-29 at 11 35 18
CleanShot 2026-07-29 at 11 36 02
CleanShot 2026-07-29 at 11 36 28

### Data Quality Tab Layout
CleanShot 2026-07-29 at 11 47 37

### Checks Configuration Section
CleanShot 2026-07-29 at 11 44 16
CleanShot 2026-07-29 at 11 44 45
CleanShot 2026-07-29 at 11 45 06

### Latest Report Section
CleanShot 2026-07-29 at 11 50 47
CleanShot 2026-07-29 at 11 51 40

### Run History 
CleanShot 2026-07-29 at 11 55 19

### Triggers
CleanShot 2026-07-29 at 11 56 43
CleanShot 2026-07-29 at 11 57 11


apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
```diff
@@ -81,9 +99,78 @@ function ModelCanvasViewContent() {
     () => (filtered ? mergeBidirectionalEdges(filtered.edges) : []),
     [filtered]
   );
+  const selectedStorageType = dataStorages.find(storage => storage.id === filters.storageId)?.type;
+  const bulkActionDataMarts = useMemo(
```
laskevych, line 138: 🔴 **Critical (blocking):** On the canvas, `bulkActionDataMarts` is built from *all* filtered nodes (`filtered?.nodes ?? []`; the search query does not narrow it, and the service pages through every mart), while the canvas has no selection at all (`elementsSelectable={false}`). Yet the shared button tooltip says "Bulk actions for **selected** data marts" and the delete dialog says "You're about to delete N **selected** data marts. This action cannot be undone."

A user can delete every mart matching the current status filter believing they're deleting a selection. Please either add real selection on the canvas or change the copy/confirmation to make "applies to all N *filtered* data marts" unmistakable. Minor related point: deletion is soft on the backend, so "cannot be undone" is also inaccurate in the other direction.

apps/backend/src/data-marts/data-quality/data-quality-snapshot-table-reference.service.ts
```diff
@@ -0,0 +1,99 @@
...
+      throw new DataQualitySnapshotStorageMismatchError();
+    }
+  }
+}
+
+function createTechnicalViewName(
+  runId: string,
+  identity: DataQualitySnapshotReferenceIdentity
+): string {
+  const runToken = runId
```
laskevych: 🔴 **Critical (blocking):** Technical views are named per run (`dq_<runToken>_<identityHash>`, token derived from `runId`), created via plain `CREATE OR REPLACE VIEW` on all five storages, and persisted into `dataQualitySnapshot.technicalViews` — but nothing ever drops them (I found no DROP/cleanup path in the backend, and per-run names mean `CREATE OR REPLACE` never reuses an object).

A daily scheduled DQ run on a SQL-definition mart with two SQL relationship targets leaks ~1,100 permanent views per year into the customer's warehouse. Before this ships we need a cleanup strategy: drop views at run finish, use a deterministic per-mart/per-definition name, or a retention job that consumes the already-persisted `technicalViews` references.

apps/backend/src/data-marts/services/data-quality-run-trigger-handler.service.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  stuckTriggerTimeoutSeconds(): number {
+    return 60 * 60;
+  }
+
+  triggerTtlSeconds(): number {
+    return 23 * 60 * 60;
+  }
+
+  protected getRunTypes(): string[] {
```
laskevych, line 126: 🟡 **Important (blocking):** The generic paths this handler inherits from `BaseRunTriggerHandlerService` — `cleanupOrphanedRuns` (base:131-155) and `failDataMartRunSafely` (base:96-110) — set `run.status = FAILED` but never touch `dataQualitySummary`, which stays in its lifecycle state `QUEUED`.

The latest-run endpoints then report `state = QUEUED` forever: the workspace polls every 2 s indefinitely, list/canvas render a permanent "Queued" spinner, and Cancel always 409s because the run is already FAILED. These paths need the DQ-specific summary handling that `markRunAndSummaryAsExecutionFailed` already does (e.g. set `EXECUTION_FAILED`). Also, the persisted error text is the connector concurrency message, which is misleading on a `DATA_QUALITY` run.

apps/web/src/features/data-marts/data-quality/components/DataQualityWorkspace.tsx
```diff
@@ -0,0 +1,791 @@
...
+                        draft: {
+                          ...current.draft,
+                          rules: current.draft.rules.map(rule =>
+                            rule.key === ruleKey ? { ...rule, enabled: true } : rule
+                          ),
+                        },
+                      }
+                    : current
+                );
+                if (addedRule?.scope.type === 'FIELD') {
```
laskevych: 🟡 **Important:** `onAddCheck` updates the draft with `rules.map(rule => rule.key === ruleKey ? {...enabled} : rule)` — it can only flip existing rules, never append. When a dirty draft predates a rule key that appeared in a refetched `effectiveConfig` (the guard at ~137 intentionally preserves dirty drafts, while the picker lists fields from the fresh config), the update is a silent no-op — but the "added … — not saved yet" toast still fires because `addedRule` is looked up in the fresh config. Save then persists a config without the check the user believes they added. Consider appending the rule when it's missing from the draft.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
```diff
@@ -0,0 +1,669 @@
...
+      }
+      if (dataMartRun.status === DataMartRunStatus.PENDING) {
+        const startedAt = this.systemClock.now();
+        const claim = await runRepository.update(
+          { id: dataMartRunId, status: DataMartRunStatus.PENDING },
+          { status: DataMartRunStatus.RUNNING, startedAt }
+        );
+        if (!claim.affected)
+          throw new Error(`Data Quality run ${dataMartRunId} could not be claimed`);
+        dataMartRun.status = DataMartRunStatus.RUNNING;
```
laskevych, line 261: 🟡 **Important:** `claimRun` accepts an already-`RUNNING` run with no ownership/liveness check, and `recoverStuckTriggers` resets a PROCESSING trigger after 1 h without knowing whether the executor is still alive (nothing refreshes `trigger.modifiedAt` during execution). Any run whose checks exceed 1 h wall-time gets a second concurrent executor re-running the remaining checks — duplicate warehouse queries and spend.

The connector handler explicitly guards this case ("already RUNNING, skipping duplicate trigger"); the DQ handler should distinguish crash recovery from a live executor the same way, or heartbeat `modifiedAt` between checks. (Consumption is deduplicated by `processRunId` and appends are serialized/upserted, so the damage is cost, not data corruption.)

laskevych (commented):
Review of the Data Quality feature. Process: multi-angle pass over the full diff (engine/SQL dialects, run lifecycle, API/persistence, web) with per-finding verification against the actual code; several initially-suspected issues were dropped as intentional or already guarded (e.g. CANCELLED→SUCCESS on late completion is test-pinned; the latest-run anti-join is already covered by the existing `(dataMartId, createdAt)` index).

One blocking finding has no inline anchor because the file is not part of this PR's diff:

🔴 **Critical (blocking)** — `packages/api-client/src/runs.ts:16`: `PROJECT_DATA_MART_RUN_TYPE_VALUES` was not updated with `DATA_QUALITY`, while the backend's `GET /data-marts/runs` now returns DQ runs (`listVisibleByProject` has no type filter). `parseProjectRunHistory` throws `OWOXApiError` when *any* row fails `isProjectDataMartRun` — it never skips rows — so a single Data Quality run makes `client.runs.list()` fail for the whole page for every consumer of the published `@owox/api-client`. Please add `'DATA_QUALITY'` to the set, and consider skipping unknown run types instead of failing the entire response so the next run type doesn't repeat this. (The web app is unaffected — it uses its own client/enum — which is why tests didn't catch it.)

Inline: 3 more blocking comments, 12 important (non-blocking unless marked), 3 questions, 10 suggestions. Deliberately **not** flagged (checked, would be churn): the "already-quoted" unwrap heuristic (codebase-wide convention), reusing the type-alias parsers (they have their own quirks and don't fit DQ equality semantics), reusing `IdentifierEscaperFacade` for column paths (it splits on dots), and assorted micro-duplications.

apps/backend/src/data-marts/data-quality/data-quality-sql-dialect.ts
```diff
@@ -0,0 +1,715 @@
...
+    .map(line => `${indentation}${line}`)
+    .join('\n');
+}
+
+function quoteIdentifierSegment(value: string, quote: '`' | '"'): string {
+  const unquoted =
+    value.length >= 2 && value.startsWith(quote) && value.endsWith(quote)
+      ? value.slice(1, -1).replaceAll(`${quote}${quote}`, quote)
+      : value;
+  return `${quote}${unquoted.replaceAll(quote, `${quote}${quote}`)}${quote}`;
```
laskevych, line 591: 🟡 **Important:** `quoteIdentifierSegment` escapes the quote char by doubling it. That's correct for Databricks, but BigQuery quoted identifiers use backslash escapes: a field name containing a backtick compiles to `` `a``b` `` (syntax error → the check reports ERROR), and a name ending with `\` produces `` `amount\` `` where the backslash escapes the closing quote and swallows the following SQL into the identifier — and this SQL is both executed and shown to the user.

Output Schema field names are user-editable and validated only as non-empty strings, so this is reachable. The BigQuery/Legacy BigQuery dialects should backslash-escape `\` and `` ` `` (note `data-quality-sql-dialect.spec.ts:31-36` currently pins the doubling behavior).

apps/web/src/features/data-marts/data-quality/components/DataQualityRuleEditor.tsx
```diff
@@ -0,0 +1,187 @@
...
+                aria-label='Null rate threshold percent'
+                type='number'
+                min={0}
+                max={100}
+                step='any'
+                className='w-36'
+                value={thresholdPercent ?? 0}
+                disabled={controlsDisabled}
+                onChange={event => {
+                  const next = Number(event.target.value);
```
laskevych: 🟡 **Important:** The threshold `onChange` commits `Number(event.target.value)`, and `Number('') === 0` passes the `Number.isFinite` guard — clearing the field instantly writes `thresholdPercent: 0` (same for `thresholdHours` at ~173) into the draft, the controlled input snaps back to "0", the field can never be emptied, and select-all→retype mangles input. Saving in that state persists the strictest thresholds (0 % null rate; freshness cutoff = now). Keep the raw string in local state and commit only valid numbers (treat `''` as no-op).

apps/backend/src/data-marts/mappers/data-mart.mapper.ts
```diff
@@ -160,6 +167,7 @@ export class DataMartMapper {
       availableForReporting: dto.availableForReporting,
       availableForMaintenance: dto.availableForMaintenance,
       contexts: dto.contexts,
+      qualitySummary: dto.qualitySummary ?? createNoRunDataQualitySummary(0),
```
laskevych: 🟡 **Important:** `qualitySummary: dto.qualitySummary ?? createNoRunDataQualitySummary(0)` (here and at ~338) fabricates an `ALL_DISABLED` summary whenever a caller didn't attach one — and four existing use cases (publish, update-schema, actualize-schema, update-owners) call `toDomainDto` without a summary. The web reducer wholesale-replaces `dataMart` on those successes, so publishing a mart with enabled checks (no run yet) flips the Output Schema status card from "has not been checked yet" to "checks are disabled" until an unrelated refetch.

A lying default is worse than an absent field: make `qualitySummary` nullable in the response, or require callers to supply it so a forgotten endpoint fails loudly instead of asserting a false state.

apps/web/src/features/data-marts/data-quality/model/use-data-quality-workspace.ts
```diff
@@ -0,0 +1,166 @@
...
+  const detailRunId = terminalRunRef.current.runId;
+  const runQuery = useDataQualityRun(projectId, dataMartId, detailRunId);
+  if (runQuery.data && !isActiveState(runQuery.data.summary.state)) {
+    terminalRunRef.current.runId = runQuery.data.runId;
+    terminalRunRef.current.run = runQuery.data;
+  }
+  const activeRun = latestIsActive ? (latestQuery.data ?? null) : null;
+  const latestRun = latestIsActive
+    ? terminalRunRef.current.run
+    : (runQuery.data ??
```
laskevych: 🟡 **Important:** On the poll tick where a newly-started run reaches a terminal state, `latestRun` falls back to `latestQuery.data`, whose `results` are hardcoded to `[]` in `api/data-quality.service.ts:88`. The `sortedResults.length > 0` gate then unmounts the whole "Latest report" card (dropping per-card expansion state) and remounts it collapsed when the detail fetch lands; if that fetch fails, the UI shows only the error alert while the summary claims N findings. Consider keeping the previous run's results until the new details arrive, or gating on the detail query state instead of `results.length`.

apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx
```diff
@@ -83,6 +86,31 @@ function estimateEdgeLabelDimensions(
 const nodeTypes = { modelCanvasNode: ModelCanvasFlowNode };
 const edgeTypes = { modelCanvasEdge: ModelCanvasFlowEdge };
 
+function getNodeTopologySignature(nodes: readonly ModelCanvasNode[]): string {
```
laskevych, line 101: 🟡 **Important:** `getNodeTopologySignature` serializes `{id, title, status, description, fieldCount}` — `qualitySummary` is omitted — so `useStableValue` keeps returning the mount-time nodes array. The layout effect rebuilds all flow nodes from that stale array, so toggling layout direction or "Show join fields" reverts every quality badge to its mount-time state; since the quality poll stops once the run is terminal, nothing corrects it afterwards (the patch effect's `[nodes]` dependency did not change either). Add `qualitySummary` (or a digest) to the signature, or source summaries in the patch effect from the live prop.

apps/web/src/features/data-marts/list/model/hooks/useDataMartList.ts
```diff
@@ -1,26 +1,52 @@
...
-    } catch (error) {
-      dispatch({
-        type: 'SET_ERROR',
-        payload: error instanceof Error ? error.message : 'Failed to load data marts',
+      try {
+        const response = await dataMartService.getDataMarts(
+          options.silent ? { skipLoadingIndicator: true, skipErrorToast: true } : undefined
+        );
+        const listItems = mapDataMartListFromDto(response);
+        dispatch({ type: 'SET_ITEMS', payload: listItems });
```
laskevych, line 27: 🟡 **Important:** `loadDataMarts` dispatches `SET_ITEMS` unconditionally after `await`, with no request-generation counter or abort. A slow silent 2 s-poll response can land after a user-triggered `refreshList` (e.g. right after bulk delete) and overwrite the fresh list with the pre-delete snapshot — and if that snapshot has no active-quality items, the poll doesn't restart and the resurrected rows persist until manual refresh. A monotonic request id checked before dispatch (or an AbortController canceled on refresh) fixes it.

apps/backend/src/data-marts/services/data-quality-run.service.ts
```diff
@@ -0,0 +1,487 @@
...
+      }
+      if (run.dataQualitySummary) {
+        run.dataQualitySummary = {
+          ...run.dataQualitySummary,
+          state: DataQualitySummaryState.EXECUTION_FAILED,
+        };
+      }
+
+      run.status = DataMartRunStatus.FAILED;
+      run.errors = [DATA_QUALITY_RUN_EXECUTION_ERROR_MESSAGE];
```
laskevych, line 363: 🟡 **Important:** `markRunAndSummaryAsExecutionFailed` receives the failure as `_error` and discards it, persisting only the generic "Data Quality run failed during execution". The sibling generic path persists `error.message` (`base-run-trigger-handler.service.ts:102`) and per-check failures already store raw provider messages, so run-level infra failures are the only case where the user gets no diagnostic (it survives only in server logs). Please persist the message.

apps/backend/src/data-marts/data-storage-types/interfaces/create-view-executor.interface.ts
```diff
@@ -7,6 +7,10 @@ export interface CreateViewResult {
   fullyQualifiedName: string;
 }
 
+export interface CreateViewOptions {
```
laskevych: 🟡 **Important:** The new `CreateViewOptions.requireFullyQualifiedName` contract is honored inconsistently: Redshift/Databricks return the *raw* name with the flag but the *escaped* one without; Snowflake keeps the old 4-arg signature and always returns a pre-escaped identifier, which is persisted verbatim into `technicalViews` (it works today only because `escapeSnowflakeIdentifier` unwraps already-quoted parts); and `LegacyBigQueryCreateViewExecutor.createView` silently drops the `options` argument when delegating — TypeScript accepts the narrower signature, so nothing flags an executor that ignores the option.

Net effect: `technicalViewReference` stores raw dotted names for four storages and an escaped one for Snowflake — a trap for any future consumer (e.g. the cleanup this field anticipates). Suggest adding `options` to all five signatures and normalizing the return contract to raw names.

apps/backend/src/data-marts/data-storage-types/athena/services/athena-create-view.executor.ts
```diff
@@ -39,10 +41,14 @@ export class AthenaCreateViewExecutor implements CreateViewExecutor {
       throw new Error('Athena storage config expected');
     }
 
-    if (!fullyQualifiedViewName || !sql) {
+    if (!viewName || !sql) {
       throw new Error('View name and SQL must be provided');
     }
 
+    const fullyQualifiedViewName =
+      options?.requireFullyQualifiedName && !viewName.includes('.')
```
laskevych: 🟡 **Important:** The `requireFullyQualifiedName` path hardcodes the `default.` Glue database and never ensures it exists, while every sibling establishes its namespace (`CREATE SCHEMA IF NOT EXISTS owox_internal` on Redshift, `CREATE DATABASE/SCHEMA IF NOT EXISTS` on Snowflake, dataset auto-create on BigQuery). Two consequences: Athena DQ views land in the user-visible `default` database instead of an internal namespace, and on accounts where `default` was pruned (it's deletable) every DQ run on SQL marts fails with "Schema default not found" while the same config works on all other storages. Suggest an `owox_internal`-style database with ensure-exists, matching the siblings.

apps/backend/src/data-marts/entities/data-mart-run.entity.ts
```diff
@@ -82,4 +91,33 @@ export class DataMartRun implements CreatorAwareEntity {
 
   @Column({ type: 'json', nullable: true })
   additionalParams?: Record<string, unknown> | null;
+
+  @Column({
+    type: 'json',
+    nullable: true,
+    select: false,
```
laskevych, line 98: 🟡 **Important:** The DQ JSON columns are validated by `.strict()` Zod schemas whose transformer `parse()`s (throws) on load, and no format-version marker is persisted (`configRevision` is transient, checked only at enqueue). Any future field rename/removal — or a rollback right after an additive deploy — makes historical rows throw inside the TypeORM transformer; and since `dataQualitySummary` is *not* `select: false` and `listByDataMartId` is a plain `find`, one unparseable legacy row would break the entire Run History list for that mart. Consider `safeParse` with a degraded fallback on read, and/or a `version` field in the snapshot.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
```diff
@@ -0,0 +1,669 @@
...
+      }
+
+      const missingRules = pendingRules.filter(
+        rule => !parsedResults.some(result => result.ruleKey === rule.key)
+      );
+      for (const rule of missingRules) {
+        const parsed = executionError(rule, error);
+        const stored = await this.appendDataQualityResult(
+          dataMartRun.id,
+          this.createStoredResult(rule.scope, parsed)
```
laskevych, line 219: 🔵 **Question:** Cancellation semantics in the cancel-to-abort-signal window (the signal is delivered by a 5 s cron): when the in-flight work *completes*, the already-persisted CANCELLED run flips to SUCCESS — clearly intentional and test-pinned ("completes successfully when all checks finish before the cancellation signal arrives"). But when the in-flight query *errors*, this catch fabricates `EXECUTION_ERROR` results for every never-attempted check and the run becomes FAILED/EXECUTION_FAILED — contradicting "partial results are retained on cancellation", and this branch has no test.

Is that intended? If "true outcome wins" is the policy, it would still be better to persist only the actually-failing check's error rather than fabricating results for checks that never ran — and worth documenting that a 200 from the cancel endpoint can be superseded within ~5 s.

apps/backend/src/migrations/1784066400000-add-data-quality-foundation.ts
```diff
@@ -0,0 +1,118 @@
...
+      }
+    }
+    if (!(await queryRunner.hasTable(this.DATA_QUALITY_RUN_TRIGGER_TABLE))) {
+      await queryRunner.createTable(this.createDataQualityRunTriggerTable());
+    }
+
+    await queryRunner.query(
+      `UPDATE ${this.DATA_MART_TABLE}
+       SET ${this.CONFIG_COLUMN} = ?
+       WHERE status <> ? AND ${this.CONFIG_COLUMN} IS NULL`,
```
laskevych, line 50: 🔵 **Question:** The backfill writes `{rules: []}` only `WHERE status <> 'PUBLISHED' AND dataQualityConfig IS NULL`, so pre-existing *published* marts keep NULL and resolve to the system preset (empty-table, PK uniqueness, null-rate on PK/join fields, relationship integrity enabled), while pre-existing *drafts* are pinned to all-disabled. The migration spec asserts exactly this split, so it looks deliberate — but is "existing published marts get default-enabled checks" the intended grandfathering?

Nothing auto-runs, but the first bulk/scheduled run someone triggers will execute billable checks the owner never opted into, and two identical pre-migration marts end up with opposite effective configs based purely on publish status at migration time.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
```diff
@@ -0,0 +1,669 @@
...
+      }
+      return { dataMartRun, dataMart: dataMartRun.dataMart };
+    });
+  }
+
+  private async publishConsumption(dataMart: DataMart, dataMartRunId: string): Promise<void> {
+    try {
+      await this.consumptionTrackingService.registerDataQualityRunConsumption(
+        dataMart,
+        dataMartRunId
```
laskevych, line 290: 🔵 **Question:** Two consumption edges worth confirming: (1) `publishConsumption` fires at the top of every `executeExistingRun` invocation, so the intended crash-recovery re-execution republishes the same command — I assume the downstream consumer dedupes on `processRunId`; is that guaranteed? (2) A failed publish is only `logger.warn`'d and never retried, so a run that completes on its first attempt after a publish failure is never billed, with no persisted trace. If both are accepted trade-offs (consistent with other consumption types), fine — just flagging that the run row carries no "consumption registered" marker either way.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
```diff
@@ -0,0 +1,669 @@
...
+    }
+    dataMartRun.dataQualitySnapshot = snapshot;
+  }
+
+  private async appendDataQualityResult(
+    dataMartRunId: string,
+    result: DataQualityStoredCheckResult
+  ): Promise<boolean> {
+    return this.dataSource.transaction(async manager => {
+      const current = await this.findRunWithDataQualityDetails(manager, dataMartRunId, true);
```
laskevych: 💡 **Suggestion (non-blocking):** `appendDataQualityResult` re-reads the full run row (snapshot + accumulated results + joined mart/storage) in a locked transaction after every check, then rewrites the whole results array — O(N²) I/O in results volume per run, while the in-memory `parsedResults` accumulator already exists. The same file already demonstrates the cheaper pattern: `persistTechnicalViewReference` uses a guarded `update(… status IN (RUNNING, CANCELLED) …)`. A conditional UPDATE (checking `affected` for the abort signal) would keep the concurrency guard without per-check re-reads; the only care point is the merge with `cancelActiveRun`'s summary write.

apps/web/src/features/data-marts/list/model/hooks/useDataMartList.ts
```diff
@@ -1,26 +1,52 @@
...
+      }
+    },
+    [dispatch]
+  );
+
+  const hasActiveQualityRun = state.items.some(
+    item => dataQualityPollingInterval(item.qualitySummary.state) !== false
+  );
+
+  useEffect(() => {
```
laskevych: 💡 **Suggestion (non-blocking):** Quality polling is implemented three different ways (workspace `refetchInterval`, this hand-rolled 2 s `setInterval`, and a shadow disabled query + 5 s interval + `dataUpdatedAt` merge on the canvas), while the codebase already has `useAutoRefresh` (visibility gating + abort) used by the other polling features. The list/canvas variants refetch the full endpoints — which now hydrate `schema` + `dataQualityConfig` per row and re-resolve effective configs for every no-run mart — every tick, per open tab, including hidden tabs.

A shared "poll quality summaries while active" hook plus a lightweight `{dataMartId, qualitySummary}` batch endpoint would serve both surfaces. Related: `calculateHasActiveRuns` (edit reducer:237) now excludes DQ runs, so Run History stays on the 30 s cadence during an active DQ run and lags the Quality tab (2 s) by up to 30 s.

apps/backend/src/data-marts/mappers/data-quality-api.mapper.ts
```diff
@@ -0,0 +1,145 @@
...
+import {
+  LatestDataQualityRunResponseApiDto,
+  RunDataQualityRequestApiDto,
+} from '../dto/presentation/data-quality-api.dto';
+import { DataMartRun } from '../entities/data-mart-run.entity';
+import { DataQualityScope } from '../enums/data-quality-scope.enum';
+
+const BatchRunDataQualityRequestSchema = z
+  .object({
+    dataMartIds: z.array(z.string().min(1).max(255)).min(1),
```
laskevych, line 17: 💡 **Suggestion (non-blocking):** `dataMartIds` has no max size, unlike the analogous batch endpoints (`@ArrayMaxSize(200)` on batch health status). Express's 100 kb JSON default caps this at ~2.5 k ids in practice, but that still allows a 2.5 k-id `IN ()`, a matching `canAccessMany` fan-out and sequential enqueues from one request. Add a `.max(…)` consistent with the precedent.

apps/backend/src/data-marts/services/data-quality-run-trigger.service.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  async requestCancellation(
+    trigger: DataQualityRunTrigger | null,
+    manager: EntityManager
+  ): Promise<void> {
+    if (!trigger) return;
+    if (trigger.status === TriggerStatus.CANCELLED) return;
+
+    const nextStatus =
+      trigger.status === TriggerStatus.PROCESSING
```
laskevych: 💡 **Suggestion (non-blocking):** `requestCancellation` re-implements the transition from the shared `stopRunTriggersForRun` (used by connector and report trigger services) but omits its `version: () => 'version + 1'` optimistic-lock bump. The practical exposure is narrow (the PubSub stale-message check; currently mitigated by `cancelTriggerIfRunAlreadyCancelled`), but extending the shared util's union with `DataQualityRunTrigger` removes the divergence for free.

apps/backend/src/data-marts/use-cases/get-data-mart-run.service.ts
```diff
@@ -34,16 +38,42 @@ export class GetDataMartRunService {
...
+        throw new NotFoundException('Run not found');
+      }
+      const targetIds = Array.from(
+        new Set(
+          (run.dataQualitySnapshot?.relationships ?? []).map(
+            relationship => relationship.targetDataMartId
+          )
+        )
+      );
+      const accessByTargetId =
```
laskevych, line 59: 💡 **Suggestion (non-blocking):** When `command.userId` is falsy, the relationship-target access map defaults to all-`true`, i.e. no redaction of related-mart SQL/examples. Today the only construction site always passes a verified `userId`, so it isn't exploitable — but the safe default for cross-mart data should be fail-closed, and extracting the access-map + `toRunDetails` orchestration into a single `getRedactedRunDetails(run, principal)` would keep the next caller (project view, MCP, export) from copying the fallback.

apps/backend/src/data-marts/services/data-quality-config-resolver.ts
```diff
@@ -0,0 +1,421 @@
...
+    ...(overrides.notApplicableReason
+      ? { notApplicableReason: overrides.notApplicableReason }
+      : {}),
+    parameters:
+      overrides.parameters ??
+      (category === DataQualityCategory.DATA_FRESHNESS ? { thresholdHours: 24 } : {}),
+  };
+}
+
+function collectCurrentFields(
```
laskevych, line 290: 💡 **Suggestion (non-blocking):** This re-implements the compiler's schema-walking layer as near-verbatim private copies (`collectCurrentFields` vs `collectFieldList`, `descendantsRequireFlattening`, `isRepeatedScalar`, path helpers, the REPEATED→COMPLEX normalization, reason strings) — and they've already drifted ("…in the *current* Output Schema" here at :154 vs "…in the Output Schema" compiler:182). These layers must agree by design: the resolver predicts applicability for the UI, the compiler enforces it at run time. A shared module for the field walker + applicability predicates would prevent divergence. (The existing `data-mart-schema.utils.ts` is not a drop-in — it prunes `isHiddenForReporting` — so this needs a new shared DQ module.)

Same spirit: `compilePkUniqueness`/`compileDuplicateRows` are copy-paste twins that could share one duplicate-group builder, and the zero-filled summary template exists in three places (run service, summary service, `DataQualityWorkspace`).

apps/backend/src/data-marts/services/data-quality-config-resolver.ts
```diff
@@ -0,0 +1,421 @@
...
+      notApplicableReason: discoveredRule.notApplicableReason,
+    };
+  });
+
+  for (const savedRule of parsedSavedConfig.rules) {
+    if (discoveredByKey.has(savedRule.key)) continue;
+    effectiveRules.push({
+      ...savedRule,
+      isApplicable: false,
+      notApplicableReason: staleScopeReason(savedRule.scope),
```
laskevych: 💡 **Suggestion (non-blocking):** Two small resolver issues: (1) a saved `DATA_FRESHNESS` rule whose field still exists but lost eligibility (type changed, or became repeated) falls into this stale-key branch and gets "Field scope no longer exists in the current Output Schema" — a false message, since the same field still produces the other field-category rules; a dedicated "field is no longer eligible for freshness" reason would be accurate. (2) `storageTypeForSchema`'s `default: return null` (~389-404) is the one silent branch in an otherwise loud dispatch chain — for a future storage type it marks every field check not-applicable in the UI while the compiler accepts them; consider throwing there.

apps/backend/src/data-marts/data-quality/data-quality-sql-dialect.ts
```diff
@@ -0,0 +1,715 @@
...
+  expectedNativeType: string,
+  expectedMode?: string
+): boolean {
+  switch (storageType) {
+    case DataStorageType.GOOGLE_BIGQUERY:
+    case DataStorageType.LEGACY_GOOGLE_BIGQUERY:
+      return matchesBigQueryStorageType(actualNativeType, expectedNativeType, expectedMode);
+    case DataStorageType.AWS_ATHENA:
+      return (
+        normalizeAthenaStorageType(actualNativeType) ===
```
laskevych: 💡 **Suggestion (non-blocking):** `matchesProviderStorageType` for Athena/Redshift/Databricks returns `true` when both sides normalize to `null` (two unrecognized types "match"), while the BigQuery variant guards `actual !== null` and Snowflake guards `!expected`. Unreachable today (schema types are enum-validated and fully mapped), but it's a landmine for the next enum member — align the three on the fail-closed guard.

apps/backend/src/data-marts/data-quality/data-quality-result-parser.ts
```diff
@@ -0,0 +1,448 @@
...
+    if (!readBoolean(readValue(row, resultShape.exampleMarkerColumn))) continue;
+    const values = safeSerializeRow(
+      Object.fromEntries(
+        resultShape.exampleColumns.map(column => [
+          column,
+          unwrapProviderScalar(readValue(row, column)),
+        ])
+      )
+    );
+    const key = stableExampleKey(values);
```
laskevych, line 251: 💡 **Suggestion (non-blocking):** Two minor example-fidelity issues: `parseExamples` dedups by serialized value, so NULL_RATE examples on a mart without PK fields all serialize to `{null_value: null}` and collapse to one row though the header promises up to 3 (consider a row locator when no PKs exist); and `unwrapProviderScalar` (~274-280) treats any `{value: <scalar>}` object as a provider wrapper, so a genuine `STRUCT<value INT64>` example cell is silently flattened to its scalar.

apps/backend/src/data-marts/services/data-quality-summary.service.ts
```diff
@@ -0,0 +1,130 @@
...
+    dataMartIds: readonly string[],
+    projectId: string
+  ): Promise<Map<string, DataQualitySummaryDto>> {
+    const ids = Array.from(new Set(dataMartIds));
+    if (ids.length === 0) return new Map();
+
+    const runs = await this.repository
+      .createQueryBuilder('run')
+      .innerJoin('run.dataMart', 'dataMart')
+      .leftJoin(
```
laskevych, line 40: 💡 **Suggestion (non-blocking):** Several hot-path queries hydrate more than they use: `getLatest`/`getActiveRunId` (2 s-polled) load `definitionRun`/`logs`/`errors` when the response needs id + summary + timestamps (`getActiveRunId` needs only `id`); `getLatestByDataMartIds` here (every list/canvas request) does `.getMany()` with default selection; and the DQ branch of `GetDataMartRunService` fetches the run twice (type probe, then heavy re-query — `addSelect` of the DQ columns in the first query would do). `.select()` projections would trim the endpoints most exposed to polling. Optional: the ideal composite index for the latest-per-mart anti-join is `(dataMartId, type, createdAt)` — `(dataMartId, createdAt)` already exists, so this is a nice-to-have.

apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
kalnik-a-a: The canvas behavior is intentional: the status and relationship controls define the Data Marts shown on the canvas, and Actions apply to all of those Data Marts. There is no separate selection model. I fixed the misleading shared copy so canvas confirmations explicitly say that the action applies to all N Data Marts shown by the current canvas filters, while the table keeps selected-item wording. Search remains a highlight-only control. I also removed the inaccurate irreversible-delete wording because Data Mart deletion is soft delete.

kalnik-a-a:
Addressing the blocking API-client finding from the review summary: fixed in `43ecd85fd`. `DATA_QUALITY` is now part of the public project Run History run-type union, with a regression test covering a project history page that contains a DQ run. Parsing remains strict for genuinely unknown future run types. The Real DB workflow also passed on this head: https://github.com/OWOX/owox-data-marts/actions/runs/30524932378

apps/backend/src/data-marts/data-quality/data-quality-snapshot-table-reference.service.ts
kalnik-a-a: Agreed. I removed DQ-owned per-run view creation, generated `dq_*` names, persisted `technicalViews`, and the snapshot-specific Create View path. SQL-based DQ sources and relationship targets now use the existing stable canonical view through `DataMartTableReferenceService`, so repeated runs do not create additional warehouse objects. The accepted trade-off is that SQL runs use the current canonical definition, matching the existing report/sample-data behavior.

apps/backend/src/data-marts/services/data-quality-run-trigger-handler.service.ts
kalnik-a-a: Fixed through a small base-handler terminalization hook with a DQ-specific override. An orphaned DQ run now persists `FAILED`, `EXECUTION_FAILED`, `finishedAt`, and a safe DQ execution error in the same save, so summary polling reaches a terminal state.

apps/web/src/features/data-marts/data-quality/components/DataQualityWorkspace.tsx
kalnik-a-a: Fixed. Add Check now enables an existing dirty-draft rule or appends a stored copy when the rule appeared in a newer effective config. The success toast is emitted only after an actual draft change. A dirty-draft plus config-refetch regression test covers the case.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
kalnik-a-a: Fixed without changing the shared scheduler recovery policy. The DQ handler now refreshes the processing trigger heartbeat while execution is alive and always clears the heartbeat in `finally`. Heartbeat updates neither trigger status nor version, so normal runner completion and cancellation keep their existing optimistic transitions.

apps/backend/src/data-marts/data-quality/data-quality-sql-dialect.ts
kalnik-a-a: Fixed. BigQuery and Legacy BigQuery now escape backslashes and backticks using GoogleSQL quoted-identifier rules, while Databricks retains doubled-backtick escaping. Contract tests cover embedded backticks, backslashes, mixed input, and a trailing backslash.

apps/web/src/features/data-marts/data-quality/components/DataQualityRuleEditor.tsx
kalnik-a-a: Fixed for both percentage and freshness thresholds. The editor keeps a local string while typing, allows an intermediate empty value, commits only finite in-range numbers, and restores the last valid value on blur. Component tests cover clear/replace, decimals, and invalid blur.

apps/backend/src/data-marts/mappers/data-mart.mapper.ts
kalnik-a-a: Fixed by removing Data Quality from generic Data Mart mapping entirely. Generic detail/list/mutation/connector responses, embedded Report Data Marts, and Models canvas topology no longer hydrate or expose `qualitySummary`. Compact status is now served by a dedicated project-scoped, access-aware batch endpoint. The list requests it only for the current page, while the canvas requests it only for nodes left by its status/relationship filters; active-run polling is scoped to the same visible IDs. A missing summary for an otherwise accessible Data Mart is treated as a contract error and is never converted into `ALL_DISABLED`. Run History still carries `qualitySummary` because it describes that specific historical run.

apps/web/src/features/data-marts/data-quality/model/use-data-quality-workspace.ts
kalnik-a-a: Fixed. The lightweight latest-run overview is no longer treated as a full report with empty results. The previous terminal report remains visible while details for the new run load and also remains available if that detail request fails, alongside the loading/error feedback. Both transitions have regression coverage.

apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx
kalnik-a-a: Fixed. Topology still stays stable to avoid rerunning Dagre on every poll, but every layout rebuild overlays the current `qualitySummary` from the live node ref. A terminal badge now survives direction and layout-setting changes.

apps/web/src/features/data-marts/list/model/hooks/useDataMartList.ts
kalnik-a-a: Fixed by removing DQ polling from the generic list state entirely. The generic list has its own monotonic load generation, while compact summaries are owned by React Query under a key derived from the stable unique IDs of the current page. Changing page, search, filters, or rows detaches and aborts the obsolete summary request, so it cannot overwrite or resurrect generic list items. Regression tests cover page-scoped IDs and request cancellation.

apps/backend/src/data-marts/services/data-quality-run.service.ts
kalnik-a-a: I kept the current security boundary intentionally. Generic `DataMartRun.errors` is exposed through common Run History paths that do not apply DQ relationship redaction, so copying a raw provider error there could expose SQL or identifiers. The raw error remains in server logs, while access-aware per-check details retain the safe structured error. A correlation ID can be added separately if operational diagnosis needs it.

apps/backend/src/data-marts/data-storage-types/interfaces/create-view-executor.interface.ts
kalnik-a-a: Resolved by removing the DQ-only snapshot-view path rather than extending the shared contract. `runFromSnapshot` and `requireFullyQualifiedName` no longer exist, and the common Create View service, facade, interface, and provider executors are restored to their pre-DQ behavior.

apps/backend/src/data-marts/data-storage-types/athena/services/athena-create-view.executor.ts
kalnik-a-a: Resolved together with the per-run view removal. DQ no longer calls a special Athena Create View path; it uses the same stable `DataMartTableReferenceService` flow as the rest of the application, so the DQ-specific `default` database branch is gone.

apps/backend/src/data-marts/entities/data-mart-run.entity.ts
kalnik-a-a: No code change here by design. DQ is still under development and has not shipped, so we do not need backward compatibility or a version wrapper yet. Keeping the schema strict helps detect corrupt state. The first post-release incompatible snapshot change should introduce an explicit version/migration strategy rather than speculative versioning now.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
kalnik-a-a: The current cooperative-cancellation semantics are intentional and covered. If cancellation is persisted but all work completes before the abort signal is observed, the completed result wins and the run becomes successful. If abort is observed between checks, the run remains cancelled with partial results. Per-check provider errors are stored as check errors; only setup/global failures use the outer failure path.

apps/backend/src/migrations/1784066400000-add-data-quality-foundation.ts
kalnik-a-a: This is the intended rollout behavior: existing published and newly created Data Marts use the system preset, while existing unpublished Data Marts are migrated to an explicit all-disabled config. The migration does not start any runs. No change was made.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
kalnik-a-a: The current handler publishes DQ consumption only after terminal `SUCCESS` has been durably saved, and terminal runs are not claimed again. This is covered by tests. I did not add a DQ-only outbox or billing marker; publish remains best effort, consistent with the existing consumption paths, and downstream deduplication uses the run ID as `processRunId`. Stronger delivery guarantees should be implemented as a shared consumption change.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
kalnik-a-a: I reduced the hot-path hydration in this PR: each per-check mutation now locks/selects only the run ID, status, summary, `finishedAt`, and embedded results instead of reloading the snapshot, Data Mart, and storage. I kept the required save-after-each-check behavior and the cross-database embedded-array contract. Replacing it with provider-specific JSON updates or a separate persistence model needs measurement and a dedicated design.

apps/web/src/features/data-marts/list/model/hooks/useDataMartList.ts
kalnik-a-a: The concrete duplication is now reduced rather than deferred: list and canvas share one React Query summary hook, including stable ID normalization, cancellation, and active-state polling. Generic list/topology data is never reloaded by that polling. Single-Data-Mart latest/details queries remain separate because they return different resources and drive the report workspace lifecycle.

apps/backend/src/data-marts/mappers/data-quality-api.mapper.ts
kalnik-a-a: No arbitrary API cap was added. The read path no longer requests every Data Mart loaded by a page: list summaries are scoped to the current paginated rows and canvas summaries to nodes left by its filters. The run batch intentionally still applies to every Data Mart explicitly selected in the list or shown by the canvas filters and returns an independent typed result for each. We can add measured chunking or limits later if production load demonstrates a need.

apps/backend/src/data-marts/services/data-quality-run-trigger.service.ts
kalnik-a-a: Fixed. `DataQualityRunTrigger` is now supported by the existing shared cancellation helper, including the version bump and the established READY/IDLE/PROCESSING transitions. The DQ service delegates to that helper and keeps only its in-memory trigger state synchronized.

apps/backend/src/data-marts/use-cases/get-data-mart-run.service.ts
kalnik-a-a: Fixed. Missing principal context now produces deny entries for every relationship target instead of granting access. The existing controller still supplies the principal, and a unit test locks in the fail-closed default.

apps/backend/src/data-marts/services/data-quality-config-resolver.ts
kalnik-a-a: I kept the resolver/compiler boundary unchanged in this PR because a shared walker refactor touches applicability and provider SQL compilation simultaneously. I fixed the concrete inconsistency reported for freshness and retained cross-layer contract coverage. A shared walker should be extracted in a focused follow-up with explicit nested-field and provider invariants.

apps/backend/src/data-marts/services/data-quality-config-resolver.ts
kalnik-a-a: Fixed. If the field still exists but no longer has a freshness-eligible type, the rule now says that the field is no longer eligible for Data Freshness. A genuinely missing field keeps the removed-field reason. Unknown future schema discriminators now fail exhaustively instead of silently becoming not applicable.

apps/backend/src/data-marts/data-quality/data-quality-sql-dialect.ts
kalnik-a-a: Fixed for Athena, Redshift, and Databricks. A normalized type must now be known on both sides before equality is considered; `null === null` can no longer produce a false match. Provider contract tests cover the unknown-type cases.

apps/backend/src/data-marts/data-quality/data-quality-result-parser.ts
kalnik-a-a: I kept NULL deduplication intentionally: without a stable PK, three identical null objects do not provide additional diagnostic value, and the UI promises up to three distinct examples. I also did not guess whether a single-key `{ value: ... }` object is a provider wrapper or a real STRUCT. Correct unwrapping requires provider/native column metadata and should be handled in a dedicated follow-up rather than with a shape-only heuristic.

apps/backend/src/data-marts/services/data-quality-summary.service.ts
kalnik-a-a: Fixed the low-risk parts. Active-run lookup selects only the ID; latest and bulk summary lookups select only compact summary/timestamp fields; detail retrieval no longer performs a probe followed by a second DQ query; and per-check persistence no longer reloads snapshot/Data Mart/storage. Generic Data Mart list/canvas polling was also removed in favor of the dedicated compact summaries endpoint. I did not add an index without measurements because the existing `(dataMartId, createdAt)` index already serves the latest-run lookup pattern.

apps/backend/src/data-marts/services/data-quality-api.service.ts
```diff
@@ -0,0 +1,368 @@
...
+
+    return { items };
+  }
+
+  async getSummaries(
+    context: AuthorizationContext,
+    requestedIds: readonly string[]
+  ): Promise<GetDataQualitySummariesResponseApiDto> {
+    const ids = Array.from(new Set(requestedIds));
+    const existing = await this.dataMartService.findByIdsAndProjectId(ids, context.projectId);
```
romandubovyi, line 160: 🔴 **Blocking:** `getSummaries` passes entities from `dataMartService.findByIdsAndProjectId(...)` into `summaryService.getCurrentByDataMarts(...)`, but that finder selects `['id']` only (`data-mart.service.ts:63`). For every Data Mart with no Data Quality run yet, `getCurrentByDataMarts` calls `resolveEffectiveDataQualityConfig(dataMart.dataQualityConfig, dataMart.schema, ...)` with both arguments `undefined`: the resolver returns the system preset for an `undefined` saved config, and `deriveSystemPreset` over an empty schema leaves only `EMPTY_TABLE` enabled — so `enabledChecks: 1`, state `NEVER_RUN`.

Concretely: a Data Mart migrated to `{rules: []}` reports `NEVER_RUN` instead of `ALL_DISABLED`, and a Data Mart with six enabled applicable checks reports one. This feeds the Output Schema status card, the Data Mart details summary and the Models canvas node badges — i.e. exactly the fabricated-summary class of problem this batch endpoint was introduced to avoid.

The sibling `getCurrentByDataMartIds` (line 115) loads the marts with no `select` at all, so the "pass me an entity" contract does assume a hydrated one. Either give `getSummaries` a finder that hydrates `schema` + `dataQualityConfig` (separate from the id-only one `runBatch` needs), or have `getCurrentByDataMarts` re-read the marts it needs instead of trusting the caller's projection. `data-quality-api.service.spec.ts` mocks `getCurrentByDataMarts`, so this interaction is currently uncovered.

🤖 Generated with Claude Code (Opus 5, 1M context)

apps/backend/src/data-marts/services/data-mart.service.ts
```diff
@@ -84,6 +93,8 @@ export class DataMartService {
         'dm.status',
         'dm.description',
         'dm.definitionType',
+        'dm.schema',
```
romandubovyi: ⚪ **Non-blocking:** `dm.schema` + `dm.dataQualityConfig` here, and `dm.definitionType` + `dm.dataQualityConfig` in `findByProjectIdAndStorageIdForCanvas` (~227), look like leftovers from the reverted "hydrate quality in the generic list" approach.

None of them has a consumer: `toListItemDto`/`toListItemResponse` (mapper lines 281–332) read neither `schema` nor `dataQualityConfig`, and `model-canvas.mapper.ts` uses only `dataMart.schema?.fields?.length` — it never references `dataQualityConfig` or `definitionType`.

Both queries page at 1000 rows, so this pulls the full Output Schema JSON and the Data Quality config JSON per row on the two hottest list endpoints for nothing. Suggest dropping the unused columns (`dm.schema` stays on the canvas query for `fieldCount`).

🤖 Generated with Claude Code (Opus 5, 1M context)

apps/web/src/features/data-marts/data-quality/model/use-data-quality-workspace.ts
```diff
@@ -0,0 +1,227 @@
...
+  const normalizedDataMartIds = Array.from(new Set(dataMartIds)).sort();
+
+  return useQuery({
+    queryKey: dataQualityQueryKeys.summaries(projectId, normalizedDataMartIds),
+    queryFn: async ({ signal }) => {
+      const summaries = await dataQualityService.getSummaries(normalizedDataMartIds, {
+        signal,
+        ...SILENT_QUERY_OPTIONS,
+      });
+      assertAllSummariesReturned(normalizedDataMartIds, summaries);
```
romandubovyi: ⚪ **Non-blocking:** `assertAllSummariesReturned` throws whenever the backend omits an id, but the backend legitimately omits Data Marts that were deleted or that the caller cannot `SEE` — `DataQualityApiService.getSummaries` only emits items for `visibleDataMarts`.

Concrete case: on the canvas, `refreshCanvas` (`ModelCanvasView.tsx:204`) fires `refetch()` and `refetchQuality()` concurrently after a bulk delete, and the summaries refetch still runs under the pre-delete id set. It throws, and `const error = topologyError ?? qualitySummariesError` (line 209) surfaces it as a page-level "Failed to load the data model" right after a successful delete, until the topology query settles.

Treating a missing entry as "no summary for this node yet" would keep the contract check without turning an expected condition into a page error. Note that `ModelCanvasView.tsx:119` repeats the same assumption inside a `useMemo` — it is currently unreachable only because this assert fails first, so it should be relaxed together with this one rather than left behind.

🤖 Generated with Claude Code (Opus 5, 1M context)

romandubovyi (commented):
Review of the Data Quality feature, focused on the batch summaries path added in the latest revision.

Three points: one blocking (the compact-summaries endpoint derives quality state from id-only Data Mart entities, producing fabricated `NEVER_RUN` / `enabledChecks` values on the list, details and canvas), and two non-blocking (dead columns on the list and canvas selects, and a missing summary being treated as a hard query failure).

Each finding was verified against the code at df02bd3 before being included; several other candidates were dropped as false or already addressed.

🤖 Generated with Claude Code (Opus 5, 1M context)

apps/backend/src/data-marts/services/data-quality-api.service.ts
kalnik-a-a: 🔴 **Blocking — fixed in `098353bae`.** The summaries endpoint now keeps the ID-only lookup only for project scoping and SEE checks, then passes the visible IDs to `DataQualitySummaryService`. That service explicitly loads only `id`, `schema`, and `dataQualityConfig` before deriving no-run summaries, so saved `{ rules: [] }` remains `ALL_DISABLED` and configured or preset counts are computed from the actual schema. Regression coverage exercises the ID-only access lookup and the hydrated summary path.

apps/backend/src/data-marts/services/data-mart.service.ts
kalnik-a-a: ⚪ **Non-blocking — fixed in `098353bae`.** I removed `schema` and `dataQualityConfig` from the list projection, and `definitionType` and `dataQualityConfig` from the canvas projection. The canvas still selects `schema` for `fieldCount`, while ERD `definitionType` and fields continue to come from the existing detail enrichment; the list keeps `definitionType` for its DTO and connector mapping.

apps/web/src/features/data-marts/data-quality/model/use-data-quality-workspace.ts
kalnik-a-a: ⚪ **Non-blocking — fixed in `098353bae`.** The batch endpoint is intentionally partial for deleted or no-longer-visible Data Marts, so the shared batch hook no longer treats an omitted ID as a query failure. On the canvas, nodes without a returned summary and their incident edges are temporarily excluded until topology catches up; no fallback summary is fabricated. The single-Data-Mart details hook remains strict. Regression tests cover the partial-response and stale-edge cases.

laskevych (approved):
✅ **Approving.** Re-reviewed the follow-up commits (`412dcb168`, `48203caa0`, `098353bae`) against the previous review: all 4 blocking findings are resolved, and @romandubovyi's blocking finding is fixed (`getCurrentByDataMartIds` now re-loads marts with `id`/`schema`/`dataQualityConfig` instead of receiving id-only entities).

Verified in code and tests: per-run technical views removed in favor of the stable `view_<dataMartId>` flow; orphaned DQ runs terminalize with an `EXECUTION_FAILED` summary; processing-trigger heartbeat (5 min ≪ 1 h timeout, cleared in `finally`); GoogleSQL backslash escaping for BigQuery identifiers; threshold editor local-string state; add-check now appends missing rules and toasts only on real changes; canvas overlays live summaries on layout rebuilds; monotonic list load generation plus one shared React Query summaries hook scoped to visible rows/nodes; fail-closed redaction default; shared cancellation helper with the version bump; `DATA_QUALITY` added to `@owox/api-client`; select projections on the polled paths; type-match null guards (double-checked: every schema enum member still normalizes non-null on all three dialects, so the new guard introduces no false positives).

Non-blocking follow-ups worth tracking, none of which should hold up the merge:

1. **Summaries endpoint fan-out** — `POST …/data-quality/summaries` has no `dataMartIds` size cap, and `canAccessMany` fans out per-id access queries via `Promise.all`. A ~1000-node canvas polling at 2 s during an active run can burst the DB pool. Consider an `@ArrayMaxSize`-style cap plus bounded concurrency (or the SQL visibility filter used elsewhere).
2. **Consumption publish is outside the terminal transaction** — a crash between the SUCCESS commit and the publish leaves a successful run unbilled (terminal runs are not re-claimed), and the residual dual-executor edge can publish twice (dedupe relies on the downstream consumer keying on `processRunId`). Fine as best-effort; flagging for the billing backlog.
3. **Canvas skeleton on filter changes** — the summaries query re-keys on every visible-id-set change without `placeholderData: keepPreviousData`, so toggling a filter unmounts the canvas to a skeleton (viewport/layout reset) until the POST resolves; and `error = topologyError ?? qualitySummariesError` blanks the whole canvas when only the summaries call fails.
4. **Snapshot semantics for SQL marts** — checks now run against the live `view_<dataMartId>` refreshed from the *current* SQL, while the run snapshot records queued-time SQL. The spec asserts this, so treating it as intended — worth a line in the docs.
5. **Pre-release data note** — snapshot rows written by earlier builds of this branch still contain `technicalViews` and will fail the `.strict()` parse on read (staging/preview environments only; consistent with the stated pre-release strictness stance).

romandubovyi (approved):

PR #1479, opened 2026-07-30
feat(web): Joinable Data Marts in the ERD design, canvas layout persistence, object labels and diagram filters

laskevych:
### Joinable Data Marts
CleanShot 2026-07-30 at 17 06 35

### Models canvas

https://github.com/user-attachments/assets/cd783b8a-ea95-4584-a526-2c2806e9d5c8

## Summary

Follow-up to #1476. Brings the **Joinable Data Marts** diagram on the Data Mart page in line with the Models canvas ERD design, and adds the next round of canvas UX: persisted layout, object-label toggles, diagram filters, and reference-matching edge colors. Also removes the edge cardinality badge per product decision.

### Joinable Data Marts diagram (Data Mart page)

- **ERD cards, same as the Models canvas** — accent stripe + definition-type badge (`SQL` / `VIEW` / `TABLE` / `PATTERN` / `CONNECTOR`), published/draft status dot, rounded card chrome. The alias badge, connected-field count, warning/attention indicators (Loop / Draft / Blocked / No primary key), no-access marker and all viewport behaviors are preserved. The root mart is subtly tinted to stand out.
- **Definition types are enriched non-blocking** (`useRelationshipDefinitionTypes`): the relationship payload doesn't carry `definitionType`, so it's fetched from the detail endpoint after first paint; cards render with a neutral accent until it arrives, and an individual fetch failure leaves that card neutral instead of breaking the diagram.
- **New filters** in a gear popover on the canvas (persisted in localStorage):
  - **Show looped Data Marts** — off by default. Well-connected marts produce N self-referencing loop stubs that blow the graph up; they're now hidden unless explicitly enabled.
  - **Status** — All (default) / Published / Draft. Filtering a node hides its subtree; the root mart is always shown.
- Background grid removed, React Flow attribution hidden, minimap colored by definition type — matching the Models canvas.

### Models canvas

- **Layout persistence** — dragged card positions are stored in localStorage per storage and survive a reload. A saved position wins over the dagre layout; picking a layout algorithm is treated as an explicit re-layout and clears them. Compact/Detailed switching and search keep positions.
- **Object labels** (ported from `owox/models`) — a new settings section with three independent toggles: **Input source** (badge + accent stripe — they encode the same thing and hide together), **# Field count**, and **• Status dot**, plus **Check all — show everything** / **Uncheck all — title only** shortcuts. Stored as the *hidden* set, so parts added later default to visible for users with an existing preference. When both the badge and the count are hidden, cards drop the meta row entirely and the layout accounts for the shorter cards.
- **Cardinality badge removed** (product decision) — the `1:1` / `N:1` label on join edges and its PK-based inference module are gone; join labels show the conditions only.

### Both canvases

- **Edges are neutral gray at rest** (corporate dark gray `#606060`, ≥3:1 contrast) and turn **brand-blue with a thicker stroke only when the user selects that specific edge**; pane click deselects. Arrow-head markers are rendered as inline SVG markers so they recolor with the stroke (React Flow's static marker props can't follow selection). Warning edges keep the orange dashed style. Shared primitives extracted to `shared/canvas/` (`erd-card.tsx`, `definition-type-accent.ts`, `edge-arrow.tsx`).
- **Clicking a data mart card highlights every edge connected to it** (and outlines the card in brand-blue), so all of the mart's relationships are visible at once. Clicking the card again or the pane clears it; single-edge selection and card selection supersede each other.
- **OWOX corporate color palette** (`shared/canvas/owox-palette.ts`, from the design team's colors sheet in `owox-models`) now sources every canvas color: definition-type accents (VIEW=blue `#4286DE`, SQL=green `#6ABA5C`, TABLE=purple `#9F4DB6`, PATTERN=red `#E15241`, CONNECTOR=orange `#E9883C`, fallback=gray), search-highlight blue + pulse ring, warning orange, missing-PK attention yellow, status dot, primary-key icon, and the neutral edge/border gray. Selection color stays on the product `--primary` brand blue.

## Testing

- `tsc -b`, ESLint and the full web build pass.
- 155 tests across the model-canvas and DataMartRelationships suites (object-labels state, per-part card hiding, loop/status filtering via the popover UI, meta-row geometry, node-click edge highlighting, delete-key hardening, controlled filters, layout re-flow).
- Manually verified: ERD cards and filters on the Joinable diagram (inline + fullscreen), position persistence across reloads, object-label toggles, edge selection color.

## Notes

- Task 6763: the existing `.changeset/6763-model-canvas-erd-redesign.md` (still unreleased) is updated in place to cover this PR — cardinality mention dropped, Joinable diagram / filters / layout persistence / object labels / edge selection colors added.
- Latest `main` is merged in (`f8b767b3c`): the data-quality canvas indicators, data-last-updated icon and topology-signature stabilization are woven into the branch's card layout; the upstream-touched `edge-cardinality` module is deleted again as part of the cardinality removal.
- Server-side enrichment of the canvas/relationship list endpoints (to avoid per-mart detail fetches) remains a possible follow-up.


apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx
```diff
@@ -544,19 +664,106 @@ function ModelCanvasInner({
...
       {ready && (
         <ReactFlow
-          nodes={flowNodes}
-          edges={flowEdges}
+          nodes={displayNodes}
+          edges={displayEdges}
           nodeTypes={nodeTypes}
           edgeTypes={edgeTypes}
           onNodesChange={onNodesChange}
+          onEdgesChange={onEdgesChange}
```
zapolsky, line 805: **[Bug — verified, empirically reproduced]** After clicking a data-mart card (or a join edge), pressing `Backspace` silently deletes it — and every edge touching it — from the Models canvas, with no way back through the UI.

**Evidence chain** (verified against `@xyflow/react` 12.11.2 as installed):

- `ModelCanvas.tsx:429-437` → `:443-449` / `:451-461` — the new `handleNodeClick` sets `selectedNodeId`, and `displayNodes` / `displayEdges` inject `selected: true` onto that node and onto every edge touching it.
- `ModelCanvas.tsx:755-756` — those arrays are handed to `<ReactFlow>`. `StoreUpdater` → `setNodes` (`@xyflow/react/dist/esm/index.js:3389-3419`) stores the user array verbatim, and `adoptUserNodes` (`@xyflow/system/dist/esm/index.js:1616-1662`) spreads `...userNode`, so `selected` survives into the store. The `selectable: false` at `ModelCanvas.tsx:194` does **not** strip it.
- `@xyflow/react/dist/esm/index.js:3728` — `deleteKeyCode` defaults to `'Backspace'`; there is no override anywhere in `apps/web/src`.
- `index.js:1235-1247` — `useGlobalKeyHandler` reads `store.getState().nodes` / `.edges` and calls `deleteElements({ nodes: nodes.filter(selected), edges: edges.filter(selected) })`. The `keydown` listener is on `document` (`index.js:352`, `:377`) with no focus-in-wrapper guard; the only suppressor is `isInputDOMNode` (`@xyflow/system:860-868`). So typing in the toolbar search box is safe — the canvas is not.
- `@xyflow/system:510-532` — `getElementsToRemove` skips only elements with `deletable === false`. Neither `buildFlowNode` (`ModelCanvas.tsx:183-217`) nor `buildFlowEdge` (`:232-252`) sets it, and it cascades through `getConnectedEdges` to every edge attached to the node. `onBeforeDelete` is not supplied, so there is no veto path.
- `index.js:3506-3536` — `triggerNodeChanges` / `triggerEdgeChanges` invoke `onNodesChange` / `onEdgesChange` **unconditionally**; controlled flows do *not* drop `remove` changes. `ModelCanvas.tsx:417-423` applies them straight into `flowNodes` / `flowEdges`.
- `ModelCanvas.tsx:415` — the rebuild effect's deps are all unchanged afterwards (`topologyNodes` / `topologyEdges` are signature-stabilised at `:160-167`), so nothing restores the removed elements.

**Reproduced** by rendering the real `ModelCanvas` against the unmocked `@xyflow/react` — the existing `ModelCanvas.test.tsx` mocks React Flow away, which is why the green suite cannot catch this:

```
AFTER CLICK      class="react-flow__node ... selected draggable"
AFTER BACKSPACE  nodes= [ 'orders', 'sessions' ]        // 'customers' gone
NODE CHANGES=    [{"id":"customers","type":"remove"}]
EDGE CHANGES=    [{"id":"e1","type":"remove"},{"id":"e2","type":"remove"}]
```

The identical test run against `origin/main` shows no `selected` class and removes nothing (`NODE CHANGES= []`), so this is a regression introduced by this PR: before it, nothing was ever `selected` and `onEdgesChange` did not exist.

**Trigger:** `/data-marts` → Models canvas → click any card (the new "light up all of its connections" gesture) → press `Backspace` with focus outside a text field. The card and all its edges disappear. Clicking a single join edge then pressing `Backspace` removes that edge — a second, independent trigger. Nothing is sent to the server, but the diagram stays wrong until the user reloads or toggles a setting in the gear popover; search and navigation do not restore it.

**Two corrections to the obvious diagnosis, both checked by counterfactual:**
1. `elementsSelectable` (`:766`) is *not* the load-bearing hop — patching it back to `elementsSelectable={false}` on this head still deletes the node, because `index.js:2228` (`isSelectable = !!(node.selectable || (elementsSelectable && typeof node.selectable === 'undefined'))`) makes it inert for elements that carry an explicit `selectable`. Reverting that line alone does not fix this.
2. `Delete` is not bound — React Flow's default is `Backspace` only.

**Fix:** add `deleteKeyCode={null}` to the `<ReactFlow>` at `ModelCanvas.tsx:754`; this canvas has no delete semantics at all. (Setting `deletable: false` in `buildFlowNode` / `buildFlowEdge` also closes it.) Worth adding the same prop to `RelationshipCanvas.tsx:987` for symmetry — that canvas is safe today only incidentally, because it passes no `onNodesChange` / `onEdgesChange`.

zapolsky (commented):
🔴 **Blocked issue** — see the inline comment. It should be addressed before merge, but this review does not formally block; a human owns the merge decision. Posted by automated OWOX frontend review.

zapolsky:

One blocked issue is posted inline (Backspace deleting the selected card from the Models canvas). Everything below is non-blocking. Nice PR overall — the localStorage handling is genuinely defensive, the subtree-filtering mechanism is correct, and the upstream `main` merge preserved the data-quality / data-last-updated / topology-signature behaviour (all verified, see the disproved list at the bottom).

**CI parity, run inside a worktree at `ac0d247`:** `npx tsc -b --pretty false` → exit 0, zero output (also with `--force`; `tsconfig.app.json:38` includes `src`, so the `.test.tsx` files *are* type-checked). `npx eslint` on all three changed directories → clean. `npx vitest run` on the two suites → **152/152 passed, 18/18 files** — the "152 tests" claim is exact. No pre-existing baseline errors; nothing here is unattributable to this PR.

### Primary notes (criticality ≥ 80)

- `apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx:523-531` — **Re-clicking the already-active layout algorithm destroys the saved layout with zero feedback.** Every radio in the *Layout algorithm* group calls `handleDirectionChange(option.value)` unconditionally (`:641-655`), including the one that already has the checkmark. The handler then unconditionally runs `savedPositionsRef.current = {}` + `storageService.remove(positionsStorageKey(...))` (`:527-530`) — but `setDirection(next)` with an identical value hits React's state bailout, so the layout effect (dep `direction`, `:415`) never re-runs and nothing moves on screen. The user clicks "Horizontal" while Horizontal is active expecting a re-flow, sees nothing happen, and their dragged arrangement is already gone from storage — it disappears on the next reload. Fix: early-return when `next === direction`, or bump a `layoutEpoch` state included in the effect deps so a re-pick actually re-flows.
- `apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx:754-774` — **On the Models canvas, edge selection does not supersede card selection** (the PR description claims they supersede each other). `handleNodeClick` clears edge selection (`:429-437`), but the `<ReactFlow>` wires no `onEdgeClick`, so nothing clears `selectedNodeId`. React Flow's own `addSelectedEdges` does emit a node-deselect change (`index.js:3547-3556`), but `displayNodes` re-injects `selected: true` on the next render (`:443-449`). Result: click card A, then click an unrelated edge → A stays outlined, all of A's edges stay blue, *and* the clicked edge is blue. The sibling implementation gets it right (`RelationshipCanvas.tsx:800-808`). Fix: `onEdgeClick={() => { setSelectedNodeId(null); }}`.
- `apps/web/src/features/data-marts/shared/canvas/edge-marker-id.ts:2` — **Inline and fullscreen Joinable diagrams emit duplicate SVG marker ids, so arrowheads stop following selection in fullscreen.** Both canvases are mounted at once (`DataMartRelationshipsContent.tsx:452` inline, `:633` inside the `<Dialog>`; `renderContent()` is not gated on `isFullscreen`), and `DialogContent` is portalled to `document.body` (`packages/ui/src/components/dialog.tsx:50`), so the inline DOM comes first. Both instances build identical edge ids (`RelationshipCanvas.tsx:500`, `:641`) and derive the marker id from the edge id alone, so `url(#rel-arrow-…-end)` (`:369`) resolves to the *inline* canvas's `<marker>`, whose fill reflects the inline instance's selection state. Click an edge in fullscreen: the stroke turns brand-blue and thickens, the arrowhead stays gray — defeating the stated reason for hand-rolling the markers, in the exact view where users read the diagram. (`origin/main` used `MarkerType.ArrowClosed`, whose ids are `rfId`-scoped precisely for this.) Fix: scope the id per instance, e.g. `useId()` or React Flow's `rfId` threaded into `edgeMarkerId`.
- `apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx:692-707` — **Filters changed in the fullscreen diagram never reach the inline one.** `showLooped` / `statusFilter` are per-instance `useState` whose lazy initializers read localStorage only at mount, and the handlers write localStorage without any `storage` listener or shared store. Since both instances stay mounted (chain above), turning on *Show looped Data Marts* in fullscreen and closing leaves the inline diagram on the old setting — it silently jumps to the new one later, on a List/Graph toggle or a reload. Fix: lift both into `DataMartRelationshipsContent` (next to `definitionTypes` at `:237`) and pass them to both instances.
- `apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx:237` — **N full data-mart detail GETs fire even in List view, where the diagram never renders.** `viewMode` defaults to `'table'` (`:127-130`) and the canvas chunk is deliberately lazy-loaded "only when the user switches to graph view" (`:46-50`), but `useRelationshipDefinitionTypes` is called unconditionally above that guard. Its only gate is `enabled: targetIds.length > 0` (`useRelationshipDefinitionTypes.ts:51`), and `targetIds` covers every accessible target at *every* depth (`:22-31`), each fetched via `GET /data-marts/:id` returning the full payload (including the SQL `definition`) to read one enum field. No `staleTime`, so it re-fires on every mount of the edit page. Fix: thread the view mode in and use `enabled: viewMode === 'graph' && targetIds.length > 0`, plus a `staleTime`.
- `apps/web/src/features/data-marts/shared/canvas/owox-palette.ts:24` — **Contrast regression on the resting edge / border / fallback-accent color.** `OWOX_GRAY_LIGHT = gray[3] = #C5C5C5` now backs `EDGE_NEUTRAL_COLOR` (`constants.ts:11`), `NODE_BORDER_COLOR` (`:4`), `SOCKET_STYLE.background` (`:21`) and `DEFINITION_TYPE_FALLBACK_ACCENT` (`definition-type-accent.ts:24`). On white that is **1.73:1** — the previous fallback `#94a3b8` was 2.56:1, and WCAG 1.4.11 wants 3:1 for graphical objects that carry meaning. It is worse on the Joinable diagram specifically, because this PR also removed the background grid, so there is no secondary cue for where edges run. `OWOX_YELLOW_BASE = #F5C344` (1.64:1) is likewise used for 10px bold text ("No primary key", `RelationshipCanvas.tsx:123` → `:180`). Separately: the PR description says edges are `#94a3b8`, but the shipped value is `#C5C5C5` — worth correcting the description either way. Fix: use a darker step (`gray[2] = #606060`, ~6:1) for edges/borders/handles in light mode, keeping `gray[3]` for fills.
- `apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx:314` — **"Cards render with a neutral accent until it arrives" is not what happens — they render a gray `—` pill.** `ErdDefinitionBadge` is rendered unconditionally, and `getInfo(null)` returns `{ displayName: '—', icon: () => null }` (`data-mart-definition-type.model.ts:38-45`), so `erd-card.tsx:37-43` paints a filled `#C5C5C5` pill with white uppercase text (1.73:1) containing a lone em dash. That is the state of **every** target card on first paint (`useRelationshipDefinitionTypes.ts:54` returns `EMPTY_MAP` until the query resolves) and the **permanent** state for marts the user has no access to (`:28` excludes them) or whose fetch failed (`:46`). Fix: `{data.definitionType && <ErdDefinitionBadge … />}`, or return `null` from `ErdDefinitionBadge` when `type === null`.
- `apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx:194` and `apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx:605` — **The new "click a card to light up all its connections" feature has no keyboard path on either canvas.** Nodes are built `focusable: false` and edges `focusable: false` / `edgesFocusable={false}` (`ModelCanvas.tsx:240`, `:767`; `RelationshipCanvas.tsx:645`, `:995`); React Flow only wires its `Enter`/`Space` selection handlers and sets `tabIndex` when `isFocusable` (`index.js:2286`, `:3002`). So the headline interaction is mouse-only. On the Joinable diagram there is not even a hover affordance: `STATIC_NODE_STYLE` forces `cursor: 'default'` (`constants.ts:28`) and the edges never match React Flow's `.selectable { cursor: pointer }` rule. Not a regression (nothing was selectable before), but a gap in brand-new functionality. Fix: `focusable: true` + an `ariaLabel` on the nodes so `Enter` reaches `handleNodeClick`; at minimum `cursor: 'pointer'`.
- `apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx:100-101` — **The two new filters are stored in one global localStorage key, with no active-filter indicator and no empty state.** `'relationship-canvas-show-looped'` / `'relationship-canvas-status-filter'` are scoped neither by data mart nor by project — unlike the position key this same PR namespaces per storage (`ModelCanvas.tsx:79-81`). Since filtering a node also drops its whole subtree (`:469-473` + `:511-513`), setting *Status = Published* on one data mart silently truncates every other data mart's diagram, potentially down to a single lone root card, and the gear button carries no badge or count to say why (`:937-946`). Compare the Models canvas, which keeps its equivalent filters in the URL and a visible toolbar plus an explicit "No data marts match the current filters" state (`ModelCanvasView.tsx:218-234`, `:264-265`). Also worth a release note: `showLooped` now defaults to `false`, so on this deploy every existing user's diagram loses its loop stubs with nothing on screen explaining it. Fix: scope both keys per data mart (or project), show an indicator on the gear when the state is non-default, and render a "no matches" message when the graph reduces to the root alone.

<details>

- `apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx:500` — node keys are positional (`${dmId}-${nodeCounter++}`), and the counter shifts whenever a filter skips an earlier node (`:513`, `:539`). `selectedNodeId` / `selectedEdgeId` are never reset when the filters change (`:766-767`), so toggling a filter usually just drops the highlight silently, and — when the same mart appears at two alias paths — can move it to the wrong occurrence. Every node id also changes on a filter toggle, so React Flow remounts every node rather than updating it. Keying by `aliasPath` (already unique, `:534`) and clearing both selections in the existing `graphIdentity` effect (`:745-749`) fixes both.
- `apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx:415` — `objectLabels` is in the big layout effect's deps, and the effect ends in `requestAnimationFrame` → `reactFlow.fitView()` (`:400-410`). So unticking "Status dot" — purely cosmetic, it does not change node heights (only `source && fields` does, `:326`) — re-runs dagre and throws away the user's pan/zoom. The same happens for a genuine no-op: `handleObjectLabelsChange(NOTHING_HIDDEN)` (`:723-725`) passes the module constant while the live state is an equal-but-distinct object from `parseObjectLabelsHidden`, so clicking "Check all" when everything is already checked still re-fits.
- **Question (unproven):** `apps/web/src/features/data-marts/model-canvas/model/erd-node.ts:22` — is `CARD_META_ROW_HEIGHT = 36` measured? The row it models (`ModelCanvasFlowNode.tsx:188-197`: `pt-1` = 4px plus a `py-0.5 text-[10px]` pill) reads as ~20–23px, and the same 36 is subtracted from both `COMPACT_NODE_HEIGHT = 116` and `ERD_HEADER_HEIGHT = 88` even though those two describe the same three rows and differ by 28px. If it over-subtracts, `node.height` — which is also React Flow's inline wrapper height — ends up shorter than the rendered card, biasing `fitView` and the pan clamping. No overlap in practice (dagre's default `nodesep`/`ranksep` of 50 absorbs it), and we could not measure it: the vitest env is happy-dom with no layout engine and the suite mocks React Flow. `erd-node.test.ts:60-68` restates the formula from the same constants, so it cannot catch a wrong value. Worth one check in the browser with "Uncheck all — title only".
- **Question:** `apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx:780` — when "Input source" is hidden, the badge and accent stripe both go (they "encode the same thing"), but the minimap still colors nodes by definition type (`nodeColor={node => definitionTypeAccent(node.data.definitionType)}`, same at `RelationshipCanvas.tsx:1013`). The comment on the module this replaced said the helper existed so "the header stripe, the badge **and the minimap dot** stay in sync". Intentional, or should the minimap fall back to neutral too?
- `apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx:463-473` — `onNodeDragStop` never prunes entries for data marts that no longer exist, so `model-canvas-positions:<storageId>` grows monotonically over the life of the browser profile.
- minor: `shared/canvas/constants.ts:7` — `EDGE_COLOR = 'steelblue'` is now a dead export (zero consumers repo-wide), and `NODE_BORDER_COLOR` (`:4`) is only referenced inside its own file; both can be dropped while the module is being reorganised. · `shared/types/data-mart-definition-type.model.ts:46` — `return this.types[type]` yields `undefined` for a definition type the FE enum does not know, which would make `erd-card.tsx:35` render `<Icon>` as `undefined` and take the canvas down; the FE and backend enums are currently identical so it is unreachable today, but a `?? { type: null, displayName: '—', icon: () => null }` would make it safe by construction. · The PR description says "14 added" tests — the real delta is +12 added / −8 deleted (net +4, `edge-cardinality.test.ts` accounting for the deletions); the "152 tests" total is exact.

</details>

<details>
<summary>Hypotheses checked and ruled out (so they don't get re-raised)</summary>

- **Unguarded `JSON.parse` / corrupt or old-shape localStorage breaking a canvas** — `localstorage.service.ts:46-52` catches and returns `null`; `loadSavedPositions` (`ModelCanvas.tsx:85-97`) re-validates every entry as `{x:number,y:number}` (so arrays, `null`s and `NaN`-serialised values are skipped); `parseObjectLabelsHidden` drops unknown tokens and handles the `''` that `serializeObjectLabelsHidden(NOTHING_HIDDEN)` writes; `parseStatusFilter` whitelists. Quota / private-mode throws are swallowed at `localstorage.service.ts:20-22`.
- **"Filtering a node hides its subtree" / "the root is always shown"** — both hold. A filtered node never registers in `aliasPathToNodeKey` (`RelationshipCanvas.tsx:504` is inside `addEdgeAndNode`, skipped by the `continue` at `:513`), so descendants die on `:512`; the root is inserted unconditionally at `:480-488` and never passes through `passesFilters`. The single-pass filter is safe because the backend emits DFS pre-order (`get-data-mart-relationship-graph.service.ts:133-152`, asserted by its own spec).
- **Stale/duplicate async writes from the enrichment hook** — it holds no local state, keys the React Query cache by the sorted id list, passes the `AbortSignal`, and `Promise.allSettled` absorbs individual rejections; `data ?? EMPTY_MAP` keeps a stable module-level identity while loading, so the `graphResult` memo does not thrash.
- **`nodeTypes` / `edgeTypes` re-created per render remounting every node** — both are module-level constants on both canvases.
- **`edgeMarkerId`'s `[^a-zA-Z0-9_-] → '_'` collapsing two distinct edge ids** — Models ids are UUIDs or `uuid+uuid`, relationship ids are `uuid-N->uuid-M`; neither alphabet contains `_`, so no aliasing *within* one canvas (the cross-instance collision above is a different mechanism).
- **Dangling references to the deleted `edge-cardinality` module** — zero matches repo-wide (no barrel, no type re-export, no test).
- **The upstream `main` merge dropping data-quality / data-last-updated / topology-signature behaviour** — all preserved: both icon components still render (`ModelCanvasFlowNode.tsx:198-207`), the live-quality patch effect (`ModelCanvas.tsx:475-485`) and the `useStableValue` stabilisation (`:142-167`) are untouched, and quality polling cannot wipe dragged positions because `getNodeTopologySignature` excludes `qualitySummary`.
- **The Joinable diagram sharing the Backspace bug** — it does not: it passes no `onNodesChange` / `onEdgesChange`, so the `remove` changes land nowhere. (Note this is incidental, not by design — controlled flows do *not* drop `remove` changes on their own.)
- **Dragging a card spuriously toggling the new highlight** — `d3-drag` installs a capture-phase click suppressor for any gesture that moved, so no `onNodeClick` follows a real drag.
- **`saved position wins over dagre` / `Compact-Detailed switching and search keep positions`** — both verified.

</details>

<sub>Automated OWOX frontend review · worktree checkout at `ac0d247` · 4 parallel analysis agents (line-by-line, spec-vs-implementation, tester, un-checklisted fresh-eyes) · the blocker was adversarially verified by an independent skeptic and reproduced against the unmocked library.</sub>

apps/web/src/features/data-marts/model-canvas/components/ModelCanvas.tsx
laskevych: Fixed in 76b788fd2 — both canvases now pass `deleteKeyCode={null}`, and the Models canvas additionally builds nodes/edges with `deletable: false` (belt and suspenders, per your note that `getElementsToRemove` honors it). Covered by tests asserting `deleteKeyCode` is null and every element is non-deletable on both canvases. Thanks for the reproduction against the unmocked library — that made this an easy fix to trust.

laskevych:
Review addressed in 76b788fd2 — thanks, this was an unusually useful pass. Status per finding:

**Blocker**
- ✅ Backspace deletion: `deleteKeyCode={null}` on both canvases + `deletable: false` on Models-canvas nodes/edges; regression-tested.

**Primary notes**
- ✅ Re-picking the active layout algorithm: a `layoutEpoch` state now forces a genuine re-flow, so clearing the saved positions is always visible and intentional (test added).
- ✅ Edge click now clears card selection on the Models canvas (`onEdgeClick` wired), so the two modes truly supersede each other (test added).
- ✅ Duplicate SVG marker ids: marker ids are scoped per rendered edge instance via `useId()`, so the fullscreen diagram's arrowheads follow its own selection.
- ✅ Fullscreen ↔ inline filter desync: `showLooped` / `statusFilter` are lifted to `DataMartRelationshipsContent` and passed to both instances (the canvas keeps an uncontrolled fallback). The shared model lives in `relationship-canvas-filters.ts` so the lazy canvas chunk stays out of the main bundle.
- ✅ N detail GETs in List view: enrichment is gated on `viewMode === 'graph'` and gets `staleTime: 5m`.
- ✅ Contrast: resting edges / borders / handles / fallback accent now use the corporate dark gray `#606060` (≥3:1); the attention label text is dark gray with the corporate yellow kept on the triangle glyph. (PR description updated — the `#94a3b8` claim was stale.)
- ✅ Gray `—` pill: `ErdDefinitionBadge` renders nothing while the type is unknown; `getInfo` also falls back safely for enum values the FE doesn't know.
- ✅/➡️ Filter UX: keys are now scoped per data mart, the gear shows an active-filter dot, and a "No related data marts match the current filters" notice appears when the graph reduces to the root. Keyboard path for the click-to-highlight gesture is acknowledged as a real gap — deferred to a follow-up (React Flow's focus/selection model needs the highlight to be derived from RF selection state, which is a larger change than this PR should absorb).

**Additional observations**
- ✅ Positional node keys → alias-path keys (stable across filter toggles); selections cleared on graph identity change.
- ✅ No-op object-label picks no longer re-run layout / reset pan-zoom.
- ✅ Saved positions are pruned of deleted marts.
- ✅ Dead `EDGE_COLOR` export removed; `NODE_BORDER_COLOR` folded into the module.
- ❓ `CARD_META_ROW_HEIGHT = 36`: fair question — it's an estimate, not a measurement. Will verify in the browser with "Uncheck all — title only" during the next manual pass and adjust if the wrapper height visibly under-measures.
- ❓ Minimap color with "Input source" hidden: intentional for now — the minimap is an orientation aid rather than a card label, and a fully neutral minimap loses its only distinguishing cue. Happy to revisit if design disagrees.

155 tests green (5 new for the fixes), `tsc -b` / ESLint / full build clean.

zapolsky (approved):

PR #1480, opened 2026-07-30
fix(connectors): stop copied config from sharing another Data Mart's credentials

prokopetsv:
## Problem

Steps to reproduce:

1. Create a connector-based Data Mart_1, select a source, fill in manual credentials, save, and publish.
2. Create Data Mart_2, select the same source, use **Copy from…** to reuse DataMart_1's configuration.
3. Save and refresh.

The credentials disappear from **both** Data Marts.

## Root cause

Manual (non-OAuth) credentials are not stored in the Data Mart definition. They live in
`connector_source_credentials`, and the config item only holds a pointer, `_secrets_id`.
That pointer is scoped 1:1 to a Data Mart — the record is stamped with `dataMartId` + `configId`,
and `deleteSecretsByDataMart` assumes that ownership.

`ConfigurationStep.handleCopyConfiguration` spreads every field of the source config except
`_id`/`_copiedFrom`, so `_secrets_id` came along, and nothing on the way in dropped it. The copied
config therefore kept pointing at Data Mart_1's record: `extractAndSaveSecrets` took its *update*
branch and wrote 2's credentials into 1's record, and a later save then deleted that record as
orphaned — hence both going blank.

(`list-data-marts-by-connector-name` returns the raw stored definition, so `_secrets_id` reaches
the browser in the first place.)

## Fix

The invariant is **a credentials record belongs to exactly one Data Mart**. Rather than patch the
copy path alone, this enforces it where every save already passes.

- **`extractAndSaveSecrets` compares the record's owner with the Data Mart being saved.** On a
  mismatch it creates a record of its own instead of writing through the pointer it was handed.
  This is the core change: it makes any path that produces a foreign pointer harmless, not just the
  one the UI uses.
- **`mergeDefinitionSecretsFromSource` drops `_secrets_id` on a copied item**, so a copy starts from
  a clean record rather than relying on the check above. This is now uniform across connectors: the
  `copySecretsByValue` special case is gone, every connector copies secrets by value.
- **Orphan cleanup compares the `_secrets_id` values of the previous and current definitions**
  instead of configuration `_id`s. A record now survives while any configuration item still
  references it — the previous check deleted records that two items shared — and a record is
  reclaimed when an item keeps its `_id` but drops its pointer, which is what happens when it
  switches to OAuth and used to leak the old record for the lifetime of the Data Mart.
- **Ownership is part of the delete API.** `deleteCredentials(id)` is replaced by
  `deleteCredentialsByIdsAndDataMart(ids, dataMartId)`, which scopes the statement in SQL and
  deletes in one round trip; ownership lookups select only `id` and `dataMartId`, so authorizing a
  delete no longer loads secret payloads. Skipped records are logged.
- **A copy is seeded with the source's rotated refresh token, in its own record.** Microsoft Ads
  mints a new refresh token on every run (`generated_refresh_token`, #1308). The copy takes the
  source's current one as the starting value of its own credentials record: Microsoft
  [does not revoke](https://learn.microsoft.com/en-us/entra/identity-platform/refresh-tokens) a
  redeemed refresh token when it issues a new one, and there is
  [no reuse detection and no cap](https://learn.microsoft.com/en-us/answers/questions/988508/multiple-valid-refresh-tokens-at-the-same-time)
  on concurrently valid tokens, so the source and the copy rotate independent token lineages and
  neither breaks the other. Without the seed, a copy of a Data Mart older than the token's ~90-day
  lifetime would start from the long-expired user-supplied token and fail with `invalid_grant`
  until re-authenticated. The seed travels only while the copy authenticates with the source's
  exact refresh token, and never into managed-OAuth configs.
- **A secret field left holding only the mask is dropped** rather than persisted as if the
  placeholder were the credential.
- **Deleting a Data Mart hands over records another Data Mart still references.** Ownership
  enforcement separates legacy shared pairs only when the referencing Data Mart is saved, so
  deleting the owner first would still have wiped the copy's credentials. The delete flow now
  transfers such a record to the first live referencer — the pointer stays valid, nothing is lost —
  and any further referencers fork onto their own records on their next save. Unreferenced records
  are deleted as before.
- **Every copy source is authorized, and secret references are validated, for every connector.**
  Both checks existed for `copySecretsByValue` connectors (Google Sheets) and were gated on that
  capability; with copy-by-value now uniform, the gates are gone. Copying requires `EDIT` on the
  source Data Mart (technical owner, or shared for maintenance), and a caller-supplied
  `_secrets_id` must match the Data Mart's own stored definition or the named copy source —
  otherwise the save is refused with a 403 instead of relying on the chokepoint drop alone.

## Already-affected Data Marts

No data migration is needed for the aliasing. Because the ownership check runs on every save, two
Data Marts that currently share a record are separated the first time each of them is saved: the
one that does not own the record forks onto its own copy of the values.

Credentials that were already **wiped** by this bug are not restored — those Data Marts need their
credentials entered again.

## Not fixed here

- Connections authorized through OAuth are still shared between Data Marts. That is by design:
  such a credential belongs to the project, not to a single Data Mart.
- The "Copy from…" picker still lists Data Marts the save will now refuse with a 403: it is fed by
  `GET /data-marts/by-connector/:connectorName`, which applies no visibility filter. Filtering the
  picker belongs together with fixing that listing — recorded as a follow-up (see #1485's closing
  notes).

## Tests

`connector-secret.service.spec.ts` + `update-data-mart-definition.service.spec.ts` +
`delete-data-mart.service.spec.ts` — 84 passing, including: a record still referenced by another
Data Mart handed over on delete while unreferenced ones are removed; the copy-source access check
and the unowned-secret-reference rejection for a connector without `copySecretsByValue` (with the
self-heal of a pointer coming from the Data Mart's own stored definition staying open); the ownership fork and that a same-project pointer into another Data Mart is not written
through; the project mismatch still throwing; a record kept while a second item references it; a
record reclaimed after a switch to OAuth; own and foreign records handled in one batch; a record
that no longer exists skipped; the mask dropped instead of stored; the copy seeded with the
source's rotated token in its own record — and not seeded when the copy's refresh token changes or
the config is managed OAuth; and a use-case level test that cleanup runs against the definition
that was actually saved.

The copy test that previously covered this bug was passing for the wrong reason — its fixture
combined `_secrets_id` with an inline plaintext value, which production never stores — so it now
mocks the credentials record the way the real flow provides it.

After merging `main`, two main-side tests were reconciled with this branch's invariant: the #1466
test that pinned pointer-sharing for non-`copySecretsByValue` connectors is replaced by the seeding
test above, and the #1495 same-type-update test now asserts the
`deleteOrphanedSecrets(dataMartId, currentDefinition, previousDefinition)` signature.

## Follow-ups (separate PRs)

- Copying requires permission on the source Data Mart, and copying from several Data Marts in one
  save.
- `update-data-mart-definition` runs credentials writes and deletes outside a transaction.


apps/backend/src/data-marts/services/connector/connector-secret.service.ts
```diff
@@ -818,10 +902,18 @@ export class ConnectorSecretService {
       ) as Record<string, unknown>;
 
       delete mergedItem._copiedFrom;
-      if (copySecretsByValue) {
-        delete mergedItem._secrets_id;
-      }
+      // A copied config must never keep the source's _secrets_id: that id is
+      // scoped 1:1 to the source DataMart's own config item, and reusing it
+      // would alias both DataMarts' manual credentials to the same record.
+      delete mergedItem._secrets_id;
```
amarchenk0, line 908: This path now copies secret values for every connector, but `UpdateDataMartDefinitionService` still gates both source `EDIT` authorization and secret-reference validation on `capabilities.copySecretsByValue`. Only Google Sheets declares that capability. For a connector such as Facebook Marketing, an editor of the target Data Mart can submit a `sourceDataMartId` and `_copiedFrom` for a source they cannot edit: `getByIdAndProjectId` performs no access decision, and this method loads the source secret record and creates a new credential record. That bypasses authorization to use manual credentials. Require source access on every copy path and add a test that a connector with `copySecretsByValue: false` rejects this request.

amarchenk0 (changes requested):
Changes requested.

The ownership-safe cleanup only runs on save. Existing aliases are not migrated, while `DeleteDataMartService` still calls `deleteSecretsByDataMart()`, which soft-deletes every record owned by the Data Mart. If the original Data Mart is deleted before its copy is manually saved, the copy still points to the deleted record and loses its credentials. These are precisely the already-affected records this fix needs to remediate safely; it cannot rely on every user saving the copy first. Split or clone aliases safely before deletion, or run a repair migration, and add coverage for this scenario.

apps/backend/src/data-marts/services/connector/connector-secret.service.ts
laskevych: Fixed in 15e1bcc29. Both checks lost their `copySecretsByValue` gate: `validateCredentialCopyAccess` now requires EDIT on the source Data Mart for every connector's copy path, and `validateSecretReferences` rejects a caller-supplied `_secrets_id` that matches neither the mart's own stored definition nor the named copy source — on top of the chokepoint drop from 75cb69799. New tests cover the 403 for a non-`copySecretsByValue` connector in both scenarios, plus the self-heal pointer staying accepted.

laskevych:
Re the requested changes on deletion: fixed in 15e1bcc29. The delete flow now hands a record still referenced by another live Data Mart over to that referencer (`DeleteDataMartService.handOverSecretsStillInUse` → `transferSecretsOwnership`) instead of soft-deleting it; unreferenced records are deleted as before. Covered by a test with a referenced and an unreferenced record, including the deleted mart's own reference not counting as an heir. This keeps the no-migration approach: pairs still separate lazily on save, and deletion can no longer race that.

amarchenk0 (approved):

PR #1481, opened 2026-07-31
docs(data-marts): Data Last Updated user guide (#6750)

laskevych:
## Summary

Adds a dedicated user-guide page for **Data Last Updated**: `docs/getting-started/setup-guide/data-last-updated.md`, registered in the docs sidebar right after Data Quality Checks.

**Base is the feature branch, not `main`** — the page documents UI that exists only on `feat/6750-data-last-updated-date` (PR #1472). Merging this PR folds the docs into the feature PR; nothing lands on `main` before the feature does.

## Why this location

`docs/getting-started/setup-guide/` is where per-feature guides live (`data-quality-checks.md`, `output-controls.md`, `report-aggregations.md`, `joinable-data-marts.md`). Data Last Updated sits beside Data Quality Checks in the sidebar because the two answer neighbouring questions about a Data Mart's state — and because they need explicit disambiguation:

- Data Quality already ships a check named **"Data freshness"** that validates a timestamp *field inside the data* against a configured threshold.
- **Data Last Updated** is *storage metadata*: when the source tables were last written to.

The page spells out the difference with the backfill example (Data Last Updated says "today", the Data freshness check still fails), so the adjacent names stop being a trap.

## What the page covers

- What the value means: a storage write time, not data recency; **Unknown** = could not determine (neither stale nor fresh); **≥** = partial coverage, a floor.
- Where it appears: Data Mart page tile, sortable list column, canvas clock icon + toolbar sweep, and the MCP `data_last_updated` block (cross-linked with `mcp.md`, which now links back).
- How to check it: the two buttons; free of credits, no run recorded; a failed check never erases the last-known value.
- When it updates: manual checks persist; MCP measures live per query without overwriting; never on a schedule.
- How OWOX determines it: views resolved to base tables, sharded sets collapsed to the newest shard (prefix-only neighbours like `events_backup` excluded), external tables and the ~50-table cap reported as partial coverage.

## Verification

- `markdownlint-cli2`: 0 errors on both touched pages.
- Full docs site build (`apps/docs npm run build`): 147 pages, links validation green — the new page and both cross-links resolve.

One heads-up from running the build locally: the docs sync step sweeps every root-level `*.md` into the site, so untracked local working notes break the local build on link validation. Not an issue for CI (the notes are untracked), but worth knowing if the build fails locally with links from files you never put in `docs/`.

### DIFF
--- a/apps/docs/astro.config.mjs
+++ b/apps/docs/astro.config.mjs
@@ -74,6 +74,7 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
                 'docs/getting-started/setup-guide/pattern-data-mart',
                 'docs/getting-started/setup-guide/joinable-data-marts',
                 'docs/getting-started/setup-guide/data-quality-checks',
+                'docs/getting-started/setup-guide/data-last-updated',
                 'docs/getting-started/setup-guide/output-controls',
                 'docs/getting-started/setup-guide/report-aggregations',
                 'docs/getting-started/setup-guide/connector-triggers',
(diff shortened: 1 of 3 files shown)


laskevych:
Re-doing per team flow: the docs branch should be cut from main, not from the feature branch. Replacement PR (base: main) incoming.

PR #1484, opened 2026-07-31
feat(data-marts): refresh Data Last Updated with every data-delivering run

laskevych:
## Summary

This PR completes the original meeting decision — **measure Data Last Updated whenever data is delivered anyway (MCP query / report run)** — so the value keeps itself current for actively-consumed Data Marts without anyone pressing a button.

- **Report runs** (manual *and scheduled*) measure in parallel with the read/write, journal the snapshot on the run record, and — for a non-blended run — save it as the Data Mart's last-known value.
- **HTTP Data streams** do the same, journalling into the run metadata.
- **MCP queries**, which already measured and journalled, now also persist for non-blended queries.
- On the **canvas**, the check moves from a standalone toolbar button into the **Actions** menu beside Check Quality, targeting the same filtered set as the other bulk actions.

## The persist rule

A non-blended run reads exactly the Data Mart's **own** sources — output controls only narrow rows, never widen the FROM — so its measurement means the same thing as the manual *Check now* and is safe to save. A blended run spans several joined Data Marts; saving its reading onto one Data Mart would overstate it, so blended runs **journal into their run record only**. Every path uses the `needsBlending` flag the run had already resolved:

| Path | Blending signal | Journal | Persist |
| --- | --- | --- | --- |
| Report run | `blendingDecision.needsBlending` | `reportDefinition.dataLastUpdated` (saved at `finish()`) | non-blended only |
| HTTP Data | `decision.needsBlending` | run metadata (`httpData`) | non-blended only |
| MCP query | `compose()` now returns the flag it already resolved | run metadata (`mcpQuery`) — as before | non-blended only (new) |

`SourceDataLastUpdatedService.resolveForDefinition()` is the shared definition-scoped entry point the non-blended run paths use — the identical scope the manual refresh measures, so the persisted value has one meaning everywhere.

## Design notes

1. **Measurement overlaps the run.** It starts right after the blending decision and is awaited after the data is written, so its 1–2s hides behind the run's own read/write. Best-effort contract throughout: it never rejects, caps itself, and a run that delivered data can never fail over its metadata.
2. **No second blending resolution.** `compose()` return gains `needsBlending` (additive; existing callers destructure only `sql`/`params`), so the MCP path learns the answer for free.
3. **Failed report runs skip journal/persist** — their measurement promise resolves silently (never-rejects contract), so nothing dangles. The HTTP stream journals and persists *before* streaming begins: the measurement describes warehouse state, not delivery, so its validity does not depend on whether the stream completes (same as the manual check, which delivers nothing at all).
4. **The Actions item is optional on `DataMartBulkActions`** (`onCheckDataLastUpdated?`), so the list-selection usage of the same menu is untouched. No confirmation dialog — unlike Check Quality, the check is free of consumption. A success toast reports the count, since the menu closes on select.
5. **Run-record schemas extend optionally** (`dataLastUpdated` on `reportDefinition` and HTTP metadata), so runs recorded before this change still parse.

## Fix found in manual testing

The report-run persist was being silently reverted: `Report.dataMart` is declared with `cascade: true`, and `reportRunService.finish()` re-saved the full Report entity when the run completed — cascading the **stale** in-memory Data Mart snapshot loaded at run start over the value persisted moments earlier. The run journal (a separate table) stayed fresh, which is how the mismatch surfaced.

Fixed structurally after review feedback: `finish()` now persists the run outcome as a **targeted column update** (`ReportService.updateLastRunOutcome` — status + error only, with an explicit NULL clear), so no terminal path full-saves the Report anymore and the cascade window is gone on all of them, not just non-blended success. The in-memory refresh in `recordDataLastUpdated` stays as defence against any future full save. On top of that, `DataMartService.updateDataLastUpdated` now drops any measurement whose `computedAt` is older than the saved one — the value can never move backwards regardless of writer ordering — and scopes the write by `projectId`.

## Testing

- `run-report.service.spec.ts` (+2): non-blended run journals **and** persists (in DB *and* on the in-memory entity the finish-time cascade re-saves); blended run measures the blended SQL, journals, and does **not** persist.
- `stream-http-data.service.spec.ts` (+2): same pair for the stream path, asserting the journalled run metadata.
- `query-data-mart.service.spec.ts` (+3): persists for non-blended, not for blended, not for unresolved.
- `report-sql-composer.service.spec.ts`: exact-match assertions aligned with the additive `needsBlending` field in the `compose()` result.
- Suites: backend 138+ across the touched specs, all green; web 842 green; typecheck at baseline on both apps; lint clean.
- Manually verified end-to-end: a manual report run against a freshly-INSERTed BigQuery table now persists the fresh value (journal and DB column agree).

## Canvas UX follow-ups (manual testing round 2)

- **Fresh values now reach the nodes without a reload.** The canvas layout effect deliberately reacts only to topology changes; quality summaries already had a data-sync effect that carried live values past it, but `dataLastUpdated` did not — so a finished sweep was invisible until refresh. The sync effect now carries both.
- **The check announces itself like Check Quality does**: while the sweep runs, every node's Data Last Updated icon spins (`RefreshCw`), and the icons settling into fresh values is the success signal — the loading/success toasts are gone. Failure toasts stay, still honest about partial coverage.
- The Actions item is renamed **Check Data Last Updated**, naming the action rather than the metric.
- **`VIEW` accent is now OWOX corporate yellow** (was blue), per design.
- **"Uncheck all — title only" now truly means title only**: the quality indicators row (Data Quality shield + Data Last Updated clock) hides too, and the node height/layout accounts for the dropped row.

## Docs & changeset

- `mcp.md`: one sentence noting the non-blended persist.
- `data-last-updated.md` (landed via #1482, updated here after merging main): the "When the value updates" table now has rows for report runs (manual *and scheduled*), HTTP Data, and MCP with the persist rule, and the "On a schedule → Never" row that triggered this discussion is reframed as "In the background, on its own → Never — a scheduled report run counts as a run". Toolbar-button mentions replaced with the Actions menu.
- The two `6750-*` changesets are merged into a single `6750-data-last-updated-date.md` (one `'owox': minor` entry covering the whole feature).

## Deployment

No migration, no env vars. Additive schema fields, additive `compose()` return. Non-BigQuery storages keep reporting `unavailable`; their runs journal an honest unknown and never persist.


zapolsky (commented):
🔴 **Blocked issue** — see the inline comment. It should be addressed before merge, but this review does not formally block; a human owns the merge decision. Posted by automated OWOX review.

apps/backend/src/common/scheduler/services/fetchers/trigger-fetcher.service.ts
```diff
@@ -205,19 +205,7 @@ export class TriggerFetcherService<T extends Trigger> {
...
-
-    const ttlStartTime = new Date(this.systemClock.now().getTime() - this.triggerTtlSeconds * 1000);
-    const { affected } = await this.repository.delete({
-      createdAt: LessThanOrEqual(ttlStartTime),
-    } as FindOptionsWhere<T>);
-
-    if (affected) {
-      this.logger.log(`[${this.entityName}] ${affected} triggers deleted due to TTL.`);
-    }
+    return;
```
zapolsky: **[Bug — verified]** `deleteTriggersByTtl()` has been replaced with a bare `return;`, silently disabling TTL cleanup for every trigger table in the product. This is unrelated to the PR's stated scope and reads like a local debugging edit that got committed.

**Evidence chain**

- `apps/backend/src/common/scheduler/services/fetchers/trigger-fetcher.service.ts:207-209` — the whole body (config guard, `LessThanOrEqual(ttlStartTime)` delete, affected-rows log) is now `return;`. The JSDoc at `:199-206` still documents the deletion it no longer performs.
- Introduced by this PR's own feature commit `9bf219770` ("refresh Data Last Updated with every data-delivering run (#6750)"), whose message says nothing about the scheduler. `git show origin/main:...trigger-fetcher.service.ts` still has the original body.
- Still invoked every dispatcher tick: `trigger-fetcher.service.ts:63` (`await this.deleteTriggersByTtl()` inside `fetchTriggersReadyForProcessing`).
- The TTL is a first-class handler contract — `common/scheduler/shared/trigger-handler.interface.ts:59` → `common/scheduler/facades/scheduler-facade.impl.ts:80` → `trigger-fetcher-factory.service.ts:50` → `trigger-fetcher.service.ts:41`. **13 handlers declare a TTL that is now dead**: `report-run-trigger-handler.service.ts:160` (23 h), `connector/connector-run-trigger-handler.service.ts:171` (23 h), `data-quality-run-trigger-handler.service.ts:122` (23 h), `schema-actualize-trigger-handler.service.ts:68` (1 h), `sql-dry-run-trigger-handler.service.ts:101` (1 h), `ai-helper-trigger-handler.service.ts:128` (1 h), `publish-drafts-trigger-handler.service.ts:72` (1 h), `insight-run-trigger-handler.service.ts:76`, `insight-template-run-trigger-handler.service.ts:79`, `insight-artifact-sql-preview-trigger-handler.service.ts:74`, `ai-assistant-run-trigger-handler.service.ts:75` (3 h each), `search/indexing/search-reindex-trigger-handler.service.ts:76` and `:119` (24 h / 2 h).
- **Nothing else reaps those rows.** There is no `repository.delete` left anywhere under `common/scheduler`; the only trigger deletes in the codebase are user-initiated and target the *config* table, not the ephemeral run triggers (`data-marts/use-cases/delete-scheduled-trigger.service.ts:33`, `data-marts/services/scheduled-trigger.service.ts:125` — both `DataMartScheduledTrigger`).
- **Collateral:** `data-marts/services/base-run-trigger-handler.service.ts:134-153` — `cleanupOrphanedRuns` exists specifically to handle "*a trigger deleted by TTL cleanup while the run is still pending*" and matches on `.andWhere('trigger.id IS NULL')`. With TTL deletion dead, that predicate never matches, so this cleanup becomes a no-op that still runs on a timer.

**Why CI is green on it** — `trigger-fetcher.service.spec.ts` has zero TTL coverage (`grep -i ttl` returns nothing; `repository.delete` is never asserted, only `find`/`update`; all 7 tests sit under `describe('fetchTriggersReadyForProcessing')`), and `triggerTtlSeconds` (`:41`) is a *parameter property*, so the now-dead constructor arg raises no unused-variable error. `tsc -b` and `eslint` are both clean on this file. Nothing in the repo would go red if this ships.

**Impact** — every trigger row created from now on stays forever. `sql_dry_run_trigger` and `schema_actualize_trigger` are UI-driven and were designed to be swept hourly. The dispatcher's own `repository.find(findOptions)` (`:120`) then scans a monotonically growing table on every tick, and on the SQLite self-hosted default the DB file grows without bound. Failure surfaces months later, in a commit titled "refresh Data Last Updated".

**Fix** — revert this hunk: `git checkout origin/main -- apps/backend/src/common/scheduler/services/fetchers/trigger-fetcher.service.ts`. If disabling TTL cleanup is genuinely intended, it belongs in its own PR with its own reasoning, a changeset, and a spec.

zapolsky:

One blocked issue is posted inline (trigger TTL cleanup gutted to `return;`). Everything below is non-blocking.

Worth saying up front: the new backend tests were **mutation-tested**, not just read — deleting `run-report.service.ts:460`, moving the `baseMetadata` assignment past `recordSuccessfulRun`, hardcoding `needsBlending`, and flipping the composer's flag each turned a test red. The suites are real, not shaped around the mocks. Backend 175/175 green across the five touched specs, scheduler 7/7, web 73/73 across the nine canvas specs.

Primary notes (criticality ≥ 80), highest first:

- `apps/backend/src/data-marts/use-cases/run-report.service.ts:456` — **the cascade clobber is fixed on one branch out of five, but the comment at `:457-459` reads as if solved.** Your diagnosis is right, and I verified the mechanism empirically rather than from memory: TypeORM 0.3.31 turns `cascade: true` (`entities/report.entity.ts:51`) into `isCascadeUpdate` (`metadata/RelationMetadata.js:156-159`), `SubjectDatabaseEntityLoader` re-SELECTs the row at save time, and `SubjectChangedColumnsComputer` diffs *in-memory vs freshly-loaded DB row* — so a column another process changed is reverted, not skipped. A standalone repro with the same entity shape emits the `UPDATE "data_mart" SET "dataLastUpdated" = …` revert. Object identity holds (`run-report.service.ts:232` `const { dataMart } = report`; `models/base-report-run.model.ts:46-48,62-65`). What `:460` closes is the *deterministic* self-clobber. Four terminal paths still reach `services/report-run.service.ts:114` → `services/report.service.ts:233-235` `repository.save(report)` carrying a run-start snapshot: **(a)** blended success — the guard is false, and your own test asserts it (`run-report.service.spec.ts:744-746`); **(b)** failure/cancel — `run-report.service.ts:397` → `:618` → `finish()`, `recordDataLastUpdated` never reached; **(c)** non-blended with `dataLastUpdatedAt === null`; **(d)** `saveActualizedSchema` (`services/data-mart.service.ts:381-385`) on *every* run, plus `stream-http-data.service.ts:265`. Fairness: these windows exist on `main` too — but `main` had one writer of this column (`use-cases/refresh-data-mart-data-last-updated.service.ts:141`) and this PR adds three, so the collision rate goes up ~4×. Patching more branches with in-memory assignments won't converge; the structural fix is to stop full-saving the DataMart from the Report path — either `cascade: ['insert']` on `report.entity.ts:51`, or make `saveReport` a targeted `repository.update(id, { lastRunStatus, lastRunError, lastRunAt })`, since `finish()` only ever changes Report's own scalars.
- `apps/backend/src/data-marts/services/data-mart.service.ts:296` — **no monotonicity guard, so the saved timestamp can move backwards — and this one is genuinely new in this PR.** `update({ id }, { dataLastUpdated: block })` is an unconditional overwrite, and the PR takes the writers from one to four (`run-report.service.ts:462`, `stream-http-data.service.ts:349`, `query-data-mart.service.ts:371`, `refresh-data-mart-data-last-updated.service.ts:141`), unordered and unlocked. The report path's measure→write gap is the wide one: the promise starts at `:255` and is written at `:462`, i.e. after the entire batch loop and destination writes — minutes to hours. Concretely: a run measures 09:00 (source = 08:00), a user hits *Check now* at 09:10 (source = 09:05), the run writes at 09:20 and reverts the column to 08:00, regressing `computedAt` too. Unlike `modifiedAt`, this column **is** rendered and sortable (`apps/web/.../DataMartTable/columns/columns.tsx:160-178`), so the list visibly re-sorts. Mitigating and worth noting: every writer already skips `dataLastUpdatedAt === null`, so a failed lookup can never erase a good value, and the error direction is the conservative one the schema already sanctions (`dto/schemas/source-data-last-updated.schema.ts:9-11`). Cheap fix: `computedAt` is already persisted (`:31`) and simply never compared — add a `WHERE json_extract(dataLastUpdated,'$.computedAt') < :computedAt` guard or a read-compare-write.
- `apps/web/src/features/data-marts/shared/components/DataMartBulkActions/DataMartBulkActions.tsx:196` — **moving the check into the dropdown removes the only progress affordance for an operation that can run for minutes.** Radix closes the menu on `onSelect` (`:199`, no `preventDefault()`), so `disabled={isCheckingDataLastUpdated}` and the "Checking Data Last Updated…" label only ever render inside a menu that just unmounted. The removed toolbar button was always visible with `animate-spin`. The sweep is genuinely long: chunks of 200 sent **sequentially** (`model-canvas/model/use-refresh-data-last-updated.ts:15,45-56`), request timeout 150 s (`shared/services/data-mart.service.ts:181`), canvas up to 1000 nodes, and `skipLoadingIndicator: true` (`:182`) suppresses the global indicator too. A user on a large canvas sees nothing at all for up to ~12 minutes. `event.preventDefault()` in `onSelect` to keep the menu open, or a `toast.loading` dismissed in the existing `finally`, would fix it.
- `apps/web/src/features/data-marts/model-canvas/model/use-refresh-data-last-updated.ts:65` — **the new success toast fires even when part of the sweep failed.** `anyRequestFailed` is set at `:54` but the toast at `:65-67` only checks `measured.size > 0`; the error toast at `:58-63` fires only when `measured.size === 0`. A 600-node canvas where 2 of 3 chunks fail shows a green "checked for 200 data marts" and no error, while 400 nodes silently keep stale values. There is also a fully silent path: `measured.size === 0` with no request throwing produces no toast at all — and the hook's own JSDoc at `:28` ("*a toast appears only when the sweep produced nothing at all*") now contradicts the code.

<details>

- `apps/backend/src/data-marts/use-cases/stream-http-data.service.ts:345` — the lookup plus an awaited DB write sit before the first byte, with no grace and no `AbortSignal`, while the MCP path bounds the identical wait to 2 s (`query-data-mart.service.ts:51,77-89,289`) and passes a signal. Verified honestly: this is *not* a new class of problem — `computeTotalsBestEffort` at `:340` already awaits a full BigQuery job with no timeout and no signal on `main`, and in the common case (any numeric column) the lookup has already settled behind it, so the realistic added TTFB is ~0. Non-BigQuery storages add exactly 0 ms (`data-storage-providers.ts:262` registers only the BigQuery resolver). Still worth reusing `withGrace` here for consistency — and the comment at `:303` ("cannot fail or **stall** the stream") is inaccurate: self-capped means capped at 15 s.
- `apps/backend/src/data-marts/services/source-data-last-updated.service.ts:26` — nothing is cached, deduplicated or TTL'd, so cost scales with run count rather than Data Mart count: 50 reports on one Data Mart produce 50 byte-identical measurements and 50 identical writes to the same row per cycle. For sharded/wildcard marts each one includes a real BigQuery query job against `__TABLES__` (`bigquery/adapters/bigquery-api.adapter.ts:301-314` — no `dryRun`, unlike `:217`), not just metadata calls. A short per-Data-Mart TTL on the persist would collapse this with no product change. (The docs' "Checking is free" claim is *not* affected — it links explicitly to OWOX consumption units and is scoped to the two manual checks; and that shard query already existed on `main` via the manual refresh.)
- `apps/backend/src/data-marts/use-cases/stream-http-data.service.ts:346` — design note 3 ("failed runs skip journal/persist") does not hold for the HTTP path: the journal and the persist both happen before `streamRows` at `:368`, so a stream that dies mid-write is recorded FAILED at `:410-419` having already written the Data Mart's saved value. Report runs behave as described; the three paths should agree.
- `apps/backend/src/data-marts/services/source-data-last-updated.service.ts:53` — `resolveForDefinition` takes `definition?: unknown` and casts `dataMart.definition as never` at `:74`, while the sibling doing the identical thing passes a properly typed value with no cast (`use-cases/refresh-data-mart-data-last-updated.service.ts:117-119`). `never` is assignable to anything, so a future signature change to `DataMartQueryBuilderFacade.buildQuery` would break the sibling at compile time and silently keep compiling here. `Pick<DataMart, 'id'|'storage'|'definitionType'|'definition'>` works — both callers already pass a real `DataMart`.
- `apps/backend/src/data-marts/use-cases/run-report.service.ts:255` / `stream-http-data.service.ts:304` — the promise is only awaited on the success path, so any earlier throw abandons it. No unhandled rejection (verified: `source-data-last-updated.service.ts:139-141`, `:178-192` catch every mode), but the work isn't cancelled either — and the HTTP path passes no signal at all, so a failed request keeps a dry run plus up to 50 `tables.get` calls in flight for up to 15 s.
- `apps/backend/src/data-marts/use-cases/run-report.service.ts:258` — `sql: blendingDecision.blendedSql ?? ''` has no guard, unlike the stream path which throws first (`stream-http-data.service.ts:298-300`); an empty string reaches `executeDryRunQuery('')` and is swallowed into `unavailable`.
- `apps/backend/src/data-marts/services/source-data-last-updated.service.ts:113-137` duplicates `use-cases/refresh-data-mart-data-last-updated.service.ts:113-137` (same `buildQuery` call, same string/object unwrap, same warn-and-degrade); folding the sibling onto the new method would also let it drop its own `DataMartQueryBuilderFacade` dependency.
- minor: the MCP path persists a measurement taken from the *composed* query (`query-data-mart.service.ts:235,369-371`) while the other two persist a *definition-scoped* one — I chased this hard as a possible wrong-value bug and it is **refuted**: both hand BigQuery an identical FROM clause (`bigquery/services/bigquery-query.builder.ts:140-142` vs `:161-163`), `_TABLE_SUFFIX` is never generated anywhere in the codebase, so no shard pruning is possible, and an unexpanded technical view degrades to `unavailable` and is blocked by the `!== null` gate — worst case a *missed* persist, never a wrong one. Still, mirroring the blended/non-blended branch and calling `resolveForDefinition` would make the invariant self-evident instead of load-bearing on a subtle argument; `dto/schemas/data-mart-run/data-mart-run-report-definition.schema.ts:24` journals the **full** block including `sources[]` and the Run History Configuration panel renders unknown keys via a blanket `JSON.stringify` with only two keys excluded (`apps/web/.../DataMartRunHistoryView/ConfigurationView.tsx:101-111`), so consider `toSourceDataLastUpdatedSummary` for the list payloads; on the same run record `reportDefinition.executionSqlQuery` (composed) sits next to `dataLastUpdated` (definition-scoped) with nothing signalling they describe different queries; `updateDataLastUpdated` matches on `{ id }` alone with no `projectId` — every caller is project-scoped today, so defence-in-depth only; `definitionType` is checked at `source-data-last-updated.service.ts:65` and then never used (builders discriminate structurally via the zod guards); `LEGACY_GOOGLE_BIGQUERY` has no resolver (`data-storage-providers.ts:262` + `bigquery-source-data-last-updated.resolver.ts:37`), so those marts never measure — pre-existing, but the new docs now promise "every run"; and `updateDataLastUpdated` bumps `@UpdateDateColumn modifiedAt` on every write (`typeorm/query-builder/UpdateQueryBuilder.js:403-406`) — I checked whether this corrupts a "Modified" column and it does **not** (no such column or sort exists in `DataMartTable`, and report runs already bumped it on `main` via `saveActualizedSchema`), but pinning `modifiedAt` would avoid adding to the existing churn on the MCP-only surfaces (`repositories/mcp-data-catalog-summary.repository.ts:82,86`).

</details>

**Checks run.** `eslint` clean on all 13 changed backend + web files. `tsc -b` in `apps/backend` reports 143 errors at this head vs **141 on `origin/main` with an identical setup** — the delta is 2 more instances of a pattern already present 4× on main (`dataMartRun.reportDefinition!.x` → `TS2339 … on type 'never'`, oldest blamed to `dad37e4076`, 2026-06-18). *Not attributable to this PR*, and an artifact of the local resolution environment. Changeset `.changeset/6750-data-last-updated-date.md` is correct (`'owox': minor`, task number in the filename, the two `6750-*` files properly folded into one, no other changeset touched). No new HTTP endpoints and no `*.facade.impl.ts` changed, so the `@Auth` and facade-unit-test gates don't apply.

<sub>Automated OWOX review · worktree multi-agent passes · all blockers adversarially verified. Four candidate blockers were refuted or downgraded by the skeptic gate and appear above as notes rather than blockers.</sub>

apps/backend/src/common/scheduler/services/fetchers/trigger-fetcher.service.ts
zapolsky: **Refinement to the above — trigger lifecycle, checked properly.** My phrasing "every trigger row created from now on stays forever" was too broad. The correct split, which actually sharpens the finding rather than softening it:

- **Run triggers are never deleted at all** — `ReportRunTrigger`, `ConnectorRunTrigger`, `DataQualityRunTrigger`, `SearchReindexTrigger` all extend `Trigger` directly (`data-marts/entities/report-run-trigger.entity.ts:7`, `connector-run-trigger.entity.ts:7`, `data-quality-run-trigger.entity.ts:8`, `entities/search/search-reindex-trigger.entity.ts:9`). After execution the runner only transitions status — `common/scheduler/services/runners/abstract-trigger-runner.service.ts:142`, `:197`, `:209` are all `repository.save(trigger)` with `SUCCESS`/`ERROR`. There is no `.delete(` or `.remove(` in any of those handlers or in `base-run-trigger-handler.service.ts`. **TTL was the only reaper for these tables**, so `report_run_trigger`, `connector_run_trigger`, `data_quality_run_trigger` and the search-reindex tables now grow unconditionally, one row per run, forever.

- **UI triggers do get deleted — but on response collection, not on completion.** `SqlDryRunTrigger`, `SchemaActualizeTrigger`, `AiHelperTrigger`, `PublishDraftsTrigger`, `InsightRunTrigger`, `InsightTemplateRunTrigger`, `InsightArtifactSqlPreviewTrigger`, `AiAssistantRunTrigger` extend `UiTrigger`, and `remove()` fires when the frontend fetches the result: `common/scheduler/shared/ui-trigger.service.ts:49,55,60` in `getTriggerResponse()` (SUCCESS / ERROR / CANCELLED) and `:79,93` in `abortTriggerRun()`. So on the happy path these do clean themselves up. They leak whenever the client never comes back for the response — closed tab, navigation away, dropped connection, backend restart between the POST and the GET. The 1-3 h TTLs on those handlers existed precisely as the safety net for that unhappy path, and that net is now gone.

- **Scheduled triggers were never in scope for TTL, and that is load-bearing.** The delete matches on `createdAt` alone with no status filter, so it would have wiped live user schedules — it doesn't, because `data-marts/services/scheduled-triggers-handler.service.ts` declares no `triggerTtlSeconds` (the method is optional on `common/scheduler/shared/trigger-handler.interface.ts:59`), leaving it `undefined` and short-circuiting on the original guard. Worth keeping in mind if the guard is ever restructured rather than restored verbatim.

Net effect on the finding: unchanged for the run-trigger tables (unbounded growth, no other reaper), and narrowed but still real for the UI-trigger tables (leak confined to abandoned requests).

zapolsky (changes requested):

apps/backend/src/common/scheduler/services/fetchers/trigger-fetcher.service.ts
laskevych: Thanks! It was for testing. I forgot return default file state.

laskevych:
Thanks for the depth here — especially the empirical verification of the cascade mechanism. Addressed in 1c5810b30 (the TTL revert landed earlier as 91d2809f4). Point by point:

## Primary notes — all four implemented

**Cascade clobber (structural fix).** Took your second suggestion: `ReportRunService.finish()` no longer full-saves the Report. `saveReport` is replaced by `ReportService.updateLastRunOutcome` — a targeted `repository.update(id, { lastRunStatus, lastRunError })` with an explicit NULL clear for the error (mirroring what `updateRunStatus` already does for the Looker path, which never had this problem). `finish()` only ever changes the Report's own scalars, so nothing is lost, and **all** terminal paths (blended success, failure/cancel, null-measurement, non-blended) stop carrying the run-start DataMart snapshot into a save. `cascade: ['insert']` was the runner-up; I preferred not to change relation semantics that other save-sites may rely on. The in-memory refresh at `recordDataLastUpdated` stays as defence-in-depth for any future full save. The `saveActualizedSchema` window you flagged remains — it's cross-process (schema actualization happens at run start, before any measurement exists to clobber deterministically) and pre-exists this PR; with the monotonicity guard below its worst case is now a skipped write, not a regression.

**Monotonicity guard.** `updateDataLastUpdated` now reads the saved snapshot and drops any measurement whose `computedAt` is not newer — the saved value can never move backwards, which closes your 09:00-measure/09:20-write scenario. I went with read-compare-write rather than a `json_extract` predicate: it has to hold on both better-sqlite3 and MySQL (`JSON_EXTRACT` returns a quoted JSON string there, so the comparison needs `JSON_UNQUOTE` and dialect-forked SQL), and the residual race window is milliseconds against the minutes-wide gap this closes. Also scoped the write by `projectId` while in there (your defence-in-depth note). Unit-tested: writes on empty/newer, drops older, skips foreign project.

**Dropdown progress affordance.** Went with the `toast.loading` option: it survives navigation away from the canvas and doesn't pin the menu open under the user's pointer for minutes. Shown for the whole sweep, dismissed in the existing `finally`.

**Partial-failure toast.** Success toast now fires only when every chunk succeeded; a partially-failed sweep gets an error toast with `checked N of M` counts; and the previously-silent path (nothing measured, nothing thrown) now reports too. The hook's JSDoc is updated to match.

## Additional observations

Implemented the cheap ones:

- `resolveForDefinition` input is now `Pick<DataMart, 'id' | 'storage' | 'definitionType' | 'definition'>` — the `as never` cast is gone.
- The report path no longer sends `''` to the warehouse when a blending decision carries no blended SQL — it resolves `unavailable` directly (the read path below still surfaces the actual error, matching the stream's throw).
- The "cannot fail or stall" comments now say what's true: never rejects, self-caps at the soft timeout.

Declined, with reasoning:

- **`withGrace` on the HTTP path** — by your own measurement the realistic added TTFB is ~0 because `computeTotalsBestEffort` already awaits an unbounded BigQuery job in front of it. A grace wrapper here would bound the wrong wait. If totals ever gets a deadline, both should get one together.
- **Persist-before-stream on the HTTP path** — kept, and the PR's design note is corrected instead. The measurement describes warehouse state, not delivery: it is exactly as valid when the stream later dies as when it completes (the manual Check now delivers nothing at all and persists the same scope). The report path differs only because its await naturally sits after the write loop.
- **Caching/TTL on measurements** — deliberately uncached (the service header explains why: a stale answer to "how fresh?" is worse than a slow one). The new `computedAt` guard already collapses redundant *writes*; the redundant *reads* ride behind runs that are paying for a warehouse round-trip anyway. Worth revisiting if `__TABLES__` job volume ever shows up in metrics.
- **Folding the manual-refresh sibling onto `resolveForDefinition`** — agreed it's the right cleanup, but it rewires the already-shipped #1472 path, so I'd rather do it as a small follow-up than widen this PR.
- **MCP composed-vs-definition scope** — your refutation matches my reading (identical FROM, no `_TABLE_SUFFIX` generation, technical views degrade to a *missed* persist, never a wrong one). Switching to `resolveForDefinition` would add a second query build per MCP call purely for symmetry; the journalled `sources[]` already makes what was measured explicit.
- **`LEGACY_GOOGLE_BIGQUERY`** — pre-existing gap, consistent with the documented "other storages report unavailable" caveat; belongs with the per-storage rollout rather than this PR.
- **Pinning `modifiedAt`** — per your own check nothing user-facing reads it and report runs already bumped it on main; not worth a nonstandard update.

apps/backend/src/common/scheduler/services/fetchers/trigger-fetcher.service.ts
laskevych: Restored to the origin/main state in 91d2809f4 — it was a local debugging edit that got swept into the feature commit. `trigger-fetcher.service.spec.ts` back to 7/7 green.

apps/web/src/features/data-marts/model-canvas/model/use-refresh-data-last-updated.ts
```diff
@@ -62,6 +65,13 @@ export function useRefreshDataLastUpdated(storageId: string | null) {
           return;
         }
 
+        if (anyRequestFailed) {
```
prokopetsv, line 74: 🟡 **Non-Blocker** — the "nothing measured, nothing failed" path (line 61, `measured.size === 0 && !anyRequestFailed`) is silent again, and it's the *guaranteed* outcome on unsupported storages.

The review response said this path "now reports too" (implemented in `1c5810b30`), but `787eb94b5` removed the success/loading toasts and the current code returns with no toast. That combination isn't an edge case: the refresh endpoint only resolves BigQuery, unmeasurable marts are simply absent from the response, and the canvas can show any storage type — so on a non-BigQuery canvas, *Check Data Last Updated* spins every icon, then stops with zero feedback and zero change, every time. The feature looks broken rather than unsupported.

Cheap fix: a toast for this branch ("couldn't measure any of the selected Data Marts"), or don't offer the menu item for storages without a resolver.

apps/backend/src/data-marts/use-cases/run-report.service.ts
```diff
@@ -413,6 +442,45 @@ export class RunReportService {
...
+  ): Promise<void> {
+    const dataLastUpdated = await dataLastUpdatedPromise;
+    if (dataMartRun?.reportDefinition) {
+      dataMartRun.reportDefinition.dataLastUpdated = dataLastUpdated;
+    }
+    if (!needsBlending && dataLastUpdated.dataLastUpdatedAt !== null) {
+      // Also refresh the in-memory entity: `dataMart` is the same instance as
+      // `report.dataMart` (cascade: true), and Report is fully re-saved when the run
+      // finishes — a stale in-memory snapshot would cascade over the value written below.
+      dataMart.dataLastUpdated = dataLastUpdated;
```
prokopetsv: 🟡 **Non-Blocker** — this in-memory "defence" can hold a value the DB guard rejected.

The assignment runs *before* `updateDataLastUpdated` applies its `computedAt` guard. If the guard drops the write (a manual check or MCP query saved a newer measurement mid-run — exactly the race the guard exists for), the in-memory entity now carries the *older* block. The assignment's stated purpose is defending against a future full cascade save — but a cascade save bypasses the guard entirely, so in that one future it defends against, this defence would itself regress the newer DB value.

No live bug today (the structural fix removed all full saves on the terminal paths), but the defence is inverted for its only purpose. Cheap fix: assign only after `updateDataLastUpdated` confirms it wrote (e.g. have it return a boolean), or drop the assignment now that the structural fix landed.

docs/getting-started/setup-guide/data-last-updated.md
```diff
@@ -20,30 +20,36 @@ Hover over any Data Last Updated value or icon to see the details: the exact tim
...
-| On a schedule | Never. OWOX does not re-measure Data Marts in the background |
+| *Actions → Check Data Last Updated* on the canvas | Measures and **saves** the value for all visible Data Marts |
+| Report run — manual or scheduled | Measures during the run and records the result in the run's history entry. When the report reads only the Data Mart's own sources, it also **saves** the value |
+| HTTP Data request | Same as a report run: records in history, **saves** when the request reads only the Data Mart's own sources |
+| MCP `query_data_mart` call | Measures live for that response and records it in the run's history entry. **Saves** the value when the query reads only the Data Mart's own sources (no joined fields) |
+| In the background, on its own | Never. OWOX only measures when someone checks manually or a run delivers data — a scheduled report run counts as a run |
+
+A run that spans several joined Data Marts records the measurement in its history entry only. It does not overwrite any single Data Mart's saved value, because a blended reading covers other Data Marts' tables too and would overstate the one it is saved on.
```
prokopetsv: 🟡 **Non-Blocker** — the second sentence breaks the repo's codified docs style rules (`.claude/CLAUDE.md`): it exceeds the 20-word cap and ends passive ("it is saved on"). The `mcp.md` addition has the same length problem. Flagged only because the rules are codified, not as preference.

Suggested rewrite: "A blended run records the measurement in its history entry only. Its reading covers other Data Marts' tables too, so saving it on one Data Mart would overstate that value."

prokopetsv (commented):
Three 🟡 non-blockers from a follow-up pass over the post-review commits (`1c5810b30`, `787eb94b5`) — inline below. Verified at head that all items from the earlier review round are genuinely addressed: TTL cleanup fully restored, `updateLastRunOutcome` correctly omits `lastRunAt`/`runsCount` (persisted at run start via `updateReportWithVersionControl`), monotonicity guard tested, no stranded `saveReport` callers. No blockers.

apps/web/src/features/data-marts/model-canvas/model/use-refresh-data-last-updated.ts
laskevych: Agreed — and thanks for catching that the two commits cancelled each other out: `1c5810b30` added the toast, `787eb94b5` swept it away with the success/loading ones while reworking the UX. You're right that this branch is the *guaranteed* path on non-BigQuery storages, which makes it the worst one to be silent on.

Fixed in e37638a1a: the `measured.size === 0` branch always reports now — "Couldn't measure any of the selected Data Marts — their storage may not support Data Last Updated yet" when nothing failed, the existing failure toast otherwise. I kept the menu item visible rather than hiding it per storage: the canvas is single-storage but the hook would need resolver knowledge the frontend doesn't have today, and an honest message beats a mysteriously missing item.

apps/backend/src/data-marts/use-cases/run-report.service.ts
laskevych: Sharp catch — the defence really was inverted for the one scenario it exists for: assign-before-guard means a rejected (older) block sits on the entity, and a future cascade save would push it over the newer DB value while bypassing the guard entirely.

Fixed in e37638a1a with your first option: `updateDataLastUpdated` now returns whether it wrote, and the run mirrors the block onto the in-memory entity only after a confirmed write. Kept the mirror rather than dropping it — the structural fix removed today's full saves, but the mirror is exactly the belt-and-braces for a future one, and now it carries the right value. Covered by a new spec case (guard returns false → in-memory untouched) plus `resolves.toBe(true/false)` assertions on the guard tests.

docs/getting-started/setup-guide/data-last-updated.md
laskevych: Took your rewrite verbatim in e37638a1a — it says the same thing in half the words. The `mcp.md` sentence is split in two as well ("A non-blended query (no joined fields) also saves the measurement as the Data Mart's last-known value. The same value appears across the OWOX UI — see Data Last Updated."). markdownlint clean on both files.

prokopetsv (approved):
Approved. The feature persists Data Last Updated safely on every data-delivering path: the report-run cascade clobber is structurally fixed, the persist is monotonic and project-scoped, and blended runs correctly journal without persisting. All review rounds (automated + follow-up) were addressed with verified fixes and regression tests; backend and web suites green. Residual notes (fail-closed balance-API-style races measured in milliseconds, BigQuery-only resolver coverage) are acknowledged, documented trade-offs — nothing blocking.

apps/backend/src/common/scheduler/services/fetchers/trigger-fetcher.service.ts
zapolsky: ✅ **Confirmed fixed at `e37638a1a`.** `git diff origin/main e37638a1a -- apps/backend/src/common/scheduler/services/fetchers/trigger-fetcher.service.ts` is empty — the file is byte-identical to `origin/main`, so the TTL body, the config guard and the affected-rows log are all back, and `triggerTtlSeconds` is a live dependency again. `trigger-fetcher.service.spec.ts` 7/7 green. Thanks for the quick turnaround, and for saying plainly that it was a debug leftover — that is exactly the kind of thing worth catching before it ships.

Closing this thread from my side.

zapolsky:
✅ **No blocking issues found at `e37638a1a`** — looks ready to approve (a maintainer should give the formal approval). Incremental pass over the four commits since `af2e087b5`.

The blocker is confirmed closed in its own thread. Everything below is non-blocking, and the highest-ranked item is a 70.

## Confirmed fixed — verified, not taken on trust

- **Cascade clobber — fixed structurally.** `saveReport` is gone from the entire backend (`grep -rn "saveReport" src/` returns only the unrelated `saveReportRunResultSafely`), and `finish()` now takes the targeted `updateLastRunOutcome` (`services/report-run.service.ts:115` → `services/report.service.ts:239-244`). I went looking for the classic failure mode of this refactor — a column that the old graph save persisted and the targeted update silently drops — and **found none**: all 20 `Report` columns enumerated, and the two candidates I most suspected, `lastRunAt` and `runsCount`, are mutated only in `initializeStarting` and already durable via `createPending` → `updateReportWithVersionControl` (`report.service.ts:255-278`) before the worker reloads a fresh Report. `modifiedAt` and `version` are appended by `update()` through the same code path `save()` used (`typeorm/query-builder/UpdateQueryBuilder.js:397-406`), so no `@UpdateDateColumn` divergence. The `() => 'NULL'` clear is correct in both transition directions.
- **Monotonicity guard — real and driver-verified.** I checked the `projectId` scoping specifically because a silently no-op'ing write would have been worse than the bug it fixes: all four callers pass the Data Mart's own `projectId` (`run-report.service.ts:465`, `stream-http-data.service.ts:350`, `query-data-mart.service.ts:371`, `refresh-data-mart-data-last-updated.service.ts:145`), and scheduled runs carry project context via the assertion at `run-report.service.ts:194-200`. **No silent no-op.** The guard was also exercised against a real better-sqlite3 DataSource, not just the mocks: older → `false`, DB unchanged; newer → `true`, DB advances.
- **In-memory mirror — correctly guard-aware.** The mirror is now strictly inside `if (written)` (`run-report.service.ts:474`), and skipped on the `catch`.
- Also closed: the `as never` cast (now `Pick<DataMart, …>`), the empty-string SQL reaching the dry run, and the inaccurate "cannot stall" comments.

Mutation-tested rather than read: breaking `report.service.ts:242`, `run-report.service.ts:474` and `data-mart.service.ts:317` each turned a specific test red.

**The declines are all reasonable and I'm not re-litigating them** — the `withGrace` argument in particular is right, and it is the correct call to bound both waits together or neither.

## One correction to the response

> "The `saveActualizedSchema` window remains — with the monotonicity guard below its worst case is now a skipped write, not a regression."

That isn't right, and it's worth knowing because it changes what the guard buys you. `updateDataLastUpdated` is a **service-method** guard, not a column invariant, so it only protects writes that go through it. `saveActualizedSchema` is a full `dataMartRepository.save(dataMart)` (`services/data-mart.service.ts:405-406`) and bypasses it entirely — TypeORM diffs the stale in-memory `dataLastUpdated` against the current row and writes the stale value back. That is a **revert, not a skip**. Reproduced on a real better-sqlite3 DataSource: after a full save of a stale entity the saved `computedAt` went from `12:00` back to `10:00`.

This is the same mechanism you documented at `report.service.ts:229-236` as the reason to replace `saveReport` — the fix just landed on one of the sites. Still live on the run path (`run-report.service.ts:540-543`), the stream path (`stream-http-data.service.ts:265-268`), `update-report.service.ts:179` and `create-report.service.ts:133` (both cascade through `Report.dataMart`), plus roughly nine `save(dataMart)` use-cases.

To be fair to the change: the windows are now seconds-to-a-request rather than minutes-to-hours, and a run can no longer revert **its own** measurement — which was the 100%-reproducible original bug and is genuinely dead. So this is "the claim is broader than the fix", not "the fix didn't work". Worth either narrowing the claim or filing the follow-up.

<details>

Secondary (50–79), highest first:

- `apps/web/.../model-canvas/components/ModelCanvas.tsx:347` — **(70)** dagre lays out nodes 30 px taller than they render in title-only mode. `computeNodeHeight` takes four parameters (`model/erd-node.ts:52-56`); the real node passes four at `:195`, the dagre input passes three at `:347`, so `statusRowHidden` is never applied there and `CARD_STATUS_ROW_HEIGHT` is never subtracted. Visible as phantom vertical gaps in exactly the mode `787eb94b5` added. `erd-node.test.ts` tests the pure function directly, so it cannot catch a miswired call site.
- `apps/web/.../shared/canvas/definition-type-accent.ts:18` — **(60)** the VIEW accent moved `OWOX_BLUE_BASE` → `OWOX_YELLOW_BASE`, and `shared/canvas/erd-card.tsx:43-45` renders that pill as `text-white` over the accent. Computed: white on `#4286DE` = 3.69:1, white on `#F5C344` = **1.64:1**, against WCAG AA's 4.5:1 for small text. These are hardcoded hex from `owox-palette.ts`, not CSS variables, so the theme toggle doesn't help either mode. Notable because `owox-palette.ts:26` already carries a comment justifying a colour choice on contrast grounds.
- `apps/backend/.../services/data-mart.service.ts:317` — **(55)** the guard-bypass correction above, in one line: it protects the four service callers, not the column.
- `apps/backend/.../services/data-mart.service.ts:316-319` — **(50)** the guard compares `computedAt` only, so a `coverage: 'partial'` reading can shadow a `complete` one. `bigquery-source-data-last-updated.resolver.ts:147-148` still returns a non-null MAX over just the subset that resolved when metadata lookups fail or the 50-table cap truncates, and every caller persists on `dataLastUpdatedAt !== null` regardless of coverage — so a newer `computedAt` carrying an older `dataLastUpdatedAt` passes the guard and the sortable column moves backwards anyway. (`unavailable` is safe: always null, gated out.) Comparing `dataLastUpdatedAt`, or refusing to let `partial` overwrite `complete`, would close it.
- `apps/web/.../shared/canvas/definition-type-accent.ts:18` — **(50)** `OWOX_YELLOW_BASE` now carries four meanings that can co-occur: the draft status dot (`erd-card.tsx:23`), the primary-key icon (`ModelCanvasFlowNode.tsx:61`), `ATTENTION_COLOR` in the relationships canvas (`RelationshipCanvas.tsx:120`), and now the VIEW type accent. A draft VIEW renders a yellow stripe next to a yellow dot in the same header row (`ModelCanvasFlowNode.tsx:144-157`), and `RelationshipCanvas.tsx` calls `definitionTypeAccent` at `:215`/`:1067` in the file that defines `ATTENTION_COLOR` — type colour and status colour became indistinguishable there.

Minor (0–49):

- `ModelCanvasFlowNode.tsx:203` — **(45)** wrapping the whole status row in `{!titleOnly && …}` also removes `DataQualityCanvasStatusIcon`, the only per-node entry point to Open / Run Data Quality (a different, already-shipped feature). The settings UI documents three parts (`ModelCanvas.tsx:101-118`) and the button says "Uncheck all — title only", so a user unticking three documented things loses an undocumented fourth — and the preference persists (`OBJECT_LABELS_LS_KEY`).
- `ModelCanvasFlowNode.tsx:203` — **(45)** same line removes the new sweep spinner, so title-only users get zero progress affordance for a multi-minute sweep. The JSDoc at `use-refresh-data-last-updated.ts:28-30` states unconditionally that every icon spins.
- `use-refresh-data-last-updated.ts:63` — **(45)** the restored "nothing measured" toast is gated on `measured.size`, which counts rows returned, not timestamps learned. The backend returns `unavailable` blocks as real map entries (`refresh-data-mart-data-last-updated.service.ts:100-101` → `data-mart.mapper.ts:202-204`), so an all-`unavailable` sweep has `measured.size > 0` and stays silent — the icons spin and settle back unchanged. `[...measured.values()].some(b => b.dataLastUpdatedAt !== null)` would gate it correctly. (This is the residual of the path @prokopetsv flagged; the no-resolver variant genuinely is fixed.)
- **(40)** "mirrors Check Quality UX" is superficial: Check Quality spins per node from its own state and its bulk path goes through a confirmation dialog; here one canvas-wide boolean spins every node (`ModelCanvas.tsx:548-556`) regardless of which 200-id chunk is in flight, and the icon is a non-interactive tooltip trigger.
- `use-refresh-data-last-updated.ts:41-96` — **(40)** switching storage mid-sweep leaves `isRefreshing` true (the hook isn't remounted, `ModelCanvasView.tsx:68`) so the new canvas spins for a sweep not measuring it, and results land under the old storage's cache key.
- **(35)** two mutations survived: `>=` → `>` on the equality boundary, and dropping `projectId` from the guard's READ `where` — `data-mart.service.spec.ts:373-382` mocks `findOne` → `null`, so it asserts the mock rather than the query.
- `ModelCanvasFlowNode.test.tsx:180` — **(35)** asserts the absence of `Data Quality checks for Orders`, which is the PopoverContent aria-label (`DataQualityCanvasStatusIcon.tsx:267`) and never mounts in that test, so it passes either way. The sibling assertion on `:181` is genuine.
- **(30)** the Actions item is offered unconditionally (`ModelCanvasView.tsx:283`) while only BigQuery has a resolver, so on five storage types the only reachable outcome is spin-then-red-toast.
- **(25)** the changeset covers the menu rename but not the VIEW colour change or the title-only behaviour change, both user-visible.
- **(25)** *Check now* caches the measured value client-side (`use-refresh-data-last-updated.ts:81-91`) while `refresh-data-mart-data-last-updated.service.ts:100-106` discards the new boolean, so a guard-dropped write leaves the canvas showing a value the DB doesn't hold until reload.
- **(20)** copy: `use-refresh-data-last-updated.ts:69` says "selected" where the canvas action targets the filter-visible set; and "storage may not support this yet" ships as `toast.error` for a non-error outcome.
- **(15)** `updateDataLastUpdated` returns `true` without consulting `result.affected`, so a row deleted between the read and the update reports a write that didn't happen. Cosmetic today — no caller can reach it — but the `@returns` contract now has a consumer.

</details>

**Checks run.** Backend: `eslint` clean on all changed files; 159 tests green across the five touched suites plus 51 across the service suites. Web: 84 tests passed, but `ModelCanvasView.test.tsx` and `use-model-canvas.test.tsx` fail to load and all 15 web `eslint` errors cascade from a single missing `react-router-dom` in `node_modules` (declared at `apps/web/package.json:58`) — **environmental, not attributable to this PR**; a clean install is needed for a trustworthy web lint/test signal. Worth noting that `ModelCanvasView.test.tsx` is the file that would cover the `isCheckingDataLastUpdated` wiring, and it stubs the hook anyway. Canvas mock trap applies to the two new `ModelCanvas.test.tsx` cases: they assert on props handed to a mocked `ReactFlow`, so `ModelCanvasFlowNode` never renders and no spinner is ever asserted — which is precisely the seam the title-only gap falls into.

<sub>Automated OWOX review · incremental pass over `af2e087b5..e37638a1a` · worktree multi-agent verification, claimed fixes checked against code and mutation-tested rather than taken on trust.</sub>

zapolsky (approved):

amarchenk0 (approved):

PR #1485, opened 2026-07-31
fix(data-marts): authorize the copy source and support copying from several Data Marts

laskevych:
Stacked on #1480 — the base is that branch, so this diff shows only the two commits below. GitHub retargets it to `main` once #1480 merges.

Both commits are about the same thing: how **Copy from…** resolves the Data Mart a configuration comes from. They rewrite the same block of `update-data-mart-definition.service.ts`, which is why they travel together.

## 1. Copying requires permission on the source Data Mart

Copying a configuration carries the source's stored credentials into the target, but only the Data Mart being edited was authorized. The source named by `sourceDataMartId` was fetched with `getByIdAndProjectId` — by project alone.

Per-Data-Mart access is a real boundary here: `apply-data-mart-visibility-filter` hides Data Marts that are neither owned nor shared, and the matrix denies a non-owner with no sharing every action. So an editor could name a Data Mart invisible to them as the copy source and have its secret values merged into a configuration their own connector then runs with. The ids needed are discoverable from `GET /data-marts/by-connector/:connectorName`, which is viewer-level and applies no visibility filter.

The copy now requires `COPY_CREDENTIALS` on the source, matching what `create-data-destination` and `update-data-storage` already do for their own copy flows. The action was missing from `EntityType.DATA_MART` entirely, so the matrix never modelled this case. It follows `EDIT`: granted to the technical owner, and to a non-owner when the Data Mart is shared for maintenance. **Sharing for reporting does not grant it** — that shares the data, not the credentials behind it.

Rationale for tying it to `EDIT`: whoever may edit a Data Mart's connector configuration can already run it with those credentials, so copying them grants no further capability.

## 2. Each copied configuration resolves against its own source

Copying one configuration from Data Mart A and another from Data Mart B in a single save failed the whole update with a 500 and lost both edits. The client derived a single `sourceDataMartId` from the first copied item, and the server then looked for every `_copiedFrom.configId` in that one Data Mart — the second item was never found and the merge threw a bare `Error`, which the global filter reports as an internal error.

Each copied item already records its origin in `_copiedFrom.dataMartId` and the web client already sends it; the server simply never read it. The merge now resolves each item against its own source, so one save may draw on as many sources as it likes. **No frontend change was needed.**

`sourceDataMartId` stays in the request as the source for items that name none, which is what clients predating the per-item metadata send.

Two things fall out of this:

- **Every source is authorized before any is read**, so a save that is only partly permitted touches nothing, rather than failing on whichever source happened to be loaded first.
- **Failures are domain errors now.** Copying between different connector types, naming a source that is not part of the request, and copying a configuration that no longer exists in its source all raise `BusinessViolationException` → 400 with a message the user can act on.

Also removed `sourceConfigurationId`, which was declared on the API DTO, the command and the mapper but never read by any use case nor sent by the web client.

## Tests

`access-decision-e2e.spec.ts`, `update-data-mart-definition.service.spec.ts`, `connector-secret.service.spec.ts`, `copy-report-as-data-mart.service.spec.ts`.

New: `COPY_CREDENTIALS` denied on a not-shared Data Mart, denied when shared for reporting only, allowed when shared for maintenance, allowed for the technical owner, denied to a business user throughout; the use case refusing a copy from a Data Mart the user cannot copy from, and touching nothing when it does; copying from two Data Marts in one save, with both authorized and both definitions reaching the merge; the whole save refused when one of several sources is not permitted; and a copy whose source was not resolved.

Backend and web typecheck and lint are clean on the changed files; CI runs the full suites.

### DIFF
--- a/apps/backend/src/data-marts/services/connector/connector-secret.service.spec.ts
+++ b/apps/backend/src/data-marts/services/connector/connector-secret.service.spec.ts
@@ -101,6 +101,49 @@ describe('ConnectorSecretService', () => {
   });
 
   describe('mergeDefinitionSecrets', () => {
+    // The merge only ever dereferences a pointer this DataMart already has
+    // stored. That is what lets a Data Mart still carrying a pointer into
+    // another one's record recover its values (extractAndSaveSecrets then forks
+    // them onto a record of its own), while a pointer supplied by the caller
+    // buys nothing.
+    it('never loads secrets for a _secrets_id supplied by the caller', async () => {
+      const { service, credentialsService } = createService(['AccessToken']);
+
+      const previous = makeDefinition([{ _id: 'a', _secrets_id: 'ours', AccountIDs: '33' }]);
+      const incoming = makeDefinition([
+        // Unknown _id, plus a pointer at another DataMart's record.
+        { _id: 'invented', _secrets_id: 'belongs-to-another-datamart', AccessToken: SECRET_MASK },
+      ]);
+
+      const merged = await service.mergeDefinitionSecrets(incoming, previous);
+      const cfg = merged.connector.source.configuration as Array<Record<string, unknown>>;
+
+      expect(credentialsService.getCredentialsById).not.toHaveBeenCalled();
+      // Nothing was resolved, so the mask stands - and extractAndSaveSecrets
+      // drops it rather than storing it as a credential.
+      expect(cfg[0].AccessToken).toBe(SECRET_MASK);
+    });
+
+    it('replaces a caller-supplied _secrets_id with the stored one', async () => {
+      const { service, credentialsService } = createService(['AccessToken']);
+      (credentialsService.getCredentialsById as jest.Mock).mockResolvedValue({
+        id: 'ours',
+        credentials: { AccessToken: 'our-token' },
+      });
+
+      const previous = makeDefinition([{ _id: 'a', _secrets_id: 'ours', AccountIDs: '33' }]);
+      const incoming = makeDefinition([
+        { _id: 'a', _secrets_id: 'belongs-to-another-datamart', AccessToken: SECRET_MASK },
+      ]);
+
+      const merged = await service.mergeDefinitionSecrets(incoming, previous);
+      const cfg = merged.connector.source.configuration as Array<Record<string, unknown>>;
+
+      expect(credentialsService.getCredentialsById).toHaveBeenCalledWith('ours');
+      expect(cfg[0]._secrets_id).toBe('ours');
+      expect(cfg[0].AccessToken).toBe('our-token');
+    });
+
     it('keeps previous secret when incoming has SECRET_MASK', async () => {
       const { service } = createService(['AccessToken']);
       const previous = makeDefinition([
--- a/apps/backend/src/data-marts/services/connector/connector-secret.service.spec.ts
+++ b/apps/backend/src/data-marts/services/connector/connector-secret.service.spec.ts
@@ -1,3 +1,4 @@
+import { BusinessViolationException } from '../../../common/exceptions/business-violation.exception';
 import type { ConnectorDefinition } from '../../dto/schemas/data-mart-table-definitions/connector-definition.schema';
 import { ConnectorService } from './connector.service';
 import { ConnectorSecretService, SECRET_MASK } from './connector-secret.service';
@@ -58,6 +59,12 @@ describe('ConnectorSecretService', () => {
       },
     } as unknown as ConnectorDefinition;
   };
+
+  // Copied items name their own source Data Mart; these cover the items that
+  // do not, which fall back to the source named by the request itself.
+  const SOURCE_DATA_MART_ID = 'source-datamart';
+  const fromSingleSource = (definition: ConnectorDefinition) =>
+    new Map([[SOURCE_DATA_MART_ID, definition]]);
 
   describe('mask', () => {
     it('masks secret fields using SECRET_MASK', async () => {
@@ -662,7 +669,7 @@ describe('ConnectorSecretService', () => {
         },
       ]);
 
-      const merged = await service.mergeDefinitionSecretsFromSource(incoming, sourceDefinition);
+      const merged = await service.mergeDefinitionSecretsFromSource(incoming, fromSingleSource(sourceDefinition), SOURCE_DATA_MART_ID);
       const cfg = merged.connector.source.configuration as Array<Record<string, unknown>>;
 
       expect(cfg[0].AccessToken).toBe('access1');
@@ -707,7 +714,7 @@ describe('ConnectorSecretService', () => {
         },
       ]);
 
-      const merged = await service.mergeDefinitionSecretsFromSource(incoming, sourceDefinition);
+      const merged = await service.mergeDefinitionSecretsFromSource(incoming, fromSingleSource(sourceDefinition), SOURCE_DATA_MART_ID);
       const cfg = merged.connector.source.configuration as Array<Record<string, unknown>>;
       const authType = cfg[0].AuthType as Record<string, Record<string, unknown>>;
 
@@ -746,7 +753,7 @@ describe('ConnectorSecretService', () => {
         },
       ]);
 
-      const merged = await service.mergeDefinitionSecretsFromSource(incoming, sourceDefinition);
+      const merged = await service.mergeDefinitionSecretsFromSource(incoming, fromSingleSource(sourceDefinition), SOURCE_DATA_MART_ID);
       const cfg = merged.connector.source.configuration as Array<Record<string, unknown>>;
       const authType = cfg[0].AuthType as Record<string, Record<string, unknown>>;
 
@@ -787,7 +794,7 @@ describe('ConnectorSecretService', () => {
         },
       ]);
 
-      const merged = await service.mergeDefinitionSecretsFromSource(incoming, sourceDefinition);
+      const merged = await service.mergeDefinitionSecretsFromSource(incoming, fromSingleSource(sourceDefinition), SOURCE_DATA_MART_ID);
       const cfg = merged.connector.source.configuration as Array<Record<string, unknown>>;
 
       expect(cfg[0]).not.toHaveProperty('generated_refresh_token');
@@ -816,8 +823,8 @@ describe('ConnectorSecretService', () => {
       } as unknown as ConnectorDefinition;
 
       await expect(
-        service.mergeDefinitionSecretsFromSource(incoming, sourceDefinition)
-      ).rejects.toThrow('Cannot copy secrets from different connector type');
+        service.mergeDefinitionSecretsFromSource(incoming, fromSingleSource(sourceDefinition), SOURCE_DATA_MART_ID)
+      ).rejects.toThrow('Cannot copy a configuration between different connector types');
     });
 
     it('returns configuration as is when _copiedFrom.configId metadata is missing (existing config)', async () => {
@@ -833,7 +840,7 @@ describe('ConnectorSecretService', () => {
         },
       ]);
 
-      const merged = await service.mergeDefinitionSecretsFromSource(incoming, sourceDefinition);
+      const merged = await service.mergeDefinitionSecretsFromSource(incoming, fromSingleSource(sourceDefinition), SOURCE_DATA_MART_ID);
       const cfg = merged.connector.source.configuration as Array<Record<string, unknown>>;
 
       // Should return the item unchanged (will be merged with previous in the next step)
@@ -855,8 +862,61 @@ describe('ConnectorSecretService', () => {
       ]);
 
       await expect(
-        service.mergeDefinitionSecretsFromSource(incoming, sourceDefinition)
-      ).rejects.toThrow('Source configuration with _id "non-existent-id" not found');
+        service.mergeDefinitionSecretsFromSource(incoming, fromSingleSource(sourceDefinition), SOURCE_DATA_MART_ID)
+      ).rejects.toThrow(BusinessViolationException);
+    });
+
+    it('copies each item from the Data Mart its own metadata names', async () => {
+      const { service, credentialsService } = createService(['AccessToken']);
+      (credentialsService.getCredentialsByIds as jest.Mock).mockResolvedValue(
+        new Map([
+          ['secrets-a', { id: 'secrets-a', credentials: { AccessToken: 'token-a' } }],
+          ['secrets-b', { id: 'secrets-b', credentials: { AccessToken: 'token-b' } }],
+        ])
+      );
+
+      const sourceA = makeDefinition([{ _id: 'a-1', _secrets_id: 'secrets-a', AccountIDs: '1' }]);
+      const sourceB = makeDefinition([{ _id: 'b-1', _secrets_id: 'secrets-b', AccountIDs: '2' }]);
+
+      const incoming = makeDefinition([
+        {
+          AccessToken: SECRET_MASK,
+          AccountIDs: '1',
+          _copiedFrom: { dataMartId: 'dm-a', configId: 'a-1' },
+        },
+        {
+          AccessToken: SECRET_MASK,
+          AccountIDs: '2',
+          _copiedFrom: { dataMartId: 'dm-b', configId: 'b-1' },
+        },
+      ]);
+
+      const merged = await service.mergeDefinitionSecretsFromSource(
+        incoming,
+        new Map([
+          ['dm-a', sourceA],
+          ['dm-b', sourceB],
+        ])
+      );
+      const cfg = merged.connector.source.configuration as Array<Record<string, unknown>>;
+
+      expect(cfg[0].AccessToken).toBe('token-a');
+      expect(cfg[1].AccessToken).toBe('token-b');
+      expect(cfg[0]).not.toHaveProperty('_copiedFrom');
+      expect(cfg[1]).not.toHaveProperty('_copiedFrom');
+      expect(cfg[0]._id).not.toBe(cfg[1]._id);
+    });
+
+    it('rejects a copy whose source Data Mart was not resolved', async () => {
+      const { service } = createService(['AccessToken']);
+
+      const incoming = makeDefinition([
+        { AccessToken: SECRET_MASK, _copiedFrom: { dataMartId: 'dm-unknown', configId: 'a-1' } },
+      ]);
+
+      await expect(
+        service.mergeDefinitionSecretsFromSource(incoming, new Map())
+      ).rejects.toThrow(BusinessViolationException);
     });
 
     it('generates new _id for each copied configuration', async () => {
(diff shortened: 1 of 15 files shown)


apps/backend/src/data-marts/services/connector/connector-secret.service.spec.ts
```diff
@@ -101,6 +108,49 @@ describe('ConnectorSecretService', () => {
   });
 
   describe('mergeDefinitionSecrets', () => {
+    // The merge only ever dereferences a pointer this DataMart already has
+    // stored. That is what lets a Data Mart still carrying a pointer into
+    // another one's record recover its values (extractAndSaveSecrets then forks
+    // them onto a record of its own), while a pointer supplied by the caller
+    // buys nothing.
+    it('never loads secrets for a _secrets_id supplied by the caller', async () => {
```
zapolsky, line 116: **[Security — verified by execution]** The new `COPY_CREDENTIALS` gate is keyed on `_copiedFrom`, but what actually grants access to another Data Mart's stored credentials is the credential pointer itself (`_secrets_id`, and equally `_source_credential_id`). A pointer sent **without** `_copiedFrom` reaches the same outcome with no permission check at all — so the invariant this test and the comment above it assert ("a pointer supplied by the caller buys nothing") does not hold in general, and neither does the changeset's "Copying from a Data Mart … not shared at all, is refused".

Four independent reviewers reached this chain; two of them reproduced it by running the real services.

**Evidence chain**

1. **Discovery.** `apps/backend/src/data-marts/controllers/data-mart.controller.ts:129-139` — `GET /data-marts/by-connector/:connectorName` is `@Auth(Role.viewer(Strategy.PARSE))` → `apps/backend/src/data-marts/use-cases/list-data-marts-by-connector-name.service.ts:17-30` → `apps/backend/src/data-marts/services/data-mart.service.ts:237-242` `find({ where: { projectId, definitionType } })` — **no visibility filter**, unlike every other listing in the module, which goes through `apps/backend/src/data-marts/utils/apply-data-mart-visibility-filter.ts` → `apps/backend/src/data-marts/mappers/data-mart.mapper.ts:205-206` → `:137-141` `mask()` → `apps/backend/src/data-marts/services/connector/connector-secret.service.ts:455-460`, which deliberately `continue`s over `_secrets_id` and `_source_credential_id` at every nesting level. Any project member reads the credential-row ids of Data Marts never shared with them.
2. **Gate skipped.** `apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts:45-58` collects sources only from `_copiedFrom.configId`; with no `_copiedFrom` anywhere the set is `[]`, so `:114-137` (the new `COPY_CREDENTIALS` loop) is never entered and control goes to `:167-172`.
3. **Survives the merge.** `connector-secret.service.ts:663-667` (`usesSourceCredentials` false ⇒ `_secrets_id` is not deleted), then `:669-673` / `:680-682` return `incomingItem` verbatim.
4. **Survives persistence.** `extractAndSaveSecrets` skips `_`-prefixed keys (`connector-secret.service.ts:99-101`), so with the secret fields left as `**********` `secrets === {}` and `:357-360` returns early — **before** `getCredentialsById` (`:363`), the `projectId` check (`:367`) and the `belongsToAnotherDataMart` fork (`:379-407`). The OAuth branch at `:338-343` returns earlier still. `deleteOrphanedSecrets` (`:744-747`) sees the id as referenced, so it is not reclaimed either.
5. **Dereferenced at run time.** `apps/backend/src/data-marts/services/connector/connector-executor.service.ts:266-270` → `connector-source-config.service.ts:34-41` → `connector-credential-injector.service.ts:199-239` (for `_secrets_id`; the only guard is `secretsEntity.projectId !== projectId` at `:220`) and `:58-77` (for `_source_credential_id`, same projectId-only guard). `connector-source-credentials.service.ts:50-55` looks up by id alone — no `dataMartId`, no re-resolution by `(dataMartId, configId)`.

**Reproduced** (real services, only repositories mocked): `mergeDefinitionSecrets` + `extractAndSaveSecrets` on `{ "_secrets_id": "<victim row>", "AccountIDs": "22" }` stored the pointer unchanged — `createSecretsForConfig` never called, no warning logged — and `injectSecrets(storedItem, "proj-1")` then returned the victim's real `AccessToken`.

**Trigger:** an editor who is technical owner of their own Data Mart calls `GET /api/data-marts/by-connector/FacebookMarketing`, lifts an `_secrets_id` from a Data Mart never shared with them, `PUT`s it onto a configuration item of their own Data Mart with the secret fields left masked, and runs it. Requires a hand-crafted request — the web client always stamps `_copiedFrom` (`apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep.tsx:260-264`).

**Attribution — please read before reacting:** the production code is inherited from the base branch (`3f4006816`); neither `e39c5e4cd` nor `fae7c5271` touches `extractAndSaveSecrets` or the injector. What *this* PR adds is the comment and test asserting the invariant holds, plus a changeset stating the boundary is now closed. It is raised here because the two PRs ship together; the fix may well belong in #1480.

**Fix:** treat a caller-supplied credential pointer as a copy, exactly as destinations already do — `apps/backend/src/data-marts/use-cases/create-data-destination.service.ts:123-148` requires `COPY_CREDENTIALS` when an incoming credential is already linked to another destination, not only when a copy marker is present. Concretely: drop `_secrets_id` from incoming items in `mergeDefinitionSecrets` (it is only ever minted at `connector-secret.service.ts:406` and legitimately restored from `previousItem` at `:699`), or move the ownership fork above the `secrets.length === 0` early return at `:357`. Worth tightening this test as well: it currently also passes via the unrelated early return at `:681` (the `_id` `'invented'` matches no previous item), and it never asserts that the foreign `_secrets_id` was dropped — which is the part that matters.

apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.spec.ts
```diff
@@ -168,6 +168,155 @@ describe('UpdateDataMartDefinitionService', () => {
...
+    expect(connectorSecretService.mergeDefinitionSecretsFromSource).not.toHaveBeenCalled();
+  });
+
+  it('refuses to copy a configuration from a Data Mart the user cannot copy credentials from', async () => {
+    const { service, dataMartService, accessDecisionService, connectorSecretService, dataMart } =
+      createService();
+
+    dataMart.definitionType = DataMartDefinitionType.CONNECTOR;
+    // EDIT on the target is granted, COPY_CREDENTIALS on the source is not.
+    accessDecisionService.canAccess.mockImplementation(async (...args: unknown[]) =>
```
zapolsky, line 283: **[CI — command-verified]** `npm run format:check` fails on this branch, so the `Code Quality OWOX / Prettier` job will be red.

**Evidence:** `.github/workflows/lint-owox.yml:5-9,30-33` runs `npm run format:check` for every pull request touching `apps/backend/**` → root `package.json` `format:check` → `apps/backend/package.json:45` `prettier --check "**/*.{ts,json}"` (`printWidth: 100` via `apps/backend/prettier.config.mjs` → `packages/prettier-config/base.js:11`; `.prettierignore` does not exclude spec files). Observed in a clean checkout of this head:

```
[warn] src/data-marts/services/connector/connector-secret.service.spec.ts
[warn] src/data-marts/services/connector/connector-secret.service.ts
[warn] src/data-marts/use-cases/update-data-mart-definition.service.spec.ts
```

**Introduced by this PR:** the arrow body on this line (`update-data-mart-definition.service.spec.ts:283-285`) — the file was Prettier-clean at `3f4006816` and is dirty now — plus ten 135-character call sites in `connector-secret.service.spec.ts` (`:672, 717, 756, 797, 826, 843, 865, 934, 967, 1010`), all of the form `await service.mergeDefinitionSecretsFromSource(incoming, fromSingleSource(sourceDefinition), SOURCE_DATA_MART_ID);`. Base versions were re-checked in place so the repo's Prettier config resolved correctly.

**Inherited from the base branch** (already red there, not this PR's regression, but it needs fixing for the job to go green): `connector-secret.service.ts:941` (stray blank line before the class-closing brace) and `connector-secret.service.spec.ts:497/548/948`.

ESLint is clean on every changed backend and web file, and `tsc -p tsconfig.build.json` reports no error in `data-marts` — so "typecheck and lint are clean" holds; `format:check` is a separate CI job and it is not.

**Fix:** `npm run format -w @owox/backend`, and the same on the `fix/copied-config-shares-credentials-row` base branch.

zapolsky (commented):

zapolsky:

Two blocked issues are posted inline. Everything below is non-blocking and does not gate merge. The permission model itself checks out: I enumerated all 75 role × ownership × sharing cells for `DATA_MART` and `EDIT ⟺ COPY_CREDENTIALS` holds in both directions, so self-copy and ordinary edits are unaffected, and `ACCESS_MATRIX`/`allActions` have no consumer that could change behaviour elsewhere. `_copiedFrom.dataMartId` does survive every hop from `ConfigurationStep.tsx` to `collectSourceDataMartIds` — "no frontend change was needed" is correct.

Primary notes (criticality ≥ 80), highest first:

- `apps/backend/src/data-marts/use-cases/list-data-marts-by-connector-name.service.ts:17-30` — **the read side of the boundary this PR closes is still wide open.** `GET /data-marts/by-connector/:connectorName` is `@Auth(Role.viewer(Strategy.PARSE))` and applies no visibility filter, so any project member — Business Users included — enumerates every connector Data Mart's title, full (masked) configuration and credential-row ids, including Data Marts `apply-data-mart-visibility-filter.ts` hides everywhere else. That listing is what makes the inline blocker's ids discoverable, and it also powers the "Copy from…" dropdown. Worth a follow-up ticket even though it predates this PR: run the result through `applyDataMartVisibilityFilter` (or `canAccessMany(…, Action.SEE)`).
- `apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts:121-143` — **a source deleted between "Copy from…" and Save is reported as a permission failure.** Because every source is authorized *before* any is read, and Data Mart deletion is a soft delete (`entities/data-mart.entity.ts:94`), `getSharingState` misses the row (`access-decision.service.ts:247-254` → `NOT_SHARED`) and a non-owning maintainer gets `403 You do not have permission to copy the configuration of this DataMart` for a Data Mart that simply no longer exists. A technical owner takes the other branch — `DataMartTechnicalOwner` rows survive a soft delete — passes the gate and gets a raw `404 … not found`. Two users, two opaque errors, same event. Resolving each source first and raising a "source no longer exists" `BusinessViolationException` would match the message you already wrote for the sibling case at `connector-secret.service.ts:891-895`.
- `apps/web/.../DataMartDefinitionSettings/form/CopyConfigurationButton.tsx:42` — **the picker still offers sources the save will now refuse, and a refused save leaves the form looking saved.** The list comes from the unfiltered endpoint above, so an editor can pick a Data Mart shared only for reporting (or one they merely *business*-own) and only learn at Save. The 403 does surface — `apiClient.ts:113-119` fires a persistent toast with the server message — but `DataMartContext.tsx:379-394` does not rethrow, so `DataMartDefinitionSettings.tsx:155-168` still runs `reset(data)` and `setShouldActualizeSchema(true)`, marking the unsaved definition clean and actualizing a schema that was never persisted. I checked the "dead end" reading and it does **not** hold — there are five in-product recoveries (pick a permitted source, delete the item, switch tabs and back, re-run setup in DRAFT, or schema actualization resetting the form) and the previously saved definition is never at risk. Cheapest fix in scope: filter the picker by `COPY_CREDENTIALS`, or rethrow so `onSubmit` skips `reset(data)`.
- `apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts:46-48` — **`_copiedFrom.dataMartId` reaches `canAccess` and the repository completely unvalidated**, while the field it effectively replaces is `@IsUUID()`-checked (`update-data-mart-definition-api.dto.ts:21-23`). `definition` carries only `@IsNotEmptyObject()` with no `@ValidateNested`/`@Type`, and the zod schema is `z.record(z.string(), z.any())` and is not parsed on the write path, so a non-string `dataMartId` produces an unhandled TypeORM error → 500 where the old code could only ever see a validated UUID. A `typeof === 'string'` guard in `collectSourceDataMartIds` closes it.
- `apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts:131-135` — **the 403 names no source.** No id in the message, no `errorDetails`, and no log line at the default level (`AccessDecisionService` logs the denial at `debug`; `DEFAULT_LEVEL` is `INFO`). With several sources in one save the user cannot tell which pick to undo, and support has nothing to correlate — while the sibling `BusinessViolationException`s 40 lines away all carry `{ sourceDataMartId, sourceConfigId }`. The id is caller-supplied, so echoing it leaks nothing.

<details>

- `.changeset/6784-authorize-copy-source-data-mart.md` — "you must own the source Data Mart" overstates the grant: a **business** owner is refused (`access-matrix.config.ts:198-221`), as is a technical owner holding only the `viewer` role (`:185-195`). Only a technical owner with the `editor` role, or an admin, qualifies. Suggest "you must be a *technical* owner of the source Data Mart, or it must be shared with you for maintenance".
- `apps/backend/src/data-marts/services/access-decision/access-matrix.config.ts:250-252` — the new comment claims `COPY_CREDENTIALS` follows `EDIT` "as it does for Storage and Destination", but Destination grants it to `Role.VIEWER` (`:354-356`, `:447-482`) while Data Mart denies it to viewers in every cell — which your own new e2e case encodes. No behavioural consequence; it is the justification a future reader will lean on.
- `apps/web/.../ConnectorEditForm/steps/ConfigurationStep.tsx:251-266` — using "Copy from…" *inside an existing configuration's* edit sheet merges into that item but the backend then re-keys it (`connector-secret.service.ts:919` `_id = randomUUID()`), so its incremental-import state is stranded in `connector_state` (`connector-state.service.ts:31,62-70`) and the next run re-imports from the connector's start date. Pre-existing, but it lives inside the path this PR rewrites; keeping the existing `_id` when the target item already had one would fix it.
- `apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts:50-64` — a copied item whose source cannot be resolved (a `configId` with no `dataMartId` and no request-level `sourceDataMartId`) is silently saved with its `_copiedFrom` intact, but 400s if an unrelated *sibling* item happens to name a source. Same item, two verdicts. The silent-save half is pre-existing and API-only (the UI has always stamped `dataMartId`); the sibling dependence is new. Rejecting a `configId` with no resolvable source in `collectSourceDataMartIds` makes it consistent.
- `apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts:60-64` — the fallback also means a request that names `sourceDataMartId` while annotating no item now needs `COPY_CREDENTIALS` on it, for a save that provably copies nothing (`connector-secret.service.ts:860-862` returns every item unchanged). API-only.
- `apps/backend/src/data-marts/dto/presentation/update-data-mart-definition-api.dto.ts` — removing `sourceConfigurationId` is a wire-level breaking change: `global-pipes.config.ts:6-10` sets `forbidNonWhitelisted`, so a client still sending it now gets a 400 rather than having it ignored. Verified that nothing in the repo ever *sent* it (only ever declared), so no first-party skew — but it deserves a line in the changeset for direct API consumers.
- `apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts:122-155` — two fully sequential `await` loops where `accessDecisionService.canAccessMany` (`access-decision.service.ts:156-169`) already exists and does exactly this, and each `getByIdAndProjectId` eager-loads seven relations to read one field. The "authorize everything before reading anything" property survives `Promise.all` on the first loop.
- `apps/backend/src/data-marts/services/connector/connector-secret.service.ts:846-853` — `getCredentialsByIds` now decrypts every configuration of every source, not just the items actually being copied, and it happens *before* the connector-type check that used to be the method's first statement. Building the id list from the resolved `sourceConfig`s only, and hoisting the type check into the load loop, restores the old fail-fast ordering.
- `apps/backend/src/data-marts/services/connector/connector-secret.service.spec.ts:852-868` and `:910-920` — both new tests assert only `rejects.toThrow(BusinessViolationException)`, and three distinct causes now throw that class. The one at `:852` previously asserted the message; deleting the `defaultSourceDataMartId` fallback would flip it to a different error and it would still pass. `update-data-mart-definition.service.spec.ts:245-275` ("refuses the whole save") likewise never asserts that `save` was not called or that the permitted source was never read — which is the headline claim.
- `apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts:178-195` — create-credentials → delete-orphans → `save` is still untransacted, and this PR multiplies the rows written per save. Pre-existing and strictly after authorization, so "a partly permitted save touches nothing" does hold.
- minor: stale numbered docblock at `connector-secret.service.ts:811-813` (step 1 is now a per-item check inside the loop); `collectSourceDataMartIds(definition, command)` takes the whole command to read one optional field; mixed paranoia between `definition.connector?.source?.configuration` (`:45`) and the unguarded `connectorDefinition.connector.source.name` (`:181`); the OAuth branch of `extractAndSaveSecrets` (`:338-343`) skips `removeMaskedSecretsRecursively`, so a copied OAuth item persists a literal `**********` in `AuthType.oauth2.RefreshToken` (harmless — the injector overwrites it — but it is a placeholder stored as if it were a credential).

</details>

**Refuted during verification, so *not* raised as findings** — recorded so nobody re-litigates them: copying an OAuth configuration aliases the source's `_source_credential_id` row, but that is documented, deliberate and pre-existing (`.changeset/6783-…`: "such a credential belongs to the project rather than to a single Data Mart"), the row can never be wiped by either Data Mart, and the one rotation race that exists self-recovers; adding `COPY_CREDENTIALS` to `allActions(DATA_MART)` changes no existing decision or API response; the removed positional constructor argument shifts nothing at any of the nine call sites; `errorDetails` echoes only caller-supplied ids; a cross-project source id is stopped by `getByIdAndProjectId`; and all four touched suites pass (83 tests).

<sub>Automated OWOX review · worktree multi-agent passes (5 specialists + fresh-eyes generalist, plus a second wave) · every blocker adversarially verified, 3 candidates refuted at the gate.</sub>

laskevych:
Closing this PR after triaging the review above.

**The blocking security finding was real, but the hole lives in the base branch's chokepoint, not in this PR's gate.** A bare `_secrets_id` with no `_copiedFrom` never entered the `COPY_CREDENTIALS` loop, and with every secret field masked the extraction path returned before its ownership check ran — so the forged pointer survived to run time. That is now fixed where every save passes, in #1480 (75cb69799): a caller-supplied pointer is dropped or forked on every save, for every connector, even when there are no secret values to re-home, and a cross-project pointer is rejected. With that in place, "a pointer supplied by the caller buys nothing" holds without a per-copy-source permission gate, and the reproduced exfiltration chain is closed by construction rather than by an additional check the caller can route around.

What else this PR carried, recorded so it isn't lost:

- **`COPY_CREDENTIALS` on the copy source** — a product-level boundary ("Copy from…" gated by sharing). The review shows it cannot be coherent while `GET /data-marts/by-connector/:connectorName` lists every connector Data Mart unfiltered: the picker offers sources the save would then refuse, a deleted source surfaces as an opaque 403, and the listing itself leaks titles and masked configurations of unshared Data Marts. If we want this boundary, it should ship together with the listing visibility fix as one coherent change.
- **Copying from several Data Marts in one save still fails with a 500** — real, low-frequency; worth a separate small PR when prioritized (the per-item `_copiedFrom.dataMartId` resolution from this branch is the right shape for it).
- The unused `sourceConfigurationId` DTO cleanup can ride any future PR.

The branch is kept for reference.

PR #1487, opened 2026-08-02
feat(web): add Copy link button for Destinations and Storages

artemfenoga:
# Copy a link to a Destination or Storage

Destinations and Storages now have a **Copy link** button right in their configuration panel, so you can grab a direct link and share it with a teammate — no more explaining which item to click through the list to find.

Screenshot 2026-08-02 221510


packages/ui/src/components/common/copy-link-button.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface CopyLinkButtonProps {
+  link: string;
+  ariaLabel: string;
+}
+
+export function CopyLinkButton({ link, ariaLabel }: CopyLinkButtonProps) {
+  const [copied, setCopied] = useState(false);
+
+  const handleCopy = () => {
+    void navigator.clipboard.writeText(link);
```
romandubovyi: ⚪ **Non-blocking:** The clipboard write is fire-and-forget, so `setCopied(true)` runs even when `writeText` rejects (insecure context, denied permission, unfocused document) — the user sees "Copied!" while the clipboard still holds the old value, and the rejection surfaces as an unhandled promise error. Only flip the state on success, e.g. `writeText(link).then(() => setCopied(true)).catch(...)`, or reuse `useClipboard` (`apps/web/src/hooks/useClipboard.ts`), which is explicitly documented as never throwing.

packages/ui/src/components/common/copy-link-button.tsx
```diff
@@ -0,0 +1,41 @@
+import { useState } from 'react';
+import { Check, Link } from 'lucide-react';
+
+interface CopyLinkButtonProps {
+  link: string;
+  ariaLabel: string;
+}
+
+export function CopyLinkButton({ link, ariaLabel }: CopyLinkButtonProps) {
```
romandubovyi, line 9: ⚪ **Non-blocking:** This duplicates `CopyToClipboardButton` in the same folder (`copy-to-clipboard-button.tsx`) — same `copied` state, same `writeText`, same 2s `setTimeout`, same `Check` swap. The only real difference is the visual treatment (plain text button with `Link` instead of a `Button` with `Copy`), which could be covered by adding an `icon`/`variant` prop there. Keeping one component means clipboard fixes (see the error-handling comment) land in one place instead of three.

apps/web/src/features/data-destination/edit/components/DataDestinationEditSheet/DataDestinationConfigSheet.tsx
```diff
@@ -46,6 +48,11 @@ export function DataDestinationConfigSheet({
 
   useIntercomLauncher(isOpen);
 
+  const { scope } = useProjectRoute();
+  const destinationLink = dataDestination
```
romandubovyi: ⚪ **Non-blocking:** `scope()` falls back to the unprefixed path when `projectId` is null (`shared/hooks/useProjectRoute.ts`). That fallback is fine for `navigate()` — which redirects to `/` in that case — but here it is interpolated straight into a shareable URL, producing `https://host/data-destinations?id=…`, which matches no route (routes live under `/ui/:projectId`, `routes/index.tsx:101-106`), and the user still gets a "Copied!" confirmation. Consider gating the button on `projectId` too: `const { scope, projectId } = useProjectRoute();` and building the link only when both are present. Same applies to `DataStorageConfigSheet.tsx:38`.

apps/web/src/features/data-destination/edit/components/DataDestinationEditSheet/DataDestinationConfigSheet.tsx
```diff
@@ -46,6 +48,11 @@ export function DataDestinationConfigSheet({
...
+    ? `${window.location.origin}${scope(`/data-destinations?id=${dataDestination.id}`)}`
```
romandubovyi: ⚪ **Non-blocking:** The URL construction is the part most likely to silently drift from what the deep link expects (`useUrlParam('id')` in `DataDestinationList.tsx:61`), and nothing asserts it. `apps/web` already has the setup for this — see `DataDestinationList.test.tsx`, which covers the deep-link path — so a single test asserting the button writes `/ui/<projectId>/data-destinations?id=<id>` to the clipboard would lock the contract down. (No test is expected in `packages/ui`, which has none.)

romandubovyi (approved):
Review complete. No blocking issues found — the deep link the button produces matches what the lists already handle via `useUrlParam('id')`, so the feature works end to end. Four non-blocking notes below: clipboard error handling, duplication with the existing `CopyToClipboardButton`, the unprefixed-URL fallback when `projectId` is missing, and test coverage for the link construction.

packages/ui/src/components/common/copy-link-button.tsx
artemfenoga: Good catch, fixed — setCopied(true) now runs only in .then() after a successful writeText, rejections are caught and logged instead of surfacing as unhandled errors.

apps/web/src/features/data-destination/edit/components/DataDestinationEditSheet/DataDestinationConfigSheet.tsx
artemfenoga: Fixed in both files — now destructuring projectId from useProjectRoute and gating the link on dataDestination/dataStorage && projectId, so the button doesn't render when there's no project scope to build a valid URL.

apps/web/src/features/data-destination/edit/components/DataDestinationEditSheet/DataDestinationConfigSheet.tsx
artemfenoga: Good idea — added DataDestinationConfigSheet.test.tsx and DataStorageConfigSheet.test.tsx, each asserting the copy button writes /ui/<projectId>/data-destinations|data-storages?id=<id> to the clipboard, plus a case confirming the button stays hidden without a resolvable projectId. Locks the deep-link contract down on both sides.

packages/ui/src/components/common/copy-link-button.tsx
artemfenoga: Good catch — but turns out CopyToClipboardButton wasn't actually used anywhere in the codebase (not even exported from common/index.ts), so it wasn't an established pattern to reuse, just dead code. I removed it instead of adapting CopyLinkButton to match it — it's built on shadcn's Button, which forces fixed height/padding/background that didn't fit the inline text-link look next to SheetDescription. Happy to reconsider if there's a reason to keep it around.

PR #1488, opened 2026-08-03
feat(data-marts): correct COUNT DISTINCT, SUM, AVG and percentiles across joined Data Marts (#6766)

zapolsky:
Fibery: #6766

## The problem

`COUNT DISTINCT`, `SUM`, `AVG`, percentiles and `Combined` were wrong on a joined Data Mart whenever the join fans out — including across a bridge, where the joined Data Mart is reached through another one.

Each of those metrics was pre-aggregated on the join key and then re-aggregated. Once the fan-out is collapsed that way, re-aggregating answers a different question than the report asked: `SUM` adds the same underlying row once per report row it fans out to, `COUNT DISTINCT` counts per-key results rather than distinct entities, `AVG` becomes an unweighted average of averages, a percentile is weighted by how often the join repeated each value, and `Combined` repeats the same text once per matching row.

## The approach

Each such metric now gets a **metric sleeve**: a CTE that re-joins the RAW (pre-dedup) path and recomputes the metric at the report's own dimension grain. The outer query pulls the result back with `ANY_VALUE` over a NULL-safe join on the dimension tuple.

Two shapes, declared in one table (`SLEEVE_ROUTING`) that also drives the output-controls gate, so adding a function is a single entry rather than four edits:

- `count-distinct` — a single-level `COUNT(DISTINCT col)` at the report grain.
- `value` — a `SELECT DISTINCT (dimensions, owner identity, value)` pass wrapped by the metric's own aggregate. Used by everything whose answer depends on how many times a value appears.

The owner identity is the joined Data Mart's own declared primary key when it has one, and a per-row surrogate otherwise. Rows agreeing on the key and on the value collapse; rows agreeing on the key but not the value stay separate, so contradictory data is never silently resolved to one version.

These are symmetric (non-additive) aggregates: an entity reachable under two dimension values counts once in each group but once overall, so per-group values need not sum to Totals.

Applies to BigQuery, Athena, Redshift, Snowflake and Databricks.

## What changes for existing reports

**Numbers change** for any saved report using one of these metrics on a fanning joined Data Mart — that is the fix. Three further changes are worth calling out:

- **Totals now honour metric (`HAVING`) filters.** They used to be dropped, so Totals summarised rows the report itself hides.
- **`COUNT` is no longer offered next to `COUNT DISTINCT` on a joined field**, in report columns as well as Totals. The two are computed at different grains and shown together can invert `COUNT DISTINCT <= COUNT`.
- **Some previously-saved reports are now rejected on their next run** with an explanation instead of returning wrong data: a relationship with no join conditions or a nested join key, two output columns Redshift cannot tell apart, and a metric filter on a joined sleeve-routed metric.

The changeset carries the full list.

## Also in here

- A joined "any value" pre-join roll-up is now deterministic. Its dedup CTE is read twice — by the outer grouping and inside every sleeve — and an engine free to evaluate the two separately could return different values, which made the correction silently produce empty results.
- Every failed validation rule now reaches the caller. `ValidationPipe` reports them as a string array, which the exception filter collapsed into the single phrase "Bad Request Exception".
- `query_data_mart` reports the type of joined columns, which came back empty for exactly the columns a caller cannot infer.

## Verification

- 6348 backend unit tests, 1212 web tests, `nest build` and web `tsc -b` clean, eslint clean.
- Live on BigQuery, arithmetic proven against seeded ground truth rather than SQL text: fan-out through a bridge, `SUM`/`AVG` set-based correctness, Totals restricted by a metric filter at a month bucket, a duplicated joined row counted once with a declared key (170) and twice without (270), a percentile over the de-duplicated distribution (median 20, not 30), and `Combined` listing each joined row once.
- Percentile SQL is pinned per dialect on all five warehouses.

## Known limitations

- A metric (`HAVING`) filter on a joined sleeve-routed metric is rejected with a clear error rather than supported.
- A declared primary key is trusted. One that is not truly unique, or whose column is empty on more than one row, merges rows — only BigQuery's schema records nullability, so this cannot be enforced.
- A percentile's answer, not just its spelling, differs per warehouse: BigQuery and Athena approximate, `PERCENTILE_CONT` interpolates.


amarchenk0 (approved):

apps/backend/src/data-marts/data-storage-types/blending/metric-sleeve.planner.ts
```diff
@@ -0,0 +1,391 @@
...
+    }
+    const entry = fieldIndex.get(metric.column);
+    if (!entry) {
+      throw new BusinessViolationException(
+        `groupValueSleeveMetrics: no fieldIndex entry for value-sleeve metric column='${metric.column}' ` +
+          `(the column is aggregated but missing from the blended field index)`
+      );
+    }
+    const dims = Array.from(dimensions);
+    const key = `${entry.cteName}\u241F${dims.join('\u241F')}`;
```
kalnik-a-a: Criticality: HIGH / BLOCKING

Could we please keep each value-sleeve dedup group scoped to a single metric column? This key contains only the owner CTE and dimensions, so `SUM(revenue)` and `SUM(tax)` from the same joined Data Mart share one `DISTINCT (dimensions, owner_identity, revenue, tax)` tuple. For two rows with the same identity and `revenue = 100`, but `tax = 10` and `tax = 11`, both tuples survive and `SUM(revenue)` becomes `200` instead of `100`.

Please include `metric.column` in the grouping key. Multiple functions for the same column (`SUM`, `AVG`, percentiles, or `Combined`) can still share one pass. A regression test should assert the arithmetic for two columns where the first value repeats while the second differs, rather than checking SQL text only.

apps/backend/src/data-marts/data-storage-types/blending/re-aggregation.ts
```diff
@@ -0,0 +1,61 @@
...
+ * matches nothing: SUM/AVG come back NULL and COUNT DISTINCT a confident 0. `MAX` picks a stable
+ * member of the same set. Without a resolvable type the declared function stands, since MAX is
+ * not valid for every type.
+ */
+export function preJoinAggregateFunctionFor(
+  aggregateFunction: AggregateFunction,
+  fieldType?: string
+): AggregateFunction {
+  if (aggregateFunction !== 'ANY_VALUE' || !fieldType) return aggregateFunction;
+  return ORDERABLE_CATEGORIES.has(categorizeFieldType(fieldType)) ? 'MAX' : aggregateFunction;
```
kalnik-a-a, line 60: Criticality: HIGH / BLOCKING

This deterministic replacement makes ordinary `MIN`/`MAX` report metrics read a pre-join value collapsed with `MAX`, while sleeve-routed `AVG` reads the raw rows. The metrics therefore use different grains for the same joined column. For one order with values `[10, 20]`, the generated query returns `MIN = 20`, `AVG = 15`, and `MAX = 20`, violating `MIN <= AVG <= MAX`.

Could we please make `MIN`/`MAX` use the same raw report grain, or explicitly reject the incompatible combination until one consistent semantic is chosen? Please cover the decision with an arithmetic regression test across the supported dialects.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -691,9 +788,18 @@ export class OutputControlsValidatorService {
...
       if (blended.isHidden) continue;
+      // `blendable-schema.service` already clamps a stored override to the effective type's
+      // supported set, so this menu is legal by the time it arrives; keep the fallback for a
+      // schema assembled without that service.
       allowedByColumn.set(
         blended.name,
-        blended.postJoinAggregations ?? resolveFieldGovernance(blended.type).allowedAggregations
+        // Same rule Totals applies: a joined column must not offer both COUNT and
+        // COUNT_DISTINCT, which are computed at different grains and can invert.
+        withoutCountBesideSleevedCountDistinct(
```
kalnik-a-a: Criticality: HIGH / BLOCKING

Applying this menu rule during every `validateForReport()` call breaks persisted configurations instead of normalizing them. A previously valid joined `COUNT` can now fail with `AGGREGATION_FUNCTION_NOT_ALLOWED_FOR_FIELD`; this also affects a `COUNT`-only config on default joined string/boolean fields because the allowed menu contains `COUNT_DISTINCT`, so this helper removes `COUNT` regardless of what the saved report selected. The changeset instead says that `COUNT` is unchanged and that a saved `COUNT + COUNT_DISTINCT` pair will show only `COUNT_DISTINCT`.

Could we please add an explicit migration, grandfathering, or controlled normalization path for both `COUNT`-only and `COUNT + COUNT_DISTINCT` persisted reports, with run-path regression coverage?

apps/backend/src/data-marts/services/blendable-schema.service.ts
```diff
@@ -273,18 +273,24 @@ export class BlendableSchemaService {
...
+        // No override → effective-type governance default; explicit `[]` = none allowed. An
+        // explicit override may only NARROW the effective type's SUPPORTED set: it used to be
+        // carried verbatim, so a stored config could offer a function the type cannot run. The
+        // validator's type floor catches the obvious ones (SUM/AVG on a non-number,
+        // COUNT_DISTINCT/STRING_AGG on a non-scalar) but NOT percentiles — `P50` on a text field
+        // passed save-time validation and only failed at the warehouse. Clamping here fixes it
+        // for every consumer at once: this menu feeds the validator, the MCP field list and the
+        // web column picker.
+        dto.postJoinAggregations = resolveFieldGovernance(effectiveType, {
+          allowedAggregations: fieldOverride?.postJoinAggregations,
```
kalnik-a-a, line 292: Criticality: HIGH / BLOCKING

This DTO is the menu consumed by both the Web picker and MCP field metadata. For a default joined string/boolean field, `resolveFieldGovernance()` publishes both `COUNT` and `COUNT_DISTINCT`, while `OutputControlsValidatorService` applies `withoutCountBesideSleevedCountDistinct()` and rejects `COUNT`. A user or AI client can therefore select an aggregation exactly as advertised and receive a 400 on save or run.

Could we please apply the same effective rule before assigning `postJoinAggregations`—ideally from one shared source of truth—and add a contract test covering `blendable schema -> Web/MCP menu -> save/run validation`?

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -253,6 +265,8 @@ If truncated is true, not all matching rows were returned: narrow the query (few
             ...(source.note ? { note: source.note } : {}),
           })),
         },
+        // The fact of the failure is what the caller acts on; the reason stays in Run History.
+        ...(res.totalsError ? { totals_error: TOTALS_UNAVAILABLE_MESSAGE } : {}),
```
kalnik-a-a, line 269: Criticality: MEDIUM / NON-BLOCKING

The comment says the detailed reason remains in Run History, but `QueryDataMartService` does not include `totalsError` in the `recordMcpQueryRun()` metadata, and `McpQueryRunMetadataSchema` does not declare the field. Since the persistence boundary parses through Zod, the reason is not retained for later diagnosis.

Could we please add optional `totalsError` metadata, pass it when recording the run, and cover the post-parse persisted value? Rows and the generic MCP failure signal still work, so I am treating this as non-blocking.

apps/web/src/app/api/apiClient.ts
```diff
@@ -122,6 +122,23 @@ apiClient.interceptors.response.use(
     if (error.response?.status === 400 && !skipErrorToast) {
       showApiErrorToast(error, 'Bad request');
     }
+
+    // A 5xx used to show the user NOTHING: only 400/403/404 were toasted, and the backend
+    // deliberately sends no `message` for a non-HTTP exception (it could carry SQL or internals),
+    // so a failed action just looked like it did nothing. Report the failure and the request id —
+    // that id is the only handle support has to find the matching server log.
+    const status = error.response?.status;
+    if (status !== undefined && status >= 500 && !skipErrorToast) {
```
kalnik-a-a, line 131: Criticality: MEDIUM / NON-BLOCKING

This global handler emits a new toast for every 5xx response. `DataMartReportsPage` silently polls each visible `RUNNING` report every five seconds via `getReportById()`/`Promise.allSettled()` without `skipErrorToast`; a rejected response leaves the stored status as `RUNNING`, so an outage can add one toast per report every five seconds indefinitely.

Could we please use a stable toast id or suppression window, or set `skipErrorToast` for silent polling and surface one controlled page-level error? A fake-timer regression for repeated polling failures would make the behavior explicit.

kalnik-a-a (changes requested):
Criticality: HIGH / BLOCKING

Thank you for the substantial work in this PR. I am requesting changes because four confirmed issues can produce incorrect aggregate results, break previously valid persisted reports, or advertise configurations that the backend rejects. I also left two non-blocking observations about MCP Run History diagnostics and repeated 5xx polling toasts. The inline comments include concrete examples and acceptance criteria.

apps/backend/src/data-marts/data-storage-types/blending/metric-sleeve.planner.ts
zapolsky: Fixed in 4fa5a8b8 — the grouping key now carries the metric column, so only metrics on the *same* column share a dedup pass. `SUM` + `AVG` + a percentile + `Combined` on one column still merge into a single `SELECT DISTINCT`, which is the saving the merge existed for.

One clarification on the scope, because it narrows what to test: the defect only bites under a **declared-primary-key** identity. With the per-row surrogate (`__owox_rid`) every raw row already carries a distinct value, and a non-identity owner reads its dedup CTE, which is one row per pre-join group key — in both shapes the identity leg alone separates the rows, so extra value slots change nothing. The declared key is the first identity that *deliberately* collapses rows, which is exactly what a second column's variation then undoes.

`resolveValueSleeveGroupCteName`'s multi-column branch became unreachable and is gone; it now throws if handed a group spanning more than one column. `dimensionsFingerprint` had no other caller and was removed with it.

Coverage as requested — arithmetic, not SQL text:
- Unit (`metric-sleeve.builder.spec.ts`): two metrics on different columns produce two `SELECT DISTINCT` passes and no `_val_0`; several functions on one column produce one.
- Live BigQuery (`bigquery.integration.ts`): the fixture holds an order with two raw rows whose values differ (40 and 41) beside one that repeats, so a shared tuple would inflate the neighbouring column.

apps/backend/src/data-marts/data-storage-types/blending/re-aggregation.ts
zapolsky: Agreed on the invariant, and fixed: `MIN` and `MAX` are now sleeve-routed (`SLEEVE_ROUTING`, 4fa5a8b8), so every metric on a joined column reads the same rows. No new builder branch was needed — the `value` shape already spells any function through the dialect's `buildAggregation`.

One correction on the cause, since it changes what the fix is for. The grain split is **older than the deterministic replacement**: `MIN`/`MAX` have always read the dedup CTE while sleeve-routed `AVG` read the raw rows. Before `ANY_VALUE` was made deterministic the roll-up picked an arbitrary member, so on `[10, 20]` the old `MIN` returned 10 *or* 20 depending on the plan — the invariant could already fail, just unpredictably. Making it `MAX` pinned the old number at 20; it did not create the disagreement. So this change replaces an undefined value with a consistent one rather than moving a stable answer.

Scope worth recording: for a **non-identity** joined field the sleeve reads the owner dedup CTE's already-aggregated column, and `MIN`/`MAX` are indifferent to repetition — those reports are unchanged. Only an `ANY_VALUE` passthrough on a relationship that actually fans out sees a different number.

Live evidence (`bigquery.integration.ts`, 84f872f9) — one order with raw rows 40 and 41:

| | before | after |
| --- | --- | --- |
| `MIN` | 41 | 40 |
| `AVG` | 40.5 | 40.5 |
| `MAX` | 41 | 41 |

Consequence I would rather state than bury: `SLEEVE_ROUTED_FUNCTIONS` also gates metric (`HAVING`) filters, so a `HAVING` on a joined `MIN`/`MAX` is now rejected with the existing clear error. Blended Totals over a joined numeric field change on the same narrow population. Both are in the changeset.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
zapolsky: Confirmed, and thank you — a `COUNT`-only saved report did break, which is the opposite of what the changeset promised. Fixed in 4fa5a8b8, though not by grandfathering.

I tried the pruned-menu-plus-grandfathering shape first and dropped it: `buildAggregationGovernance` derives the joined allowed set from `blended.postJoinAggregations`, the same field a menu prune would narrow, so re-admitting `COUNT` in the validator is the only way to make that combination work — and it would make an *explicit* per-field exclusion (`postJoinAggregations: ['COUNT_DISTINCT']`, set by an admin) unenforceable, since the validator cannot tell an intentional exclusion from a grain-rule one.

What is in place instead:

- **The validator no longer prunes.** A saved joined `COUNT` validates exactly as before, alone or beside `COUNT_DISTINCT`. `COUNT` on its own is legitimate — it answers a different question rather than a distorted one.
- **The contradictory pair is normalised when the query is built**, so the two are never shown together. `COUNT` is dropped, `COUNT_DISTINCT` kept, and the persisted config is untouched.
- **Totals keep pruning** their derived set, which creates no mismatch because nothing was selected there.

The normalisation lives in `BlendedReportDataService.resolveBlendingDecision`, not in the composer. `run-report.service`, `stream-http-data.service` and `report-data-cache.service` each resolve the decision themselves and only fall through to `compose` when blending is not needed, so a compose-time fix would have missed every scheduled, Looker and HTTP Data run. The decision also carries the normalised list, because readers emit one header per (column, function) and would otherwise return a `<col> | COUNT` header with no matching SQL column.

Run-path coverage: `blended-report-data.service.spec.ts` asserts both the SQL built for the joined query and the normalised list handed back for headers.

apps/backend/src/data-marts/services/blendable-schema.service.ts
zapolsky: Right, and resolved by the same change as the previous comment — from the other end, so it is worth saying why the menu was left alone.

The mismatch was that the menu advertised `COUNT` while the validator rejected it. Rather than narrow the menu to match the rejection, the rejection is gone: the validator accepts a joined `COUNT` again, so what is advertised is exactly what is accepted, and no client can be told yes and then handed a 400.

Pruning `postJoinAggregations` here would have relocated the regression rather than fixed it. This DTO field is the same one `OutputControlsValidatorService.buildAggregationGovernance` derives the joined allowed set from, so narrowing it makes a saved `COUNT`-only report fail validation instead of the menu. It also conflates two different statements — "an admin excluded `COUNT` for this field" and "the grain rule removed it" — leaving no way to enforce the first.

So the single source of truth is the rule, not a pre-pruned field: `withoutCountBesideSleevedCountDistinct` is applied where a *set* is derived with nothing selected (Totals) and where a *selection* is normalised before the query is built. The picker can still offer both; selecting both yields `COUNT_DISTINCT`, which the changeset states.

Contract coverage: `blendable-schema.service.spec.ts` pins the advertised menu for a default joined string field, and `output-controls-validator.service.spec.ts` covers both directions — a joined `COUNT` validates, while an explicit per-field menu that excludes it is still enforced.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Confirmed and fixed in 4fa5a8b8. `McpQueryRunMetadataSchema` gained an optional `totalsError`, and the successful run records it — optional so runs written before the field existed still parse, matching every other optional key there. It is set only when totals actually failed, and the shape now matches the HTTP Data run metadata, which already recorded the reason.

Chasing it turned up the same false claim a second time: the tool's failure path carried "Full error stays in Run History", and nothing preserved it there either — `recordMcpQueryRun` has a single caller and it is success-path only, so a failed, timed-out or cancelled query is deliberately never recorded. That half of the sentence is removed rather than given an invented destination. If those runs should be preserved, that is a separate change.

Coverage asserts twice, because the passing value was never the problem: on what reaches `recordMcpQueryRun`, and on `McpQueryRunMetadataSchema.parse(...)` of that same object — the second is what fails on the old code, since Zod's key-stripping is what dropped it.

One consequence worth flagging: the run mapper forwards the whole `mcpQuery` subtree and the web Run History renders its keys generically, so the raw warehouse message now appears there without any frontend change. That is consistent with `executionSqlQuery`, already in the same subtree; the redaction stays where it matters, on the MCP tool response.

apps/web/src/app/api/apiClient.ts
zapolsky: Confirmed and fixed in 4fa5a8b8, at both layers since they defend different things.

**The silent polling no longer toasts.** `getReportById` takes an optional Axios config, the same way its siblings in that service already do, and `DataMartReportsPage` passes `{ skipErrorToast: true }` from a module-level constant — matching the `SILENT_REQUEST_OPTIONS` idiom used by the other silent refreshes. Existing single-argument callers are unaffected.

**The 5xx toast no longer stacks.** It now carries `id: server-error:<status>`. Keyed on the status rather than the message on purpose: every 5xx body carries a fresh `requestId`, so a message-keyed id — the shape the persistent 403 branch uses — would still produce one toast per failed request. It is not made persistent; only de-duplicated.

Coverage: repeated 500s with different request ids assert the same stable id, and the polling path asserts `skipErrorToast` is forwarded.

Two things left alone deliberately, both easy follow-ups: the poll also trips the global loading indicator every five seconds (`skipLoadingIndicator` is a separate behaviour change nobody asked for), and `report-status-polling.service` still toasts on failure — but it stops polling inside its `catch`, so it produces at most one toast per run rather than a stream.

kalnik-a-a (approved):

PR #1490, opened 2026-08-03
fix(data-quality): verify project balance before execution

kalnik-a-a:
## Summary

- verify the project can perform operations immediately before executing pending Data Quality rules
- stop Data Quality compilation and warehouse queries when the project balance blocks operations
- add regression coverage that also prevents consumption from being registered for blocked runs

## Root cause

The Data Quality executor registered consumption after successful runs but, unlike other billable execution flows, did not call the shared project balance guard before querying the warehouse.

## Impact

Blocked projects now terminate through the existing Data Quality failure path before any Data Quality query runs or consumption is published.

## Validation

- `AWS_EC2_METADATA_DISABLED=true npm test -w @owox/backend -- --runInBand` — 461 suites and 6063 tests passed
- `npm run lint -w @owox/backend` — 0 errors; one unrelated pre-existing warning in `sql-clause-renderer.spec.ts`
- Prettier check and `git diff --check`


apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
```diff
@@ -104,6 +106,7 @@ export class RunDataQualityService {
     try {
       signal?.throwIfAborted();
       if (pendingRules.length > 0) {
+        await this.projectBalanceService.verifyCanPerformOperations(dataMart.projectId);
```
prokopetsv, line 110: 🟡 **Non-Blocker** — `canPerformOperations` rethrows on Balance API infra errors (`project-balance.service.ts:80-85`), not just on a blocked verdict. So a transient Balance API outage now fails DQ runs that previously succeeded, marking every pending rule ERROR. This fail-closed behavior matches the sibling flows, so it may be intended — but it's a new availability dependency for DQ. If product prefers fail-open on infra errors, catch non-`ProjectOperationBlockedException` here, log, and proceed. Otherwise fine as is.

apps/backend/src/data-marts/use-cases/run-data-quality.service.spec.ts
```diff
@@ -1018,6 +1026,21 @@ describe('RunDataQualityService', () => {
     expect(runQueryBuilder.innerJoinAndSelect).toHaveBeenCalledWith('dataMart.storage', 'storage');
   });
 
+  it('does not execute Data Quality queries when project balance blocks operations', async () => {
```
prokopetsv: 🟡 **Non-Blocker** — good coverage of the blocked path, but the placement's other half is untested: a resumed run with **all results already persisted** (`pendingRules.length === 0`) must finish successfully *without* calling `verifyCanPerformOperations` — those results are already paid for. A refactor that hoists the guard above the `pendingRules` check would silently break that, and no test would go red. One test asserting `verifyCanPerformOperations` is not called and the run finishes SUCCESS would pin it.

prokopetsv (commented):
The fix does what the description says: the shared balance guard now runs before any Data Quality compilation or warehouse query, matching the placement and argument (`dataMart.projectId`) used by `run-report.service.ts:237`, `connector-executor.service.ts:107`, `stream-http-data.service.ts:275`, and `query-data-mart.service.ts:131`. The guard sits inside `pendingRules.length > 0`, so resumed runs with fully persisted results still finish without a balance call. On block, `ProjectOperationBlockedException`'s user-facing message lands in each rule's stored result via `failedResult` → `description`, so run history explains the failure. Constructor change is DI-only; the trigger-handler spec mocks the service, so nothing else breaks. Two 🟡 non-blockers inline, no blockers.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
kalnik-a-a: We intend to keep the fail-closed behavior here. Data Quality should only start warehouse work after the shared balance check succeeds; failing open on Balance API errors would allow execution without confirming available credits and would also diverge from the sibling flows. We'll keep the current behavior.

apps/backend/src/data-marts/use-cases/run-data-quality.service.spec.ts
kalnik-a-a: Added a regression test for the fully persisted resume path in 87489482d. It verifies that verifyCanPerformOperations is not called when pendingRules is empty and that the run finishes with SUCCESS.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
laskevych, line 110: 🟡 **Important (non-blocking):** Every other balance-guarded flow maps `ProjectOperationBlockedException` to the warning-level `RESTRICTED` status: `connector-executor.service.ts:172-176` + `:734-740`, `base-report-run.model.ts:103-106`, and the UI/API distinguish it (`DataMartHealthStatusCell.tsx:42` maps RESTRICTED → WARNING, `StatusBadge.tsx:80` renders a dedicated yellow badge, `mcp-reports.facade.impl.ts:119` maps it to `'restricted'`). Here the exception falls into the generic catch, so a blocked DQ run (a) lands as a red `FAILED` — DQ becomes the only flow that never uses `RESTRICTED` for this exact exception; (b) fabricates a `DATA_QUALITY_EXECUTION_ERROR` result per pending rule, so N checks look like they ran and failed; and (c) gets the generic constant in `run.errors`, so `FailedRunsAllDmNotification` emails/webhooks carry no billing cause (the connector path persists the actual exception message into `run.errors`, so its notifications do). Net effect: a blocked connector run on a project shows "Restricted" with an actionable message while the DQ run on the same project looks like a data-quality regression.

I realize DQ run status is derived from `DataQualitySummaryState`, which has no RESTRICTED analog, so aligning this needs a small design decision (a new summary state, or special-casing the exception before the generic catch) — fine as a follow-up issue rather than this PR, but worth capturing so it doesn't get lost.

laskevych, line 110: 🔵 **Question (non-blocking):** Resume scenario — attempt 1 passes the guard, executes and persists 8 of 10 rules (real warehouse spend), then the worker dies while the run is RUNNING. On re-processing, `claimRun` re-enters via the RUNNING branch, `pendingRules.length > 0`, and if the project got blocked between attempts the guard now rejects → the run ends FAILED → the `status === SUCCESS` gate at line 232 means consumption for the 8 already-executed queries is never registered — exactly for the project that is over its limit, with no retry since FAILED is terminal in `claimRun`. This loss mode pre-exists for other mid-run failures, and the new test intentionally pins "no consumption for blocked runs" — but for partially executed resumed runs the spend already happened. Is that an accepted trade-off, or should consumption be registered for the persisted results in this case?

apps/backend/src/data-marts/use-cases/run-data-quality.service.spec.ts
```diff
@@ -1018,6 +1026,21 @@ describe('RunDataQualityService', () => {
...
+    );
+
+    await service.executeExistingRun('run-1', 'project-1');
+
+    expect(projectBalance.verifyCanPerformOperations).toHaveBeenCalledWith('project-1');
+    expect(snapshotTableReference.resolve).not.toHaveBeenCalled();
+    expect(compiler.compile).not.toHaveBeenCalled();
+    expect(executor.executeChecks).not.toHaveBeenCalled();
+    expect(consumption.registerDataQualityRunConsumption).not.toHaveBeenCalled();
+    expect(dataMartRun.status).toBe(DataMartRunStatus.FAILED);
```
laskevych: 💡 **Suggestion (non-blocking):** The neighboring failure-path tests (storage-mismatch at ~923, definition-snapshot at ~945) also assert the persisted contract — `dataQualityResults` content, `dataQualitySummary.state === EXECUTION_FAILED`, and `errors`. This test stops at run status, so a refactor that swallowed the exception without appending per-rule results (while still ending FAILED) would stay green, leaving users with a FAILED run and zero explanatory results. Mirroring lines ~954-964 costs ~5 lines with the existing mock plumbing: `error.code: 'DATA_QUALITY_EXECUTION_ERROR'` with the credit-limit message as `description`, plus the `EXECUTION_FAILED` summary state — that would pin what the UI actually shows for a blocked run.

laskevych (approved):
Reviewed the guard placement, the failure path, and the downstream consumers (run status semantics, run history, notifications, consumption). The fix itself is correct and consistent with the sibling flows: the guard sits inside `pendingRules.length > 0` so resumed fully-persisted runs skip the Balance call, the claim-then-check ordering matches both scheduled siblings (connector and report trigger handlers also claim RUNNING before the balance check fires), and the DI wiring is fine. Three inline comments below — none blocking. One more note that doesn't anchor to the diff:

💡 **Suggestion (non-blocking, follow-up):** This adds the 8th uncached `verifyCanPerformOperations` call site. Scheduled DQ/connector/report triggers fan out one run per data mart per slot (processed in parallel on a 5s cron via `Promise.all`), so a project with N data marts issues N identical can-perform HTTP round-trips per slot. A short-TTL per-project memo inside `ProjectBalanceService` would collapse this for all call sites transparently — out of scope here, but probably worth a ticket. (Caveat for that ticket: caching `allowed: true` lets a just-blocked project slip a few runs through the TTL window.)

Approving — no blockers.

apps/backend/src/data-marts/use-cases/run-data-quality.service.ts
kalnik-a-a: Implemented in 2b627ec10. A blocked Data Quality run now persists both `DataMartRunStatus.RESTRICTED` and `DataQualitySummaryState.RESTRICTED`, stores the original `ProjectOperationBlockedException` message in `run.errors`, and does not synthesize per-rule `ERROR` results for checks that never ran. The web UI/API now expose a warning-level Restricted presentation as well. Thanks for catching the inconsistency.
CleanShot 2026-08-03 at 16 01 13

kalnik-a-a: Thanks for calling this out. We’re keeping the existing success-only consumption behavior unchanged in this PR; partial-run accounting needs a separate product decision and durable idempotent design, so we won’t address it here.

apps/backend/src/data-marts/use-cases/run-data-quality.service.spec.ts
kalnik-a-a: Updated in 2b627ec10. Since the blocked path now has a dedicated restricted contract, the regression asserts the new persisted behavior: `RESTRICTED` run and summary states, no synthetic rule results, the actionable credit-limit message in `run.errors`, no query/consumption, and `finishedAt`. I also added coverage for terminal `RESTRICTED` handling and the web mappings. Thanks!

kalnik-a-a:
Thanks also for the balance-check caching suggestion in the review summary. We’ll keep caching out of scope for this PR and revisit it separately.

PR #1491, opened 2026-08-03
feat(data-marts): name joined fields as "Field (Data Mart)" in Google Sheets

laskevych:
## Problem

Columns coming from a joined Data Mart lead with the Data Mart name — `RFM_SEGMENT Recency Score`, `CLINIC city`. In a spreadsheet header cell that is the worst possible place for it: the part you are actually looking for gets pushed out of view, on every joined column at once, and you end up hovering cells to read the real field name.

"Just use shorter Data Mart titles" is not a workaround. Connector-based Data Marts carry the endpoint in the title because several endpoints are needed to cover the required fields, and prepared client tables frequently cannot be renamed at all without colliding with a name already taken elsewhere in the model.

## Change

Google Sheets now writes the Data Mart name **after** the field name, in parentheses:

| Before | After |
| --- | --- |
| `RFM_SEGMENT Recency Score` | `Recency Score (RFM_SEGMENT)` |
| `CLINIC city` | `city (CLINIC)` |

Details:

- The suffix is **unconditional**, not "only when the field name is ambiguous within the report". A uniqueness-driven label would make a column's header depend on which other columns happen to be selected, so adding an unrelated field would silently rename an existing one.
- For a chain of joins the name is the Data Mart the field actually comes from, not the whole path — the same string the picker groups that field under.
- Fields from the report's own Data Mart are untouched, and a blank Output Alias still yields a bare field name (never `city ()`).
- An aggregated joined field reads `Cost (Products) | SUM` — the Data Mart name stays glued to the field it qualifies and the function token lands last, as for every other column.

**Scope: the Google Sheets destination only.** Data Studio, email-based destinations, the HTTP data endpoint and MCP field metadata keep the Data Mart name as a prefix.

## How the destination choice is expressed

Rather than comparing enum members inside the blending service, this follows the convention already in the codebase for per-destination behaviour — `ReportExecutionPolicyResolver` asks `report.isEmailBasedDestination()`, never `type === EMAIL`:

- `usesSuffixedJoinedFieldNames(type)` — a named capability predicate in `data-destination-type.enum.ts`, next to its existing siblings `isEmailBasedDataDestinationType` / `isPullBasedDataDestinationType`. This is the module whose job is classifying destination types.
- `usesSuffixedJoinedFieldNames(report)` — the `ReportLike` narrowing helper in `report-like-read-plan.ts`, alongside `shouldIncludeRowCount` / `hasOutputControls`, which exist for exactly this "the union may or may not carry this field" problem.
- `BlendedReportDataService` asks the question by name and imports no destination enum at all.

A read plan carries no destination — the totals query, the HTTP data endpoint and MCP all build one — so those reads keep the prefix. A newly added destination type defaults to the prefix until it opts in deliberately, and the enum test is driven off `Object.values` so it is covered the moment it appears.

## Backward compatibility

The label is display-only: it never reaches SQL. Identifiers come from `ReportDataHeader.name` / `buildBlendedFieldUnifiedName`, so there is no migration, no config change and no effect on saved column selections or generated SQL.

Existing sheets migrate **in place** on their next refresh. `OWOX_COLUMNS` developer metadata still holds the previous `{name, alias}` pairs, so `ColumnPlanBuilder` resolves the old prefixed strings in row 1 back to canonical names, emits no structural ops, and `writeHeaders()` rewrites row 1 with the new labels. No column is inserted, deleted or reordered, and nothing shifts under user content to the right of the imported range. Covered by two new cases in `column-plan-builder.spec.ts`, including one where the user had reordered the joined columns by hand.

## Tests

- `blended-field-display-name.spec.ts` — both styles, alias fallback chain, nested struct paths, and empty / whitespace-only Data Mart name for each style.
- `data-destination-type.enum.spec.ts` — enum-driven, pins the prefix default for every non-Sheets type.
- `blended-report-data.service.spec.ts` — the header alias per destination: Google Sheets, Looker Studio, and no destination at all.
- `report-data-headers.utils.spec.ts` — the aggregation token lands after the joined-field label.
- `column-plan-builder.spec.ts` — the in-place migration, with and without a user reorder.

The MCP facade, totals composer and report-data-cache suites were run unchanged to confirm the out-of-scope surfaces stay on the prefix.

## Docs

`joinable-data-marts.md`: the Output Alias and per-field Alias sections now describe both conventions and which destination uses which. Also fixes a pre-existing inaccuracy in the Transitive Joins section, which claimed fields are qualified with the whole alias chain — only the leaf Data Mart's alias is ever used.


apps/backend/src/data-marts/services/blended-field-display-name.ts
```diff
@@ -5,11 +5,43 @@ export interface BlendedFieldDisplayNameSource {
...
+ * name at all. A blank alias therefore falls through to the next candidate, and a blank Data Mart
+ * name yields the bare field name — never `Field name ()`.
  */
-export function formatBlendedFieldDisplayName(field: BlendedFieldDisplayNameSource): string {
-  const fieldName = field.alias || field.originalFieldName || field.name;
+export function formatBlendedFieldDisplayName(
+  field: BlendedFieldDisplayNameSource,
+  style: BlendedFieldNameStyle = 'prefix'
+): string {
+  const fieldName = field.alias?.trim() || field.originalFieldName?.trim() || field.name;
```
prokopetsv, line 41: 🔵 **Question** — the trim/fall-through normalization applies to the **prefix** style too, so it changes existing labels on Looker Studio, MCP `displayName`, and HTTP `title` for any user whose alias carries padding or is whitespace-only. That sits in mild tension with the enum-side rationale for keeping those surfaces on the prefix ("changing them would break naming continuity for their existing consumers"). The changeset documents it and the affected input is degenerate, so this is likely fine — just confirming the continuity exception is intentional rather than incidental. If Looker Studio matches fields by name/ID only, no action needed.

docs/getting-started/setup-guide/joinable-data-marts.md
```diff
@@ -77,18 +77,22 @@ By default, all fields from the target Data Mart are available in reports built
 The **Output Alias** at the top of the tab controls how the joined Data Mart is presented to report editors:
 
 - It's the **group label** shown in the Report Columns picker.
-- It's the **prefix** added to every joined field name in the report output (e.g., `orders revenue` if the alias is `orders`).
+- It's the **Data Mart name** attached to every joined field name in the report output. Where it goes depends on the destination — with the alias `orders`, a field named `revenue` becomes:
+  - `revenue (orders)` in **Google Sheets**, so the field name stays readable in a narrow header cell;
+  - `orders revenue` in **Data Studio**, **email-based destinations** and the **HTTP data endpoint**.
+
+The position follows the surface that renders the label, not the report. Reading a Google Sheets report through the HTTP data endpoint therefore returns `orders revenue`, even though its sheet shows `revenue (orders)` — match columns on the technical field name, which is identical everywhere, rather than on the label.
```
prokopetsv: 💡 **Suggestion** — the second sentence here runs ~35 words, over the codified 20-word cap (repo CLAUDE.md docs style). Possible split: "Reading a Google Sheets report through the HTTP data endpoint therefore returns `orders revenue`, even though its sheet shows `revenue (orders)`. Match columns on the technical field name — it is identical everywhere." The new Alias table cell has the same issue if the rule applies to cells.

prokopetsv (commented):
Reviewed at head `e265071`. No blockers and nothing important found — I verified the load-bearing claims rather than taking them on faith. The in-place Sheets migration holds: `ColumnPlanBuilder` resolves old row-1 labels through the previous `OWOX_COLUMNS` snapshot (tested, including a user reorder), a sheet with no metadata falls to the first-run plan with no structural ops, and the unknown-string fallback only bites on hand-edited headers — a pre-existing class this PR doesn't widen. The prefix default is genuinely preserved everywhere else: `report-data-cache` serves only Looker Studio (prefix), `buildBlendedTotalsHeaders` and the MCP facade call the formatter with the default, and the web picker uses `outputPrefix` purely as a group label, so no client-side divergence. The soft-deleted-destination guard with its regression test, the enum-driven default pinning, and the HTTP read-plan exclusion test are all careful work. One 🔵 question and one 💡 docs-style item inline.

apps/backend/src/data-marts/services/blended-field-display-name.ts
laskevych: Intentional, and thanks for pushing on it — the tension you spotted was real, but it was in my wording rather than in the behaviour.

I checked whether any of the three prefix surfaces actually keys on the label:

- **Looker Studio** — `buildSchemaField` sets `name: header.name` (the field ID) and `label: header.alias || header.name`. Saved charts reference the ID, so a label change is cosmetic.
- **MCP** — the `fields` input on `query_data_mart` takes *exact field names*; `displayName` is output-only.
- **HTTP data** — `toMetadataDataDescription` emits `title` alongside `name`, and consumers match on `name`.

One detail worth knowing: `tabular-serializer.ts:29` already does `column.displayName.trim() || column.name`, so for MCP the whitespace-only case was **already** rendering as the bare field name before this PR. The only thing my change alters there is an internal doubled space (`Orders  revenue` → `Orders revenue`). So the visible delta is confined to Looker Studio, and only on degenerate input.

What I did fix is the sentence you were reacting to. "changing them would break naming continuity for their existing consumers" reads as a promise of byte-exact labels, which is not what the predicate guarantees. The JSDoc now says it pins the **position** of the Data Mart name, notes that every style still normalizes whitespace, and lists what each surface actually keys on — so the next reader does not infer a stronger contract than exists. `7c660382d`.

docs/getting-started/setup-guide/joinable-data-marts.md
laskevych: Applied the split, near enough verbatim to your wording — thank you.

One correction to the premise, though: there is no `CLAUDE.md` in this repository, and no 20-word cap anywhere in it (I grepped every `.md`). The surrounding prose does not observe one either — the opening sentence of this same guide runs 36 words, the relationship description 28, and Step 3's second sentence 23. So the rule as stated is not something I could have been following.

The suggestion stands on its own merits regardless: that sentence was carrying a fact and an instruction stitched together with an em dash, and the instruction reads better standing alone. It is now:

> …even though its sheet shows `revenue (orders)`. Match columns on the technical field name — it is identical everywhere.

I gave the **Alias** cell the same treatment: the example is its own sentence instead of a trailing parenthetical. Not for a word count, but because it is the part a reader scans for. `7c660382d`.

apps/backend/src/data-marts/services/blended-field-display-name.ts
```diff
@@ -5,11 +5,43 @@ export interface BlendedFieldDisplayNameSource {
...
+  const dataMartName = field.outputPrefix?.trim();
+
+  if (!dataMartName) return fieldName;
 
-  return field.outputPrefix ? `${field.outputPrefix} ${fieldName}` : fieldName;
+  return style === 'suffix' ? `${fieldName} (${dataMartName})` : `${dataMartName} ${fieldName}`;
```
amarchenk0, line 46: The suffix convention can create a display-alias collision that did not exist before. For example, a native field aliased `Revenue (Orders)` and a joined `revenue` field with Output Alias `Orders` used to render as `Revenue (Orders)` and `Orders Revenue`, but the first refresh now writes `Revenue (Orders)` twice. That duplicate is persisted in `OWOX_COLUMNS`. On the next refresh `ColumnPlanBuilder` resolves both cells to the first canonical name and builds `['revenue', 'revenue', 'orders__revenue']`: the first column is cleared, an extra column is inserted, and data/user content shifts. Please guarantee unique rendered labels, or make the snapshot diff unambiguous for duplicates, with a two-refresh regression test.

amarchenk0 (changes requested):
Please address the blocking issue in the inline comment.

apps/backend/src/data-marts/services/blended-field-display-name.ts
laskevych: Confirmed and fixed in `15cb43e5d` — thank you, the mechanism is exactly as you described.

I reproduced it against `ColumnPlanBuilder` before touching anything. Second refresh, given row 1 `['Revenue (Orders)', 'Revenue (Orders)', 'user stuff']` and a snapshot holding both under the same label:

```
finalImportedNames: ["revenue","revenue","orders__revenue"]
ops:                [["insert", 2, "orders__revenue"]]
nameToFinalIndex:   revenue→1, orders__revenue→2
```

So index 0 never receives a header, is written blank and loses its column's data; `orders__revenue` looks new and is inserted, pushing `user stuff` right. Your description was precise.

**One correction on provenance.** The failure turns on two columns sharing a *display string*, not on which convention produced it. Re-running the same repro with prefix-style labels (`Orders Revenue` twice) gives byte-identical output — same duplicate, same insert, same blank cell. So this predates the suffix change; what the PR does is relocate *which* configurations collide, not introduce the mechanism. `ColumnPlanBuilder`'s C6 comment already documents the class as known. That is not a reason to leave it, just context for how far back it reaches.

**Fix** — I took your first option, guaranteeing unique rendered labels, since it also closes the pre-existing case. `prepareToWriteReport` now runs a `findDuplicateRenderedLabels` check mirroring the existing C5 name check one level up, on `alias || name`, and throws a `BusinessViolationException` before any sheet mutation:

> Duplicate column headers in report output: Revenue (Orders). Two columns would be written to the sheet under the same header. Change one of their aliases, or the Output Alias of the joined Data Mart they come from.

C5 could not catch this on its own — the names are distinct precisely in the case that breaks.

**Tests**, including the two-refresh regression you asked for:

- `column-plan-builder.spec.ts` — pins the corruption the second refresh produces, asserting all three of the outputs above, so the reason the guard exists stays visible.
- `google-sheets-report-writer.spec.ts` — the guard rejects your exact native/joined pair and asserts `updateValues` and `clearValuesInRange` were never called, i.e. the sheet is untouched; a second case proves it fires on the prefix form too; a third proves no false positive when an unaliased column falls back to its name.

Also noted in the changeset, since a report that used to write silently will now fail with a message.

amarchenk0 (approved):

PR #1492, opened 2026-08-03
feat(data-marts): share new Data Marts for maintenance by default

laskevych:
## Problem

A new Data Mart was created with **Shared for maintenance** off. That toggle is what gates joining: creating a relationship requires `Edit` on *both* Data Marts, and a non-owner Technical User only gets `Edit` through it.

So the advertised "add one more column from another Data Mart" flow did not work out of the box. The analyst who wanted the join had to notice that a toggle exists, find the other Data Mart's owner, and wait for them — or, more likely, build a duplicate Data Mart instead.

The default was protecting a corner case (someone edits my Data Mart) at the cost of the main one (handing a joined model over to a business user).

## What changed

- `CreateDataMartService` now creates Data Marts with `availableForMaintenance: true`. This covers the UI create flow and **Copy report as Data Mart**, which goes through the same service.
- `SyncLegacyDataMartService` sets both sharing flags explicitly for Data Marts it creates. It previously set neither and fell through to the DB column default, so it would have kept producing maintenance-private Data Marts. In the legacy extension anyone with access to the same GCP project could already edit the query, so the new default matches the access those Data Marts used to have.
- `DataMart.availableForMaintenance` column metadata flipped to `default: true`, mirroring `availableForReporting`.

## Not changed

- **No migration, no backfill.** Data Marts that already exist keep their current sharing settings.
- **Storages and Destinations** keep their defaults: shared for use, not shared for maintenance. The maintenance flag there governs credentials and connection configs, which is a different risk.
- **The sharing gate itself.** Joining a Data Mart that is *not* shared for maintenance is still refused. The owner turns the toggle off and everything behaves exactly as before.
- No DB default rewrite: every insert path now passes the value explicitly, so the column default is metadata only.

## Defaults after this change

| Resource | Shared for reporting / use | Shared for maintenance |
|---|---|---|
| Data Mart | On | **On** |
| Storage | On | Off |
| Destination | On | Off |

## Tests

- `data-mart-default-sharing.e2e-spec.ts` (new) — a non-owner Technical User creates a relationship to a freshly created Data Mart with no availability call anywhere in the setup, and gets 403 on the same call once the owner turns maintenance sharing off.
- `sync-legacy-data-mart.service.spec.ts` — a newly synced Data Mart gets both flags; an existing one keeps its own settings on re-sync.
- `create-data-mart.service.spec.ts`, `sharing-fields.spec.ts`, `ownership.e2e-spec.ts` — updated expectations.
- The `permissions-*`, `report-operate-access` and `project-activity-lists` e2e suites pass unchanged; they set sharing explicitly rather than relying on the default.

## Docs

`ownership-and-sharing.md` claimed new resources start with both toggles off, which was already wrong for the first toggle. Replaced with a per-resource defaults table, a note on how to opt out, and a line in the Joinable Data Mart section.

### DIFF
--- a/.changeset/6721-shared-for-maintenance-default.md
+++ b/.changeset/6721-shared-for-maintenance-default.md
@@ -1,0 +1,11 @@
+---
+'owox': minor
+---
+
+# New Data Marts are shared for maintenance by default
+
+A newly created Data Mart now starts with both sharing toggles on: **Shared for reporting** and **Shared for maintenance**. Other Technical Users can join it to their own Data Marts straight away, instead of first having to find the owner and ask them to share it.
+
+Data Marts that already exist are not changed. The owner can still turn **Shared for maintenance** off on any Data Mart in one click — non-owners then keep reporting access and lose the ability to edit or join it.
+
+Storages and Destinations keep their current defaults: shared for use, not shared for maintenance.
--- a/apps/backend/src/data-marts/entities/data-mart.entity.ts
+++ b/apps/backend/src/data-marts/entities/data-mart.entity.ts
@@ -77,7 +77,7 @@ export class DataMart implements CreatorAwareEntity {
   @Column({ type: 'boolean', default: true })
   availableForReporting: boolean;
 
-  @Column({ type: 'boolean', default: false })
+  @Column({ type: 'boolean', default: true })
   availableForMaintenance: boolean;
 
   @OneToMany(() => DataMartBusinessOwner, owner => owner.dataMart)
--- a/docs/project/ownership-and-sharing.md
+++ b/docs/project/ownership-and-sharing.md
@@ -86,7 +86,17 @@ The second toggle is **Shared for maintenance** for all resource types. The comb
 | First toggle off, second on | Can see, use, edit, and delete the resource |
 | Both toggles on | Both of the above |
 
-> ☝️ New resources start with both toggles off. Existing resources were migrated to both toggles on to preserve previous access patterns. Owners can gradually reconfigure sharing to match their intended access model.
+### Defaults for new resources
+
+| Resource | Shared for reporting / use | Shared for maintenance |
+|---|---|---|
+| Data Mart | On | On |
+| Storage | On | Off |
+| Destination | On | Off |
+
+A new Data Mart is shared for maintenance so that other Technical Users can join it to their own Data Marts right away, without asking the owner for access first. If you do not want that, turn **Shared for maintenance** off on the Data Mart — non-owners keep reporting access and lose the ability to edit or join it.
+
+> ☝️ Changing the default does not touch Data Marts that already exist. Resources created before the sharing model was introduced were migrated to both toggles on to preserve previous access patterns.
 
 ![Resource settings page with the Shared for reporting and Shared for maintenance toggles](https://imagedelivery.net/zKr-4bdC5CBGL2DuuEmvYw/f8622cf6-8d75-46ee-e9e2-b869f987a500/public)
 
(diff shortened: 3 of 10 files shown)


apps/backend/src/data-marts/entities/data-mart.entity.ts
```diff
@@ -77,7 +77,7 @@ export class DataMart implements CreatorAwareEntity {
   @Column({ type: 'boolean', default: true })
   availableForReporting: boolean;
 
-  @Column({ type: 'boolean', default: false })
+  @Column({ type: 'boolean', default: true })
```
prokopetsv, line 80: 🟡 **Non-Blocker** — this repo runs `synchronize: false` with migrations (`data-source-options.config.ts:115`), so flipping the metadata default without a migration leaves entity and DB defaults disagreeing: existing and fresh DBs keep `DEFAULT false`. Benign today (both insert paths are explicit, and `sharing-fields.spec` pins the metadata), but the next `migration:generate` will emit this ALTER into someone else's unrelated migration. Fix: add the one-line migration now, or note the intentional drift here.

.changeset/6721-shared-for-maintenance-default.md
```diff
@@ -0,0 +1,11 @@
+---
+'owox': minor
+---
+
+# New Data Marts are shared for maintenance by default
+
+A newly created Data Mart now starts with both sharing toggles on: **Shared for reporting** and **Shared for maintenance**. Other Technical Users can join it to their own Data Marts straight away, instead of first having to find the owner and ask them to share it.
```
prokopetsv, line 7: 🟡 **Non-Blocker** — the changeset sells the default as "other Technical Users can join it", but per the sharing table in `ownership-and-sharing.md`, maintenance sharing grants **see, use, edit, and delete**. An owner reading only this changeset won't realize every Technical User in the project can now also edit or delete their new Data Mart by default. Worth one honest sentence naming the full grant — the opt-out line already covers the remedy.

docs/project/ownership-and-sharing.md
```diff
@@ -86,7 +86,17 @@ The second toggle is **Shared for maintenance** for all resource types. The comb
...
-> ☝️ New resources start with both toggles off. Existing resources were migrated to both toggles on to preserve previous access patterns. Owners can gradually reconfigure sharing to match their intended access model.
+### Defaults for new resources
+
+| Resource | Shared for reporting / use | Shared for maintenance |
+|---|---|---|
+| Data Mart | On | On |
+| Storage | On | Off |
+| Destination | On | Off |
+
+A new Data Mart is shared for maintenance so that other Technical Users can join it to their own Data Marts right away, without asking the owner for access first. If you do not want that, turn **Shared for maintenance** off on the Data Mart — non-owners keep reporting access and lose the ability to edit or join it.
```
prokopetsv, line 97: 🟡 **Non-Blocker** — both new sentences here exceed the codified 20-word cap and open passive ("is shared"), per the repo docs style rules; the Joinable Data Mart addition at line 205 has the same issue. Example rewrite: "A new Data Mart starts shared for maintenance. Other Technical Users can join it right away, without asking the owner. To opt out, turn **Shared for maintenance** off — non-owners keep reporting access only."

prokopetsv (commented):
Verified the PR's key claims at head `72295cd`: only two `dataMartService.create` call sites exist and both now pass the flags explicitly, so no insert path falls through to the DB default. The e2e suite tests the right things — the default via the API, a non-owner join succeeding with zero availability calls in setup, and the 403 after the owner opts out. Legacy re-sync preserving existing settings is pinned by a unit test. No blockers; three 🟡 non-blockers inline.

amarchenk0 (approved):

PR #1493, opened 2026-08-03
feat(plugins): plugin gallery and the plugin SDK

romandubovyi:
One GitHub repository is one plugin, deployment-wide. Releases are read from GitHub, the highest eligible SemVer becomes the current version, and every active installation follows it. Publication scopes decide who sees a plugin; installations are per member per project and only ever soft-deleted. The deployment — not the member — owns activation: every plugin under maintenance is checked daily in a stable slot derived from its id, and **Check now** only brings that check forward.

## Reviewing this

The commits are grouped by subsystem and ordered by dependency, so they can be read one at a time:

| Commit | What |
|---|---|
| `feat(backend)` machine-readable code | optional `code` on API failures, so clients branch on `PLUGIN_SUSPENDED` rather than on a sentence |
| `fix(backend)` fail closed on unresolvable host | SSRF guard extracted from the webhook service; an unresolvable host is now a refusal, not a pass |
| `feat(plugin-sdk)` | the package plugin authors install from npm — protocol, handshake, typed calls |
| `feat(auth)` plugin runtime access tokens | short-lived token bound to one installation, marked `authFlow: plugin` |
| `feat(plugins)` plugin host backend | the module itself: records, authorization, GitHub sync, daily schedule, audit |
| `feat(api-client)` / `feat(ctl)` | typed control-plane operations and their CLI |
| `feat(web)` gallery and runtime host | gallery, detail, history, and the iframe host bridge |
| `fix(destinations)` withhold credentials | a plugin runtime token no longer receives stored destination credentials |
| `docs(plugins)` | authoring guide, environment variables, changeset |
| `build(plugin-sdk)` publish standalone | nothing in the repo imports the SDK; the publish workflow builds it explicitly |
| `fix(plugins)` / `refactor(plugins)` | spec-conformance copy on every surface, and the drift risks review found |

## Security-relevant decisions

- The plugin holds **no credential**. Its SDK transport posts to the host over a `MessagePort`; the host page performs the request with the runtime token, which never enters the plugin document. Path resolution compares origins rather than string prefixes, so `//evil.example/x` cannot make the host send the token elsewhere.
- The iframe is sandboxed with exactly `allow-scripts allow-downloads` — adding `allow-same-origin` alongside scripts is the classic escape, so the value is asserted in a test.
- A runtime token is refused on the installation lifecycle: installing another plugin or activating a version is the member's decision, not the third-party page's.
- Stored destination credentials (e.g. a Looker Studio secret key) are withheld from plugin traffic. Business data reaching a plugin is accepted by the design; a credential is not.
- Delivery URLs are validated against private and metadata networks through redirects and DNS, and must permit iframe embedding.

## Deliberate deviations from the specs

Both were reviewed against the spec text and accepted rather than missed:

1. **Navigation gate** (Gallery §10). The sidebar shows **Plugins** when the Gallery has an installable plugin; the spec also lists "active installation or restorable soft-uninstalled installation". A member whose publication was withdrawn keeps the plugin but loses the nav entry to it.

## Testing

- backend: 339 plugin-host tests across 32 suites; 4 e2e suites including an entity/migration drift guard and a route-registration guard for the catch-all `GET :pluginId`
- web: full suite green (1342), including the host bridge, sandbox attributes and XSS regressions
- api-client 163, plugin-sdk 23, owox-ctl 66
- `owox` and `@owox/web` build clean

## Notes

- Schema arrives in a single migration (`1784000000000-create-plugin-tables`); there is no follow-up column migration to apply.
- `@owox/plugin-sdk` is published for plugin authors and is imported nowhere in this repo, so the publish workflow builds it explicitly and waits for it on npm.
- Deployment publishing authority comes from `OWOX_DEPLOYMENT_PLUGIN_PUBLISHER_API_KEY_IDS`; an unset value denies everyone and must never mean "any".


apps/backend/src/idp/guards/idp.guard.ts
```diff
@@ -207,6 +224,54 @@ export class IdpGuard implements CanActivate {
     }
   }
 
+  private async checkPluginRuntimeAuthorization(
+    tokenPayload: Payload,
+    rejectPluginAuth: boolean
+  ): Promise<void> {
+    if (tokenPayload.authFlow !== 'plugin') {
```
amarchenk0, line 231: Plugin runtime tokens are accepted on every `@Auth` route unless the route explicitly opts into `@RejectPluginAuth()`. For example, `POST /api/project-member-api-keys` accepts a viewer token and returns the newly created API-key secret. A plugin can proxy that request through the bridge, receive the secret, and retain project access after its runtime token expires or the installation is removed. Please make plugin auth default-deny with an explicit allowlist (or protect every credential/minting endpoint) and cover this path with an integration test.

apps/backend/src/common/helpers/safe-url.helper.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+/** True for ::1, ::, fe80::/10 and fc00::/7. */
+export function isPrivateIpv6(address: string): boolean {
+  const normalized = address.toLowerCase();
+  return (
+    normalized === '::1' ||
+    normalized === '::' ||
+    normalized.startsWith('fe80:') ||
```
amarchenk0: `isPrivateIpv6` only rejects literal `fe80::`, although the link-local range is `fe80::/10`, so `https://[fe81::1]/` passes. IPv4-mapped loopback addresses such as `https://[::ffff:127.0.0.1]/` pass too. Both forms can direct the server-side probe to an internal network. Please classify the full IPv6 ranges and mapped IPv4 addresses before fetching, with tests.

apps/backend/src/plugin-host/services/remote-url-validator.service.ts
```diff
@@ -0,0 +1,238 @@
...
+  private async run(rawUrl: string): Promise<RemoteUrlCheckResult> {
+    let currentUrl = rawUrl;
+
+    for (let hop = 0; hop < this.config.maxRedirectHops; hop++) {
+      const guarded = await this.guard(currentUrl);
+      if (!guarded.ok) {
+        return guarded;
+      }
+
+      const response = await this.probe(currentUrl);
```
amarchenk0, line 81: DNS is checked separately before `fetch`, but `fetchWithBackoff` resolves the original hostname again when it connects. Publisher-controlled DNS can return a public IP during `guard()` and a private IP for the actual request. The tests cover a private answer during the first check, not this TOCTOU/DNS-rebinding path. Please pin the connection to a vetted IP or validate the lookup used by the actual transport for every probe, including the descriptor request.

amarchenk0 (changes requested):
Please address the three security issues in the inline comments.

apps/backend/src/plugin-host/use-cases/update-plugin.service.ts
```diff
@@ -0,0 +1,97 @@
...
+        : null;
+      const refreshed = await this.pluginService.findById(plugin.id);
+      diagnostics = buildPublisherDiagnostics(refreshed, version, result.report);
+    }
+
+    const scheduled = await this.pluginService.findById(plugin.id);
+
+    return {
+      pluginId: result.pluginId,
+      repository: result.repository,
```
amarchenk0: `POST /api/plugins/:pluginId/update` is available to an ordinary viewer, including one without an installation, and returns `repository: owner/name`. For a private plugin, the member view deliberately withholds the repository name and URL, so this reveals the private repository to every member who can reach the page. Return the canonical repository only to an authorized publisher, or redact it for ordinary members, and add a regression test.

packages/plugin-sdk/src/iframe-transport.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    putJson: <T>(path: string, jsonBody: unknown) =>
+      json<T>({ kind: 'api', method: 'PUT', path, body: jsonBody }),
+
+    async getStream(path: string, query?: URLSearchParams): Promise<Response> {
+      const response = await send({
+        kind: 'api',
+        method: 'GET',
+        path,
+        query: query ? Object.fromEntries(query) : undefined,
```
amarchenk0: `Object.fromEntries(query)` drops duplicate query keys. For example, `ctx.owox.dataMarts.traverseData(..., { column: ['a', 'b'] })` normally emits two `column` parameters, but only the last one reaches the host; the `Record<string, string>` protocol cannot restore them. The plugin SDK therefore returns a different dataset from the regular API client. Preserve multi-value query entries in the protocol and append them at the host, with a regression test.

apps/backend/src/plugin-host/use-cases/install-plugin.service.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    // Any installed member can move a plugin forward at any moment, so the version the
+    // member read about may already be gone. Reject rather than install something they
+    // never saw; the new version travels back so the screen can ask again.
+    if (command.expectedVersionId !== plugin.currentVersionId) {
+      const current = await this.versionService.findById(plugin.currentVersionId);
+      throw new StaleVersionConfirmationError(plugin.currentVersionId, current?.semver ?? null);
+    }
+
+    const existing = await this.installations.findOne(
```
amarchenk0, line 75: Initial concurrent install/publish requests use a `find → create` sequence, so one can hit a unique-constraint 500 even though the operation is documented as idempotent. This does not corrupt data, but a retry or double-click can fail. Handle the unique conflict by re-reading/restoring the row and add a concurrency test.

amarchenk0 (changes requested):

apps/backend/src/idp/guards/idp.guard.ts
romandubovyi: Fixed in `f8179c35f`. Inverted the default: `@RejectPluginAuth` is gone and the guard now refuses a plugin runtime token unless the route carries `@AllowPluginAuth()`. `POST /api/project-member-api-keys` is closed, with a regression test asserting none of its handlers opt in.

The allowlist is 17 routes, derived from what `ctx.owox` actually calls: data marts (list/detail), runs, insight templates, the NDJSON traversal, storages, destinations, search, model canvas, auth context, project settings, setup progress, the markdown transform, and the member's own installation reads. Publication management and every credential path are out.

Two boundary calls worth your view: I left `PUT /api/projects/settings/description` closed (the only write in the non-publisher `ctx.owox` surface), and left `GET /api/data-destinations` open on the basis that credentials are already withheld from plugin traffic by this PR. Happy to move either.

apps/backend/src/common/helpers/safe-url.helper.ts
romandubovyi: Fixed in `958c9ed90`. `isPrivateIpv6` no longer prefix-matches — it classifies via `ipaddr.js` (already resolved in the tree, now declared): `ipv4Mapped` unwraps to the existing IPv4 rules so `::ffff:8.8.8.8` still works, and anything but plain `unicast` is refused, which also covers multicast, 6to4, Teredo and reserved space.

Tests cover `fe81::1` / `fe90::1` / `feaf::1` / `febf::1`, `::ffff:127.0.0.1` and its hex spelling `::ffff:7f00:1`, `::ffff:10.0.0.1`, `::ffff:169.254.169.254`, `ff02::1`, `2002:7f00:1::`, and the deprecated `::127.0.0.1`, which does not parse and is refused rather than assumed public. Four of them are also asserted through `assertPublicHttpUrl`, not just the predicate.

One range I deliberately left alone: `fec0::/10` (site-local) classifies as `unicast` and stays allowed — deprecated by RFC 3879 and not routed. Say the word if you'd rather close it.

apps/backend/src/plugin-host/services/remote-url-validator.service.ts
romandubovyi: Fixed in `958c9ed90`, taking your second option. Pinning to a vetted IP breaks SNI and certificate validation for HTTPS, so instead the check moved inside the transport's own resolution: a shared `undici.Agent` whose `lookup` validates every returned address and refuses the name if any of them points inside. The address vetted is the address connected to, so the window is gone rather than narrowed.

Both callers use it — the webhook path and `probe()`. Since `probe()` is the only way the validator reaches the network, that covers every redirect hop and the descriptor request you called out.

Tests drive the second resolution specifically, including the case that matters most: a public and a private address in the same answer is refused.

apps/backend/src/plugin-host/use-cases/update-plugin.service.ts
romandubovyi: Fixed in `c9a2f3258`. `repository` now goes through the same publisher check that already gated `diagnostics`: ordinary members get `owner/***`, deployment publishers get the canonical value. The owner is kept because `toSource` discloses it anyway. Regression tests cover private+member, private+publisher, and public unchanged; the OpenAPI description was updated, as it claimed the value was unconditionally canonical.

One visible consequence: `owox-ctl plugins update` on a private plugin under a non-allowlisted key now prints `owner/***`.

packages/plugin-sdk/src/iframe-transport.ts
romandubovyi: Fixed in `7376bf4cd`, and thank you — the `traverseData` example made it concrete. The protocol now carries ordered pairs instead of `Record<string, string>` (structuredClone moves them intact), and the host appends rather than sets. `getJson` sends pairs too, so the host reads one shape. Both copies of the protocol were updated together.

Regression tests on both sides: the SDK keeps both values of a repeated key, and the bridge asserts `url.searchParams.getAll('column')` returns both.

apps/backend/src/plugin-host/use-cases/install-plugin.service.ts
romandubovyi: Fixed in `a9bccd685`, for both paths you named. `installOrFind` and `createOrFindForRepo` re-read the winner's row on a unique violation, using the `isUniqueConstraintViolation` helper the version service already uses. A conflict that leaves no row still surfaces — reporting success for a row that does not exist would be worse than the 500 it replaces.

One behavioural detail: the request that loses the race records `RESTORE`, not a second `INSTALL`, since that event means a first grant of a member's authority to a third party and the winner already wrote it. Eight tests across the two services plus two at the use-case level.

apps/backend/src/idp/guards/idp.guard.ts
romandubovyi: Correction to my previous reply: we moved off default-deny in `851061589`, and took the other option you offered — protect every credential and minting endpoint explicitly.

Why we switched: per-action plugin permissions are the next piece of work here, and both decorators will be retired by it. The allowlist was short-lived machinery with a wide blast radius — 18 opt-ins meant every controller in the app had to know about plugins, and all of it would have been unwound shortly.

What is there now: `@RejectPluginAuth` is back, and sits next to `@RejectApiKeyAuth` on all 21 sites. The reasoning is that an endpoint closed to an API key returns or mints something no credential should reach, and a plugin runtime token is a credential held by a third party. `POST /api/project-member-api-keys` — the path you named — is covered by the controller-level pair.

They stay two decorators rather than one unified check, because the sets are not identical: `owox-ctl plugins update` runs under an API key while a plugin must not, so the installation lifecycle refuses plugins and admits API keys. A single flag cannot express that.

The honest tradeoff, since it is the same failure mode that produced the original finding: default-allow means a new endpoint is plugin-reachable unless someone remembers otherwise. The safety net is a test rather than the type system — `plugin-auth-coverage.spec.ts` reads the controller sources and requires `@RejectPluginAuth()` on the line after every `@RejectApiKeyAuth()`, adjacency rather than a tally, plus the five lifecycle routes. I confirmed it fails when a decorator is removed.

Its limit is worth stating: it catches a `@RejectApiKeyAuth()` that arrives without its pair, but not a brand-new credential endpoint carrying neither decorator. That gap closes when per-action permissions land.

docs/plugins/authoring-guide.md
```diff
@@ -0,0 +1,163 @@
+# Building an OWOX plugin
+
+A plugin is a web app you host yourself. OWOX learns about it from a GitHub repository,
```
max-voloshyn:
```
A plugin is a web app you host yourself.
```
It's temporary, we should talk about it as a feature, not the main principle

docs/plugins/authoring-guide.md
```diff
@@ -0,0 +1,163 @@
+# Building an OWOX plugin
```
max-voloshyn: It should be a dedicated section between API and Contributing

docs/plugins/authoring-guide.md
```diff
@@ -0,0 +1,163 @@
...
+**Your entry page must be embeddable.** It must not send `X-Frame-Options`, and if it
+sends a `Content-Security-Policy` with a `frame-ancestors` directive, that directive must
+permit OWOX — `*`, `https:`, or your OWOX deployment's exact origin. OWOX checks this at
+publication time and refuses a page it could never display.
+
+**Your entry page must announce itself.** Serve a descriptor beside it, or publication is
+refused with `SDK_HANDSHAKE_FAILED`:
+
+```jsonc
+// https://plugin.example.com/.well-known/owox-plugin.json
```
max-voloshyn: May we handle it on plugin-sdk level without requiring serving a special endpoint?

docs/plugins/authoring-guide.md
```diff
@@ -0,0 +1,163 @@
...
+running inside an OWOX frame, and if no host answers within 10 seconds.
+
+|                                                          |                                                                                                                                                                  |
+| -------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ctx.owox`                                               | OWOX API client. The SDK owns its transport; you cannot replace or inspect it.                                                                                   |
+| `ctx.openExternal(url)`                                  | Ask the host to open an external https URL in a new tab. The sandbox denies you both navigation and popups, so the host opens it, not you.                       |
+| `ctx.navigate(path)`                                     | Ask the host to go to a page inside OWOX — `/ui/${ctx.projectId}/data-marts/${id}` — in place of your frame. Anything resolving off the app's origin is refused. |
```
max-voloshyn: Should we namespace `openExternal` and `navigate` under `ui` ?

docs/plugins/authoring-guide.md
```diff
@@ -0,0 +1,163 @@
...
+| `ctx.setHeight(px)`                                      | Report content height so the host can size your frame.                                                                                                           |
```
max-voloshyn: It seems a redundant detail, may we avoid it in sdk contract?

docs/plugins/authoring-guide.md
```diff
@@ -0,0 +1,163 @@
...
+| `ctx.signal`                                             | Aborts when the host tears your plugin down.                                                                                                                     |
+| `ctx.member`, `ctx.projectId`, `ctx.theme`, `ctx.locale` | Display context. No tokens.                                                                                                                                      |
```
max-voloshyn: Does `ctx.member` mean `ctx.userId` ?

docs/plugins/authoring-guide.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Publishing
+
+Any project member can publish a plugin for themselves, and a Project Admin can publish it
+to their whole project — both from the OWOX web app. Publishing to an entire deployment is
+done with `owox-ctl` by an allowlisted publisher key:
+
+```bash
+owox-ctl plugins publish OWOX/example-plugin --scope deployment --all-projects
+owox-ctl plugins publish OWOX/example-plugin --scope deployment --project-id p1 --project-id p2
```
max-voloshyn: I think we shouldn't promote deployment level publishing in general docs because it's mostly cloud thing. General use case is publishing with project / member scope

amarchenk0 (approved):

apps/backend/src/plugin-host/config/plugin-host.config.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * Origin a plugin vendor may name in `frame-ancestors` for us to accept it.
+   *
+   * Undefined means we cannot know our own deployment origin, so a vendor allowlisting
+   * a specific host is rejected rather than wrongly accepted.
+   */
+  get publicOrigin(): string | undefined {
+    const raw = this.trimmed('PLUGIN_HOST_PUBLIC_ORIGIN');
```
max-voloshyn: Could we reuse the existing `PUBLIC_ORIGIN` here instead of introducing `PLUGIN_HOST_PUBLIC_ORIGIN`? The runtime host is served from `window.location.origin`, while `PublicOriginService` already represents the deployment’s public application origin. Having two values for the same origin can drift and make the publication-time `frame-ancestors` check disagree with the actual host.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/config/env-validation.config.ts
```diff
@@ -27,6 +27,18 @@ const configSchema = z
 
     MAX_CONNECTOR_RUNS_PER_PROJECT: z.coerce.number().int().min(1).max(1000).default(3),
     MAX_REPORT_RUNS_PER_PROJECT: z.coerce.number().int().min(1).max(1000).default(1000),
+
+    // Plugin host. GITHUB_* and PLUGIN_HOST_PUBLIC_ORIGIN stay unvalidated pass-through
+    // strings: they are optional and mode-dependent, and PluginHostConfigService already
+    // treats a blank or malformed value as absent.
+    PLUGIN_HOST_SYNC_MIN_INTERVAL_SEC: z.coerce.number().int().min(0).max(86_400).default(300),
+    PLUGIN_HOST_MAX_RELEASE_PAGES: z.coerce.number().int().min(1).max(10).default(1),
```
max-voloshyn: Does this cap need to be deployment configuration? It controls how far back an implementation detail—the GitHub release scan—looks, rather than deployment-specific behavior. Unless operators have a concrete reason to tune it, could we keep it as an internal constant and avoid expanding the public environment-variable surface? If configurability is required, it would help to document when an operator should change it.

🤖 Codex (GPT-5.6 Sol High)

docs/plugins/authoring-guide.md
romandubovyi: Fixed in `ef94067a8`. You were right that it read as the definition — it was the first sentence of the document.

A plugin is now defined by the contract that does not change: a page in the sandbox, running with the authority of the member who installed it. Delivery moved into its own paragraph in the present tense, so if OWOX Data Marts ever hosts plugins itself, only that paragraph changes.

That paragraph also earns its place: it explains why publishing takes a repository and where the https and embeddable requirements come from, which the guide previously asserted further down without a reason.

docs/plugins/authoring-guide.md
romandubovyi: Fixed in `ef94067a8` — and it was worse than placement: `docs/plugins` appeared nowhere in `astro.config.mjs`, so the page built, had a URL, and nothing on the site linked to it. Reachable only by direct link or search.

Added as a `Plugins` group between API and Contributing, exactly where you suggested, using `autogenerate` so later pages need no config change. Verified in the built site: sidebar order is now Sources → API → Plugins → Contributing, and `dist/index.html` links to the guide, which it did not before.

docs/plugins/authoring-guide.md
romandubovyi: Removed entirely in `11543c97f`.

The literal answer is no — the check runs at publish time, server-side, when no browser exists, so the SDK cannot stand in for it. But tracing it further, the endpoint bought nothing that justified asking authors for it.

The descriptor and `plugin.json` were read once, at the same moment, from a publisher declaring intent either way. The descriptor could not track a redeploy any better, since the delivery URL is deliberately unpinned and neither file is read again. Its only real gate was `sdkVersion` against a supported list of exactly `['1']` — so it refused nothing a publisher could legitimately write, while costing them a second file with its own CORS requirements.

Compatibility is already settled where a browser does exist: `isPluginReady` and `isPluginHello` check the protocol version from both ends, so a page built against an SDK the host cannot speak fails to start. That gate stays. When an SDK v2 arrives, a publish-time check comes back with it rather than sitting here refusing nothing.

One consequence worth naming: publishing no longer checks SDK compatibility at all, so an incompatible plugin now surfaces at load time rather than at publish.

docs/plugins/authoring-guide.md
romandubovyi: Done in `c224fcfc0` — `ctx.ui.openExternal` and `ctx.ui.navigate`.

The grouping says something the flat shape did not: both are requests the host is free to refuse, which is a different relationship from `ctx.owox`, where the plugin is the one acting. It also leaves an obvious place for anything else host-mediated later.

The postMessage protocol is unchanged; this is only the shape of the object `connect()` returns.

docs/plugins/authoring-guide.md
romandubovyi: Removed in `c224fcfc0`, and it was more redundant than it looked: the whole channel was inert end to end.

The SDK posted a `resize` message, the bridge routed it, and `PluginRuntimePage` answered with `onResize: () => undefined` while the frame is sized by CSS (`h-full w-full`). So the contract advertised a capability that discarded its input.

`setHeight`, the `resize` message in both protocol copies, the bridge option and the handler are all gone.

docs/plugins/authoring-guide.md
romandubovyi: In effect yes, which is why it is now literally `ctx.userId` (`c224fcfc0`).

The type declared `member: { userId, fullName?, avatar? }`, but the host only ever sent `{ userId: user.id }` — so the struct promised a profile and delivered one field. `fullName` and `avatar` are available from `GET /api/auth/context` to any plugin that wants them, where they cannot go stale as a second copy in the handshake would.

`ctx.locale` went with it: it was hardcoded `'en'` with no i18n behind it, a constant declared as a variable.

One of the same family I left alone, since you did not raise it: `ctx.theme` is read once when the bridge is created, so a plugin does not learn about a theme switch. Happy to make it live or drop it if you would rather it were not there either.

docs/plugins/authoring-guide.md
romandubovyi: Reordered in `ef94067a8`: the examples are `--scope member` and `--scope project` now, with deployment described as the administrator operation it is and why it will not apply to most readers — it needs a key named in `OWOX_DEPLOYMENT_PLUGIN_PUBLISHER_API_KEY_IDS`.

Checking this against the CLI also turned up a plain error: the guide said member and project publishing happens "from the OWOX Data Marts web app", but `owox-ctl` takes all three scopes, and its own examples lead with `--scope member`. Corrected.

One thing I would rather ask than guess: did you mean deployment scope should not be documented in this guide at all? I read "promote" as emphasis and kept it as a reference line, because there is nowhere else it lives — `docs/editions/` does not mention plugins, so removing it here would leave the command undocumented rather than relocated. If you want it out, it is worth deciding where it goes instead.

apps/backend/src/plugin-host/config/plugin-host.config.ts
romandubovyi: Removed it entirely in `480cdb07f` — `publicOrigin` now delegates to `PublicOriginService`.

Your drift argument is the right one, and tracing where the variable came from makes it sharper: the plan recorded the open question as "should we add this so vendors who allowlist our exact origin pass validation? **Yes for cloud** — without it a correctly-configured vendor is rejected, because the validator cannot know the deployment origin."

That premise was wrong. `PublicOriginService.getPublicOrigin()` reads `PUBLIC_ORIGIN`, every deployment guide requires it, and five other call sites already use it. The variable was solving a problem that was already solved — and it had been carried far enough to sit in the notes as a P0 asking someone to go and decide what the cloud origin is.

Nothing regresses in either direction:

- `PUBLIC_ORIGIN` set (the normal case): a vendor who correctly names your origin is now **accepted**, where before every existing deployment rejected them, since the new variable was unset everywhere.
- nothing set: comparison falls back to `http://localhost:PORT`, which no real vendor names, so the refusal is the same as today.

The deployment guide also claimed "the server has no way to know what that origin is", which was untrue; that paragraph now explains that the origin compared against is `PUBLIC_ORIGIN` and why a plugin-specific copy could only drift from it.

The config specs construct a real `PublicOriginService` rather than a stub — what is worth testing here is what the delegation produces, and a stub would only assert my own fixture back at me.

apps/backend/src/config/env-validation.config.ts
romandubovyi: Agreed, and done in `488313248` — it is an internal constant now.

The same file already made this distinction: `MAX_REDIRECT_HOPS` sits a few lines up as a constant marked "not configurable". The release page cap was simply on the wrong side of that line.

To your "unless operators have a concrete reason": I looked for one and could not construct it. A sync walks releases newest first and stops at the first that validates, so a second page can only hold versions that could never become current. Wanting one would mean the newest hundred releases are all drafts, prereleases or broken manifests — the publisher's problem to fix, not something a deployment should tune around.

The two neighbours stay configurable, since both do have scenarios: `PLUGIN_HOST_SYNC_MIN_INTERVAL_SEC` is pressure against GitHub's rate limit, which differs between cloud and self-managed, and `PLUGIN_HOST_REMOTE_PROBE_TIMEOUT_MS` follows network conditions.

Worth flagging plainly: this drops a documented environment variable, so anyone who had set it will find it silently inert. The feature has not shipped, so the practical risk is nil, but it is a contract change rather than a pure refactor.

One test needed inverting rather than editing: it was called "never pages past the configured cap" and set the variable to 2 to prove paging worked. It now asserts what actually happens — a single request.

docs/plugins/authoring-guide.md
max-voloshyn: I'm OK to leave it in current docs with reordering 👍

.changeset/6772-plugin-gallery.md
```diff
@@ -0,0 +1,27 @@
...
+
+**Emergency control.** An allowlisted publisher key can suspend a plugin across the deployment. Suspension blocks opening, installing and restoring while leaving uninstalling and update checking available, so a corrective version can still become current before the plugin is resumed. It changes no publication or installation record.
+
+Plugin authors build against the new `@owox/plugin-sdk`, which owns the host handshake and hands the plugin a working OWOX API client. `owox-ctl` gains `plugins publish`, `unpublish`, `publications list`, `update`, `suspend` and `resume`.
+
+New deployment variables: `OWOX_DEPLOYMENT_PLUGIN_PUBLISHER_API_KEY_IDS`, `GITHUB_TOKEN`, `GITHUB_APP_ID`, `GITHUB_APP_SLUG`, `GITHUB_APP_PRIVATE_KEY`, `GITHUB_API_BASE_URL`, `PLUGIN_HOST_PUBLIC_ORIGIN`, `PLUGIN_HOST_SYNC_MIN_INTERVAL_SEC`, `PLUGIN_HOST_MAX_RELEASE_PAGES`, `PLUGIN_HOST_REMOTE_PROBE_TIMEOUT_MS`.
```
max-voloshyn: This line should be actualized

romandubovyi: Fixed in `647c97468`. Both `PLUGIN_HOST_PUBLIC_ORIGIN` and `PLUGIN_HOST_MAX_RELEASE_PAGES` were removed by the two changes above, so the line now lists the eight that remain and says the `frame-ancestors` origin comes from `PUBLIC_ORIGIN`.

Checked the rest of the note for the same staleness — the `@owox/api-client` paragraph still holds.

PR #1495, opened 2026-08-03
feat(data-marts): allow changing the input source of a data mart

laskevych:
## Problem

The Input Source (definition) type of a Data Mart was frozen the moment it was first saved. Switching a table-based Data Mart to a SQL query, or repointing a view at a real table once that table exists, was impossible — the only path was to rebuild the Data Mart from scratch and recreate every relationship and report attached to it. That cost is prohibitive as soon as you stop starting from a blank slate.

Two guards enforced this, one per side:

- backend rejected any type change with `BusinessViolationException('DataMart already has definition')`
- frontend rendered the type selector only while the Data Mart had no type yet

Changing the definition *within* the same type already worked.

## What changed

The Input Source type of an existing Data Mart can now be changed, within the same storage, across `SQL`, `TABLE`, `VIEW` and `TABLE_PATTERN`.

The Data Mart keeps its id through the change, so what is built on top of it survives untouched:

- relationships in both directions (`DataMartRelationship` keys on Data Mart ids and field names)
- reports and their column configuration
- field-level metadata — alias, description, aggregation role, primary key, hidden-for-reporting
- data quality rules, scheduled triggers, run history

Output-schema drift is handled by the machinery that already exists: the per-storage schema mergers mark fields missing from the new source as `DISCONNECTED`, flag fields whose type changed, and preserve user-set metadata. That is the same behaviour a same-type definition edit produces today, so no new machinery was added for it.

### Safety around the change

- The type change is *staged* in the UI — picking another type swaps the form and shows a warning banner, but nothing is written until Save.
- Save opens a confirmation dialog naming the transition and stating how many relationships and reports depend on this Data Mart, so the blast radius is visible before committing.
- Switching to a `TABLE`, `VIEW` or `TABLE_PATTERN` validates the new target against the storage **before** it is persisted, so an unreachable source is refused rather than saved and left to fail later at schema actualization. `SQL` targets are exempt, mirroring `PublishDataMartService`, which already defers SQL to the editor's dry run rather than repeating the round trip server-side.
- The use case is now `@Transactional()` — a failed switch can no longer leave a Data Mart with a new type and a stale definition.
- After saving, the schema is re-actualized automatically.
- A published Data Mart stays published.

### Schema refresh now reports its result

Actualizing the output schema used to toast a bare "Output schema actualized". It now reports what it found — how many fields are connected, and how many came back disconnected or with a type mismatch. This is how a user learns which fields the new source stopped providing, and it applies to every schema refresh, not only to a type change.

### Out of scope

**Connector-backed Data Marts are untouched in both directions.** A connector owns a write target, stored secrets, an incremental cursor and its own run triggers, none of which can be handed over to — or picked up from — a plain source. Their type stays fixed, their editing experience is unchanged, and no other type can be switched to a connector. No teardown code was written, so there is no half-safe cleanup path in the codebase.

Legacy Google BigQuery storages keep their existing SQL-only rule; the selector is hidden there.

Moving a Data Mart to a different storage remains out of scope.

## Notes

- No migration: the database schema is unchanged.
- No storage capability map was introduced. Every non-legacy storage's query builder already handles all four static definition types, so the only real restriction is the legacy BigQuery rule that was already enforced.
- Relationship counts for the dialog reuse the existing storage-wide edge endpoint, which is the only read that exposes inbound edges; no new endpoint was needed.

## Testing

- Backend unit: every transition among the four static types, both connector rejections, validation refusing an unresolvable target without saving, SQL staying exempt, same-type edits not being re-validated, a regression lock that a same-type connector update still runs its secrets merge/extract/orphan-cleanup flow unchanged, the legacy BigQuery rule, and event semantics.
- Frontend component: selector visibility for existing / connector / legacy Data Marts, connector excluded as a switch target, staged-change warning, discard restoring the saved type, confirmation gating the write, and same-type edits saving without a dialog. Plus unit coverage of the schema summary, including nested fields and singular/plural wording.
- E2E: a Data Mart with inbound and outbound relationships plus a report keeps all of them across a type change; a switch to a target the storage cannot resolve is refused and the Data Mart is left on its original definition; a connector Data Mart is rejected and stays on `CONNECTOR`.

Full `apps/web` data-marts suite and `apps/backend` use-case suite pass locally. E2E was not run locally and is left to CI.


apps/web/e2e/specs/datamart-data-setup.spec.ts
```diff
@@ -290,3 +290,92 @@ test.describe('Data Setup - Connector Definition', () => {
...
+    await apiHelpers.setDefinition(neighbour.id);
+
+    // Outbound: subject -> neighbour. Inbound: neighbour -> subject.
+    await apiHelpers.createRelationship(subject.id, neighbour.id, 'neighbour', [
+      { sourceFieldName: 'id', targetFieldName: 'id' },
+    ]);
+    await apiHelpers.createRelationship(neighbour.id, subject.id, 'subject', [
+      { sourceFieldName: 'id', targetFieldName: 'id' },
+    ]);
+    const { reportId } = await apiHelpers.setupDestinationWithReport(subject.id);
```
amarchenk0: This fixture never publishes `subject`, but `CreateReportService` rejects reports unless the Data Mart is `PUBLISHED`. CI fails here on both attempts, before the TABLE → SQL request at line 326, so this test never exercises the feature it claims to cover. Please build a publishable fixture (or use an allowed report setup path), execute the transition, and keep the E2E suite green.

apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts
```diff
@@ -129,6 +137,15 @@ export class UpdateDataMartDefinitionService {
       dataMart.definition = command.definition;
     }
 
+    // A type change repoints the Data Mart at a different kind of source, so the new definition is
+    // checked against the storage before it lands. Same-type edits keep their existing behaviour:
+    // they are validated on publish and on schema actualization, not on every save.
+    // SQL is exempt for the same reason publishing exempts it — a SQL definition is dry-run from
+    // the editor, and re-running it here would duplicate that round trip on every switch.
+    if (definitionTypeChanged && command.definitionType !== DataMartDefinitionType.SQL) {
```
amarchenk0: The SQL exception makes a TABLE → SQL transition an unvalidated write. The UI dry run is advisory: its state is not used to gate Save, and API callers bypass it altogether. A non-empty invalid query can therefore replace a working table definition. Validate a SQL target server-side (or require a server-verifiable dry run) and add a regression test that a TABLE → invalid SQL transition persists neither the type nor the definition.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
```diff
@@ -66,11 +71,36 @@ export function DataMartDefinitionSettings({
   const {
     definition: initialDefinition,
     id: dataMartId,
+    reportsCount,
```
amarchenk0: `reportsCount` is always `0` here: both the detail and update paths call `DataMartMapper.toDomainDto()` without counters, which defaults the count to zero. The new confirmation can therefore say “Nothing else depends on this Data Mart yet” even when reports exist. Fetch the real count at an authorized backend boundary (or return it in a dedicated impact response) and add an integration test with a non-zero report count.

apps/web/src/features/data-marts/edit/model/context/DataMartContext.tsx
```diff
@@ -593,7 +597,9 @@ export function DataMartProvider({ children }: DataMartProviderProps) {
       const response = await dataMartService.actualizeDataMartSchema(id);
       const dataMart = await mapDataMartFromDto(response);
       dispatch({ type: 'ACTUALIZE_DATA_MART_SCHEMA_SUCCESS', payload: dataMart });
-      toast.success('Output schema actualized');
+      // Report what the refreshed schema looks like, not just that it ran. After an input source
+      // change this is how the user learns which fields the new source no longer provides.
+      toast.success(describeSchemaFieldSummary(summarizeSchemaFields(dataMart.schema)));
```
amarchenk0, line 608: The new summary is not shown on the user-facing path. Input Source uses `runSchemaActualization`, which goes through `useSchemaActualizeTrigger` and still emits the fixed “Output schema actualized” toast; `actualizeDataMartSchema`, where this summary was added, is never invoked by the UI. Wire the summary into the trigger success path and add a UI/integration test for the actual toast.

amarchenk0 (changes requested):
Please address the three blocking issues in the inline comments. The fourth comment is non-blocking.

apps/backend/src/data-marts/use-cases/update-data-mart-definition.service.ts
laskevych: Confirmed and fixed in bc69251b3 — the exemption is gone, every type change is validated before it is persisted.

You are right that the dry run is advisory: `DataMartDefinitionSettings` discards the validator state (`const [, setSqlValidationState]`) and Save is gated only on the zod check, so nothing on the server side was standing between an invalid query and a working table definition.

I had modelled the exemption on `PublishDataMartService`, which skips SQL — but publishing an already-saved definition and replacing one are not the same risk, so copying that rule was wrong here.

Regression test added: TABLE → invalid SQL asserts the validator was called, the error propagates, and `save` was never reached. A backend e2e covers the same path end-to-end and asserts the Data Mart still holds its original TABLE definition afterwards.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
laskevych: Confirmed and fixed in bc69251b3. `GetDataMartService` passes `undefined` for counters, so `reportsCount` on the detail response is always 0 — the dialog could indeed have said "Nothing else depends on this Data Mart yet" while reports existed.

Went with the dedicated impact response you suggested: `GET /data-marts/:id/input-source-change-impact`, gated on EDIT (the action the answer feeds into), returning `outboundRelationshipsCount`, `inboundRelationshipsCount` and `reportsCount`. `reportsCount` is counted in `ReportService`; relationships are counted in both directions in `DataMartRelationshipService`.

Two related cleanups fell out of this:
- reverted the `reportsCount` plumbing I had added to the web edit model — it would have been a permanently-zero field.
- the hook no longer fetches the storage-wide edge list to derive relationship counts client-side, which was both heavier and wider than needed.

On a failed read the dialog now shows no counts at all rather than zeros, since zeros read as "nothing depends on this".

Integration coverage with a non-zero count: the backend e2e asserts `{outbound: 1, inbound: 1, reports: 1}` on a Data Mart that has a real report and relationships in both directions.

apps/web/src/features/data-marts/edit/model/context/DataMartContext.tsx
laskevych: Confirmed and fixed in bc69251b3. You traced it exactly: the Input Source flow goes `runSchemaActualization` → `useSchemaActualizeTrigger`, which emitted the fixed string, while I had put the summary in `actualizeDataMartSchema` — a context method the UI never calls for this flow. The summary was dead on the user-facing path.

`onSuccess` now resolves with the refreshed Data Mart (`getDataMart` returns it instead of void), and the trigger derives the toast from that schema. The fallback string stays for callers that return nothing.

Tests on the actual toast path in `useSchemaActualizeTrigger.test.ts`:
- a schema with 3 fields, 1 disconnected → `Output schema actualized: 3 fields, 1 disconnected`
- `onSuccess` returning nothing → the plain `Output schema actualized`

apps/web/e2e/specs/datamart-data-setup.spec.ts
laskevych: Confirmed and fixed in bc69251b3. `CreateReportService` requires PUBLISHED and the fixture never published `subject`, so the test failed at report creation and never reached the transition. I flagged that I had not run the Playwright suite locally, but that is not an excuse for shipping a test that cannot pass — thanks for catching it.

Rather than patch the fixture, I moved the scenario to where it can actually run. Once SQL targets are validated too (your other comment), *no* transition can succeed in the Playwright harness, which has no warehouse to validate against — a publishable fixture alone would not have been enough.

New `apps/backend/test/data-mart-input-source-change.e2e-spec.ts` follows the existing pattern of stubbing `DataMartDefinitionValidatorFacade` (as the output-controls e2e specs do) and drives the real HTTP API: publish → relationships in both directions → destination + report → TABLE → SQL transition. It asserts both relationship graphs, the report still pointing at the Data Mart, the status still PUBLISHED, and the impact counts. Also covers the rejection paths and the rollback. Passing locally: 3/3.

The Playwright suite keeps only what it can honestly cover without a warehouse — the two rejection paths — and a comment explains where the successful transition is tested and why. Unused helpers I had added there are removed.

laskevych:
All four addressed in bc69251b3 — three were real defects, and the reviews were correct on each.

**What changed**

1. **SQL targets are validated server-side.** The exemption I copied from `PublishDataMartService` was wrong here: the editor dry run never gated Save and API callers skip it, so an invalid query could replace a working table definition. Every type change is now checked before it is persisted.
2. **Dependency counts come from `GET /data-marts/:id/input-source-change-impact`** (EDIT-gated), counting inbound and outbound relationships and reports server-side. The old path read `reportsCount` off the detail response, which is always 0. The permanently-zero field I had added to the web model is reverted, and the client no longer derives relationship counts from the storage-wide edge list.
3. **The schema summary is emitted from `useSchemaActualizeTrigger`**, the path the UI actually uses; `onSuccess` now resolves with the refreshed Data Mart so the toast can describe it.
4. **The survival scenario moved to a backend e2e** (`data-mart-input-source-change.e2e-spec.ts`) with the validator stubbed, as the output-controls specs do. It publishes properly and drives a real transition over HTTP. This was not just a fixture fix: with SQL validated too, no transition can succeed in the Playwright harness at all, since it has no warehouse. Playwright keeps the two rejection paths.

**Verification**

- backend unit: 32 in the definition use-case spec, including TABLE → invalid SQL persisting neither type nor definition
- backend e2e: 3/3 — relationships both directions, report, PUBLISHED status and impact counts `{1, 1, 1}` all survive the transition
- web: 892 data-marts tests, including the two new toast-path assertions
- lint and typecheck clean on both apps

Two full-suite backend specs (`project-list-sql-dialects`, `looker-studio-connector-api-data`) time out under parallelism on my machine and pass in isolation; unrelated to this change.

apps/backend/src/data-marts/use-cases/get-data-mart-input-source-change-impact.service.ts
```diff
@@ -0,0 +1,56 @@
...
+        command.projectId
+      );
+      if (!canEdit) {
+        throw new ForbiddenException('You do not have permission to edit this DataMart');
+      }
+    }
+
+    const [outbound, inboundSourceIds, reportsCount] = await Promise.all([
+      this.relationshipService.findBySourceDataMartId(command.id),
+      this.relationshipService.findSourceDataMartIdsByTargetDataMartId(
```
amarchenk0: `findSourceDataMartIdsByTargetDataMartId()` removes duplicates with a `Set`, but its result is presented as the number of inbound relationships. One source can have several relationships to the same target under different aliases—the unique key is source plus alias. The dialog therefore underreports the real dependency count. Count relationship records rather than distinct source Data Marts, and add a case with two aliases.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/useInputSourceChangeImpact.ts
```diff
@@ -0,0 +1,67 @@
...
+    dataMartService
+      .getInputSourceChangeImpact(dataMartId, { signal: abortController.signal })
+      .then(response => {
+        setImpact({
+          outboundRelationships: response.outboundRelationshipsCount,
+          inboundRelationships: response.inboundRelationshipsCount,
+          reports: response.reportsCount,
+        });
+      })
+      .catch(() => {
```
amarchenk0, line 63: When the impact request fails, the hook sets `impact = null`, and the dialog shows “Nothing else depends on this Data Mart yet” once loading ends. An API or network failure is therefore indistinguishable from a zero impact and gives the user false reassurance. Add a distinct error state with neutral copy and preferably a retry; do not present zero dependencies without a successful response.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
```diff
@@ -188,6 +220,26 @@ export function DataMartDefinitionSettings({
...
+      runSubmitGuarded(data);
+    },
+    [isTypeChangeStaged, runSubmitGuarded]
+  );
+
+  const handleConfirmTypeChange = useCallback(() => {
+    if (!pendingTypeChange) return;
+    const data = pendingTypeChange;
+    setPendingTypeChange(null);
+    runSubmitGuarded(data);
```
amarchenk0, line 242: The backend now rejects invalid SQL, but `updateDataMartDefinition()` catches the API error and resolves normally. `onSubmit` therefore still resets the form and actualizes the old Data Mart after `await`. The user loses the entered SQL although nothing was saved. Rethrow the error or return an explicit operation result, and reset/actualize only after success; add a UI test for the rejected transition.

amarchenk0 (changes requested):
Changes requested.

apps/backend/src/data-marts/use-cases/get-data-mart-input-source-change-impact.service.ts
laskevych: Confirmed and fixed in 0ad3d58c4. `findSourceDataMartIdsByTargetDataMartId` exists for search-index invalidation, where distinct sources are the right unit — reusing it here silently changed the question from "how many joins depend on this" to "how many data marts".

Added `countByTargetDataMartId` to `DataMartRelationshipService` (a plain row count by target + project) and the impact use case now uses it. The e2e creates two inbound relationships from the same source under different aliases and asserts `inboundRelationshipsCount: 2`.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/useInputSourceChangeImpact.ts
laskevych: Confirmed and fixed in 0ad3d58c4. My earlier comment even said zeros "read as nothing depends on this" — and then `impact = null` fell through to exactly that copy. The failure state existed in the hook but not in the UI.

The hook now exposes `hasError` and `retry` (an attempt counter that re-arms the fetch effect; aborts no longer count as failures). The dialog has three distinct states: loading, failed — neutral "Couldn't check what depends on this Data Mart" with a try-again link, still allowing the user to proceed — and loaded, which is the only state that may say "Nothing else depends on this Data Mart yet".

Component test covers the failure copy, the absence of the reassuring line, and retry turning a later successful read into real counts.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
laskevych: Confirmed and fixed in 0ad3d58c4. The context's `updateDataMartDefinition` dispatched the error state and resolved normally, so the component's `try` block continued as if the save had succeeded: form reset, schema actualization of an unchanged Data Mart, and the entered SQL gone.

The context now rethrows after dispatching (the pattern `createDataMart` and `deleteDataMart` already follow); `onSubmit` reaches `reset`/actualization only on success. Checked the callers first: the definition form is the only consumer of the context method — the bulk-create path uses the raw service, which already throws — so the rethrow changes no other flow.

Component test: a rejected transition keeps the entered SQL in the field and `runSchemaActualization` is never called.

laskevych:
Second round addressed in 0ad3d58c4 — all three confirmed against the code before fixing.

1. **Inbound count** — now a row count via a new `countByTargetDataMartId` in `DataMartRelationshipService`. The deduplicating method I had reused was built for search-index invalidation, where distinct sources are the right unit; here every source+alias row is its own dependency. E2E asserts inbound `2` for one source joining under two aliases.
2. **Failed impact read** — the hook exposes `hasError`/`retry`; the dialog gained a distinct failure state with neutral copy and a try-again link. "Nothing else depends on this Data Mart yet" is now reachable only from a successful response.
3. **Swallowed save error** — the context rethrows after dispatching, so a rejected transition keeps the user's input and skips schema actualization. Verified the definition form is the only consumer of the context method before changing its contract.

Verification: backend definition spec 32/32, backend e2e 3/3 (with the two-alias case), web data-marts suite 894/894 including the two new component tests (rejected transition preserves input; unknown impact never rendered as zero), lint and typecheck clean on both apps.

amarchenk0 (approved):

PR #1496, opened 2026-08-04
feat(data-marts): export Models canvas as SVG, PNG, JSON, or OKF bundle

laskevych:
## Preview

https://github.com/user-attachments/assets/8b051d59-c188-42b5-b1eb-850dc52328ab

## What

The **Models** canvas gets an **Export** submenu in the **Actions** menu, with four formats:

| Format | File | Notes |
| --- | --- | --- |
| **Image (SVG)** | `<storage>-<date>.svg` | Vector snapshot of the whole visible model — pan/zoom never crops it |
| **Image (PNG)** | `<storage>-<date>.png` | Same capture rasterized at 2× via an offscreen composite |
| **JSON** | `<storage>-<date>.json` | Model graph in the [OWOX Model Canvas](https://model.owox.com/) interchange format, sanitized of storage and data mart identifiers |
| **OKF (Markdown)** | `<storage>-<date>.zip` | One cross-linked Markdown document per data mart (overview, schema with PK/FK notes, joins) plus an `index.md` catalog |

Both images take a theme-aware background layer (dark-theme exports stay readable outside the app) and an OWOX watermark in the bottom-right corner.

The PR also ships feedback-driven canvas polish:

- The **Actions** menu moves from a canvas-corner overlay into the toolbar row — consistent with the Data Marts list, and it no longer covers canvas content.
- The **field count** joins the status icons row instead of occupying its own line, so every card is one row shorter.
- The Object labels settings list drops its decorative glyphs.

## How

- The export targets exactly the **filtered/visible** model — the same set the other Actions cover — and the submenu appears only on the canvas surface, following the established optional-prop pattern in `DataMartBulkActions`.
- Image framing uses `getNodesBounds` with a transform override on the React Flow viewport, so the whole model renders at 1:1 regardless of the user's pan/zoom. The background is composited as a separate layer because `html-to-image`'s own `backgroundColor` would paint on the translated viewport and end up offset.
- The canvas exposes the export API through an imperative handle: the Actions menu renders outside the `ReactFlowProvider`, while measured nodes and the viewport element live inside it.
- New dependencies `html-to-image` and `fflate` load in a separate lazy chunk (~31 kB) on first export, keeping the canvas bundle unchanged.
- Docs: a new user-guide page describing the formats and what OKF is.

## Testing

- New unit tests: model-graph mapping + sanitization, OKF serializer (documents, joins, FK notes, index, slug de-duplication), download helper, Export submenu rendering and callbacks; card-layout tests updated for the merged field-count row.
- All existing model-canvas and list tests pass; `tsc -b`, eslint, prettier, markdownlint, and the production build are clean.

### DIFF
--- a/apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
+++ b/apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
@@ -155,7 +155,12 @@ function ModelCanvasViewContent({ onActiveQualityRunChange }: ModelCanvasViewPro
   const canvasStyle = { height: 'calc(100vh - 220px)', minHeight: 480 };
 
   const canvasExportRef = useRef<ModelCanvasExportHandle>(null);
+  // Image capture can take seconds on a large model — swallow repeat clicks
+  // instead of kicking off parallel captures and duplicate downloads.
+  const isExportingRef = useRef(false);
   const handleExport = useCallback((format: DataMartCanvasExportFormat) => {
+    if (isExportingRef.current) return;
+    isExportingRef.current = true;
     void (async () => {
       try {
         await canvasExportRef.current?.exportCanvas(format);
@@ -168,6 +173,8 @@ function ModelCanvasViewContent({ onActiveQualityRunChange }: ModelCanvasViewPro
       } catch (caught) {
         console.error('Canvas export failed:', caught);
         toast.error("Couldn't export the model — please try again.");
+      } finally {
+        isExportingRef.current = false;
       }
     })();
   }, []);
--- a/apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
+++ b/apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
@@ -163,7 +163,15 @@ function ModelCanvasViewContent({ onActiveQualityRunChange }: ModelCanvasViewPro
     isExportingRef.current = true;
     void (async () => {
       try {
-        await canvasExportRef.current?.exportCanvas(format);
+        // The handle registers once the lazy canvas chunk mounts, and the
+        // export itself declines until the first layout pass — in both windows
+        // nothing downloads, so say so instead of silently succeeding, and
+        // record analytics only for real downloads.
+        const exported = (await canvasExportRef.current?.exportCanvas(format)) ?? false;
+        if (!exported) {
+          toast('The canvas is still loading — please try again in a moment.');
+          return;
+        }
         trackEvent({
           event: 'model_canvas_exported',
           category: 'DataMart',
--- a/apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
+++ b/apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
@@ -247,6 +247,23 @@ function ModelCanvasViewContent({ onActiveQualityRunChange }: ModelCanvasViewPro
         onRelChange={filters.setRel}
         searchQuery={filters.searchQuery}
         onSearchChange={filters.setSearchQuery}
+        actions={
+          <DataMartBulkActions
+            onExport={handleExport}
+            onCheckDataLastUpdated={() => {
+              // Meeting decision: the check covers what the user actually sees — the same
+              // filtered set the other bulk actions target.
+              void refreshDataLastUpdated(bulkActionDataMarts.map(dataMart => dataMart.id));
+            }}
+            isCheckingDataLastUpdated={isRefreshingDataLastUpdated}
+            dataMarts={bulkActionDataMarts}
+            projectId={projectId ?? ''}
+            deleteDataMart={deleteDataMart}
+            publishDataMart={publishDataMart}
+            onCompleted={refreshCanvas}
+            targetScope='canvas'
+          />
+        }
       />
       {storageLoadError ? (
         <CanvasMessage role='alert'>
@@ -302,23 +319,6 @@ function ModelCanvasViewContent({ onActiveQualityRunChange }: ModelCanvasViewPro
             isCheckingDataLastUpdated={isRefreshingDataLastUpdated}
             storageTitle={dataStorages.find(storage => storage.id === filters.storageId)?.title}
             exportApiRef={canvasExportRef}
-            topLeftControls={
-              <DataMartBulkActions
-                onExport={handleExport}
-                onCheckDataLastUpdated={() => {
-                  // Meeting decision: the check covers what the user actually sees — the same
-                  // filtered set the other bulk actions target.
-                  void refreshDataLastUpdated(filtered.nodes.map(node => node.id));
-                }}
-                isCheckingDataLastUpdated={isRefreshingDataLastUpdated}
-                dataMarts={bulkActionDataMarts}
-                projectId={projectId ?? ''}
-                deleteDataMart={deleteDataMart}
-                publishDataMart={publishDataMart}
-                onCompleted={refreshCanvas}
-                targetScope='canvas'
-              />
-            }
             style={canvasStyle}
           />
         </Suspense>
--- a/apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
+++ b/apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
@@ -14,7 +14,11 @@ import { useModelCanvasFilters } from '../model/use-model-canvas-filters';
 import { ModelCanvasToolbar } from './ModelCanvasToolbar';
 import { dataQualityService } from '../../data-quality/api/data-quality.service';
 import { dataMartService } from '../../shared';
-import { DataMartBulkActions } from '../../shared/components/DataMartBulkActions';
+import {
+  DataMartBulkActions,
+  type DataMartCanvasExportFormat,
+} from '../../shared/components/DataMartBulkActions';
+import type { ModelCanvasExportHandle } from '../export';
 import { trackEvent } from '../../../../utils/data-layer';
 import { isDataQualityActivityState } from '../../shared/components/RunActivityIndicator';
 import { useDataQualitySummaries } from '../../data-quality/model/use-data-quality-workspace';
@@ -150,6 +154,24 @@ function ModelCanvasViewContent({ onActiveQualityRunChange }: ModelCanvasViewPro
 
   const canvasStyle = { height: 'calc(100vh - 220px)', minHeight: 480 };
 
+  const canvasExportRef = useRef<ModelCanvasExportHandle>(null);
+  const handleExport = useCallback((format: DataMartCanvasExportFormat) => {
+    void (async () => {
+      try {
+        await canvasExportRef.current?.exportCanvas(format);
+        trackEvent({
+          event: 'model_canvas_exported',
+          category: 'DataMart',
+          action: 'CanvasExport',
+          label: format,
+        });
+      } catch (caught) {
+        console.error('Canvas export failed:', caught);
+        toast.error("Couldn't export the model — please try again.");
+      }
+    })();
+  }, []);
+
   const runQuality = useCallback(
     async (dataMartId: string) => {
       try {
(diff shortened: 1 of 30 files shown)


apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
```diff
@@ -0,0 +1,153 @@
...
+  const files: Record<string, string> = {};
+  for (const node of graph.nodes) {
+    files[`${folder}/${slugByKey.get(node.key) ?? node.key}.md`] = renderNode(
+      node,
+      graph,
+      slugByKey
+    );
+  }
+
+  const rows = graph.nodes
```
prokopetsv, line 51: 🟡 **Non-Blocker** — user-controlled strings go into Markdown structure unescaped. A Data Mart titled `Orders | EU` breaks this index table row and its link (`[title](./slug.md)` also breaks on `]`/`)` in titles); the same applies to aliases and types in the schema table below, and raw descriptions are injected straight into the document body. Titles with pipes are plausible, and one such mart corrupts the table for the whole bundle. Fix: escape `|` → `\|` in table cells and brackets in link text. Related: line 112 emits `# Schema` (H1) while siblings are `## Overview`/`## Joins` — a second H1 per document, worth fixing to `##` in the same pass since the format claims model.owox.com compatibility.

apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
```diff
@@ -0,0 +1,153 @@
...
+    type: 'OWOX Data Mart',
+    title: node.title,
+    description: node.description === '' ? undefined : node.description,
+    tags: ['owox', node.inputSource.toLowerCase()],
+  });
+
+  const overview = [
+    '## Overview',
+    '',
+    `- **ID:** \`${node.owoxId ?? '—'}\``,
```
prokopetsv: 🟡 **Non-Blocker** — the sanitization postures of the two shareable formats contradict each other. JSON deliberately strips `owoxId` and the storage label ("the same rules model.owox.com applies to its share links"), but the OKF bundle — explicitly pitched in the docs as context to hand to AI assistants — embeds every Data Mart UUID here plus the storage title in each Overview and the index. Not credentials, but the same identifiers one format treats as unsafe to share, the other ships. Either drop the ID line from OKF or document why OKF intentionally keeps it.

apps/web/src/features/data-marts/model-canvas/export/export-image.ts
```diff
@@ -0,0 +1,144 @@
...
+    style,
+    skipFonts: true,
+    pixelRatio: PNG_PIXEL_RATIO,
+  });
+
+  // Composite onto a fresh canvas: background fill first (see the SVG note on
+  // why html-to-image's own backgroundColor option cannot be used), then the
+  // capture, then the watermark rendered from a standalone SVG.
+  const output = document.createElement('canvas');
+  output.width = width * PNG_PIXEL_RATIO;
```
prokopetsv: 🟡 **Non-Blocker** — no guard on output canvas size. The canvas pages up to 1000 nodes; a wide layout at 2× can exceed browser canvas limits (Chrome fails → `toBlob` null → the thrown error lands in the view's toast, acceptable; Safari instead silently succeeds with a **blank or truncated** PNG — the user gets a broken file and no error). Cheap fix: clamp the effective pixel ratio by a max-area budget, e.g. `const ratio = Math.min(PNG_PIXEL_RATIO, Math.sqrt(MAX_CANVAS_AREA / (width * height)))`, so huge models degrade to 1× instead of failing.

docs/getting-started/setup-guide/models-canvas-export.md
```diff
@@ -0,0 +1,33 @@
+# Export the Models Canvas
```
prokopetsv, line 1: 🟡 **Non-Blocker** — this page is not registered in the docs-site sidebar: `apps/docs/astro.config.mjs` lists every sibling setup-guide page explicitly (`joinable-data-marts`, `data-last-updated`, …) and `models-canvas-export` is absent, so the page won't appear in the docs navigation. Add `'docs/getting-started/setup-guide/models-canvas-export'` to that sidebar list (this is a functional gap verified against the config, not a style point).

prokopetsv (commented):
Reviewed at head `23a0c7d`. No existing comments to account for; no blockers found. The architecture claims check out under verification: the lazy-chunk claim holds (only type-only static imports cross into the canvas chunk; the real work loads via dynamic `import('../export')`), zip entry paths are traversal-safe by construction (slugify strips to `[a-z0-9-]`), the empty-canvas "menu is not shown" docs claim matches the code (`dataMarts.length === 0 → return null`), export errors surface as a toast with analytics fired only on success, and the enum-keyed `INPUT_SOURCE_BY_DEFINITION_TYPE` record makes a future definition type a compile error rather than a silent `undefined`. Test coverage is broad and tests real behavior (slug de-dup, bidirectional join flipping, sanitization shape). Four 🟡 items inline: OKF markdown breaks on `|`/brackets in user titles, the OKF/JSON sanitization postures contradict each other, PNG export has no canvas-size guard, and the new docs page is missing from the docs-site sidebar registration.

apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
laskevych: Thanks — addressed across two commits:

- **Pipes in table cells**: fixed in 55a5ae39c — every schema-table cell and the index row (title, storage label) goes through an `escapeTableCell` pass, with tests.
- **Brackets in link text**: fixed in 09ce98abb — `escapeLinkText` escapes `[`/`]` in the index links, FK notes, and join lists, so a title like `Customers [EU]` can no longer terminate its own link; covered by a test. Parens in titles only matter inside the link *target*, and targets are slugified to `[a-z0-9-]`, so they stay safe by construction.
- **Raw descriptions in the body**: intentional — the document body is Markdown and the description is rendered as prose there (same as the model.owox.com serializer); only structural contexts (tables, link text) are escaped.
- **`# Schema` as H1**: deliberately kept. The OKF spec itself (okf-format.md on model.owox.com) canonically uses `# Schema` in all its examples, and its parser accepts both depths (`/^##?\s+Schema/i`). Since the format claims compatibility with that spec, mirroring its H1 is the safer choice — happy to revisit if the spec changes.

apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
laskevych: Agreed, and aligned in 55a5ae39c: `owoxId` is gone from the graph shape entirely, so the OKF Overview no longer emits the ID line — both shareable formats now carry zero data mart identifiers (a test asserts this for the bundle). Rationale is in the code comment: these exports feed documentation and LLM context, where opaque UUIDs add nothing — the semantic value is in titles, schemas and joins.

The storage *title* intentionally stays in OKF: it's a human-readable label in a human-readable format and useful context for a reader/assistant, while JSON — the stricter machine format — keeps dropping it.

apps/web/src/features/data-marts/model-canvas/export/export-image.ts
laskevych: Fixed in 55a5ae39c with exactly this shape: `pngPixelRatio()` clamps the effective ratio by `Math.sqrt(MAX_PNG_AREA / (width * height))` with a floor of 1×, where `MAX_PNG_AREA` is an 8192×8192 equivalent — comfortably inside the strictest (Safari) limits, so huge models degrade to lower resolution instead of a null `toBlob`/blank file. The watermark and output dimensions scale with the effective ratio. Good catch on the Safari silent-truncation mode — that's the case the clamp really protects.

docs/getting-started/setup-guide/models-canvas-export.md
laskevych: Fixed in 09ce98abb — added `'docs/getting-started/setup-guide/models-canvas-export'` to the sidebar list (right after `joinable-data-marts`, keeping the data-marts pages grouped). Verified with a local docs build: the page renders and appears in the Setup Guide navigation, and the link validator reports no issues for it. Thanks for checking this against the config — easy one to miss.

apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
```diff
@@ -150,6 +154,31 @@ function ModelCanvasViewContent({ onActiveQualityRunChange }: ModelCanvasViewPro
...
+  const canvasExportRef = useRef<ModelCanvasExportHandle>(null);
+  // Image capture can take seconds on a large model — swallow repeat clicks
+  // instead of kicking off parallel captures and duplicate downloads.
+  const isExportingRef = useRef(false);
+  const handleExport = useCallback((format: DataMartCanvasExportFormat) => {
+    if (isExportingRef.current) return;
+    isExportingRef.current = true;
+    void (async () => {
+      try {
+        await canvasExportRef.current?.exportCanvas(format);
```
amarchenk0: The Actions menu renders before the lazy `<ModelCanvas>` inside `Suspense`. While the canvas chunk is still loading, `canvasExportRef.current` is `null`, and optional chaining turns export into a successful no-op: no file is downloaded, no toast is shown, yet `model_canvas_exported` is recorded. There is also a short window after mount and before the effect populates `flowNodes`; JSON/OKF then receive an empty position map and serialize every node at `(0, 0)`. Do not allow export until the handle and measured nodes are ready; report an explicit unavailable state and emit analytics only after a real download. Add a test with a deferred lazy canvas.

docs/getting-started/setup-guide/models-canvas-export.md
```diff
@@ -0,0 +1,33 @@
...
+
+The **Models** canvas shows your Data Marts as an entity-relationship diagram: every Data Mart is a card, every join is an arrow. The canvas can be exported — as an image for a presentation or documentation, or as a machine-readable model you can archive, diff, or open in other tools.
+
+## Where to find it
+
+Open **Data Marts → Models**, pick a storage, and open the **Actions** menu in the toolbar above the canvas. The **Export** submenu lists the available formats.
+
+The export always covers **what the canvas currently shows**: the same filtered set of Data Marts the other Actions target. Narrow the canvas with the storage, status, relationship, or search filters first if you want to export a subset. When the canvas is empty, there is nothing to export and the menu is not shown.
```
amarchenk0: The documentation promises that search narrows the exported subset. In reality, `searchQuery` is only used for dimming/highlighting in `computeCanvasHighlight`; `filtered` and `bulkActionDataMarts` are filtered only by status and relationships. After searching for `Orders`, the export still contains the other Data Marts. `ModelCanvasView.test.tsx` even codifies this by expecting the Actions count not to narrow by search. Either remove search from the instructions and describe it as highlighting, or actually filter the export; add a regression test for the chosen contract.

amarchenk0 (changes requested):
Changes requested.

apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
laskevych: Good catch on both windows — fixed in 347145eb1:

- `exportModelCanvas` now returns whether a download actually happened, and declines (`false`) while the canvas is not ready: no nodes, no measured `flowNodes` yet (the (0,0)-positions window), or no viewport element for the image formats.
- The view treats `ref === null` (lazy chunk still loading) and a declined export the same way: an explicit toast — "The canvas is still loading — please try again in a moment." — and **no** `model_canvas_exported` event. Analytics fires only after a real download.
- Tests: three readiness unit tests on `exportModelCanvas` (declines with empty `flowNodes` / missing viewport, downloads once measured nodes exist), plus a view test where the mocked canvas never registers the handle — the same observable state as the deferred lazy chunk — asserting the toast and the absence of the analytics event.

docs/getting-started/setup-guide/models-canvas-export.md
laskevych: You're right — the docs overpromised. Fixed in 347145eb1 by correcting the documentation rather than changing the behavior: the export deliberately shares its target set (`bulkActionDataMarts`) with the other Actions, and search narrows none of them — making search narrow only the export would let the same menu Publish/Delete one set and export a different one. The guide now says: "The search box only highlights matching Data Marts — it does not narrow the export, just as it does not narrow the other Actions."

For the regression contract: the existing `counts the Data Marts left by canvas filters without narrowing the count by search` test is exactly that shared-set contract (one `bulkActionDataMarts` feeds both the badge and the export scope); I added a comment there pinning export to it so a future change to either surface trips the test's intent.

amarchenk0 (approved):

PR #1498, opened 2026-08-05
feat(data-marts): include the data mart definition in the OKF export

laskevych:
## What

Every document in the OKF bundle exported from the Models canvas gains a **Definition** section:

- **TABLE / VIEW** — the fully qualified name (`project.dataset.table`)
- **TABLE PATTERN** — the pattern
- **SQL** — the query text
- **CONNECTOR** — deliberately excluded: its source config can carry sensitive parameters, unlike a plain table path or query text

Definitions are included for **published** data marts only — a draft's definition is not settled yet (product decision). Draft documents keep their overview, schema, and joins.

This mirrors the [OWOX Model Canvas](https://model.owox.com/) serializer, which renders the same section.

## Why

The bundle is pitched as context for AI assistants — but without physical references an assistant has to ask for the dataset and emits `<YOUR_DATASET>` placeholders instead of runnable SQL. With the definitions included, the generated queries use the real project, dataset, and table names. Found while preparing the feature demo.

## How

- The canvas enrichment already fetched the full data mart detail; a new `extractDefinitionText` helper flattens the per-type definition DTO into a single string, threaded through `ModelCanvasNode` → `ModelGraph` → the OKF renderer.
- The **JSON export intentionally stays definition-free** — it mirrors the sanitized Model Canvas share format, and the sanitize test now pins that.

## Testing

- Unit tests: `extractDefinitionText` for all five definition types (including mismatched-shape degradation), the graph passthrough, the sanitize exclusion, and the rendered `## Definition` section (sql vs text fencing, omitted when absent).
- Full model-canvas suite green; `tsc -b`, eslint, prettier, markdownlint, and the production build are clean.

### DIFF
--- a/apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
+++ b/apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
@@ -136,8 +136,13 @@ function renderNode(
       '\n\n'
     : '';
 
-  const definition = node.definition?.trim()
-    ? `## Definition\n\n\`\`\`${node.inputSource === 'SQL' ? 'sql' : 'text'}\n${node.definition.trim()}\n\`\`\`\n\n`
+  // A definition line starting with three-plus backticks would close the fence
+  // early (and the Model Canvas parser cuts at the first such line). A single
+  // leading space keeps the line inert for both renderers; inside SQL it can
+  // only occur in comments or string literals, where the space is harmless.
+  const definitionText = node.definition?.trim().replace(/^( {0,3})(`{3,})/gm, ' $1$2');
+  const definition = definitionText
+    ? `## Definition\n\n\`\`\`${node.inputSource === 'SQL' ? 'sql' : 'text'}\n${definitionText}\n\`\`\`\n\n`
     : '';
 
   const outgoing = graph.edges.filter(
--- a/apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
+++ b/apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
@@ -136,6 +136,10 @@ function renderNode(
       '\n\n'
     : '';
 
+  const definition = node.definition?.trim()
+    ? `## Definition\n\n\`\`\`${node.inputSource === 'SQL' ? 'sql' : 'text'}\n${node.definition.trim()}\n\`\`\`\n\n`
+    : '';
+
   const outgoing = graph.edges.filter(
     edge => edge.from === node.key || (edge.bidirectional && edge.to === node.key)
   );
@@ -159,5 +163,5 @@ function renderNode(
       '\n'
     : '';
 
-  return `---\n${frontmatter}\n---\n\n# ${node.title}\n${node.description ? '\n' + node.description + '\n' : ''}\n${overview}${schema}${joins}`;
+  return `---\n${frontmatter}\n---\n\n# ${node.title}\n${node.description ? '\n' + node.description + '\n' : ''}\n${overview}${schema}${definition}${joins}`;
 }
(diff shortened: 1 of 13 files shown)


apps/web/src/features/data-marts/model-canvas/model/use-model-canvas.ts
```diff
@@ -54,6 +55,7 @@ async function enrichNodes(
       enriched[start + index] = {
         ...target,
         definitionType: detail.definitionType,
+        definition: extractDefinitionText(detail.definitionType, detail.definition),
```
prokopetsv, line 58: 🔴 Blocker: Definitions exist only after the background detail query finishes, but export becomes available when the base canvas is ready. On larger canvases, sequential batches can leave a substantial window where OKF export succeeds without any definitions. A rejected detail request is also silently retained as an unenriched node, with no retry or export warning. Gate OKF export on successful enrichment, or fetch the required definitions during export.

apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
```diff
@@ -136,6 +136,10 @@ function renderNode(
       '\n\n'
     : '';
 
+  const definition = node.definition?.trim()
+    ? `## Definition\n\n\`\`\`${node.inputSource === 'SQL' ? 'sql' : 'text'}\n${node.definition.trim()}\n\`\`\`\n\n`
```
prokopetsv: 🟡 Non-Blocker: The fixed triple-backtick fence does not safely contain definitions with a line beginning with three backticks, which is valid inside a SQL block comment. That line closes the fence, causing the remaining query to become Markdown and truncating the definition on OKF re-import. Use a content-safe delimiter or encoding supported by the parser, with a round-trip test.

apps/web/src/features/data-marts/model-canvas/model/use-model-canvas.ts
laskevych: Agreed on both windows — fixed in ff0b43629:

- `useModelCanvas` now exposes `isEnriching` (base nodes present, detail query not yet settled), and the view declines export during that window with an explicit toast ("still loading details") — no download, no analytics event. This closes the gap where a large canvas could quietly serialize OKF/JSON without definitions and schemas.
- For the failure mode: a rejected detail request keeps its node compact (`fields` absent). The export still proceeds after enrichment settles — blocking it entirely on one inaccessible mart felt worse — but it is no longer silent: a toast reports "Exported without schema details for N data marts — reload the page to retry."
- Tests: a view test with `isEnriching: true` (toast + no analytics), a view test where the export handle resolves against unenriched nodes (warning toast fires, export succeeds), and the hook test pins `isEnriching` settling to `false`.

apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
laskevych: Nice catch — fixed in ff0b43629. Lines inside a definition matching `/^( {0,3})`{3,}/` get one leading space before fencing. That keeps the line inert for both consumers: CommonMark no longer treats it as a closing fence, and the Model Canvas parser regex (`/\n```/`, which cuts at the FIRST such line regardless of our fence length — so a longer outer fence alone wouldn't have helped re-import) now extracts the full definition. Inside SQL a leading-backtick line can only occur in comments or string literals, where the extra space is harmless. The test round-trips the emitted document through the reference parser's own regex and asserts the tail of the query survives.

prokopetsv (approved):
✅ Approving.

Traced the full path: detail fetch → `extractDefinitionText` → `ModelCanvasNode` →
`canvasToModelGraph` → OKF renderer / JSON sanitizer.

- Definitions reach the OKF bundle only; the JSON export stays sanitized
  (field whitelist + pinned regression test).
- Connector definitions excluded by construction; the published-only gate
  reuses the existing status predicate.
- Both review findings resolved in ff0b43629: export now waits for detail
  enrichment (explicit toast, plus a warning when some detail fetches failed),
  and a backtick-fence line inside a SQL definition can no longer truncate the
  definition on Model Canvas re-import (round-trip tested against the
  reference parser's regex).
- No new endpoints, requests, or dependencies; export remains a local download.
- CI: build, lint, and all test suites green. The Audit failure is the
  pre-existing astro/sharp advisories (also failing on main), unrelated to
  this PR.

PR #1500, opened 2026-08-05
feat(data-marts): turn on STRING_AGG and ANY_VALUE by default for text fields (#6791)

zapolsky:
Closes [#6791](https://owox.fibery.io/Sprint_Work/Work_Item/Add-by-default-STRING_AGG-and-ANY_VALUE-possible-aggs-for-string-fields-6791)

## What

Text fields now ship with `STRING_AGG` (Combined) and `ANY_VALUE` (Sample) turned on, alongside `COUNT` and `COUNT_DISTINCT`. Both were already in the supported menu for text — they were just off by default, so analysts had to enable them field by field before any report could use them.

```diff
-string: { role: 'dimension', allowedAggregations: ['COUNT', 'COUNT_DISTINCT'] },
+string: {
+  role: 'dimension',
+  allowedAggregations: ['COUNT', 'COUNT_DISTINCT', 'STRING_AGG', 'ANY_VALUE'],
+},
```

Applied in all three governance copies the source comments require to stay in sync: backend `field-aggregation-governance.ts`, web `aggregation-governance.ts`, and the extension mirror (separate PR in `google-sheets-extension`).

Only the `string` category changes. Numeric, date, time, boolean and complex types are untouched.

## Existing Data Marts pick this up without a migration

`allowedAggregations` is optional on the schema field — *"absent = derive defaults by type"* — and no schema provider writes the resolved defaults back. Governance is resolved on read, so:

- a field with **no** explicit list gains the two functions immediately;
- a field where someone deliberately narrowed the list keeps exactly what they chose.

That is the semantics the ticket asks for, and it is why there is no migration in this PR.

## Checked for blast radius

| Area | Result |
| --- | --- |
| MCP | `STRING_AGG` / `ANY_VALUE` are not in `MCP_AGGREGATE_FUNCTIONS`, so the advertised MCP matrix is unchanged |
| Totals | `NON_SUMMARIZABLE_AGGREGATIONS` already excludes both — no meaningless grand totals appear |
| Looker Studio | the aggregation mapper already maps both to `DIMENSION` |

## Visible UI change worth knowing

`AllowedAggregationsSelect` prints the joined function names only while `value.length <= 2`, otherwise `"N selected"`. Text fields now have four defaults, so the **Σ available** column reads **"4 selected"** instead of `COUNT, COUNT_DISTINCT`. The test for that is updated explicitly rather than deleted, so the change is visible in review.

## Verification

| Check | Result |
| --- | --- |
| backend jest | 511 suites / 6873 tests pass |
| web vitest | 181 files / 1370 tests pass |
| `nest build` | clean |
| web `tsc -b` | clean |
| eslint, markdownlint | clean |

Three test expectations that pinned the old default were updated: the governance specs, `BlendableSchemaService` post-join defaults, and the two schema-settings table tests.

## Docs

`docs/getting-started/setup-guide/report-aggregations.md` — the "Default (on) / Also available" table now lists both functions under Default for Text.

## Out of scope

Date and time fields have the same supported menu and the same argument could be made for them, but the ticket scopes this to text fields only.

### DIFF
--- a/.changeset/6791-string-default-aggregations.md
+++ b/.changeset/6791-string-default-aggregations.md
@@ -1,0 +1,9 @@
+---
+'owox': minor
+---
+
+# Combined and Sample are now on by default for text fields
+
+Text columns now offer `Combined` (`STRING_AGG`) and `Sample` (`ANY_VALUE`) out of the box, alongside `Count` and `Count Unique`. Previously both were supported but switched off, so you had to enable them field by field before a report could use them.
+
+Existing Data Marts pick this up automatically — a field only keeps a narrower list if you explicitly chose one for it. Numeric, date, time, boolean, and complex-type fields are unchanged, and neither function is ever included in Totals.
--- a/apps/backend/src/data-marts/dto/schemas/field-aggregation-governance.spec.ts
+++ b/apps/backend/src/data-marts/dto/schemas/field-aggregation-governance.spec.ts
@@ -32,10 +32,10 @@ describe('resolveFieldGovernance — default allowed aggregations (no explicit o
     expect(resolveFieldGovernance('TIME').allowedAggregations).toEqual(['MIN', 'MAX']);
   });
 
-  it('string → dimension with [COUNT, COUNT_DISTINCT]', () => {
+  it('string → dimension with [COUNT, COUNT_DISTINCT, STRING_AGG, ANY_VALUE]', () => {
     expect(resolveFieldGovernance('STRING')).toEqual({
       role: 'dimension',
-      allowedAggregations: ['COUNT', 'COUNT_DISTINCT'],
+      allowedAggregations: ['COUNT', 'COUNT_DISTINCT', 'STRING_AGG', 'ANY_VALUE'],
     });
   });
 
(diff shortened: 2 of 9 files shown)


laskevych (approved):
Reviewed the governance change, tests, docs, and the blast-radius claims — verified locally that `STRING_AGG`/`ANY_VALUE` are filtered out of every MCP surface via `MCP_AGG_SET`, that `NON_SUMMARIZABLE_AGGREGATIONS` keeps both out of Totals, and that the remaining `['COUNT', 'COUNT_DISTINCT']` occurrences in specs are explicit fixtures rather than derived defaults. No blocking issues — approving. Two non-blocking notes:

💡 **Suggestion (non-blocking):** `SourceFieldsTable.test.tsx` (~line 238, the #6733 regression test) — the fixture comment still describes `postJoinAggregations: ['COUNT', 'COUNT_DISTINCT']` as "The backend's STRING default". After this PR the derived STRING default is the four-item list, so the comment is stale and the scenario no longer mirrors what the backend would actually derive. Consider updating the fixture to the new default (or rewording the comment) so the regression case keeps representing reality.

🔵 **Question (non-blocking):** Could you link the `google-sheets-extension` mirror PR here? The source comments require the three governance copies to stay in sync, so the merge/release order matters — if the extension mirror lags, the extension briefly derives the old two-item STRING default and shows a narrower aggregation menu than the web UI for the same field. Is that window acceptable, or is the rollout coordinated?

PR #1503, opened 2026-08-06
fix(data-marts): carry field descriptions into the canvas exports

laskevych:
## Problem

Follow-up to #1496. The Models canvas export dropped the business descriptions of Output Schema fields, even though they are stored in the Data Mart and shown in the Output Schema editor:

- **OKF bundle** — the schema table's Description column carried only the `PK.` / `FK to [Target]` notes, so a field like `customers.id` exported as `FK to [Orders](./orders.md)` while its actual description ("Unique numerical identifier assigned to each customer account.") was lost, and fields without joins exported with an empty cell.
- **JSON model graph** — schema fields had no `description` property at all.

The descriptions were dropped at the very first mapping step (`mapSchemaFields`), so no export format ever saw them.

## Fix

Thread `description` through the whole chain — `CanvasNodeField` → `ModelGraphSchemaField` → serializers:

- `mapSchemaFields` now carries the field description from the Data Mart schema (blank values are omitted).
- The OKF Description cell renders `PK.` → description. Multi-line descriptions collapse to a single line so they cannot break the Markdown table row.
- The `FK to [Target]` notes left the schema table entirely: every join already appears in the Joins section with its condition, so the note was a duplicate that crowded out the business description. This deliberately diverges from the model.owox.com serializer — its parser reads joins from the Joins keys first and falls back to FK notes only when the keys are missing, and our export always writes the keys, so the round-trip is unaffected.
- The JSON export picks the field up automatically via `sanitizeModelGraph` (schema passes through unchanged); the reference `SchemaField` type already declares `description?: string`, so the shape stays compatible.

The canvas cards themselves are unchanged — the description is export-only data.

## Tests

- `serialize.test.ts`: description rendered in the Description cell; multi-line description collapses to one line; no FK notes anywhere in the schema table.
- `model-graph.test.ts`: description mapped into the graph; explicit assertion that it survives `sanitizeModelGraph` into the JSON export.
- All model-canvas tests pass; `tsc --noEmit` clean.

### DIFF
--- a/apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
+++ b/apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
@@ -2,9 +2,15 @@ import type { ModelGraph, ModelGraphNode } from '../model-graph';
 import { slugify } from '../slug';
 import { renderFrontmatter } from './frontmatter';
 
-/** Keep a value from breaking out of its Markdown table cell. */
+/**
+ * Keep a value from breaking out of its Markdown table cell. The HTML entity is
+ * used instead of the GFM `\|` escape because the Model Canvas importer splits
+ * rows on every raw pipe without recognizing backslash escapes, which would
+ * truncate the cell on re-import; `&#124;` renders as `|` and survives the
+ * split intact.
+ */
 function escapeTableCell(value: string): string {
-  return value.replace(/\|/g, '\\|');
+  return value.replace(/\|/g, '&#124;');
 }
 
 /** Keep a title from terminating its `[text](./slug.md)` link early. */
--- a/apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
+++ b/apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
@@ -64,32 +64,6 @@ export function serializeOkfBundle(graph: ModelGraph, bundleTitle = 'Data Marts'
   return { files };
 }
 
-// Map each of a node's own FK columns to the target mart it points at, so the
-// FK note can be rendered inside that column's Description cell.
-function fkColumns(
-  node: ModelGraphNode,
-  graph: ModelGraph,
-  slugByKey: Map<string, string>
-): Map<string, { title: string; slug: string }> {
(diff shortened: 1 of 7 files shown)


apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
```diff
@@ -123,10 +96,16 @@ function renderNode(
...
+          // deliberate divergence from model.owox.com, which also annotates FK
+          // columns here — its parser reads joins from the Joins keys first,
+          // so the round-trip is unaffected).
           const parts: string[] = [];
           if (field.pk) parts.push('PK.');
-          const ref = fk.get(field.name);
-          if (ref) parts.push(`FK to [${escapeLinkText(ref.title)}](./${ref.slug}.md)`);
+          // Newlines would end the Markdown table row, so multi-line
+          // descriptions collapse to a single line.
+          if (field.description) parts.push(field.description.replace(/\s+/g, ' ').trim());
```
prokopetsv, line 114: 🟡 Non-Blocker: Descriptions containing `|` do not round-trip through Model Canvas. `escapeTableCell()` emits `\|`, but the current importer splits rows on every raw pipe without recognizing escapes (OWOX/models `packages/okf/src/parse.ts:192`). For example, `Eligible A | B customers` is re-imported as `Eligible A \`, silently losing data. Add coverage for this delimiter and either emit an importer-compatible representation or coordinate escaped-cell parsing in the consumer.

prokopetsv (approved):

apps/web/src/features/data-marts/model-canvas/export/okf/serialize.ts
laskevych: Fixed in ded630313. `escapeTableCell()` now emits `&#124;` instead of the GFM `\|` escape — the entity renders as a pipe in Markdown viewers but contains no raw `|`, so the importer's raw-pipe row split keeps the cell whole. Added a test that mirrors the importer's split (`row.split('|').slice(1, -1)`) and asserts `Eligible A | B customers` survives as one cell. Titles and aliases get the same treatment since they share the escaper.

PR #1507, opened 2026-08-06
feat(web): move Joinable Data Marts filters to the toolbar and apply them to the list view

laskevych:
## Summary

Follow-up to #1479 per PM feedback on the Joinable Data Marts block (task 6776-1) — first step towards full design parity with the Models canvas:

- The **Status** and **looped data marts** filters move out of the diagram's gear popover onto the toolbar, next to search, as dropdowns styled after the Models canvas toolbar (`All statuses` / `Published only` / `Draft only`, `Hide looped data marts` / `Show looped data marts`).
- Filters now apply to **both views**. Previously they only trimmed the Graph, so switching to List showed a different set of relationships. The List applies the same rules via the new pure `filterTransientRows`, including the diagram's subtree semantics: a filtered-out data mart also hides everything joined through it.
- `relationship-canvas-filters.ts` is renamed to `relationship-filters.ts` (it is no longer canvas-only). Per-mart localStorage keys are unchanged, so saved preferences carry over.
- `RelationshipCanvas` loses the filter popover, its dead uncontrolled fallback state and the un-scoped localStorage keys; `showLooped`/`statusFilter` become required props with no change callbacks.

The gear button disappears in this PR; it returns in the next phase carrying view settings (View Compact/Detailed, Layout algorithm, Show join fields, Object labels) shared with the Models canvas.

## Test plan

- `vitest run src/features/data-marts/edit/components/DataMartRelationships/` — 61/61 passing (canvas filter tests rewritten as prop-driven; new unit tests for `filterTransientRows` cover status, looped, subtree drop and the sibling-alias-prefix edge case).
- Manual: toggle Status/Looped on the toolbar and confirm List and Graph show the same set; check inline vs fullscreen diagram stay in sync; confirm previously saved filter prefs are picked up.


apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
```diff
@@ -445,16 +455,49 @@ export function DataMartRelationshipsContent({
...
-        <SearchInput
-          id='search-relationships'
-          placeholder='Search data marts'
-          value={searchInput}
-          onChange={setSearchInput}
-          debounceTime={0}
-          className='border-muted dark:border-muted/50 rounded-md border bg-white pl-8 text-sm dark:bg-white/4 dark:hover:bg-white/8'
-          aria-label='Search data marts'
-        />
+      <div className='flex min-w-0 items-center justify-between gap-2 pb-4'>
```
prokopetsv: 🟡 Non-Blocker: The toolbar is now one non-wrapping row. The new selects enforce 150px and 180px minimum widths, while search and the relationship/view controls share the row. In the 600px-wide edit layout this must overflow or overlap, making filters or List/Graph controls inaccessible. Allow the toolbar/control group to wrap or stack at a breakpoint, and add a narrow-viewport regression test.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
```diff
@@ -276,14 +285,15 @@ export function DataMartRelationshipsContent({
   );
 
   const filteredRows = useMemo(() => {
-    if (!searchQuery) return transientRows;
+    const rows = filterTransientRows(transientRows, { showLooped, statusFilter });
```
prokopetsv, line 288: 🟡 Non-Blocker: The new tests cover `filterTransientRows` and prop-driven canvas filtering, but none render `DataMartRelationshipsContent`. The primary workflow—changing toolbar selections, filtering list rows, and preserving existing per-mart localStorage keys—can regress while all 61 tests pass. Add a parent-level test that seeds stored preferences, changes both selects, and asserts visible rows plus persisted values/canvas props.

prokopetsv (approved):

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
laskevych: Done in ae4ebdd95 — the toolbar row is now `flex-wrap` (the counter + List/Graph group keeps right alignment via `ml-auto`), so at ~600px the controls wrap instead of overflowing. Added a regression guard in the new container test; since jsdom can't measure overflow, it asserts the wrap class — real layout coverage would need an e2e viewport test.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
laskevych: Done in ae4ebdd95 — added `DataMartRelationshipsContent.test.tsx` covering exactly this workflow: it seeds the per-mart localStorage keys and asserts the list is filtered on first render, changes both selects and asserts the visible rows plus the persisted values, and switches to Graph to assert the canvas receives the same `showLooped`/`statusFilter` props. The design-system Select is stubbed with a native `<select>` (Radix needs pointer-capture/ResizeObserver APIs jsdom lacks); the test targets the container's wiring rather than Radix internals.

PR #1457, opened 2026-07-23
fix(api): audit project run history contract

max-voloshyn:
## Scope

Audit and repair the complete contract for `GET /api/data-marts/runs` only:

- reconcile viewer authentication and project/Data Mart visibility behavior;
- document exact `limit`/`offset` normalization and newest-first pagination;
- make every response field, enum, nested object, nullability rule, and format explicit;
- preserve `definitionRun` as required but nullable when a historical snapshot is unavailable;
- identify `createdByUser` unambiguously as the run author, including its nullable cases and
  `userId`, `fullName`, `email`, and `avatar`;
- expose the client operation consistently as `runs.list({ limit, offset })` and align generated
  OpenAPI, runtime validation, focused tests, the public guide, and the canonical workflow
  changeset.

The only cross-cutting repair is the shared `DataMartRunResponseApiDto` metadata required by this
endpoint. Its stricter required-property schema also reaches existing run-bearing consumers, but
`toRunResponse` already emits every key for those consumers, so no unrelated runtime payload
changes are included. `docs/api/openapi.md` is unchanged.

Audited base: `44d89941a37207d4a0e18a9db6b083b3d534e077`

Reconciled `main` through `a54bf5dd8c798f0ad1714d067c3e24a9db59c967`. The intervening
releases, MCP-reporting, changelog-link corrections, and Search API-client coverage do not change
this endpoint's service, mapper, or client source. The first release consumed the prior canonical
changeset and released `runs.getHistory(...)`; the canonical file contains the merged Search
capability, which this branch preserves before appending the run-history contract and the required
`getHistory(...)` to `list(...)` migration. Current `main`
`8724c41f0b1fae5a2834f0deda625fce8f099529` adds only an unrelated connector-settings web fix and
its separate changeset, so it requires no endpoint or canonical-workflow reconciliation.

## Coverage matrix

| Surface | Before | After |
| --- | --- | --- |
| Swagger/OpenAPI | Covered since 2026-07-21 | Covered since 2026-07-23 |
| `@owox/api-client` | Covered since 2026-07-21 | Covered since 2026-07-23 |

Coverage classification:

- Swagger/OpenAPI: materially repaired; its covered-since date is reset to 2026-07-23.
- `@owox/api-client`: materially repaired; its covered-since date is reset to 2026-07-23.

The Support Matrix links remain valid:

- Swagger: `#/Run%20History/ProjectDataMartRunsController_list`
- API client: `#read-project-run-history`

## Contract checklist

[...]

PR #1508, opened 2026-08-07
feat(api): cover Data Mart run lifecycle

max-voloshyn:
## Summary

- add a typed, Data-Mart-scoped `@owox/api-client` run lifecycle under `runs.forDataMart(id)`
- make the four lifecycle operations explicit, named OpenAPI contracts and preserve existing scoped pagination behavior
- document the client surface and mark both coverage dimensions as covered

## Scope

Audited base: `60b1151a86b7ede6d877f78b9e6b7cae91e40d9f`

| Endpoint | OpenAPI | API client |
| --- | --- | --- |
| `POST /api/data-marts/{id}/manual-run` | Unassessed → Covered | Unassessed → Covered (`runs.forDataMart(id).start`) |
| `GET /api/data-marts/{id}/runs` | Unassessed → Covered | Unassessed → Covered (`runs.forDataMart(id).list`) |
| `GET /api/data-marts/{id}/runs/{runId}` | Unassessed → Covered | Unassessed → Covered (`runs.forDataMart(id).get`) |
| `POST /api/data-marts/{id}/runs/{runId}/cancel` | Unassessed → Covered | Unassessed → Covered (`runs.forDataMart(id).cancel`) |

The existing project-wide `runs.list()` remains a separate endpoint with unchanged server-owned pagination normalization. Project-wide and scoped run reads remain compatible with older deployments that omit Data Quality fields.

## Contract checklist

- Authorization: manual run and cancel retain editor access; list and detail retain viewer access. Generated OpenAPI security continues to come from the shared auth decorator.
- Manual run request: named closed incremental and manual-backfill DTO variants own the generated `oneOf` contract. Backend request validation rejects explicit `null` payloads, unknown keys, unsupported run types, and non-object `data`; it accepts retained object-valued `data` on incremental requests and fieldless manual backfills for existing web forms. The public API-client type and runtime validator are intentionally stricter: any `data` requires explicit `MANUAL_BACKFILL`, and implicit or explicit incremental runs with `data` are rejected before authentication or network access. The endpoint-level 1 MiB limit is reachable over HTTP; the shared request-body transport contract publishes and returns `413` beyond the parser ceiling. Packaged authentication body parsers are scoped to `/auth`, so they do not preempt the backend `/api` ceiling.
- Manual run behavior: non-connector and unpublished connector Data Marts return the documented `400` before connector dispatch.
- Pagination: the scoped endpoint preserves its existing caller-provided limit/offset behavior and 100/0 omission defaults. The scoped client rejects invalid pagination, including integers outside JavaScript's safe range, before authentication or network access. Project-wide `runs.list()` remains outside this slice and continues to rely on server normalization.
- Responses: manual run has a named UUID response; list and detail use named run DTOs; status/type/trigger and mapper-owned totals are required; Data Quality summaries/details are required-nullable; cancel is `204 No Content`.
- Runtime client validation: blank, dot-segment, separator-bearing, and percent-bearing Data Mart/run IDs; malformed run states, timestamps, known nested Data Quality values, scoped start/list options, and response shapes are rejected with `OWOXApiError`. Safe ID characters that require encoding remain encoded. Scoped reads normalize Data Quality fields omitted by older deployments to `null`; additive Data Quality response fields and server-owned invariants remain forward-compatible.
- Operation identities: generated-spec tests pin `manualRun`, `getRunHistory`, `getRunById`, and `cancelRun` operation IDs used by the coverage matrix.
- Cross-cutting ownership: authentication stays in the shared auth decorator; request-body `413` responses stay in the generated-document transform; request/response shapes use named DTOs; shared run literals and validators have one client source.
- Exclusions: none.

## Verification

- backend request-DTO and generated-OpenAPI suites on the final rebased head: 15 tests passed
- backend Data Mart runs e2e suite: 10 tests passed, including omitted-payload behavior, 1 MiB validation, and transport-level `413`
- packaged-server body-parser regression: 2 tests passed, including a 128 KiB `/api` request and retained `/auth` JSON parsing
- legacy Better Auth focused provider suite: 34 tests passed
- focused API-client lifecycle, project run-history, and transport-seam suites: 81 tests passed, including the strict manual-backfill discriminator, backend-compatible fieldless backfills, legacy scoped-response normalization, safe ID encoding, unsafe-ID rejection, and unchanged project-wide pagination behavior
- backend production TypeScript graph: passed (`tsconfig.build.json`)
- API-client and both changed IDP provider typechecks: passed
- affected backend, API-client, IDP-provider, and OWOX ESLint checks: passed
- Prettier, Markdown lint, and diff whitespace checks: passed
- API coverage validator: passed (139 endpoints, 16 fully covered)
- Changesets status against current `origin/main`: passed with the single canonical minor changeset
- `docs/api/openapi.md`: unchanged from current `main`
- current-main update `ec28b5ea7`: rebased and re-audited; its low-level PATCH/DELETE release section is preserved in the canonical changeset, and scoped resource paths comply with its authenticated-path security contract
- independent full-diff and repeated delta reviews: all published blocking feedback addressed; final review found no blocker or non-blocker

The broader backend spec-inclusive TypeScript graph still reports pre-existing errors outside this diff; the production graph and every affected focused suite pass.

## Blocker register

- Review blockers: none.
- Merge blockers: all reported checks passed on `f361648f520e55adb2cb0082c3c4ec9e0324b4a1`; an authorized engineer approval is still required.

🤖 Codex (GPT-5.6 Sol High)

### DIFF
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -26,9 +26,9 @@ consumers should prefer typed resource abstractions when available.
 the Data-Mart-scoped `runs.forDataMart(id)` client and its `start()`, `list()`, `get()`, and `cancel()`
 methods. Starting and cancelling require Technical User access; listing and inspecting require
 Business User access. `start()` accepts typed incremental or manual-backfill options, including
-connector-specific backfill fields. Serialized start options are limited to 1 MB. List pagination
-defaults to 100 items, caps the limit at 100 and offset at 100,000, and normalizes invalid numeric
-values. Responses are type-checked at runtime, including Data Quality details.
+connector-specific backfill fields. List pagination defaults to 100 items, caps the limit at 100 and
+offset at 100,000, and normalizes invalid numeric values. Responses are type-checked at runtime,
+including Data Quality details.
 
 The existing project-wide `runs.list()` method is unchanged, so current integrations require no
 migration.
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -23,12 +23,12 @@ consumers should prefer typed resource abstractions when available.
 ## Manage Data Mart run lifecycles through the API client
 
 `@owox/api-client` now supports starting, listing, inspecting, and cancelling Data Mart runs through
-`runs.start()`, `runs.listForDataMart()`, `runs.get()`, and `runs.cancel()`. Starting and cancelling
-require Technical User access; listing and inspecting require Business User access. `runs.start()`
-accepts typed incremental or manual-backfill options, including connector-specific backfill fields.
-Serialized start options are limited to 1 MB. List pagination defaults to 100 items, caps the limit
-at 100 and offset at 100,000, and normalizes invalid numeric values. Responses are type-checked at
-runtime, including Data Quality details.
+the Data-Mart-scoped `runs.forDataMart(id)` client and its `start()`, `list()`, `get()`, and `cancel()`
+methods. Starting and cancelling require Technical User access; listing and inspecting require
+Business User access. `start()` accepts typed incremental or manual-backfill options, including
+connector-specific backfill fields. Serialized start options are limited to 1 MB. List pagination
+defaults to 100 items, caps the limit at 100 and offset at 100,000, and normalizes invalid numeric
+values. Responses are type-checked at runtime, including Data Quality details.
 
 The existing project-wide `runs.list()` method is unchanged, so current integrations require no
 migration.
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -2,7 +2,9 @@
 'owox': minor
 ---
 
-# Low-level API client transport methods
+# API surface maintenance
+
+## Low-level API client transport methods
 
 OWOX API clients can now use `patchJson()` and `deleteJson()` alongside
 `getJson()`, `postJson()`, `putJson()`, and `getStream()` for `/api/...`
@@ -17,3 +19,19 @@ The existing plugin protocol adds PATCH and DELETE through `ctx.owox` as an
 additive capability while preserving compatibility with existing plugins.
 Low-level JSON return values are caller-typed and are not runtime-validated;
 consumers should prefer typed resource abstractions when available.
+
+## Add Data Mart run lifecycle client support
+
+The Data Mart run lifecycle endpoints now publish their complete request, response, visibility,
+pagination, and cancellation contracts. Manual runs and cancellation require Technical User
+access; listing and inspecting runs require Business User access to the Data Mart. Manual connector
+payloads must be JSON objects no larger than 1 MB. Run list pagination defaults to 100, caps the
+limit at 100 and the offset at 100,000, and normalizes fractional, non-finite, and non-positive
+values.
+
+`@owox/api-client` adds `dataMarts.run(dataMartId, request)`,
+`dataMarts.listRuns(dataMartId, options)`, `dataMarts.getRun(dataMartId, runId)`, and
+`dataMarts.cancelRun(dataMartId, runId)`. It exports the Data Mart run, request, response,
+pagination, status, type, author, and Data Quality types and validates the complete runtime
+response before returning it. Existing `runs.list()` remains the separate project-wide run-history
+method; no existing call needs migration.
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -24,10 +24,11 @@ consumers should prefer typed resource abstractions when available.
 
 `@owox/api-client` now supports starting, listing, inspecting, and cancelling Data Mart runs through
 `runs.start()`, `runs.listForDataMart()`, `runs.get()`, and `runs.cancel()`. Starting and cancelling
-require Technical User access; listing and inspecting require Business User access. Manual payloads
-must be objects no larger than 1 MB, and list pagination defaults to 100 items, caps the limit at 100
-and offset at 100,000, and normalizes invalid numeric values. Responses are type-checked at runtime,
-including Data Quality details.
+require Technical User access; listing and inspecting require Business User access. `runs.start()`
+accepts typed incremental or manual-backfill options, including connector-specific backfill fields.
+Serialized start options are limited to 1 MB. List pagination defaults to 100 items, caps the limit
+at 100 and offset at 100,000, and normalizes invalid numeric values. Responses are type-checked at
+runtime, including Data Quality details.
 
 The existing project-wide `runs.list()` method is unchanged, so current integrations require no
 migration.
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -26,19 +26,19 @@ consumers should prefer typed resource abstractions when available.
 the Data-Mart-scoped `runs.forDataMart(id)` client and its `start()`, `list()`, `get()`, and `cancel()`
 methods. Starting and cancelling require Technical User access; listing and inspecting require
 Business User access. `start()` accepts typed incremental or manual-backfill options, including
-connector-specific backfill fields. Manual-backfill `data` now requires an explicit
-`runType: 'MANUAL_BACKFILL'`; incremental runs reject `data` instead of silently ignoring it. The
-client also rejects empty or dot-segment Data Mart and run IDs before sending a request. Serialized
-manual-run options are limited to 1 MiB by both the client and HTTP API; requests above the HTTP
-transport ceiling return `413` instead of an internal-server error. Packaged authentication
-middleware now limits its body parsers to `/auth`, so it no longer overrides the backend API's 2 MiB
-transport ceiling.
+connector-specific fields. Manual-backfill `data` is optional for connectors without backfill
+fields, and object-valued `data` remains accepted for incremental runs so existing run forms and
+API consumers do not fail when they retain hidden field values. The client also rejects empty or
+dot-segment Data Mart and run IDs before sending a request. Serialized manual-run options are
+limited to 1 MiB by both the client and HTTP API; requests above the HTTP transport ceiling return
+`413` instead of an internal-server error. Packaged authentication middleware now limits its body
+parsers to `/auth`, so it no longer overrides the backend API's 2 MiB transport ceiling.
 
 Scoped list pagination defaults to 100 items when omitted and preserves valid caller-provided limits
 and offsets without silently capping them. Both scoped and project-wide list methods reject unknown,
 zero or negative limits, negative offsets, non-integer values, non-finite values, and integers outside
 JavaScript's safe range before authentication or network access.
-Project-wide `runs.list()` remains compatible with older self-hosted deployments that omit
-`qualitySummary`, and Data Quality response validation tolerates additive server fields while still
-checking known values. Existing typed integrations need no migration unless they passed backfill
-`data` without the required manual-backfill run type or relied on invalid pagination values.
+Project-wide and Data-Mart-scoped run methods remain compatible with older self-hosted deployments
+that omit Data Quality fields, and Data Quality response validation tolerates additive server fields
+while still checking known values. Existing typed integrations need no migration unless they relied
+on invalid pagination values.
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -36,8 +36,8 @@ transport ceiling.
 
 Scoped list pagination defaults to 100 items when omitted and preserves valid caller-provided limits
 and offsets without silently capping them. Both scoped and project-wide list methods reject unknown,
-zero or negative limits, negative offsets, non-integer values, and non-finite values before
-authentication or network access.
+zero or negative limits, negative offsets, non-integer values, non-finite values, and integers outside
+JavaScript's safe range before authentication or network access.
 Project-wide `runs.list()` remains compatible with older self-hosted deployments that omit
 `qualitySummary`, and Data Quality response validation tolerates additive server fields while still
 checking known values. Existing typed integrations need no migration unless they passed backfill
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -30,17 +30,17 @@ connector-specific fields. Manual-backfill `data` is optional for connectors wit
 fields. The API client requires an explicit `MANUAL_BACKFILL` run type whenever `data` is supplied
 and rejects `data` on implicit or explicit incremental runs before sending a request. The backend
 HTTP endpoint separately accepts retained object-valued `data` on incremental requests so existing
-run forms remain compatible. The client also rejects empty or dot-segment Data Mart and run IDs
-before sending a request. Serialized manual-run options are limited to 1 MiB by both the client and
-HTTP API; requests above the HTTP transport ceiling return `413` instead of an internal-server
-error. Packaged authentication middleware now limits its body parsers to `/auth`, so it no longer
-overrides the backend API's 2 MiB transport ceiling.
+run forms remain compatible. The client also rejects empty, dot-segment, or separator-bearing Data
+Mart and run IDs before sending a request. Serialized manual-run options are limited to 1 MiB by
+both the client and HTTP API; requests above the HTTP transport ceiling return `413` instead of an
+internal-server error. Packaged authentication middleware now limits its body parsers to `/auth`,
+so it no longer overrides the backend API's 2 MiB transport ceiling.
 
 Scoped list pagination defaults to 100 items when omitted and preserves valid caller-provided limits
-and offsets without silently capping them. Both scoped and project-wide list methods reject unknown,
-zero or negative limits, negative offsets, non-integer values, non-finite values, and integers outside
-JavaScript's safe range before authentication or network access.
+and offsets without silently capping them. The scoped list method rejects unknown, zero or negative
+limits, negative offsets, non-integer values, non-finite values, and integers outside JavaScript's
+safe range before authentication or network access. Project-wide `runs.list()` pagination remains
+unchanged and relies on server normalization.
 Project-wide and Data-Mart-scoped run methods remain compatible with older self-hosted deployments
 that omit Data Quality fields, and Data Quality response validation tolerates additive server fields
-while still checking known values. Existing typed integrations need no migration unless they relied
-on invalid pagination values.
+while still checking known values. Existing typed integrations need no migration.
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -26,9 +26,17 @@ consumers should prefer typed resource abstractions when available.
 the Data-Mart-scoped `runs.forDataMart(id)` client and its `start()`, `list()`, `get()`, and `cancel()`
 methods. Starting and cancelling require Technical User access; listing and inspecting require
 Business User access. `start()` accepts typed incremental or manual-backfill options, including
-connector-specific backfill fields. List pagination defaults to 100 items, caps the limit at 100 and
-offset at 100,000, and normalizes invalid numeric values. Responses are type-checked at runtime,
-including Data Quality details.
+connector-specific backfill fields. Manual-backfill `data` now requires an explicit
+`runType: 'MANUAL_BACKFILL'`; incremental runs reject `data` instead of silently ignoring it. The
+client also rejects empty or dot-segment Data Mart and run IDs before sending a request. Serialized
+manual-run options are limited to 1 MiB by both the client and HTTP API; requests above the HTTP
+transport ceiling return `413` instead of an internal-server error.
 
-The existing project-wide `runs.list()` method is unchanged, so current integrations require no
-migration.
+Scoped list pagination defaults to 100 items when omitted and preserves valid caller-provided limits
+and offsets without silently capping them. Both scoped and project-wide list methods reject unknown,
+zero or negative limits, negative offsets, non-integer values, and non-finite values before
+authentication or network access.
+Project-wide `runs.list()` remains compatible with older self-hosted deployments that omit
+`qualitySummary`, and Data Quality response validation tolerates additive server fields while still
+checking known values. Existing typed integrations need no migration unless they passed backfill
+`data` without the required manual-backfill run type or relied on invalid pagination values.
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -20,18 +20,8 @@ additive capability while preserving compatibility with existing plugins.
 Low-level JSON return values are caller-typed and are not runtime-validated;
 consumers should prefer typed resource abstractions when available.
 
-## Add Data Mart run lifecycle client support
+## Manage Data Mart run lifecycles through the API client
 
-The Data Mart run lifecycle endpoints now publish their complete request, response, visibility,
-pagination, and cancellation contracts. Manual runs and cancellation require Technical User
-access; listing and inspecting runs require Business User access to the Data Mart. Manual connector
-payloads must be JSON objects no larger than 1 MB. Run list pagination defaults to 100, caps the
-limit at 100 and the offset at 100,000, and normalizes fractional, non-finite, and non-positive
-values.
-
-`@owox/api-client` adds `dataMarts.run(dataMartId, request)`,
-`dataMarts.listRuns(dataMartId, options)`, `dataMarts.getRun(dataMartId, runId)`, and
-`dataMarts.cancelRun(dataMartId, runId)`. It exports the Data Mart run, request, response,
-pagination, status, type, author, and Data Quality types and validates the complete runtime
-response before returning it. Existing `runs.list()` remains the separate project-wide run-history
-method; no existing call needs migration.
+`@owox/api-client` now supports starting, listing, inspecting, and cancelling Data Mart runs through
+`runs.start()`, `runs.listForDataMart()`, `runs.get()`, and `runs.cancel()`. These methods enforce the
+same access, payload, pagination, and response contracts as the service API.
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -30,7 +30,9 @@ connector-specific backfill fields. Manual-backfill `data` now requires an expli
 `runType: 'MANUAL_BACKFILL'`; incremental runs reject `data` instead of silently ignoring it. The
 client also rejects empty or dot-segment Data Mart and run IDs before sending a request. Serialized
 manual-run options are limited to 1 MiB by both the client and HTTP API; requests above the HTTP
-transport ceiling return `413` instead of an internal-server error.
+transport ceiling return `413` instead of an internal-server error. Packaged authentication
+middleware now limits its body parsers to `/auth`, so it no longer overrides the backend API's 2 MiB
+transport ceiling.
 
 Scoped list pagination defaults to 100 items when omitted and preserves valid caller-provided limits
 and offsets without silently capping them. Both scoped and project-wide list methods reject unknown,
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -27,12 +27,14 @@ the Data-Mart-scoped `runs.forDataMart(id)` client and its `start()`, `list()`, 
 methods. Starting and cancelling require Technical User access; listing and inspecting require
 Business User access. `start()` accepts typed incremental or manual-backfill options, including
 connector-specific fields. Manual-backfill `data` is optional for connectors without backfill
-fields, and object-valued `data` remains accepted for incremental runs so existing run forms and
-API consumers do not fail when they retain hidden field values. The client also rejects empty or
-dot-segment Data Mart and run IDs before sending a request. Serialized manual-run options are
-limited to 1 MiB by both the client and HTTP API; requests above the HTTP transport ceiling return
-`413` instead of an internal-server error. Packaged authentication middleware now limits its body
-parsers to `/auth`, so it no longer overrides the backend API's 2 MiB transport ceiling.
+fields. The API client requires an explicit `MANUAL_BACKFILL` run type whenever `data` is supplied
+and rejects `data` on implicit or explicit incremental runs before sending a request. The backend
+HTTP endpoint separately accepts retained object-valued `data` on incremental requests so existing
+run forms remain compatible. The client also rejects empty or dot-segment Data Mart and run IDs
+before sending a request. Serialized manual-run options are limited to 1 MiB by both the client and
+HTTP API; requests above the HTTP transport ceiling return `413` instead of an internal-server
+error. Packaged authentication middleware now limits its body parsers to `/auth`, so it no longer
+overrides the backend API's 2 MiB transport ceiling.
 
 Scoped list pagination defaults to 100 items when omitted and preserves valid caller-provided limits
 and offsets without silently capping them. Both scoped and project-wide list methods reject unknown,
--- a/.changeset/api-surface-maintenance.md
+++ b/.changeset/api-surface-maintenance.md
@@ -23,5 +23,11 @@ consumers should prefer typed resource abstractions when available.
 ## Manage Data Mart run lifecycles through the API client
 
 `@owox/api-client` now supports starting, listing, inspecting, and cancelling Data Mart runs through
-`runs.start()`, `runs.listForDataMart()`, `runs.get()`, and `runs.cancel()`. These methods enforce the
-same access, payload, pagination, and response contracts as the service API.
+`runs.start()`, `runs.listForDataMart()`, `runs.get()`, and `runs.cancel()`. Starting and cancelling
+require Technical User access; listing and inspecting require Business User access. Manual payloads
+must be objects no larger than 1 MB, and list pagination defaults to 100 items, caps the limit at 100
+and offset at 100,000, and normalizes invalid numeric values. Responses are type-checked at runtime,
+including Data Quality details.
+
+The existing project-wide `runs.list()` method is unchanged, so current integrations require no
+migration.
--- a/apps/backend/src/bootstrap.ts
+++ b/apps/backend/src/bootstrap.ts
@@ -12,6 +12,7 @@ import { GlobalExceptionFilter } from './common/exceptions/global-exception.filt
 import { createLogger } from './common/logger/logger.service';
 import { DEFAULT_PORT } from './config/constants';
 import { setupGlobalPipes } from './config/global-pipes.config';
+import { setupBodyParsers } from './config/body-parser.config';
 import { runMigrationsIfNeeded } from './config/migrations.config';
 import { PROTOCOL_ROUTE_EXCLUSIONS } from './config/protocol-route-exclusions.config';
 import { setupSwagger } from './config/swagger.config';
@@ -44,6 +45,8 @@ export async function bootstrap(options: BootstrapOptions): Promise<NestExpressA
     new ExpressAdapter(options.express),
     { logger }
   );
+
+  setupBodyParsers(app);
 
   app.useLogger(createLogger());
   app.useGlobalFilters(new GlobalExceptionFilter(), new BaseExceptionFilter());
--- a/apps/backend/src/data-marts/controllers/data-mart.controller.ts
+++ b/apps/backend/src/data-marts/controllers/data-mart.controller.ts
@@ -13,6 +13,7 @@ import { DataMartValidationResponseApiDto } from '../dto/presentation/data-mart-
 import { ListDataMartsQueryApiDto } from '../dto/presentation/list-data-marts-query-api.dto';
 import { PaginatedDataMartsResponseApiDto } from '../dto/presentation/paginated-data-marts-response-api.dto';
 import { RunDataMartRequestApiDto } from '../dto/presentation/run-data-mart-request-api.dto';
+import { RunDataMartResponseApiDto } from '../dto/presentation/run-data-mart-response-api.dto';
 import { UpdateDataMartDefinitionApiDto } from '../dto/presentation/update-data-mart-definition-api.dto';
 import { UpdateBlendedFieldsConfigApiDto } from '../dto/presentation/update-blended-fields-config-api.dto';
 import { UpdateDataMartDescriptionApiDto } from '../dto/presentation/update-data-mart-description-api.dto';
@@ -78,6 +79,7 @@ import {
   ValidateDataMartDefinitionSpec,
   DataMartAiHelperAvailabilitySpec,
 } from './spec/data-mart.api';
+import { normalizeProjectListPagination } from '../utils/normalize-project-list-pagination';
 
 @Controller('data-marts')
 @ApiTags('DataMarts')
@@ -285,7 +287,7 @@ export class DataMartController {
     @AuthContext() context: AuthorizationContext,
     @Param('id') id: string,
     @Body() dto: RunDataMartRequestApiDto
-  ): Promise<{ runId: string }> {
+  ): Promise<RunDataMartResponseApiDto> {
     const command = this.mapper.toRunCommand(id, context, dto.payload);
     const runId = await this.runDataMartService.run(command);
     return { runId };
@@ -348,10 +350,16 @@ export class DataMartController {
   async getRunHistory(
     @AuthContext() context: AuthorizationContext,
     @Param('id') id: string,
-    @Query('limit') limit: number = 100,
-    @Query('offset') offset: number = 0
+    @Query('limit') limit?: string | number,
+    @Query('offset') offset?: string | number
   ): Promise<DataMartRunsResponseApiDto> {
-    const command = this.mapper.toGetDataMartRunsCommand(id, context, limit, offset);
+    const pagination = normalizeProjectListPagination(limit, offset);
+    const command = this.mapper.toGetDataMartRunsCommand(
+      id,
+      context,
+      pagination.limit,
+      pagination.offset
+    );
     const runs = await this.getDataMartRunsService.run(command);
     return this.mapper.toRunsResponse(runs);
   }
--- a/apps/backend/src/data-marts/controllers/data-mart.controller.ts
+++ b/apps/backend/src/data-marts/controllers/data-mart.controller.ts
@@ -79,7 +79,6 @@ import {
   ValidateDataMartDefinitionSpec,
   DataMartAiHelperAvailabilitySpec,
 } from './spec/data-mart.api';
-import { normalizeProjectListPagination } from '../utils/normalize-project-list-pagination';
 
 @Controller('data-marts')
 @ApiTags('DataMarts')
@@ -350,16 +349,10 @@ export class DataMartController {
   async getRunHistory(
     @AuthContext() context: AuthorizationContext,
     @Param('id') id: string,
-    @Query('limit') limit?: string | number,
-    @Query('offset') offset?: string | number
+    @Query('limit') limit: number = 100,
+    @Query('offset') offset: number = 0
   ): Promise<DataMartRunsResponseApiDto> {
-    const pagination = normalizeProjectListPagination(limit, offset);
-    const command = this.mapper.toGetDataMartRunsCommand(
-      id,
-      context,
-      pagination.limit,
-      pagination.offset
-    );
+    const command = this.mapper.toGetDataMartRunsCommand(id, context, limit, offset);
     const runs = await this.getDataMartRunsService.run(command);
     return this.mapper.toRunsResponse(runs);
   }
--- a/apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
+++ b/apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
@@ -33,11 +33,7 @@ import { UpdateEntityContextsRequestApiDto } from '../../dto/presentation/contex
 import { DATA_MARTS_PAGE_SIZE } from '../../use-cases/list-data-marts.service';
 import { BatchDataMartDataLastUpdatedResponseApiDto } from '../../dto/presentation/data-mart-data-last-updated-response-api.dto';
 import { RefreshDataMartDataLastUpdatedRequestApiDto } from '../../dto/presentation/refresh-data-mart-data-last-updated-request-api.dto';
-import {
-  DEFAULT_PROJECT_LIST_LIMIT,
-  MAX_PROJECT_LIST_LIMIT,
-  MAX_PROJECT_LIST_OFFSET,
-} from '../../utils/normalize-project-list-pagination';
+import { DEFAULT_PROJECT_LIST_LIMIT } from '../../utils/normalize-project-list-pagination';
 
 export function CreateDataMartSpec() {
   return applyDecorators(
@@ -221,8 +217,6 @@ export function GetDataMartRunsSpec() {
       required: false,
       type: Number,
       default: DEFAULT_PROJECT_LIST_LIMIT,
-      minimum: 1,
-      maximum: MAX_PROJECT_LIST_LIMIT,
       description:
         'Maximum number of runs to return. Non-positive or non-finite values use the default; fractions are floored and larger values are capped.',
     }),
@@ -231,8 +225,6 @@ export function GetDataMartRunsSpec() {
       required: false,
       type: Number,
       default: 0,
-      minimum: 0,
-      maximum: MAX_PROJECT_LIST_OFFSET,
       description:
         'Number of runs to skip. Non-positive or non-finite values use zero; fractions are floored and larger values are capped.',
     }),
--- a/apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
+++ b/apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
@@ -27,11 +27,17 @@ import { DataMartRunDetailResponseApiDto } from '../../dto/presentation/data-mar
 import { UpdateDataMartOwnersApiDto } from '../../dto/presentation/update-data-mart-owners-api.dto';
 import { PaginatedDataMartsResponseApiDto } from '../../dto/presentation/paginated-data-marts-response-api.dto';
 import { RunDataMartRequestApiDto } from '../../dto/presentation/run-data-mart-request-api.dto';
+import { RunDataMartResponseApiDto } from '../../dto/presentation/run-data-mart-response-api.dto';
 import { UpdateDataMartAvailabilityApiDto } from '../../dto/presentation/update-availability-api.dto';
 import { UpdateEntityContextsRequestApiDto } from '../../dto/presentation/context-api.dto';
 import { DATA_MARTS_PAGE_SIZE } from '../../use-cases/list-data-marts.service';
 import { BatchDataMartDataLastUpdatedResponseApiDto } from '../../dto/presentation/data-mart-data-last-updated-response-api.dto';
 import { RefreshDataMartDataLastUpdatedRequestApiDto } from '../../dto/presentation/refresh-data-mart-data-last-updated-request-api.dto';
+import {
+  DEFAULT_PROJECT_LIST_LIMIT,
+  MAX_PROJECT_LIST_LIMIT,
+  MAX_PROJECT_LIST_OFFSET,
+} from '../../utils/normalize-project-list-pagination';
 
 export function CreateDataMartSpec() {
   return applyDecorators(
@@ -168,24 +174,20 @@ export function DeleteDataMartSpec() {
 
 export function RunDataMartSpec() {
   return applyDecorators(
-    ApiOperation({ summary: 'Manual run DataMart' }),
-    ApiParam({ name: 'id', description: 'DataMart ID' }),
+    ApiOperation({
+      summary: 'Start a manual Data Mart run',
+      description:
+        'Starts a connector Data Mart run. Technical User access to the Data Mart is required.',
+    }),
+    ApiParam({ name: 'id', description: 'Data Mart ID' }),
     ApiBody({ type: RunDataMartRequestApiDto, required: false }),
     ApiResponse({
       status: 201,
       description: 'DataMart run created',
-      schema: {
-        type: 'object',
-        required: ['runId'],
-        properties: {
-          runId: {
-            type: 'string',
-            format: 'uuid',
-            example: '123e4567-e89b-12d3-a456-426614174000',
-          },
-        },
-      },
-    })
+      type: RunDataMartResponseApiDto,
+    }),
+    ApiResponse({ status: 400, description: 'Invalid payload or unsupported Data Mart type' }),
+    ApiResponse({ status: 404, description: 'Data Mart not found' })
   );
 }
 
@@ -208,32 +210,53 @@ export function UpdateDataMartSchemaSpec() {
 
 export function GetDataMartRunsSpec() {
   return applyDecorators(
-    ApiOperation({ summary: 'Get DataMart run history' }),
-    ApiParam({ name: 'id', description: 'DataMart ID' }),
+    ApiOperation({
+      summary: 'List Data Mart runs',
+      description:
+        'Returns newest-first run history for one visible Data Mart. Business User access is required.',
+    }),
+    ApiParam({ name: 'id', description: 'Data Mart ID' }),
     ApiQuery({
       name: 'limit',
       required: false,
       type: Number,
-      example: 100,
-      description: 'Maximum number of runs to return. Defaults to 100.',
+      default: DEFAULT_PROJECT_LIST_LIMIT,
+      minimum: 1,
+      maximum: MAX_PROJECT_LIST_LIMIT,
+      description:
+        'Maximum number of runs to return. Non-positive or non-finite values use the default; fractions are floored and larger values are capped.',
     }),
     ApiQuery({
       name: 'offset',
       required: false,
       type: Number,
-      example: 0,
-      description: 'Number of runs to skip before returning results',
+      default: 0,
+      minimum: 0,
+      maximum: MAX_PROJECT_LIST_OFFSET,
+      description:
+        'Number of runs to skip. Non-positive or non-finite values use zero; fractions are floored and larger values are capped.',
     }),
-    ApiOkResponse({ type: DataMartRunsResponseApiDto })
+    ApiOkResponse({
+      description: 'Newest-first page of Data Mart runs.',
+      type: DataMartRunsResponseApiDto,
+    }),
+    ApiResponse({ status: 404, description: 'Data Mart not found' })
   );
 }
 
 export function CancelDataMartRunSpec() {
   return applyDecorators(
-    ApiOperation({ summary: 'Cancel a DataMart run' }),
-    ApiParam({ name: 'id', description: 'DataMart ID' }),
+    ApiOperation({
+      summary: 'Cancel a Data Mart run',
+      description:
+        'Cancels an active connector, standard report, or Data Quality run. Technical User access to the Data Mart is required.',
+    }),
+    ApiParam({ name: 'id', description: 'Data Mart ID' }),
     ApiParam({ name: 'runId', description: 'Run ID' }),
-    ApiNoContentResponse({ description: 'DataMart run cancelled' })
+    ApiNoContentResponse({ description: 'Data Mart run cancelled' }),
+    ApiResponse({ status: 400, description: 'The run type cannot be cancelled' }),
+    ApiResponse({ status: 404, description: 'Data Mart or run not found' }),
+    ApiResponse({ status: 409, description: 'The run is no longer active' })
   );
 }
 
@@ -247,10 +270,15 @@ export function ListDataMartsByConnectorNameSpec() {
 
 export function GetDataMartRunByIdSpec() {
   return applyDecorators(
-    ApiOperation({ summary: 'Get DataMart run by ID' }),
-    ApiParam({ name: 'id', description: 'DataMart ID' }),
+    ApiOperation({
+      summary: 'Get a Data Mart run',
+      description:
+        'Returns one run belonging to a visible Data Mart, including Data Quality detail when applicable. Business User access is required.',
+    }),
+    ApiParam({ name: 'id', description: 'Data Mart ID' }),
     ApiParam({ name: 'runId', description: 'Run ID' }),
-    ApiOkResponse({ type: DataMartRunDetailResponseApiDto })
+    ApiOkResponse({ type: DataMartRunDetailResponseApiDto }),
+    ApiResponse({ status: 404, description: 'Data Mart or run not found' })
   );
 }
 
--- a/apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
+++ b/apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
@@ -182,7 +182,10 @@ export function RunDataMartSpec() {
       description: 'DataMart run created',
       type: RunDataMartResponseApiDto,
     }),
-    ApiResponse({ status: 400, description: 'Invalid payload or unsupported Data Mart type' }),
+    ApiResponse({
+      status: 400,
+      description: 'Invalid payload, unsupported Data Mart type, or unpublished Data Mart',
+    }),
     ApiResponse({ status: 404, description: 'Data Mart not found' })
   );
 }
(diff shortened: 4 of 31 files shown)


.changeset/api-surface-maintenance.md
```diff
@@ -0,0 +1,21 @@
...
+# API surface maintenance
+
+## Add Data Mart run lifecycle client support
+
+The Data Mart run lifecycle endpoints now publish their complete request, response, visibility,
+pagination, and cancellation contracts. Manual runs and cancellation require Technical User
+access; listing and inspecting runs require Business User access to the Data Mart. Manual connector
+payloads must be JSON objects no larger than 1 MB. Run list pagination defaults to 100, caps the
+limit at 100 and the offset at 100,000, and normalizes fractional, non-finite, and non-positive
+values.
```
max-voloshyn: There too many implementation details here

.changeset/api-surface-maintenance.md
```diff
@@ -0,0 +1,21 @@
...
+
+`@owox/api-client` adds `dataMarts.run(dataMartId, request)`,
+`dataMarts.listRuns(dataMartId, options)`, `dataMarts.getRun(dataMartId, runId)`, and
+`dataMarts.cancelRun(dataMartId, runId)`. It exports the Data Mart run, request, response,
```
max-voloshyn: runs should be consistently managed via `runs.*`, not inside `dataMarts.*`

.changeset/api-surface-maintenance.md
max-voloshyn: Addressed in 282b7b75c. The changeset is now concise, release-facing prose while retaining the standalone role, payload, pagination, validation, and compatibility details required for adoption.

🤖 Codex (GPT-5.6 Sol High)

.changeset/api-surface-maintenance.md
max-voloshyn: Addressed in 5053e0a20. The lifecycle now lives exclusively under `runs.start`, `runs.listForDataMart`, `runs.get`, and `runs.cancel`; the existing project-wide `runs.list` remains unchanged.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
```diff
@@ -241,6 +241,52 @@ unknown enum value, malformed user, storage, or context, invalid timestamp, miss
 field, or invalid pagination value with `OWOXApiError`. Invalid list options are rejected before
 authentication or any network request.
 
+## Manage Data Mart runs
+
+Use `dataMarts.run(dataMartId, request)` to start a manual connector run. Technical User access
+to the Data Mart is required. The optional `payload` must be a JSON object no larger than 1 MB;
+omit it to use the Data Mart's default connector payload. The method returns the new run ID.
```
max-voloshyn: Why we should mention payload size here?

docs/api/api-client.md
```diff
@@ -241,6 +241,52 @@ unknown enum value, malformed user, storage, or context, invalid timestamp, miss
...
+
+```ts
+const { runId } = await client.dataMarts.run('data-mart-id', {
+  payload: { cursor: 'next-page' },
```
max-voloshyn: is there more idomatic interface than general "payload"?

docs/api/api-client.md
```diff
@@ -241,6 +241,52 @@ unknown enum value, malformed user, storage, or context, invalid timestamp, miss
...
+newest-first page, while `runs.get()` returns one run and includes full Data Quality detail
+when the run is a Data Quality run.
+
+```ts
+const history = await client.runs.listForDataMart('data-mart-id', {
+  limit: 50,
+  offset: 0,
+});
+
+const run = await client.runs.get('data-mart-id', history.runs[0].id);
```
max-voloshyn: why do we need data mart id here if we have run id?

docs/api/api-client.md
```diff
@@ -241,6 +241,52 @@ unknown enum value, malformed user, storage, or context, invalid timestamp, miss
...
+  payload: { cursor: 'next-page' },
+});
+```
+
+Business Users can inspect runs for a Data Mart they can see. `runs.listForDataMart()` returns a
+newest-first page, while `runs.get()` returns one run and includes full Data Quality detail
+when the run is a Data Quality run.
+
+```ts
+const history = await client.runs.listForDataMart('data-mart-id', {
```
max-voloshyn: how is it compatible with client.runs.list ? may they me merged somehow ? (should they?)

docs/api/api-client.md
```diff
@@ -241,6 +241,52 @@ unknown enum value, malformed user, storage, or context, invalid timestamp, miss
...
+console.log(run.status, run.qualitySummary, run.dataQuality);
+```
+
+The list endpoint defaults to 100 runs, caps `limit` at 100 and `offset` at 100,000, floors finite
+fractions, and replaces non-finite or non-positive values with its defaults. The response has no
+total or next-page marker. Increment `offset` by the number of returned runs and stop when a page
+contains fewer runs than the effective limit or the next offset would exceed 100,000. New runs can
+shift offset pages, so deduplicate by `run.id` while paging.
+
+Use `runs.cancel(dataMartId, runId)` to cancel an active connector, standard report, or
```
max-voloshyn: why do we need both dataMartId and runId?

docs/api/api-client.md
max-voloshyn: Retained as a concise public constraint in b5d3b3e53 because both the backend DTO and the API client reject serialized start options over 1 MiB. Without documenting it, type-valid options could fail at runtime unexpectedly. The guide now describes the limit on typed scoped options rather than exposing the generic wire payload.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
max-voloshyn: Addressed in 341586b5a and de0c288f1. The public start interface is now typed as `runType` plus connector-specific `data`, exposed through `runs.forDataMart(id).start(options)`. The `{ payload: ... }` wrapper remains only an internal mapping to the existing backend route.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
max-voloshyn: Addressed in de0c288f1. The Data Mart ID is now supplied once when creating `runs.forDataMart(dataMartId)`; inspection is `scope.get(runId)`, so consumers no longer repeat the Data Mart ID on each operation while the client still calls the scoped backend route.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
max-voloshyn: Addressed in de0c288f1. Existing `client.runs.list()` remains the compatible project-wide history and includes each run's Data Mart reference. `client.runs.forDataMart(id).list()` is intentionally scoped to one Data Mart and its response omits that redundant reference. Merging them into one overload would make the return shape scope-dependent and weaken the existing `runs.list()` contract.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
max-voloshyn: Addressed in de0c288f1. Consumers now create `const runs = client.runs.forDataMart(dataMartId)` once and call `runs.cancel(runId)` (or `runs.get(runId)`), while the scope retains the Data Mart ID required by the backend route and authorization boundary.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
```diff
@@ -168,24 +174,20 @@ export function DeleteDataMartSpec() {
...
-            type: 'string',
-            format: 'uuid',
-            example: '123e4567-e89b-12d3-a456-426614174000',
-          },
-        },
-      },
-    })
+      type: RunDataMartResponseApiDto,
+    }),
+    ApiResponse({ status: 400, description: 'Invalid payload or unsupported Data Mart type' }),
```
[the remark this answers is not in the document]
max-voloshyn: Addressed in b5d3b3e53. `RunDataMartService` now throws `BadRequestException` for non-connector manual runs, matching the documented `400`, and the focused behavior test verifies both the exception class and that connector execution is not dispatched.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
```diff
@@ -208,32 +210,53 @@ export function UpdateDataMartSchemaSpec() {
...
+    ApiParam({ name: 'id', description: 'Data Mart ID' }),
     ApiQuery({
       name: 'limit',
       required: false,
       type: Number,
-      example: 100,
-      description: 'Maximum number of runs to return. Defaults to 100.',
+      default: DEFAULT_PROJECT_LIST_LIMIT,
+      minimum: 1,
+      maximum: MAX_PROJECT_LIST_LIMIT,
```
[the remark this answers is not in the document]
max-voloshyn: Addressed in b5d3b3e53. The generated OpenAPI schemas no longer publish `minimum` or `maximum` for `limit` and `offset`; they retain defaults and the normalization description. The generated-document test now pins the absence of bounds so clients do not reject values accepted and normalized by runtime.

🤖 Codex (GPT-5.6 Sol High)

.changeset/api-surface-maintenance.md
```diff
@@ -0,0 +1,18 @@
...
+
+# API surface maintenance
+
+## Manage Data Mart run lifecycles through the API client
+
+`@owox/api-client` now supports starting, listing, inspecting, and cancelling Data Mart runs through
+the Data-Mart-scoped `runs.forDataMart(id)` client and its `start()`, `list()`, `get()`, and `cancel()`
+methods. Starting and cancelling require Technical User access; listing and inspecting require
+Business User access. `start()` accepts typed incremental or manual-backfill options, including
+connector-specific backfill fields. Serialized start options are limited to 1 MB. List pagination
```
max-voloshyn: `Serialized start options are limited to 1 MB.`
Is there realistic case that call options will not match this wide limit?

apps/backend/src/data-marts/controllers/data-mart.controller.ts
```diff
@@ -348,10 +350,16 @@ export class DataMartController {
   async getRunHistory(
     @AuthContext() context: AuthorizationContext,
     @Param('id') id: string,
-    @Query('limit') limit: number = 100,
-    @Query('offset') offset: number = 0
+    @Query('limit') limit?: string | number,
+    @Query('offset') offset?: string | number
```
max-voloshyn: Why default values are omited?

max-voloshyn:
Addressed the latest review feedback in 75f5e9dda:

- The 1 MiB cap is a defensive validation limit rather than a realistic adoption concern for normal incremental or backfill options, so it no longer appears in the release note. It remains in the public API-client guide because both the client and backend enforce it.
- Pagination defaults remain intentionally owned by the shared `normalizeProjectListPagination` helper. Nest supplies raw query strings (or `undefined`), and the helper applies the documented 100/0 defaults together with flooring, caps, and invalid-value normalization for both run-history controllers. OpenAPI continues to publish `default: 100` and `default: 0`; duplicating defaults in the controller signature would create a second contract source without changing behavior.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
```diff
@@ -241,6 +241,65 @@ unknown enum value, malformed user, storage, or context, invalid timestamp, miss
 field, or invalid pagination value with `OWOXApiError`. Invalid list options are rejected before
 authentication or any network request.
 
+## Manage Data Mart runs
+
+Create a Data-Mart-scoped run client with `runs.forDataMart(dataMartId)`. Its `start(options)` method
+starts a manual connector run and requires Technical User access to the Data Mart. Omit the options
+for an incremental run, or set `runType` to `MANUAL_BACKFILL` and pass the connector-specific
+backfill fields in `data`. The method returns the new run ID. The serialized options must not exceed
+1 MB.
```
max-voloshyn: `The serialized options must not exceed
1 MB.`

Is it realistic limit? Should it be mentioned at all?

docs/api/api-client.md
```diff
@@ -241,6 +241,65 @@ unknown enum value, malformed user, storage, or context, invalid timestamp, miss
...
+  runType: 'MANUAL_BACKFILL',
+  data: {
+    StartDate: '2026-07-01',
+    EndDate: '2026-07-31',
+  },
+});
+```
+
+Business Users can inspect runs for a Data Mart they can see. The scoped `list()` method returns a
+newest-first page, while `get(runId)` returns one run and includes full Data Quality detail when the
```
max-voloshyn, line 325: isn't that get(runId) competes with general get for runId?

docs/api/api-client.md
max-voloshyn: Retained after revalidation. Normal incremental and backfill options are not expected to approach 1 MiB, which is why the limit was removed from the release note. The public `data` field is connector-specific and unbounded at the type level, though, and both the API client and backend reject serialized options above 1 MiB. This method guide is the consumer-facing disclosure that prevents a type-valid request from failing unexpectedly, so the concise warning remains here.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
max-voloshyn: There is no general `runs.get(runId)` surface today. Project-wide runs expose only `GET /api/data-marts/runs` through `client.runs.list()`, while detail is available only at `GET /api/data-marts/{id}/runs/{runId}` and the Data Mart ID participates in the visibility boundary. `runs.forDataMart(id).get(runId)` therefore maps the supported route without competing with a project-wide `get`; adding `runs.get(runId)` would imply a backend capability that does not exist or hide a required identifier.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/controllers/data-mart.controller.ts
```diff
@@ -348,10 +350,16 @@ export class DataMartController {
...
   ): Promise<DataMartRunsResponseApiDto> {
-    const command = this.mapper.toGetDataMartRunsCommand(id, context, limit, offset);
+    const pagination = normalizeProjectListPagination(limit, offset);
```
romandubovyi: 🔴 **Blocking:** This endpoint previously had no upper bound on `limit`: `@Query('limit') limit: number = 100` combined with the global `ValidationPipe({ transform: true })` passed the caller's value straight into `listByDataMartId` as `take` (`data-mart-run.service.ts:155`), so `?limit=500` returned 500 runs. `normalizeProjectListPagination` now clamps it to `MAX_PROJECT_LIST_LIMIT` (100).

An existing API-key integration that requests `limit=500` and advances `offset` by 500 will now silently receive 100 runs per page and skip the other 400, with no error to detect it. The changeset states that current integrations require no migration, which does not hold for `GET /api/data-marts/{id}/runs`.

Either keep this route's previous unbounded behaviour, or treat the cap as an intentional breaking change and call it out in the changeset.

packages/api-client/src/data-mart-runs.ts
```diff
@@ -0,0 +1,475 @@
...
+): string {
+  if (scope.type === 'DATA_MART') return `${category}:data_mart`;
+  if (scope.type === 'FIELD') return `${category}:field:${JSON.stringify(scope.fieldPath)}`;
+  return `${category}:relationship:${scope.relationshipId}`;
+}
+
+function isDataQualityRule(value: unknown): value is OWOXDataQualityRule {
+  if (!isRecord(value) || !isRecord(value.parameters)) return false;
+  if (
+    !hasOnlyKeys(value, [
```
romandubovyi: ⚪ **Non-blocking:** `hasOnlyKeys` here rejects any rule object that carries a field outside this eight-key list. Adding one optional field to `EffectiveDataQualityRuleConfigSchema` is a backwards-compatible backend change, but after it every already-published `@owox/api-client` version would throw `OWOXApiError` from `runs.forDataMart(id).get(runId)` for every DATA_QUALITY run — including for consumers who never read `dataQuality`.

Lines 347-348 have the same shape: the client re-derives the rule key with its own copy of `buildDataQualityRuleKey`, and `DATA_QUALITY_CATEGORY_SCOPES` maps each category to exactly one scope, while the backend's map (`data-quality-config.schema.ts:45`) allows an array of scopes per category. Both turn a future backend-side relaxation into a client-side rejection of a valid response.

Suggest keeping unknown-key rejection for request options (where it catches caller typos) and dropping it from response validation, along with the re-derived key and single-scope assertions.

packages/api-client/src/runs.ts
```diff
@@ -185,6 +115,86 @@ function parseProjectRunHistory(response: unknown): OWOXProjectDataMartRunsRespo
...
+      !response.runs.every(isDataMartRun)
+    ) {
+      throw new OWOXApiError('OWOX Data Mart Runs API returned an unexpected response shape', {
+        details: response,
+      });
+    }
+    return response as OWOXDataMartRunsResponse;
+  }
+
+  async get(runId: string): Promise<OWOXDataMartRunDetail> {
```
romandubovyi, line 217: ⚪ **Non-blocking:** `start()` and `list()` reject invalid input before authentication and any network request, which is what the client guide promises. `get(runId)` and `cancel(runId)` do not validate `runId` at all: an empty or whitespace-only string produces `/api/data-marts/{id}/runs/` (or `/runs//cancel`), which is still sent to the API and comes back as an opaque 404 from an unrelated route.

A short non-empty check throwing `OWOXApiError` would keep the module consistent and surface the real cause to the caller.

packages/api-client/src/runs.ts
```diff
@@ -185,6 +115,86 @@ function parseProjectRunHistory(response: unknown): OWOXProjectDataMartRunsRespo
...
+      throw new OWOXApiError(
+        'OWOX Data Mart Manual Run API returned an unexpected response shape',
+        { details: response }
+      );
+    }
+    return response as OWOXRunDataMartResponse;
+  }
+
+  async list(options: OWOXDataMartRunListOptions = {}): Promise<OWOXDataMartRunsResponse> {
+    validateRunListOptions(options);
```
romandubovyi, line 192: ⚪ **Non-blocking:** This scoped `list()` validates its options, but the project-wide `RunsApi.list()` (line 201 below, outside this diff) takes the same `{ limit, offset }` shape and builds its query with no validation at all — `limit: NaN` becomes the literal string `"NaN"` in the query, and unknown keys are silently ignored.

Now that `validateRunListOptions` is a shared helper, applying it in `RunsApi.list()` too would make the two sibling methods behave consistently.

romandubovyi (commented):
Multi-agent review of the Data Mart run lifecycle API surface. Four agents analysed the PR independently; findings were merged, de-duplicated against the existing threads, and validated against the current code before being included here.

One blocking point: `GET /api/data-marts/{id}/runs` previously accepted an unbounded `limit`, so the new normalization is a silent behaviour change for existing API-key consumers. The remaining three are client-side consistency and forward-compatibility notes.

🤖 Generated with Claude Code (Opus 5 (1M context)) + Antigravity CLI (Gemini 3.5 Flash) + Kimi CLI + Grok CLI (grok-4.5)

packages/api-client/src/runs.ts
```diff
@@ -96,78 +44,60 @@ export type OWOXProjectRunHistoryOptions = {
...
+    (options.offset !== undefined && typeof options.offset !== 'number')
+  ) {
+    throw new OWOXApiError('Invalid OWOX Data Mart run-list options', { details: options });
+  }
 }
 
 function isProjectDataMartRun(value: unknown): value is OWOXProjectDataMartRun {
-  if (!isRecord(value) || !isRecord(value.dataMart)) {
+  const projectRun = value as Record<string, unknown>;
+  if (!isDataMartRun(value) || !isRecord(projectRun.dataMart)) {
```
romandubovyi: ⚪ **Non-blocking:** Reusing `isDataMartRun` here tightens `client.runs.list()`. On `main`, `isProjectDataMartRun` validated its fields inline and never looked at `qualitySummary`; the shared validator requires the key to be present (`data-mart-runs.ts`, `isCompactDataQualitySummary` branch), and `project-run-history.test.ts` now asserts that a run without it is rejected.

The backend only started returning `qualitySummary` on project-wide run history in `61fc1bc60` (2026-07-30). A published `@owox/api-client` pointed at any self-hosted deployment older than that will now throw `OWOXApiError` from `runs.list()` — a call the changeset describes as unchanged and requiring no migration.

Treating the field as optional in response validation (`value.qualitySummary === undefined || value.qualitySummary === null || isCompactDataQualitySummary(...)`) would keep the shared validator without narrowing the set of backends `runs.list()` works against.

zapolsky (commented):
🔴 **Blocked issues** — see the 2 inline comments. Both were adversarially verified (an independent skeptic tried to refute each and failed; one was empirically reproduced). These should be addressed before merge, but this review does not formally block — a human owns the merge decision.

Six further candidate blockers were **refuted** by the same gate and are not posted. Non-blocking notes are in the separate top-level comment. Posted by automated OWOX review.

packages/api-client/src/runs.ts
```diff
@@ -96,78 +44,60 @@ export type OWOXProjectRunHistoryOptions = {
...
-const PROJECT_DATA_MART_RUN_TRIGGERS = new Set<string>(PROJECT_DATA_MART_RUN_TRIGGER_VALUES);
-
-function isNullableRecord(value: unknown): boolean {
-  return value === null || isRecord(value);
-}
-
-function isNullableString(value: unknown): boolean {
-  return value === null || typeof value === 'string';
-}
+const MAX_MANUAL_RUN_PAYLOAD_BYTES = 1024 * 1024;
```
zapolsky, line 52: **[Contract — verified, reproduced]** The 1 MiB manual-run payload limit this PR publishes does not exist. The real ceiling is body-parser's **100 KiB** default, and exceeding it returns **HTTP 500**, not a documented status.

**Evidence chain (every hop verified against installed deps in the worktree):**
1. `apps/backend/src/bootstrap.ts:41-45` — `NestFactory.create(AppModule, new ExpressAdapter(...), { logger })`. No `bodyParser: false`, no `app.useBodyParser(...)`, no `express.json({ limit })`. A repo-wide search (`useBodyParser`, `bodyParser`, `express.json`, `limit:`, `'1mb'`, `raw-body`, `client_max_body_size`, all `NestFactory.create` call sites, `*.middleware.ts`, `apps/owox/**`, `Dockerfile`) found no limit anywhere; the only `express.json()` calls are the IDP providers, also without a limit.
2. `node_modules/@nestjs/core/nest-application.js:101-102` → `registerParserMiddleware()` → `node_modules/@nestjs/platform-express/adapters/express-adapter.js:188-199` → `express.json({})`.
3. `node_modules/body-parser/lib/utils.js:60-63` — `limit === undefined ? 102400 /* 100kb default */ : bytes.parse(...)`. Installed: `@nestjs/platform-express@11.1.28`, `express@5.2.1`, `body-parser@2.3.0`.
4. `apps/backend/src/data-marts/dto/presentation/run-data-mart-request-api.dto.ts:5,15` — `MAX_PAYLOAD_SIZE_BYTES = 1024 * 1024` + `@MaxJsonSize(...)` runs inside the `ValidationPipe` (`apps/backend/src/config/global-pipes.config.ts:5-11`), i.e. **after** the parser → unreachable above 100 KiB.
5. `PayloadTooLargeError` carries `status: 413` but **no `getStatus()` method**, and `apps/backend/src/common/exceptions/global-exception.filter.ts:23-28` gates precisely on `'getStatus' in e && typeof e.getStatus === 'function'` → falls through to `:92-93` `HttpStatus.INTERNAL_SERVER_ERROR` + a `logger.error('Unhandled exception caught by GlobalExceptionFilter')` per occurrence.
6. `packages/api-client/src/runs.ts:50,74-77` (this line) — the client pre-validates against 1 MiB, so it *passes through* every payload in the 100 KiB–1 MiB window that the server cannot accept. `docs/api/api-client.md:249-250` states "The serialized options must not exceed 1 MB." `apps/backend/src/data-marts/controllers/spec/data-mart.api.ts:171-187` documents only 201/400/404 — no 413 and no 500.

**Reproduced:** a minimal Nest 11 app booted against this repo's own `node_modules` with the same `ExpressAdapter` + a filter mirroring `global-exception.filter.ts` — middleware stack `["jsonParser","urlencodedParser","handle"]`; ~50 kB → `201`; ~200 kB and ~600 kB → `500 {"statusCode":500,"message":"request entity too large"}`, filter saw `{name:"PayloadTooLargeError", isHttp:false, ownStatus:413}`.

**Trigger:** `runs.forDataMart(id).start({ runType: 'MANUAL_BACKFILL', data: { ids: <150 kB> } })` — client guard passes, server 500s with a message the OpenAPI contract says cannot occur.

**Attribution:** the unreachable `@MaxJsonSize(1 MiB)` is pre-existing (`git show 60b1151a8:...run-data-mart-request-api.dto.ts` already has it). What this PR newly ships is the *claim* — in a published npm client, in public docs, in a green unit test that asserts the limit via `validate(dto)` without ever crossing HTTP, and in a newly-authored OpenAPI response set.

**Fix:** make the limit real — `app.useBodyParser('json', { limit: '1mb' })` in `bootstrap.ts` before `setupGlobalPipes(app)` (plus `urlencoded`), which makes the DTO validator reachable and all three published claims true at once. Otherwise drop the 1 MB language from `runs.ts:50`, `docs/api/api-client.md:249`, and the changeset, and document the real limit. Either way an over-limit case in `apps/backend/test/data-mart-runs.e2e-spec.ts` would catch this class — the `validate(dto)` unit test cannot, by construction.

packages/api-client/src/data-mart-runs.ts
```diff
@@ -0,0 +1,475 @@
...
+  qualitySummary: OWOXCompactDataQualitySummary | null;
+};
+
+export type OWOXDataMartRunDetail = OWOXDataMartRun & {
+  dataQuality: OWOXDataQualityRunDetail | null;
+};
+
+export type OWOXDataMartRunsResponse = { runs: OWOXDataMartRun[] };
+export type OWOXDataMartRunListOptions = { limit?: number; offset?: number };
+export type OWOXDataMartRunStartOptions = {
```
zapolsky: **[Bug — verified]** `start({ data })` without `runType` typechecks, passes client validation, returns `201` with a real `runId`, and the backfill fields are then **silently discarded**. No error at any layer.

`runType` and `data` are independent optionals here, so nothing — not the type, not the validator, not the backend, not the connector — enforces that `data` is only meaningful under `MANUAL_BACKFILL`.

**Evidence chain:**
1. `packages/api-client/src/data-mart-runs.ts:183-186` (this line) — `{ runType?: 'INCREMENTAL' | 'MANUAL_BACKFILL'; data?: Record<string, unknown> }`. Not a discriminated union, no overloads. Confirmed by this PR's own test: `packages/api-client/src/data-mart-runs.test.ts:224` calls `start({ data: { value: ... } })` with **no `as never` cast**, unlike lines 212/218 which do need one — so `{data}`-without-`runType` typechecks.
2. `packages/api-client/src/runs.ts:52-62` — `validateRunStartOptions` rejects only unknown keys, a bad `runType` literal, and a non-record `data`; `options.runType !== undefined && ...` means an absent `runType` passes.
3. `packages/api-client/src/runs.ts:139` — wire body `{ payload: options }`.
4. `apps/backend/src/data-marts/dto/presentation/run-data-mart-request-api.dto.ts:13-23` — `payload` gets only `@IsOptional() @IsObject() @MaxJsonSize(...)`; no inner shape validation. → `data-mart.controller.ts:291` → `use-cases/run-data-mart.service.ts:59-64` passes `command.payload` straight through with no runType/data cross-check → `services/connector/connector-executor.service.ts:489` `buildRunConfig(payload, configState)`.
5. `apps/backend/src/data-marts/services/connector/connector-source-config.service.ts:67-70` — `const type = typeof runTypeRaw === 'string' ? runTypeRaw : 'INCREMENTAL';`
6. `packages/connectors/src/Core/AbstractRunConfig.js:25-33` — `validate()` routes `INCREMENTAL` past `_validateManualBackfill` without inspecting `data`, returns `true`.
7. `packages/connectors/src/Core/AbstractConnector.js:67-89` — `_processRunConfig()` applies `runConfig.data` **only** in the `MANUAL_BACKFILL` branch; the `INCREMENTAL` branch is exactly `this.config.logMessage('Incremental mode activated');`. Exhaustive grep: `runConfig.data` is read in only two places repo-wide (`AbstractConnector.js:70`, `AbstractRunConfig._validateManualBackfill`), both `MANUAL_BACKFILL`-gated. `AbstractConnector.js:210-217` then computes the window from `_getIncrementalDateRange()` — the caller's values are gone.

*Escape routes checked and closed:* `buildSourceConfig` (`connector-source-config.service.ts:21-53`) is a separate method fed the stored Data Mart config (`connector-executor.service.ts:478-485`), not the payload — so the `data` entries are **not** folded into the source config regardless of run type. The payload reaches only `RunConfigDto` → `OW_RUN_CONFIG` (`connector-process-spawner.service.ts:53`).

**Trigger:** `await runs.forDataMart(id).start({ data: { StartDate: '2026-07-01', EndDate: '2026-07-31' } })` → `201` + a real runId, connector logs "Incremental mode activated", imports its normal incremental window, and the requested range is discarded with no signal in run history.

**Attribution:** the backend's `'INCREMENTAL'` fallback and the connector's gate are pre-existing. The `start()` method, its type, and its validator are **all new in this PR** — this is the first surface that offers `data` as an independently-typed, IDE-autocompleted option. `docs/api/api-client.md:246-249` documents the *pairing* but never says `data` alone is a silent no-op, and nothing enforces it.

**Fix:** make it a discriminated union — `{ runType?: 'INCREMENTAL'; data?: never } | { runType: 'MANUAL_BACKFILL'; data: Record<string, unknown> }` — plus the matching check in `validateRunStartOptions`. Worth doing before publish: this shape is a permanent commitment once it lands on npm.

apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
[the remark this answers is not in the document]
zapolsky: Independently verified this fix in the worktree — it lands, but it covers only **one** of the two rejection branches on this path.

`RunDataMartService.run()` now 400s on `definitionType !== CONNECTOR` (`use-cases/run-data-mart.service.ts:46-50`). A **connector** Data Mart that is not yet PUBLISHED passes that guard and reaches the second branch: `connector-run.service.ts:35` → `validateDataMartForConnector` → `:131` `throw new ConnectorExecutionError('DataMart is not published')`. That class extends plain `Error` (`errors/connector-execution.error.ts:8`), so `@Catch(BusinessViolationException)` (`common/exceptions/base-exception.filter.ts:6`) misses it and `global-exception.filter.ts:93` returns **500** plus an ERROR-level "Unhandled exception" log line.

Reproduced end-to-end with a throwaway e2e that runs the standard `setupConnectorDataMart` flow minus the publish step:

```
ERROR [GlobalExceptionFilter] Unhandled exception caught by GlobalExceptionFilter
ConnectorExecutionError: DataMart is not published
  at ConnectorRunService.validateDataMartForConnector (connector-run.service.ts:132:13)
{ status: 500, method: 'POST', url: '/api/data-marts/.../manual-run', isHttpError: false }
```

DRAFT is the default on creation (`entities/data-mart.entity.ts:68`) and `setup-connector-data-mart.ts:64-88` shows the canonical lifecycle passes through CONNECTOR+DRAFT — so this is the common case, not an exotic one, and `runs.forDataMart(id).start()` bypasses the UI's client-side guard (`ConnectorDefinitionField.tsx:261`) entirely.

The sibling "connector already running" case already demonstrates the correct pattern — `BusinessViolationException` → 400 via `BaseExceptionFilter`. Making `ConnectorExecutionError` extend `BusinessViolationException`, or moving the status check next to the new `definitionType` guard, would close the second branch too. Also worth noting `connector-run.service.ts:125-130`'s `definitionType` re-check is now dead on the manual-run path.

zapolsky:

Worktree checkout at `75f5e9d`, 5 parallel analysis agents + a second finder wave, every blocker through an adversarial skeptic gate. **2 blockers posted inline**; 6 further candidate blockers were **refuted** by the gate and are not posted (listed at the bottom, since two of them contradict plausible-looking claims).

Primary notes below (criticality ≥ 80), highest first — none of these block merge.

- `apps/backend/src/data-marts/use-cases/run-data-mart.service.ts:46` — **the `BadRequestException` fix covers only one of two rejection branches; the more common one still 500s.** A **DRAFT** (unpublished) connector Data Mart passes the new `definitionType` guard, reaches `connector-run.service.ts:35` → `validateDataMartForConnector` → `:131` `throw new ConnectorExecutionError('DataMart is not published')`, which extends plain `Error` (`errors/connector-execution.error.ts:8`), is not caught by `@Catch(BusinessViolationException)` (`common/exceptions/base-exception.filter.ts:6`), and lands on `global-exception.filter.ts:93` → **500** + an ERROR-level "Unhandled exception" log line. Reproduced end-to-end: `SKEPTIC manual-run HTTP status: 500` with an empty response body. DRAFT is the default on creation (`entities/data-mart.entity.ts:68`) and every connector Data Mart passes through CONNECTOR+DRAFT. The code path is byte-identical to base — but this PR is what publishes `400/404` for the endpoint and flips it to `Covered`, and `runs.forDataMart(id).start()` bypasses the UI's client-side guard entirely. The sibling "already running" case already proves the correct pattern (`BusinessViolationException` → 400). **Fix:** move the status check next to the `definitionType` guard as a `BadRequestException`, or make `ConnectorExecutionError` extend `BusinessViolationException`. (Related: `connector-run.service.ts:125-130`'s `definitionType` check is now dead on the manual-run path.)
- `.changeset/api-surface-maintenance.md:23-24` — **"`runs.list()` is unchanged, so current integrations require no migration" is not accurate.** The rewritten `isProjectDataMartRun` (`packages/api-client/src/runs.ts:93-102`) now delegates to `isDataMartRun`, which ends with `(value.qualitySummary === null || isCompactDataQualitySummary(...))` (`data-mart-runs.ts:465`) — a **missing key** satisfies neither branch, and one bad run rejects the whole page via `.every()` (`runs.ts:104-116`). The base validator (`git show 60b1151a8:packages/api-client/src/runs.ts`) never referenced the field; the PR's own test locks the new rejection in (`project-run-history.test.ts:223-240`), and the base fixture had to gain `+ qualitySummary: null`. The backend only emits it on the project response since `61fc1bc60` (#1474) → **affected combination: this client against owox 0.27.0–0.30.1**. Also a compile-time break for anyone constructing `OWOXProjectDataMartRun` (fixtures, adapters). Impact is genuinely limited — `docs/api/api-client.md:449` says matched versions only, the `fixed` changesets group keeps them in lockstep, and #1457 set precedent for exactly this tightening — so the fix is one sentence, not a revert: say the method now requires the shared `qualitySummary` field. `docs/api/api-client.md:120-129` ("Read project run history") also still enumerates the validated fields without it.
- `docs/api/api-client.md:274` — **the section's only "read a run" example doesn't compile under this repo's own config and throws on an empty history.** `const run = await dataMartRuns.get(history.runs[0].id);` — `packages/typescript-config/base.json:14` sets `noUncheckedIndexedAccess: true`, so `history.runs[0]` is `T | undefined`. Compiling the snippet verbatim against `packages/api-client/src` with that base config gives `error TS2532: Object is possibly 'undefined'`. At runtime it's worse: `list()` returns `{ runs: [] }` for a Data Mart with no runs (`[].every(...)` is `true`), so a fresh Data Mart throws `TypeError` before reaching `get()`. The same document does it correctly elsewhere (`for (const run of history.runs)`, `if (firstPage.nextOffset !== null)`). **Fix:** `const [latest] = history.runs; if (!latest) return; const run = await dataMartRuns.get(latest.id);`
- `apps/backend/src/data-marts/dto/presentation/data-mart-run-response-api.dto.ts:212` — **`totals` optionality is asymmetric between the two run schemas while the shared client guard requires it on both.** Verified by generating the real document, not by reading decorators: `DataMartRunResponseApiDto.required` and `DataMartRunDetailResponseApiDto.required` **omit** `totals`; `ProjectDataMartRunResponseApiDto.required` **includes** it (`project-data-mart-runs-response-api.dto.ts:63` passes `true`). Meanwhile `isNullableTotals` rejects `undefined` and gates all three endpoints (`data-mart-runs.ts:464` via `runs.ts:95,169,182`). The asymmetry is pre-existing, but this PR is what newly binds a present-`totals` guard to a schema that publishes it as optional — and it normalized `qualitySummary` across exactly these two DTOs while leaving the neighbour alone. **Fix:** `@DataMartRunTotalsApiProperty(true)` on the scoped DTO and drop the override.
- `packages/api-client/src/runs.ts:132` — **`forDataMart('.')` silently returns project-wide runs.** `encodeURIComponent('.') === '.'`, and `new URL(path, apiOrigin)` (`transport.ts:36`) collapses dot segments, so `/api/data-marts/./runs` → **`/api/data-marts/runs`** — the project-wide endpoint. The response validates (the extra `dataMart` key isn't rejected), so `list()` resolves with runs from every Data Mart in the project, typed as one Data Mart's history. No privilege escalation (same credentials), but silent cross-scope data from a scoped API. The `/` case *is* covered (`data-mart-runs.test.ts:130` asserts `data%2Fmart`); dots weren't considered. Related: `get('')` reaches the list route (Express non-strict routing), fails `isDataMartRunDetail`, and surfaces as "unexpected response shape" instead of 404. Notably `docs/api/api-client.md:241-243` promises "Invalid list options are rejected before authentication or any network request" — ids get no such check. **Fix:** reject empty/blank/`.`/`..` ids in `forDataMart`/`get`/`cancel`.
- `packages/api-client/src/data-mart-runs.ts:324` — **closed-world Data-Quality validators make additive backend changes break already-published clients.** `hasOnlyKeys` is applied to the rule object, `rule.parameters`, and all three `scope` variants (`:296,299,307,324,334`), and the client re-derives server invariants: `buildDataQualityRuleKey` is a byte-for-byte copy of `dto/schemas/data-quality/data-quality-config.schema.ts:61-73` and is asserted with `===` at `:347`; `MAX_DATA_QUALITY_THRESHOLD_HOURS` (`:197`), the threshold couplings (`:362-363`), and `DATA_QUALITY_CATEGORY_SCOPES` (`:198-210`) are duplicated across a package boundary with no shared fixture. The category map is also **narrower** than the server's — a single scope string compared with `===`, where the backend holds `readonly DataQualityScope[]` and uses `.includes()` (`data-quality-config.schema.ts:45-59,102`). To be precise about the blast radius (an earlier framing overstated it): the run, summary, and check-result predicates are open-world, so this is *not* "any DQ field breaks it" — the trigger set is a new key inside `scope` or `rule.parameters`, or a new member in one of the six mirrored enum lists, which then rejects the whole page/response for every deployed older client. No sibling module does this (`insight-templates.ts:46-67`, `model-canvas.ts:37-65` ignore extras). Worth noting the mutation testing: these are exactly the checks with **zero positive fixture coverage** — `hasOnlyKeys` on rule/parameters/scope, the key derivation, the category↔scope check, and the whole 12-field per-result validator can all be deleted and the 60-test suite stays green (`qualitySummary` is `null`, `results` is `[]`, `relationships` is `[]` in every accepting fixture). **Fix:** drop `hasOnlyKeys` and the re-derived invariants; keep type/enum/nullability/timestamp checks.

<details>

- `apps/backend/src/data-marts/controllers/spec/data-mart.api.ts:181` — the 400 a run-lifecycle client hits most often, `BusinessViolationException('Connector is already running…')` (`connector-run.service.ts:37-41` → `base-exception.filter.ts:13` → 400), is described as "Invalid payload or unsupported Data Mart type". A caller branching on status can't tell "retry later" from "my request is malformed". Consider mapping it to 409, matching `CancelDataMartRunSpec:251`.
- `spec/data-mart.api.ts` — all four new specs omit **403**, though each use-case throws `ForbiddenException` (`run-data-mart.service.ts:41`, `list-data-mart-runs.service.ts:32`, `get-data-mart-run.service.ts:36`, `cancel-data-mart-run.service.ts:48`). `manual-run` also has an undocumented **401** (`CredentialsExpiredException` → `google-oauth.exceptions.ts:100`) which `transports/api-key-transport.ts:114-117` treats as a stale token: it clears the token and **replays the POST**, then surfaces "Reconnect Storage" as `OWOXAuthError` (`errors.ts:121`). Benign today only because `checkStorageAccess` runs before any run row is created. Restricting the retry to GET would make that a guarantee rather than an accident.
- `apps/backend/src/data-marts/controllers/data-mart.controller.ts:356` — `GET /api/data-marts/{id}/runs` now clamps `limit` at 100 / `offset` at 100 000 (base passed the raw value through: `@Query('limit') limit: number = 100` with `transform: true` really did yield `take: 500`). This is deliberate alignment — the helper is pre-existing and four sibling controllers already used it — and the endpoint was `Unassessed` at base, so the framing "breaks existing integrations" doesn't hold. Two residues worth a line each: the cap value is deliberately undiscoverable (`data-mart.controller.openapi.spec.ts:352` asserts `not.toHaveProperty('maximum')` while the description says only "larger values are capped"), and no changeset line mentions the endpoint's own behavior changing.
- `packages/api-client/src/runs.ts:202` — `RunsApi.list()` (project-wide) validates nothing, while the scoped `list()` calls `validateRunListOptions`. `docs/api/api-client.md:241-242` already claims the guarantee for both. `client.runs.list({ limit: '25' as never, offsett: 5 as never })` silently ships a typo'd param.
- `packages/api-client/src/runs.ts:86-87` — `validateRunListOptions` only checks `typeof === 'number'`, so `NaN`, `Infinity`, `-5`, `3.7` all reach the wire as `String(...)`; the server then silently defaults them. `data-marts.ts:229-242` already has the right pattern (`Number.isInteger(v) && v >= 0`).
- `apps/backend/src/data-marts/services/connector/connector-run.service.ts:36-43` — the "already running" guard is a TOCTOU: a non-locking `findOne` (`:139-148`) inside `@Transactional()` with no unique index over non-terminal CONNECTOR runs per `dataMartId`. Two concurrent `start()` calls can both get 201. Pre-existing, but newly reachable programmatically — and the 401 replay above can produce two POSTs from one client call.
- `packages/api-client/src/data-mart-runs.ts` — the module holds only types + guards and no API class, while `DataMartRunsScope` and both option validators live in `runs.ts`. Every sibling resource module (`model-canvas.ts`, `insight-templates.ts`) owns types + guards + `parse*` + the `*Api` class in one file, and `runs.ts → data-mart-runs.js` is the package's first resource→resource import (the established home for shared guards is `validation.ts`). `index.ts` now exports one public surface from two modules.
- `docs/api/api-client.md:288-290` — "cancelling a terminal run returns a conflict error" is only half true: `cancel-data-mart-run.service.ts:64-72` checks the run **type** first and returns 400 for types that don't support cancellation. The OpenAPI spec is right; only the prose is wrong.
- `packages/api-client/src/data-mart-runs.ts:120-152` — `docs/api/api-client.md:296-298` says the package exports the Data Quality "nested-object types", but the three a consumer actually needs are anonymous inline shapes: `snapshot` (`:121`), `relationships[]` (`:124`), `results[]` (`:135`). Callers must write `OWOXDataQualityRunDetail['results'][number]`. Worth naming and exporting before the first publish.
- `spec/data-mart.api.ts:216,222` — the caps are described in prose without the numbers, while the sibling `project-data-mart-runs.api.ts:17-19,29-31` spells out 100 / 100,000; `MAX_PROJECT_LIST_LIMIT`/`MAX_PROJECT_LIST_OFFSET` are already importable in this file. Same file also has vocabulary drift for the identical `Role.viewer(Strategy.PARSE)` guard: `:212,268` say "Business User access", `:51` and `project-data-mart-runs.api.ts:11` say "Viewer access".
- `packages/api-client/src/runs.ts:184` — `get()`/`list()` put the whole response in `OWOXApiError.details`, which for a DQ run includes `examples[].values` (real warehouse rows, `data-quality-api.mapper.ts:87-89`) and `sql`; `errors.ts:18` makes it an own enumerable property with no `toJSON`, so `JSON.stringify(err)` emits it. To be fair: `details: response` is the established convention across 13 modules and `errors.ts` is untouched here — but this PR does newly make row-level values reachable through it, so a package-wide truncation/`toJSON` on `details` would be worth a follow-up.
- minor: `OWOXRunDataMartDefinitionType` (`data-mart-runs.ts:68-75`) is a byte-identical duplicate of the already-public `OWOXDataMartDefinitionType` (`data-marts.ts:17-24`) — two exported names for one concept is a permanent semver commitment; `runType` now means two disjoint vocabularies on the same scoped object (`OWOXDataMartRunStartOptions.runType` is `INCREMENTAL|MANUAL_BACKFILL`, `OWOXDataMartRun.runType` is `manual|scheduled`), and `OWOXDataMartRunType` types neither; `OWOXRunDataMartResponse` inverts the sibling noun order; `index.ts:73-82` appends `OWOXDataMartRunsScope` out of alphabetical order unlike every other export block; `OWOXDataMartRunsScope` is typed as a plain object with method members but implemented as a class reading `this.path`, so `const { list } = client.runs.forDataMart(id)` typechecks and throws at runtime; `start()`'s RFC-4122 version+variant regex (`runs.ts:144`) is far stricter than the `typeof === 'string'` used for the same identifier at `data-mart-runs.ts:444`, is unexercised by tests (the negative case passes a number), and a rejection orphans an already-created run with no id to poll or cancel; `data-mart-run.service.ts:154` lacks the `.addOrderBy('run.id','DESC')` tie-breaker its sibling at `:184` has (see refuted list — no observable effect, but worth parity).

</details>

<details>
<summary>Refuted by the verification gate — recorded because several look plausible</summary>

- **Offset paging duplicates/skips runs** (no `id` tie-breaker, second-precision `createdAt`). Mechanism is formally correct and ties are trivially reachable (10 TypeORM inserts → 1 distinct `createdAt`), but **empirically it does not manifest**: 4 configurations on the real DDL — SQLite with and without `idx_dmr_dataMart_createdAt`, MySQL 9.4 InnoDB with the index, and InnoDB with the index dropped and `sort_buffer_size=32768` to force the priority-queue filesort over a 3000-row tie group — all returned 0 skipped / 0 duplicated. Also, the paging recipe is not new: `git show 60b1151a8:docs/api/api-client.md:99-106` already had it, including the "next offset would exceed 100,000" stop condition.
- **Changeset filename must start with a Fibery number.** `api-surface-maintenance.md` has been added and merged three times before under exactly this name, alongside `api-project-settings.md`, `google-sheets-source-connector.md`, etc. The numeric prefix is the majority pattern, not a rule.
- **`'owox': minor` is the wrong bump — `@owox/api-client` is what changed.** `.changeset/config.json:5-19` puts both in the same `fixed` group; all workspaces are at `0.31.0` today.
- **The coverage matrix was loosened to make coverage easier to claim.** `git diff` on `api-coverage-matrix.mjs` is 28 insertions / 0 deletions, all inside the `COVERED_TARGETS` data map; every validator is byte-identical. Counts reconcile (139 endpoints, 16 covered).
- **The new OpenAPI test doesn't really pin operation IDs.** It does — verified by mutation, not inference: renaming the handler to `manualRunV2` fails the suite with the exact diff, and removal fails too. All four IDs the matrix deep-links are pinned. (Both Codex P2 findings — the 400 alignment and the removed `minimum`/`maximum` — were also independently verified as landed.)
- **`@IsObject()` / the `limit` cap / de-nullifying `status`/`type`/`runType` break existing consumers.** All disproved: `apps/web` sends objects and uses page sizes 5/20/50; the three columns are non-nullable `@Column()`s backfilled by `1762812222908-refactor-nullable-fields-in-data-mart-run.ts`. **IDOR on `runId`** also disproved — `getByIdAndDataMartId` and `getDataQualityDetailsByIdAndDataMartId` both filter on `dataMartId`, and cancel is an atomic conditional UPDATE, not a read-modify-write.

</details>

**CI parity** (run in the worktree): `tsc -p tsconfig.build.json` ✅ · `eslint` (backend changed files + `packages/api-client/src`) ✅ 0 errors · `prettier --check` ✅ · backend specs 23/23 ✅ · api-client 174/174 ✅ · `api-coverage-matrix.mjs` ✅. *Not attributable to this PR:* the spec-inclusive `tsc -b` reports 14 pre-existing errors in untouched files (`project-member-api-key.service.spec.ts`, 9 e2e specs under `test/`) — CI type-checks via `tsconfig.build.json`, which is clean.

<sub>Automated OWOX review · worktree multi-agent passes (5 agents + second wave) · all blockers adversarially verified; 6 candidates refuted.</sub>

packages/api-client/src/runs.ts
max-voloshyn: Addressed in 1eeaefda3. Shared path-ID validation rejects blank, whitespace-only, dot, and dot-dot Data Mart and run IDs before authentication or network access. Focused client coverage includes the forDataMart dot-collapse case plus get and cancel paths.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/data-mart-runs.ts
max-voloshyn: Addressed in 1eeaefda3. The public start options are now a discriminated union, and both client and backend runtime validation reject data without MANUAL_BACKFILL, incremental runs with data, manual backfills without data, and unknown keys. Generated OpenAPI uses two closed named oneOf variants. Runtime and OpenAPI tests pin the pairing, and the package typecheck passes.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/controllers/data-mart.controller.ts
max-voloshyn: Addressed in 1eeaefda3. The scoped controller again forwards caller values unchanged with only the existing 100/0 omission defaults; the shared cap is no longer used. A focused controller behavior test pins 500 and 100001 passthrough, and the docs and changeset now distinguish scoped behavior from project-wide caps.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/runs.ts
max-voloshyn: Addressed in 1eeaefda3. The project-run type now makes qualitySummary optional for compatibility with older deployments, and the parser normalizes a missing value only for shared validation. A regression test accepts the missing field while current present values remain validated; the docs and changeset were updated accordingly.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/controllers/spec/data-mart.api.ts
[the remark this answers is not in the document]
max-voloshyn: Follow-up addressed in 1eeaefda3. Manual RunDataMartService execution now checks for PUBLISHED state and returns BadRequest before connector dispatch; scheduled runs preserve their existing failed-history semantics. A focused behavior test asserts the BadRequest and zero dispatch calls.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/runs.ts
max-voloshyn: Addressed in 1eeaefda3. The shared 2 MiB body-parser ceiling now makes the endpoint-level 1 MiB DTO validation reachable: payloads above 1 MiB return 400, while payloads above 2 MiB map to 413 rather than 500. The 413 contract is owned once by the generated-document transform for every request-body operation. Focused e2e coverage pins the normal, 400, and 413 paths.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/runs.ts
max-voloshyn: Addressed in 1eeaefda3. Both list methods now call the shared pagination validator: limit must be a positive integer and offset a nonnegative integer. Unknown values, NaN, negatives, fractions, and zero limits are rejected before authentication or network access, with focused tests for both scopes.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/data-mart-runs.ts
max-voloshyn: Addressed in 1eeaefda3. Response validation is now open to additive rule, scope, and parameter fields and no longer re-derives server-owned rule keys, category-scope relationships, threshold coupling, or uniqueness invariants. It still validates known shapes, types, enums, nullability, and timestamps; a positive compatibility fixture covers additive fields and relaxed invariants.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/dto/presentation/run-data-mart-request-api.dto.ts
```diff
@@ -1,23 +1,95 @@
...
 export class RunDataMartRequestApiDto {
   /**
-   * Payload for the manual run. If not provided, the data mart will be run with the default payload.
-   * The payload is specific to the data mart definition type.
-   * For example, for a connector data mart, the payload is the connector configuration fields with unknown structure.
+   * Payload for the manual run. Omit it or select INCREMENTAL without data for an incremental run.
+   * MANUAL_BACKFILL requires connector-specific fields in data.
    */
   @IsOptional()
+  @IsObject()
```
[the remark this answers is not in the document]
max-voloshyn: Addressed in e202822ef. The request DTO now skips validation only when payload is undefined; an explicit null reaches the object and run-payload validators and is rejected before controller dispatch, while an omitted payload remains valid. The regression test was observed failing under IsOptional and now passes; the focused DTO/OpenAPI suites pass 14/14, the manual-run e2e suite passes 10/10, backend production typecheck and affected lint/format checks pass, and the independent delta review found no blocker.

🤖 Codex (GPT-5.6 Sol High)

docs/api/api-client.md
```diff
@@ -96,12 +96,11 @@ can also see shared Data Marts available for reporting or maintenance, subject t
...
-at 100. `offset` defaults to 0, floors finite fractions, falls back to 0 for non-finite or
-non-positive values, and caps at 100,000. The response has no total or next-page marker. Prefer a
-`limit` from 1 through 100, increment `offset` by the number of returned runs, and stop when a page
-contains fewer runs than the server-normalized effective limit or the next offset would exceed
+Pass a positive integer `limit` and a non-negative integer `offset` to page through the newest-first
+history. The client rejects invalid values before authentication or a network request. When omitted,
+the server defaults `limit` to 100 and `offset` to 0; it caps them at 100 and 100,000 respectively.
+The response has no total or next-page marker. Increment `offset` by the number of returned runs and
+stop when a page contains fewer runs than the requested limit or the next offset would exceed
 100,000. Because new runs can shift newest-first offset pages while a consumer is paging, deduplicate
```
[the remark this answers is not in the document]
max-voloshyn: Addressed in af379f7f5. The guide now compares each page with the server-effective limit using Math.min(requestedLimit, 100), or 100 when omitted. A full 100-row page therefore continues pagination even when the caller requested more than 100, while the existing 100000 offset boundary remains accurate. Prettier and Markdown lint pass, current-main ownership checks are clean, and the independent delta review found no blocker.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/bootstrap.ts
```diff
@@ -45,6 +46,8 @@ export async function bootstrap(options: BootstrapOptions): Promise<NestExpressA
     { logger }
   );
 
+  setupBodyParsers(app);
```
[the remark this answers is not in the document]
max-voloshyn: Addressed in d3d79e6e9. Both packaged Better Auth providers now scope JSON and URL-encoded parsers to /auth, so /api requests reach the backend 2 MiB parser. A focused packaged-server regression reproduces the prior 128 KiB 413, now passes, and separately verifies /auth JSON parsing remains intact. Provider builds, focused legacy tests, both provider typechecks and lints, OWOX lint, formatting, and changeset status pass.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/runs.ts
```diff
@@ -96,78 +46,80 @@ export type OWOXProjectRunHistoryOptions = {
...
-      ))
-  );
+function validateRunListOptions(options: unknown): asserts options is OWOXDataMartRunListOptions {
+  if (
+    !isRecord(options) ||
+    Object.keys(options).some(key => key !== 'limit' && key !== 'offset') ||
+    (options.limit !== undefined &&
+      (!Number.isInteger(options.limit) || (options.limit as number) <= 0)) ||
+    (options.offset !== undefined &&
+      (!Number.isInteger(options.offset) || (options.offset as number) < 0))
```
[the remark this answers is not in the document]
max-voloshyn: Addressed in 817bf20a5. Shared pagination validation now uses Number.isSafeInteger for limit and offset, so unsafe values are rejected before authentication or network access for both scoped and project list methods. The regression was observed failing after the unsafe limit reached authentication and now passes; the focused client pair passes 64/64 with typechecking, API-client lint, Prettier, Markdown lint, and Changesets status.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/dto/presentation/run-data-mart-request-api.dto.ts
```diff
@@ -1,23 +1,95 @@
...
+    {
+      name: 'isRunDataMartPayload',
+      validator: {
+        validate(value: unknown): boolean {
+          if (!isRecord(value)) return false;
+          if (Object.keys(value).some(key => key !== 'runType' && key !== 'data')) return false;
+          if (value.runType === 'MANUAL_BACKFILL') return isRecord(value.data);
+          return (
+            (value.runType === undefined || value.runType === 'INCREMENTAL') &&
+            value.data === undefined
```
romandubovyi: 🔴 **Blocking:** This branch requires `data` to be absent for an incremental run, but the web Run sheet can send it — manual runs from the UI will start failing with 400.

`ConnectorRunForm` registers the `data.*` inputs only while `runType === MANUAL_BACKFILL` (`ConnectorRunForm.tsx:140-178`), and the form is created with `useForm()` without `shouldUnregister` — react-hook-form's default is `false`, so unmounted field values are retained in form state, and `validateField` returns early for unmounted fields (`!mount → {}`), so the Run button stays enabled. Sequence inside a single open sheet:

1. user picks "Backfill (custom period)" and fills `StartDate` / `EndDate`
2. user switches back to "Incremental load"
3. user clicks Run

`handleSubmit` receives `{ runType: 'INCREMENTAL', data: { StartDate: '…', EndDate: '…' } }`, and both call sites forward it verbatim (`DataMartDetails.tsx:334-338`, `ConnectorDefinitionField.tsx:158`) into `dataMartService.runDataMart(id, payload)`, which posts `{ payload }`. On `main` the backend accepted this and the connector ignored `data`; with this DTO it now fails validation.

Second case on the same branch, line 25: selecting `MANUAL_BACKFILL` for a connector that declares no `MANUAL_BACKFILL` fields leaves `data` undefined and is rejected too. GoogleSheets is exactly that connector today — it is the only source under `packages/connectors/src/Sources` with no `MANUAL_BACKFILL` attribute.

Either relax the incremental branch to tolerate an extra/empty `data`, or clear `data` on `runType` change in the web form — though a web-only fix would still 400 users running an older cached bundle. Worth a regression test covering the backfill → incremental toggle, since the DTO spec currently only exercises hand-built payloads.

packages/api-client/src/runs.ts
```diff
@@ -185,10 +137,94 @@ function parseProjectRunHistory(response: unknown): OWOXProjectDataMartRunsRespo
...
+      ...(options.offset === undefined ? {} : { offset: String(options.offset) }),
+    };
+    const response = await this.requester.getJson<unknown>(
+      `${this.path}/runs`,
+      Object.keys(query).length === 0 ? undefined : query
+    );
+    if (
+      !isRecord(response) ||
+      !Array.isArray(response.runs) ||
+      !response.runs.every(isDataMartRun)
```
romandubovyi: ⚪ **Non-blocking:** `isProjectDataMartRun` (line 116) normalizes a missing `qualitySummary` to `null` specifically for compatibility with older self-hosted deployments, but the scoped methods call `isDataMartRun` / `isDataMartRunDetail` directly, and both require the key to be present: `data-mart-runs.ts:397` accepts only `null` or a compact summary, and `:405` accepts only `null` or a full detail object — `undefined` fails either way.

`qualitySummary` and `dataQuality` were both added to `DataMartRunResponseApiDto` / `DataMartRunDetailResponseApiDto` in `61fc1bc60` (#1474, 2026-07-30). Against a deployment older than that, `client.runs.list()` succeeds while `client.runs.forDataMart(id).list()` and `.get(runId)` (line 206) throw `OWOXApiError` on every run — from the same client, against the same server.

Applying the same `undefined → null` normalization in the scoped paths, or moving it into `isDataMartRun` / `isDataMartRunDetail`, would make the compatibility story uniform across the run surface.

apps/backend/src/data-marts/dto/presentation/run-data-mart-request-api.dto.ts
max-voloshyn: Fixed in `1b80d6584`. The backend request DTO and public client now accept object-valued `data` for incremental runs and allow `MANUAL_BACKFILL` to omit `data` when a connector has no backfill fields. Unsupported run types, unknown keys, and non-object `data` remain rejected, and the generated `oneOf` schema now publishes the compatible variants.

The regressions were observed failing before the implementation change. On the rebased head, the DTO and generated-OpenAPI suites pass 15/15 and the client lifecycle/history suites pass 67/67 with typechecking; backend production typechecking, affected lints, formatting, Markdown lint, and Changesets validation also pass.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/runs.ts
max-voloshyn: Fixed in `1b80d6584`. Scoped `list()` now normalizes a missing `qualitySummary` to `null`, while scoped `get()` normalizes both missing `qualitySummary` and missing `dataQuality` to `null` before applying the existing validators. Present malformed values are still rejected.

The compatibility test failed against the previous implementation and now passes as part of the 67/67 client lifecycle/history tests on the rebased head, including API-client typechecking and lint.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/data-mart-runs.ts
romandubovyi: 🔴 **Blocking:** This is reopened on `1b80d6584`. The relaxation made in response to my DTO comment ([r3736004315](https://github.com/OWOX/owox-data-marts/pull/1508#discussion_r3736004315)) went one layer too far and undid the discriminated union that closed this thread in `1eeaefda3`.

`data-mart-runs.ts:175-177` is now:

```ts
export type OWOXDataMartRunStartOptions =
  | { runType?: 'INCREMENTAL'; data?: Record<string, unknown> }
  | { runType: 'MANUAL_BACKFILL'; data?: Record<string, unknown> };
```

`data` no longer discriminates, so `start({ data: { StartDate: '2026-07-01' } })` with no `runType` matches the first member: it typechecks, passes `validateRunStartOptions` (`runs.ts:42-43` permits `runType === undefined` together with an object `data`), passes the backend validator, and returns `201` — with the backfill fields silently discarded, exactly as originally reported here.

The test delta confirms it: the rejection case changed from `start({ data: { value: 'silently ignored' } })` to `start({ data: [] })`, so the object-valued case is no longer covered and now simply passes.

The two layers need different strictness. My comment was about the **backend** DTO, because the web Run sheet posts through `data-mart.service.ts` and can legitimately carry retained `data` on an incremental run. `@owox/api-client` has no such constraint — nothing in the web app goes through it — so the client can stay strict:

```ts
export type OWOXDataMartRunStartOptions =
  | { runType?: 'INCREMENTAL'; data?: never }
  | { runType: 'MANUAL_BACKFILL'; data?: Record<string, unknown> };
```

plus the matching branch in `validateRunStartOptions`. That keeps the web form working and still fails a consumer who expects a backfill.

Raising this as blocking rather than non-blocking only because `runs.forDataMart()` ships as new public surface in this `minor` changeset: tightening the union after release is a breaking change for anyone who wrote `start({ data })`, whereas right now it is a two-line fix.

max-voloshyn: Fixed in `f361648f5`. The public API-client contract again uses a strict discriminated union: `data` is allowed only with an explicit `MANUAL_BACKFILL` run type. Runtime validation now rejects object-valued `data` when `runType` is omitted or `INCREMENTAL`, before authentication or network access.

The backend DTO remains intentionally compatible with the web form: retained object-valued `data` is accepted on incremental requests, and `MANUAL_BACKFILL` may omit `data` for connectors without backfill fields. Compile-time and runtime regressions cover the client distinction, while the backend DTO/OpenAPI tests cover the compatibility half.

After rebasing onto `ec28b5ea7`, the focused client lifecycle, project-history, and transport suites pass 81/81; the backend DTO/OpenAPI suites pass 15/15; typecheck, lint, formatting, Markdown lint, coverage validation, Changesets validation, and the final independent delta review are clean.

🤖 Codex (GPT-5.6 Sol High)

romandubovyi (approved):
Approving on `f361648f5`. All three issues I raised are closed, and every review thread on this PR is resolved.

**Blocker (reopened `start({ data })` no-op) — closed in `8bd1489ed`** at both levels, with the split the two layers actually need:

- Client type is a discriminated union again (`data?: never` on the incremental member), and `validateRunStartOptions` requires `data === undefined` for incremental runs.
- The backend DTO was deliberately left permissive, so the web Run sheet still works when it carries retained `data` after a backfill → incremental toggle, and when `MANUAL_BACKFILL` has no connector fields.

The `@ts-expect-error` pin on `{ data: { StartDate } }` is the right guard — a future relaxation of the union fails typecheck, not just a runtime assertion.

**Earlier findings**, both still in place after the rebase: the request DTO accepts the payload variants the existing run form produces, and the scoped `list()` / `get()` normalize missing `qualitySummary` / `dataQuality` so older self-hosted deployments behave like the project-wide history.

**Verified locally on the head**: `@owox/api-client` typecheck plus 213/213 tests, and 39/39 across the focused backend DTO, OpenAPI, Swagger, exception-filter, and run-service suites. CI is green on all jobs.

One note, not a request: `f361648f5` drops client-side pagination validation from project-wide `runs.list()` while the scoped `list()` keeps it. That asymmetry is intentional and documented in both the changeset and the client guide, so I am leaving it as decided.

PR #1509, opened 2026-08-07
fix(data-marts): SQL editor loses keystrokes on connected data marts

laskevych:
## Symptom

After changing an existing Data Mart's input source to SQL, spaces could not be typed into the query. On a freshly created Data Mart the editor worked fine.

## Root cause

**React Flow (the Joinable Data Marts diagram) swallows Space for the whole page.** Its key hooks listen for keydown on the entire document and call `preventDefault()` on their matches; the is-this-an-input check recognizes only `input`/`select`/`textarea`/`contenteditable`. Monaco in current versions accepts text through **EditContext** — a plain `div` that fails that check. The default pan shortcut, `panActivationKeyCode="Space"`, therefore killed every space before the editor ever saw it.

That is also why a new Data Mart seemed unaffected: the diagram mounts only when the Data Mart has relationships and the diagram view is selected (the view mode is persisted in localStorage). Letters survive because they match no shortcut; Backspace survives because Monaco handles command keys through its own keydown path; only plain Space depends on the browser's default action. The input-source change itself was never involved — the type is simply what people change on long-lived, well-connected Data Marts.

Reproduced and verified live against a running instance: same Data Mart, same diagram view, same keystrokes — `select*from` on the unfixed bundle, `select * from x` with the fix.

## What changed

- `RelationshipCanvas` no longer registers any global key shortcut: `panActivationKeyCode={null}` (joining the already-disabled `deleteKeyCode`), with a regression test.
- The `DataMartCodeEditor` container carries React Flow's `nokey` marker — the library's own opt-out — so no present or future canvas on the page can grab keystrokes typed into Monaco.

## Deliberate tradeoff (no separate fix planned)

The full-page **Models canvas** keeps React Flow's default global shortcuts, including Space-to-pan. That is safe today: the Models page has no editable surface, so there is nothing to steal keystrokes from, and the shortcut is genuinely useful there. The mechanism is unchanged though — if a Monaco or other EditContext-based editor ever lands next to that canvas (a side panel, an inline SQL view), this class of bug returns. The `nokey` marker on the SQL editor container protects that editor on any page; for new editable surfaces next to an xyflow canvas the rule is: either disable the canvas key codes or mark the editor with `nokey`. We are intentionally not opening a separate task for this.

## Also in this PR (found on the same path)

The Input Source form re-read the saved definition whenever the Data Mart object was rebuilt in the page context — schema actualization after a save, a publish, a relationship change — discarding everything typed since the last save. In the staged type-change state the same reset left Save silently disabled with a valid query on screen. The form now resets only when the definition type actually changes, and never over unsaved edits; the SQL editor is fully controlled, so the form value and the editor content cannot drift apart.

## Testing

Full `apps/web` data-marts suite (945 tests), lint, and type-check pass. Regression tests: the canvas registers no global key shortcuts; an unsaved query survives a Data Mart refresh in both the normal and the staged state; picking another type still re-shapes the form.

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useEffect, useCallback, useRef } from 'react';
 import { useForm, FormProvider, type SubmitHandler, type Resolver } from 'react-hook-form';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { DataMartDefinitionTypeSelector } from './form/DataMartDefinitionTypeSelector.tsx';
@@ -144,11 +144,22 @@ export function DataMartDefinitionSettings({
   const currentDefinition = watch('definition');
   const sqlCode = getSqlQueryFromDefinition(definitionType, currentDefinition);
 
+  const lastResetTypeRef = useRef<DataMartDefinitionType | null>(null);
+
   useEffect(() => {
-    if (definitionType) {
-      reset(getInitialFormValues());
-    }
-  }, [definitionType, reset, getInitialFormValues]);
+    if (!definitionType) return;
+
+    // Picking another type has to re-shape the form, so that reset always runs. Every other reason
+    // this effect re-runs is a Data Mart refresh handing us a new `definition` object — schema
+    // actualization after a save, a relationship change, a publish. Those must not overwrite what
+    // the user is typing: `getInitialFormValues` returns the *saved* definition (or an empty one
+    // while a type change is staged), so resetting mid-edit throws the unsaved query away.
+    const isTypeSwitch = definitionType !== lastResetTypeRef.current;
+    lastResetTypeRef.current = definitionType;
+    if (!isTypeSwitch && isDirty) return;
+
+    reset(getInitialFormValues());
+  }, [definitionType, reset, getInitialFormValues, isDirty]);
 
   useEffect(() => {
     if (!definitionType && !initialDefinitionType && preset?.definitionType) {
--- a/apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
@@ -203,7 +203,10 @@ export function DataMartDefinitionSettings({
         try {
           await updateDataMartDefinition(dataMartId, data.definitionType, data.definition);
           setShouldActualizeSchema(true);
-          reset(data);
+          // Re-baseline dirtiness on the snapshot that was actually saved, but keep the live
+          // values: anything typed while the request was in flight stays in the editor and
+          // simply leaves the form dirty again instead of being snapped back to the snapshot.
+          reset(data, { keepValues: true });
         } catch (error) {
           console.error('Failed to update data mart definition:', error);
         }
--- a/.changeset/6614-1-input-source-form-discards-unsaved-query.md
+++ b/.changeset/6614-1-input-source-form-discards-unsaved-query.md
@@ -2,10 +2,11 @@
 'owox': minor
 ---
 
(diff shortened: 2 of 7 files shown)


apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
```diff
@@ -144,11 +144,22 @@ export function DataMartDefinitionSettings({
...
+    if (!definitionType) return;
+
+    // Picking another type has to re-shape the form, so that reset always runs. Every other reason
+    // this effect re-runs is a Data Mart refresh handing us a new `definition` object — schema
+    // actualization after a save, a relationship change, a publish. Those must not overwrite what
+    // the user is typing: `getInitialFormValues` returns the *saved* definition (or an empty one
+    // while a type change is staged), so resetting mid-edit throws the unsaved query away.
+    const isTypeSwitch = definitionType !== lastResetTypeRef.current;
+    lastResetTypeRef.current = definitionType;
+    if (!isTypeSwitch && isDirty) return;
```
prokopetsv, line 159: 🟡 Non-Blocker: one residual wipe window remains — keystrokes typed while a save request is in flight. `onSubmit` (line 206, unchanged) calls `reset(data)` with the *submitted* snapshot when the request resolves, so anything typed between pressing Save and the response arriving is discarded, and the now fully-controlled editor snaps back visibly. This guard can't catch it because `reset(data)` legitimately clears `isDirty`.

Pre-existing behavior and a narrow window, so nothing to change in this PR — noting it only because the changeset says the query survives "right after saving": that holds for text typed *before* Save, not during the request.

prokopetsv (commented):
## Review summary

The fix is correct and well-scoped. Verified against the installed packages: `@xyflow/react` does default `panActivationKeyCode` to `'Space'` with a document-level listener, and `@xyflow/system`'s `isInputDOMNode` does honor the `.nokey` marker (`target.closest('.nokey')`), so both halves of the keystroke fix — disabling the canvas shortcut and marking the editor container — are real library mechanisms, not conventions. The form-reset guard is sound: the save path clears `isDirty` via `reset(data)` before the Data Mart refresh lands, type switches always re-shape the form, and a clean form still picks up externally refreshed definitions. Making `DataMartCodeEditor` fully controlled removes the state mirror that allowed editor/form drift; both call sites of the renamed prop are updated. Both changed test suites pass (30/30) and the three regression tests match the claimed behaviors.

### Findings

**🔴 Blocker** — none.

**🟡 Non-Blocker** — one residual wipe window during an in-flight save (see inline comment). Pre-existing, narrow, fine to leave out of this PR.

No architectural, security, or performance concerns. The per-keystroke `onChange` → form update path is unchanged from before the refactor.

prokopetsv (approved):
✅ **Approved.**

Verified the fix against the installed packages, not just the PR description: `@xyflow/react` does default `panActivationKeyCode` to `'Space'` with a document-level listener, and `isInputDOMNode` in `@xyflow/system` honors the `.nokey` marker — so disabling the canvas shortcut plus marking the editor container are both real library mechanisms, giving defense in depth. The form-reset guard is sound (save clears `isDirty` before the refresh lands, type switches always re-shape the form, a clean form still adopts refreshed definitions), and the fully controlled editor removes the state mirror that let the form and editor drift. Both changed suites pass (30/30), regression tests cover all three claimed behaviors.

One 🟡 non-blocker left inline: keystrokes typed while a save request is in flight are still wiped by `reset(data)` in `onSubmit` — pre-existing, narrow window, fine outside this PR.

apps/web/src/features/data-marts/edit/components/DataMartDefinitionSettings/DataMartDefinitionSettings.tsx
laskevych: Good catch, and thank you for verifying the library mechanisms directly. Rather than softening the changeset claim, the window is now closed: `onSubmit` re-baselines with `reset(data, { keepValues: true })`, so settling the request updates the dirty baseline to the saved snapshot without touching the live values — anything typed mid-flight stays in the editor and marks the form dirty again (react-hook-form recomputes `isDirty` via `_getDirty()` on this path, so Save re-enables immediately, not on the next keystroke). Covered by a regression test that edits while the save promise is pending. beab19267

PR #1510, opened 2026-08-07
fix(api-client): expose in/not_in filters and new relative-date presets in traversal types

laskevych:
Follow-up to #1435 (task 6761-1), prompted by an external user asking whether the `in` filter takes `'Germany, Ukraine'` or `['Germany', 'Ukraine']` — and where that is documented.

Release 0.31.0 taught the traversal HTTP API the `in`/`not_in` operators and the `this_week`/`last_week`/`this_quarter`/`last_quarter`/`next_n_days` relative-date presets, but the public `@owox/api-client` types were never updated: a TypeScript caller writing `operator: 'in'` got a compile error even though the request works at runtime (the client base64url-encodes the rules without validating them).

## Changes

- **`TraverseDataFilterRule`** gains the `in`/`not_in` branch — `value: (string | number)[]`, matching the backend contract (1..500 values, same-type list, booleans rejected in favor of `is_true`/`is_false`).
- **`TraverseDataRelativeDatePreset`** gains `this_week`, `last_week`, `this_quarter`, `last_quarter`, and `next_n_days`.
- **Tests**: the typed-options compile check now covers `in`/`not_in` and the new presets, plus two new `@ts-expect-error` cases locking in that a comma-separated string and a boolean list are rejected (ts-jest typechecks these during transform).
- **Docs (`docs/api/api-client.md`)**: the `traverseData` example now includes an `in` filter, and a new value-shape-by-operator table documents every operator family — array-only `in`/`not_in` with the 500-value cap and same-type rule, `between` bounds, the full relative-date preset list (ISO weeks, `n` up to 3650, `last_n_days`/`next_n_days` include today), no-value operators, and the NULL-inclusive semantics of negative operators.
- **OpenAPI**: the `filter` query parameter description gains an `in` example stating the array-of-1..500-same-type-values rule.

## Verification

- `@owox/api-client`: 163 tests green behind a new `tsc --noEmit -p tsconfig.test.json` gate in the test script — review found the compile-time cases were previously inert (`isolatedModules` puts ts-jest in transpile-only mode), so the gate now enforces the positive and `@ts-expect-error` cases and surfaced 9 latent type errors in neighboring test files, fixed here
- Backend: `http-data.openapi.spec.ts` green (7 tests)
- ESLint + markdownlint clean; changeset included (`6761-1`, `owox: minor`)


packages/api-client/src/data-marts.test.ts
```diff
@@ -29,6 +37,10 @@ const invalidTraverseDataOptions: TraverseDataOptions = {
   filter: [
     // @ts-expect-error unsupported filter operators must be rejected by the public options type
     { column: 'Event Date (local)', operator: 'approximately', value: '2026-01-01' },
+    // @ts-expect-error in/not_in take an array of values, not a comma-separated string
```
prokopetsv, line 54: 🔴 **Blocker — these compile-time assertions are never typechecked in CI; the PR's only test mechanism is inert.**

`packages/typescript-config/base.json` sets `"isolatedModules": true`, and ts-jest reads that flag from the resolved tsconfig and switches to transpile-only mode — types are stripped, never checked, and `@ts-expect-error` is never evaluated. Nothing else covers this file: `packages/api-client/tsconfig.json` excludes `**/*.test.ts` from `tsc`, the package ESLint config ignores test files, and CI runs only the jest script for this package.

Empirically reproduced: appending `const deliberateTypeError: number = 'not a number';` to this file and running `npm test` still reports **163 passed**. Inverting the existing `@ts-expect-error` case (line 43) also stays green.

Impact: the two new `@ts-expect-error` cases and the positive `TraverseDataOptions` cases assert nothing in CI, so the PR description's claim that "ts-jest performs full typechecking, so the positive and `@ts-expect-error` cases are both enforced" does not hold today, and any future regression to these unions merges green. The root cause is pre-existing (the earlier `approximately` case is equally inert), but this PR's entire test contribution and its Verification section rest on it.

Minimal fix: add a `tsconfig.test.json` in this package that includes test files and wire `tsc --noEmit -p tsconfig.test.json` into the test/CI script (or enable ts-jest `diagnostics` in the transform options for this package). Alternatively, correct the PR description and track the typecheck gate separately — but as written the tests promise a guarantee they don't provide.

packages/api-client/src/data-marts.ts
```diff
@@ -131,6 +136,13 @@ export type TraverseDataFilterRule = (
       column: string;
       operator: 'is_empty' | 'is_not_empty' | 'is_null' | 'is_not_null' | 'is_true' | 'is_false';
     }
+  | {
+      column: string;
+      operator: 'in' | 'not_in';
+      // The API accepts 1..500 values, all of the same type (all strings or all
```
prokopetsv: 🟡 **Non-Blocker — `value: string[] | number[]` would encode the same-type rule this comment documents.**

The backend refine rejects mixed-type lists, but `(string | number)[]` accepts them, so `value: ['Germany', 5]` compiles and only fails as a 400 at request time — decoupled from where the array was built. `string[] | number[]` expresses the homogeneity invariant at zero runtime cost: both positive cases in the test file still compile, both `@ts-expect-error` cases still error, and `[]` assignability is unchanged. The only caller that regresses is one holding a value pre-annotated as `(string | number)[]`, which is an unlikely shape for this API and easily cast.

Non-blocking because the runtime failure is loud (pre-stream 400 with the zod message via `OWOXApiError`), and the sibling `between` branch has the same looseness today — but here the fix is one line.

packages/api-client/src/data-marts.test.ts
```diff
@@ -13,11 +13,19 @@ const typedTraverseDataOptions: TraverseDataOptions = {
...
   filter: [
     { column: 'Event Date (local)', operator: 'gte', value: '2026-01-01' },
+    { column: 'Country', operator: 'in', value: ['Germany', 'Ukraine'] },
+    { column: 'Channel', operator: 'not_in', value: ['(direct)'] },
     {
       column: 'Event Date (local)',
       operator: 'relative_date',
       value: { kind: 'last_n_months', n: 3 },
     },
+    { column: 'Event Date (local)', operator: 'relative_date', value: { kind: 'this_week' } },
```
prokopetsv: 🟡 **Non-Blocker — three of the five new preset kinds are never referenced, and three expressible negative cases are untested.**

`last_week`, `this_quarter`, and `last_quarter` appear in neither options object, so a typo in one of those literals would slip through even with typechecking in place. Additionally, these three shapes are real compile errors today but have no `@ts-expect-error` lock: `value` omitted on an `in`/`not_in` rule, `n` omitted on `next_n_days`, and `n` supplied on a no-arg preset such as `this_week` (TypeScript narrows excess-property checking to the matched discriminant branch, so that case does error).

Worth adding once the typecheck gate from the blocker comment actually runs — until then, additional cases here assert nothing.

prokopetsv (commented):
## Multi-perspective review

Six specialized passes: contract correctness, docs/comment accuracy, test coverage, silent-failure hunt, type design, simplification. Every factual claim in the PR was verified against the backend source of truth; the one blocker is about test enforcement, not the shipped types.

**Verdict:** the type changes and documentation are correct. One 🔴 Blocker and two 🟡 Non-Blockers, all inline.

**Verified good (no action needed):**

- Every docs/OpenAPI/changeset claim checks out against `filter-config.schema.ts` and all five storage clause renderers: the 1..500 cap (`IN_LIST_MAX_VALUES`), the same-type refine, boolean rejection, `n <= 3650` (`RELATIVE_DATE_MAX_N`), ISO weeks (BigQuery `ISOWEEK`, Snowflake `DAYOFWEEKISO`), `last_n_days`/`next_n_days` including today, and the NULL-inclusive negatives (`neq`/`not_contains`/`not_regex`/`not_in`).
- The new union branches exactly match the backend contract — no supported operator or preset is still missing from the client types, and nothing the types now allow is rejected by the backend beyond what TypeScript cannot express (cardinality, homogeneity — see inline).
- `placement`/`function` compose correctly onto the new `in`/`not_in` branch: the `& TraverseDataFilterPlacement` intersection distributes over the whole union, mirroring the backend's `z.intersection`.
- No runtime code changed, and a type-blessed-but-invalid rule fails loudly: the backend validator rejects it pre-stream and the client surfaces a thrown `OWOXApiError` with the zod message before the NDJSON reader is constructed. No silent-failure paths.
- No redundancy worth consolidating across the four prose locations; the PR is minimal.

packages/api-client/src/data-marts.test.ts
laskevych: Confirmed — reproduced the inert gate exactly as described (deliberate type error and an inverted `@ts-expect-error` both stayed green). Fixed in 12015b995: added `tsconfig.test.json` (includes test files, jest types) and wired `tsc --noEmit -p tsconfig.test.json` into the package's `test` script, so CI now runs the gate. It immediately surfaced 9 latent `noUncheckedIndexedAccess` errors in the neighboring test files (`insight-templates`, `project-run-history`, `reports`, `search`), fixed with non-null assertions. Re-verified both probes now fail the build, and corrected the PR description's typechecking claim.

packages/api-client/src/data-marts.ts
laskevych: Adopted in 12015b995 — `value` is now `string[] | number[]`, and a new `@ts-expect-error` case locks in that a mixed `['Germany', 5]` list fails at compile time. Left `between` as is per the comment.

packages/api-client/src/data-marts.test.ts
laskevych: Done in 12015b995 — positive cases added for `last_week`, `this_quarter`, and `last_quarter`, plus the three suggested negative locks (`in` without `value`, `next_n_days` without `n`, `n` on `this_week`) and the mixed-type list case. All assert for real now that the typecheck gate from the blocker runs in the test script.

romandubovyi (commented):
Reviewed the contract sync itself against `apps/backend/src/data-marts/dto/schemas/filter-config.schema.ts`: the new `in`/`not_in` branch and all five added relative-date presets match the backend schema exactly, and the NULL-inclusive claim in the guide is backed by `inListSql()` in `sql-clause-renderer.ts`. Blast radius is clean — the only consumer outside the package, `apps/ctl/src/commands/data-marts/stream.ts`, casts parsed JSON and gains nothing to update.

Three non-blocking notes on the new documentation prose, all about constraints that are stricter or differently-shaped than what the text promises.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Kimi CLI + Grok CLI (Grok 4.5) + Claude Code (Opus 5)

docs/api/api-client.md
```diff
@@ -304,6 +307,22 @@ Do not pass comma-separated column lists. Column names are opaque strings and ma
...
+| `in`, `not_in` | Array of values — match any of / none of the listed values | `{ column: 'Country', operator: 'in', value: ['Germany', 'Ukraine'] }` |
+| `between` | `{ from, to }` bounds of the same type | `{ column: 'Revenue', operator: 'between', value: { from: 100, to: 500 } }` |
+| `relative_date` | A preset object, see below | `{ column: 'Date', operator: 'relative_date', value: { kind: 'last_n_days', n: 7 } }` |
+| `is_empty`, `is_not_empty`, `is_null`, `is_not_null`, `is_true`, `is_false` | No `value` | `{ column: 'Country', operator: 'is_null' }` |
+
+`in` and `not_in` take an **array** of 1 to 500 values, never a comma-separated string — `value: 'Germany, Ukraine'` is rejected, exactly like comma-separated column lists, because stored values are opaque and may themselves contain commas. All values in one list must be the same type (all strings or all numbers); booleans are not allowed — use `is_true` / `is_false` instead. Which operators a given column accepts depends on its field type; the API rejects a mismatched operator with a validation error naming the allowed set.
```
romandubovyi: ⚪ **Non-blocking:** The 500-value cap is `IN_LIST_MAX_VALUES`, but on this endpoint it is not the binding limit: `filter` is also capped at `HTTP_DATA_MAX_ENCODED_PARAM_LENGTH = 8192` base64url characters (`apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts:7,55`), i.e. ~6144 bytes of JSON. 500 values only fit for values of roughly 9 characters or less — a list of UUIDs tops out near 155 before the request is rejected, and the resulting error is the generic zod string-length message, not a value-count message. Worth one sentence here (and in the OpenAPI description at `http-data.api.ts:80`, which states the same 1..500 rule) noting that the encoded `filter` parameter length can bind before the 500-value cap does.

romandubovyi: ⚪ **Non-blocking:** This sentence says a type-mismatched operator is rejected "with a validation error naming the allowed set". On the HTTP path this guide documents (`traverseData` / raw `filter` query), `OutputControlsValidatorService.throwIfInvalid` returns `message: 'Output controls validation failed'` with `details.errors[]` entries shaped as `{ code: 'INVALID_OPERATOR_FOR_TYPE', column, type, operator }` — no allowed-operator list. That list is assembled only in the MCP error mapper (`apps/backend/src/ee/mcp/tools/output-controls-error.mapper.ts:192`). Suggest rephrasing to what the HTTP caller actually receives (an `INVALID_OPERATOR_FOR_TYPE` error naming the column, its type, and the rejected operator) so the client docs do not promise MCP-only messaging.

docs/api/api-client.md
```diff
@@ -304,6 +307,22 @@ Do not pass comma-separated column lists. Column names are opaque strings and ma
...
+
+`relative_date` presets: `{ kind }` with `today`, `yesterday`, `this_week`, `last_week`, `this_month`, `last_month`, `this_quarter`, `last_quarter`, or `this_year`, and `{ kind, n }` with `last_n_days`, `last_n_months`, or `next_n_days` (`n` up to 3650). Weeks are ISO weeks (Monday start) on every storage, and `last_n_days` / `next_n_days` both include today.
```
romandubovyi: ⚪ **Non-blocking:** `n` is documented only as "up to 3650", but the backend validates `z.number().int().positive().max(RELATIVE_DATE_MAX_N)` (`apps/backend/src/data-marts/dto/schemas/filter-config.schema.ts:32-43`), so `0`, negatives, and non-integers are rejected too. The client type is `n: number`, so this paragraph is the only place a caller can learn the constraint — suggest "a positive integer up to 3650".

romandubovyi (approved):
No blockers from my side.

`prokopetsv`'s blocker is addressed by `12015b99` and I verified the gate empirically rather than taking the script change on trust — checked out the PR head and ran `tsc --noEmit -p tsconfig.test.json`:

- clean on the PR as-is;
- appending `const deliberateTypeError: number = 'not a number';` to `src/data-marts.test.ts` now fails with `TS2322` (the exact reproduction that stayed green before);
- turning the new `value: 'Germany, Ukraine'` case into a valid one fails with `TS2578: Unused '@ts-expect-error' directive`.

`tsconfig.test.json` extends the base and only widens `exclude`, so it keeps `include: ["src/**/*"]` — the rewired `typecheck` script is a strict superset of the old `tsc --noEmit`, not a narrowing. The `value: string[] | number[]` change lands the same-type invariant in the type as well.

My three inline notes above are documentation-prose only and are the author's call.

docs/api/api-client.md
laskevych: Fixed in 53c6f4a51 — both the guide and the Swagger `filter` description now note that the whole encoded parameter is capped at 8192 base64url characters (~6 KB of JSON), so with long values the length cap can bind before the 500-value cap; the Swagger text interpolates `HTTP_DATA_MAX_ENCODED_PARAM_LENGTH` so it cannot drift.

laskevych: Right — this was corrected in 1f988acd9: the guide now says the 400 names the column, its field type, and the rejected operator (`INVALID_OPERATOR_FOR_TYPE`), without promising the MCP-only allowed-set messaging.

docs/api/api-client.md
laskevych: Fixed in 53c6f4a51 — the preset paragraph now reads "`n` is a positive integer, up to 3650".

PR #1511, opened 2026-08-07
feat(web): give the Joinable Data Marts diagram the Models canvas view settings

laskevych:
## Summary

Phase 2 of the Joinable Data Marts / Models canvas design parity (PM feedback on #1479, task 6776; phase 1 was #1507). The diagram's gear button returns — now carrying the exact same view settings as the Models canvas, rendered by one shared component:

- **Shared `CanvasSettingsPanel`** (`shared/canvas/canvas-settings-panel.tsx`): View Compact/Detailed, Layout algorithm Horizontal/Vertical, Show join fields, Object labels with Check/Uncheck-all shortcuts. Both ModelCanvas and RelationshipCanvas render it inside their gear popovers, so the two designs cannot drift apart.
- **Shared modules moved to `shared/canvas/`** (git-mv, history preserved): `object-labels`, `canvas-direction`, `path-point`, `dagre-layout` (+`estimateEdgeLabelDimensions`), plus new `view-mode` and `erd-fields`/`erd-fields-section` (the ERD field-row model and card body extracted from `ModelCanvasFlowNode`).
- **RelationshipCanvas** now lays out with the same dagre engine (honoring Horizontal/Vertical), supports Detailed cards with field rows from the blendable schema (name + raw type, hidden-for-reporting rows dimmed, collapsed to 4 with "+N more"), draws join-condition labels on edges, and honors object-label toggles (the alias badge always stays — it is join configuration, not an object label).
- **Settings state** lives in `DataMartRelationshipsContent` (global per-browser LS keys `relationship-canvas-view-mode/layout/show-join-fields/object-labels`), keeping the inline and fullscreen diagram instances in sync.

Known scope cuts: the root card stays compact in Detailed view (native root fields are untyped in the schema payload), and rows carry no primary-key icons (PK info is not part of the blendable schema).

## Test plan

- 195/195 tests green across `DataMartRelationships`, `model-canvas` and `shared/canvas` (new: gear settings delegation, Detailed card sizing, join-label toggling; parent test now asserts view props reach the canvas).
- `tsc`, eslint and prettier clean on all touched areas.
- Manual: switch View/Layout/Show join fields/Object labels on both canvases and compare; verify inline vs fullscreen sync and per-browser persistence.


apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx
```diff
@@ -559,48 +684,47 @@ function buildRelationshipFlow({
...
+    width: widths.get(nodeKey) ?? NODE_W,
+    height: heights.get(nodeKey) ?? TGT_H,
+  }));
+  const dagreEdges: DagreLayoutEdge[] = edgeInfos.map(edge => ({
+    id: `${edge.sourceId}->${edge.targetId}`,
+    sourceId: edge.sourceId,
+    targetId: edge.targetId,
+    label: estimateEdgeLabelDimensions(edge.joinLabel),
+  }));
+  const { positions } = runDagreLayout(dagreNodes, dagreEdges, direction);
```
prokopetsv, line 716: 🔴 **Blocker — toggling any new view setting discards the user's pan/zoom and clears their selection.**

`getRelationshipFlowGraphIdentity` (line 503, unchanged) fingerprints every node's `position.x/y`, `width`, and `height`. This PR wires `viewMode`, `direction`, and `showJoinFields` into exactly that geometry — card sizing varies with view mode, and this `runDagreLayout(..., direction)` call repositions every node when direction flips or join-label dimensions change. So a purely cosmetic toggle now changes the identity string, which fires the reset effect (lines 903–910: selection cleared, `userInteractedRef.current = false`), and the follow-up refit effect (lines 1000–1007) sees `userInteractedRef.current === false` and calls `fitFull()`.

User-facing impact: zoom into a subgraph, select a node, then flip Compact↔Detailed, Horizontal↔Vertical, or "Show join fields" — viewport and selection are wiped, every time. ModelCanvas avoids this by fingerprinting topology separately from cosmetics, so this also breaks the design parity the PR exists to deliver.

Fix: build the identity from node/edge membership and structural flags (`isCycleStub`, `isBlocked`, `warning`, …) only — drop `position`/`width`/`height` from the fingerprint. Note no current test catches this: the new settings tests assert the `on*Change` callbacks fire, never viewport/selection state after a toggle.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
```diff
@@ -284,6 +302,62 @@ export function DataMartRelationshipsContent({
...
+  const handleObjectLabelsChange = useCallback((next: ObjectLabelsHidden) => {
+    setObjectLabels(next);
+    storageService.set(CANVAS_OBJECT_LABELS_KEY, serializeObjectLabelsHidden(next));
+  }, []);
+
+  // ERD rows for the Detailed canvas view, keyed by aliasPath. The root
+  // mart's native fields are untyped in the schema payload, so the root card
+  // stays compact. Primary-key info is not part of the blendable schema —
+  // rows render without key icons.
+  const canvasFieldsByAliasPath = useMemo(() => {
```
prokopetsv, line 347: 🟡 **Non-Blocker — a failed blendable-schema fetch now silently defeats the entire Detailed view.**

The `.catch(() => setBlendableSchema(null))` swallow in `fetchBlendableSchema` (lines 220–238) is pre-existing, but this new consumer gives it a much bigger blast radius: with `blendableSchema` null this map stays empty, every card gets `fields: []`, and `showFieldRows` is false everywhere — so "Detailed" renders identically to "Compact" with no error, no log, and no way to distinguish "schema call failed" from "these marts have no fields". Before this PR the same swallow only degraded the field-count badges.

Suggested minimum: log the error in the catch; ideally also surface a lightweight hint (or disable the Detailed option) when the schema is unavailable.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
```diff
@@ -284,6 +302,62 @@ export function DataMartRelationshipsContent({
...
+    const map = new Map<string, ErdCardField[]>();
+    if (!blendableSchema) return map;
+    for (const field of blendableSchema.blendedFields) {
+      const rows = map.get(field.aliasPath) ?? [];
+      if (!map.has(field.aliasPath)) map.set(field.aliasPath, rows);
+      rows.push({
```
prokopetsv, line 357: 🟡 **Non-Blocker — this mapper ignores the blended field's real alias and its sibling's `UNKNOWN` filter.**

`alias: field.originalFieldName` fills the exact property `ErdCardFieldsSection` renders (`{field.alias}`, plus the tooltip), so a field renamed via alias always shows its raw column name on this canvas — while the Models canvas mapper (`mapSchemaFields`: `field.alias?.trim() ? field.alias : field.name`) shows the rename. If raw names are intentional (the PR body says "name + raw type"), this still contradicts `ErdCardField.alias`'s documented contract ("businessName / displayName when set") and makes the two canvases render the same shared component differently — worth either mapping `field.alias` with the same fallback, or documenting the divergence at this call site.

Separately, `connectedFieldCounts` (lines 255–264) filters `type === 'UNKNOWN'` out of the badge count, but this mapper does not — so a Detailed card can render a row whose type column literally says `UNKNOWN` while the same card's field-count badge disagrees with the number of visible rows. Applying the same exclusion here keeps the card self-consistent.

A shared `toErdCardField(...)` factory in `erd-fields.ts` would prevent the two per-source mappers drifting on exactly this kind of fallback logic again.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.test.tsx
```diff
@@ -388,6 +396,80 @@ describe('RelationshipCanvas filters', () => {
...
+    reactFlowHarness.latestProps = null;
+    reactFlowHarness.store.width = 800;
+    reactFlowHarness.store.height = 600;
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('offers the shared canvas settings in the gear popover and delegates changes', async () => {
```
prokopetsv, line 417: 🟡 **Non-Blocker — the new tests stop at the data/callback layer; three render-level contracts are unguarded.**

1. Detailed field rows are never rendered in any relationship-canvas test: sizing (`relationshipNodeHeight`) and rendering (`data.fields`) come from two separate Map operations in `RelationshipCanvas.tsx` (~lines 693–703 vs 751), and this suite only asserts the mocked ReactFlow's `width`/`height`. A future desync (wrong key, stale closure) ships correctly-sized but empty Detailed cards with CI green.
2. Object-label toggles on the rendered card are untested for this canvas — `RelationshipFlowNode` has its own `withSource`/`withFieldCount`/`withStatus` implementation (lines 254–257), parallel to the one `ModelCanvasFlowNode.test.tsx` protects. Nothing renders a node with all labels hidden and asserts the badge/count/status dot vanish while the alias badge stays — the exact "always stays" rule the PR calls out.
3. The fullscreen `<RelationshipCanvas>` block in `DataMartRelationshipsContent.tsx` (lines ~799–826) is never mounted in its test file, so the inline/fullscreen settings sync the PR advertises — wired via hand-duplicated props in two JSX blocks — is verified nowhere; a future setting added to only one block would slip through.

prokopetsv (commented):
## Multi-perspective review

Seven specialized passes: contract correctness, React behavior, test coverage, silent-failure hunt, type design, comment/claim accuracy, simplification. The PR has no prior reviews or comments, so nothing below duplicates existing feedback. Every finding was verified against the PR-side code; the two most serious were independently re-verified by reading the exact code paths.

**Verdict:** the refactor itself is in very good shape — one 🔴 Blocker (inline) on how the new settings interact with the viewport-reset logic, plus three 🟡 inline comments and one 🟡 cleanup below.

**Verified good (no action needed):**

- ModelCanvas's four `model-canvas-*` localStorage keys are byte-identical before/after — existing users' saved settings do not reset. Moved modules are byte-identical renames (`dagre-layout` 82% only from the verbatim `estimateEdgeLabelDimensions` move) with git history preserved.
- `CanvasSettingsPanel` is a faithful 1:1 extraction (same ARIA attributes, class strings, check/uncheck-all semantics), and all nine of its props are required — the compiler itself keeps both canvases identically wired, which closes the drift risk this PR exists to eliminate.
- Inline and fullscreen instances are literal siblings sharing one state source; the Radix dialog renders through a portal in the same fiber tree, so no remount/desync. `useId()` prevents duplicate switch ids when both mount.
- The `relationship-canvas-*` parsers are validating string checks (no `JSON.parse`), matching the repo's `parseRelationshipStatusFilter` precedent — corrupt values cannot throw or reset anything silently.
- All PR-description claims check out in code (alias badge exemption, collapse-at-4 via named constant, hidden-row dimming, exact LS key strings, join-label format); changeset follows repo rules.
- Independently re-ran the suites on the PR branch: 195/195 in the touched dirs, 944 tests across the data-marts tree, `tsc` and eslint clean.

🟡 **Non-Blocker — finish the de-dup this PR started.** `ModelCanvas.tsx:277` still inlines `=== 'erd' ? 'erd' : 'compact'` instead of the `parseCanvasViewMode` helper this PR created (its sibling call site uses it); if `CanvasViewMode` ever grows a variant, this call site silently keeps two-value behavior. Related, take or leave: the `model/erd-node.ts:9-16` re-export shim now exists only so the untouched `erd-node.test.ts` keeps compiling — moving the `orderFields`/`collapsedRowCount` test blocks to a new `shared/canvas/erd-fields.test.ts` would give the shared module its own test home and let the shim shrink; and the four settings' `useState` + `storageService` plumbing is duplicated across both components with only the key prefix differing (a small shared `usePersistedCanvasSetting` hook would close it).

prokopetsv (changes requested):

apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx
laskevych: Fixed in d0f73b742 (with a first pass in f22760a47). `getRelationshipFlowGraphIdentity` now fingerprints membership + structural flags only — `position`/`width`/`height` are dropped, so Compact↔Detailed, Horizontal↔Vertical and Show join fields no longer clear the selection or reset the viewport. The reset effect additionally keeps a selection whenever its node/edge id survives a structural change. New test: `keeps the selection and viewport across view-settings toggles` asserts fitView is not re-invoked and the selected node stays selected across all three toggles.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
laskevych: Done in d0f73b742 — the catch now logs `Failed to load blendable schema` with the error (matching the repo's console-trace pattern) before falling back to null, so a Detailed view that renders compact everywhere is diagnosable. Left the UI hint / disabling the Detailed option out for now — that needs a PM/UX call (same bucket as the drafts-excluded-from-schema case); happy to follow up if you want it in this PR.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
laskevych: Both fixed: f22760a47 added the `UNKNOWN` exclusion (mirroring `connectedFieldCounts`, with a comment tying the two together) and the alias mapping; d0f73b742 aligned the fallback to exactly the Models canvas mapper semantics (`alias.trim() ? alias : originalFieldName`). Skipped the shared `toErdCardField` factory for now — the two mappers consume different DTOs (`DataMartSchemaField` vs `BlendedField`) and only the one-line alias fallback is common; extracting it felt like abstraction without leverage. Can revisit if a third mapper appears.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.test.tsx
laskevych: All three covered in d0f73b742: (1) the Detailed sizing test now also asserts the six rows reach `node.data.fields`, and `RelationshipFlowNode.test` renders them — 4 collapsed rows, `+2 more fields` expand-in-place, none in Compact; (2) object-label render tests on the relationship card — `ALL_HIDDEN` drops the field count while the alias badge stays (the "always stays" rule), `NOTHING_HIDDEN` shows it; (3) a fullscreen-sync test opens the dialog via `onRequestFullscreen`, asserts the second instance carries the same state (and no fullscreen handler of its own), then changes a setting through it and asserts both the shared state and the persisted LS key update.

laskevych:
Re the review-body 🟡 (finish the de-dup): `parseCanvasViewMode` at the ModelCanvas call site — done in f22760a47, along with a shared `CanvasSettingsPopover` (gear shell + `useId`, removing the hardcoded switch id), a shared `EdgeJoinLabel`, and shared `ERD_NODE_WIDTH`/`erdFieldsBodyHeight` so the card geometry has one home. The two take-or-leave items I'm leaving for a follow-up: relocating the `orderFields`/`collapsedRowCount` tests into a `shared/canvas/erd-fields.test.ts` (and shrinking the `erd-node` shim), and a `usePersistedCanvasSetting` hook — both are pure moves with no behavior change, and this PR is already large. Blocker + inline comments are addressed in f22760a47 + d0f73b742 (see per-thread replies); suites are green (201 tests in the touched dirs).

prokopetsv (approved):
All review points addressed in `f22760a47` + `d0f73b742` — verified in code: `graphIdentity` is structural-only so settings toggles keep viewport and selection, the schema-fetch failure now leaves a trace, the field mapper matches the Models-canvas alias fallback and `UNKNOWN` handling, and the new render-level tests cover collapsed rows, the alias-badge rule, and fullscreen settings sync. Re-ran the touched suites on `d0f73b742`: 201/201 green, `tsc` clean. Nice extras: structural-identity comment, `updateNodeInternals` on expand, and the `CanvasSettingsPopover`/`EdgeJoinLabel` dedup. LGTM 🚢

PR #1512, opened 2026-08-07
fix(connectors): bump LinkedIn and Shopify off sunset-risk API versions

prokopetsv:
# Problem

The LinkedIn Ads and LinkedIn Pages connectors send every request with a hardcoded `LinkedIn-Version: 202509` header. LinkedIn's own versioning docs confirm that `202507` — just two months older — has already sunset, and each version is guaranteed a minimum support window of only one year, so `202509` was on track to start failing requests around September 2026. The Shopify connector had the same problem: it called Admin API version `2025-10`, which had roughly two months of support left before its own cutoff. This is the same failure mode that recently broke the Google Ads connector when its hardcoded `v21` was sunset by Google mid-run with no grace period.

# Solution

Bumped both connectors to their current stable API versions per each vendor's official documentation, rather than waiting for them to fail in production like Google Ads did:

- LinkedIn Ads and LinkedIn Pages: `LinkedIn-Version` header updated from `202509` to `202607`, the latest published version per Microsoft Learn's LinkedIn Marketing API docs
- Shopify: Admin API version updated from `2025-10` to `2026-07`, the current version recommended for production per Shopify's official release schedule
- Also swept each connector's field-schema documentation links (`ShopifyAPIReference/*.js`) for the same stale version string, so in-app documentation links stay accurate
- Updated the existing Google Ads API-version changeset to describe all three connector fixes together instead of adding a separate entry

# Changes

- Updated `LinkedInAds/Source.js` and `LinkedInPages/Source.js` to send `LinkedIn-Version: 202607`
- Updated `Shopify/Source.js` to call Admin API version `2026-07`
- Updated 19 files under `Shopify/ShopifyAPIReference/` to point documentation links at `2026-07`
- Rewrote `.changeset/6810-google-ads-api-v25.md` to cover the Google Ads, LinkedIn, and Shopify version fixes as one changeset

### DIFF
--- a/.changeset/6810-google-ads-api-v25.md
+++ b/.changeset/6810-google-ads-api-v25.md
@@ -2,9 +2,16 @@
 'owox': minor
 ---
 
-# Google Ads connector uses a supported API version
+# Connectors use supported API versions
 
 Previously, the Google Ads connector called API version v21, which Google
 sunset on 2026-08-05, causing every import to fail with an
-`UNSUPPORTED_VERSION` error. Now, the connector uses v25, the current stable
-version. This means Google Ads imports run successfully again.
+`UNSUPPORTED_VERSION` error. The LinkedIn Ads and LinkedIn Pages connectors
+were also close to hitting the same wall: they called a version only two
+months newer than one LinkedIn had already sunset. The Shopify connector was
+two releases behind, with its version nearing its own support cutoff.
+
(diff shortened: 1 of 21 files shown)


laskevych (approved):
Verified the bump end to end: LinkedIn `202607` and Shopify `2026-07` are both real, current stable versions per vendor docs, and no stale `202509` / `2025-10` strings remain anywhere on the branch. Proactively moving off sunset-track versions is the right call after the Google Ads incident. LGTM 👍

PR #1513, opened 2026-08-07
feat(api-client): add secure low-level PATCH and DELETE methods

max-voloshyn:
## Scope

Add the documented low-level API-client escape hatch for API-key-compatible endpoints:

- add `patchJson()` and `deleteJson()` beside `getJson()`, `postJson()`, `putJson()`, and `getStream()`;
- preserve source compatibility for existing custom transports while requiring built-in transports to implement the new methods;
- reject unsafe authenticated paths and redirects before acquiring or forwarding credentials;
- carry PATCH and DELETE additively through plugin SDK protocol v1 and the web host bridge without exposing runtime credentials;
- preserve existing protocol-v1 handshake and bodyless POST/PUT compatibility while rejecting unsupported versions;
- document that these caller-typed methods do not count as typed API-client endpoint coverage.

Case: `owox-data-marts/api-client-low-level-json-methods`

Audited base: `0225afe7d6f7412949b8ccacaa6ef06351804db6`

Grouping rationale: this is one cross-cutting transport capability spanning the public API client, plugin SDK, and host trust boundary. It does not claim coverage for any endpoint, so the final endpoint slice size is 0 and no unrelated endpoint work is included.

## Coverage results

| Surface | Before | After | Classification |
| --- | --- | --- | --- |
| OpenAPI endpoint rows | Unchanged | Unchanged | `unchanged` |
| API-client endpoint rows | Unchanged | Unchanged | `unchanged` |
| Low-level transport policy | Implicit | Explicitly excluded from typed coverage | Policy clarification |

Support Matrix statuses, evidence links, dates, counts, and percentages are unchanged. `docs/api/openapi.md` is unchanged.

## Contract checklist

| Category | Source of truth | Independent oracle |
| --- | --- | --- |
| Authentication and visibility | API-key transport and plugin-host runtime-token boundary | zero-token-exchange, zero-token-mint, zero-fetch, hostile-header, redirect, and protected-route tests |
| Request methods and bodies | `OWOXTransport`, `OWOXApiClient`, iframe transport, and protocol-v1 request unions | wire-level PATCH/DELETE assertions; bodyless GET/DELETE, legacy bodyless POST/PUT, and required-body PATCH tests |
| Authenticated path constraints | boundary-local validators with a documented trust-boundary exception | one shared accepted/rejected path conformance oracle consumed by both focused suites; mutation checks catch either boundary drifting from the 2,048-character rule |
| Response behavior | established JSON response parser | JSON DELETE and empty/204 DELETE behavior tests |
| Types and runtime validation | public method generics, legacy-compatible custom transport seam, required built-in transport capability, and protocol-v1 unions | API-client and SDK contract TypeScript configs enforced by normal `typecheck`; unsupported custom-transport calls reject deliberately; runtime docs state caller validation is required |
| Protocol compatibility | additive v1 wire contract in both protocol copies | literal-v1 ready/hello/host-init assertions; bodyless POST/PUT compatibility; PATCH/DELETE parity; unsupported v2 rejection |
| Documentation and coverage | API-client/plugin guides and Support Matrix policy | README parity tests, Markdown lint, and Support Matrix validator |
| Generated OpenAPI | Not applicable; no endpoint contract is changed | `docs/api/openapi.md` and endpoint rows have no diff |

Cross-cutting path admission remains local to each credential-bearing trust boundary because the standalone public client and raw plugin-message host require different adapters and error contracts. A shared conformance oracle prevents their pure admission rule from drifting.

## Verification

- API client focused tests: 3 suites, 58 tests passed.
- Plugin SDK focused tests: 3 files, 30 tests passed.
- Web host bridge: 1 file, 79 tests passed.
- Backend plugin authorization: 4 suites, 27 tests passed.
- Support Matrix validator: 23 tests passed.
- API client, plugin SDK (including protocol contract config), and web no-emit/build-mode type checks passed.
- API client, plugin SDK, and affected web lint passed.
- Affected package/source formatting and documentation Markdown lint passed.
- `changeset status --since=origin/main` passed with the canonical `.changeset/api-surface-maintenance.md`.
- `git diff --check` passed.

## Exemptions and blockers

Exemptions: none.

Blocker register: none.

🤖 Codex (GPT-5.6 Sol High)


.changeset/api-surface-maintenance.md
```diff
@@ -0,0 +1,21 @@
+---
+'@owox/api-client': minor
+'@owox/plugin-sdk': minor
+'@owox/web': minor
+'owox': minor
```
max-voloshyn, line 2: Mention only owox package here

.changeset/api-surface-maintenance.md
```diff
@@ -0,0 +1,21 @@
...
+# Low-level API client transport methods
+
+API-key clients can now use `patchJson()` and `deleteJson()` alongside
+`getJson()`, `postJson()`, `putJson()`, and `getStream()` for API-key-compatible
+endpoints that do not yet have typed resource abstractions. Authenticated
+low-level requests accept only root-relative `/api/...` paths up to 2,048
+characters and refuse unsafe paths and redirects, preventing credentials from
+being sent to an unintended destination.
+
+Plugin SDK protocol v2 adds PATCH and DELETE through `ctx.owox`. The web host
```
max-voloshyn: Don't introduce new protocol version, evolve existing, it seems not breaking

.changeset/api-surface-maintenance.md
```diff
@@ -0,0 +1,21 @@
...
+'@owox/api-client': minor
+'@owox/plugin-sdk': minor
+'@owox/web': minor
+'owox': minor
+---
+
+# Low-level API client transport methods
+
+API-key clients can now use `patchJson()` and `deleteJson()` alongside
+`getJson()`, `postJson()`, `putJson()`, and `getStream()` for API-key-compatible
```
max-voloshyn: 1. Should these methods be namespaces as `owox.raw.*`?
2. omit "Json" suffix?
3. getStream -> stream ?

.changeset/api-surface-maintenance.md
max-voloshyn: Updated the changeset frontmatter to list only `owox`. `npx changeset status --since=origin/main` confirms that the configured fixed group still propagates the release to the affected packages.

🤖 Codex (GPT-5.6 Sol High)

.changeset/api-surface-maintenance.md
max-voloshyn: Kept the wire protocol at v1 and evolved it additively. Legacy bodyless POST/PUT remain accepted, PATCH requires a JSON body, DELETE remains bodyless, and unsupported v2 ready/hello/host-init messages are rejected. The SDK tests now pin literal v1 independently of the implementation constant. Fixed in `3dc201014`.

🤖 Codex (GPT-5.6 Sol High)

.changeset/api-surface-maintenance.md
max-voloshyn: Keeping the method-per-shape API follows the confirmed case direction and preserves the established public surface: `getJson()`, `postJson()`, `putJson()`, and `getStream()` already exist, so `patchJson()` and `deleteJson()` extend the same naming scheme. Introducing `owox.raw.*` or renaming the existing methods would broaden this transport-focused case and risk breaking current consumers. The docs instead make the low-level/unchecked tradeoff explicit and exclude these methods from typed endpoint coverage.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/plugins/runtime/protocol.ts
```diff
@@ -6,14 +6,12 @@
...
- * Change one side and the version must move with it. The handshake guards check the
- * version on every message, so a mismatch fails the handshake instead of being read
- * against the wrong rules.
+ * satisfy anyway. Additive request methods may evolve v1 when the host preserves every
+ * existing wire shape; incompatible wire changes require a new negotiated version.
  */
 
-export const PLUGIN_PROTOCOL_VERSION = 1;
+export const PLUGIN_PROTOCOL_VERSION = 1 as const;
+export type PluginProtocolVersion = typeof PLUGIN_PROTOCOL_VERSION;
```
max-voloshyn: leave only nessesary changes in this file

packages/plugin-sdk/README.md
```diff
@@ -34,6 +34,19 @@ against an SDK the deployment cannot speak fails to start rather than misbehavin
...
 which attaches the token — so requests act with **the authority of the member who
 installed your plugin**, and never more. Do not assume you are trusted beyond that.
+Protected routes still apply their server-side authorization and reject calls the member
+may not make.
+
+The existing protocol adds `ctx.owox.patchJson()` and `ctx.owox.deleteJson()` as an
+additive capability for the same low-level API escape hatch available to API-key clients.
+Existing plugins remain wire-compatible.
+
+Use low-level methods only for API-key-compatible endpoints without a typed resource:
```
max-voloshyn: > API-key-compatible endpoints

Don't hardcode "API-key" here, plugin-sdk relies on api-client interface but don't use API keys

packages/plugin-sdk/README.md
```diff
@@ -34,6 +34,19 @@ against an SDK the deployment cannot speak fails to start rather than misbehavin
 Your plugin never holds a credential. `ctx.owox` calls are brokered by the host page,
 which attaches the token — so requests act with **the authority of the member who
 installed your plugin**, and never more. Do not assume you are trusted beyond that.
+Protected routes still apply their server-side authorization and reject calls the member
+may not make.
+
+The existing protocol adds `ctx.owox.patchJson()` and `ctx.owox.deleteJson()` as an
```
max-voloshyn: `The existing protocol adds`

README != changelog

packages/plugin-sdk/src/protocol.ts
```diff
@@ -70,13 +70,42 @@ export type PluginRequest =
   | {
       id: string;
       kind: 'api';
-      method: 'GET' | 'POST' | 'PUT';
+      method: 'POST' | 'PUT';
       path: string;
       query?: PluginQuery;
+      /** Optional for compatibility with plugins built against the original v1 shape. */
```
max-voloshyn: version menions seems redundant

packages/api-client/src/transport.ts
```diff
@@ -6,24 +6,27 @@
...
  * own has nothing to do.
  */
+import { OWOXConfigError, createNetworkError } from './errors.js';
+
 export type OWOXTransport = {
   getJson<T>(path: string, query?: Record<string, string>): Promise<T>;
   postJson<T>(path: string, jsonBody: unknown, accept?: string): Promise<T>;
   putJson<T>(path: string, jsonBody: unknown): Promise<T>;
+  patchJson<T>(path: string, jsonBody: unknown): Promise<T>;
+  deleteJson<T = void>(path: string): Promise<T>;
```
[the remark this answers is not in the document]
max-voloshyn: Made `patchJson` and `deleteJson` optional on the public `OWOXTransport`, so existing custom transports remain source-compatible. `OWOXApiClient` now rejects unsupported new calls with `OWOXConfigError`, while a stronger internal capability type still requires both methods on the API-key client and iframe transport. The legacy shape is enforced by the normal API-client typecheck. Fixed in `4d1035bfd`.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/plugins/runtime/protocol.ts
max-voloshyn: Reduced this file's diff to only the request-shape changes required for PATCH/DELETE and legacy bodyless POST/PUT compatibility. Removed the negotiated-version alias/plumbing, host-only request alias, and explanatory churn; the existing v1 ready/hello guards remain unchanged. Fixed in `355168130`.

🤖 Codex (GPT-5.6 Sol High)

packages/plugin-sdk/README.md
max-voloshyn: Removed the API-key eligibility wording from the plugin SDK README and authoring guide. They now describe these as host-brokered low-level methods on `ctx.owox`, without implying that plugin runtime authentication uses API keys. Fixed in `eb1176ce0`.

🤖 Codex (GPT-5.6 Sol High)

packages/plugin-sdk/README.md
max-voloshyn: Replaced the changelog-style sentence with direct reference documentation: `patchJson()` and `deleteJson()` are available for low-level requests brokered by the host. Fixed in `eb1176ce0`.

🤖 Codex (GPT-5.6 Sol High)

packages/plugin-sdk/src/protocol.ts
max-voloshyn: Removed the redundant v1 compatibility comment from both protocol copies while keeping their request shapes aligned. Fixed in `eb1176ce0`.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/client.test.ts
```diff
@@ -1,10 +1,16 @@
 import { jest } from '@jest/globals';
 
 import { OWOXApiClient, OWOXApiError, OWOXAuthError, OWOXConfigError } from './index.js';
+import { requestApi, resolveAuthenticatedApiUrl } from './transport.js';
+import {
+  acceptedAuthenticatedApiPaths,
+  rejectedAuthenticatedApiPaths,
+} from '../../../apps/web/src/test/authenticated-api-path-contract.js';
```
romandubovyi: ⚪ **Non-blocking:** This is the only `packages/* → apps/*` import in the repo, and it inverts the dependency direction: the published `@owox/api-client` test suite can no longer run without `apps/web` on disk. The package is otherwise self-contained (`files: ["README.md", "dist"]`, its own `tsconfig.json` with `rootDir: "./src"`, its own Jest with `roots: ['<rootDir>/src']`), and the imported file sits outside that `rootDir`. It is also unchecked by either typecheck path — `tsconfig.json` excludes `**/*.test.ts` and `tsconfig.contracts.json` includes only `transport-seam.test.ts` — so the shared oracle is enforced by ts-jest alone. `apps/web/package.json` does not depend on `@owox/api-client`, so the mirror-image fix is not available either. This is the same coupling the host protocol header at `apps/web/src/features/plugins/runtime/protocol.ts:1-12` deliberately refuses. Either keep a local copy of the table on each side (consistent with the existing host-keeps-its-own-copy convention), or move it into a private internal package both workspaces may legitimately depend on.

🤖 Generated with Grok CLI (Grok 4.5) + Claude Code (Opus 5)

apps/web/src/features/plugins/runtime/pluginHostBridge.ts
```diff
@@ -222,10 +231,28 @@ export function createPluginHostBridge(options: PluginHostBridgeOptions): Plugin
 
     inFlight += 1;
     try {
-      const response = await forward(request);
+      // Admission covers validation as well as I/O. In particular, JSON serialization
+      // can be attacker-controlled work and must not remain unbounded at capacity.
+      const request = validateRequest(candidate);
+
+      if (request.kind === 'openExternal') {
```
romandubovyi, line 257: ⚪ **Non-blocking:** `openExternal` and `navigate` used to be dispatched before the `MAX_IN_FLIGHT` check; they now sit after it (line 223). Neither performs I/O, and both are fire-and-forget on the SDK side — `index.ts:165-171` posts them straight to the port with their own id, bypassing `send()`, so no `pending` entry exists and `iframe-transport.ts:44-49` drops the `PROTOCOL_ERROR` reply silently. With 32 API requests in flight, a plugin's navigation simply does nothing, with no error surfaced anywhere. The stated reason for moving validation inside admission is bounding attacker-controlled JSON serialization, but that work happens in `serializeJsonBody()`, after these branches — so validating and dispatching `openExternal`/`navigate` above the in-flight check keeps the bound and restores the previous behavior.

🤖 Generated with Grok CLI (Grok 4.5) + Claude Code (Opus 5)

apps/web/src/features/plugins/runtime/pluginHostBridge.ts
```diff
@@ -87,20 +97,32 @@ export function createPluginHostBridge(options: PluginHostBridgeOptions): Plugin
...
 
-    if (!ALLOWED_METHODS.has(request.method)) {
-      throw forbidden(`${request.method} is not allowed from a plugin`);
+    const url = new URL(request.path, apiOrigin);
+
+    if (url.origin !== apiOrigin || !url.pathname.startsWith(API_PATH_PREFIX)) {
+      throw forbidden('Requests must target a root-relative path under /api/');
     }
 
+    decodePathForValidation(url.pathname);
```
romandubovyi: ⚪ **Non-blocking:** After constructing the URL, the API client re-decodes `url.pathname` and re-applies the backslash and `.`/`..` segment checks (`assertAuthenticatedApiUrl` in `transport.ts`). The host only calls `decodePathForValidation(url.pathname)` and discards the result, so the post-URL segment checks are missing here. I could not find an input that reaches this line with an unsafe pathname — the pre-URL check catches the encoded and literal cases — so this is parity, not an exploitable gap. But the shared oracle only pins accept/reject decisions, so it would not catch the two boundaries drifting on the internal steps either. Re-running the same post-URL checks the client uses would keep the documented "same admission rule" claim literally true.

packages/api-client/src/transport.ts
```diff
@@ -50,6 +162,13 @@ function buildApiUrl(apiOrigin: string, path: string, query: QueryParams | undef
 }
 
 export async function requestApi(options: ApiRequestOptions): Promise<Response> {
+  const resolvedUrl = resolveAuthenticatedApiUrl(options.apiOrigin, options.path, options.query);
+  if (options.url && options.url.href !== resolvedUrl.href) {
```
romandubovyi: ⚪ **Non-blocking:** `requestApi` already recomputes `resolvedUrl` from the same `apiOrigin`/`path`/`query`, and there are only two callers: `auth.ts:79` passes no `url` at all, and `api-key-transport.ts:113` passes exactly `resolveAuthenticatedApiUrl(this.apiOrigin, path, options.query)` — the identical value. The href comparison therefore cannot fail for any input, and the `assertAuthenticatedApiUrl` call right below it is a repeat pass, since `resolveAuthenticatedApiUrl` already asserts before returning. Dropping the `url` option would remove the redundant passes, one field from `ApiRequestOptions`, and the `client.test.ts` cases that exist only to exercise this seam. The property you want — resolve before the token exchange — is already carried by `resolveAuthenticatedApiUrl` being the first statement in `requestApi`.

packages/plugin-sdk/src/iframe-transport.ts
```diff
@@ -6,6 +6,8 @@ import type {
   PluginResponse,
 } from './protocol.js';
 
+type IframeTransport = OWOXTransport & Required<Pick<OWOXTransport, 'patchJson' | 'deleteJson'>>;
```
romandubovyi: ⚪ **Non-blocking:** This is structurally a verbatim re-declaration of `OWOXTransportWithLowLevelWrites` from `packages/api-client/src/transport.ts`. The SDK cannot reuse it only because `packages/api-client/src/index.ts:34` exports just `OWOXTransport` from that module — the package is already a dependency here and the line above already imports `OWOXTransport` from it. Adding the new type to the api-client barrel and importing it keeps one definition of the contract instead of two that can drift.

romandubovyi (approved):
Approving — no blocking findings. I checked the parts that carry the security weight of this PR rather than only the diff shape: `API_KEY_EXCHANGE_PATH = '/api/auth/api-keys/exchange'` passes the new validator so the token exchange still works, `redirect: 'error'` cannot be overridden through `fetchInit` because the spread precedes it on both boundaries, and encoded traversal (`%2e%2e`, `%2f`, `%5c`) is caught by the pre-URL check on both the client and the host. I also checked whether the new `usableRequestId` gate could drop `openExternal`/`navigate` for lack of an id — it cannot, the SDK stamps one.

Five non-blocking notes below. The one worth reading first is the `MAX_IN_FLIGHT` ordering change: it is the only finding here with a user-visible behavior regression, and it fails silently in both directions.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Grok CLI (Grok 4.5) + Claude Code (Opus 5)

packages/api-client/src/client.test.ts
max-voloshyn: Moved the shared path cases to the package-neutral `test/contracts/authenticated-api-path-contract.mjs` and pointed both focused suites at it, removing the `packages/* → apps/*` dependency while preserving one cross-boundary drift detector. Current `main` also typechecks all API-client tests through `tsconfig.test.json`, and the shared table now independently covers same-origin absolute URL rejection. Fixed in `ad3aef737`.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/plugins/runtime/pluginHostBridge.ts
max-voloshyn: Restored fire-and-forget behavior by validating and dispatching `openExternal` and `navigate` before the API in-flight admission check. API validation and JSON serialization remain bounded by the 32-request limit. A regression test fills all API slots and verifies navigation still reaches the host. Fixed in `9f7478ec7`.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/plugins/runtime/pluginHostBridge.ts
max-voloshyn: Re-applied the decoded backslash and traversal-segment checks to `url.pathname` after URL construction, using the same local predicate as the pre-construction pass. The package-neutral path contract remains shared by the host and API-client suites. Fixed in `9f7478ec7` and `ad3aef737`.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/transport.ts
max-voloshyn: Removed the optional precomputed `url` field and redundant href/assertion passes from `requestApi`. `ApiKeyTransport` still resolves the caller path before acquiring a token, and `requestApi` performs the final resolution used by fetch. The focused credential non-leakage tests remain green. Fixed in `9f7478ec7`.

🤖 Codex (GPT-5.6 Sol High)

packages/plugin-sdk/src/iframe-transport.ts
max-voloshyn: Exported `OWOXTransportWithLowLevelWrites` from the API-client barrel and changed the iframe transport to use that shared capability type. The duplicate SDK declaration is gone, and both package typechecks pass. Fixed in `9f7478ec7`.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/transports/api-key-transport.ts
```diff
@@ -59,6 +63,14 @@ export class ApiKeyTransport implements OWOXTransport {
     return this.requestJsonWithAuth<T>(path, { method: 'PUT', jsonBody });
   }
 
+  async patchJson<T>(path: string, jsonBody: unknown): Promise<T> {
+    return this.requestJsonWithAuth<T>(path, { method: 'PATCH', jsonBody });
```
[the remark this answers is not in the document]
max-voloshyn: Added an `OWOXApiClient.patchJson()` guard that rejects `undefined` after the legacy transport capability check but before delegation, authentication, fetch, or MessagePort dispatch. The regression test uses a valid fresh API key and asserts `OWOXConfigError` with zero fetch calls; bodyless POST/PUT compatibility is unchanged. Fixed in `7a8f1d203`.

🤖 Codex (GPT-5.6 Sol High)

.changeset/api-surface-maintenance.md
```diff
@@ -0,0 +1,19 @@
+---
+'owox': minor
+---
+
+# Low-level API client transport methods
+
+API-key clients can now use `patchJson()` and `deleteJson()` alongside
+`getJson()`, `postJson()`, `putJson()`, and `getStream()` for API-key-compatible
```
max-voloshyn: Use transport-neutral wording here because the api-client is also used by plugin-sdk, whose requests are brokered by the host rather than authenticated with API keys.

🤖 Codex (GPT-5.6 Sol High)

max-voloshyn (commented):
Release-note wording review.

🤖 Codex (GPT-5.6 Sol High)

.changeset/api-surface-maintenance.md
max-voloshyn: Fixed in 7a3679bbf. The changeset now describes OWOX API clients and `/api/...` endpoints without tying the escape hatch to API-key authentication; the package READMEs retain the transport-specific authentication details.

Verification: Prettier passed, markdownlint reported 0 errors, changeset status succeeded, and `git diff --check` passed.

🤖 Codex (GPT-5.6 Sol High)

packages/plugin-sdk/README.md
```diff
@@ -34,6 +34,18 @@ against an SDK the deployment cannot speak fails to start rather than misbehavin
 Your plugin never holds a credential. `ctx.owox` calls are brokered by the host page,
 which attaches the token — so requests act with **the authority of the member who
 installed your plugin**, and never more. Do not assume you are trusted beyond that.
+Protected routes still apply their server-side authorization and reject calls the member
+may not make.
+
+`ctx.owox.patchJson()` and `ctx.owox.deleteJson()` are available for low-level PATCH and
+DELETE requests brokered by the host.
```
max-voloshyn: I don't understand why mention patchJson and deleteJson specifically along with next paragraph

packages/api-client/src/client.ts
```diff
@@ -77,6 +78,23 @@ export class OWOXApiClient implements OWOXTransport {
     return this.transport.putJson<T>(path, jsonBody);
   }
 
+  async patchJson<T>(path: string, jsonBody: unknown): Promise<T> {
+    if (!this.transport.patchJson) {
+      throw new OWOXConfigError('Injected OWOX transport does not support patchJson()');
+    }
+    if (jsonBody === undefined) {
+      throw new OWOXConfigError('patchJson() requires a defined JSON body');
```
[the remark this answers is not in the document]
max-voloshyn: Fixed in 25385111d. `patchJson()` now preflights `JSON.stringify()` before delegating to any transport, rejecting values whose result is `undefined` and serialization errors as `OWOXConfigError` before authentication. The regression covers top-level symbols and functions and verifies no fetch occurs.

Verification: focused client test 46/46 passed; API-client typecheck and lint passed; affected-file formatting, markdownlint, `git diff --check`, and changeset status passed.

🤖 Codex (GPT-5.6 Sol High)

packages/plugin-sdk/README.md
max-voloshyn: Fixed in 25385111d. Removed the redundant PATCH/DELETE sentence; the following low-level-method list already names both methods, while the preceding paragraph explains that all `ctx.owox` requests are brokered by the host.

Verification: plugin README formatting and markdownlint passed.

🤖 Codex (GPT-5.6 Sol High)

packages/api-client/src/client.ts
```diff
@@ -77,6 +78,29 @@ export class OWOXApiClient implements OWOXTransport {
...
+    let serializedBody: string | undefined;
+    try {
+      serializedBody = JSON.stringify(jsonBody);
+    } catch (cause) {
+      throw new OWOXConfigError('patchJson() requires a JSON-serializable body', { cause });
+    }
+    if (serializedBody === undefined) {
+      throw new OWOXConfigError('patchJson() requires a JSON-serializable body');
+    }
+    return this.transport.patchJson<T>(path, jsonBody);
```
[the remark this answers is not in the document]
max-voloshyn: Fixed in cce16a5e2. After validating serialization, `patchJson()` now parses that serialized representation and forwards the resulting JSON-compatible snapshot. Nested functions and `undefined` values therefore follow JSON semantics before either API-key or iframe transport delegation, avoiding `DataCloneError` divergence.

Verification: the new transport-boundary regression failed against the original delegation and now passes; focused client suites passed 61/61, with API-client typecheck and lint, affected-file formatting, markdownlint, `git diff --check`, and changeset status also passing.

🤖 Codex (GPT-5.6 Sol High)

PR #1514, opened 2026-08-07
fix(google-ads): remap fields renamed in API v25

prokopetsv:
# Problem

Google Ads renamed four fields between API v21 and v25: `metrics.video_views`, `metrics.video_view_rate`, `campaign.start_date`, and `campaign.end_date`. The connector's earlier v21→v25 migration updated the endpoint URL but missed these four field mappings, which still requested the old, now-unrecognized names.

Google Ads rejects a GAQL query entirely when it contains even one unrecognized field, so any `campaigns` or `campaigns_stats` import selecting one of these fields failed outright with `UNRECOGNIZED_FIELD` / `INVALID_ARGUMENT`. This affected new data marts too, since `campaign_start_date` and `campaign_end_date` are selected by default on the `campaigns` node.

# Solution

Remapped the four affected fields to their current v25 API names, keeping the same field names in the schema and UI so existing data mart configurations and destination tables are unaffected:

- `metrics.video_views` → `metrics.video_trueview_views`
- `metrics.video_view_rate` → `metrics.video_trueview_view_rate`
- `campaign.start_date` → `campaign.start_date_time`
- `campaign.end_date` → `campaign.end_date_time`

Added a regression test that builds the real field schema from the GAS-style field-definition files and asserts, through `GoogleAdsSource._getAPIFields()`, that the mapped API names match v25 and that none of the four dead v21 names remain anywhere in the schema — so a future version bump that reintroduces a stale field name fails the test suite instead of failing silently in production.

# Changes

- Updated `apiName` for `video_views` and `video_view_rate` in `campaignsStatsFields.js`
- Updated `apiName` for `campaign_start_date` and `campaign_end_date` in `campaignsFields.js`
- Added a test suite in `Source.test.js` covering the v25 field mappings and asserting the old field names are gone from the schema
- Added changeset `6812-google-ads-v25-field-renames.md`

### DIFF
--- a/packages/connectors/src/Sources/GoogleAds/GoogleAdsAPIReference/campaignsFields.js
+++ b/packages/connectors/src/Sources/GoogleAds/GoogleAdsAPIReference/campaignsFields.js
@@ -28,12 +28,12 @@ var campaignFields = {
   },
   'campaign_start_date': {
     'description': 'Campaign Start Date',
-    'apiName': 'campaign.start_date',
+    'apiName': 'campaign.start_date_time',
     'type': DATA_TYPES.STRING
   },
   'campaign_end_date': {
     'description': 'Campaign End Date',
-    'apiName': 'campaign.end_date',
+    'apiName': 'campaign.end_date_time',
     'type': DATA_TYPES.STRING
   },
   'campaign_bidding_strategy_type': {
--- a/packages/connectors/src/Sources/GoogleAds/GoogleAdsAPIReference/campaignsFields.js
+++ b/packages/connectors/src/Sources/GoogleAds/GoogleAdsAPIReference/campaignsFields.js
@@ -29,12 +29,18 @@ var campaignFields = {
   'campaign_start_date': {
     'description': 'Campaign Start Date',
     'apiName': 'campaign.start_date_time',
-    'type': DATA_TYPES.STRING
+    'type': DATA_TYPES.STRING,
+    // v25 returns "yyyy-MM-dd HH:mm:ss" (zeroed to 00:00:00 for daily-granularity
+    // campaigns); v21's campaign.start_date returned a bare "yyyy-MM-dd".
+    'dateOnly': true
   },
   'campaign_end_date': {
     'description': 'Campaign End Date',
     'apiName': 'campaign.end_date_time',
-    'type': DATA_TYPES.STRING
+    'type': DATA_TYPES.STRING,
+    // Same as campaign_start_date: v25 returns "yyyy-MM-dd HH:mm:ss" (23:59:59 for
+    // daily-granularity campaigns) where v21's campaign.end_date was a bare date.
+    'dateOnly': true
   },
   'campaign_bidding_strategy_type': {
     'description': 'Bidding Strategy Type (MANUAL_CPC, TARGET_CPA, TARGET_ROAS, etc.)',
--- a/packages/connectors/test/Sources/GoogleAds/Source.test.js
+++ b/packages/connectors/test/Sources/GoogleAds/Source.test.js
@@ -4,11 +4,48 @@ import { describe, expect, it } from 'vitest';
 import { loadGasClass } from '../../support/loadGasClass.js';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const apiReferenceDir = path.join(
+  __dirname,
+  '../../../src/Sources/GoogleAds/GoogleAdsAPIReference'
+);
 
 loadGasClass(path.join(__dirname, '../../../src/Constants/HttpConstants.js'));
 loadGasClass(path.join(__dirname, '../../../src/Core/AbstractSource.js'));
 loadGasClass(path.join(__dirname, '../../../src/Sources/GoogleAds/Source.js'));
 const proto = globalThis.GoogleAdsSource.prototype;
+
+// Each field file declares its object with `var`, so it attaches to globalThis once
+// loaded (GAS-style scripts, no imports/exports). GoogleAdsFieldsSchema.js itself uses
+// `const`, which vm.runInThisContext does NOT attach to globalThis, so it's rebuilt here
+// from the field globals instead of being loaded directly — this mirrors the node -> fields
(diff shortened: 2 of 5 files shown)


packages/connectors/src/Sources/GoogleAds/GoogleAdsAPIReference/campaignsFields.js
```diff
@@ -28,12 +28,12 @@ var campaignFields = {
   },
   'campaign_start_date': {
     'description': 'Campaign Start Date',
-    'apiName': 'campaign.start_date',
+    'apiName': 'campaign.start_date_time',
```
laskevych, line 31: 🟡 **Important (non-blocking, please verify):** `campaign.start_date_time` / `campaign.end_date_time` almost certainly return a **datetime** string, not the `yyyy-MM-dd` value the old `start_date`/`end_date` fields returned — Google renamed these fields precisely because campaigns now support a time-of-day component for start/end ([About campaign start and end times](https://support.google.com/google-ads/answer/16072972)).

If so, the values stored in the `campaign_start_date` / `campaign_end_date` columns silently change format (e.g. `2026-01-15` → `2026-01-15 00:00:00`): existing destination tables end up with mixed formats, and any downstream SQL that parses or compares the column as a date may break — which would contradict the "destination tables are unaffected" note in the description.

Could you check the actual response format from your real run (the one referenced in the test comment)? If it does include a time component, truncating to the date part during mapping would preserve backward-compatible values — the field is still named and described as a *date*. Fine as a fast-follow if you'd rather keep this PR minimal.

packages/connectors/test/Sources/GoogleAds/Source.test.js
```diff
@@ -4,12 +4,49 @@ import { describe, expect, it } from 'vitest';
...
 loadGasClass(path.join(__dirname, '../../../src/Sources/GoogleAds/Source.js'));
 const proto = globalThis.GoogleAdsSource.prototype;
 
+// Each field file declares its object with `var`, so it attaches to globalThis once
+// loaded (GAS-style scripts, no imports/exports). GoogleAdsFieldsSchema.js itself uses
+// `const`, which vm.runInThisContext does NOT attach to globalThis, so it's rebuilt here
+// from the field globals instead of being loaded directly — this mirrors the node -> fields
+// mapping in src/Sources/GoogleAds/GoogleAdsAPIReference/GoogleAdsFieldsSchema.js.
+loadGasClass(path.join(__dirname, '../../../src/Constants/DataTypes.js'));
+for (const file of [
```
laskevych, line 24: 💡 **Suggestion (non-blocking):** the "no dead v21 names" sweep iterates a hardcoded list of nine field files that mirrors `GoogleAdsFieldsSchema.js`. If a new field file/node is added later with a stale pre-v25 `apiName`, the sweep never loads it and the suite stays green — the exact failure mode this test exists to catch. For the sweep test specifically, deriving the list dynamically (`fs.readdirSync(apiReferenceDir)`, excluding `GoogleAdsFieldsSchema.js`) would keep future files covered automatically. The two explicit mapping tests are fine as they are.

laskevych (approved):
✅ **Approved.** The rename is consistent end-to-end — both the GAQL `SELECT` and the response mapping derive from the same `apiName`, so no second location needed updating — and the regression test that rebuilds the schema from the real field files and asserts the dead v21 names are gone from every node is exactly the right guard for future version bumps. Ran the test suite locally: 5/5 passing.

The video-metrics rename is also corroborated externally ([Google renamed Views → TrueView Views](https://ppcnewsfeed.com/ppc-news/2025-10/google-renames-views-metric-to-trueview-views/) in late 2025), with the underlying values unchanged, so historical re-imports of `campaigns_stats` should stay consistent.

One important question inline about the value format of `campaign.start_date_time` — worth a quick check against your real run, but it doesn't block merging a fix for a fully broken import path.

packages/connectors/src/Sources/GoogleAds/GoogleAdsAPIReference/campaignsFields.js
prokopetsv: Confirmed — checked Google's v25 field reference docs directly. `campaign.start_date_time` / `end_date_time` return `"yyyy-MM-dd HH:mm:ss"`, with the time zeroed to `00:00:00` (start) / `23:59:59` (end) for daily-granularity campaigns — so the plain `yyyy-MM-dd` format would indeed have broken.

`GoogleAdsSource` had no existing value-casting step, so that datetime string would've flowed straight to storage as-is.

I also checked whether retyping these fields to `DATA_TYPES.DATE` would fix it — it doesn't, and would've made things worse: `GoogleBigQueryStorage` maps `DATA_TYPES.DATE` to a BigQuery `DATE` column vs `STRING` for `DATA_TYPES.STRING`, so that would attempt to change an existing column's SQL type, a bigger break than the one being fixed.

Kept `type: STRING` and added a `dateOnly: true` flag on these two field configs. `_mapResultToColumns` now drops the time component when that flag is set, before the value reaches storage — so `campaign_start_date`/`campaign_end_date` keep the exact `yyyy-MM-dd` format existing destination tables already have. Added a regression test with a synthetic datetime API response asserting the truncation, and confirming a sibling non-`dateOnly` field is untouched.

Pushed in `292728ca2`.

packages/connectors/test/Sources/GoogleAds/Source.test.js
prokopetsv: Good catch — agreed, that's exactly the blind spot this test exists to close. Changed the "no dead v21 names" sweep to read `apiName` values straight from every `.js` file in `GoogleAdsAPIReference/` via `fs.readdirSync` (excluding `GoogleAdsFieldsSchema.js`), instead of going through the hardcoded load-loop list / `fieldsSchemaSelf`. Any new field file added later is swept automatically with no test-file change needed.

Left the two explicit mapping tests (`_getAPIFields` for video metrics and campaign dates) on the hardcoded list as suggested, since those need the real node-name structure to call `_getAPIFields` correctly.

Pushed in `292728ca2`.

PR #1515, opened 2026-08-07
fix(data-marts): reliable bulk publishing of Data Mart drafts

prokopetsv:
# Problem

Publishing Data Mart drafts in bulk from a Storage's "Publish drafts" menu action failed for every draft, showing a generic "Failed to publish N data mart drafts. Please check them independently." message with no way to diagnose the cause.

The root cause was in `PublishDataStorageDraftsService`: the per-draft publish call was built with a hardcoded empty `roles: []` array instead of the publisher's actual project roles. `AccessDecisionService.resolveRole([])` resolves an empty roles array to `VIEWER`, so every draft failed its EDIT permission check regardless of the user's real role — indistinguishable from the user genuinely lacking access. The bulk loop also swallowed every error with a bare `catch { ++failedCount }`, discarding the real reason before it ever reached the response.

Two more defects surfaced while fixing this: the response leaked internal error text (including live warehouse validation output, which can contain SQL fragments and credential hints) to any project viewer regardless of their role, and a failure in a post-publish side effect (schema-actualization scheduling, search reindex, or event dispatch) was reported as a failed *publish* — even though the Data Mart had already been saved as `PUBLISHED` — sending the user to a DRAFT-filtered list where it no longer appeared.

# Solution

- Resolve the publisher's real roles from the identity service (`IdpProjectionsFacade.getProjectForUser`) instead of hardcoding `[]`, and treat an empty or unresolvable role list as a hard error rather than silently degrading to viewer-level access — the same failure mode this PR fixes, just reintroduced through a different path.
- Skip the roles lookup entirely when there are no drafts to publish, so a transient identity-service issue can't turn "nothing to publish" into an error.
- Separate a draft's publish outcome from its post-publish side effects (schema-actualization scheduling, search reindex, event dispatch). None of these can fail the reported result anymore — they're logged and skipped, since by the time they run the Data Mart is already committed as published.
- Replace the swallow-and-count catch with a small allowlist: only error messages this codebase authors itself are returned to the client (`PUBLISH_DATA_MART_ERRORS`); anything else — including warehouse driver output — is replaced with a generic message, with the full detail preserved in the server log.
- Return deduplicated failure *reasons* only, with no Data Mart IDs or titles: creating the trigger only requires EDIT access to the storage, but the batch covers every draft inside it, including ones the publisher may not have visibility into. Only the reasons travel back to the client.
- Frontend: the failure toast now states the actual reason instead of a generic "check them independently" message.

# Changes

- `publish-data-mart.service.ts`: extracted `PUBLISH_DATA_MART_ERRORS` allowlist constants; wrapped the post-`save()` side effects (search reindex, `DataMartPublishedEvent` dispatch) so their failure can't be reported as a failed publish
- `publish-data-storage-drafts.service.ts`: resolve real publisher roles per batch instead of hardcoding `[]`; fail loudly (not silently) when roles can't be determined; skip the roles lookup when there are no drafts; separated per-draft publish outcome from schema-actualization scheduling; replaced the bare `catch { ++failedCount }` with logged, sanitized failure reasons
- `publish-data-storage-drafts-result.dto.ts` / `publish-data-storage-drafts-response-api.dto.ts`: added `failureReasons: string[]` (deduplicated, no per-draft identifiers)
- `data-storage.mapper.ts` / `publish-drafts-trigger-handler.service.ts`: wired the response through the existing (previously unused) `DataStorageMapper.toPublishDraftsResponse` mapper
- `usePublishDraftsTrigger.ts` / `buildPublishFailureMessage.ts`: the failure toast now renders the real reason(s) instead of a generic message
- Added unit test coverage for: role resolution and its failure modes, the schema-actualization/publish-outcome split, the error allowlist (including a case asserting no internal identifiers leak into the response), and the toast message builder


apps/web/src/features/data-storage/shared/utils/buildPublishFailureMessage.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Builds the toast text for drafts that could not be published.
+ *
+ * `failedCount` comes from the server rather than `reasons.length`: the API
+ * returns deduplicated reasons and no per-draft identifiers, because editing a
+ * storage does not imply visibility of every Data Mart inside it.
+ */
+export function buildPublishFailureMessage(failedCount: number, reasons: string[]): string {
+  const plural = failedCount !== 1 ? 's' : '';
+  const reason = reasons.length === 1 ? `: ${reasons[0]}` : ' due to different errors';
```
laskevych: 🟡 **Important (non-blocking):** An empty `reasons` array falls into the second branch and renders “due to different errors” when zero reasons are known. This will actually happen during a rolling deploy: a trigger completed by the pre-deploy backend persists `uiResponse` without `failureReasons` (triggers live up to the 1-hour TTL), the new frontend polls it, the hook passes `response.failureReasons ?? []`, and with `failedCount: 3` the toast asserts “due to different errors” with no reasons at all. Worth special-casing `reasons.length === 0` with a neutral fallback (e.g. the old generic wording).

apps/web/src/features/data-storage/shared/utils/buildPublishFailureMessage.ts
```diff
@@ -0,0 +1,13 @@
...
+
+  return `Failed to publish ${String(failedCount)} Data Mart draft${plural}${reason}. Review them in the Data Marts list and try again.`;
```
laskevych: 🟢 **Nitpick (non-blocking):** Two copy details: (1) “Review **them**” is hardcoded, so a single failed draft reads “Failed to publish 1 Data Mart draft: … Review them…” — the message this replaces switched `it`/`them` on `failedCount`. (2) The unconditional “and try again” suffix contradicts some reasons: if a draft was published concurrently mid-batch (single-publish endpoint or a second trigger), its reason is “Data Mart is already published” — retrying can never succeed and the mart is no longer in the DRAFT-filtered list this points to.

apps/backend/src/data-marts/services/publish-drafts-trigger-handler.service.ts
```diff
@@ -37,20 +42,22 @@ export class PublishDraftsTriggerHandlerService
...
       );
       trigger.uiResponse = {
         successCount: 0,
         failedCount: 0,
-        error: e instanceof Error ? e.message : 'Unknown error',
+        // PublishDataStorageDraftsService only raises BusinessViolationException
+        // with text it authored itself. Anything else reaching here is
+        // infrastructure (TypeORM, NotFound, a bug) whose message can carry
+        // internal detail, and this response is readable by any project viewer.
+        error: e instanceof BusinessViolationException ? e.message : GENERIC_TRIGGER_ERROR,
```
laskevych: 💡 **Suggestion (non-blocking):** The genericization is a clear improvement (the old code leaked storage/project IDs into the toast), but it labels some *permanent* conditions as retryable. Concretely: `DataStorageService.getByProjectIdAndId` throws `NotFoundException` when the storage was deleted while the trigger was pending (`DeleteDataStorageService` has no pending-trigger guard, and a stuck trigger can sit up to the 15-min timeout / 1-h TTL), and the user then gets “Please try again.” forever. Consider mapping `NotFoundException` to its own message (“This Storage no longer exists”) while keeping the generic fallback for everything else.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -37,32 +69,110 @@ export class PublishDataStorageDraftsService {
...
+   * trigger is processed asynchronously with no live request, so roles cannot
+   * be read from a JWT and are fetched fresh here.
+   *
+   * An unresolved role list must never fall back to `[]`: AccessDecisionService
+   * resolves `[]` to VIEWER, which fails every draft with a permission error
+   * indistinguishable from the user genuinely lacking access — the same silent
+   * all-fail this service was fixed to eliminate. Fail loudly instead.
+   */
+  private async resolvePublisherRoles(command: PublishDataStorageDraftsCommand): Promise<string[]> {
+    const project = await this.idpProjectionsFacade
```
laskevych, line 155: 💡 **Suggestion (non-blocking):** The blanket `.catch` maps *every* rejection to the “Please try again.” message, but on `idp-owox-better-auth` the client throws typed, definitive errors: `handleAxiosError` produces `ForbiddenException` on 403 and `IdpNotFoundException` on 404 (e.g. the user was removed from the project between trigger creation and processing). Those users will be told to retry indefinitely. Consider letting `ForbiddenException`/`IdpNotFoundException` map to the non-retry `UNRESOLVED_ROLES_ERROR`-style wording and keeping the retry wording for transient failures/timeouts only.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -37,32 +69,110 @@ export class PublishDataStorageDraftsService {
...
+      .catch((error: unknown) => {
+        this.logger.warn(
+          `Failed to resolve roles for user ${command.userId} in project ${command.projectId}: ` +
+            (error instanceof Error ? error.message : String(error)),
+          error instanceof Error ? error.stack : undefined
+        );
+        throw new BusinessViolationException(PERMISSIONS_LOOKUP_FAILED_ERROR);
+      });
+
+    if (!project?.roles?.length) {
```
laskevych, line 174: 🔵 **Question (non-blocking):** `ProjectSchema.roles` is optional in `@owox/idp-protocol` (`packages/idp-protocol/src/types/models.ts`), and `IdentityOwoxClient.getProjectForUser` parses the remote response with exactly that schema — so a hosted OWOX Identity deployment is contractually allowed to omit roles. `McpProjectContextFacadeImpl` defensively falls back to JWT roles for this exact case, but this async path has no JWT to fall back to, so such a deployment would fail every bulk publish with `UNRESOLVED_ROLES_ERROR`. That’s still strictly better than the pre-PR silent all-fail (and the in-repo providers always return roles), but: do we know the hosted endpoint always populates `roles` for this call? If not, is fail-loud the intended behavior there, or should this log at error level so it’s noticed?

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -37,32 +69,110 @@ export class PublishDataStorageDraftsService {
...
 
     const draftIds = await this.dataMartService.findDraftIdsByStorage(dataStorage);
 
+    // Nothing to authorize: skip the remote role lookup so an unrelated IDP
+    // blip cannot turn "no drafts to publish" into a hard error.
+    if (draftIds.length === 0) {
+      return new PublishDataStorageDraftsResultDto(0, 0);
+    }
+
+    const roles = command.userId ? await this.resolvePublisherRoles(command) : [];
```
laskevych: 🔵 **Question (non-blocking):** The `: []` branch quietly contradicts the invariant `resolvePublisherRoles`’s doc-comment establishes (“must never fall back to `[]`”). Today it’s unreachable — the trigger entity’s `userId` column is non-nullable and both call sites pass it — but nothing validates non-emptiness: if a trigger were ever persisted with `userId: ''`, this takes the `[]` branch *and* `PublishDataMartService`’s `if (command.userId)` gate then skips `canAccess` entirely, publishing every draft with no authorization check. Since the branch is effectively dead, would it be safer to drop the ternary and let `resolvePublisherRoles` fail loudly on a missing userId too?

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -37,32 +69,110 @@ export class PublishDataStorageDraftsService {
...
+  /**
+   * Publish failures are returned to the browser and readable by any project
+   * viewer, so only reasons this codebase authored are echoed back. Errors from
+   * deeper in the publish path (storage drivers, warehouse dry-run validation)
+   * can embed SQL, table paths or credential hints and are replaced with a
+   * generic reason; the full error stays in the server log above.
+   */
+  private toUserFacingReason(error: unknown): string {
+    const message = error instanceof Error ? error.message : '';
+    return USER_FACING_FAILURE_REASONS.has(message) ? message : GENERIC_FAILURE_REASON;
```
laskevych, line 197: 💡 **Suggestion (non-blocking, follow-up material):** The allowlist keys on byte-exact message strings, but the codebase already has the mechanism this wants: `BusinessViolationException` accepts a `code` as its third argument, `base.exception.ts` documents it as “Clients branch on this instead of matching human-readable messages”, and `base-exception.filter.ts` serializes it (precedent: `PLUGIN_SUSPENDED`). Throwing the `PUBLISH_DATA_MART_ERRORS` with stable codes and allowlisting codes would keep the sanitization while decoupling it from display text — right now any copy edit to one of the three strings silently downgrades that reason to the generic message with no test or type error. Related: the trigger handler’s `instanceof BusinessViolationException` filter relies on the convention that BVE text is always safe, which isn’t structural — e.g. `actualize-data-mart-schema.service.ts:44` wraps arbitrary `error.message` into a BVE and `data-mart-definition-validator-facade.service.ts:42` throws BVE with raw warehouse text. It’s safe on today’s call graph (this PR’s own comment documents the audit), but a code-based allowlist would make the guarantee structural instead of conventional. (One wrinkle: `NO_PERMISSION` travels via Nest’s `ForbiddenException`, which doesn’t carry the `code` contract — it would need to switch to BVE-with-code or stay matched separately.)

apps/backend/src/data-marts/use-cases/publish-data-mart.service.ts
```diff
@@ -15,6 +15,20 @@ import { ConnectorExecutionService } from '../services/connector/connector-execu
...
 
+/**
+ * Failure reasons authored here, and therefore safe to show to end users.
+ *
+ * Anything else raised while publishing — storage driver errors, warehouse
+ * dry-run validation output — may carry SQL, table paths or credential hints,
+ * so callers that surface failures to the UI must not echo it back. See
+ * PublishDataStorageDraftsService.toUserFacingReason.
+ */
+export const PUBLISH_DATA_MART_ERRORS = {
```
laskevych, line 26: 🔵 **Question (non-blocking):** For non-SQL drafts, `run()` also throws `BusinessViolationException`s whose text this codebase authors — the definition validators’ “Invalid identifier format. Expected: project.dataset.table” family (bigquery/snowflake/databricks/athena/redshift) and the facade’s “DataMart storage credentials not found”. Because only these three strings are allowlisted, those all degrade to the generic “Publishing failed. Open the Data Mart to see details.” in the bulk path, while the changeset promises the notification “explains the actual reason”. The conservative default is right (the same BVE channel also carries raw dry-run driver text, so a blanket pass-through would leak), but was excluding the authored identifier-format class deliberate? A `code` on `ValidationResult` — the pattern `toUserFacingStorageError` already uses — would let those through safely.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -37,32 +69,110 @@ export class PublishDataStorageDraftsService {
...
+        ++failedCount;
+        continue;
+      }
+
+      // The draft is published and saved by this point. Scheduling schema
+      // actualization is a follow-up, so its failure must not be reported as a
+      // publish failure: that would send the user to a DRAFT-filtered list the
+      // Data Mart is no longer in, and a retry would say it is already published.
+      try {
         await this.schemaActualizeTriggerService.createTrigger(
```
laskevych, line 117: 🔵 **Question (non-blocking):** Separating publish outcome from post-publish effects is the right call, but the failure mode is now fully invisible to the user: if the scheduler hiccups during a 20-draft batch, the toast says “Published 20 Data Mart drafts” while none get schema actualization — schemas stay stale until each mart is touched manually, and only the server log knows. Is a follow-up planned to surface this (e.g. a `warnings` field next to `failureReasons`, or a reconciliation for marts without a scheduled actualization)? Also a small note: this catch hand-rolls the same “published but side effect failed” policy as `runPostPublishEffect` in `PublishDataMartService` — if the policy ever grows (metrics, retry), the two copies will drift.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -37,32 +69,110 @@ export class PublishDataStorageDraftsService {
...
+            command.projectId,
+            command.userId,
+            roles,
+            command.userId
+          )
         );
+        ++successCount;
+      } catch (error) {
+        this.logger.warn(
+          `Failed to publish draft ${draftId}: ${error instanceof Error ? error.message : String(error)}`,
```
laskevych, line 104: 🟢 **Nitpick (non-blocking):** The stack passed as `logger.warn`’s second argument is silently dropped in production: `bootstrap.ts` installs `CustomLoggerService`, whose `warn()` treats a trailing string as the context label and then overwrites it with the service name — the stack ends up in neither `trace` nor `metadata` (only `error(msg, stack)` maps to `trace`). This affects all four new warn-with-stack sites (three in this file, one in the trigger handler), which undercuts the PR’s “full detail preserved in the server log” intent on these paths. Established patterns in the codebase: interpolate the stack into the message, pass `{ stack }`, or use `logger.error` where a stack matters.

laskevych (approved):
Nice work — the root-cause analysis (hardcoded `roles: []` → VIEWER → silent all-fail) is spot on, the fix deliberately refuses to reintroduce the same failure mode through the roles lookup, and the test coverage is thorough, including the leak-prevention assertions. DI wiring, module imports, and the frontend polling contract all check out.

No blocking issues — approving. The comments below are all **non-blocking**: one 🟡 important edge case worth fixing before/shortly after merge, a few 🔵 questions about deliberate trade-offs, and 💡/🟢 suggestions that can be follow-ups.

Legend: 🟡 Important (should be addressed, not blocking) · 🔵 Question · 💡 Suggestion · 🟢 Nitpick

apps/web/src/features/data-storage/shared/utils/buildPublishFailureMessage.ts
prokopetsv: Confirmed and fixed — thank you, the rolling-deploy path is exactly right and I had missed it.

`reasons.length === 0` now produces no cause clause at all rather than falling through to the "different errors" branch. A trigger completed by the previous backend now reads *"Failed to publish 3 Data Mart drafts. Review them in the Data Marts list and try again."* Claiming the reasons differ when we have none was an assertion we cannot support.

Covered by a test asserting the exact string and that `different errors` is absent.

apps/web/src/features/data-storage/shared/utils/buildPublishFailureMessage.ts
prokopetsv: **(1) Fixed.** The pronoun now switches on `failedCount` (`Review it` / `Review them`), matching the message this replaced. Test added.

**(2) Acknowledged, not changed yet.** You are right that it contradicts `Data Mart is already published`. Making the suffix conditional needs a retryable/non-retryable classification per reason — a second allowlist alongside the message one — and the case requires a concurrent publish landing mid-batch. It is also mildly self-correcting, since reviewing the mart shows it is already published.

Happy to take it now if you would rather not wait; otherwise it folds naturally into the code-based work discussed in the `toUserFacingReason` thread.

apps/backend/src/data-marts/services/publish-drafts-trigger-handler.service.ts
prokopetsv: Good catch — a deleted storage told to "try again" forever is a bad outcome. Fixed.

Added `toTriggerErrorMessage`, which maps `NotFoundException` to *"This Storage no longer exists. No Data Mart drafts were published."* Deliberately a fixed string rather than the exception own message, since that one embeds the storage and project ids. `BusinessViolationException` still passes through, and everything else keeps the generic fallback.

Handler test updated, plus a new one asserting an unrecognized infrastructure error still genericizes and leaks no host name.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: Verified against `handleAxiosError` — 403 produces `ForbiddenException`, 404 produces `IdpNotFoundException`, both definitive. Fixed.

The branch keys on the `status` field, which `BaseException` exposes publicly, so `data-marts` does not have to import provider-specific exception types: 403/404 now map to the non-retry `UNRESOLVED_ROLES_ERROR`, and everything else keeps the retry wording. `status` is now logged too.

Table-driven test covers both codes and asserts the message contains no "try again".

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: No, I cannot confirm the hosted endpoint always populates `roles` for this call — worth someone with Identity access checking.

Fail-loud is intentional. The alternative is `[]`, which `resolveRole` reads as VIEWER, i.e. precisely the silent all-fail this PR removes, and there is no JWT here to fall back to the way `McpProjectContextFacadeImpl` does.

Taking your suggestion on visibility: that branch now logs at **error**, not warn. Identity answering successfully but omitting roles would break every bulk publish in a deployment — a contract problem, not routine noise.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: Agreed, and this was the most valuable comment in the set. Fixed.

You are right that it is worse than dead code: an empty `userId` takes the `[]` branch **and** makes `PublishDataMartService` skip `canAccess` entirely, publishing every draft unauthorized.

Dropped the ternary — `resolvePublisherRoles(command)` is now unconditional — and added an explicit guard that logs at error and throws when `userId` is missing. I deliberately did not rely on the IDP rejecting an empty user id: on the self-hosted provider, `getUserRole("")` flows through `toRoleOrViewer` and would come back as a valid-looking `["viewer"]`.

Test asserts it throws and that neither the IDP nor the publisher is called.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: Agreed on the direction, including the sharper half: the handler `instanceof BusinessViolationException` filter is conventional, not structural — `data-mart-definition-validator-facade.service.ts:42` throws BVE wrapping `result.errorMessage` (raw warehouse dry-run output for SQL-backed definitions), and `actualize-data-mart-schema.service.ts` wraps an arbitrary `error.message`. It is safe today only because per-draft errors are caught inside the loop and never reach that catch, which is a property of the current call graph rather than of the types.

**Correcting one point from my original reply (edited):** I agreed too quickly that a copy edit to one of the three strings would silently downgrade that reason. That is not true here — `USER_FACING_FAILURE_REASONS` is derived from `Object.values(PUBLISH_DATA_MART_ERRORS)` and the throw sites use those same constants, so editing the wording updates both sides together.

Your concern does hold via a different route, though: adding a **new** authored message and not registering it is silently downgraded, with no test or type failure. That is exactly the identifier-format case in the other thread.

Deferring to a follow-up rather than doing it here, because the blast radius is wider than it first looks:

- `NO_PERMISSION` is thrown as Nest `ForbiddenException` (403), while `base-exception.filter.ts:13` hardcodes `BusinessViolationException` to 400 — so giving it a code silently changes the single-publish endpoint status. Keeping it matched separately avoids that but leaves the allowlist half codes, half special case.
- Letting the validator messages through needs a `code` on `ValidationResult`, which is shared across 23 files (every storage validator plus storage-access validation).

Tracked internally.

apps/backend/src/data-marts/use-cases/publish-data-mart.service.ts
prokopetsv: Deliberate, and you have identified the right fix.

The three allowlisted strings are the ones I could prove safe by construction; the validator family shares a channel with dry-run driver output, so passing the class through wholesale would leak. Erring toward the generic message was the conservative call.

You are right that it undersells the changeset promise. `ValidationResult.code` — the same discriminator `toUserFacingStorageError` already relies on — is the clean way to let the authored identifier-format messages through, and it composes with the code-based allowlist discussed in the other thread. Same follow-up.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: No follow-up planned yet; agreed there should be. A `warnings` field beside `failureReasons` is the shape I would pick — it reuses the plumbing this PR added and keeps publish outcome separate from side-effect outcome. Reconciliation is more robust but a bigger piece of work.

Worth noting the blast radius is smaller than it looks: `actualizeSchema*IfExpired` runs on the real consumption paths (report runs via `stream-http-data.service.ts:272`, MCP, AI context) and `isSchemaExpired` returns true when `schemaActualizedAt` is null, so a missed trigger self-heals on first use. The gap is the UI showing a stale schema until then.

Also taking your duplication note — two copies of the same "published but side effect failed" policy.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: Confirmed by reading `CustomLoggerService`, and this one genuinely undercut the intent of the PR. Fixed.

`warn(message, ...params)` destructures the second argument as `context` and passes `undefined` as trace, so the stack was not just dropped — it replaced the context label. Only `error(message, trace)` maps to `trace`.

All four sites now pass `{ stack }`, which `destructureParams` spreads into metadata, matching `llm-tool-runner.ts:520`. Checking the repo, those four were the only places passing a bare stack string to `warn`; every other site already used `{ stack }`, `logger.error(msg, stack)`, or interpolation.

apps/web/src/features/data-storage/shared/hooks/usePublishDraftsTrigger.ts
```diff
@@ -96,26 +97,30 @@ export function usePublishDraftsTrigger(onSuccess?: () => void): UsePublishDraft
...
+                    duration: 5000,
+                    id: triggerId,
+                  });
                 }
 
                 onSuccessRef.current?.();
               }
             } catch (e) {
-              const errorMessage = e instanceof Error ? e.message : 'Publishing drafts failed';
+              const errorMessage =
```
amarchenk0: The trigger-level safe errors added by this PR never reach the user. The handler stores `uiResponse.error` and marks the trigger as `ERROR`, but `UiTriggerService.getTriggerResponse()` removes an `ERROR` trigger and returns HTTP 400 via `BadRequestException(response)`. `getPublishDraftsTriggerResponse()` therefore always throws an Axios error: the `if (response.error)` branch above is unreachable, while this catch displays only `e.message`—normally “Request failed with status code 400”—instead of `response.data.error`. This hides the Storage, role-resolution, and permission-lookup reasons that this PR deliberately prepares. Extract the allowlisted `error` from the API error body and add a UI/integration test for a terminal `ERROR` trigger.

amarchenk0 (changes requested):
Changes requested.

apps/web/src/features/data-storage/shared/hooks/usePublishDraftsTrigger.ts
prokopetsv: Good catch on the user impact — fixed, with one correction to the mechanics.

The `if (response.error)` branch is actually reachable, and is the normal path today: `Trigger.onSuccess()` has no transition guard, and `AbstractTriggerRunner.executeTrigger` unconditionally calls it after `handleTrigger` returns — our handler swallows its error, so the runner overwrites ERROR with SUCCESS within milliseconds and the reason arrives on a 200.

But that's load-bearing luck, not design: a poll landing between the handler's ERROR save and the runner's SUCCESS save does hit the 400 — which also **deletes the trigger row**, losing the reason permanently. And any future guard on `onSuccess` (or a rethrowing handler) would flip every trigger-level error onto the 400 path.

So I took the hardening as proposed: `handleError` now extracts the allowlisted `error` from the 400 body (it survives `GlobalExceptionFilter`'s spread) before falling back to the Axios message, and the duplicated response-fetch catch now routes through `handleError`. Added hook-level tests for a terminal ERROR trigger, the no-`error`-body fallback, and the existing 200 path.

amarchenk0 (approved):

PR #1516, opened 2026-08-07
fix(data-marts): make AI helper failures visible — human-readable persistent errors and traceable logs

laskevych:
## Preview

https://github.com/user-attachments/assets/99c77803-b788-4826-a35a-aa960943d8c1

## Summary

During a client demo (2026-08-05, production), AI generation of field aliases/descriptions failed twice for a BigQuery-backed SQL data mart, and the presenter never learned why: the run failed on a missing BigQuery permission (`bigquery.datasets.create`, required by the technical-view flow that backs sampling), and the error toast auto-dismissed in seconds. Finding the cause afterwards was also hard — the ERROR log entry carried only the trigger id, so filtering Cloud Logging by dataMartId missed it.

This PR improves how such failures are communicated to the user and to whoever reads the logs. It deliberately does **not** change when or how the internal dataset and technical view are created — that is a separate product decision.

### Web — error visibility

- AI helper failures (poll errors, trigger-creation errors, empty AI results) now show as **persistent, dismissible** errors rendered through the app's standard error-toast pattern (react-hot-toast, red, top-center, explicit close button — same as `showApiErrorToast`'s persistent mode) instead of auto-dismissing after seconds.
- BigQuery permission errors are rewritten into a human-readable message naming the project and the missing permission; the raw error stays available under an expandable "Show technical details".
- Navigating away mid-generation (which aborts the run) now leaves a persistent notice that the run was cancelled, so returning users are not left guessing whether generation succeeded.

### Backend — observability

- AI helper trigger logs (error + success/debug/cancel) now include `dataMartId` and `projectId`.
- `AiInsightsFacadeImpl.generateDataMartMetadata` now logs the caught error with details (previously only a timing entry with `status=failed`).

## Tests

- New `ai-helper-toast.test.tsx` (permission-error humanization + persistent toast options) and `use-ai-helper.test.tsx` (cancellation notice on unmount with a run in flight; silence otherwise).
- Related component suites (`DataMartSchemaSettings`, `BaseSchemaTable`, `DataMartDetails`) still pass.
- `tsc` and eslint clean on changed files.


prokopetsv (commented):
## Review summary

The PR does what it says: AI helper failures become persistent, dismissible, humanized toasts on the web side, and backend logs become searchable by `dataMartId`/`projectId`. I verified the load-bearing claim — a failed generation really does arrive as status SUCCESS + HTTP 200 with `{ error }`, because the handler swallows the error into `uiResponse` and `AbstractTriggerRunnerService.executeTrigger` calls `trigger.onSuccess()` unconditionally (abstract-trigger-runner.service.ts:141). The new `response.error` branch closes a genuinely silent production path, and both this path and the 400 path are covered by tests. Backend changes are log-only plus a rethrown catch, consistent with the existing `answerPrompt` pattern in the same facade — no architectural concerns.

**No 🔴 blockers.** Four 🟡 non-blockers below as inline comments: stale persistent toasts surviving a successful retry (the main one), a missed cancellation notice when unmounting during trigger creation, duplicated dismiss-button markup vs `showApiErrorToast`, and a stale DTO docblock.

apps/web/src/features/data-marts/edit/model/hooks/use-ai-helper.ts
```diff
@@ -147,9 +151,18 @@ export function useAiHelper(): UseAiHelperResult {
                 dataMartId,
                 triggerId
               );
-              outcome = response.result
-                ? { kind: 'ok', data: response.result }
-                : { kind: 'cancelled' };
+              if (response.result) {
+                outcome = { kind: 'ok', data: response.result };
```
prokopetsv, line 159: 🟡 **Non-blocker (HIGH, UX):** Persistent toasts are keyed per data mart but never dismissed on success. Evidence: `duration: Infinity` with id `ai-helper-error-${dataMartId}`, and nothing in the `ok` path clears it. A user who fixes the permission and retries successfully still sees the old "can't access BigQuery" error (and any earlier "cancelled" notice) on screen. Fix: on `kind: 'ok'` — or at the start of `generate()` — `toast.dismiss` both `ai-helper-error-${dataMartId}` and `ai-helper-cancelled-${dataMartId}`.

apps/web/src/features/data-marts/edit/model/hooks/use-ai-helper.ts
```diff
@@ -121,7 +121,11 @@ export function useAiHelper(): UseAiHelperResult {
 
   useEffect(() => {
     return () => {
+      // Cancelling on unmount is silent by design (the run's results have nowhere to go),
+      // but the user must not come back to untouched fields with no explanation.
+      const activeRun = runStateRef.current;
       cancelCurrentRun();
+      if (activeRun) showAiHelperCancelledToast(activeRun.dataMartId);
```
prokopetsv, line 132: 🟡 **Non-blocker (LOW):** If the user leaves while `createAiHelperTrigger` is still in flight, `runStateRef.current` is null, so no abort is sent and no cancellation notice shows — yet the trigger is created and runs server-side. The missing abort is pre-existing, but the new notice inherits the same blind window, undermining the "returning users are not left guessing" goal for that case. A flag set before the POST (cleared on completion) would cover it; fine as a follow-up.

apps/web/src/features/data-marts/edit/model/hooks/ai-helper-toast.tsx
```diff
@@ -0,0 +1,158 @@
...
+    };
+  }
+
+  return { message: raw };
+}
+
+/** Same dedicated close button the app's persistent API-error toasts use
+ * (see shared/utils/showApiErrorToast.ts) — text stays selectable, dismissal is explicit. */
+// eslint-disable-next-line react-refresh/only-export-components -- toast content, not a route component
+function DismissButton({ toastId, label }: { toastId: string; label: string }) {
```
prokopetsv, line 46: 🟡 **Non-blocker (MEDIUM, DRY):** `DismissButton` and the flex-span wrapper duplicate `showApiErrorToast`'s persistent-mode markup verbatim (shared/utils/showApiErrorToast.ts:88–118), as the comment itself acknowledges. Two copies of the same button style will drift on the next styling change. Fix: export a shared `PersistentToastShell`/`DismissButton` from `shared/utils` and use it in both places. Not blocking — the third consumer can be the trigger for the extraction.

apps/web/src/features/data-marts/shared/services/data-mart.service.ts
```diff
@@ -456,7 +456,11 @@ export class DataMartService extends ApiService {
 
   /**
    * Fetch the trigger response when generation is complete.
-   * Backend returns HTTP 400 with `{ error }` if the trigger errored.
+   *
+   * A failed generation normally arrives as HTTP 200 with `{ error }`: the trigger
+   * handler swallows the error into `uiResponse` and the scheduler runner then flips
```
prokopetsv, line 461: 🟡 **Non-blocker (LOW, docs):** This corrected comment now contradicts the docblock on `AiHelperTriggerResponseDto` (types/api/response/ai-helper-trigger.response.dto.ts:6–8), which still states "On SUCCESS, `result` is populated… On ERROR, the backend returns HTTP 400". That stale claim is exactly what hid this bug; the next reader may trust whichever comment they find first. Fix: update the DTO docblock to the same SUCCESS-with-`{ error }` contract.

prokopetsv (approved):
**Approving** ✅

The core fix is verified: failed generations really do arrive as SUCCESS + HTTP 200 with `{ error }` (the handler swallows the error into `uiResponse`, and `AbstractTriggerRunnerService.executeTrigger` calls `trigger.onSuccess()` unconditionally), so the new `response.error` branch closes a genuinely silent production path. Both failure routes are covered by tests, backend changes follow the existing `answerPrompt` logging pattern, and everything is backward compatible.

No blockers. Four non-blocking notes left inline — the one worth a fast follow-up is dismissing the persistent error/cancelled toasts on a successful retry, so users who fix the permission don't keep seeing the stale error. The rest (shared dismiss-button extraction, the trigger-creation blind window, the stale `AiHelperTriggerResponseDto` docblock) can land separately.

Nice work turning a demo incident into both a UX fix and searchable logs 🚢

apps/web/src/features/data-marts/edit/model/hooks/use-ai-helper.ts
laskevych: Fixed in 309b4a2a6 — went with the "start of `generate()`" option: `dismissAiHelperToasts(dataMartId)` clears both the stale error and the cancellation notice the moment a new run starts, so a fixed permission never shows next to an old failure. Covered by tests (hook + toast module).

apps/web/src/features/data-marts/edit/model/hooks/use-ai-helper.ts
laskevych: Agreed — deferring. The missing abort for an in-flight `createAiHelperTrigger` is pre-existing, and covering the notice for that window needs a pending-create flag with its own cleanup semantics; will handle both together in a follow-up.

apps/web/src/features/data-marts/edit/model/hooks/ai-helper-toast.tsx
laskevych: Agreed — deferring per your note. Extracting a shared `PersistentToastShell`/`DismissButton` means touching `showApiErrorToast`, which sits on the global interceptor path — safer as its own small refactor once this lands (we now officially have the second consumer, so it's next in line).

apps/web/src/features/data-marts/shared/services/data-mart.service.ts
laskevych: Fixed in 309b4a2a6 — the `AiHelperTriggerResponseDto` docblock now documents the real contract: failed generations arrive as HTTP 200 with `{ error }` (handler swallows the error, runner flips the trigger to SUCCESS), and a 400 occurs only for terminal ERROR/CANCELLED triggers. Same wording as the service comment so the two can't disagree again.

laskevych:
Thanks for the review — especially for independently verifying the SUCCESS-with-`{ error }` path.

Addressed in 309b4a2a6:
- **Stale persistent toasts**: a new run now dismisses the data mart's error + cancellation toasts at the start of `generate()` (your suggested option), with tests.
- **`AiHelperTriggerResponseDto` docblock**: rewritten to the real contract (HTTP 200 + `{ error }` for failed generations; 400 only for terminal ERROR/CANCELLED).

Deferred as agreed:
- **Trigger-creation blind window** — pre-existing abort gap; will fix together with the notice in a follow-up.
- **Shared `PersistentToastShell`/`DismissButton` extraction** — separate small refactor, since it touches `showApiErrorToast` on the global interceptor path.

PR #1517, opened 2026-08-07
feat(plugin-host): add managed plugin collections

amarchenk0:
## Summary

- add manifest-declared, host-managed JSON collections for installed plugins
- expose typed `list`, `get`, `put`, and `delete` operations through `@owox/plugin-sdk`
- support project- and member-scoped namespaces with optional entity bindings and action-based authorization
- persist documents, quota counters, and body-free audit events through a dedicated named TypeORM data source
- validate collection compatibility across plugin releases and serialize concurrent release synchronization
- include collection migrations in migration, dump/restore, and health-check flows
- document the SDK contract, storage limits, and deployment configuration

## Why

Sandboxed frontend plugins cannot use browser storage or safely hold OWOX credentials. Collections give them a host-owned persistence layer while keeping runtime credentials in the ODM host and applying the current member's authorization to every operation.

Task: [Plugins collections — 6788](https://owox.fibery.io/Sprint_Work/Work_Item/Plugins-collections-6788)

## Deployment behavior

Without `PLUGIN_COLLECTIONS_DB_*` overrides, the named collections connection falls back to the main `DB_*` configuration. This is suitable for local, self-hosted, and stage deployments.

Cloud production can set `PLUGIN_COLLECTIONS_DB_*` to place plugin documents in a dedicated MySQL database and Cloud SQL instance. Collection migrations use their own `plugin_collections_migrations` history and run together with the regular up migrations.

## Validation

- backend changed unit suites: 13 suites, 144 tests passed
- plugin collections/schema e2e: 2 suites, 16 tests passed
- plugin SDK: 4 suites, 35 tests passed
- plugin SDK typecheck passed
- backend build passed
- backend and SDK lint/format checks passed
- Markdown lint passed
- manually verified with [OWOX/data-mart-dashboardization](https://github.com/OWOX/data-mart-dashboardization): create, persist across reload, duplicate, update, and delete


apps/backend/src/plugin-host/collections/use-cases/delete-plugin-collection-document.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    } catch (error) {
+      if (!(error instanceof PluginCollectionAuthorizationDeniedError)) throw error;
+      await this.audit.record(command.context, {
+        collectionName: command.collectionName,
+        documentId: command.documentId,
+        parentType: resolved.declaration.entityBinding?.type,
+        action: 'DELETE',
+        outcome: 'AUTHORIZATION_DENIED',
+      });
+      throw error;
```
romandubovyi: 🔴 **Blocking:** `GetPluginCollectionDocumentService` deliberately rewrites `PluginCollectionAuthorizationDeniedError` to `PluginCollectionNotFoundError` so a member cannot probe for documents whose parent entity they cannot see. This path rethrows the 403 unchanged, while a missing document yields 404 from `persistence.delete`. A caller with a plugin runtime token can therefore enumerate existing document ids in a project-scoped, entity-bound collection by comparing 403 against 404 — exactly the distinction the GET path is written to hide, and the same property the ordering comment in `plugin-collection-persistence.service.ts:168` protects for PUT. Suggest throwing `PluginCollectionNotFoundError` here after recording the audit event, mirroring the GET use case, plus an e2e assertion that an unauthorized delete returns 404.

🤖 Generated with Claude Code (Opus 5) + Grok CLI (grok-4.5)

apps/backend/src/plugin-host/collections/services/plugin-collection-audit.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  @Transactional({ connectionName: PLUGIN_COLLECTIONS_DATA_SOURCE })
+  async record(
+    context: PluginCollectionRuntimeContext,
+    input: PluginCollectionAuditInput
+  ): Promise<void> {
+    const cutoff = new Date(
+      Date.now() - PLUGIN_COLLECTION_LIMITS.auditRetentionDays * 24 * 60 * 60 * 1000
+    );
+    await this.repository.delete({ projectId: context.projectId, createdAt: LessThan(cutoff) });
```
romandubovyi, line 39: ⚪ **Non-blocking:** Every audited operation runs a project-wide `DELETE ... WHERE projectId AND createdAt < cutoff` plus two `COUNT(*)` queries (per plugin/project capped at 50,000 rows, and per project capped at 500,000) before inserting a single row. That is three-plus extra statements per PUT/DELETE — and per LIST that denied at least one document — all inside the caller's write transaction, which on SQLite contends for the global write lock. Consider moving retention and trimming to a scheduled sweep, or amortising it (every N inserts, or only when a cached watermark is exceeded), so the hot path performs just the insert.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Claude Code (Opus 5) + Grok CLI (grok-4.5)

romandubovyi (commented):
Review complete. Nine inline comments below, plus one point that cannot be anchored inline.

**`apps/backend/src/plugin-host/utils/plugin-manifest.util.spec.ts`**

⚪ **Non-blocking:** This file now contains a literal `0x00` byte inside a string literal in `never throws, whatever it is handed`, so git classifies it as binary (`Bin 2875 -> 5569 bytes`) and the ~2.7 KB of new manifest tests added by this PR cannot be reviewed in the diff, grepped, or patched. Writing the same input as escapes (`'\u0000\uFFFF'`) keeps the test semantics identical and restores the file to text. (GitHub does not accept inline comments on binary files, hence this note.)

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Claude Code (Opus 5) + Grok CLI (grok-4.5) + Kimi CLI

apps/backend/src/plugin-host/collections/services/plugin-collection-persistence.service.ts
```diff
@@ -0,0 +1,339 @@
...
+    command: ListPluginCollectionCommand
+  ): Promise<PluginCollectionPageDto> {
+    const after = command.cursor ? this.decodeCursor(command.cursor) : null;
+    const selectedIds: string[] = [];
+    // Reserve enough for the page envelope, nextCursor and item separators.
+    let responseBytes = 128;
+    // Authorization is deliberately a bounded scan, not "scan until the requested number
+    // of visible rows is found". Otherwise a member who can see few parents can turn one
+    // request into thousands of main-DB permission checks. A short page with a cursor is valid.
+    const scanLimit = resolved.declaration.entityBinding
```
romandubovyi, line 67: ⚪ **Non-blocking:** For an entity-bound collection `scanLimit` is clamped to `maxAuthorizationScanRows` (10), so a page never returns more than 10 documents even when the caller asks for `limit=100` — the `bounds the raw rows scanned for an authorization-filtered page` e2e test pins exactly that. The bound itself is well reasoned, but the published contract says otherwise: `docs/plugins/authoring-guide.md:249` states "List pages contain at most 100 documents" and the SDK README example passes `limit: 50`. Please document that bound collections page in units of 10, and that a short or empty page with a non-null `nextCursor` is normal, so plugin authors do not treat it as end-of-data.

🤖 Generated with Claude Code (Opus 5) + Grok CLI (grok-4.5)

apps/backend/src/plugin-host/collections/entities/plugin-collection-document.collection.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+  Index,
+  PrimaryGeneratedColumn,
+  UpdateDateColumn,
+} from 'typeorm';
+
+@Entity('plugin_collection_document')
+@Index('UQ_plugin_collection_document_namespace_key', ['namespaceKey', 'documentKey'], {
+  unique: true,
+})
+@Index('idx_plugin_collection_document_list', ['namespaceKey', 'documentKey'])
```
romandubovyi: ⚪ **Non-blocking:** `idx_plugin_collection_document_list` covers exactly the same columns in the same order as the unique index `UQ_plugin_collection_document_namespace_key` declared two lines above, so it can never serve anything the unique index does not already serve — both the point lookup in `findForMutation` and the `namespaceKey` + `ORDER BY documentKey` scan in `list()`. It only adds write amplification and storage. The same duplication exists in `1785931300000-create-plugin-collection-tables.ts`.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Claude Code (Opus 5)

apps/backend/src/common/exceptions/global-exception.filter.ts
```diff
@@ -37,6 +38,33 @@ function isPayloadTooLargeError(e: unknown): e is Error & { status: 413; type: s
...
+        ? (exception.driverError as Record<string, unknown>)
+        : {};
+    const safeDetails = Object.fromEntries(
+      ['code', 'errno', 'sqlState'].flatMap(key => {
+        const value = driver[key];
+        return typeof value === 'string' || typeof value === 'number' ? [[key, value]] : [];
+      })
+    );
+    const code = safeDetails.code ?? safeDetails.sqlState ?? safeDetails.errno;
+    const error = new Error(`Database query failed${code ? ` (${String(code)})` : ''}`);
```
romandubovyi, line 58: ⚪ **Non-blocking:** Stripping the bound parameters is right, but the substitute `Error` is constructed inside `errorForLogging`, so its `stack` points at this filter rather than at the service or repository that issued the failing query. This applies to every `QueryFailedError` in the backend, not just collection queries, so all database-caused 500s lose their origin. The original stack carries no bound parameters, so it can be preserved — `error.stack = exception.stack`, or `{ cause: exception }` with the message and parameters still stripped.

apps/backend/src/dump/apply-dump.ts
```diff
@@ -35,28 +40,59 @@ async function enableForeignKeys(queryRunner: QueryRunner, dbType: string) {
...
+
+  try {
+    const claimedFiles = new Set<string>();
+    for (const { name, dataSource } of sources) {
+      const tableNames = new Set(dataSource.entityMetadatas.map(entity => entity.tableName));
+      const sourceFiles = files.filter(file => tableNames.has(path.basename(file, FILE_EXT)));
+      sourceFiles.forEach(file => claimedFiles.add(file));
+      await applyFiles(dataSource, sourceFiles, name);
+    }
+    const unknownFiles = files.filter(file => !claimedFiles.has(file));
```
romandubovyi, line 73: ⚪ **Non-blocking:** The `unknownFiles` guard runs after the loop that already applied and committed a transaction per data source, so a dump containing a file for an unrecognised table fails only once the recognised tables have been written, with nothing left to roll back. The table name sets are known right after `initialize()`, so computing `claimedFiles` and throwing before the first `applyFiles` call would keep the restore all-or-nothing.

apps/backend/src/plugin-host/use-cases/sync-plugin-releases.service.spec.ts
```diff
@@ -334,19 +372,41 @@ describe('SyncPluginReleasesService', () => {
...
+        },
+      ] as never);
+
+      const result = await run(s);
+
+      expect(result.report.rejections[0]).toMatchObject({
+        code: ReleaseRejectionCode.COLLECTIONS_INCOMPATIBLE,
+      });
+      expect(s.validator.validate).not.toHaveBeenCalled();
+      expect(s.versionService.insertVersionForLease).not.toHaveBeenCalled();
```
romandubovyi, line 407: ⚪ **Non-blocking:** (Anchored here because the lines below are outside this diff — this is about the `throttling` block, lines 427 and 434.) `tryClaimSyncSlot` now returns `string | null`, but both mocks there still resolve to `false`. The tests pass because `!false` and `!null` are equivalent, yet the mock no longer matches the real contract and will mislead the next refactor. `mockResolvedValue(null)` keeps them honest.

🤖 Generated with Kimi CLI

packages/test-utils/src/helpers/create-test-app.ts
```diff
@@ -24,6 +24,7 @@ export async function createTestApp(
   // each Jest worker runs in a separate process and createTestApp() is always called first.
   process.env.DB_TYPE = 'sqlite';
   process.env.SQLITE_DB_PATH = ':memory:';
+  process.env.PLUGIN_COLLECTIONS_SQLITE_DB_PATH = ':memory:';
```
romandubovyi, line 27: ⚪ **Non-blocking:** Setting `PLUGIN_COLLECTIONS_SQLITE_DB_PATH=':memory:'` gives the collections connection its own in-memory database, which is not the default deployment shape — with no `PLUGIN_COLLECTIONS_*` overrides both data sources open the *same* SQLite file through two `better-sqlite3` connections. That is precisely what the shared `queuesByDatabase` in `serializeSqliteTransactions` exists for, and it is currently covered only by `sqlite-transaction-serializer.spec.ts`, which mocks `dataSource.transaction` rather than opening a real file. A file-backed run that leaves this variable unset would exercise the actual cross-connection locking path.

apps/backend/src/plugin-host/collections/controllers/plugin-collections.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+  put(
+    @Param('collectionName') collectionName: string,
+    @Param('documentId') documentId: string,
+    @Body() dto: PutPluginCollectionDocumentApiDto,
+    @AuthContext() context: AuthorizationContext
+  ): Promise<PluginCollectionDocumentApiDto> {
+    return this.putDocument.run(this.mapper.toPutCommand(collectionName, documentId, dto, context));
+  }
+
+  @Post(':collectionName/documents/:documentId/delete')
```
romandubovyi: ⚪ **Non-blocking:** Deletion is exposed as `POST :documentId/delete` rather than `@Delete(':collectionName/documents/:documentId')`, while the rest of the chain supports DELETE end to end (`pluginHostBridge.ts` `ALLOWED_METHODS`, `protocol.ts`, `iframe-transport.deleteJson`). I suspect the reason is that `CollectionRequester` in `collections.ts` deliberately narrows to `getJson | putJson | postJson` so an injected transport without `deleteJson` still works (`client.ts` throws `OWOXConfigError` otherwise). If that is the intent, a short comment on either side would help — otherwise `@Delete` plus `requester.deleteJson` matches the other plugin-host controllers.

PR #200, opened 2025-06-25
feat(web): Implement data mart publishing 

romandubovyi:

PR #1519, opened 2026-08-10
fix(facebook-marketing): skip inaccessible ad accounts instead of failing the import

cristianolugara:
## Problem

In `FacebookMarketingConnector.startImportProcessOfTimeSeriesData`, accounts are iterated one after another and `fetchData` is not guarded. A single ad account that the access token can no longer reach — typically a client that stopped sharing it — throws and breaks out of all three loops, so:

- the accounts already fetched for that day are discarded;
- every account queued behind it is never requested;
- `updateLastRequstedDate` is never reached, so progress is not recorded and the next scheduled run starts from the same date and fails at the same account.

The import therefore stays stuck until someone edits the account list by hand. With a long account list this means one departing client silently freezes the data for all the others.

Graph API returns `(#200) Ad account owner has NOT grant ads_management or ads_read permission` in this case, and `isValidToRetry` is false for code 200, so the run dies on the first blocked account.

## Fix

- Each account iteration is wrapped in `try/catch`: a failing account is logged as `Skipped account <id>: <reason>` and the loop continues with the remaining accounts.
- If **every** account was skipped, the run throws instead of completing. That case does not mean "each client revoked access on the same day": it points to a global cause such as an expired access token. Without this guard the connector would report a successful run that imported nothing, turning a total outage into a silent one — which is worse than the original bug, because nothing surfaces it.

Progress is still only recorded when at least one account succeeded, so a partially failing day does not silently advance past data that was never imported.

## Testing

Verified on a self-hosted instance importing ~40 Facebook ad accounts:

- **One inaccessible account** (client that revoked access): before the change every run failed and no data was imported for any account; with the change that account is logged and skipped and the remaining ones import normally.
- **All accounts failing** (access token expired): every account is skipped and the run now fails with an explicit error. This is precisely the case that motivated the guard — an earlier version without it reported `SUCCESS` while importing zero rows, and the outage went unnoticed for five days.
- **Normal run** (no failing accounts): unchanged behaviour, same rows imported.

### DIFF
--- a/.changeset/facebook-skip-inaccessible-ad-accounts.md
+++ b/.changeset/facebook-skip-inaccessible-ad-accounts.md
@@ -1,0 +1,17 @@
+---
+'owox': minor
+---
+
+# Facebook imports survive an inaccessible ad account
+
+Previously, the Facebook Marketing connector aborted the whole import as soon as a single ad
+account returned an error. Because accounts are fetched one after another, one account the access
+token could no longer reach — typically a client that stopped sharing it — discarded the accounts
+already fetched and every account still queued behind it. The failure repeated on each following
+run, since the import never got far enough to record its progress, so the data stayed frozen until
+someone removed that account from the configuration by hand.
+
+Now, an account that fails is logged and skipped, and the import carries on with the remaining
+ones. If every account fails, the run stops with an error instead of reporting success: that points
+to a global cause, such as an expired access token, and finishing quietly there would hide an
+outage behind a run that imported nothing at all.
--- a/.changeset/facebook-skip-inaccessible-ad-accounts.md
+++ b/.changeset/facebook-skip-inaccessible-ad-accounts.md
@@ -5,13 +5,17 @@
 # Facebook imports survive an inaccessible ad account
 
 Previously, the Facebook Marketing connector aborted the whole import as soon as a single ad
-account returned an error. Because accounts are fetched one after another, one account the access
-token could no longer reach — typically a client that stopped sharing it — discarded the accounts
-already fetched and every account still queued behind it. The failure repeated on each following
-run, since the import never got far enough to record its progress, so the data stayed frozen until
-someone removed that account from the configuration by hand.
+account returned a permission error. Because accounts are fetched one after another, one account
+the access token could no longer reach — typically a client that stopped sharing it — discarded the
+accounts already fetched and every account still queued behind it. The failure repeated on each
+following run, since the import never got far enough to record its progress, so the data stayed
+frozen until someone removed that account from the configuration by hand.
 
-Now, an account that fails is logged and skipped, and the import carries on with the remaining
-ones. If every account fails, the run stops with an error instead of reporting success: that points
-to a global cause, such as an expired access token, and finishing quietly there would hide an
-outage behind a run that imported nothing at all.
+Now, an account that returns a permission error is skipped and the import carries on with the
+remaining ones, for catalog and time-series data alike. Each skip is raised as a warning on the run,
+so an account that is quietly missing from the destination is visible without reading the run log.
+Only permission failures are skipped: a storage write or an exhausted transient error still fails
+the run, which keeps the import from moving past a day whose data was never stored. If every account
+fails, the run stops with an error naming each one, because that points to a global cause such as an
+expired access token — and finishing quietly there would hide an outage behind a run that imported
+nothing at all.
--- a/packages/connectors/src/Sources/FacebookMarketing/Connector.js
+++ b/packages/connectors/src/Sources/FacebookMarketing/Connector.js
@@ -11,8 +11,12 @@ var FacebookMarketingConnector = class FacebookMarketingConnector extends Abstra
       constructor(config, source, storageName = "GoogleBigQueryStorage", runConfig = null) {
     
     super(config, source, null, runConfig);
-    
+
     this.storageName = storageName;
+
+    // accountId -> errors that made this account unreachable, kept so a partial failure can be
+    // reported per account instead of collapsing into whichever error happened to come last
+    this.skippedAccounts = new Map();
 
     }
 
@@ -60,6 +64,83 @@ var FacebookMarketingConnector = class FacebookMarketingConnector extends Abstra
 
       }
 
+      // A run that skipped an account still completes, so without this the only trace would be a
+      // log line: the run would report plain success while that account's data is missing.
+      if( this.skippedAccounts.size ) {
+        this.config.addWarningToCurrentStatus(
+          `${this.skippedAccounts.size} out of ${accountsIds.length} accounts were skipped and their data is missing. `
+          + `Skipped accounts: ${[...this.skippedAccounts.keys()].join(', ')}`
+        );
+      }
+
+    }
+
+  //---- skipOrRethrow -------------------------------------------------
+    /**
+     * Decides whether a failure lets the import move on to the next account.
+     *
+     * Only account-scoped permission failures are safe to skip: the account stays unreachable
+     * however often it is retried, so continuing costs nothing. Everything else — a storage
+     * write, an exhausted transient error — must fail the run: swallowing it would let
+     * LastRequestedDate advance past a day whose data was never stored, and once
+     * ReimportLookbackWindow has passed that day is never requested again.
+     *
+     * @param {string} accountId - The account being imported
+     * @param {Error} error - The failure to classify
+     * @param {string} context - What was being attempted, for the warning text
+     * @throws {Error} The original error, when it is not an account-scoped permission failure
+     * @private
+     */
+    _skipOrRethrow(accountId, error, context) {
+
+      if( !error?.isWarning ) {
+        throw error;
+      }
+
+      if( !this.skippedAccounts.has(accountId) ) {
+        this.skippedAccounts.set(accountId, []);
+      }
+      this.skippedAccounts.get(accountId).push(error);
+
+      this.config.addWarningToCurrentStatus(`${context}: skipped account ${accountId}: ${error.message}`);
+
+    }
+
+  //---- throwIfAllAccountsSkipped -------------------------------------------------
+    /**
+     * Fails the run when no account was importable.
+     *
+     * Every account failing at once is not a set of individual accounts losing access on the same
+     * day: it points to a global cause, typically an expired access token. Reporting success there
+     * would hide a total outage behind a run that imported nothing.
+     *
+     * @param {array} accountIds - Every account the run was asked to import
+     * @param {Set} skipped - The accounts skipped in the current pass
+     * @param {string} context - What was being imported, for the error text
+     * @throws {Error} When every account was skipped
+     * @private
+     */
+    _throwIfAllAccountsSkipped(accountIds, skipped, context) {
+
+      if( !accountIds.length || skipped.size < accountIds.length ) {
+        return;
+      }
+
+      const details = accountIds.map(accountId => {
+        const errors = this.skippedAccounts.get(accountId) || [];
+        const last = errors[errors.length - 1];
+        return `${accountId}: ${last ? last.message : 'unknown error'}`;
+      });
+
+      const error = new Error(
+        `All ${accountIds.length} accounts were skipped ${context}, so nothing was imported. `
+        + `This points to a global failure, such as an expired access token, rather than individual `
+        + `accounts being inaccessible. Errors: ${details.join('; ')}`
+      );
+      // Every skipped account was skipped because of a permission failure, which is something the
+      // customer can act on: keep the readable message instead of a stack.
+      error.isWarning = true;
+      throw error;
 
     }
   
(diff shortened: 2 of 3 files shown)


packages/connectors/src/Sources/FacebookMarketing/Connector.js
```diff
@@ -115,29 +115,49 @@ var FacebookMarketingConnector = class FacebookMarketingConnector extends Abstra
...
+              this.config.logMessage(data.length ? `${data.length} records were fetched` : `No records have been fetched`);
 
-            if( data.length || this.config.CreateEmptyTables?.value ) {
-              const storage = await this.getStorageByNode(nodeName);
-              await storage.saveData(data);
             }
 
-            this.config.logMessage(data.length ? `${data.length} records were fetched` : `No records have been fetched`);
+          // an account the token can no longer access must not abort the whole import
+          } catch( error ) {
```
amarchenk0, line 242: The new catch treats every failure in the try block as an inaccessible account: not only the intended Graph permission error, but also an exhausted retryable Facebook error, `getStorageByNode()` failure, or `storage.saveData()` failure. If another account succeeds, the code still advances `LastRequestedDate` and the run completes successfully although that account/day was not stored. Once the `ReimportLookbackWindow` passes, that date is no longer retried. Catch only the recognised account-scoped permission error around fetch; all other failures must fail the run and preserve the cursor. Add tests for a storage/retry failure while another account succeeds.

.changeset/facebook-skip-inaccessible-ad-accounts.md
```diff
@@ -0,0 +1,17 @@
...
+---
+
+# Facebook imports survive an inaccessible ad account
+
+Previously, the Facebook Marketing connector aborted the whole import as soon as a single ad
+account returned an error. Because accounts are fetched one after another, one account the access
+token could no longer reach — typically a client that stopped sharing it — discarded the accounts
+already fetched and every account still queued behind it. The failure repeated on each following
+run, since the import never got far enough to record its progress, so the data stayed frozen until
+someone removed that account from the configuration by hand.
```
amarchenk0: The release note promises that Facebook imports continue past an inaccessible account, but the fix applies only to time-series nodes. `startImportProcess()` executes every catalog node first, and `startImportProcessOfCatalogData()` still awaits `fetchData` directly. A supported configuration containing catalog nodes still aborts the whole run on a revoked account before time-series processing starts. Apply the chosen policy to catalog imports with a test, or narrow the title/changelog to time-series imports.

amarchenk0 (changes requested):
Changes requested.

packages/connectors/src/Sources/FacebookMarketing/Connector.js
```diff
@@ -115,29 +115,49 @@ var FacebookMarketingConnector = class FacebookMarketingConnector extends Abstra
...
+
+            skippedAccounts.push( accountId );
+            lastAccountError = error;
+            this.config.logMessage(`Skipped account ${accountId}: ${error?.message || error}`);
```
prokopetsv: 🔴 **Blocker — skipped accounts are invisible in the run status.**

A skipped account surfaces only as a run-log line; the run status stays plain SUCCESS with no warning. A permanently revoked account therefore makes every future scheduled run report clean success while that account's data is silently missing — the exact "silent outage" failure mode the PR description argues is worse than a loud failure, just scoped to one account instead of all of them.

The codebase already solves this precise problem: the TikTok Ads connector routes non-fatal per-advertiser failures through `this.config.addWarningToCurrentStatus(...)` and aggregates them in `_logFailure` / `_checkAndReportErrors` (`packages/connectors/src/Sources/TikTokAds/Connector.js:302-380`), and `NodeJsConfig` implements the mechanism.

Reusing that pattern also resolves two adjacent weaknesses of this diff for free:

- it classifies errors (permission failure → warning, anything else → real error), which is the remediation the existing review already asks for;
- its all-failed throw aggregates one error per account, whereas `lastAccountError` here discards every failure reason except the last one.

packages/connectors/src/Sources/FacebookMarketing/Connector.js
```diff
@@ -115,29 +115,49 @@ var FacebookMarketingConnector = class FacebookMarketingConnector extends Abstra
...
 
           }
 
         }
 
+        // Every account failing points to a global cause, such as an expired access token, rather
+        // than to individual accounts losing access. Reporting success here would hide the outage:
+        // the run would be marked as successful while importing nothing at all.
+        if( accountsIds.length && skippedAccounts.length === accountsIds.length ) {
+          throw new Error(`All ${accountsIds.length} accounts were skipped for ${DateUtils.formatDate(startDate)}, so nothing was imported. This usually means a global failure, like an expired access token, rather than individual accounts being inaccessible. Last error: ${lastAccountError?.message || lastAccountError}`);
```
prokopetsv: 🟡 **Non-Blocker — no tests for either headline behavior.**

Beyond the storage-failure test already requested in the existing review, neither headline behavior of this change is tested: skip-and-continue on a partial failure, and this all-accounts-skipped throw — including that it fires *before* `updateLastRequstedDate`, preserving the cursor.

The harness exists: `packages/connectors/test/Sources/FacebookMarketing/Source.test.js`, and `packages/connectors/test/Sources/TikTokAds/warningClassification.test.js` tests exactly this class of logic and is a ready template. This guard is the part protecting against the five-day silent outage described in the PR, so it's the last logic that should rely on manual verification only.

prokopetsv (commented):
Inline comments below. Existing review by @amarchenk0 accounted for — the over-broad catch / cursor advancement and the uncovered catalog nodes are not repeated here.

prokopetsv (changes requested):
Requesting changes, converging with @amarchenk0's review.

The intent and the failure-handling shape are right: the all-skipped guard throws before `updateLastRequstedDate`, so a total outage preserves the cursor and retries, and the per-day reset means a token expiring mid-run keeps the days already completed. The problem analysis in the description is accurate and well-evidenced.

Two things keep this from merging as-is:

1. 🔴 Partial failures are invisible at the run-status level. A skipped account only appears in the run log, so a permanently revoked account turns into a permanent silent gap behind green runs — the same silent-outage mode this PR sets out to eliminate, scoped to one account. Please surface skips via `addWarningToCurrentStatus`, following the TikTok Ads pattern (`_logFailure` / `_checkAndReportErrors`). Adopting that pattern also addresses the error-classification concern already raised: only recognized account-scoped permission errors should be skippable, everything else must fail the run.
2. The new control flow ships without tests. Together with the storage-failure scenario already requested, please cover skip-and-continue and the all-skipped throw; `test/Sources/TikTokAds/warningClassification.test.js` is a ready template.

packages/connectors/src/Sources/FacebookMarketing/Connector.js
cristianolugara: You were right, and this was the more dangerous of the two problems: swallowing a storage or exhausted-retry failure let `LastRequestedDate` advance past a day whose data was never stored, so the change introduced a silent data loss of its own once `ReimportLookbackWindow` passed.

Skipping is now limited to failures already classified as account-scoped permission errors — `_isAuthError` in the Facebook source flags `OAuthException` payloads, so the classification exists upstream and I did not add a second one. `_skipOrRethrow` rethrows anything else, which fails the run and preserves the cursor.

Covered by tests: a storage write failing while another account succeeds, and an exhausted transient error — both assert the run fails and `updateLastRequstedDate` was never called.

.changeset/facebook-skip-inaccessible-ad-accounts.md
cristianolugara: Good catch — the release note promised more than the diff delivered. Rather than narrow the wording, I applied the policy to catalog imports too, since they run first and an unreachable account there aborted the run before a single day was requested.

`startImportProcessOfCatalogData` now goes through the same `_skipOrRethrow` and all-skipped guard as the time-series path. Two tests cover it: skip-and-continue across accounts, and the all-skipped throw. A storage failure there is rethrown as well, with its own test.

The changeset has been rewritten to describe the behaviour that actually ships, including the warning surfacing and the fact that only permission failures are skipped.

packages/connectors/src/Sources/FacebookMarketing/Connector.js
cristianolugara: Agreed, and thanks for naming the existing pattern — reusing it was clearly better than inventing a second mechanism.

Skips now go through `this.config.addWarningToCurrentStatus`, twice over: once per skipped account at the point of failure, and once as a run-level summary (`N out of M accounts were skipped`) at the end of `startImportProcess`. A permanently revoked account therefore shows up on every run instead of only in the log.

On classification, the Facebook source already distinguishes these cases: `_isAuthError` flags `OAuthException` payloads, which is exactly the account-scoped permission failure, so `_skipOrRethrow` keys off `isWarning` and rethrows everything else. That is the same remediation @amarchenk0 asked for, so both are handled by the one change.

On aggregation, you were right that `lastAccountError` threw away every reason but the last: the all-skipped error now lists one entry per account, and carries `isWarning` so it surfaces as a readable message rather than a stack.

packages/connectors/src/Sources/FacebookMarketing/Connector.js
cristianolugara: Fair point — this guard is the part meant to prevent a silent outage, so leaving it on manual verification only was the wrong call.

Added `packages/connectors/test/Sources/FacebookMarketing/skipInaccessibleAccounts.test.js`, following the `warningClassification.test.js` structure. 12 tests covering:

- skip-and-continue on a partial failure, asserting the remaining account is still stored and the cursor moves;
- the skip surfacing as a run warning;
- a storage-write failure and an exhausted transient failure, each asserting the run fails and `updateLastRequstedDate` was never called;
- the all-skipped throw firing **before** the cursor moves, and naming every account;
- a token expiring mid-run keeping the days already completed (the per-day reset);
- the catalog paths: skip-and-continue, all-skipped, and storage failure;
- `_skipOrRethrow` classification in isolation.

`npx vitest run` in `packages/connectors`: 130 tests pass, no regressions.

cristianolugara:
Thanks both — the reviews found a real regression risk in the first version, not just rough edges. Pushed as `abe10d9`, replies on the individual threads.

Summary of what changed:

| Review point | Change |
|---|---|
| Over-broad catch advancing the cursor | Only failures classified as account-scoped permission errors are skipped (`_skipOrRethrow` keys off `isWarning`, which the Facebook source already sets for `OAuthException`). Storage writes and exhausted transient errors fail the run and preserve the cursor. |
| Skips invisible at run-status level | Routed through `addWarningToCurrentStatus`, per account and as a run-level `N out of M` summary, per the TikTok Ads pattern. |
| Catalog nodes uncovered | `startImportProcessOfCatalogData` uses the same skip policy and all-skipped guard. |
| `lastAccountError` discarding reasons | The all-skipped error lists one entry per account and carries `isWarning`. |
| No tests | 12 tests in `skipInaccessibleAccounts.test.js`, including the storage-failure and cursor-preservation cases. |

The one deliberate judgement call worth flagging: the all-skipped guard stays **per day** rather than per run. A token expiring mid-run then keeps the days already imported, and only the day that actually failed is retried — the alternative would discard completed work. Happy to change it if you'd rather have a single run-level check.

`npx vitest run` in `packages/connectors`: 130 passed, 11 files.

amarchenk0 (approved):

prokopetsv (approved):
Approving.

I re-checked the parts that carry the risk rather than reading the diff alone:

- **The skip really is limited to permission failures.** `_skipOrRethrow` keys off `isWarning`, and that flag is genuinely set for the motivating error — `Source.test.js` already covers Facebook code 200 returning `true` from `_isAuthError`, and `AbstractSource.js:94` promotes it once retries stop. The new tests stub `isWarning` directly, so it's worth saying explicitly that the upstream seam is covered too.
- **The data-loss path is closed.** An exhausted transient error keeps `isWarning` false, because `_isAuthError` defers to `_evaluateRetry`, so it rethrows and `LastRequestedDate` stays put. Same for a storage write. Both are asserted with the cursor check, which is the assertion that actually matters.
- **Skips are visible.** Per-account warning at the point of failure plus the run-level `N out of M` summary, through the existing `addWarningToCurrentStatus` mechanism rather than a second one.
- **Catalog nodes now follow the same policy**, so the changeset no longer promises more than the diff delivers.
- Reusing `isWarning` on the all-skipped error is sound by construction here: only warning-classified errors can produce a skip, so the flag needs no `.every()` guard like TikTok's.

On the per-day question you flagged: keep it per day. Per-run would discard days that imported cleanly before the token expired, and your test pins that behavior. The narrower guard is the right trade.

cristianolugara:
Thanks both for the approvals, and for re-checking the risky paths rather than just re-reading the diff — the cursor-preservation assertions were exactly the part I wanted a second pair of eyes on.

The Prettier check was failing on the test file I added. Fixed in `4a27e5d`: formatting only, no behaviour change. Verified locally with the package's own command (`prettier --check "**/*.{js,mjs,json}"`) rather than a neighbouring package's — the patterns differ, and the wrong one reports files this repo considers fine. `packages/connectors`: format check clean, `vitest run` 130 passed across 11 files.

The workflows for that commit are sitting in `action_required`, so they need a maintainer to approve the run whenever one of you gets a moment.

On your note about the upstream seam: agreed, and worth stating for the record — the new tests stub `isWarning` deliberately, to pin this connector's branching independently of how the flag gets set. The seam itself is already covered by `Source.test.js` (Facebook code 200 → `_isAuthError` → `true`) plus `AbstractSource.js:94` promoting it once retries stop, so the two suites meet without overlapping.

amarchenk0:
@cristianolugara Thanks a lot for your contribution!

PR #1522, opened 2026-08-11
feat(data-marts): Unique Count for joined Data Marts (#6792)

zapolsky:
## What

Adds a `Unique Count` field to every joined Data Mart in the report column picker: `COUNT(DISTINCT <that Data Mart's primary key>)`, computed correctly across join fan-out.

It answers the ticket's two cases — "customers with their order count", "customers with the number of unique products across their orders" — without adding the order or product key to the report as a column. Composite primary keys are supported. The metric can be selected and sorted by; it cannot be filtered or fed into another aggregation, and every entry point rejects that explicitly rather than falling through to a misleading "disconnected column" error.

## Why the numbers move, and why the new ones are right

A joined aggregate computed after the join is distorted by fan-out. The metric is therefore built in its own CTE over the source's raw rows and joined back at the report's dimension grain. Three pre-existing defects fall out of that, all fixed here:

- **`Unique Count` on a composite key was too high.** A row with an empty key component counted as one extra record, while a single-column key had always ignored such rows — the same declaration behaved differently depending only on how many columns it had. Counts go down.
- **`Sum` and `Average` on a joined Data Mart were too low.** All rows sharing an empty key collapsed into one, so the metric read a single row instead of all of them. Verified live on BigQuery: twenty such rows summed to the value of one row before, to all twenty after. `Min`, `Max`, the percentiles and `Combined` read the same rows and shift for the same reason.
- **A key unique only *within* what it is joined on was counted on its own.** A line number restarting at 1 in every order reported 12 — the largest order's line count — for a book of 4,300 line items. The counted identity is now scoped by the parent join key, so it reports 4,300, matching what `Sum`/`Average` on that key always meant.

A Data Mart with no declared primary key, or one whose key is always filled, is untouched. The changeset carries all of this in customer language.

## Scope

| Area | Change |
|---|---|
| SQL | Counting sleeve with a two-level `SELECT DISTINCT` → `COUNT`, join-key scoping, NULL-safe join-back (incl. a NaN leg — `IS NOT DISTINCT FROM` is not portable, and `NaN = NaN` is FALSE on BigQuery/Trino) |
| Dialects | All five — BigQuery, Snowflake, Redshift, Athena, Databricks. `textCastType()` is abstract, not defaulted, so a new dialect cannot pass the single-key tests and fail on its first composite key |
| Config | `uniqueCountConfig: boolean \| string[]` (alias paths, `''` = main mart). Column is already `json`, so no migration; legacy `true`/`false`/`null` parse through a union schema |
| Web | Per-source row in the picker, four distinct disabled reasons, sort entry under the SQL-safe name with the display label shown |
| MCP | `query_data_mart` selects and sorts the metric; the other four clauses are rejected with a dedicated error code |
| OpenAPI | The whole `blendable-schema` response was publishing as an empty object (no Swagger plugin, no decorators) — now fully documented, `uniqueCountAvailability` included, pinned by a spec test |
| Docs | `report-aggregations.md` + changeset |

`api-client` and `owox-ctl` need no change: the DM-level HTTP-data parameters (`columns`, `filter`, `sort`, `aggregation`, `dateTrunc`, `limit`) never carried a `uniqueCount` for any Data Mart, and `traverseData` streams the saved report's own configuration, so the metric flows through untouched.

## Testing

- Backend: 530 suites / 7352 tests. Per-dialect coverage is `describe.each` over the five real builder+renderer pairs.
- Web: 206 files / 1679 tests. `nest build` and `tsc -b` both clean.
- Live BigQuery, twice, in the dev project. The second run showed the pre-fix query returning 3 and 2 where the truth is 6 and 3, and a real backend was booted in-process to prove the MCP chain end to end.

## Notes for the reviewer

- Two names per metric: the SQL-safe `orders__unique_count` (alias, header `name`, what a sort rule stores) and the display string (header `alias`). The free-form join prefix must never reach SQL. A backend drift guard pins the name against the hand-written web mirror.
- Main and joined primary keys follow **different** backend rules — the main one keeps nested keys and prunes hidden ones, the joined one does the opposite. They are separated into scope-tagged vocabularies so the compiler rejects a swap; an earlier revision shipped exactly that bug.
- A joined Data Mart with a source column literally named `__owox_rid` is now rejected with an error naming the column, in the one case that previously returned wrong numbers instead.


apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -652,18 +718,41 @@ export class OutputControlsValidatorService {
...
+        // UniqueCountFieldUnsupportedClauseError already gives.
+        const isUniqueCountFilter = (rule: FilterRule) =>
+          uniqueCountOutputColumns.has(rule.column) && !homeFieldTypes.has(rule.column);
+        for (const rule of parsedFilters.filter(isUniqueCountFilter)) {
+          errors.push({
+            code: 'UNIQUE_COUNT_FILTER_UNSUPPORTED',
+            column: rule.column,
+            message: `"${rule.column}" is a Unique Count metric: it can be selected and sorted by, but not filtered or sliced. Remove the filter on it.`,
+          });
+        }
```
prokopetsv, line 776: 🟡 **Non-Blocker:** This dedicated, honest rejection exists for filters only. An aggregation or date-trunc rule targeting a Unique Count output column (e.g. `orders__unique_count` in `aggregationConfig`) instead falls through to the pre-existing `type === undefined` branches in `validateAggregations`/`validateDateTruncs` and is rejected as `AGGREGATION_FUNCTION_NOT_ALLOWED_FOR_TYPE` / `DATE_TRUNC_REQUIRES_DATE_COLUMN` with `type: 'unknown'` — pointing the caller at a schema/type problem that doesn't exist. Still a correct 400, so no data risk, but it's the exact "misleading fall-through" this comment block says the filter path avoids, and the PR description claims every entry point rejects explicitly. Exposure is raw HTTP report create/update payloads only (the web picker never offers the metric in aggregation dropdowns; the MCP tool rejects earlier with `UniqueCountFieldUnsupportedClauseError`). Suggest a `UNIQUE_COUNT_AGGREGATION_UNSUPPORTED`-style code mirroring this one.

prokopetsv (commented):
Reviewed the full diff (blending SQL core across all five dialects, backend validation/services, MCP layer, web UI, and behavioral test coverage). **No blockers.** Identifier quoting, the NULL/NaN-safe join-back, the `__owox_rid` guard, the validation chokepoint covering every entry point, cache invalidation, and the two-name (SQL-safe vs display) contract all check out. Five non-blockers total — three inline, two here:

🟡 **Non-Blocker (test coverage):** `apps/backend/test/output-controls.e2e-spec.ts` covers persistence round-trips (legacy boolean/array/null) and generic validator wiring, but no HTTP-level test sends a `filterConfig` entry targeting a Unique Count column and asserts 400 + `UNIQUE_COUNT_FILTER_UNSUPPORTED` through the real controller/DTO/validator chain. The rule is thoroughly unit-tested in `output-controls-validator.service.spec.ts` (~4108-4210), so this is belt-and-braces for the boundary the PR's contract leans on.

🟡 **Non-Blocker:** Passing the human-readable display name (e.g. `Orders Unique Count`) in the MCP tool's `fields` misses `hasUniqueCountFieldCandidate` (no `__unique_count` suffix) and falls through to a generic `field_not_found` instead of the purpose-built `UnmatchedUniqueCountFieldError` that lists valid pseudo-field names. The tool bills per call, so a model that reasonably tries the display name pays for a wasted round-trip before it can retry. Already acknowledged as a known tradeoff in `query-data-mart.tool.spec.ts:58-66` — flagging as a candidate follow-up (e.g. match display names in `splitUniqueCountFields` and return the dedicated error).

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -214,17 +221,55 @@ If truncated is true, not all matching rows were returned: narrow the query (few
...
+        fields = split.columns;
+        uniqueCountConfig = split.uniqueCountConfig;
+
+        // A matched pseudo-field is no longer in `columns`, so if it is ALSO named in filters/
+        // slices/aggregations/date_buckets, letting the request through would hit the validator's
+        // "column not selected" family — whose fix-it message says to add the field back to
+        // "fields", where the model already put it. Catch it here instead, before that
+        // unresolvable loop can start. `sort` is not in that set: the metric IS orderable, and
+        // its sortConfig entry is forwarded untouched.
+        const violations = findUniqueCountClauseViolations(split.matchedNames, {
```
prokopetsv, line 250: 🟡 **Non-Blocker (test coverage):** This throw site is never exercised end-to-end. `findUniqueCountClauseViolations` and `UniqueCountFieldUnsupportedClauseError` are each unit-tested in isolation in `query-data-mart.input.spec.ts`, but no test in `query-data-mart.tool.spec.ts` feeds `handler()` a filter/slice/aggregation/date_bucket referencing a Unique Count pseudo-field and asserts the MCP error surfaces (the error class never appears in that spec). A refactor that mis-assembles `violations` (wrong clause labels, wrong `matchedNames`) or drops the `if (violations.length > 0)` check would pass the current suite. One handler-level test per clause type would pin the wiring.

apps/web/src/features/data-marts/shared/utils/unique-count-availability.ts
```diff
@@ -0,0 +1,134 @@
...
+  return { scope: 'main', availability };
+}
+
+/**
+ * Whether a STORED selection may stay. Only a state the client RECOGNISES as unavailable — or the
+ * source vanishing from the schema (`undefined`, the one case pruning exists for) — may take one
+ * away. A value it cannot read must never destroy one: the user has no way back, and the backend
+ * remains the authority that rejects a bad save.
+ */
+export function canKeepUniqueCount(state: UniqueCountSourceState | undefined): boolean {
```
prokopetsv, line 81: 🟡 **Non-Blocker:** An unrecognized joined-availability value (version skew: older web bundle, newer backend failure reason) is treated as keepable — the right call, and well documented here — but the row then renders exactly like a confirmed-healthy metric: `UniqueCountRow.tsx:97-101` computes `hint = undefined`, so no warning, normal Σ badge. A selection that is actually failing server-side is visually indistinguishable from a live one, and the user only learns at save/run time. Consider a third "unverified" visual state for the unknown case, distinct from both "live" and "not emitted".

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -634,6 +637,32 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
   }
 
+  /**
+   * A joined Data Mart's Unique Count is built by the blended query builder, which requires an
+   * EXPLICIT column projection — and `['*']` is this tool's only way to say "all fields", so
+   * mapping it to null would make such a report impossible to update at all. Materialize the
+   * data mart's reportable columns instead, exactly as the web picker does when a joined
+   * source's Unique Count is toggled on while the selection is still implicit (#6792).
+   */
+  private toUpdatedColumnConfig(fields: string[], current: ReportDto): ReportColumnConfig {
```
zapolsky, line 647: **[Bug — verified]** `add_report` / `update_report` accept the advertised `<alias>__unique_count` pseudo-field in `fields`, store it as a projected column, and the report then fails on **every** run with a misleading "restore the schema" error — after the Google Sheet has already been created.

**Evidence chain:**
1. `facades/mcp-data-marts.facade.impl.ts:195-206` — the pseudo-field is pushed into the *same* array real blended fields live in: `return { joinedFields: [...blendedFields, ...uniqueCountFields], uniqueCountSources };`
2. `ee/mcp/tools/data-mart-details.tool.ts:184` — reaches the model verbatim as `joined_fields`.
3. `ee/mcp/instructions/mcp-system-instructions.ts:20` — sits in the **global** `Rules:` block (`:15`), prepended to every session by `composeMcpInstructions` (`:33-40`): *"Copy its \"name\" (e.g. \"orders__unique_count\") … It can be selected in \"fields\" and ordered by in \"sort\" … but never placed in filters, slices, aggregations, or date_buckets."* The exclusion list names four clauses and affirmatively blesses `fields`/`sort` — the literal parameter names on `add_report` (`add-report.tool.ts:49-52`, `:81-87`) and `update_report` (`update-report.tool.ts:30-36`, `:61-66`), whose descriptions read "same as query_data_mart's" and whose tool description tells the model to copy a query's parameters verbatim into an export.
4. **No split on this path.** `splitUniqueCountFields` is imported by `query-data-mart.tool.ts:27` and nowhere else. `mcp-reports.facade.impl.ts:636-638` — `toColumnConfig` returns `fields` untouched; `createReportWithConfig` (`:492-509`) and `addGoogleSheetsReport` (`:542-563`) stop the `CreateReportCommand` argument list at `dateTruncConfig`, never passing `uniqueCountConfig`; `updateReport` (`:230`) only forwards `current.uniqueCountConfig`.
5. **Nothing rejects it at save.** Pre-flight passes `uniqueCountConfig: null` (`:630`). In `services/output-controls-validator.service.ts:570-576`, `hasOutputControls` **omits `columnConfig`** from its terms, so a projection-only report takes the early return at `:578-586` after `validateOutputColumnNames` — which only detects *duplicates* (`:502-546`), never membership. The sort-carrying variant passes too: `validateSort`'s `selectedSet` is seeded from `columnConfig` (`:769`) and `knownOutputColumns` from every available source's `buildJoinedUniqueCountColumnName` (`:721-728`).
6. **Failure lands at run time.** `services/blended-report-data.service.ts:168-169` → `:624-631`: the name is neither native nor in `blendableSchema.blendedFields` (the pseudo-field exists only inside the MCP facade, never in `computeBlendableSchema`), so `throwDisconnectedReportColumnsError` fires — *"…missing from the current Data Mart output schema. Uncheck them to remove them from the report, or contact your analyst to restore the schema."* Not remapped: `output-controls-error.mapper.ts` wraps only the save calls, and this is a run-time `BusinessViolationException`.
7. **Ordering makes it irreversible.** `mcp-reports.facade.impl.ts:527` pre-flight (blind to this) → `:530-538` Google Sheet created → `:542` report saved.

**Why anchored here:** `toUpdatedColumnConfig` is the proof this path *knows* joined Unique Counts exist — a dedicated helper plus a dedicated error for `fields: ["*"]` on a report that already carries one. But it can only **preserve** `current.uniqueCountConfig`. There is no path to **set** it from `fields`, and no rejection either, so the metric is unreachable through MCP report creation while being actively advertised to the model.

**Trigger:** *"create a Google Sheets report of unique orders per customer"* → `add_report({ fields: ['customer_id','orders__unique_count'], … })`. Sheet created, report saves clean, every manual and scheduled run fails afterwards.

**Fix:** minimally — reject `<alias>__unique_count` names matching an advertised source in `McpReportsFacadeImpl` with a named `BadRequestException` **before** `createGoogleSheetDocumentService.run`. Full support — reuse the `query-data-mart.tool.ts:228-259` block (`hasUniqueCountFieldCandidate` → `splitUniqueCountFields` → `findUniqueCountClauseViolations`), pass the result as `CreateReportCommand`'s `uniqueCountConfig`, and replace the hard-coded `uniqueCountConfig: null` at `:630` so the pre-flight can fail before the sheet exists.

apps/web/src/features/data-marts/reports/edit/hooks/useGoogleSheetsReportForm.ts
```diff
@@ -84,7 +84,7 @@ export function useGoogleSheetsReportForm({
       limitConfig: initialReport?.limitConfig ?? null,
       aggregationConfig: initialReport?.aggregationConfig ?? null,
       dateTruncConfig: initialReport?.dateTruncConfig ?? null,
-      uniqueCountConfig: initialReport?.uniqueCountConfig ?? false,
+      uniqueCountConfig: initialReport?.uniqueCountConfig ?? [],
```
zapolsky, line 87: **[Contract regression — verified]** The wire value for "Unique Count is off" changes from `false` to `[]` for every report saved through the new SPA. `[]` is **truthy**, and the *currently released* Google Sheets add-on (v0.15.1, `fcb6ba4`) reads the field as a boolean — so it renders a ticked "Unique count" row, with the Σ "auto-generated column" badge, on reports that emit no such column.

**Evidence chain:**
1. This PR widens the persisted union — `dto/schemas/unique-count-config.schema.ts:5`; `origin/main` had `z.boolean().nullable()`.
2. **This line** plus `:36` (`z.array(z.string())`) replace the baseline's `z.boolean()` / `?? false`, and the value is submitted **unconditionally, with no dirty-gating**, at `:131` (create) and `:155` (update). Identical in `useEmailReportForm.ts` and `useLookerStudioReportForm.ts`. The send-side DTOs no longer admit a boolean at all (`create-report.request.dto.ts:24`, `update-report.request.dto.ts:88`).
3. The backend cannot fold it: `use-cases/create-report.service.ts:113,131` and `update-report.service.ts:122` use `command.uniqueCountConfig ?? null`, and `[]` is not nullish. `entities/report.entity.ts:112-118` → `common/zod/zod-transformer.ts` short-circuits only `null`/`undefined`. The read mapper `mappers/report.mapper.ts:127` sends `[]` straight back out.
4. Released add-on reads it as boolean: `fcb6ba4:src/shared/types/output-config.ts:178` → `uniqueCountConfig: boolean;`; `fcb6ba4:src/features/reports/hooks/use-report-card.ts:87` → `report.uniqueCountConfig ?? false` (a no-op on `[]`); `fcb6ba4:.../use-output-controls.ts:90-99` spreads it into `effectiveOutputConfig` normalizing only filter/sort/limit. It reads SPA-saved reports over the same API (`fcb6ba4:src/features/reports/api/report-service.ts:7,10,17`).
5. Two live consumers misread it: `fcb6ba4:.../ReportColumnPicker/index.tsx:970-981` passes the raw value to `<UniqueCountRow checked={…}>` → `:1144-1146` `<input type="checkbox" checked={checked} …>` (WebIDL ToBoolean → ticked) and `:1152` `{checked && <Σ badge>}`; and `:540-551`, a repair effect on a keyless mart, sees `[]` as truthy and commits `uniqueCountConfig: false`, dirtying a report nobody edited.
6. The window does not close on its own: `fcb6ba4` is the newest release commit, the array-aware fix (`ea02b13`) is unmerged on `feature/6792-joined-unique-count`, and the add-on's release workflow is `on: workflow_dispatch` only.

**Honest scoping** (an adversarial pass trimmed the original claim): a third named consumer, `hasAnyOutputControls` (`fcb6ba4:output-config.ts:198`), is **dead code** — its only call site is gated on `!outputControlsSupported`, and the add-on's supported-storage set is byte-identical to the backend's full `DataStorageType` enum. Nothing is destroyed and nothing 400s; the add-on's save path round-trips `[]` into a backend that now accepts it, the mislabel self-corrects the moment anyone toggles the row, and only reports touched by the new SPA are affected (MCP-created ones stay `null`, old-SPA ones stay `false`).

What keeps it in this bucket rather than the notes: it is a **false statement about report content shown to Marketplace users**, it is persistent, and the wire change buys nothing — `[]` and `false` are indistinguishable to every backend consumer (`normalizeUniqueCountSources` maps `false`, `null`, `undefined` and `[]` all to `[]`) and to this PR's own web (`report.mapper.ts:44-49` maps `false`/`null`/absent → `[]`).

**Fix:** fold the empty array at the **persist boundary**, not in the web layer — any API client can write `[]`. And apply the fold **before** the change comparison at `update-report.service.ts:121-124`: folding only at assignment would make a stored `null` vs an incoming `[]` compare as changed, flipping `rejectUnavailableUniqueCountSources` to `true` on an untouched config — exactly the permanent-400-on-rename the comment at `:138-141` exists to prevent. No test pins the current behaviour: the e2e round-trip block covers `['', 'users']`, `true` and `null` only.

zapolsky (commented):
🔴 **Blocked issues** — see inline comments. These should be addressed before merge, but this review does not formally block; a human owns the merge decision. Both findings survived an independent adversarial verification pass (a third candidate, rollback-poisoning of the widened persisted union, was **refuted** as a blocker and demoted to a note). CI is 20/20 green on this head. Posted by automated OWOX review.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: **Correction — this one appears to be already covered.** Re-verified against the checked-out head; the throw site *is* exercised end to end.

- `query-data-mart.tool.spec.ts:746-777` drives the real `tool.handler` with `aggregations: [{ field: 'orders__unique_count', function: 'SUM' }]` and asserts `error_code: 'unique_count_selection_only'`, the offending clause name in the message, and `facade.queryDataMart` not called.
- `:779-808` is an `it.each` over `filters`, `slices` and `date_buckets`, same handler, same assertions — so all four clause labels and their `matchedNames` are pinned.
- That error code has exactly one producer: `query-data-mart.tool.ts:449-450`, inside `if (err instanceof UniqueCountFieldUnsupportedClauseError)`. And that class has exactly one throw site: `:257`. So asserting the code *is* asserting this line ran.

The observation that the class name never appears in the spec is accurate — the tests bind to its `error_code` instead, which is the stronger of the two (it pins the wire contract the model actually sees, not the class identity). A refactor that mis-assembles `violations` or drops the `if (violations.length > 0)` check would fail `:768`/`:805`-`:806`, which assert the field name and the clause label inside the message.

<sub>Posted by the automated OWOX review — flagging because a stale "missing coverage" note costs more than silence.</sub>

zapolsky:

Two 🔴 blocked findings are posted inline. CI is **20/20 green** on this head (ESLint, Prettier, Markdownlint, every workspace suite, 4 API E2E shards, Browser E2E, Build, Audit) — no pre-existing red to attribute either.

**Primary notes (≥ 80), highest first — none of these block merge.**

- `apps/backend/src/data-marts/services/output-controls-validator.service.ts:660` — **the paired Sheets add-on collapses `columnConfig` back to `null` and hits this PR's new `JOINED_UNIQUE_COUNT_REQUIRES_COLUMN_CONFIG` on the feature's headline flow** *(cross-repo; the fix lives in the paired Sheets add-on repo, not here)*. The add-on materializes for the metric exactly like `apps/web` does (`ReportColumnPicker/index.tsx:488-490`, and its `needsExplicitColumnConfig` at `output-controls-utils.ts:142` *does* know about joined sources) — but neither of its save paths uses that helper: `use-report-card.ts:305-308` and `CreateReportForm.tsx:126-130` still compute `requireExplicit: aggregationActive || hasPreJoinSlice`, so `normalizeColumnConfig` (`column-config-utils.ts:88-95`) recognises the materialized list as the default and returns `null`. Payload goes out `columnConfig: null, uniqueCountConfig: ['orders']` → 400, with no way out through the add-on UI except deselecting an unrelated column. One-line fix there: pass `requireExplicit: needsExplicitColumnConfig(draftOutputConfig)`.
- `apps/backend/src/data-marts/dto/presentation/create-report-request-api.dto.ts:130` — **the unreleased add-on sends an array on every save, which any pre-#1522 backend rejects.** This PR replaces `@IsBoolean()` with `@IsZodValid(UniqueCountConfigRequestSchema, …)`, so every backend released before it still carries the boolean-only validator; the add-on posts `uniqueCountConfig: []` unconditionally (`use-report-card.ts:319`, `CreateReportForm.tsx:141`, seeded from `EMPTY_OUTPUT_CONFIG`). A self-hosted user on last month's backend whose add-on auto-updates gets `400 uniqueCountConfig must be a boolean value` on a title-only rename, and cannot roll the add-on back. Worth coordinating the two releases explicitly, or having the add-on omit the field when the list is empty. *(The read direction degrades cleanly — `readJoinedUniqueCountState(undefined) → 'unknown'`, `canOfferUniqueCount` false, stored selections preserved. That half is correct on both sides.)*

<details>

**50–79**

- `apps/backend/src/data-marts/services/blended-report-data.service.ts:351` — a joined source whose Output Alias is blank or whitespace renders its Unique Count header as the bare string `Unique Count`, identical to the main mart's. `defaultAlias` is `??`-coalesced only (`blendable-schema.service.ts:239,245`), and `formatBlendedFieldDisplayName` documents that `"  "` survives (`blended-field-display-name.ts:41-46`). The save-time gate compares `name`s only (`output-controls-validator.service.ts:503-517`), so `Unique Count` vs `orders__unique_count` never collide there — Google Sheets then fails at run time in `findDuplicateRenderedLabels`, and Looker Studio silently shows two identically-labelled fields. The picker already does the right thing (`ReportColumnPicker.tsx:1079`, `joined.defaultAlias.trim() || joined.title`); mirroring that in `resolveUniqueCountSources` would close it.
- `apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts:349` — a joined source contributing *only* a Unique Count still pays for a full dedup CTE and an outer `LEFT JOIN` that nothing reads. Its `chainBlendedFields` is empty (a UC references its source by alias path, not by column), yet `blend-cte.builder.ts:167` emits `buildAggregationCte` unconditionally and `buildJoinParts` joins every root; the sleeve reads `<cte>_raw` directly (`metric-sleeve.builder.ts:1015`) and the outer `agg` reads no column of that source. Emitted a second time inside `_kept_groups` for a Totals report (`:159-160`). Same class of waste `withoutChainsOnlyADroppedUniqueCountNeeded` was written to avoid — it just prunes a *dropped* source's chain, not a *surviving* one's unused dedup layer. Pure cost on a bytes-scanned warehouse; result is unaffected.
- `apps/backend/src/data-marts/services/report-data-cache.service.ts:132` — the cache lookup is keyed on `report.id` + `expiresAt` alone; `ReportDataCache` carries no config hash or SQL fingerprint, invalidation fires only on a config/relationship edit, and this PR ships no migration touching the table. So an unedited report keeps serving pre-fix numbers after deploy — and on a hit for BigQuery the new SQL is never even executed, because `initFromState` re-attaches the old result table (`bigquery-report-reader.service.ts:314-317`) while `readReportDataBatch` skips `initializeReportData` (`:111-113`). `getCacheLifetime` has no upper bound (`z.number().min(60)`). Given the changeset announces three classes of number change, purging `report_data_cache` at release would be worth putting in the runbook.
- `apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx:1313` — *(cross-repo)* this file deliberately attaches the Unique Count row to the **source** rather than to a surviving field group, so an **excluded** source still gets a clearable row (`isEmitted: false`, with the `NOT_EMITTED_NOTE` explaining it). The add-on builds groups only from included sources (`ReportColumnPicker/index.tsx:185-187`) and has no `isEmitted` concept at all, while its own pruner deliberately *keeps* excluded sources — so there the config entry survives every save, the column never appears, and no control exists to remove it. Follow-up in the other repo.

**0–49**

- `apps/backend/src/data-marts/data-storage-types/utils/primary-key-identity.utils.ts:99` — a joined `Unique Count` reports **0** for rows with an empty key component (`renderPrimaryKeyCountedSlotRef`), while `Sum`/`Average` on the same source now **include** them via the `__owox_rid` surrogate (`:126-135`, projected for every identity owner at `blend-cte.builder.ts:149-151`). One report can therefore read "0 orders, worth 2000". Deliberate and consistent with `COUNT(DISTINCT …)` semantics — but the changeset documents the count going down for the **main** mart only; one sentence beside the Sum/Average bullet would cover the joined case too.
- `apps/backend/src/data-marts/services/blended-report-data.service.ts:339` — excluding a joined source from reporting silently drops its Unique Count column, while every *other* field of that same source keeps working (the PR's own comment at `:365-370` says so). The file has no `Logger` at all, so this drop, the stale-sort prune (`:262-270`) and the COUNT-beside-COUNT_DISTINCT drop (`:417-439`) are all silent — a nightly Sheet loses a column and every `VLOOKUP` to its right shifts, with no run warning.
- `apps/backend/src/data-marts/entities/report.entity.ts:116` — the entity re-parses `UniqueCountConfigSchema` on **read** and `createZodTransformer.from()` throws rather than degrading, so a rollback past this PR cannot hydrate a row holding an array. **Refuted as a blocker**: this is the tenth instance of a year-old repo-wide pattern (20 `createZodTransformer` columns, zero `.catch(...)`; nine prior breaking-on-rollback widenings across four columns and eight PRs, incl. `filterConfig` twice — a sibling column on this same entity). Worth a central `createZodTransformer` fix in its own PR, not a gate on this one. Note the two newest data-quality schemas use `.strict()`, which raises the future cost.
- `apps/backend/src/data-marts/data-storage-types/blending/metric-sleeve.builder.ts:1014` — `buildCountingSleeveCte` re-resolves the owner chain already resolved at `:901`, discarding the result (a bare statement kept only for its throw); and `groupByParts` (`:1032`) is dead in the nested branch, which returns at `:1083`.
- `apps/backend/src/data-marts/services/blended-report-data.service.ts:291` — the "scope-tagged vocabularies so the compiler rejects a swap" note holds for the availability *verdicts* (`apps/web/.../unique-count-availability.ts:36-49` is a real discriminated union on `scope`), but not for the two *key column lists*: `getReportablePrimaryKeyFields` (prunes hidden, keeps nested) and `collectPrimaryKeyRowIdentity` (keeps hidden, rejects nested) both return bare `string[]` and meet as sibling properties here, and again at `abstract-blended-query-builder.ts:267` vs `:335`. Branding them would make the swap fail to compile.
- `apps/backend/src/data-marts/data-storage-types/blending/metric-sleeve.builder.ts:729` — `buildSleeveCte`'s returned `alias` is now dead (both `buildAll` call sites read `built.pulls`); it is a weaker second copy of `pulls[0].alias` a future caller could desynchronise.
- `apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts:650` — `buildRowSurrogate`'s doc comment still says the window is *"emitted only for an owner whose joined Data Mart declares NO usable primary key"*, but `blend-cte.builder.ts:149-151` now emits it for every value-sleeve owner. The behaviour change is intended and disclosed in the changeset; only the comment is stale — and it is the comment a future reader will use to reason about the cost.
- `apps/backend/src/data-marts/data-storage-types/blending/metric-sleeve.builder.ts:517` — adding `_oid` to the primary-key branch's reserved set newly 400s a composite-key report whose dimension is literally named `_oid` (and frees `_oid_0`). The guard's own comment (`:581-585`) argues the collision cannot occur now that dimensions are positionally aliased, so this is a false positive on a name that previously worked.
- `apps/web/src/features/data-marts/shared/types/relationship.types.ts:196` — `uniqueCountAvailability` is typed required while the sibling `uniqueCountKeyFields?` added by the same PR is optional with a comment explaining why a cached response may omit it. The single read site defends itself via `readJoinedUniqueCountState`, so this is a type lie rather than a crash — but it invites a second consumer to read the field directly.
- `apps/backend/MODULAR_CONVENTIONS.md:70` — §3 says facade interfaces live in `shared/facades/`, but only two modules do that (`common/email`, `common/scheduler`); 30 of 47 facade files put the interface in `facades/`, and `data-marts` has no `shared/` directory at all. This PR follows the dominant precedent correctly — the doc is what's wrong, and it will keep generating false review findings on every facade-touching PR.
- `apps/backend/src/data-marts/use-cases/update-report.service.ts:122` — a legacy boolean write against a stored array replaces rather than merges, so a client on the old contract (a stale tab, or a GET→PUT integration) can wipe joined selections silently. **Rescored 50 → 18** by the verification pass: web and backend ship as one version-locked artifact (`apps/owox` depends on both, served by the same Express app), the replace-semantics hazard is pre-existing and repo-wide (`origin/main` already did `dto.uniqueCountConfig ?? null`, and `columnConfig`/`filterConfig`/… share it), what is lost is a re-tickable setting rather than data, and it is visible on the next run. The claim that "MCP was given a preservation rule and REST was not" is **wrong** — `mcp-reports.facade.impl.ts:227` reads `current.uniqueCountConfig` in `origin/main` already, and only because `update_report` has no such parameter.
- `apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx:1058` — *(cross-repo)* this file's `taken` set guards the two-sources-one-SQL-name case (`a.b` and a top-level `a_b` both build `a_b__unique_count`); the add-on's `buildUniqueCountSortColumns` has no such set and would render two identically-keyed picker entries. Rendering quality only — the backend rejects the save either way.
- `apps/backend/test/output-controls.e2e-spec.ts` — the `uniqueCountConfig persistence round trip` block shares one report across three order-dependent tests (GET `['', 'users']` → PUT `true` → PUT `null`). Fine under Jest's declaration order; `--testNamePattern` or any future randomization breaks the first one.
- minor: the picker's auto-repair writes with `shouldDirty: false` (`apply-output-config-change.ts:62-77`), so a pruned config either never persists or rides along invisibly with an unrelated title edit; and `docs/getting-started/setup-guide/report-aggregations.md` carries no note that each ticked source adds a scanned `SELECT DISTINCT` CTE, which users choosing this in the picker have no way to know.

</details>

- **Correction:** the earlier note on `apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts:250` (throw site "never exercised end-to-end") is not accurate — `query-data-mart.tool.spec.ts:746-808` drives the real handler across all four clause types and asserts the `error_code` that only that branch produces. Details posted in that thread.

<sub>Automated OWOX review · worktree multi-agent passes · all blockers adversarially verified · nothing executed locally (CI read via `gh pr checks`).</sub>

laskevych (approved):
Ran a deep multi-angle review of the full diff: the blending SQL core (sleeve planner/builder, CTE assembly, join-back NULL/NaN handling, netstring composite-key encoding) across all five dialects, backend validation/services/use-cases, the MCP layer, the web picker, and the test suites. **No blocking issues** — the two-name contract, the validation chokepoints, cache invalidation wiring, and the per-dialect overrides all check out, and the fan-out fix is verifiably the mechanism the changeset describes. Approving; non-blocking comments follow inline (a few edge-case diagnostics, two warehouse-cost follow-up candidates, and some dedup suggestions).

laskevych (commented):
Inline follow-ups to the approval above — all non-blocking.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
```diff
@@ -150,11 +167,48 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
...
+      const uniqueCountFields: McpJoinedFieldDto[] = [];
+      // `orders__unique_count` is byte-identical to the unified name of a real flat field called
+      // `unique_count` on the `orders` source (and to a native column of that literal name).
+      const realFieldNames = new Set([...nativeFieldNames, ...blendedFields.map(f => f.name)]);
+      const eligibleSources = accessibleSources.filter(
+        s =>
+          s.uniqueCountAvailability === 'available' &&
+          // The real field wins: shadowing it would answer a request for that column with
+          // COUNT(DISTINCT pk), and selecting both would collide on one output alias anyway.
+          !realFieldNames.has(buildJoinedUniqueCountColumnName(s.aliasPath))
```
laskevych, line 184: 🟡 **Important (non-blocking):** The MCP surface is the one place where the documented `a.b` / `a_b` name collision resolves *silently to the wrong data*. The eligibility filter here only guards against real fields (`realFieldNames.has(...)`), so two sources whose alias paths both build `a_b__unique_count` are advertised under one name; `splitUniqueCountFields` then resolves it last-wins via `new Map(availableSources.map(s => [s.name, s.aliasPath]))` (query-data-mart.input.ts:514), and since only one aliasPath survives into `uniqueCountConfig`, the validator's `OUTPUT_COLUMN_NAME_COLLISION` (which needs both present) never fires — the query returns `COUNT(DISTINCT pk)` of the wrong mart with no error. The web picker resolves the same collision first-wins and REST rejects the pair, so the three surfaces disagree. Suggest skipping the second colliding source while advertising (mirroring the picker's `taken` set) or rejecting the ambiguous name in `splitUniqueCountFields`.

apps/backend/src/data-marts/services/blended-report-data.service.ts
```diff
@@ -249,11 +304,111 @@ export class BlendedReportDataService {
...
+    sortConfig: SortRule[],
+    stalePaths: string[],
+    dataMart: DataMart,
+    blendedFieldsByName: ReadonlyMap<string, BlendedFieldDto>
+  ): SortRule[] {
+    const nativeNames = new Set(collectSchemaFieldPaths(dataMart.schema?.fields ?? []));
+    const staleColumns = new Set(
+      stalePaths
+        .map(buildJoinedUniqueCountColumnName)
+        .filter(name => !blendedFieldsByName.has(name) && !nativeNames.has(name))
```
laskevych: 🟡 **Important (non-blocking):** The "a real field owns the `<alias>__unique_count` name" tie-break is re-derived in five places with diverging field universes — and the hidden-field treatment already disagrees observably: this run-path pruner checks `blendedFieldsByName`, built from *all* blended fields including hidden ones (line 168), while the picker's `knownFieldNames` (ReportColumnPicker.tsx:669) excludes hidden fields. So for a stored sort on `orders__unique_count` where the real blended field with that name is *hidden*, a scheduled run keeps rendering `ORDER BY` the hidden field, but merely opening the picker classifies the same rule as a stranded metric sort and deletes it on the next save. Consider one shared ownership resolver (or at minimum aligning hidden-field treatment between this pruner and the picker's repair effect).

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -652,18 +718,41 @@ export class OutputControlsValidatorService {
...
+        for (const name of uniqueCountOutputColumns) knownOutputColumns.add(name);
+
+        // A filter naming one of those columns is rejected HERE, and its rule is kept out of
+        // validateFilters below: unknown to the field index, it would otherwise become
+        // FILTER_COLUMN_UNKNOWN and route to throwDisconnectedReportColumnsError, which tells the
+        // caller to repair a schema link that is not broken. Same honesty the MCP tool's
+        // UniqueCountFieldUnsupportedClauseError already gives.
+        const isUniqueCountFilter = (rule: FilterRule) =>
+          uniqueCountOutputColumns.has(rule.column) && !homeFieldTypes.has(rule.column);
```
laskevych: 🟡 **Important (non-blocking):** A filter on a *since-hidden or deleted* real field named `orders__unique_count` now gets misdiagnosed. `homeFieldTypes` skips hidden fields (line 709) while `uniqueCountOutputColumns` includes every available source regardless of the report's `uniqueCountConfig` (lines 725–726), so this check matches and the caller is told the column "is a Unique Count metric… Remove the filter" — factually wrong for that state (it's a real field that went hidden, and the report may have no Unique Count enabled at all). On main this state produced `FILTER_COLUMN_UNKNOWN` → the disconnected-columns error naming the broken schema link and the repair. Suggest preferring the disconnected diagnosis when the blended schema still has a real (hidden included) field with that name.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
```diff
@@ -141,18 +142,23 @@ export class QueryDataMartService {
...
     const queryMetadata = {
+      // `fields` no longer names the Unique Count the caller asked for — the tool splits that
+      // pseudo-field out into `uniqueCountConfig` before calling. Journal both, or Run History
+      // shows a run whose extra column came from nowhere.
       fields: r.fields,
       ...(r.filterConfig ? { filters: r.filterConfig } : {}),
       ...(r.sortConfig ? { sort: r.sortConfig } : {}),
       ...(r.aggregationConfig ? { aggregations: r.aggregationConfig } : {}),
       ...(r.dateTruncConfig ? { dateBuckets: r.dateTruncConfig } : {}),
+      ...(r.uniqueCountConfig?.length ? { uniqueCountConfig: r.uniqueCountConfig } : {}),
```
laskevych, line 161: 🟡 **Important (non-blocking):** This journaled `uniqueCountConfig` never reaches storage: `McpQueryRunMetadataSchema.query` (mcp-query-run-metadata.schema.ts:11–20) was not extended with the key, zod's non-strict `z.object` strips unknown keys, and `recordMcpQueryRun` (data-mart-run.service.ts:600–609) persists the *parsed* result. Since the pseudo-field is also split out of `fields`, the persisted run shows neither — exactly the outcome the comment above this line says it exists to prevent. Fix is one line in the schema plus a "(survives schema parse)" test like the existing sort one.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
```diff
@@ -322,14 +329,19 @@ export abstract class AbstractBlendedQueryBuilder implements BlendedQueryBuilder
...
+    // A joined Unique Count counts `<cte>_raw.<key>`, and a declared key is frequently referenced
+    // by nothing else — so it rides the same projection path the value sleeve's identity uses.
+    const uniqueCountKeyColumns = new Map<string, readonly string[]>(
+      uniqueCountSources.map(s => [s.cteName, s.pkColumns])
+    );
 
     for (const root of roots) {
-      const { ctes, params } = cteBuilder.buildSubtreeCtes(
-        root,
+      const { ctes, params } = cteBuilder.buildSubtreeCtes(root, {
```
laskevych, line 339: ⚡ **Performance (non-blocking, follow-up candidate):** A source referenced *only* by its Unique Count still emits its dedup CTE here (`SELECT <join keys> FROM x_raw GROUP BY …` — with empty `blendedFields` it projects join keys only) and `buildJoinParts` LEFT JOINs it into the outer body, although no output expression ever qualifies to it — the sleeve joins `x_raw` independently. That's an extra billed aggregation scan of the source on every run (and twice when Totals runs, via `_kept_groups`). The join is 1:≤1 on the grouped keys, so skipping the dedup leg for chains with no referenced columns and no pre-join filter looks grain-safe.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -214,17 +221,55 @@ If truncated is true, not all matching rows were returned: narrow the query (few
       const dateTruncConfig = mapMcpDateBuckets(parsed.date_buckets);
       const sortConfig = mapMcpSort(parsed.sort);
 
+      // Only a data mart that joins others carries a Unique Count pseudo-field, and most
+      // requests select none — skip the extra schema lookup unless a field could be one.
+      let fields = parsed.fields;
+      let uniqueCountConfig: string[] | undefined;
+      if (hasUniqueCountFieldCandidate(parsed.fields)) {
+        const details = await this.dataMarts.getDataMartDetails({
```
laskevych, line 229: ⚡ **Performance (non-blocking, follow-up candidate):** A pseudo-field query now resolves the blendable schema at least three times in one request: `getDataMartDetails` here (access check + `actualizeSchemaIfExpired`, potentially a live warehouse call, + `computeBlendableSchema`), then `queryDataMart` recomputes it in both the validator and the blending-decision path (blended-report-data.service.ts:153–159) because nothing passes the schema through. The `precomputedBlendableSchema` plumbing already exists but is wired only for the totals path — routing the pseudo-field split through the query pipeline (or passing the schema along) would do the heavy resolution once.

apps/backend/src/data-marts/use-cases/update-report.service.ts
```diff
@@ -118,6 +118,11 @@ export class UpdateReportService {
       }
     }
 
+    const previousUniqueCountConfig = report.uniqueCountConfig ?? null;
+    const nextUniqueCountConfig = command.uniqueCountConfig ?? null;
+    const uniqueCountChanged =
```
laskevych, line 130: 💡 **Suggestion:** Normalize both sides before this `JSON.stringify` comparison. A legacy report storing `true` (or `null`) is loaded by the web mapper as `['']` (or `[]`) and re-posted unchanged on any save, so the first no-op save — e.g. a pure title rename — reads as a config change and invalidates the report data cache, forcing a fresh billed warehouse query for a semantically identical config. One-time per legacy report (the stored value is rewritten on that save), but free to avoid by comparing normalized forms.

apps/backend/src/data-marts/services/blended-report-data.service.ts
```diff
@@ -249,11 +304,111 @@ export class BlendedReportDataService {
...
+   * its CTE built and LEFT JOINed — a paid warehouse scan feeding a column nobody reads.
+   *
+   * Deliberately narrow, and it must stay that way. `buildRelationshipChains` is NOT filtered by
+   * `isIncluded`, and that is load-bearing: an excluded source's fields remain selectable,
+   * filterable and sortable, and an ordinary `ORDER BY orders__amount` on one resolves precisely
+   * BECAUSE the join is built unconditionally. So this recomputes what the chain set would have
+   * been had `uniqueCountConfig` never seeded it, then adds back the SURVIVING sources (and their
+   * ancestors, which a nested one still needs). A chain goes only when nothing else wanted it.
+   */
+  private withoutChainsOnlyADroppedUniqueCountNeeded(
```
laskevych, line 395: 💡 **Suggestion:** `withoutChainsOnlyADroppedUniqueCountNeeded` can be deleted by filtering `uniqueCountAliasPaths` to emittable sources *before* seeding the chain build. Every drop predicate in `resolveUniqueCountSources` (source missing, `isIncluded === false`, empty `pkColumns`) is knowable pre-seed — `availableSources` comes from the blendable schema computed earlier, and the access assertion throws before chains are built — leaving only the cheap `!chain` guard. That removes the mirror-coupling this method's own comment warns about ("Mirrors buildRelationshipChains' own `requestedBlendedFields` — the two must not drift").

apps/backend/src/data-marts/services/blended-report-data.service.ts
```diff
@@ -249,11 +304,111 @@ export class BlendedReportDataService {
...
+    availableSources: AvailableSourceDto[],
+    nameStyle: BlendedFieldNameStyle
+  ): JoinedUniqueCountSource[] {
+    const chainByCteName = new Map(chains.map(c => [c.cteName, c]));
+    const sourceByAliasPath = new Map(availableSources.map(s => [s.aliasPath, s]));
+
+    return [...uniqueCountAliasPaths]
+      .map(aliasPath => {
+        // `cteName` is the alias path with dots replaced by underscores — see buildRelationshipChains.
+        const chain = chainByCteName.get(aliasPath.split('.').join('_'));
```
laskevych: 💡 **Suggestion:** These two inline `aliasPath.split('.').join('_')` sites (here and line 385) re-derive the aliasPath→cteName mapping that `aliasPathToCteName` (filter-config.schema.ts:142) already owns and `blended-field-index.ts:46` already calls. A lookup miss here silently drops the metric from SQL and headers by design (lines 339–341), so keeping the mapping in one place is worth it even with the integration specs pinning today's agreement.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
```diff
@@ -93,8 +96,69 @@ export function collectPrimaryKeyRowIdentity(fields: readonly DataMartSchemaFiel
...
+export type JoinedUniqueCountAvailability =
+  (typeof JOINED_UNIQUE_COUNT_AVAILABILITY_VALUES)[number];
+
+// Every isPrimaryKey field anywhere in the RAW tree, with the two facts that can disqualify it: its
+// dotted path (nested when it holds a dot) and whether it AND every ancestor are still connected.
+// Status is not pruned — unlike collectSchemaFieldPathDescriptors, which drops those subtrees and so
+// would make a declared-but-unreachable key silently disappear (reported as "no key" instead of
+// "key exists but broken"). Reachability is computed exactly as collectPrimaryKeyRowIdentity walks
+// it, so the diagnosis can never disagree with the predicate it explains.
+function collectDeclaredPrimaryKeys(
```
laskevych, line 144: 💡 **Suggestion:** `collectDeclaredPrimaryKeys` repeats `collectPrimaryKeyRowIdentity`'s walk verbatim (same prefix build, reachability, recursion guard), and `classifyJoinedUniqueCountAvailability` runs both over the same tree. The row identity is exactly derivable from the declared keys (`declared.every(d => d.reachable && !d.path.includes('.')) ? declared.map(d => d.path) : []`), byte-identical for all callers — one traversal would make the picker's diagnosis and the query path's emit predicate structurally inseparable rather than fixture-pinned.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -377,7 +396,19 @@ export class ReportSqlComposerService {
...
-    const projectedExplicit = report.columnConfig && report.columnConfig.length > 0;
+    // An EMPTY projection is "the caller selected no dimensions" (a Unique-Count-only MCP
+    // request), NOT "project everything" — totalling every numeric column would bill a second
+    // warehouse query for numbers nobody asked for. That reading holds only for a METRICS-ONLY
+    // plan, exactly as in resolveReportDataHeaders: `[]` is also what PERSISTED legacy rows carry
+    // (report-column-config.schema.ts). Such a row predates both aggregations and Unique Count, so
+    // it is NOT metrics-only and keeps projecting every native column — a report that has been
+    // totalling those for months does not lose its Totals to this. A metrics-only `[]` emits no
+    // dimension columns at all, so an empty Totals block there is the report's own shape.
+    const metricsOnly =
```
laskevych: 💡 **Suggestion (minor):** This metrics-only predicate is textually identical to the validator's `isMetricsOnly` (output-controls-validator.service.ts:859–861), synced only by "Mirror resolveReportDataHeaders" comments; `report-like-read-plan.ts` (already home to `hasOutputControls`) is a natural place for a shared helper covering these two. The headers variant is deliberately different (PK-gated main Unique Count) and should stay separate.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -1353,54 +1675,54 @@ export function ReportColumnPicker({
...
+              filterableTypeFor={filterableTypeFor}
+              filtersByColumn={filtersByColumn}
+              onAddFilter={outputControlsAvailable ? handleAddFilter : undefined}
+              onRemoveFilterAt={outputControlsAvailable ? handleRemoveFilterAt : undefined}
+              onReplaceFilterAt={outputControlsAvailable ? handleReplaceFilterAt : undefined}
+              preJoinByAliasPathColumn={preJoinByAliasPathColumn}
+              aggregationByColumn={aggregationByColumn}
+              onApplyAggregation={outputControlsAvailable ? handleApplyAggregation : undefined}
+              hasSearchQuery={hasSearchQuery}
+              uniqueCount={
```
laskevych, line 1707: 🧹 **Nit:** This copies `group.uniqueCount`'s five identically-named fields one-by-one into the structurally identical local `GroupUniqueCount`; `{ ...group.uniqueCount, state, onCheckedChange }` with the prop typed as `NonNullable<BlendedGroup['uniqueCount']> & { state; onCheckedChange }` drops both the duplicate shape and the copy. As is, an optional sixth field forgotten here still compiles and silently renders without it — and `BlendedGroupItem` already reads `group.uniqueCount?.checked` separately for its initial-open state, so there are two channels that can drift.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
zapolsky: Fixed in d6a476c73.

Two codes now mirror `UNIQUE_COUNT_FILTER_UNSUPPORTED`: `UNIQUE_COUNT_AGGREGATION_UNSUPPORTED` and `UNIQUE_COUNT_DATE_TRUNC_UNSUPPORTED`. They are raised from the same `isUniqueCountColumn` predicate the filter refusal uses (so a real field owning the name is still that field), and the offending rules are excluded from `validateAggregations` / `validateDateTruncs` afterwards — otherwise the caller would get the honest code *and* the misleading `type: 'unknown'` one in the same response.

`output-controls-error.mapper.ts` maps both into a `unique_count_selection_only` section, so the MCP caller gets the same guidance the filter path already gave it.

`output-controls-validator.service.spec.ts:4191` pinned the old behaviour (`AGGREGATION_COLUMN_NOT_SELECTED`) and was inverted: the test now asserts the dedicated code is present and the misleading one is absent, for both the aggregation and the date-bucket case. Mapper coverage added in `output-controls-error.mapper.spec.ts`.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Agreed with the correction below — re-checked, and `query-data-mart.tool.spec.ts:746-808` does drive the real `handler` across all four clause types, asserting the `error_code` that only this branch produces. No coverage change needed here.

The neighbouring gap in the same area *was* real and is fixed in d6a476c73: passing the human-readable `Orders Unique Count` in `fields` used to miss `hasUniqueCountFieldCandidate` entirely and land on the generic `field_not_found`. It is now recognised as a candidate shape, so it reaches `UnmatchedUniqueCountFieldError`, which names the exact `orders__unique_count` to copy. A real field owning such a name still wins — `realFieldNames` is checked first (`query-data-mart.input.spec.ts` pins both directions).

apps/web/src/features/data-marts/shared/utils/unique-count-availability.ts
zapolsky: Fixed in d6a476c73 (and mirrored in the Sheets add-on).

A checked row on an `unknown` verdict is now a third state, distinct from both "live" and "not emitted":

- the Σ badge is muted instead of blue, and its tooltip says the column is unconfirmed;
- the row carries a note through the tooltip *and* `aria-describedby`, so it reaches a screen reader too: the app cannot confirm the Data Mart can still be counted, the selection is kept, reload and escalate if it persists;
- `not emitted` still wins where both apply — an excluded source has a concrete reason, and that is the more useful one.

The row stays clearable and re-tickable: the verdict is unknown, not refused, so nothing about it may become inert. Covered in `UniqueCountRow.test.tsx` (and the add-on's `UniqueCountRow.test.ts`), including the negative case — a verdict the client *can* read still renders silently.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
zapolsky: Confirmed, and fixed in d6a476c73 — by **rejecting** the pseudo-field rather than supporting it, and at the validator chokepoint rather than in this facade.

New `UNIQUE_COUNT_COLUMN_NOT_PROJECTABLE`: a Unique Count output column named in `columnConfig` is refused whenever no real field owns that name. The important half is the gate — `hasOutputControls` / `needsSchema` gained a term for a projection-only report whose column list contains such a name, because without it the early return you identified at `:578-586` skipped the check entirely, which is exactly why nothing rejected it at save. It is deliberately *not* part of the capability decision: on a storage without output controls there are no joined sources, so a column of that shape is an ordinary field name.

That places the refusal inside `assertCanCreateReport`, i.e. **before** `createGoogleSheetDocumentService.run` — no orphaned sheet, and no report that fails every subsequent run. `output-controls-error.mapper.ts` translates it to `unique_count_not_reportable`, which tells the model to drop the field, tells it a human turns the metric on in the OWOX UI, and points it back at `query_data_mart` where the field does work. The global MCP rule was also narrowed: the pseudo-field is now scoped to `query_data_mart`'s `fields`/`sort` and explicitly excluded from `add_report`/`update_report`, so the model is not steered into a guaranteed rejection in the first place.

Full support — splitting `fields` into `uniqueCountConfig` on the report paths — is deliberately out of this PR. `splitUniqueCountFields` lives in `ee/mcp/tools` and `data-marts` imports nothing from `src/ee`, so wiring it means moving or duplicating the source resolution into the reports facade (and `update_report` only knows a report id, not a data mart id). That is a feature with its own review surface; it gets its own ticket.

Coverage: `output-controls-validator.service.spec.ts` (unit), `output-controls-error.mapper.spec.ts` (the model-facing message), and `output-controls.e2e-spec.ts` — a PUT naming `users__unique_count` in `columnConfig` now 400s with this code through the real controller/DTO/validator chain.

apps/web/src/features/data-marts/reports/edit/hooks/useGoogleSheetsReportForm.ts
zapolsky: Confirmed, and fixed in d6a476c73 at the persist boundary as you suggested — including the ordering caveat.

`foldEmptyUniqueCountConfig` maps anything that normalizes to an empty source list (`[]`, `false`, `null`) to `null`. `create-report.service.ts` applies it to both the validate call and the insert; `update-report.service.ts` applies it to **both sides** of the `previous` / `next` comparison at `:121-124`, so a stored `null` against an incoming `[]` no longer compares as changed and cannot flip `rejectUnavailableUniqueCountSources` to `true` on an untouched config — the permanent-400-on-rename the comment at `:138-141` exists to prevent.

The e2e round-trip block now pins it: a PUT of `[]` reads back as `null`. That block was also collapsed into one ordered test while I was there — it shared a single report across three cases and only passed in declaration order.

One half is genuinely outside this repo, and it is the mirror of your `create-report-request-api.dto.ts:130` note: the add-on still posts `uniqueCountConfig: []` unconditionally, which any pre-#1522 backend rejects with `must be a boolean value`. Both the old create and update DTOs carry `@IsOptional()` alongside `@IsBoolean()`, and `@IsOptional()` skips validation for `null` as well as `undefined` — so having the add-on send `null` when the list is empty is accepted by old and new backends alike, and matches this fold exactly. Two lines in the add-on's two save paths; happy to land it there.

zapolsky:
Thanks — this was a useful pass. Both blockers and the notes that only live in the review body are answered here; the two inline threads each have their own reply.

Everything below landed in d6a476c73 unless stated otherwise.

## Fixed in this repo

- **`blended-report-data.service.ts:351` — blank Output Alias.** `resolveUniqueCountSources` now falls back to the source's title, mirroring the picker's `defaultAlias.trim() || title`, so a blank alias can no longer render the bare `Unique Count` and collide with the main Data Mart's own header. The existing test pinned the old label and was inverted.
- **`report-data-cache.service.ts:132` — pre-fix numbers after deploy.** Rather than a runbook note, this ships as a migration that purges `report_data_cache`, so it happens on upgrade instead of depending on someone reading a checklist. The changeset says so in the numbers section. (Caveat worth naming: during a rolling deploy an old pod can repopulate an entry after the migration ran, so a report touched in that window still needs its cache to expire.)
- **`blended-report-data.service.ts:339` — silent drops.** The service has a `Logger` now, and warns on the two drops that change what a file contains: the joined Unique Count column (naming the alias paths and the reason class) together with the sort rule that goes with it, and the joined `COUNT` removed beside a `COUNT_DISTINCT`.
- **`relationship.types.ts:196` — the type lie.** `uniqueCountAvailability` is optional now, matching its sibling and matching what a cached response can actually carry. The single read site already defended itself; this stops a second consumer reading it directly.
- **`abstract-blended-query-builder.ts:650` — stale doc.** Rewritten to say the surrogate is emitted for every value-sleeve owner, and why (a declared key with a NULL component falls back to it).
- **`MODULAR_CONVENTIONS.md:70`.** §3 now records that a facade interface may live in `facades/` beside its implementation, which is what most modules do. You are right that the doc was the thing that was wrong.
- **`output-controls.e2e-spec.ts` — order-dependent block.** Collapsed into one ordered test, extended with the empty-array case.
- **Docs and changeset.** The joined-Unique-Count docs now say each ticked source adds its own scanned `SELECT DISTINCT` pass; the changeset gained the deliberate `0 orders, worth 2000` divergence you flagged in the 0–49 list, and the cache line above.

## Fixed in the paired Sheets add-on repo

- **The `columnConfig: null` → `JOINED_UNIQUE_COUNT_REQUIRES_COLUMN_CONFIG` 400 on the headline flow.** Exactly your one-liner: both save paths now compute `requireExplicit` from `needsExplicitColumnConfig(...)` instead of `aggregationActive || hasPreJoinSlice`. This was the most valuable finding in the review — the feature did not work end to end in the add-on without it.
- **Excluded source with a stored Unique Count.** The add-on keeps a group for such a source so the row has somewhere to render, and the row now carries an `isEmitted` state that says the column is not generated instead of showing a live Σ. That closes the "no control exists to remove it" half of the cross-repo note.
- **`buildUniqueCountSortColumns` collision.** Added the `taken` set, so two alias paths that build one SQL name no longer render two identically-keyed entries.

## Not changed, with reasons

- **`create-report-request-api.dto.ts:130` — release coordination.** Acknowledged; the server-side half is done (an empty list now persists as `null`), but that does not help a pre-#1522 backend receiving `[]` from a newer add-on. See the reply on the wire-value thread: both old DTOs pair `@IsBoolean()` with `@IsOptional()`, which skips `null`, so having the add-on send `null` when the list is empty is accepted by both backends and removes the coupling. Two lines, in the other repo; the release ordering still wants an explicit decision.
- **`abstract-blended-query-builder.ts:349` — dedup CTE for a Unique-Count-only source.** Real, and worth doing, but it changes SQL generation across five dialects plus the nested and `_kept_groups` cases for a pure-cost win with no effect on results. Separate ticket rather than more surface on this PR.
- **`metric-sleeve.builder.ts:729` — `alias` "dead".** It is not: four tests in `metric-sleeve.builder.spec.ts` read `sleeve.alias`.
- **`metric-sleeve.builder.ts:1014` — the discarded `resolveSleeveOwnerChain`.** Also not dead. `buildCountingSleeveCte` has a second caller at `:839` (the joined `COUNT_DISTINCT` sleeve) which does *not* resolve its own owner chain, so this call is that path's only guard against emitting joins onto an undeclared CTE. Removing it would trade a sentence naming what was being counted for a raw engine error. `groupByParts` is likewise used — by the non-nested branch at `:1092`.
- **`apply-output-config-change.ts` — `shouldDirty: false`.** Deliberate: marking the form dirty just because the picker repaired a config the user never touched would put every affected report into an unsaved state on open. The un-persisted repair is cosmetic — the run path drops the stale source and its sort rule in memory anyway, so nothing depends on the write landing.
- **`_oid` reserved name, `report.entity.ts:116` rollback, `update-report.service.ts:122` replace semantics.** Agreed with your own passes: the first is a false positive on a name the positional aliasing already frees, and the other two are pre-existing repo-wide patterns that deserve their own PRs rather than a gate on this one.

## Verification

`nest build` clean; backend 530 suites / 7360 tests; `output-controls` e2e 7 suites / 43 tests locally; web `tsc -b` and ESLint clean, 206 files / 1679 tests; the add-on's `lint`/`prettier`/`tsc --noEmit` clean with 37 suites / 423 tests. Markdownlint clean on the changeset and docs. CI on this head is running.

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
zapolsky: Confirmed and fixed in 920df8d73 — this was the sharpest finding of the batch, and the "silently the wrong mart" part is exactly right: the eligibility filter deduped only against real fields, so nothing downstream could tell the two apart.

Fixed on both ends:

- **Producer** (`mcp-data-marts.facade.impl.ts`): a `takenNames` set while advertising, first source winning — the same resolution the picker already uses, so the three surfaces now agree instead of disagreeing three ways.
- **Consumer** (`splitUniqueCountFields`): the `new Map(...)` is replaced by a loop that **throws** on a name two sources claim, rather than resolving it to the last one. Unreachable now that the producer refuses to emit one — which is the point: it is a drift alarm, and answering with the wrong Data Mart's count is not a failure mode worth degrading gracefully into.

Covered by a facade test (`a.b` + top-level `a_b` → one advertised source) and a splitter test asserting the refusal.

apps/backend/src/data-marts/services/blended-report-data.service.ts
zapolsky: Confirmed — the hidden-field disagreement is real and observable exactly as described. Fixed in 920df8d73.

`withoutStaleUniqueCountSorts` no longer treats a hidden blended field as owning the metric's name: a hidden field is not projected and cannot be sorted by, the picker's repair already says so, and disagreeing meant the scheduled run kept an `ORDER BY` that opening the editor deleted. The run path now drops the rule, matching the picker. New test pins it.

The single shared ownership resolver is the right end state and I did not go that far here — five call sites with different field universes is a refactor of its own, and this PR has had enough surface. What I did do is the same-shaped consolidation one level down: `aliasPathToCteName` now owns the alias-path mapping that two sites in this file were re-deriving inline (your other note), and the validator's ownership check was fixed in the same direction (the note below). Worth a follow-up that unifies all five.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
zapolsky: Confirmed and fixed in 920df8d73 — and worse than you saw: a commit earlier that same day extended this exact predicate to aggregations, date buckets and the projection, so the misdiagnosis had grown beyond filters by the time you flagged it.

`isUniqueCountColumn` now checks against a set that **keeps** hidden blended fields, not `homeFieldTypes`. A real field that went hidden is a broken schema link, so it falls through to `FILTER_COLUMN_UNKNOWN` and the disconnected diagnosis that names it and says how to repair it — the pre-PR behaviour, restored for every clause the predicate now guards. New test asserts the disconnected error rather than `UNIQUE_COUNT_FILTER_UNSUPPORTED` for that state.

apps/backend/src/data-marts/use-cases/query-data-mart.service.ts
zapolsky: Verified and fixed in 920df8d73 — the key really was absent from `McpQueryRunMetadataSchema.query`, so the parse dropped it and the comment above that line was describing an intent the code did not deliver.

One line in the schema, plus the "(survives schema parse)" test beside the existing sort one: a record carrying `uniqueCountConfig: ['orders']` now reaches `additionalParams.mcpQuery.query` intact.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
zapolsky: Agreed on the analysis — and this is now the second independent confirmation, since the earlier automated pass found the same waste from the other direction (`_kept_groups` doubling it for Totals). It is on the follow-up list with a raised priority.

Deliberately not in this PR. `buildAggregationCte` is called inside a recursive walk and a child's passthroughs land in its **parent's** aggregation CTE, so "skip a chain with no referenced columns" is not a local predicate — it needs the recursion's result, and getting it wrong changes emitted SQL on five dialects for a pure-cost win with no effect on results. Two approvals raise the priority of the ticket, not the risk tolerance of a branch that is already green.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Fair, and measured correctly. Follow-up rather than this PR.

The blocker is contract-shaped, not plumbing-shaped: `precomputedBlendableSchema` exists on both `compose` and `validateForReport`, but `getDataMartDetails` returns a DTO and never exposes the `BlendableSchemaDto` it computed — so threading it means widening a facade contract for a performance reason. And the heaviest part of that first call, `actualizeSchemaIfExpired`, is not avoidable either way. Worth doing with the MCP latency work rather than bolted on here.

apps/backend/src/data-marts/use-cases/update-report.service.ts
zapolsky: Fixed in 920df8d73. Both sides are normalized before the compare now, so a legacy `true` against its `['']` round-trip — and a stored `[]` against `null` — no longer read as a config change on a title-only save, and the cache survives.

Worth noting the neighbouring half that landed in the previous commit for a different reason: an empty list is folded to `null` at the persist boundary (the released Sheets add-on reads this field as a boolean, and `[]` is truthy). The two together mean the stored value is now canonical and the comparison is shape-insensitive. Two cases pinned in `update-report.service.spec.ts`.

apps/backend/src/data-marts/services/blended-report-data.service.ts
zapolsky: Agreed that this is the better shape — every drop predicate really is knowable before the chain build, and the mirror-coupling the method's own comment warns about is the cost of doing it after.

Not taking it in this PR. It is a behaviour-preserving refactor of the chain seeding, in a file that has taken several correctness fixes across the last two commits, and its whole value is structural. Filed rather than folded in — the risk/benefit points the other way while the branch is green and approved.

apps/backend/src/data-marts/services/blended-report-data.service.ts
zapolsky: Fixed in 920df8d73 — you are right that a silent drop is exactly the failure this duplication invites.

Both sites now go through the module that owns the mapping. `aliasPathToCteName` throws on a malformed path, which is correct for the schema-derived site but wrong for the one reading a **stored** config on the run path — there a malformed path is a dropped metric (the `!chain` branch already handles it), not a failed nightly run. So it gained a `tryAliasPathToCteName` sibling returning `undefined`, defined immediately beside it. Same rule, one place, two error postures.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
zapolsky: Agreed on the substance — the row identity really is derivable from the declared keys, and making the picker's diagnosis and the emit predicate structurally inseparable rather than fixture-pinned is worth more than the traversal it saves.

Not in this PR. `classifyJoinedUniqueCountAvailability` feeds the picker's disabled-row reasons *and* the query path's emit predicate, and this branch already changed how both read hidden fields (two notes up). Collapsing the two walks on top of that is a change I would rather land on its own, where a regression has somewhere obvious to point.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
zapolsky: Fixed in 920df8d73 — and you called it just in time: the drift you predicted had already happened. A commit earlier that day changed the validator's copy to read a filtered aggregation list, so the two predicates were no longer textually identical, synced by a comment that was now wrong.

`isMetricsOnlyProjection` lives in `report-like-read-plan.ts` as you suggested, next to `shouldIncludeRowCount`, and both call sites use it. The headers variant stays separate for the reason you gave — its main Unique Count is PK-gated, which is a different question.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
zapolsky: Fixed in 920df8d73, including the typing half — the local shape is gone entirely: `type GroupUniqueCount = NonNullable<BlendedGroup['uniqueCount']> & { state; onCheckedChange }`, so a sixth field added to the group cannot fail to arrive here. Both copies collapsed to spreads, at the construction site and at the `UniqueCountRow` call.

The second channel you noted — `BlendedGroupItem` reading `group.uniqueCount?.checked` for its initial-open state — is left as is on purpose: that one asks a different question (should this group start expanded) at a point where the props object does not exist yet.

zapolsky:
Thanks for the deep pass — this batch found things the earlier reviews did not, and two of them landed on code from the commit before yours. Every comment has an inline reply; 920df8d73 is the summary.

**Fixed (8):** the MCP name collision that answered with the wrong Data Mart's count; the since-hidden real field misdiagnosed as a metric (which the previous commit had widened from filters to aggregations, date buckets and the projection); the run path and the picker disagreeing about whether a hidden field owns the metric's name; the journalled `uniqueCountConfig` that zod was stripping; the raw config compare that re-billed a warehouse query on a title-only save; the metrics-only predicate, whose predicted drift had already happened; the duplicated alias-path → cteName mapping; and the hand-copied Unique Count row props.

**Filed, not folded in (4):** the dedup CTE for a Unique-Count-only source, the triple schema resolution, deleting `withoutChainsOnlyADroppedUniqueCountNeeded`, and the single primary-key traversal. Reasons are in each thread; the short version is that all four are structural or cost wins with no effect on results, and this branch has already taken two rounds of correctness fixes.

Verification on this head: `nest build` and ESLint clean, backend 530 suites / 7367 tests, `output-controls` e2e 7 / 43 locally, web `tsc -b` and ESLint clean with 206 files / 1682 tests. Eleven tests added for the fixes above.

PR #1523, opened 2026-08-11
feat(data-marts): sample SQL data marts via subquery — no bigquery.datasets.create required

laskevych:
## Why

Generating AI metadata (field aliases and descriptions, data mart title and description) for a SQL-based data mart resolved the data mart to a table reference, which materializes a technical view in the `owox_internal_<location>` dataset. Creating that dataset on first use runs `CREATE SCHEMA IF NOT EXISTS` — a DDL that requires project-level `bigquery.datasets.create` even when the dataset already exists. Users whose BigQuery access is data-only (e.g. `roles/bigquery.jobUser` + `roles/bigquery.dataViewer`) could not use the AI helper at all.

## What changed

`DataMartSampleDataService.sampleColumns` now builds the sample for SQL definitions as an inline derived table instead of resolving the technical view:

```sql
SELECT <escaped columns> FROM (<data mart sql>
) AS owox_sample_source LIMIT <n>
```

- Plain `SELECT` definitions are inlined on every storage type except `LEGACY_GOOGLE_BIGQUERY`.
- `WITH`/CTE definitions are additionally inlined on `GOOGLE_BIGQUERY`, whose dialect accepts a full query expression — including `WITH` — inside parenthesized `FROM (...)`. Other storages keep the technical-view path for CTEs, since support there varies (e.g. Athena, Redshift).
- `LEGACY_GOOGLE_BIGQUERY` is never inlined: its `definition.sqlQuery` holds the raw legacy SQL that is executable only after `LegacyBigQuerySqlPreprocessor` runs, and the sample path bypasses the query builder where that happens. Legacy marts keep the technical-view path, where the preprocessor runs at view creation.
- Wrap safety: the closing parenthesis starts on a new line so a trailing `--`/`#` line comment cannot swallow it; trailing semicolons and whitespace are stripped; a residual `;` (multi-statement script) falls back to the technical view.
- An explicitly passed `fullyQualifiedTableName` keeps priority over inlining.
- Non-SQL definitions (table, view, pattern, connector) and SQL that cannot be inlined are unaffected and keep the technical-view path, with the existing persistent human-readable error if permissions are missing there (#1516).

Out of scope (unchanged): `DataMartTableReferenceService`, `BigQueryCreateViewExecutor`, joined/blended reports, reports with output controls, data quality.

## Tests

`data-mart-sample-data.service.spec.ts` (12 tests passing):

- plain `SELECT` inlined as a derived table, technical view untouched, trailing `;` stripped
- `WITH`/CTE inlined on `GOOGLE_BIGQUERY`
- `LEGACY_GOOGLE_BIGQUERY` keeps the technical-view path (raw legacy SQL needs preprocessing)
- `WITH`/CTE falls back to the technical-view path on a non-BigQuery storage
- plain `SELECT` inlined on a non-BigQuery storage
- a definition ending in a `--` line comment still produces a valid derived table
- a multi-statement definition (`SELECT 1; SELECT 2`) falls back to the technical-view path
- SQL that is neither `SELECT` nor `WITH` (e.g. leading `DECLARE`) falls back to the technical-view path
- explicitly provided table name keeps priority over inlining
- existing identifier-escaping tests still pass

### DIFF
--- a/.changeset/6831-ai-metadata-sql-data-marts-subquery-sampling.md
+++ b/.changeset/6831-ai-metadata-sql-data-marts-subquery-sampling.md
@@ -8,6 +8,6 @@ Generating AI metadata (field aliases and descriptions, data mart title and desc
 
 Sample rows for AI generation are now fetched through an inline derived table built from the data mart's own SQL (`SELECT <columns> FROM (<data mart sql>) LIMIT <n>`), so the AI helper needs exactly the permissions a regular report run needs:
 
-- Plain `SELECT` definitions are inlined on every storage type.
-- `WITH`/CTE definitions are additionally inlined on Google BigQuery storages (including legacy), where the SQL dialect accepts a CTE inside a parenthesized derived table; other storages keep the previous behavior for CTEs.
-- SQL that cannot be inlined, explicit table references, and non-SQL definitions (table, view, pattern, connector) continue to use the technical-view path, with the existing human-readable error if permissions are missing there.
+- Plain `SELECT` definitions are inlined on every storage type except legacy Google BigQuery.
+- `WITH`/CTE definitions are additionally inlined on Google BigQuery storage, where the SQL dialect accepts a CTE inside a parenthesized derived table; other storages keep the previous behavior for CTEs.
+- Legacy Google BigQuery data marts, SQL that cannot be inlined, explicit table references, and non-SQL definitions (table, view, pattern, connector) continue to use the technical-view path, with the existing human-readable error if permissions are missing there.
--- a/.changeset/6831-ai-metadata-sql-data-marts-subquery-sampling.md
+++ b/.changeset/6831-ai-metadata-sql-data-marts-subquery-sampling.md
@@ -1,0 +1,13 @@
+---
+'owox': minor
+---
+
+# AI metadata generation for SQL data marts no longer requires dataset-creation permission
+
+Generating AI metadata (field aliases and descriptions, data mart title and description) for a SQL-based data mart previously materialized a technical view in the `owox_internal_<location>` dataset, and creating that dataset on first use required the project-level `bigquery.datasets.create` permission — so users with data-only access (e.g. `BigQuery Job User` plus `BigQuery Data Viewer`) could not use the AI helper at all.
+
+Sample rows for AI generation are now fetched through an inline derived table built from the data mart's own SQL (`SELECT <columns> FROM (<data mart sql>) LIMIT <n>`), so the AI helper needs exactly the permissions a regular report run needs:
+
+- Plain `SELECT` definitions are inlined on every storage type.
+- `WITH`/CTE definitions are additionally inlined on Google BigQuery storages (including legacy), where the SQL dialect accepts a CTE inside a parenthesized derived table; other storages keep the previous behavior for CTEs.
+- SQL that cannot be inlined, explicit table references, and non-SQL definitions (table, view, pattern, connector) continue to use the technical-view path, with the existing human-readable error if permissions are missing there.
--- a/apps/backend/src/data-marts/services/data-mart-sample-data.service.spec.ts
+++ b/apps/backend/src/data-marts/services/data-mart-sample-data.service.spec.ts
@@ -93,29 +93,79 @@ describe('DataMartSampleDataService', () => {
     expect(dataMartTableReferenceService.resolveTableName).not.toHaveBeenCalled();
     expect(dataMartSqlTableService.executeSqlToTable).toHaveBeenCalledWith(
       dataMart,
-      'SELECT `report_date` FROM (SELECT report_date, amount FROM `src.dataset.orders`) AS owox_sample_source LIMIT 5',
+      'SELECT `report_date` FROM (SELECT report_date, amount FROM `src.dataset.orders`\n) AS owox_sample_source LIMIT 5',
       { limit: 5 }
     );
   });
 
-  it.each([DataStorageType.GOOGLE_BIGQUERY, DataStorageType.LEGACY_GOOGLE_BIGQUERY])(
-    'inlines a WITH/CTE SQL data mart as a derived table on %s',
-    async storageType => {
-      const { service, dataMart, dataMartSqlTableService, dataMartTableReferenceService } =
-        createService(storageType, {
-          sqlQuery: 'WITH src AS (SELECT 1 AS report_date) SELECT * FROM src;',
-        });
+  it('inlines a WITH/CTE SQL data mart as a derived table on BigQuery', async () => {
+    const { service, dataMart, dataMartSqlTableService, dataMartTableReferenceService } =
+      createService(DataStorageType.GOOGLE_BIGQUERY, {
+        sqlQuery: 'WITH src AS (SELECT 1 AS report_date) SELECT * FROM src;',
+      });
 
-      await service.sampleColumns('data-mart-1', 'project-1', ['report_date']);
+    await service.sampleColumns('data-mart-1', 'project-1', ['report_date']);
 
-      expect(dataMartTableReferenceService.resolveTableName).not.toHaveBeenCalled();
-      expect(dataMartSqlTableService.executeSqlToTable).toHaveBeenCalledWith(
-        dataMart,
-        'SELECT `report_date` FROM (WITH src AS (SELECT 1 AS report_date) SELECT * FROM src) AS owox_sample_source LIMIT 5',
-        { limit: 5 }
-      );
-    }
-  );
+    expect(dataMartTableReferenceService.resolveTableName).not.toHaveBeenCalled();
+    expect(dataMartSqlTableService.executeSqlToTable).toHaveBeenCalledWith(
+      dataMart,
+      'SELECT `report_date` FROM (WITH src AS (SELECT 1 AS report_date) SELECT * FROM src\n) AS owox_sample_source LIMIT 5',
+      { limit: 5 }
+    );
+  });
+
+  it('keeps the technical-view path for legacy BigQuery, whose raw ODM SQL needs preprocessing', async () => {
+    const { service, dataMartSqlTableService, dataMartTableReferenceService } = createService(
+      DataStorageType.LEGACY_GOOGLE_BIGQUERY,
+      { sqlQuery: 'SELECT report_date FROM {{ source }}' }
+    );
+
+    await service.sampleColumns('data-mart-1', 'project-1', ['report_date']);
+
+    expect(dataMartTableReferenceService.resolveTableName).toHaveBeenCalledWith(
+      'data-mart-1',
+      'project-1'
+    );
+    expect(dataMartSqlTableService.executeSqlToTable).toHaveBeenCalledWith(
+      expect.anything(),
+      'SELECT `report_date` FROM `test-project`.`TestDataset`.`dashed-table-name` LIMIT 5',
+      { limit: 5 }
+    );
+  });
+
+  it('survives a trailing line comment in the inlined SQL definition', async () => {
+    const { service, dataMart, dataMartSqlTableService } = createService(
+      DataStorageType.GOOGLE_BIGQUERY,
+      { sqlQuery: 'SELECT report_date FROM `src.dataset.orders` -- daily orders' }
+    );
+
+    await service.sampleColumns('data-mart-1', 'project-1', ['report_date']);
+
+    expect(dataMartSqlTableService.executeSqlToTable).toHaveBeenCalledWith(
+      dataMart,
+      'SELECT `report_date` FROM (SELECT report_date FROM `src.dataset.orders` -- daily orders\n) AS owox_sample_source LIMIT 5',
+      { limit: 5 }
+    );
+  });
+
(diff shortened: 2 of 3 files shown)


apps/backend/src/data-marts/services/data-mart-sample-data.service.ts
```diff
@@ -45,4 +71,37 @@ export class DataMartSampleDataService {
...
+  private async resolveSampleSource(
+    dataMart: DataMart,
+    fullyQualifiedTableName?: string
+  ): Promise<string> {
+    const storageType = dataMart.storage.type;
+
+    if (!fullyQualifiedTableName) {
+      const definition = dataMart.definition;
+      if (definition && isSqlDefinition(definition)) {
+        const inlineSql = toInlineSubquerySql(definition.sqlQuery, storageType);
```
prokopetsv, line 105: 🔴 **Blocker — legacy BigQuery SQL bypasses the mandatory ODM preprocessor.**

For `LEGACY_GOOGLE_BIGQUERY`, `definition.sqlQuery` holds the raw legacy ODM query (`sync-legacy-data-mart.service.ts:114` stores `projection.query` verbatim). It is not executable BigQuery SQL until `LegacyBigQuerySqlPreprocessor.prepare` → `LegacyDataMartsService.parseQuery` expands it — which is exactly why every other path does it:

- `legacy-bigquery-create-view.executor.ts:27` (the technical-view path this PR replaces)
- `legacy-bigquery-sql-dry-run.executor.ts:26`
- `legacy-bigquery-query.builder.ts:59` — with the comment "the legacy SQL must be ODM-preprocessed first"

The sample path never reaches the query builder: `executeSqlToTable` puts a non-empty string on `SqlRunCommand.sql`, and `BigQuerySqlRunExecutor.execute` only calls `queryBuilder.buildQuery(definition)` when `sql` is falsy. So the raw legacy query is inlined and executed as-is.

**Trigger:** any legacy BigQuery SQL data mart whose query uses ODM syntax (the `parseQuery` error contract carries `variableName`, so variables are expected) → AI metadata sampling fails with a BigQuery syntax error, or resolves against the wrong source if the raw text happens to parse. On `main` this worked because the view was preprocessed at creation time.

Note this is **not** limited to the CTE case — plain `SELECT` is inlined on every storage type, so legacy marts hit it on the `/^select\b/i` branch as well.

Suggested fix: skip inlining for `LEGACY_GOOGLE_BIGQUERY` entirely (keep the technical-view path — legacy storages are not the audience for the `bigquery.datasets.create` fix), or await the preprocessor before wrapping.

apps/backend/src/data-marts/services/data-mart-sample-data.service.ts
```diff
@@ -45,4 +71,37 @@ export class DataMartSampleDataService {
...
+        if (inlineSql) {
+          return `(${inlineSql}) AS owox_sample_source`;
```
prokopetsv: 🔴 **Blocker — a trailing line comment comments out the closing parenthesis.**

`(${inlineSql}) AS owox_sample_source` puts `)` on the same line as the last line of the data mart SQL. For a definition ending in a `--` (or BigQuery `#`) comment:

```
SELECT order_id, amount FROM `p.d.orders` -- daily orders
```

the generated sample query is

```
SELECT `order_id` FROM (SELECT order_id, amount FROM `p.d.orders` -- daily orders) AS owox_sample_source LIMIT 5
```

Everything after `--` is a comment, so the statement ends at `FROM (SELECT order_id, amount FROM \`p.d.orders\`` → syntax error. `trim().replace(/[;\s]+$/, '')` strips semicolons and whitespace but not comments, and a trailing comment is common in hand-written data mart SQL. This is a regression: `CREATE OR REPLACE VIEW … AS <sql>` appends the SQL last, so the technical-view path tolerated it.

One-character fix:

```
return `(${inlineSql}\n) AS owox_sample_source`;
```

Worth covering with a test whose `sqlQuery` ends in `-- comment`. Same wrap-safety class: `SELECT 1; SELECT 2` also passes the `/^select\b/i` gate and yields an invalid derived table.

apps/backend/src/data-marts/services/data-mart-sample-data.service.spec.ts
```diff
@@ -79,6 +82,116 @@ describe('DataMartSampleDataService', () => {
...
+
+    expect(dataMartTableReferenceService.resolveTableName).not.toHaveBeenCalled();
+    expect(dataMartSqlTableService.executeSqlToTable).toHaveBeenCalledWith(
+      dataMart,
+      'SELECT `report_date` FROM (SELECT report_date, amount FROM `src.dataset.orders`) AS owox_sample_source LIMIT 5',
+      { limit: 5 }
+    );
+  });
+
+  it.each([DataStorageType.GOOGLE_BIGQUERY, DataStorageType.LEGACY_GOOGLE_BIGQUERY])(
```
prokopetsv: 🟡 **Non-blocker (follows from the blocker above).** This case asserts that the raw `sqlQuery` is inlined for `LEGACY_GOOGLE_BIGQUERY`, which pins the missing ODM preprocessing as expected behavior. Once the legacy path is fixed, this should assert the fallback to `resolveTableName` (or the preprocessed SQL, depending on which fix you take) rather than the verbatim query.

prokopetsv (changes requested):
## Review — PR #1523 (`claude/ai-metadata-sql-data-marts-1932ad`)

The permission fix is the right idea and the shape of the change is clean: sampling a SQL data mart through an inline derived table removes the `CREATE SCHEMA` DDL from the AI-metadata path, keeps the technical-view fallback for definitions that cannot be inlined, and preserves identifier escaping for the projected columns. Two correctness problems block merge. First, `LEGACY_GOOGLE_BIGQUERY` data marts store the raw legacy ODM query in `definition.sqlQuery`; every other execution path (view creation, dry run, non-output-controls query build) sends it through `LegacyBigQuerySqlPreprocessor.prepare` first, but `executeSqlToTable` receives an explicit SQL string and therefore skips `queryBuilder.buildQuery` entirely, so the un-preprocessed legacy SQL is executed verbatim. Second, the derived table is wrapped with the closing parenthesis on the same line as the data mart SQL, so any definition ending in a `--` or `#` line comment has its closing paren commented out. Both are hard failures with small fixes.

### Findings

**🔴 Blocker — legacy BigQuery SQL is inlined without ODM preprocessing** (`data-mart-sample-data.service.ts:94`)
Raw `projection.query` from `sync-legacy-data-mart.service.ts:114` is stored as-is. `LegacyBigQueryQueryBuilder.buildQuery` (line 59) states the constraint directly: "the legacy SQL must be ODM-preprocessed first". The sample path passes a non-empty `sql` to `BigQuerySqlRunExecutor.execute`, which only calls the query builder when `sql` is falsy — so the preprocessor never runs. Affects plain `SELECT` too, not just the CTE case added to `CTE_INLINE_STORAGE_TYPES`.

**🔴 Blocker — trailing line comment swallows the closing parenthesis** (`data-mart-sample-data.service.ts:96`)
`SELECT a, b FROM t -- daily orders` becomes `SELECT \`a\` FROM (SELECT a, b FROM t -- daily orders) AS owox_sample_source LIMIT 5`. Everything from `--` to end of line is a comment, so the query is truncated mid-expression and BigQuery returns a syntax error. The old technical-view path was immune because the DDL appends the SQL last. `[;\s]+$` does not strip comments.

**🟡 Non-blocker — the new legacy test asserts the un-preprocessed SQL** (`data-mart-sample-data.service.spec.ts:101`)
The `it.each` case pins the raw-SQL expectation for `LEGACY_GOOGLE_BIGQUERY`, so the first blocker is locked in by a passing test.

**🟡 Non-blocker — multi-statement definitions pass the prefix gate**
`SELECT 1; SELECT 2` matches `/^select\b/i`, and `[;\s]+$` only strips the trailing separator, producing an invalid derived table. Same class as the comment issue; a dedicated wrap-safety check would cover both.

No security or performance regressions found: the projected columns stay escaped through `IdentifierEscaperFacade`, the inlined SQL is the user's own definition executed under the same credentials as a report run, and the derived table costs the same scan as the equivalent view.

🔴 Blocker · 🟡 Non-Blocker

apps/backend/src/data-marts/services/data-mart-sample-data.service.ts
laskevych: Fixed in c7789978d — went with the first option: `LEGACY_GOOGLE_BIGQUERY` is now excluded from inlining entirely (early return in `toInlineSubquerySql`), so legacy marts keep the technical-view path where the preprocessor runs at view creation. Covered by a test asserting the fallback for a raw legacy-syntax query.

apps/backend/src/data-marts/services/data-mart-sample-data.service.ts
laskevych: Fixed in c7789978d — the closing parenthesis now starts on a new line (`(${inlineSql}\n) AS owox_sample_source`), with a test whose definition ends in `-- daily orders`. Also added the wrap-safety check for the multi-statement case: after stripping the trailing separator, any remaining `;` falls back to the technical view (test included). A semicolon inside a string literal would also trigger the fallback, but that only costs the view path, not a failure.

apps/backend/src/data-marts/services/data-mart-sample-data.service.spec.ts
laskevych: Updated in c7789978d — the `it.each` is now a BigQuery-only test, and the legacy case asserts the `resolveTableName` fallback instead of the verbatim query.

prokopetsv (commented):
## Re-review — PR #1523 (after c7789978d)

Both blockers from the previous review are verified fixed: `LEGACY_GOOGLE_BIGQUERY` is excluded from inlining via the early return in `toInlineSubquerySql` (with a test asserting the `resolveTableName` fallback), the closing parenthesis now starts on a new line (trailing `--` comment test included), and the residual-semicolon guard covers multi-statement definitions. The spec passes (12/12). Inlining plain `SELECT` on every storage matches what each storage's query builder already does with `(cleanQuery)` on other paths, so no dialect risk is introduced. One 🟡 remains: the changeset and PR description still describe the pre-fix behavior.

### Findings

**🟡 Non-Blocker — changeset and PR description are stale after c7789978d** (`.changeset/6831-ai-metadata-sql-data-marts-subquery-sampling.md:12`)
Both say WITH/CTE is inlined on Google BigQuery storages "including legacy", but the fix excludes `LEGACY_GOOGLE_BIGQUERY` from inlining entirely. This text ships in release notes. See inline comment.

**🟡 Non-Blocker — optional follow-up, outside this diff**
The same trailing-line-comment hazard fixed here still exists in the query builders' own wrapping, e.g. `bigquery-query.builder.ts:137` and `athena-query.builder.ts:113` use `(${cleanQuery})` with `)` on the last SQL line. Not introduced by this PR; fine as a separate task.

No blockers. Correctness, security (column escaping via `IdentifierEscaperFacade` unchanged), and test coverage look solid.

.changeset/6831-ai-metadata-sql-data-marts-subquery-sampling.md
```diff
@@ -0,0 +1,13 @@
...
+---
+
+# AI metadata generation for SQL data marts no longer requires dataset-creation permission
+
+Generating AI metadata (field aliases and descriptions, data mart title and description) for a SQL-based data mart previously materialized a technical view in the `owox_internal_<location>` dataset, and creating that dataset on first use required the project-level `bigquery.datasets.create` permission — so users with data-only access (e.g. `BigQuery Job User` plus `BigQuery Data Viewer`) could not use the AI helper at all.
+
+Sample rows for AI generation are now fetched through an inline derived table built from the data mart's own SQL (`SELECT <columns> FROM (<data mart sql>) LIMIT <n>`), so the AI helper needs exactly the permissions a regular report run needs:
+
+- Plain `SELECT` definitions are inlined on every storage type.
+- `WITH`/CTE definitions are additionally inlined on Google BigQuery storages (including legacy), where the SQL dialect accepts a CTE inside a parenthesized derived table; other storages keep the previous behavior for CTEs.
```
prokopetsv: 🟡 **Non-Blocker — "including legacy" is no longer true after c7789978d.**

`toInlineSubquerySql` now returns `null` for `LEGACY_GOOGLE_BIGQUERY`, so legacy marts always keep the technical-view path — for plain `SELECT` too. This changeset ships in release notes, so users would expect legacy inlining that doesn't happen.

Fix: "`WITH`/CTE definitions are additionally inlined on Google BigQuery storage; legacy BigQuery and other storages keep the previous technical-view behavior." The PR description has the same stale claim (`GOOGLE_BIGQUERY` and `LEGACY_GOOGLE_BIGQUERY`).

prokopetsv (approved):

.changeset/6831-ai-metadata-sql-data-marts-subquery-sampling.md
laskevych: Fixed in 913b3a41c — the changeset now says plain `SELECT` is inlined on every storage except legacy Google BigQuery, WITH/CTE only on Google BigQuery, and legacy marts keep the technical-view path. The PR description is updated the same way.

PR #1524, opened 2026-08-12
feat(data-marts): measure Data Last Updated Date for AWS Redshift storage

laskevych:
## What

Data Last Updated Date is now measured for AWS Redshift Data Marts, on every surface where it already worked for BigQuery: the Data Mart page tile, the Data Marts list, the canvas (badge + Actions sweep), MCP `query_data_mart` responses, and run history. No changes outside the storage layer — the new resolver registers into the existing `SourceDataLastUpdatedResolver` family and everything above it picks Redshift up automatically.

## How

- **Source discovery** — `EXPLAIN` over the composed SQL (the same mechanism the existing Redshift dry-run uses). The planner expands views, nested and late-binding included, down to base-table scans, so no SQL parsing is needed.
- **Schema resolution** — Redshift prints local scan targets without their schema, so bare names are mapped back through `svv_redshift_tables` within the connection's database. A name matching several schemas is reported as an unknown source with a note rather than guessed: `partial` coverage promises a value that is only ever older than the truth, and a guess could be newer.
- **Modification times** — `SHOW TABLES FROM SCHEMA`, currently the only Redshift surface that reports `last_modified_time` (it lags real writes by up to ~5 minutes). Read defensively by column name: older releases without the column degrade to unknown sources instead of failing.
- Spectrum external tables are listed as unknown sources with a note (their data lives outside Redshift). Materialized-view backing tables (`mv_tbl__…`) are renamed back to the view, with the value meaning the last refresh.
- Catalog and per-schema lookups are cached across a batch, so a canvas-wide sweep over one storage pays each schema once. The Data API polling chain gained optional `pollIntervalMs`/`signal` (defaults unchanged) so metadata statements poll at 250ms and abort with the orchestrator's soft timeout.

## Tests

- 13 resolver specs mirroring the BigQuery resolver suite: complete/partial/unavailable coverage, Spectrum, ambiguous and unlocatable names, missing modification times, per-item failure isolation, batch-level lookup sharing, abort before and between items, non-Redshift config.
- 10 specs for the EXPLAIN plan parser (provisioned/serverless shapes, aliases, quoted identifiers, Spectrum scans, `volt_` internals, deduplication).
- All 10 existing Redshift suites pass (137 tests); ESLint and `tsc` clean against the baseline.

Docs updated (`data-last-updated.md`, `mcp.md`) with the Redshift caveats: ~5-minute metadata lag, older releases reporting Unknown, sources outside the connection's database.

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
+++ b/apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
@@ -14,7 +14,28 @@ import { RedshiftConfig } from '../schemas/redshift-config.schema';
 import { RedshiftCredentials } from '../schemas/redshift-credentials.schema';
 import { RedshiftConnectionType } from '../enums/redshift-connection-type.enum';
 
+/**
+ * Tuning for the execute-and-poll cycle of a single statement. Defaults preserve the historic
+ * behaviour (1s polling, no cancellation) for the report/read paths; metadata lookups pass a
+ * shorter interval so a handful of tiny catalog queries does not cost seconds of idle waiting.
+ */
+export interface RedshiftQueryOptions {
+  pollIntervalMs?: number;
+  /** Stops the polling wait early; the statement itself keeps running server-side. */
+  signal?: AbortSignal;
+}
+
 export class RedshiftApiAdapter {
+  /** Overall per-statement deadline; unchanged from the historic 300 × 1s polling loop. */
+  private static readonly QUERY_TIMEOUT_MS = 300_000;
+  private static readonly DEFAULT_POLL_INTERVAL_MS = 1000;
+  /**
+   * Identifiers interpolated into SHOW TABLES, which takes no bind parameters and (unlike
+   * regular SQL) has no documented quoting syntax — so only plain identifiers are accepted and
+   * anything else is refused rather than escaped.
+   */
+  private static readonly SAFE_IDENTIFIER_RE = /^[a-zA-Z_][a-zA-Z0-9_$]*$/;
+
   private readonly logger = new Logger(RedshiftApiAdapter.name);
   private readonly redshiftDataClient: RedshiftDataClient;
   private readonly config: RedshiftConfig;
@@ -65,12 +86,19 @@ export class RedshiftApiAdapter {
   /**
    * Polls until query completes (FINISHED, FAILED, or ABORTED)
    */
-  public async waitForQueryToComplete(statementId: string): Promise<void> {
-    const maxAttempts = 300; // 5 minutes with 1s intervals
+  public async waitForQueryToComplete(
+    statementId: string,
+    options?: RedshiftQueryOptions
+  ): Promise<void> {
+    const intervalMs = options?.pollIntervalMs ?? RedshiftApiAdapter.DEFAULT_POLL_INTERVAL_MS;
+    const maxAttempts = Math.ceil(RedshiftApiAdapter.QUERY_TIMEOUT_MS / intervalMs);
     let attempts = 0;
 
     while (attempts < maxAttempts) {
-      await new Promise(resolve => setTimeout(resolve, 1000));
+      await this.sleep(intervalMs, options?.signal);
+      if (options?.signal?.aborted) {
+        throw new Error(`Aborted while waiting for statement ${statementId}`);
+      }
 
       const describeCommand = new DescribeStatementCommand({ Id: statementId });
       const response = await this.redshiftDataClient.send(describeCommand);
@@ -92,8 +120,24 @@ export class RedshiftApiAdapter {
     }
 
     throw new Error(
-      `Query execution timeout after ${maxAttempts} seconds for statement ${statementId}`
+      `Query execution timeout after ${RedshiftApiAdapter.QUERY_TIMEOUT_MS / 1000} seconds for statement ${statementId}`
     );
+  }
+
+  /** Resolves early (without throwing) when the signal aborts mid-wait. */
+  private async sleep(ms: number, signal?: AbortSignal): Promise<void> {
+    if (signal?.aborted) return;
+    await new Promise<void>(resolve => {
+      const timer = setTimeout(() => {
+        signal?.removeEventListener('abort', onAbort);
+        resolve();
+      }, ms);
+      const onAbort = () => {
+        clearTimeout(timer);
+        resolve();
+      };
+      signal?.addEventListener('abort', onAbort, { once: true });
+    });
   }
 
   /**
@@ -135,10 +179,11 @@ export class RedshiftApiAdapter {
    * Execute a SELECT query and return rows as plain JS objects
    */
   public async executeQueryAndGetRows(
-    query: string
+    query: string,
+    options?: RedshiftQueryOptions
   ): Promise<Array<Record<string, string | null>>> {
     const { statementId } = await this.executeQuery(query);
-    await this.waitForQueryToComplete(statementId);
+    await this.waitForQueryToComplete(statementId, options);
 
     const rows: Array<Record<string, string | null>> = [];
     let columns: string[] | undefined;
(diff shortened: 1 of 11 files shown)


apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
```diff
@@ -243,6 +288,98 @@ export class RedshiftApiAdapter {
...
+        tableName: row.table_name,
+        lastModifiedTime: this.toIsoUtcTimestamp(row.last_modified_time),
+      }));
+  }
+
+  /**
+   * `2025-11-18 15:44:34.856073` (Redshift timestamps are UTC, reported without a zone) →
+   * ISO-8601 with millisecond precision, or null for anything absent or unrecognised.
+   */
+  private toIsoUtcTimestamp(value: string | null | undefined): string | null {
```
prokopetsv: 🟡 This is the only place that parses Redshift's actual `2025-11-18 15:44:34.856073` shape, and no test exercises it — every resolver spec stubs `lastModifiedTime` as an already-ISO string. A format the regex rejects (e.g. a `+00` offset) returns null and surfaces as `no modification time reported by Redshift`, indistinguishable from an old Redshift release.

Suggest exporting it and adding three cases: microseconds, no fractional part, unparseable.

apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
```diff
@@ -243,6 +288,98 @@ export class RedshiftApiAdapter {
...
+   * so callers cache the result per schema. On older Redshift releases the column does not
+   * exist yet; those rows come back with a null `lastModifiedTime`.
+   */
+  public async getSchemaTablesInfo(
+    database: string,
+    schema: string,
+    options?: RedshiftQueryOptions
+  ): Promise<Array<{ tableName: string; lastModifiedTime: string | null }>> {
+    for (const identifier of [database, schema]) {
+      if (!RedshiftApiAdapter.SAFE_IDENTIFIER_RE.test(identifier)) {
```
prokopetsv, line 355: 🟡 The guard refuses identifiers that Redshift accepts when quoted — mixed case (`CREATE SCHEMA "Analytics"`), hyphens, leading digits. The throw is caught in `schemaTables`, so every table in that schema reports `could not read table metadata` and coverage drops to partial.

Graceful and logged, so not a blocker. Worth adding to the Redshift caveats in `data-last-updated.md`, next to the ~5-minute lag note, so the Unknown is explainable.

apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
```diff
@@ -243,6 +288,98 @@ export class RedshiftApiAdapter {
...
+    if (tableNames.length === 0) {
+      return [];
+    }
+
+    const nameList = tableNames.map(name => `'${this.escapeLiteral(name)}'`).join(', ');
+    const query = `
+      SELECT schema_name, table_name
+      FROM svv_redshift_tables
+      WHERE database_name = '${this.escapeLiteral(database)}'
+        AND table_type = 'TABLE'
```
prokopetsv, line 328: 🟡 The materialized-view path assumes `svv_redshift_tables` lists `mv_tbl__…` backing tables with `table_type = 'TABLE'`, and that `SHOW TABLES` reports a time for them. If either does not hold, the MV resolves to `table not found in database` and coverage silently drops to partial.

The rename spec stubs both adapter calls, so it cannot catch that. Worth one check against a real cluster before merge.

apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
```diff
@@ -65,12 +86,19 @@ export class RedshiftApiAdapter {
...
    * Polls until query completes (FINISHED, FAILED, or ABORTED)
    */
-  public async waitForQueryToComplete(statementId: string): Promise<void> {
-    const maxAttempts = 300; // 5 minutes with 1s intervals
+  public async waitForQueryToComplete(
+    statementId: string,
+    options?: RedshiftQueryOptions
+  ): Promise<void> {
+    const intervalMs = options?.pollIntervalMs ?? RedshiftApiAdapter.DEFAULT_POLL_INTERVAL_MS;
+    const maxAttempts = Math.ceil(RedshiftApiAdapter.QUERY_TIMEOUT_MS / intervalMs);
```
prokopetsv, line 100: 🟡 `pollIntervalMs: 0` makes `maxAttempts` `Infinity` and turns the wait into a zero-delay loop hammering `DescribeStatement`. No caller does this today, but the option is public and unvalidated.

One-line guard:

```ts
const intervalMs = Math.max(1, options?.pollIntervalMs ?? RedshiftApiAdapter.DEFAULT_POLL_INTERVAL_MS);
```

prokopetsv (approved):
## Review — PR #1524 (`feat/6794-data-last-updated-redshift`)

The PR delivers what the description promises: a `SourceDataLastUpdatedResolver` for Redshift that registers into the existing family, so every surface above it (tile, lists, canvas, MCP, run history) picks Redshift up with no changes outside the storage layer. It mirrors the BigQuery resolver closely — same batch contract, same abort handling, same `complete`/`partial`/`unavailable` semantics, same "never guess a value that could be newer than the truth" rule — which makes it easy to read and to maintain alongside its sibling. The adapter changes are additive and default-preserving (`pollIntervalMs`/`signal` optional, 1s polling and the 300s deadline unchanged), so existing report and dry-run paths are unaffected. Ignoring `item.params` is correct here, since the Redshift clause renderer inlines every value and emits `params: []` by design. Test coverage is good where it counts: 13 resolver specs and 10 parser specs cover coverage transitions, Spectrum, ambiguity, per-item failure isolation, batch cache sharing, and both abort points.

**No major issues detected.** Four non-blocking notes below, mostly about assumptions that only stubs currently verify.

### Findings

🔴 Blocker — none.

🟡 Non-Blocker

1. `redshift-api.adapter.ts:372` — `toIsoUtcTimestamp` is the only code touching Redshift's real timestamp format, and nothing tests it; every resolver spec feeds pre-formatted ISO strings.
2. `redshift-api.adapter.ts:349` — schemas whose names need quoting (mixed case, hyphen) are refused and silently report Unknown; not listed in the docs caveats.
3. `redshift-api.adapter.ts:322` — the materialized-view rename depends on catalog assumptions that only stubs verify.
4. `redshift-api.adapter.ts:94` — `pollIntervalMs: 0` would make `maxAttempts` infinite.

apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
laskevych: Fixed in 943354cc4 — the conversion now lives in `redshift-timestamp.utils.ts` with its own spec: the real microsecond shape SHOW TABLES returns, no fractional part, short-fraction padding, and unparseable input. While at it, explicit UTC markers (`Z`, `+00`, `+0000`, `+00:00`) are now tolerated, and non-zero offsets still return null rather than being silently relabeled as UTC.

apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
laskevych: Documented in 943354cc4 — added to the Redshift caveats in `data-last-updated.md`, next to the ~5-minute lag note: schemas whose names require quoting (mixed case, hyphens, non-ASCII) cannot be identified and report unknown entries with partial coverage.

apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
laskevych: Verified on a live cluster during manual testing: `svv_redshift_tables` does list the `mv_tbl__…` backing table with `table_type = 'TABLE'`, `SHOW TABLES` reports a modification time for it, and `REFRESH MATERIALIZED VIEW` moves the value (after the ~5-minute metadata lag). The source entry renders under the view's own name with the "time of the last refresh" note, as intended.

apps/backend/src/data-marts/data-storage-types/redshift/adapters/redshift-api.adapter.ts
laskevych: Fixed in 943354cc4 with exactly this clamp — a zero or negative interval now falls back to a 1ms floor, keeping the attempt budget finite.

PR #1525, opened 2026-08-12
fix(plugin-host): improve sync rate limiting

amarchenk0:
## What changed

- use a 30-second sync cooldown for GitHub App and server-token access
- retain the safer 300-second cooldown for anonymous GitHub access
- reuse the last validated plugin version when publishing is blocked only by the cooldown
- distinguish an active synchronization from the cooldown with `PLUGIN_SYNC_IN_PROGRESS`
- return the actual remaining retry time when no validated version is available
- document the mode-aware defaults and add a changeset

## Why

The previous per-plugin limiter applied the same 300-second window to every GitHub access mode and treated both an active synchronization and a completed synchronization's cooldown as the same failure. Publish therefore returned `PLUGIN_SYNC_RATE_LIMITED` even when ODM already had a validated version it could safely publish.

The claim result now reports whether the sync is active or merely cooling down. Publish uses the stored version only for the cooldown case; it still waits for an active sync and never invents a version when the first sync failed or produced no eligible release.

Task: [Improve PLUGIN_SYNC_RATE_LIMITED-6829](https://owox.fibery.io/Sprint_Work/Work_Item/Improve-PLUGIN_SYNC_RATE_LIMITED-6829)

## Validation

- 80 targeted backend unit tests
- 6 plugin-host schema e2e tests
- ESLint for all changed TypeScript files
- production TypeScript build check
- Markdownlint for changed documentation


.env.example
```diff
@@ -279,13 +279,14 @@ GITHUB_API_BASE_URL=
...
-# mutual-exclusion window between concurrent syncs.
-# PLUGIN_HOST_SYNC_MIN_INTERVAL_SEC=300
+# Numeric settings stay commented out. A present-but-empty value is not the same as an
+# unset value: it can fail boot-time validation or disable the sync throttle. Uncomment
+# only to override.
+
+# Minimum seconds between two synchronization starts for the same plugin.
+# Defaults to 30 with GitHub App/server-token access and 300 for anonymous access.
+# Setting this overrides the mode-aware default for every GitHub access mode.
+# PLUGIN_HOST_SYNC_MIN_INTERVAL_SEC=
```
romandubovyi: ⚪ **Non-blocking:** Uncommenting this line as written sets an empty value, which `z.coerce.number()` turns into `0`; `.min(0)` accepts it, `trimmed()` yields the truthy string `"0"`, and `getSyncMinIntervalMs` ends up at `0` — the per-plugin throttle is silently off. Before this PR the sample read `=300` and was safe to uncomment. Keeping a concrete value here (e.g. `# PLUGIN_HOST_SYNC_MIN_INTERVAL_SEC=30`) would keep the most natural action safe.

🤖 Generated with Claude Code (Opus 5) + Grok CLI (grok-4.5)

apps/backend/src/plugin-host/use-cases/sync-plugin-releases.service.spec.ts
```diff
@@ -422,27 +422,85 @@ describe('SyncPluginReleasesService', () => {
...
-        accessMode: GithubAccessMode.ANONYMOUS,
-        acceptedSemvers: ['1.0.0'],
-        unchangedSemvers: [],
-        rejections: [],
-      };
-      s.pluginService.findByGithubRepoId.mockResolvedValue({
-        id: 'p1',
-        currentVersionId: 'v1',
-        lastSyncReport: stored,
+      s.pluginService.tryClaimSyncSlot.mockResolvedValue({
```
romandubovyi, line 523: ⚪ **Non-blocking:** `SyncPluginReleasesCommand` documents that background sweeps may quietly return the stored or empty state while a first sync is still running, but this is the only `requireCurrentVersion: false` case in the block and it mocks `rate_limited`; both `in_progress` cases above run with the default `true` and assert a throw. Worth adding an `in_progress` + `requireCurrentVersion: false` example asserting no throw, `throttled: true`, and the stored (or empty) report — that is the path `RunPluginUpdateCheckService` relies on during an active lease.

🤖 Generated with Grok CLI (grok-4.5)

romandubovyi (approved):
Reviewed the mode-aware cooldown, the `PluginSyncSlotClaim` union and the new `PLUGIN_SYNC_IN_PROGRESS` path. The claim/in-progress/rate-limited split reads cleanly and the lease-cutoff reuse keeps the re-read consistent with the UPDATE predicate. No blockers — two non-blocking notes inline.

🤖 Generated with Claude Code (Opus 5) + Grok CLI (grok-4.5) + Kimi CLI

PR #1526, opened 2026-08-12
refactor(data-marts): sanitize publish errors by exception code

prokopetsv:
# Problem

Publishing Data Mart drafts in bulk decides which error text is safe to show, because the trigger response is readable by any project **viewer** while publishing requires **editor**, and errors on that path can carry SQL fragments, table paths, and service-account names from warehouse drivers.

That decision was made by comparing each message against three approved strings, with a second checkpoint that trusted any `BusinessViolationException`. Both were safe but conventional rather than structural, with two consequences. First, authored messages that were not among the three were discarded: a malformed table name produced "Publishing failed. Open the Data Mart to see details." instead of "Invalid identifier format. Expected: project.dataset.table" — the most actionable failure a user can get. Second, the "BVE implies safe text" assumption held only because per-draft errors are caught inside the loop; the same type is thrown with raw warehouse output by the definition validator facade and with an arbitrary `error.message` by `ActualizeDataMartSchema`, so reordering the error handling could have reopened the leak with no failing test.

# Solution

Both checkpoints now key on a stable **code** rather than message text, using the `BaseException.code` contract the codebase already documents ("Clients branch on this instead of matching human-readable messages"). A message is trusted because it is tagged as authored, not because its bytes match a list.

- `ValidationResult` gained a `code` field and an `authoredFailure()` constructor alongside the existing `failure()`. Keeping both matters: `failure()` stays uncoded, so anything not explicitly migrated still gets genericized. Safe-by-default rather than opt-out.
- Only demonstrably authored failures were coded. The dry-run `catch` blocks and Databricks' `EXPLAIN` output were deliberately left uncoded so raw driver text is still replaced.
- `NO_PERMISSION` intentionally stays a Nest `ForbiddenException`. Giving it a code would mean converting it to `BusinessViolationException`, which `BaseExceptionFilter` hardcodes to HTTP 400 — silently turning a 403 into a 400 for API clients. A permission denial genuinely is a 403, so it is matched by identity instead, with the reasoning recorded at the call site.
- The batch-level checkpoint matches **allowlist membership**, not merely the presence of a code. A code alone only proves some thrower opted in, not that this path authored the wording, so the two checkpoints use separate sets scoped to what can legitimately reach each.

Trade-off considered and rejected: converting `NO_PERMISSION` for uniformity. Consistency was not worth changing a public status code.

# Changes

- Added `DataMartValidationCode` and `ValidationResult.authoredFailure()` to the definition-validator interface; left `failure()` uncoded as the safe default
- Tagged identifier-format failures across all five warehouse validators (Athena, BigQuery, Databricks, Redshift, Snowflake), plus the facade's definition/config/credentials messages
- Propagated the validation code through `DataMartDefinitionValidatorFacade.checkIsValid` so the publish path can distinguish authored text from driver output (the `validate-definition` response does not expose `code` yet)
- Converted `PUBLISH_DATA_MART_ERRORS` to `{ code, message }` pairs and gave the trigger-level failures in `PublishDataStorageDraftsService` their own codes
- Replaced both string allowlists with code allowlists; the trigger handler now requires membership in an exported set rather than trusting any coded exception
- Kept `NO_PERMISSION` as a `ForbiddenException` to preserve the endpoint's 403, matched by identity with an explanatory comment
- Added a table-driven parity spec asserting every warehouse validator codes its identifier failure, so a newly added warehouse is caught automatically
- Added a facade spec pinning code propagation, including that an uncoded failure stays uncoded
- Extended the publish-drafts specs to cover an authored validator message passing through, an uncoded exception being genericized, and an unrecognized code being genericized
- Updated the existing changeset to mention that the notification now names the malformed-table-name reason

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
+++ b/apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
@@ -3,6 +3,8 @@ import { BigQueryDataMartValidator } from './bigquery/services/bigquery-datamart
 import { DatabricksDataMartValidator } from './databricks/services/databricks-datamart.validator';
 import { RedshiftDataMartValidator } from './redshift/services/redshift-datamart.validator';
 import { SnowflakeDataMartValidator } from './snowflake/services/snowflake-datamart.validator';
+import { LegacyBigQueryDataMartValidator } from './bigquery/services/legacy/legacy-bigquery-datamart.validator';
+import { validatorProviders } from './data-storage-providers';
 import {
   DataMartValidationCode,
   DataMartValidator,
@@ -14,8 +16,12 @@ import {
  * output, so it replaces it with a generic reason — which is how Athena drifted
  * out of parity with the other four.
  *
- * Each validator checks identifiers before touching credentials or an adapter,
- * so stub dependencies are never used.
+ * The list is asserted against the registered validator providers, so adding a
+ * warehouse without adding it here fails rather than passing silently.
+ *
+ * Identifiers are checked before config, credentials or an adapter in every
+ * validator, so the stub dependencies below are never dereferenced. Redshift
+ * was the exception until it was reordered to match.
  */
 describe('Data Mart validators — identifier failure parity', () => {
   const validators: [string, DataMartValidator][] = [
@@ -24,7 +30,12 @@ describe('Data Mart validators — identifier failure parity', () => {
     ['databricks', new DatabricksDataMartValidator(null as never, null as never)],
     ['redshift', new RedshiftDataMartValidator(null as never, null as never)],
     ['snowflake', new SnowflakeDataMartValidator(null as never, null as never)],
+    ['legacy-bigquery', new LegacyBigQueryDataMartValidator(null as never, null as never)],
   ];
+
+  it('covers every registered Data Mart validator', () => {
+    expect(validators).toHaveLength(validatorProviders.length);
+  });
 
   it.each(validators)('%s codes its invalid-identifier failure', async (_name, validator) => {
     const result = await validator.validate(
--- a/apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
+++ b/apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
@@ -1,0 +1,40 @@
+import { AthenaDataMartValidator } from './athena/services/athena-datamart.validator';
+import { BigQueryDataMartValidator } from './bigquery/services/bigquery-datamart.validator';
+import { DatabricksDataMartValidator } from './databricks/services/databricks-datamart.validator';
+import { RedshiftDataMartValidator } from './redshift/services/redshift-datamart.validator';
+import { SnowflakeDataMartValidator } from './snowflake/services/snowflake-datamart.validator';
+import {
+  DataMartValidationCode,
+  DataMartValidator,
+} from './interfaces/data-mart-validator.interface';
+
+/**
+ * Every warehouse validator must tag its identifier-format failure with a code.
+ * Without one the publish path cannot tell the authored message from raw driver
+ * output, so it replaces it with a generic reason — which is how Athena drifted
+ * out of parity with the other four.
+ *
+ * Each validator checks identifiers before touching credentials or an adapter,
+ * so stub dependencies are never used.
+ */
+describe('Data Mart validators — identifier failure parity', () => {
+  const validators: [string, DataMartValidator][] = [
+    ['athena', new AthenaDataMartValidator(null as never, null as never)],
+    ['bigquery', new BigQueryDataMartValidator(null as never, null as never)],
+    ['databricks', new DatabricksDataMartValidator(null as never, null as never)],
+    ['redshift', new RedshiftDataMartValidator(null as never, null as never)],
+    ['snowflake', new SnowflakeDataMartValidator(null as never, null as never)],
+  ];
+
+  it.each(validators)('%s codes its invalid-identifier failure', async (_name, validator) => {
+    const result = await validator.validate(
+      { fullyQualifiedName: 'this is !! not a valid identifier' } as never,
+      {} as never,
+      {} as never
+    );
+
+    expect(result.valid).toBe(false);
+    expect(result.errorMessage).toContain('Invalid identifier format');
+    expect(result.code).toBe(DataMartValidationCode.INVALID_IDENTIFIER_FORMAT);
+  });
+});
--- a/apps/backend/src/data-marts/data-storage-types/facades/data-mart-definition-validator-facade.service.ts
+++ b/apps/backend/src/data-marts/data-storage-types/facades/data-mart-definition-validator-facade.service.ts
@@ -1,7 +1,11 @@
 import { Inject, Injectable } from '@nestjs/common';
 import { TypeResolver } from '../../../common/resolver/type-resolver';
 import { DataStorageType } from '../enums/data-storage-type.enum';
-import { DataMartValidator, ValidationResult } from '../interfaces/data-mart-validator.interface';
+import {
+  DataMartValidationCode,
+  DataMartValidator,
+  ValidationResult,
+} from '../interfaces/data-mart-validator.interface';
 import { DATA_MART_VALIDATOR_RESOLVER } from '../data-storage-providers';
 import { DataMart } from '../../entities/data-mart.entity';
 import { BusinessViolationException } from '../../../common/exceptions/business-violation.exception';
@@ -18,16 +22,25 @@ export class DataMartDefinitionValidatorFacade {
   async validate(dataMart: DataMart): Promise<ValidationResult> {
     const definition = dataMart.definition;
     if (!definition) {
-      return new ValidationResult(false, 'DataMart definition not found');
+      return ValidationResult.authoredFailure(
+        DataMartValidationCode.DEFINITION_NOT_FOUND,
+        'Data Mart definition not found'
+      );
     }
 
     const config = dataMart.storage.config;
     if (!config) {
-      return new ValidationResult(false, 'DataMart storage config not found');
+      return ValidationResult.authoredFailure(
+        DataMartValidationCode.STORAGE_CONFIG_NOT_FOUND,
+        'Data Mart storage config not found'
+      );
     }
 
     if (!dataMart.storage.credentialId) {
-      return new ValidationResult(false, 'DataMart storage credentials not found');
+      return ValidationResult.authoredFailure(
+        DataMartValidationCode.STORAGE_CREDENTIALS_NOT_FOUND,
+        'Data Mart storage credentials not found'
+      );
     }
 
     const credentials = await this.credentialsResolver.resolve(dataMart.storage);
@@ -39,7 +52,7 @@ export class DataMartDefinitionValidatorFacade {
   async checkIsValid(dataMart: DataMart): Promise<void> {
     const result = await this.validate(dataMart);
     if (!result.valid) {
-      throw new BusinessViolationException(result.errorMessage!, result.details);
+      throw new BusinessViolationException(result.errorMessage!, result.details, result.code);
     }
   }
 }
--- a/apps/backend/src/data-marts/data-storage-types/facades/data-mart-definition-validator-facade.service.ts
+++ b/apps/backend/src/data-marts/data-storage-types/facades/data-mart-definition-validator-facade.service.ts
@@ -24,7 +24,7 @@ export class DataMartDefinitionValidatorFacade {
     if (!definition) {
       return ValidationResult.authoredFailure(
         DataMartValidationCode.DEFINITION_NOT_FOUND,
-        'Data Mart definition not found'
+        'DataMart definition not found'
       );
     }
 
@@ -32,14 +32,14 @@ export class DataMartDefinitionValidatorFacade {
     if (!config) {
       return ValidationResult.authoredFailure(
         DataMartValidationCode.STORAGE_CONFIG_NOT_FOUND,
-        'Data Mart storage config not found'
+        'DataMart storage config not found'
       );
     }
 
     if (!dataMart.storage.credentialId) {
       return ValidationResult.authoredFailure(
         DataMartValidationCode.STORAGE_CREDENTIALS_NOT_FOUND,
-        'Data Mart storage credentials not found'
+        'DataMart storage credentials not found'
       );
     }
 
--- a/apps/backend/src/data-marts/data-storage-types/interfaces/data-mart-validator.interface.ts
+++ b/apps/backend/src/data-marts/data-storage-types/interfaces/data-mart-validator.interface.ts
@@ -12,23 +12,53 @@ export interface DataMartValidator {
   ): Promise<ValidationResult>;
 }
 
+/**
+ * Stable identifiers for validation failures whose message this codebase
+ * authored, and which are therefore safe to show end users.
+ *
+ * Consumers that sanitize error text (see PublishDataStorageDraftsService)
+ * allowlist these codes. A failure with no code may carry raw driver output —
+ * a warehouse dry-run error, a Databricks EXPLAIN message — and must not be
+ * echoed back to the browser.
+ */
+export enum DataMartValidationCode {
+  INVALID_IDENTIFIER_FORMAT = 'DATA_MART_INVALID_IDENTIFIER_FORMAT',
+  DEFINITION_NOT_FOUND = 'DATA_MART_DEFINITION_NOT_FOUND',
+  STORAGE_CONFIG_NOT_FOUND = 'DATA_MART_STORAGE_CONFIG_NOT_FOUND',
+  STORAGE_CREDENTIALS_NOT_FOUND = 'DATA_MART_STORAGE_CREDENTIALS_NOT_FOUND',
+}
+
 export class ValidationResult {
   constructor(
     public readonly valid: boolean,
     public readonly errorMessage?: string,
     public readonly details?: Record<string, unknown>,
-    public readonly reason?: Record<string, unknown>
+    public readonly reason?: Record<string, unknown>,
+    public readonly code?: DataMartValidationCode
   ) {}
 
   static success(details?: Record<string, unknown>): ValidationResult {
     return new ValidationResult(true, undefined, details);
   }
 
+  /**
+   * Failure carrying a message of unknown provenance — it may come from a
+   * storage driver. Left uncoded so sanitizing consumers replace it.
+   */
   static failure(
     errorMessage?: string,
     details?: Record<string, unknown>,
     reason?: Record<string, unknown>
   ): ValidationResult {
     return new ValidationResult(false, errorMessage, details, reason);
+  }
+
+  /** Failure whose message this codebase authored; safe to show end users. */
+  static authoredFailure(
+    code: DataMartValidationCode,
+    errorMessage: string,
+    details?: Record<string, unknown>
+  ): ValidationResult {
+    return new ValidationResult(false, errorMessage, details, undefined, code);
   }
 }
--- a/apps/backend/src/data-marts/use-cases/publish-data-mart.service.ts
+++ b/apps/backend/src/data-marts/use-cases/publish-data-mart.service.ts
@@ -24,9 +24,18 @@ import { SearchableEntityType } from '../../common/search/search.facade';
  * PublishDataStorageDraftsService.toUserFacingReason.
  */
 export const PUBLISH_DATA_MART_ERRORS = {
-  NO_PERMISSION: 'You do not have permission to publish this Data Mart',
-  ALREADY_PUBLISHED: 'Data Mart is already published',
-  NO_DEFINITION: 'Data Mart has no definition',
+  NO_PERMISSION: {
+    code: 'DATA_MART_PUBLISH_FORBIDDEN',
+    message: 'You do not have permission to publish this Data Mart',
+  },
+  ALREADY_PUBLISHED: {
+    code: 'DATA_MART_ALREADY_PUBLISHED',
+    message: 'Data Mart is already published',
+  },
+  NO_DEFINITION: {
+    code: 'DATA_MART_NO_DEFINITION',
+    message: 'Data Mart has no definition',
+  },
 } as const;
 
 @Injectable()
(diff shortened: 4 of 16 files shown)


apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  it.each(validators)('%s codes its invalid-identifier failure', async (_name, validator) => {
+    const result = await validator.validate(
+      { fullyQualifiedName: 'this is !! not a valid identifier' } as never,
+      {} as never,
+      {} as never
+    );
+
+    expect(result.valid).toBe(false);
+    expect(result.errorMessage).toContain('Invalid identifier format');
```
laskevych, line 48: 🔴 **Blocking:** this spec fails on the `redshift` case. `RedshiftDataMartValidator.validate` checks `isRedshiftConfig(config)` and `isRedshiftCredentials(credentials)` **before** `validateIdentifiers` (`redshift-datamart.validator.ts:37`), so the stubbed `{}` config short-circuits the identifier check:

```
Expected substring: "Invalid identifier format"
Received string:    "Incompatible data storage config"

Tests: 1 failed, 4 passed, 5 total
```

`RedshiftConfigSchema` is a discriminated union on `connectionType`, so `{}` can never satisfy it. The docblock above states the opposite invariant ("Each validator checks identifiers before touching credentials or an adapter, so stub dependencies are never used") — which four of the five validators do hold.

Worth noting the runtime consequence, not just the red build: on a Redshift storage that is not fully configured, a malformed table name reports `Incompatible data storage config`, which is uncoded and therefore genericized by the publish path — so the actionable message this PR exists to deliver never arrives for Redshift. Moving `validateIdentifiers` above the two guards fixes both the test and the behaviour; relaxing the assertion instead would leave Redshift outside the parity this PR sets out to establish.

apps/backend/src/data-marts/data-storage-types/facades/data-mart-definition-validator-facade.service.ts
```diff
@@ -18,16 +22,25 @@ export class DataMartDefinitionValidatorFacade {
   async validate(dataMart: DataMart): Promise<ValidationResult> {
     const definition = dataMart.definition;
     if (!definition) {
-      return new ValidationResult(false, 'DataMart definition not found');
+      return ValidationResult.authoredFailure(
+        DataMartValidationCode.DEFINITION_NOT_FOUND,
+        'Data Mart definition not found'
```
laskevych: 🔴 **Blocking:** this message was reworded from `DataMart definition not found` to `Data Mart definition not found`, but `apps/backend/test/validation-schema.e2e-spec.ts:121` still asserts the old spelling and is not among this PR's changed files:

```ts
expect(res.body.errorMessage).toContain('DataMart definition not found');
```

The added space makes `toContain` fail, and `.github/workflows/e2e-api.yml` runs `npm run test:e2e -w @owox/backend`, so the VALID-05 case goes red. Either update the assertion or keep the original wording.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -12,33 +13,66 @@ import { SchemaActualizeTriggerService } from '../services/schema-actualize-trig
...
-/** Reasons the publish path authors itself; anything else is not shown to users. */
-const USER_FACING_FAILURE_REASONS: ReadonlySet<string> = new Set(
-  Object.values(PUBLISH_DATA_MART_ERRORS)
-);
+/**
+ * Failure codes whose message this codebase authored. Sanitization keys on the
+ * code rather than the text, so an authored message is safe because it is
+ * tagged as ours — not because its bytes match a list.
+ */
+const USER_FACING_FAILURE_CODES: ReadonlySet<string> = new Set<string>([
```
laskevych, line 28: 🟡 **Important:** spreading `Object.values(DataMartValidationCode)` wholesale is the inverse of the reasoning applied one layer up. `PUBLISH_DRAFTS_TRIGGER_ERROR_CODES` is an explicit membership set precisely because "a code alone only proves some thrower opted in" — but at the per-draft level, any member added to that enum, in another module, becomes user-facing here with no change to this file and no failing test.

Given the PR deliberately leaves the dry-run `catch` blocks and Databricks' `EXPLAIN` output uncoded, a future `DataMartValidationCode.DRY_RUN_FAILED` would reopen the leak silently — the same "safe but conventional rather than structural" gap the PR describes fixing. Listing the codes explicitly, or splitting the enum into safe-to-echo vs internal, would make this checkpoint as structural as the trigger-level one.

apps/backend/src/data-marts/use-cases/publish-data-mart.service.ts
```diff
@@ -24,9 +24,18 @@ import { SearchableEntityType } from '../../common/search/search.facade';
  * PublishDataStorageDraftsService.toUserFacingReason.
  */
 export const PUBLISH_DATA_MART_ERRORS = {
-  NO_PERMISSION: 'You do not have permission to publish this Data Mart',
-  ALREADY_PUBLISHED: 'Data Mart is already published',
-  NO_DEFINITION: 'Data Mart has no definition',
+  NO_PERMISSION: {
+    code: 'DATA_MART_PUBLISH_FORBIDDEN',
```
laskevych, line 28: 🟡 **Important:** this code is declared but never attached to anything — `grep -rn DATA_MART_PUBLISH_FORBIDDEN` returns only this line, since the throw site uses `new ForbiddenException(...message)`. It still enters `USER_FACING_FAILURE_CODES` via `Object.values(PUBLISH_DATA_MART_ERRORS).map(error => error.code)`, as an entry that can never match.

That breaks the invariant the allowlist is built on (every entry is thrown with its code), and it is an easy trap: someone reading `PUBLISH_DATA_MART_ERRORS` will reasonably assume permission denials are covered by the code allowlist and drop the message-identity fallback in `toUserFacingReason`, silently genericizing the permission reason again. Either drop the field, or make it real — see the note on `toUserFacingReason`.

laskevych (commented):
Reviewed the code-based sanitization end to end. The core idea holds up well: keying both checkpoints on `BaseException.code` instead of message bytes, keeping `failure()` uncoded as the safe default, and scoping two separate allowlists to what can legitimately reach each level. The reasoning recorded at the `NO_PERMISSION` call site about `BaseExceptionFilter` hardcoding 400 is the right call.

Two blocking items, both CI-red rather than design problems:

1. The new parity spec fails on the `redshift` case (`Tests: 1 failed, 4 passed`) — Redshift validates config/credentials before identifiers.
2. `apps/backend/test/validation-schema.e2e-spec.ts:121` still asserts the pre-rename `DataMart definition not found`, and `e2e-api.yml` runs that suite.

Details inline, plus a few non-blocking notes.

---

🔵 **Question** (no diff line to anchor to — `apps/backend/src/data-marts/mappers/data-mart.mapper.ts:510`)

Was leaving `code` off the `validate-definition` response intentional? `toDefinitionValidationResponse` maps `valid` / `errorMessage` / `reason` / `details` but not the new field, so the HTTP consumer there still cannot distinguish authored text from driver output — and that is the endpoint where raw dry-run output reaches the UI today. Fine if `code` is meant as an internal marker for the publish path only; just flagging that "so consumers can distinguish authored text from driver output" reads broader than what ships.

---

Everything else checked out: the other three touched specs pass (27 tests), the new code set is a strict superset of the old string set, and the pre-loop path has no `BusinessViolationException` whose message was surfaced before and is genericized now — `ValidateDataStorageAccessService` is called without `userId`, so its `ForbiddenException` branch is unreachable from here.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -12,33 +13,66 @@ import { SchemaActualizeTriggerService } from '../services/schema-actualize-trig
...
- * raw text from credential resolution or a warehouse driver, which must not
- * reach the browser.
+ * A storage ValidationResult only carries a `code` when this codebase authored
+ * the message (`unconfigured`, `oauthReauthRequired`). A bare `failure()` can
+ * hold raw text from credential resolution or a warehouse driver, which must
+ * not reach the browser.
  */
-function toUserFacingStorageError(result: ValidationResult): string {
-  return result.code && result.errorMessage ? result.errorMessage : STORAGE_ACCESS_FAILED_ERROR;
+function toStorageAccessError(result: ValidationResult): BusinessViolationException {
```
laskevych, line 86: 🟡 **Important:** this keeps the authored message but overwrites its code, so `code` and `message` can describe different failures. For `ValidationResult.oauthReauthRequired(...)` the result is:

- `message`: "Google authorization could not be refreshed. Reconnect this Storage to restore access."
- `code`: `PUBLISH_DRAFTS_STORAGE_ACCESS`, whose own message is "Check its connection settings and try again."

`BaseException` documents `code` as the thing clients "branch on instead of matching human-readable messages", so that pairing is actively misleading — a client branching on the code tells the user to check connection settings when the fix is reconnecting Google.

There is no HTTP consumer today (this service is reached only through the trigger handler), so it is latent rather than live. But if discarding `result.code` is deliberate, the docblock should say so — the current shape reads like propagation.

apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Every warehouse validator must tag its identifier-format failure with a code.
+ * Without one the publish path cannot tell the authored message from raw driver
+ * output, so it replaces it with a generic reason — which is how Athena drifted
+ * out of parity with the other four.
+ *
+ * Each validator checks identifiers before touching credentials or an adapter,
+ * so stub dependencies are never used.
+ */
+describe('Data Mart validators — identifier failure parity', () => {
+  const validators: [string, DataMartValidator][] = [
```
laskevych, line 27: 💡 **Suggestion:** the description says this spec means "a newly added warehouse is caught automatically", but the list is hardcoded and already misses the sixth registered validator: `validatorProviders` (`data-storage-providers.ts:190`) includes `LegacyBigQueryDataMartValidator` (`LEGACY_GOOGLE_BIGQUERY`).

It inherits `BigQueryDataMartValidator.validate` unchanged, so behaviour is correct today — but an override would ship unnoticed, and a seventh warehouse tomorrow passes this suite untouched. Deriving the cases from `DataStorageType` (and asserting every member is covered) would make the claim true, which is what makes this test worth its runtime.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -193,7 +229,20 @@ export class PublishDataStorageDraftsService {
...
+    if (error instanceof BusinessViolationException && error.code) {
+      return USER_FACING_FAILURE_CODES.has(error.code) ? error.message : GENERIC_FAILURE_REASON;
+    }
+
+    // NO_PERMISSION travels as a Nest ForbiddenException so the single-publish
+    // endpoint keeps its 403; HttpException carries no `code`, so this one
+    // authored message is matched by identity instead.
+    if (
+      error instanceof ForbiddenException &&
+      error.message === PUBLISH_DATA_MART_ERRORS.NO_PERMISSION.message
```
laskevych: 💡 **Suggestion:** this is the one comparison still matching message bytes, and it is avoidable without touching the status code.

A `class PublishForbiddenException extends ForbiddenException` carrying `code` keeps `getStatus() === 403` — `BaseExceptionFilter` is `@Catch(BusinessViolationException)`, so the hardcoded-400 problem the description rules out does not apply to a `ForbiddenException` subclass. The sanitizer then matches by `instanceof`, `NO_PERMISSION.code` stops being dead, and rewording the message can no longer silently downgrade permission denials — today the spec asserts against the same constant, so it would not catch that.

apps/backend/src/data-marts/data-storage-types/interfaces/data-mart-validator.interface.ts
```diff
@@ -12,23 +12,53 @@ export interface DataMartValidator {
...
+ * authored, and which are therefore safe to show end users.
+ *
+ * Consumers that sanitize error text (see PublishDataStorageDraftsService)
+ * allowlist these codes. A failure with no code may carry raw driver output —
+ * a warehouse dry-run error, a Databricks EXPLAIN message — and must not be
+ * echoed back to the browser.
+ */
+export enum DataMartValidationCode {
+  INVALID_IDENTIFIER_FORMAT = 'DATA_MART_INVALID_IDENTIFIER_FORMAT',
+  DEFINITION_NOT_FOUND = 'DATA_MART_DEFINITION_NOT_FOUND',
```
laskevych, line 26: 💡 **Suggestion:** `DATA_MART_DEFINITION_NOT_FOUND` and `PUBLISH_DATA_MART_ERRORS.NO_DEFINITION.code` (`DATA_MART_NO_DEFINITION`) are two codes for one condition, with two different messages.

`PublishDataMartService` checks `!dataMart.definition` first and throws its own, so this one is unreachable from the publish path and surfaces only on `validate-definition`. Since codes are becoming the client-facing contract here, a client has to handle both to cover one user-visible situation, and the two messages will drift. One code with one message would age better.

apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { AthenaDataMartValidator } from './athena/services/athena-datamart.validator';
```
laskevych, line 1: 💡 **Suggestion:** importing all five validators pulls in the full warehouse SDK graph. The spec takes ~23s for five assertions, and both new specs emit:

```
ReferenceError: You are trying to `require` a file after the Jest environment has been torn down
  at getInstanceMetadataRegion (@smithy/core/.../config/index.js:458)
  at hasAwsIdentity (snowflake-sdk/lib/telemetry/platform_detection.ts:95)
```

The imports kick off a real IMDS region lookup that outlives the test. That is new noise on every backend run plus an open-handle flake risk, from specs that only need to inspect a `ValidationResult`.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
```diff
@@ -12,33 +13,66 @@ import { SchemaActualizeTriggerService } from '../services/schema-actualize-trig
...
+function toStorageAccessError(result: ValidationResult): BusinessViolationException {
+  const authored = Boolean(result.code && result.errorMessage);
+  return new BusinessViolationException(
+    authored ? result.errorMessage! : PUBLISH_DRAFTS_ERRORS.STORAGE_ACCESS.message,
+    undefined,
+    PUBLISH_DRAFTS_ERRORS.STORAGE_ACCESS.code
+  );
+}
+
+/** Raises a trigger-level failure with its stable code attached. */
```
laskevych: 💡 **Nit:** "Raises" — the function returns the exception; callers do the throwing. Given what the guards below protect (refusing to publish drafts with unresolved roles), a caller who follows the comment and omits `throw` gets a silent no-op guard. "Builds" would match the code.

apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
prokopetsv: Confirmed and fixed. `RedshiftDataMartValidator` did check `isRedshiftConfig`/`isRedshiftCredentials` before `validateIdentifiers`, so the stubbed `{}` short-circuited. My docblock asserted an invariant I had verified on four validators and generalised to the fifth.

Took the reorder rather than relaxing the assertion, for the runtime reason you gave: on an unconfigured Redshift storage a malformed table name reported `Incompatible data storage config`, which is uncoded and therefore genericized by the publish path, so the actionable message never arrived. `validateIdentifiers` only inspects the definition, so moving it first is safe and matches the other four.

Docblock corrected to record that Redshift was the exception until it was reordered.

apps/backend/src/data-marts/data-storage-types/facades/data-mart-definition-validator-facade.service.ts
prokopetsv: Confirmed. Restored the original `DataMart ...` spelling on all three facade messages rather than editing the assertion.

The rename was incidental scope creep in a PR about codes, and reverting keeps both the API contract and the e2e suite untouched. A "DataMart" to "Data Mart" wording pass across these messages is worth doing, but as its own change with its own justification.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: Agreed, and you are right that it inverts the reasoning applied one layer up.

The four codes are now listed explicitly, with a comment recording why: a code added to that enum in another module must not become user-facing here without a deliberate edit. `DRY_RUN_FAILED` is exactly the scenario the spread would have let through silently, which is the gap this PR claims to close.

apps/backend/src/data-marts/use-cases/publish-data-mart.service.ts
prokopetsv: Both this and the `toUserFacingReason` suggestion are taken, and the latter resolves this one.

Added `PublishForbiddenException extends ForbiddenException` carrying the code. As you noted, `BaseExceptionFilter` is `@Catch(BusinessViolationException)`, so its hardcoded 400 never reaches a `ForbiddenException` subclass and the endpoint keeps its 403. `NO_PERMISSION.code` is now attached at the throw site, so the allowlist entry can actually match and the invariant holds again.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: Agreed — the OAuth example makes it concrete: a client branching on `PUBLISH_DRAFTS_STORAGE_ACCESS` would tell the user to check connection settings when the fix is reconnecting Google.

Added an explicit `STORAGE_VALIDATION_CODES` map from each storage-access validation code to its own trigger-level code (`PUBLISH_DRAFTS_STORAGE_UNCONFIGURED`, `PUBLISH_DRAFTS_STORAGE_OAUTH_REAUTH_REQUIRED`), so an authored message now travels with a code describing the same failure. Both are added to the trigger allowlist.

Chose an explicit map over deriving codes from the enum, to stay consistent with the reasoning in the enum-spread thread.

apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
prokopetsv: Fixed. Added `LegacyBigQueryDataMartValidator`, and rather than relying on someone remembering, a case asserting the list length matches `validatorProviders` — so a seventh warehouse fails this suite instead of passing untouched. That required exporting `validatorProviders`.

You are right that the claim was doing work the test did not back up; it does now.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: Taken — this was the sharpest suggestion in the set.

`PublishForbiddenException extends ForbiddenException` keeps `getStatus() === 403`, so the hardcoded-400 concern genuinely does not apply, and `toUserFacingReason` now matches by type and code instead of message bytes.

Your point about the spec asserting against the same constant was the part that convinced me: the old shape could not distinguish a real permission denial from any exception repeating the wording. Added a test that a bare `ForbiddenException` with identical text is **not** trusted — it would have passed under the old comparison.

apps/backend/src/data-marts/data-storage-types/interfaces/data-mart-validator.interface.ts
prokopetsv: Agreed in principle — two codes with two messages for one condition will drift, and making clients handle both to cover one user-visible situation is the wrong contract.

Not changing it here: collapsing them means choosing one message and altering what `validate-definition` returns, which I would rather do deliberately than fold into a PR about sanitization. Filing it as a follow-up.

apps/backend/src/data-marts/data-storage-types/datamart-validator-identifier-parity.spec.ts
prokopetsv: Confirmed — I see the same locally: ~23s for the parity spec and the same `getInstanceMetadataRegion` / `hasAwsIdentity` teardown error from both new specs.

The real fix is extracting the pure sanitization logic so these specs stop pulling the warehouse SDK graph, which would also make them runnable in about a second instead of twenty-three. That is a genuine piece of work rather than a tweak, so I am tracking it separately rather than bolting it on here.

apps/backend/src/data-marts/use-cases/publish-data-storage-drafts.service.ts
prokopetsv: Fixed: "Builds a trigger-level failure with its stable code attached; the caller throws."

The consequence you describe is real — a caller following the old wording and omitting `throw` would turn a guard that refuses to publish with unresolved roles into a silent no-op.

prokopetsv:
Answering the 🔵 question about `data-mart.mapper.ts:510`, since it has no diff line to thread onto.

Yes, deliberate — and you are right that my wording oversold it. `code` is currently an internal marker for the publish path, and adding it to `toDefinitionValidationResponse` would be a public API addition rather than an internal refactor, so I did not want to make it implicitly in a PR about sanitization.

I have reworded the PR description from "so consumers can distinguish authored text from driver output" to say the publish path specifically, so the claim matches what ships.

Worth doing properly though: as you note, `validate-definition` is the endpoint where raw dry-run output reaches the UI today, so it is arguably where `code` earns the most. Filing it alongside the duplicate-code cleanup.

---

Everything else from the review is addressed and pushed. Two follow-on defects surfaced while fixing, neither in the review: the parity spec imported `validatorProviders` before it was exported, and `STORAGE_VALIDATION_CODES` was declared after its first use. Both caught by typecheck.

The `PublishForbiddenException` change also invalidated two spec sites that threw a bare `ForbiddenException` — migrated, plus a new test asserting a bare one is not trusted.

Three items tracked as follow-ups rather than changed here: the duplicate definition-not-found codes, the SDK-graph import weight in the new specs, and exposing `code` on `validate-definition`.

laskevych (approved):

amarchenk0 (approved):

PR #1527, opened 2026-08-12
feat(data-marts): measure Data Last Updated Date for AWS Athena storage

laskevych:
## What

Data Last Updated Date is now measured for AWS Athena Data Marts, on every surface where it already works for BigQuery and Redshift: the Data Mart page tile, the Data Marts list, the canvas (badge + Actions sweep), MCP `query_data_mart` responses, and run history. No changes outside the storage layer — the new resolver registers into the existing `SourceDataLastUpdatedResolver` family and everything above it picks Athena up automatically.

## How

- **Source discovery** — `EXPLAIN (TYPE IO, FORMAT JSON)` over the composed SQL. Unlike the text-plan parsing the Redshift resolver needs, Athena's IO plan is a structured input-table list (catalog/schema/table) with views already expanded by the planner; the parser deep-walks the JSON so engine-version differences in the surrounding structure don't break it, and an unrecognised shape degrades to `unavailable` with the raw plan logged. Positional parameters are inlined via the existing `inlineAthenaPositionalParams` — byte-for-byte what Athena would bind as literals anyway — so nothing depends on parameterized EXPLAIN support. EXPLAIN scans no data, so discovery is free.
- **Per-format modification times** — table kind is decided from Athena's own catalog API (`GetTableMetadataCommand`; no Glue SDK dependency added):
  - **Iceberg** tables report the exact time of the last data commit via `max(committed_at)` over their `$snapshots` metadata. All Iceberg tables of one lookup are measured in a single UNION ALL query, since every Athena statement costs a full submit-and-poll round trip. Only exact times can produce `complete` coverage.
  - **Hive** tables have no data-change time anywhere in the catalog, so they report the catalog's metadata time (`transient_lastDdlTime`, falling back to creation time) with an explicit note — that time can drift either way from the last data change (a pure DDL touch moves it with no data written), so such sources always cap coverage at `partial`. The precise alternative (walking S3 object timestamps under the table location) is deliberately not done: its cost scales with object count.
  - Federated catalogs, tables missing from the catalog, unexpanded views, and snapshot-less Iceberg tables degrade to unknown sources with notes.
- Per-table classifications and measurements are cached across a batch, so a canvas-wide sweep touching one table through twenty Data Marts pays for it once. The Athena polling chain gained optional `pollIntervalMs`/`signal` (defaults unchanged, interval clamped to ≥1ms) so metadata statements poll at 250ms and stop with the orchestrator's soft timeout.

## Tests

- 14 resolver specs mirroring the sibling suites: Iceberg complete/MAX, single-query batching of snapshot lookups, Hive metadata-time capping at partial, creation-time fallback, federated catalogs, empty snapshots, metadata-read failure isolation, unexpanded-view drop, empty IO plan, per-item failure isolation, batch-level caching, abort before and between items, non-Athena config.
- 6 specs for the IO-plan parser (documented shape, dedup, leading noise, unfamiliar nesting, empty, unparseable) and 5 for the timestamp conversions (varchar `timestamp with time zone` shapes, epoch-seconds parameters, non-UTC rejection).
- All 13 Athena suites pass (206 tests); ESLint, Prettier, and `tsc` clean against the baseline.

Docs updated (`data-last-updated.md`, `mcp.md`) with the Athena semantics: exact for Iceberg, catalog metadata time with a note for Hive, federated catalogs not measured.


apps/backend/src/data-marts/data-storage-types/athena/services/athena-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,380 @@
...
+
+    if (metadata.parameters['table_type']?.toUpperCase() === 'ICEBERG') {
+      return 'iceberg';
+    }
+
+    const metadataTime = this.hiveMetadataTime(metadata);
+    cache.set(key, {
+      entry: {
+        table: name,
+        dataLastUpdatedAt: metadataTime,
```
amarchenk0: Returning `transient_lastDdlTime`/`CreateTime` as `dataLastUpdatedAt` breaks the existing `partial` contract. The schema and MCP consumer treat this value as a lower bound and tell users the sources are “at least as recent as X.” However, a metadata-only change can make `metadataTime` newer than the last data change, causing stale data to be presented as recently updated. Please keep the Hive value null/unavailable, or first introduce separate approximation semantics and update every consumer.

amarchenk0 (changes requested):
The Hive timestamp path breaks the existing Data Last Updated contract and can overstate data recency. Please address the blocking inline comment before approval.

apps/backend/src/data-marts/data-storage-types/athena/services/athena-source-data-last-updated.resolver.ts
laskevych: Confirmed and fixed in ccb267d96. You're right that this inverted the `partial` contract — the schema and every consumer present the value as "at least as recent as", while a metadata-only touch can move `transient_lastDdlTime` forward with no data written, so stale data could read as fresh. Hive sources now resolve to a null timestamp with the note `Hive table — the catalog does not track data modification time`; only exact Iceberg commit times produce a value (a pure-Hive query reports `unavailable`, a mixed one `partial` with the MAX built from Iceberg times only). The approximation-semantics route (a separate precision marker carried through the schema, UI, MCP, and persistence) is deliberately left as a possible follow-up rather than squeezed into this PR. Docs, changeset, and specs updated accordingly.

apps/backend/src/data-marts/data-storage-types/athena/services/athena-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,370 @@
...
+    });
+    return 'settled';
+  }
+
+  /**
+   * One UNION ALL query answers `max(committed_at)` for every Iceberg table of this lookup —
+   * each Athena statement costs a full submit-and-poll round trip, so per-table queries would
+   * multiply latency for nothing.
+   */
+  private async measureIcebergTables(
```
prokopetsv, line 266: 🟡 Non-Blocker: `measureIcebergTables` runs one UNION ALL query for every Iceberg table in the lookup. If any single `$snapshots` read fails (a table dropped between EXPLAIN and measurement, a Lake Formation denial on one table), the whole query throws and every Iceberg table in the batch is cached as failed — degrading a whole canvas sweep to `partial`/`unavailable` for every Data Mart that shares any of those tables. Correctness holds (failed entries are null, never a wrong value), so this is availability granularity rather than a contract break, but a fallback to per-table queries when the batched query fails would keep one bad table from poisoning its neighbors.

apps/backend/src/data-marts/data-storage-types/athena/services/athena-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,370 @@
...
+        rows.map(row => [row[0] ?? '', athenaTimestampToIsoUtc(row[1])])
+      );
+
+      for (const ref of refs) {
+        const time = timeByTable.get(`${ref.schema}.${ref.table}`) ?? null;
+        cache.set(refKey(ref), {
+          entry: {
+            table: this.displayName(ref),
+            dataLastUpdatedAt: time,
+            ...(time === null ? { note: 'Iceberg table with no snapshots' } : {}),
```
prokopetsv: 🟡 Non-Blocker: When `row[1]` is non-null but `athenaTimestampToIsoUtc` rejects its shape (non-UTC zone suffix, or Athena changing its varchar rendering), the entry silently gets `note: 'Iceberg table with no snapshots'` — indistinguishable from a genuinely empty table, and nothing is logged. This contradicts the observability the PR applies to the IO-plan parser ("an unrecognised shape degrades to unavailable with the raw plan logged"). Coverage still degrades correctly, but a format drift here would never surface. Consider logging a warning and using a distinct note when a non-null value fails to parse.

docs/getting-started/setup-guide/data-last-updated.md
```diff
@@ -65,6 +65,7 @@ Storage-specific caveats:
 
 - **BigQuery** reports modification times immediately and exactly.
 - **Redshift** metadata can lag real writes by up to ~5 minutes, so a load that finished a moment ago may not show yet. Older Redshift releases do not report modification times at all — those Data Marts show Unknown. Source tables outside the connection's database, and tables in schemas whose names require quoting (mixed case, hyphens, non-ASCII characters), cannot be identified and appear as unknown entries with partial coverage.
+- **Athena** answers exactly for **Iceberg** tables (the time of the last data commit). Classic **Hive** tables show Unknown: the catalog stores no data-change time, and its metadata timestamps can be newer than the last data change, so reporting them would break the "at least as recent as" promise. Tables from federated catalogs (connectors other than the Glue Data Catalog) are not measured.
```
prokopetsv: 🟡 Non-Blocker: The Hive sentence here (~35 words) exceeds this repo's documented docs-style limit of 20 words/sentence for `docs/`. Markdownlint won't catch it, but it's a written project rule. Splitting into two sentences resolves it.

docs/getting-started/setup-guide/mcp.md
```diff
@@ -272,7 +272,7 @@ Read the value precisely. It is a **storage** timestamp, not a statement about t
...
+Coverage is best effort per storage. Google BigQuery, AWS Redshift, and AWS Athena are supported, and OWOX resolves views and SQL data marts through to their underlying base tables. Sharded and wildcard table sets collapse into one entry. Other storages currently report `unavailable`. `sources` deliberately omits views: a view's own modification time reflects a change to its definition, not to any data. Redshift metadata can lag real writes by up to ~5 minutes, and older Redshift releases do not report modification times at all; Athena is exact for Iceberg tables, while Hive tables report `null` with a note (their catalog stores no data-change time) — see [Data Last Updated](./data-last-updated.md) for storage-specific caveats.
```
prokopetsv: 🟡 Non-Blocker: Same docs-style note as data-last-updated.md — the amended sentence ("Coverage is best effort per storage... see Data Last Updated for storage-specific caveats.") is well over the 20-word/sentence limit for `docs/`. Consider breaking it up, particularly the new Athena/Redshift clause added at the end.

prokopetsv (commented):
Reviewed with the PR description and amarchenk0's blocking thread in mind — the Hive contract-break amarchenk0 flagged is confirmed fixed in ccb267d96 (Hive now resolves to null with a note; verified against the specs), so no further comment there. All CI checks are green. Three non-blocking findings below; one general note on the PR description itself (not tied to a diff line): the "How" section still describes the pre-fix Hive behavior (reporting transient_lastDdlTime/creation time capped at partial) — the exact semantics the blocking review rejected and the code no longer implements. Worth updating so the PR body (and any release notes generated from it) match what shipped.

prokopetsv (approved):

apps/backend/src/data-marts/data-storage-types/athena/services/athena-source-data-last-updated.resolver.ts
laskevych: Fixed in c5fd5bf95 — a failed batched query now retries per table (single-table batches keep the direct failure path), so one broken table degrades only itself: its entry gets `could not read Iceberg snapshots` while its neighbours still resolve. Covered by a spec where one of two tables throws and the other still reports its commit time, with the call count asserting batch-then-per-table order.

apps/backend/src/data-marts/data-storage-types/athena/services/athena-source-data-last-updated.resolver.ts
laskevych: Fixed in c5fd5bf95 — a non-null value that fails to parse is now logged (`Unrecognised Iceberg snapshot timestamp for <table>: <raw value>`) and carries its own note, `unrecognised snapshot timestamp format`, distinct from `Iceberg table with no snapshots`. Same observability treatment as the IO-plan parser. Spec added with a non-UTC-zone value.

docs/getting-started/setup-guide/data-last-updated.md
laskevych: Fixed in c5fd5bf95 — split into three short sentences (Unknown, why the metadata timestamps are not reported, federated catalogs).

docs/getting-started/setup-guide/mcp.md
laskevych: Fixed in c5fd5bf95 — the paragraph is now short sentences throughout, including the amended Redshift/Athena clauses at the end.

amarchenk0 (approved):
Confirmed the Hive freshness-contract fix at `e547b4d`: Hive sources now remain unknown, mixed Hive/Iceberg coverage preserves the lower-bound contract, and the regression tests and documentation match the behavior.

PR #1528, opened 2026-08-12
feat(web): enhance Data Mart publishing flow and report creation

artemfenoga:
Creating a report from a draft Data Mart no longer dead-ends at a disabled "New Report" button. Clicking it now takes the user straight to the next actionable step instead of forcing them to leave the page, find the publish button, and come back.

- If the Data Mart is ready to publish → a confirmation dialog offers **"Publish and create report"**, which publishes the Data Mart and opens the new report form in one action.
- If something is still missing (e.g. Storage or Input Source) → the dialog explains exactly what's missing and offers **"Review Data Setup"**, which navigates to the Data Setup tab.

Screenshot 2026-08-12 222335

### DIFF
--- a/apps/web/src/features/data-marts/reports/list/components/DestinationCard/AddReportButton.tsx
+++ b/apps/web/src/features/data-marts/reports/list/components/DestinationCard/AddReportButton.tsx
@@ -6,8 +6,8 @@ interface AddReportButtonProps {
 }
 
 /**
- * Add Report Button component with conditional rendering and tooltip
- * Only shows for Google Sheets destinations with proper validation
+ * Button that triggers report creation.
+ * Destination-type filtering and visibility are handled by the parent DestinationCard.
  */
 export function AddReportButton({ onAddReport }: AddReportButtonProps) {
   return (
--- a/apps/web/src/features/data-marts/reports/list/components/DestinationCard/AddReportButton.tsx
+++ b/apps/web/src/features/data-marts/reports/list/components/DestinationCard/AddReportButton.tsx
@@ -1,11 +1,7 @@
 import { Button } from '@owox/ui/components/button';
 import { PlusIcon } from 'lucide-react';
-import { Tooltip, TooltipContent, TooltipTrigger } from '@owox/ui/components/tooltip';
-import { DataMartStatus } from '../../../../shared/enums/data-mart-status.enum';
-import type { DataMartStatusInfo } from '../../../../shared/types/data-mart-status.model';
 
 interface AddReportButtonProps {
-  dataMartStatus?: DataMartStatusInfo;
   onAddReport: () => void;
 }
 
@@ -13,32 +9,18 @@ interface AddReportButtonProps {
  * Add Report Button component with conditional rendering and tooltip
  * Only shows for Google Sheets destinations with proper validation
  */
-export function AddReportButton({ dataMartStatus, onAddReport }: AddReportButtonProps) {
-  const isDisabled = dataMartStatus?.code === DataMartStatus.DRAFT;
-
+export function AddReportButton({ onAddReport }: AddReportButtonProps) {
   return (
-    <Tooltip>
-      <TooltipTrigger asChild>
-        <span>
-          <Button
-            onClick={onAddReport}
-            variant='outline'
-            size='sm'
-            aria-label='Add new report'
-            disabled={isDisabled}
-            data-testid='reportCreateButton'
-            className='text-foreground'
-          >
-            <PlusIcon className='text-foreground h-4 w-4' />
-            New Report
-          </Button>
-        </span>
-      </TooltipTrigger>
-      {isDisabled && (
-        <TooltipContent>
-          <p>To create a report, publish the Data Mart first</p>
-        </TooltipContent>
-      )}
-    </Tooltip>
+    <Button
+      onClick={onAddReport}
+      variant='outline'
+      size='sm'
+      aria-label='Add new report'
+      data-testid='reportCreateButton'
+      className='text-foreground'
+    >
+      <PlusIcon className='text-foreground h-4 w-4' />
+      New Report
+    </Button>
   );
 }
--- a/apps/web/src/features/data-marts/reports/list/components/DestinationCard/DestinationCard.tsx
+++ b/apps/web/src/features/data-marts/reports/list/components/DestinationCard/DestinationCard.tsx
@@ -153,7 +153,6 @@ export function DestinationCard({
             <ReportListRenderer
               destination={destination}
               onEditReport={handleEditReport}
-              dataMartStatus={dataMartStatus}
               onAddReport={handleAddReportRequest}
             />
           </CollapsibleCardContent>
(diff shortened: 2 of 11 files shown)


apps/web/src/features/data-marts/reports/list/components/DestinationCard/DestinationCard.tsx
```diff
@@ -65,6 +93,36 @@ export function DestinationCard({ destination, dataMartStatus }: DestinationCard
     reportsCount === 0 &&
     googleSheetsCount === 1;
 
+  const isDraft = dataMartStatus?.code === DataMartStatus.DRAFT;
+  const requiredSetupActions = getRequiredSetupActions(validationErrors);
+
+  const handleAddReportRequest = useCallback(() => {
```
romandubovyi, line 102: ⚪ **Non-blocking:** The new branching is not covered by any test: `handleAddReportRequest` has three paths (published → open the form, draft + `canPublish` → publish dialog, draft without `canPublish` → setup dialog) and `handlePublishAndCreateReport` must publish *then* open the form. There is no test file in this directory, and `apps/web/e2e/specs/report.spec.ts:146` only clicks `reportCreateButton` on an already-published Data Mart, so a regression here would not be caught. Sibling flows already have unit tests (`DataMartDetails.test.tsx`, `use-schema-unsaved-guard.test.tsx`); a small Vitest + RTL test for these branches would pin the feature down.

🤖 Generated with Grok CLI (Grok 4.6) + Claude Code (Opus 5)

apps/web/src/features/data-marts/reports/list/components/DestinationCard/DestinationCard.tsx
```diff
@@ -119,6 +176,41 @@ export function DestinationCard({ destination, dataMartStatus }: DestinationCard
...
+      <AlertDialog
+        open={dialog !== null}
+        onOpenChange={open => {
+          if (!open && !isPublishing) setDialog(null);
+        }}
+      >
+        <AlertDialogContent>
+          <AlertDialogHeader>
+            <AlertDialogTitle>
+              {dialog === 'publish'
```
romandubovyi: ⚪ **Non-blocking:** `dialog` drives both `open` and the dialog contents, so `setDialog(null)` flips the title, description and footer button to the `setup` variant while Radix is still playing the close animation (`AlertDialogContent` carries `data-[state=closed]:animate-out … duration-200`). After a successful publish — and on Cancel from the publish variant — the user briefly sees "Complete Data Mart setup first" / "Review Data Setup" as the dialog fades out. `SchemaUnsavedChangesDialog` avoids this by keeping the content state (`intent`) separate from `open`; rendering from the last non-null `dialog` value here would keep the closing frame consistent.

romandubovyi (commented):
Review of the draft → publish → create report flow. Four non-blocking comments: missing test coverage for the new branching, a dialog-content flip during the close animation, a stale doc comment, and wording in the setup dialog. Nothing blocking.

apps/web/src/features/data-marts/reports/list/components/DestinationCard/AddReportButton.tsx
```diff
@@ -1,44 +1,26 @@
...
-import { DataMartStatus } from '../../../../shared/enums/data-mart-status.enum';
-import type { DataMartStatusInfo } from '../../../../shared/types/data-mart-status.model';
 
 interface AddReportButtonProps {
-  dataMartStatus?: DataMartStatusInfo;
   onAddReport: () => void;
 }
 
 /**
  * Add Report Button component with conditional rendering and tooltip
```
romandubovyi: ⚪ **Non-blocking:** The doc comment still describes the tooltip and the "proper validation" behaviour this PR removed, and the destination-type filtering now lives in `DestinationCard` (`reportDestinationTypes`). Worth updating or dropping it so it does not describe behaviour the component no longer has.

🤖 Generated with Kimi CLI + Claude Code (Opus 5)

apps/web/src/features/data-marts/reports/list/components/DestinationCard/DestinationCard.tsx
```diff
@@ -119,6 +176,41 @@ export function DestinationCard({ destination, dataMartStatus }: DestinationCard
...
+                ? 'Publish Data Mart to create a report?'
+                : 'Complete Data Mart setup first'}
+            </AlertDialogTitle>
+            <AlertDialogDescription>
+              {dialog === 'publish'
+                ? 'Reports are available only for published Data Marts. We’ll open the new report form when publishing is complete.'
+                : requiredSetupActions.length > 0
+                  ? `Complete the required setup before publishing this Data Mart and creating a report: ${requiredSetupActions.join(', ')}.`
```
romandubovyi: ⚪ **Non-blocking:** `DATA_MART_REQUIRED_ACTIONS` values are written as list items ("complete storage configuration", "configure an input source") — that is how the publish tooltip renders them in `DataMartDetails.tsx`. Inlined into this sentence the result reads "Complete the required setup … : complete storage configuration, configure an input source.", duplicating the verb. Rendering them as a bulleted list under the description would match the tooltip and read better.

romandubovyi (approved):
Approving — no blocking issues found. The four comments above are non-blocking polish (test coverage, dialog close-animation frame, stale doc comment, wording) and can be addressed at the author's discretion.

apps/web/src/features/data-marts/reports/list/components/DestinationCard/DestinationCard.tsx
artemfenoga: Good catch, thank you! I’ve added unit coverage for each new report-creation path, including the publish-then-open-form flow. All tests are passing now.

apps/web/src/features/data-marts/reports/list/components/DestinationCard/DestinationCard.tsx
artemfenoga: Good catch, fixed — split it into isDialogOpen (controls open/close) and dialogContent (only set right before opening, never reset on close), same pattern as SchemaUnsavedChangesDialog. So during the close animation it now just holds the last shown variant instead of flipping to "setup".

apps/web/src/features/data-marts/reports/list/components/DestinationCard/AddReportButton.tsx
artemfenoga: Good catch — comment was stale, updated it to reflect what the component actually does now: AddReportButton.tsx:8-11. Filtering/visibility logic lives in DestinationCard (reportDestinationTypes), so I pointed the doc there instead of describing behaviour this PR removed.

apps/web/src/features/data-marts/reports/list/components/DestinationCard/DestinationCard.tsx
artemfenoga: Moved the actions out of the inline sentence. When there's just one (the common case, since only storage/input-source can be invalid here), it now reads inline: "Before creating a report, complete storage configuration." Only when there are 2+ does it switch to a bulleted list under the description — avoids the awkward single-bullet list.

PR #1529, opened 2026-08-12
feat: add managed Cloud licensing for Report Runs

kalnik-a-a:
# Managed Cloud licensing for Report Runs

## Summary

- unify project balance checks and consumption reporting behind binding-aware billing services
- require a valid managed Cloud license before every Report Run reads or processes data
- keep Process Runs operational, with consumption emitted only for the internal billing binding
- add origin-bound license key issuance, one-time reveal, rotation, revocation, and Cloud gateway validation
- add license key management to Project Settings with admin-only mutations and read-only member access
- keep all destination configuration available in Community while restricting only Report Run execution

## Why

Report Run authorization and consumption attribution must use the verified billing binding instead of deployment-supplied project identity. Deployments without a confirmed license remain usable for configuration and Process Runs, while Report Runs fail closed with an auditable status.

## Notable behavior

- authorization runs before schema actualization, cache preparation, reader creation, or sample extraction
- denied runs are recorded as `RESTRICTED`; unavailable per-run authorization is recorded as `FAILED`
- JWK rotation supports a serialized, throttled refresh when a previously unseen `kid` appears
- gateway requests validate kind-specific consumption envelopes while allowing additive fields for rolling updates
- consumption publication remains best-effort and cannot turn a successful run into a failure
- customer license keys are bound to one public origin, shown once, and expire after 365 days

## Validation

- backend: 546 suites, 7,565 tests
- web: 211 files, 1,695 tests
- internal helpers: 13 suites, 76 tests
- CLI: 52 tests
- production build: `npm run build -w owox`
- Chromium E2E: create, reveal, rename, and revoke a license key
- lint, type-check, formatting, Markdown lint, and `git diff --check`


apps/backend/src/common/config/app-edition-config.service.ts
```diff
@@ -21,40 +42,38 @@ const LICENSE_KEY_ISSUER = 'license@owox-registry.iam.gserviceaccount.com';
...
+        claims.jti,
+        claims.payload as LicenseKeyClaims | undefined
+      );
+      this.apply(AppEdition.CLOUD_BILLED_ENTERPRISE, licenseContext);
       if (verbose) {
         this.logger.log(`${this.activeEdition} App Edition activated based on license key.`);
       }
     } catch (error) {
-      this.setEdition(AppEdition.COMMUNITY);
+      this.apply(AppEdition.COMMUNITY, null);
```
romandubovyi, line 73: 🔴 **Blocking:** This `catch` treats a transient JWKS delivery failure the same as an invalid license. A single network blip against `googleapis.com` clears `licenseContext`, and `AppEditionLicenseRefresherService` only retries on the hourly cron, so the deployment stays unlicensed for up to an hour. Via `BindingAwareProjectBillingService.delegate()` that routes everything to `RestrictedProjectBillingService`: all Report Runs fail with `LICENSE_REQUIRED`, and on an `INTERNAL` binding Process Runs keep executing while their consumption is silently dropped (`RestrictedProjectBillingService.skip()`) — unbilled usage on OWOX's own Cloud. Please distinguish signature/claim rejection from key-fetch failure, and keep the previously verified `licenseContext` when only the fetch failed (or retry with backoff before degrading).

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Claude Code (Opus 5) + Grok CLI (Grok 4.6)

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  private skipProcessRun(kind: RunKind): void {
+    this.logger.debug(`${kind} is not billed through the license gateway, skipping...`);
+  }
+
+  private async sendConsumption(kind: RunKind, payload: Record<string, unknown>): Promise<void> {
+    try {
+      await this.callCloud('consumption', { kind, payload });
+    } catch (error) {
```
romandubovyi, line 113: ⚪ **Non-blocking:** Consumption delivery here has no guarantee in either direction. On the loss side, this `catch` swallows the failure after `fetchWithBackoff` exhausts its 3 attempts, so a gateway outage drops the record permanently — there's no queue, no retry, no persisted marker to reconcile later. On the duplication side, `fetchWithBackoff` retries on 5xx, 429 and the 25s timeout, but `POST /license/consumption` carries no idempotency key and `publishForwardedConsumption` publishes straight to PubSub — a request the gateway already processed but that timed out client-side gets billed twice. The payload already carries a stable `reportRunId` for the call; consider having the gateway deduplicate on it, and persisting failed sends for reconciliation.

🤖 Generated with Claude Code (Opus 5) + Grok CLI (Grok 4.6)

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
```diff
@@ -0,0 +1,156 @@
...
+  ) {
+    super();
+  }
+
+  public async verifyCanPerformOperations(projectId: string, runKind: RunKind): Promise<void> {
+    if (!isReportRun(runKind)) {
+      return;
+    }
+
+    const response = await this.callCloud('can-perform');
```
romandubovyi, line 46: ⚪ **Non-blocking:** This authorization call runs synchronously before every Report Run with no caching and no dedicated time budget. `callCloud` uses the `fetchWithBackoff` defaults (3 attempts × 25s timeout plus backoff), so a degraded gateway stalls each run for over a minute before it fails — visible to the user as a hanging run, and eating into the execution window for scheduled runs. Consider passing a shorter timeout/retry budget for the authorization path specifically, or caching an `allowed` decision for a short TTL the way `LicenseKeyGuard` already caches verified tokens.

romandubovyi (commented):
Multi-agent review of the managed Cloud licensing change. Four agents analyzed the PR independently; findings were merged, de-duplicated and re-verified against the PR head before posting.

Main concern is the fail-closed path in `AppEditionConfig.actualizeAppEdition` — a transient JWKS fetch failure is indistinguishable from an invalid license there, and the hourly refresh makes the degradation last. The other two comments cover delivery guarantees and latency on the license gateway calls.

apps/backend/src/data-marts/services/project-billing/internal-project-billing.service.ts
```diff
@@ -0,0 +1,295 @@
...
+
+  private async publish(
+    kind: RunKind,
+    command: Record<string, unknown>,
+    destinationType?: DataDestinationType
+  ): Promise<void> {
+    try {
+      const topic = this.resolveTopic(kind, destinationType);
+      if (!this.pubSubService || !topic) {
+        this.logger.debug(`${kind} consumption tracking is not configured, skipping...`);
```
laskevych: 🔴 **Blocking:** The gateway ACKs `/license/consumption` with 202 and can then silently drop the record. `publish()` treats a missing `CONSUMPTION_*_TOPIC` (or missing PubSub project config) as a debug-level "not configured, skipping", and its `catch` swallows PubSub failures — including `resolveTopic()`'s own throw for an unsupported destination type. Meanwhile `LicenseGatewayController`'s constructor validates only the balance integration, so a Cloud deployment with one mistyped topic var returns 202 for every forwarded consumption record while dropping all of them — and the self-managed sender rightly treats the 2xx as final and never resends. Unlike a transient PubSub blip (best-effort is fine there), a config gap is deterministic and permanent. Suggestion: extend the existing constructor guard to also require the consumption topic config when the gateway is mounted, and log dropped *forwarded* records at error level — the debug-level skip was written for the internal path where it's a legitimate opt-out.

apps/backend/src/ee/license-keys/guards/license-key.guard.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    try {
+      const claims = await verifyJwtClaims(token, LICENSE_KEY_ISSUER);
+      const expiresAtMs = typeof claims.exp === 'number' ? claims.exp * 1000 : 0;
+      this.verifiedTokens.set(token, {
+        claims,
+        until: Math.min(expiresAtMs, Date.now() + VERIFIED_TOKEN_TTL_MS),
+      });
+      return claims;
+    } catch {
```
laskevych: 🔴 **Blocking:** This complements the existing blocking comment on `AppEditionConfig.actualizeAppEdition` — the same transient-vs-invalid collapse exists on the Cloud side. The bare `catch` here turns any `verifyJwtClaims` failure into 401 "License key is not valid", and `ensureCertsCache` throws on a failed refresh even when a stale cert cache exists. So during a JWKS outage (or right after a pod restart, when `verifiedTokens` is empty), the gateway 401s valid customers; `LicenseProjectBillingService.callCloud` maps 401/403 to `LICENSE_REQUIRED`, so their Report Runs fail with "create a managed license key…" during a pure infrastructure incident, and their concurrently sent consumption records are dropped too. Whatever fix distinguishes signature/claim rejection from key-fetch failure should cover this call site as well — return 503 (retryable) for fetch failures, and consider falling back to the stale cert cache instead of throwing.

apps/backend/src/common/config/public-origin.service.ts
```diff
@@ -61,6 +61,26 @@ export class PublicOriginService {
...
+   * @param value Raw origin (scheme optional)
+   * @returns Canonical origin string
+   * @throws Error if the value is not a valid http/https URL
+   */
+  normalizeOrigin(value: string): string {
+    const trimmed = value.trim();
+    // Only a missing scheme may be defaulted; `ensureHttp` would otherwise turn
+    // `ftp://host` into `http://ftp` instead of rejecting it.
+    const hasScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(trimmed);
+    const withProtocol = hasScheme ? trimmed : this.ensureHttp(trimmed);
```
laskevych, line 77: 🟡 **Important (non-blocking):** Origin normalization silently defaults a scheme-less value to `http://` (here and in `getPublicOrigin`'s fallback), while key issuance requires an explicit protocol (`@IsUrl({require_protocol: true})`, and the UI suggests `https://…`). An HTTPS deployment with `PUBLIC_ORIGIN=data-marts.example.com` — valid pre-PR, everything else works — computes audience `http://data-marts.example.com`, so every issuable key fails `aud` verification and the deployment silently falls back to COMMUNITY on each hourly refresh, with all Report Runs finishing RESTRICTED. The jwt error log does reveal the expected audience, but the root cause (silent scheme defaulting) is far from the symptom. A warning log when the scheme is defaulted — or refusing to default when a license key is configured — would make this diagnosable in minutes instead of a support ticket.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
```diff
@@ -103,8 +104,12 @@ export class LookerStudioConnectorApiService {
    * @returns Schema response with field definitions
    */
   public async getSchema(request: GetSchemaRequest): Promise<GetSchemaResponse> {
-    // Get report and cached reader centrally
-    const { report, cachedReader } = await this.getReportAndCachedReader(request);
+    const report = await this.getReport(request);
+    await this.projectBillingService.verifyCanPerformOperations(
```
laskevych: 🟡 **Important (non-blocking):** `getSchema` now awaits `verifyCanPerformOperations` — an uncached remote round-trip (Balance API on INTERNAL, app.owox.com with `fetchWithBackoff` defaults on LICENSE) — where the pre-PR code gated only the data-extraction paths. Looker Studio issues schema requests on every report open/edit and they bill nothing, so this is added latency and remote load per call; during a billing-endpoint brownout each schema fetch can stall for up to ~75s of retries, making dashboard editing crawl. If gating the schema is intentional fail-closed semantics, a short-TTL cached decision (the way `LicenseKeyGuard` caches verified tokens) would keep the semantics without the per-call round-trip.

apps/backend/src/data-marts/services/project-billing/binding-aware-project-billing.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    private readonly license: LicenseProjectBillingService,
+    private readonly restricted: RestrictedProjectBillingService
+  ) {
+    super();
+  }
+
+  private delegate(): ProjectBillingService {
+    const binding = this.appEditionConfig.getLicenseContext()?.binding;
+    if (binding === ProjectBinding.INTERNAL) {
+      return this.internal.isBalanceConfigured() ? this.internal : this.restricted;
```
laskevych, line 27: 🟡 **Important (non-blocking):** The rule "INTERNAL binding requires balance config" is encoded twice with different outcomes: `projectBillingProvider` throws a loud, actionable bootstrap error, while `delegate()` silently downgrades the same condition to `RestrictedProjectBillingService`. The runtime path is reachable — boot with an absent/invalid `LICENSE_KEY` (the factory sees no binding and skips its guard), then the hourly refresh validates an INTERNAL-binding key on a deployment missing `BALANCE_ENDPOINT_*` vars — and then every Report Run is rejected with the misleading `LICENSE_REQUIRED` message even though a valid license *is* configured. Consider keeping the rule in one place, e.g. route INTERNAL to `internal` unconditionally and let it fail with its own descriptive config error.

apps/backend/src/ee/license-keys/guards/license-key.guard.ts
```diff
@@ -0,0 +1,92 @@
...
+    request.licenseKey = record;
+    return true;
+  }
+
+  private async verify(token: string): Promise<JwtPayload> {
+    const cached = this.verifiedTokens.get(token);
+    if (cached && cached.until > Date.now()) {
+      return cached.claims;
+    }
+    this.verifiedTokens.delete(token);
```
laskevych, line 88: 💡 **Suggestion (non-blocking):** `verifiedTokens` only evicts an expired entry when the exact same token string is presented again, and entries are keyed by the full multi-KB JWT. Keys that rotate per the documented flow leave their old entries in the map forever, so this singleton slowly leaks across months of uptime. Keying by `jti` (so a rotated key replaces its predecessor) or sweeping expired entries on insert bounds it cheaply.

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
```diff
@@ -0,0 +1,156 @@
...
+      RunKind.MCP_QUERY_RUN,
+      this.mcpQueryConsumptionPayload(dataMart, runId)
+    );
+  }
+
+  private skipProcessRun(kind: RunKind): void {
+    this.logger.debug(`${kind} is not billed through the license gateway, skipping...`);
+  }
+
+  private async sendConsumption(kind: RunKind, payload: Record<string, unknown>): Promise<void> {
```
laskevych, line 110: 💡 **Suggestion (non-blocking):** Extending the existing latency comment to the consumption path: `sendConsumption` awaits `callCloud` with the default 3×25s budget inline in run finalization, yet its failure is only logged. During a gateway brownout every finishing run hangs ~75s holding its graceful-shutdown registration. Since the semantics are already fire-and-forget, not awaiting the call (or passing a short timeout budget) gives identical behavior without stalling run throughput.

apps/backend/src/data-marts/use-cases/run-report.service.ts
```diff
@@ -386,10 +384,11 @@ export class RunReportService {
 
     try {
       this.gracefulShutdownService.registerActiveProcess(processId);
-      const accessor = await this.resolveAccessor(runByUserId, reportRun.getDataMart().projectId);
-      this.availableDestinationTypesService.verifyIsAllowed(
-        reportRun.getReport().dataDestination.type
+      await this.projectBillingService.verifyCanPerformOperations(
```
laskevych, line 387: 💡 **Suggestion (non-blocking):** Billing enforcement is a manual per-call-site convention — `verifyCanPerformOperations` is hand-inserted at ~11 sites across 7 files (4 in the Looker service alone). Nothing structural ensures a future run path performs the check, so the next MCP tool or export endpoint can compile clean, pass tests, and execute unbilled Report Runs. Not asking to restructure in this PR, but a choke point (e.g. in the run-recording layer every path already goes through) or at least a lint/test convention would protect the revenue path this PR creates.

apps/web/src/features/license-keys/components/EditLicenseKeySheet.tsx
```diff
@@ -0,0 +1,267 @@
...
+  'The license key is only shown once, right after creation. If you no longer have it, create a new key and revoke this one.';
+
+interface EditLicenseKeySheetProps {
+  licenseKey: LicenseKey | null;
+  onClose: () => void;
+  onUpdated: () => void;
+  onRevoke: (key: LicenseKey) => void;
+}
+
+function MetadataItem({
```
laskevych, line 50: 💡 **Suggestion (non-blocking):** The shared-extraction this PR started (`SecretRevealDialog`, `ExpirationValue` → `shared/`) stops one step short: `MetadataItem` is re-implemented from api-keys, the copy-value-with-toast cell now exists in 4 hand-rolled variants (two in `columns.tsx`, `CopyableItem` here, and the api-keys original — already drifted in `stopPropagation`/icon size/clipboard error handling), and `useLicenseKeys` mirrors `useApiKeys` line-for-line. Extracting `MetadataItem` + a `CopyableValue` component the same way would finish the job.

apps/backend/src/data-marts/use-cases/create-data-destination.service.ts
```diff
@@ -57,7 +55,6 @@ export class CreateDataDestinationService {
 
   @Transactional()
   async run(command: CreateDataDestinationCommand): Promise<DataDestinationDto> {
-    this.availableDestinationTypesService.verifyIsAllowed(command.type);
     const availableForUse = command.availableForUse ?? true;
```
laskevych, line 58: 🔵 **Question (non-blocking):** Keeping all destination configuration available in Community is clearly intentional, but what's the intended experience for a scheduled report created on a deployment where Report Runs are restricted? As written, the scheduler fires forever, producing an unbounded stream of `RESTRICTED` runs with no signal at create/schedule time and no back-off. Is a UI hint at configuration time (or pausing a schedule after N consecutive restricted runs) planned as a follow-up, or is the run history considered signal enough?

laskevych (commented):
Automated multi-angle review (8 finder angles, each candidate re-verified against the PR head; refuted candidates dropped). Two blocking items — both cheap to fix — plus non-blocking latency/consistency/dedup notes. Deliberately does not repeat the three earlier review comments; two comments extend them to adjacent call sites.

🤖 Generated with Claude Code (Fable 5)

apps/backend/src/common/config/app-edition-config.service.ts
kalnik-a-a: We are intentionally keeping the background edition refresh fail-closed. If
the existing bounded JWKS retries are exhausted, the deployment falls back
to Community instead of retaining a stale license context or introducing an
offline grace period. We changed scheduled revalidation from hourly to every
15 minutes, so after JWKS availability returns the context is restored on the
next scheduled check, within one 15-minute interval plus the bounded refresh
attempt. The one-hour verified-key cache remains unchanged, so healthy checks
normally reuse cached JWKS. Separately, when a licensed Report Run reaches
the Cloud gateway, the shared verifier now preserves JWKS infrastructure
failure as a typed error and the gateway returns 503. The client can then use
its existing transient retries and mark the run FAILED after exhaustion
instead of treating the infrastructure failure as invalid credentials.

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
kalnik-a-a: The approved contract for successful-run consumption is best-effort and
explicitly non-transactional. The current client reuses the same serialized
payload and stable run identifier across its bounded transient retries, logs
terminal failure, and preserves the successful run status. The downstream
consumption service already deduplicates those retries: it checks for an
existing record using the stable run identifier and its consumption and
attribution dimensions, while summary calculation additionally counts
distinct run identifiers per decomposition branch. Replaying the same
payload therefore does not increase calculated consumption. A persisted
producer queue would establish a stronger delivery guarantee than the
approved contract and should be designed separately. We are therefore not
adding that machinery in this PR.

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
kalnik-a-a: We are keeping a fresh Cloud authorization decision before every Report Run.
The token cache in the gateway caches signature-verified claims only; it does
not cache the billing `allowed` decision, and the active license record and
origin are still checked on every request. The existing retry budget is
bounded. If we define a tighter authorization latency objective, we can
shorten that dedicated timeout/retry budget separately, without adding an
authorization lease or cached `allowed` result.

apps/backend/src/data-marts/services/project-billing/internal-project-billing.service.ts
kalnik-a-a: This is now addressed. When the license gateway is enabled, startup validates
the Pub/Sub project plus every accepted Report Run topic, including all
destination-specific email-based topics. Forwarded consumption uses a strict
publication path: missing configuration, routing errors, and Pub/Sub failures
propagate through the controller, so the endpoint does not return 202 unless
Pub/Sub accepted the command. The pre-existing INTERNAL consumption path
remains best-effort. Tests cover every required setting, startup rejection,
missing topics, publication failure, and controller propagation.

apps/backend/src/ee/license-keys/guards/license-key.guard.ts
kalnik-a-a: This is now addressed with a typed `JwksUnavailableError`. The shared verifier
preserves that error only for JWKS retrieval/refresh infrastructure failure,
and `LicenseKeyGuard` maps it to HTTP 503. Invalid signatures, claims,
issuer/audience, expiry, revoked records, and key/origin mismatches still
return 401. We did not add a duplicate retry loop: JWKS retrieval already has
bounded retries, and the self-managed Cloud client already retries transient
5xx responses. Tests cover both the typed failure and the 503 mapping.

apps/backend/src/common/config/public-origin.service.ts
kalnik-a-a: The current behavior remains fail-closed, and license setup documentation
requires `PUBLIC_ORIGIN` to contain the exact scheme, host, and port used when
issuing the key. Scheme-less origin support predates this feature, and
changing its generic normalization here could affect existing deployments.
We are leaving it unchanged in this PR. An edition-specific startup warning
or validation error can be considered separately as a diagnosability
improvement.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
kalnik-a-a: `getSchema` no longer performs Report Run billing authorization. The
application models it as a schema/configuration request: it does not create a
run or publish consumption, so applying only the authorization part of the
Report Run lifecycle was inconsistent. Report lookup, schema generation, and
actual data extraction paths remain unchanged. No `allowed` decision cache
was added. A focused service test verifies that schema retrieval does not call
the billing authorization service.

apps/backend/src/data-marts/services/project-billing/binding-aware-project-billing.service.ts
kalnik-a-a: The normal INTERNAL bootstrap path already requires complete balance
configuration and fails startup otherwise. We are not routing a later
INTERNAL binding unconditionally to `InternalProjectBillingService`, because
its legacy unconfigured behavior is `allowed`; doing so would violate the
required fail-closed invariant. The background edition refresh also remains
intentionally fail-closed rather than retaining stale license state. Runtime
binding transitions can be handled as a separate invariant refactor if they
become a supported lifecycle.

apps/backend/src/ee/license-keys/guards/license-key.guard.ts
kalnik-a-a: The observation about expired entries remaining until the same token is seen
again is correct, but the map is bounded by the small number of distinct
license tokens presented to a process. Every request still checks the active
database record, project, origin, and expiry. Keying by a decoded bare `jti`
would be unsafe and would not make rotation replace the old entry because a
rotated key receives a new `jti`. We are keeping the exact-token binding and
deferring optional sweeping/LRU housekeeping until the retained population is
operationally meaningful.

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
kalnik-a-a: Consumption remains awaited after the run has been persisted as successful.
Terminal delivery failure is caught and monitored without changing SUCCESS,
which matches the approved best-effort contract. Replacing this with a bare
fire-and-forget promise would weaken shutdown and error-observation
guarantees. If product defines a tighter run-finalization SLA, the minimal
follow-up is a dedicated shorter transient-delivery budget or an owned
background task with explicit shutdown semantics, rather than an untracked
promise.

apps/backend/src/data-marts/use-cases/run-report.service.ts
kalnik-a-a: We added an enforceable source-level authorization-boundary contract. It maps
every member of `REPORT_RUN_KINDS` to the current execution entrypoints for
Sheets/email-based reports, Looker streaming and non-streaming reports, HTTP
Data, and MCP Query. For each mapped entrypoint it verifies that billing
authorization is present before the first storage boundary. The test fails
for a newly added unmapped Report Run kind, a moved/removed entrypoint, or
authorization moved after storage work. This gives CI enforcement without a
speculative cross-use-case runtime abstraction.

apps/web/src/features/license-keys/components/EditLicenseKeySheet.tsx
kalnik-a-a: We are keeping these feature-local implementations for now. The apparent
duplicates already differ in click propagation, sizing, clipboard failure
handling, labels, layout, hook gating, and error/refresh behavior. A shared
abstraction would need options for those differences without adding required
product behavior. If the implementations later drift into a concrete bug, we
can extract the smallest genuinely shared clipboard control at that point.

apps/backend/src/data-marts/use-cases/create-data-destination.service.ts
kalnik-a-a: The current behavior is intentional for this scope: Community keeps report
and schedule configuration, while each execution passes through the normal
authorization boundary and records an actionable RESTRICTED run. Rejecting
schedule creation would conflict with configuration availability, and
automatically pausing or backing off requires product policy for thresholds,
notification, and reactivation. We will treat a UI hint and schedule-specific
policy as a separate follow-up rather than infer that behavior in this PR.

apps/backend/src/data-marts/services/project-billing/internal-project-billing.service.ts
laskevych: Verified at 5002061dc: `assertForwardedConsumptionConfigured()` now validates the PubSub project and every report-run topic at gateway startup, and the forwarded path uses the strict `publish(..., throwOnFailure)` so missing config, routing errors, and PubSub failures propagate — 202 is only returned after PubSub accepted the record. Unit tests cover the missing-topic, missing-project, and publish-failure paths. Resolved, thanks!

apps/backend/src/ee/license-keys/guards/license-key.guard.ts
laskevych: Verified at 5002061dc: `JwksUnavailableError` is raised only for JWKS retrieval/refresh failures, preserved through `verifyJwtClaims`, and mapped to 503 in the guard, while signature/claim/record failures stay 401. Combined with the 15-minute revalidation cadence on the client side, this addresses the concern. Resolved, thanks!

apps/backend/src/common/config/public-origin.service.ts
laskevych: Understood — scheme-less support predates this feature and the setup docs do require the exact scheme/host/port, so keeping normalization unchanged here is reasonable. An edition-specific startup warning as a follow-up would close the diagnosability gap. Resolving as non-blocking.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
laskevych: Verified at 5002061dc: the billing check is removed from `getSchema` (schema modeled as a configuration request — no run, no consumption), the data-extraction paths still authorize, and the new regression test asserts the billing service is not called on schema-only requests. Resolved, thanks!

apps/backend/src/data-marts/services/project-billing/binding-aware-project-billing.service.ts
laskevych: Fair point — routing INTERNAL unconditionally to the internal service would default to `allowed` when unconfigured, which breaks the fail-closed invariant and is worse than the misleading message. Agreed this belongs to a separate refactor if runtime binding transitions become a supported lifecycle. Resolving.

apps/backend/src/ee/license-keys/guards/license-key.guard.ts
laskevych: Agreed on both counts: the population is bounded by distinct presented tokens, and you're right that a rotated key receives a new `jti`, so replace-on-rotation wouldn't work as I suggested. Deferring sweeping/LRU housekeeping is fine. Resolving.

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
laskevych: OK — keeping the send awaited after SUCCESS is persisted, for shutdown and observability guarantees, is a deliberate and defensible trade-off; a dedicated shorter delivery budget as a follow-up works if a finalization SLA gets defined. Resolving.

apps/backend/src/data-marts/use-cases/run-report.service.ts
laskevych: The authorization-boundary contract test is a neat solution — CI now fails on an unmapped run kind, a moved/removed entrypoint, or authorization placed after the first storage boundary, which is exactly the protection I was after without a speculative runtime abstraction. Resolved, thanks!

apps/web/src/features/license-keys/components/EditLicenseKeySheet.tsx
laskevych: Reasonable — the variants do differ in more dimensions than I credited, and extracting the smallest genuinely shared control once real drift shows up is a fine policy. Resolving.

apps/backend/src/data-marts/use-cases/create-data-destination.service.ts
laskevych: Thanks for the clarification — agreed that pause/back-off thresholds and notifications need product definition, and a UI hint as a separate follow-up sounds right. Resolving.

laskevych (approved):
Re-reviewed at 5002061dc. Both blocking items from my previous review are verified fixed in code and tests: (1) the license gateway now validates the PubSub project and every report-run topic at startup and no longer ACKs a forwarded consumption record unless PubSub accepted it; (2) JWKS infrastructure failures surface as a typed error mapped to 503 in the guard, keeping 401 for genuine credential failures. The getSchema billing round-trip is removed with a regression test, and the new authorization-boundary contract test gives CI enforcement for the per-entrypoint billing checks. The declined non-blocking items all come with reasonable justifications and agreed follow-ups. Range-diff against the previous head shows the rebase introduced no other feature changes. Approving.

🤖 Generated with Claude Code (Fable 5)

apps/backend/src/common/config/app-edition-config.service.ts
romandubovyi: Verified at `5002061dc`:

- Revalidation cadence: `EVERY_HOUR` on `main` → `@Cron('0 */15 * * * *')`, with a spec asserting the cron metadata. Worst-case degradation window drops from 60 to 15 minutes.
- Gateway side: `JwksUnavailableError` (`google-jwt-body.decorator.ts:41`, thrown at `:183`) is preserved through `verifyJwtClaims` and mapped to 503 in `LicenseKeyGuard:99`.
- Client side: `callCloud` maps 401/403 to `ProjectOperationBlockedException` and any other non-OK status (503 included) to a plain `Error`; `base-report-run.model.ts:103` then records `RESTRICTED` vs `FAILED`/`ERROR` respectively. So a JWKS outage no longer reads as invalid credentials.

That covers the blocking half. Un-blocking.

One residual, explicitly **non-blocking**, that the reply does not touch: the second half of the original comment. While `licenseContext` is null, `BindingAwareProjectBillingService.delegate()` routes an INTERNAL deployment to `RestrictedProjectBillingService`. Report Runs correctly fail closed, but Process Runs (`CONNECTOR_RUN`, `DATA_QUALITY_RUN`, `AI_PROCESS_RUN`) keep executing while `skip()` drops their consumption at `logger.debug` — usage `InternalProjectBillingService` would have published to PubSub. The 15-minute cadence makes the window 4× smaller but does not close it, and at debug level it is invisible in production.

Minimal ask for a follow-up (not this PR): raise that `skip()` to `warn` so the gap is at least observable.

Resolving.

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
romandubovyi: Accepting — a fresh authorization decision before every Report Run with a bounded retry budget is a defensible trade-off, and a tighter dedicated timeout can follow once an authorization latency objective exists.

Worth noting the surface did shrink in this PR from a neighbouring thread: `getSchema` no longer calls `verifyCanPerformOperations`, so Looker no longer pays a remote round-trip on every report open/edit.

Resolving as non-blocking.

apps/backend/src/data-marts/services/project-billing/license-project-billing.service.ts
romandubovyi: Accepting the best-effort, non-transactional contract for successful-run consumption.

One note for the record: the deduplication argument is not verifiable in this repository — there is still no idempotency key, persisted marker or queue on the producer side, and `sendConsumption` catches and logs terminal failure as before. The dedup by stable run identifier lives in the Cloud consumption service, so we are taking it on trust from outside this PR. That is fine given the approved contract; a persisted producer queue is rightly a separate design.

Resolving as non-blocking.

romandubovyi (approved):
Approving at `5002061dc`.

All three of my threads are verified and resolved:

- **JWKS transient vs invalid license** — addressed. Revalidation moved from hourly to `0 */15 * * * *` (with a test on the cron metadata), `JwksUnavailableError` is preserved and mapped to 503 at the gateway, and the client separates 401/403 (`RESTRICTED`) from other failures (`FAILED`), so an outage no longer reads as bad credentials. Un-blocking.
- **Consumption delivery guarantees** — accepting the approved best-effort, non-transactional contract.
- **Authorization latency per Report Run** — accepting; the surface also shrank since `getSchema` no longer authorizes.

Two non-blocking follow-ups, neither of which needs to hold this PR:

1. While `licenseContext` is null, an INTERNAL deployment routes to `RestrictedProjectBillingService` and Process Runs keep executing with their consumption dropped at `logger.debug`. The 15-minute cadence bounds the window; raising that `skip()` to `warn` would at least make it observable.
2. Nit: the cadence change left stale wording behind — `binding-aware-project-billing.service.ts:13` still says "The hourly license refresh", and `project-billing.providers.spec.ts:71,79` still name the hourly refresh.

CI note: `Audit all` is red on a transitive `nanoid` advisory (GHSA-2v37-7h3g-55p8), unrelated to this change and present repo-wide. Everything else is green.

PR #1531, opened 2026-08-13
Improve MCP add_report to run new reports immediately

amarchenk0:
## Summary

- run newly created push-destination reports immediately by default from MCP `add_report`
- add an explicit `run_immediately: false` opt-out for configuration-only creation
- return a structured `initial_run` outcome and polling guidance
- keep Looker Studio pull-based and reject an explicit immediate run before creation
- preserve a successfully created report when the initial run cannot be queued, directing the client to retry with `run_report` instead of creating a duplicate
- update MCP instructions, tool descriptions, tests, documentation, billing guidance, and changeset

## Why

MCP previously treated `add_report` and `run_report` as separate steps. An AI model could stop after creation, leaving an empty Google Sheet or a configured report that had never delivered data.

The default now matches the common user intent: create the report and populate or deliver it once. `run_report` remains available for later reruns and for retrying a partial `failed_to_queue` outcome.

## Contract

For Google Sheets, Email, Slack, Microsoft Teams, and Google Chat:

- omitted `run_immediately` queues one billed initial Report Run
- `run_immediately: false` creates configuration only
- `initial_run.status: queued` includes the `run_id` to poll
- `initial_run.status: failed_to_queue` means the report already exists and must not be recreated

For Looker Studio:

- omitted or false returns `not_applicable`
- true is rejected before report creation

The existing top-level `status: created` response remains unchanged; `initial_run` is additive.

## Validation

- 111 targeted Jest tests across the facade, MCP tools, and instructions
- backend build
- backend ESLint
- Prettier check for all changed files
- markdownlint for the MCP setup guide
- `git diff --check`

## Feature demo

1. Ask an MCP client to export a data mart to a new Google Sheet without mentioning a separate run.
2. Verify `add_report` returns `initial_run.status: queued` and a `run_id`.
3. Poll `get_report_run_status` until completion and show that the sheet contains data.
4. Repeat with `run_immediately: false` and verify the report is created without a run.
5. Show Looker Studio returning `not_applicable`.

Fibery work item: https://owox.fibery.io/Sprint_Work/Work_Item/6802


apps/backend/src/ee/mcp/tools/run-report.tool.ts
```diff
@@ -20,7 +20,7 @@ type RunReportInput = z.infer<typeof inputSchema>;
...
   readonly description =
-    'Start an existing report by report_id and deliver fresh data to its push destination (Google Sheets, Email, Slack, Microsoft Teams, or Google Chat). Returns immediately with report_id and run_id; it does not wait or report final status. Then poll get_report_run_status with those ids until should_poll is false, waiting up to 15 seconds between checks when possible. Do not call run_report again for the same report while that run is in progress; each call starts a new billed Report Run, and concurrent runs are rejected as "Report is already running or pending".';
+    'Start an existing report by report_id and deliver fresh data to its push destination. Use this for a later rerun or when add_report returned initial_run.status="failed_to_queue". Do not call it when add_report already returned a queued run_id. Returns immediately with report_id and run_id; it does not wait for final status. Poll get_report_run_status with those ids until should_poll is false, waiting up to 15 seconds between checks when possible. Each call starts a new billed Report Run; concurrent runs are rejected as "Report is already running or pending".';
```
romandubovyi, line 23: ⚪ **Non-blocking:** This description was correctly updated to mention `add_report`'s `failed_to_queue` path, but the sibling `get_report_run_status` tool (`apps/backend/src/ee/mcp/tools/get-report-run-status.tool.ts`, description at ~L38, outside this diff) still reads *"the status of a report run started with `run_report`, identified by `report_id` and the `run_id` it returned"*. After this PR `add_report` also returns a pollable `run_id`, and models often weight the target tool's own description over the previous tool's output. Worth updating it to name `add_report`'s `initial_run.run_id` as a valid source too.

docs/getting-started/setup-guide/mcp.md
```diff
@@ -582,7 +587,7 @@ Once the OWOX server is connected, just ask your assistant in plain language. Yo
 - "Create an email destination for `analytics-alerts@example.com`."
 - "What reports and schedules already exist for the Sales data mart?"
 - "Run the Weekly Ads Report now and tell me when it finishes." **(costs credits)**
-- "Export the Ads data mart to a new Google Sheet called 'Weekly Ads Report'."
+- "Export the Ads data mart to a new Google Sheet called 'Weekly Ads Report' and tell me when the initial run finishes." **(costs credits)**
 - "Create a Looker Studio report from the Sales data mart with all fields."
 - "Send the daily revenue table to the Alerts Slack destination with the message 'Yesterday's numbers'."
```
romandubovyi: ⚪ **Non-blocking:** This example now triggers `add_report`'s default initial run, so it will post the Slack message immediately and consume one Report Run. The Google Sheets prompt two lines above was updated in this PR and marked **(costs credits)**; this one was left unchanged. Please mark it the same way — and consider noting that the message is delivered on create, since for a channel that side effect is not reversible.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
```diff
@@ -419,6 +419,7 @@ describe('McpReportsFacadeImpl.addReport', () => {
     destinationId: 'dest-1',
     fields: ['channel', 'revenue'],
     name: 'Weekly revenue',
+    runImmediately: false,
```
romandubovyi, line 422: ⚪ **Non-blocking:** Setting `runImmediately: false` on the shared `addRequest` fixture moves every pre-existing `addReport` test onto the opt-out path, so the new default — the point of this PR — is asserted in exactly one test, and only for Google Sheets. Nothing covers an email-family destination (EMAIL/SLACK/MS_TEAMS/GOOGLE_CHAT) queueing its initial run by default, i.e. that those types are actually members of `MCP_RUN_REPORT_DESTINATION_TYPE_SET`. Those are the destinations with an outbound, non-reversible side effect, so a regression there is the most expensive one. Consider leaving the fixture on the default and setting `runImmediately: false` only in the tests that specifically exercise the opt-out.

🤖 Generated with Claude Code (Opus 5 (1M context)) + Kimi Code CLI

apps/backend/src/data-marts/facades/mcp-reports.facade.ts
```diff
@@ -90,19 +90,32 @@ export interface McpAddReportRequest {
   name?: string;
   /** Required for email-family destinations; rejected for any other type. */
   message?: McpAddReportMessage;
+  /**
+   * Whether to enqueue the first Report Run after creating the report. Omitted
+   * means true for push destinations and false for pull-based destinations.
+   */
+  runImmediately?: boolean;
```
romandubovyi, line 97: ⚪ **Non-blocking:** This field and `initial_run` below are both well documented, but the `addReport` docstring on the interface itself (~L235-243, outside this diff) still describes only the per-destination branching and says nothing about the new behaviour — that push destinations queue their first run by default and that the result now carries `initial_run`. Since that docstring is the contract someone reads without opening the impl, worth adding a line there too.

🤖 Generated with Kimi Code CLI

romandubovyi (approved):
Reviewed with four agents in parallel against the PR head commit. No blocking issues — the change is well scoped, and the tests, docs, MCP instructions, and changeset were updated together. Approving; four non-blocking comments below.

🤖 Generated with Antigravity (Gemini 3.5 Flash) + Kimi Code CLI + Grok CLI (Grok 4.6) + Claude Code (Opus 5 (1M context))

PR #1532, opened 2026-08-13
feat(mcp): return full project context on demand

amarchenk0:
## Summary

- keep the complete static OWOX workflow/security/tool guidance in MCP initialization
- return the full admin-maintained project description from `get_project_context`
- instruct assistants to fetch project context before the first project-specific operation
- make description updates visible on the next tool call without reconnecting
- document the existing 10,000-character hard limit without adding a soft limit

## Why

The project description was appended to `initialize.result.instructions`. OWOX's SDK path preserves that value, but MCP clients are not required to inject server instructions into the model context, so client/model behavior can lose project-specific context even when the server response is complete.

This change keeps the stable overall OWOX instructions in initialization and moves the dynamic project description to an explicit, inspectable tool result.

Fibery: https://owox.fibery.io/Sprint_Work/Work_Item/6786

## Contract and compatibility

`get_project_context.current_project.description` is a new nullable string field. The change is additive for existing clients. The stored description is returned unchanged and without truncation; the existing project settings API hard limit remains 10,000 characters.

The project description remains supplemental and cannot override the static workflow, security, or tool-usage rules.

## Validation

- focused MCP tests: 21 passed
- exact 10,000-character description verified in both structured and JSON text tool output
- full static instructions round-tripped through the MCP SDK initialization handshake
- targeted ESLint passed
- Markdown lint passed
- backend build passed

## Manual client matrix before ready for review

- [ ] Claude: configure a 10,000-character description with unique start/middle/end markers; start a fresh conversation, make a project-specific request, and verify `get_project_context` is called and all markers are available
- [ ] OpenAI/ChatGPT: repeat the same fresh-connection and project-specific request checks
- [ ] In both clients, update the description and verify a subsequent `get_project_context` call returns the new full value without reconnecting
- [ ] In both clients, verify the static overall MCP hint remains available and is not duplicated in the tool result

### DIFF
--- a/apps/backend/src/ee/mcp/instructions/mcp-instructions.service.spec.ts
+++ b/apps/backend/src/ee/mcp/instructions/mcp-instructions.service.spec.ts
@@ -1,20 +1,18 @@
-import { Logger } from '@nestjs/common';
-import type { ProjectSettingsFacade } from '../../../project-settings/facades/project-settings.facade';
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js';
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { McpInstructionsService } from './mcp-instructions.service';
-import { composeMcpInstructions, MCP_SYSTEM_INSTRUCTIONS } from './mcp-system-instructions';
+import { MCP_SYSTEM_INSTRUCTIONS } from './mcp-system-instructions';
 import { hasUniqueCountFieldCandidate } from '../tools/query-data-mart.input';
 
 const UNIQUE_COUNT_EXAMPLE_FIELD = 'orders__unique_count';
 const DISPLAY_FORM_NON_EXAMPLES = ['Orders Unique Count', '<Prefix> Unique Count'];
 
 describe('MCP instructions', () => {
-  afterEach(() => {
-    jest.restoreAllMocks();
-  });
+  it('returns the complete static system instructions', () => {
+    const service = new McpInstructionsService();
 
-  it('returns only system instructions when the project description is empty', () => {
-    expect(composeMcpInstructions(null)).toBe(MCP_SYSTEM_INSTRUCTIONS);
-    expect(composeMcpInstructions('   ')).toBe(MCP_SYSTEM_INSTRUCTIONS);
+    expect(service.getInstructions()).toBe(MCP_SYSTEM_INSTRUCTIONS);
   });
 
   // The rule the tool ENFORCES (UniqueCountFieldUnsupportedClauseError): a model that reads only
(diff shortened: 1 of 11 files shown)


apps/backend/src/ee/mcp/tools/project-context.tool.ts
```diff
@@ -48,16 +55,20 @@ export class GetProjectContextTool implements McpToolDefinition<GetProjectContex
...
-      projectId: context.projectId,
-      roles: context.roles,
-    });
+    const [result, description] = await Promise.all([
+      this.projectContext.getProjectContext({
+        userId: context.userId,
+        projectId: context.projectId,
+        roles: context.roles,
+      }),
+      this.projectSettings.getDescription(context.projectId),
```
romandubovyi: 🔴 **Blocking:** `getDescription` was previously wrapped in a try/catch in `McpInstructionsService` (removed in this PR), so a project-settings failure only dropped the optional description and logged a warning. Here it sits in a bare `Promise.all`, so any failure of that lookup now fails the entire `get_project_context` call — including `id`, `title`, `roles` and `status`, which do not depend on project settings at all. Since the updated instructions make this the first tool an assistant calls in a conversation, an optional-field lookup becomes a hard entry-point failure.

This repo already has the pattern for exactly this case: `tryGetMcpProjectSummary` in `apps/backend/src/ee/mcp/tools/mcp-project-summary.util.ts`, whose docstring says the enrichment is supplemental and "the primary catalog result stays useful when that enrichment is temporarily unavailable". Please mirror it — catch, log a warning, fall back to `description: null` — and add a test asserting the tool still returns project context when `getDescription` rejects (no such test exists today).

🤖 Generated with Antigravity (Gemini 3.5 Flash) + Kimi Code CLI + Claude Code (Opus 5 (1M context))

apps/backend/src/ee/mcp/instructions/mcp-system-instructions.ts
```diff
@@ -10,7 +10,7 @@ For a concrete analytical question:
 Discovery:
 - Use list_data_marts only when the user explicitly asks to list or browse data marts.
 - Use summarize_data_catalog when the user asks what data is available, what can be analyzed, or does not know where to start.
-- Use get_project_context only when the user asks about the current project.
+- Call get_project_context before the first project-specific operation in a conversation so you receive the current project metadata and its complete admin-maintained description. Reuse that context for subsequent requests unless the user asks you to refresh it.
```
romandubovyi, line 13: ⚪ **Non-blocking:** This PR's whole benefit depends on the model actually calling `get_project_context` first, but the rule sits under `Discovery:`, whose other two bullets say the tool should be used *only* when the user explicitly asks — the section reads as a list of opt-in tools. It also isn't reflected in the numbered "For a concrete analytical question" workflow, which still starts at step 1 with `get_relevant_data_marts_by_prompt`, and that numbered list is what a model follows for a concrete question. Promoting this to its own step above the numbered list (or making it step 1) would make it much harder to skip.

Related: `mcp-instructions.service.spec.ts` already locks critical phrasing with `toContain` assertions (the Unique Count rules, with comments explaining why each half must survive an edit). This new rule carries the same "a model that skips it defeats the feature" property but has no such lock — worth adding one.

🤖 Generated with Claude Code (Opus 5 (1M context)) + Grok CLI (Grok 4.6)

apps/backend/src/ee/mcp/instructions/mcp-instructions.service.spec.ts
```diff
@@ -41,39 +39,25 @@ describe('MCP instructions', () => {
...
 
-  it('appends project context after the system instructions', () => {
-    const instructions = composeMcpInstructions('  Revenue means net revenue.  ');
-
-    expect(instructions.startsWith(MCP_SYSTEM_INSTRUCTIONS)).toBe(true);
-    expect(instructions).toContain('Project context:');
-    expect(instructions.endsWith('Revenue means net revenue.')).toBe(true);
-    expect(instructions.indexOf('Project context:')).toBeGreaterThan(
-      instructions.indexOf('Project-specific context')
+  it('round-trips the complete system instructions through MCP initialization', async () => {
```
romandubovyi, line 42: ⚪ **Non-blocking:** This test builds its own `McpServer` with `{ instructions: MCP_SYSTEM_INSTRUCTIONS }` rather than going through `McpSdkServerFactory.create()`, so it asserts SDK behavior rather than OWOX's own initialization path — if the handler or factory stopped passing `instructions` through, it would still pass. That seam matters more after this PR, since the handler's call site changed (`getInstructions()` is no longer async and no longer takes a `projectId`). Driving the server through `McpSdkServerFactory.create(context, instructionsService.getInstructions())` would make it cover the path the PR description credits it with.

apps/backend/src/ee/mcp/instructions/mcp-instructions.service.ts
```diff
@@ -1,29 +1,9 @@
...
-      const description = await this.projectSettings.getDescription(projectId);
-      return composeMcpInstructions(description);
-    } catch (error) {
-      this.logger.warn('Failed to load project-specific MCP instructions', {
-        projectId,
-        error: error instanceof Error ? error.message : String(error),
-      });
-      return composeMcpInstructions(null);
-    }
+  getInstructions(): string {
```
romandubovyi, line 6: ⚪ **Non-blocking:** With the `ProjectSettingsFacade` dependency gone, this service has no state, no dependencies and no logic — it returns a module-level constant. Its only consumer (`McpStreamableHttpTransportHandler`) could import `MCP_SYSTEM_INSTRUCTIONS` directly, which would drop the class, its module provider and the mock in the handler spec. Worth removing unless the DI seam is wanted for something planned.

romandubovyi (commented):
Reviewed with four agents in parallel against the PR head commit. One blocking finding: moving `getDescription` out of `McpInstructionsService` dropped its try/catch, so a project-settings failure now fails the whole `get_project_context` call — and the updated instructions make that the first tool of every conversation. Three non-blocking comments follow.

Process note: the "Manual client matrix before ready for review" checklist in the PR description is still unchecked, though the PR is not a draft. Since this change trades a guaranteed init-time injection for a model-dependent tool call, that matrix is the main evidence that the new path actually works in real clients.

🤖 Generated with Antigravity (Gemini 3.5 Flash) + Kimi Code CLI + Grok CLI (Grok 4.6) + Claude Code (Opus 5 (1M context))

PR #1533, opened 2026-08-13
feat(data-marts): measure Data Last Updated Date for Snowflake storage

laskevych:
## What

Data Last Updated Date is now measured for Snowflake Data Marts, on every surface where it already works for BigQuery and Redshift: the Data Mart page tile, the Data Marts list, the canvas (badge + Actions sweep), MCP `query_data_mart` responses, and run history. No changes outside the storage layer — the new resolver registers into the existing `SourceDataLastUpdatedResolver` family and everything above it picks Snowflake up automatically.

## How

- **Source discovery** — the adapter's existing `EXPLAIN USING JSON` dry run already returns the compiled plan with every scanned object as a fully-qualified name, views expanded by the compiler. The resolver just flattens and deduplicates `Operations[].objects`. Snowflake's query builder inlines literals and asserts zero bound params, so the composed SQL EXPLAINs as-is. EXPLAIN compiles without waking the warehouse.
- **Per-table times** — `SYSTEM$LAST_CHANGE_COMMIT_TIME`, chosen deliberately over the `INFORMATION_SCHEMA.TABLES` / `ACCOUNT_USAGE.TABLES` timestamps the task suggested investigating: their `LAST_ALTERED` is documented to move on DDL and background maintenance too ("even when no rows are affected"), so it can be NEWER than the last data change — exactly the inversion of the `partial` contract ("at least as recent as") that blocked the same approach for Athena Hive tables. The commit time moves on DML only. It is documented as approximate (clock precision/skew), which is bounded noise rather than a semantic lie; the docs caveat says so.
- All tables of one lookup are measured in a single UNION ALL query; a failed batch retries per table, so one broken object (an external table the function rejects, a shared or unauthorized table, one dropped since the EXPLAIN) degrades only itself to an unknown source with a note. A zero commit time (a table with no recorded data changes) reports null with its own note; an unrecognised value is logged and noted distinctly.
- Per-table conclusions are cached across a batch, and the adapter's dedicated connection is destroyed when the batch ends — Snowflake connections require explicit cleanup, and a canvas sweep must not leak one per storage. `executeQueryAndFetchAll` gained an optional abort signal (the driver cancels the statement natively).

## Tests

- 14 resolver specs mirroring the sibling suites: complete/MAX across tables, single batched query with quoted identifiers, string-typed commit times, zero-value and unrecognised-value notes, per-table fallback isolating a broken object, empty plan, per-item failure isolation, batch-level caching, connection destroyed on success and on failure, abort before and between items, non-Snowflake config.
- All 10 Snowflake suites pass (118 tests); ESLint, Prettier, and `tsc` clean against the baseline.

Docs updated (`data-last-updated.md`, `mcp.md`): Snowflake reports the approximate commit time of the last data change; schema changes and background maintenance do not move it; tables with no recorded changes show Unknown.


prokopetsv (commented):
Reviewed for correctness, readability, maintainability, test coverage, and backward compatibility. **No blockers.** The resolver mirrors the Redshift sibling faithfully, the UNION ALL + per-table fallback degradation is well designed, identifier/literal escaping is sound, and the adapter change is additive. Four non-blocking notes inline: one contract edge case (abort window can report `complete` over a truncated source set), one signal-threading gap, one test gap, one docs wording nit.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,330 @@
...
+      .map(source => source.dataLastUpdatedAt)
+      .filter((value): value is string => value !== null);
+
+    if (resolvedTimes.length === 0) {
+      return { dataLastUpdatedAt: null, computedAt, coverage: 'unavailable', sources };
+    }
+
+    // ISO-8601 UTC strings sort lexicographically in chronological order, so a plain max works.
+    const dataLastUpdatedAt = resolvedTimes.reduce((a, b) => (a > b ? a : b));
+    const isPartial = anyFailed || resolvedTimes.length < sources.length;
```
prokopetsv: 🟡 **Medium** — if the signal aborts after EXPLAIN but before `measureTables`, tables cached by earlier items still populate `sources` while unmeasured ones are skipped, so coverage reports `complete` over a truncated source set. Redshift returns `unavailable` mid-abort. Latent today (multi-item batches come only from the sweep, which passes no signal), but it breaks the schema's `complete` contract. Fix:

```ts
const isPartial =
  anyFailed || resolvedTimes.length < sources.length || sources.length < tables.length;
```

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,330 @@
...
+
+  private async resolveOne(
+    adapter: SnowflakeApiAdapter,
+    item: ResolveSourceDataLastUpdatedItem,
+    cache: Map<string, CachedSource>,
+    signal?: AbortSignal
+  ): Promise<SourceDataLastUpdated> {
+    const computedAt = new Date().toISOString();
+
+    const plan = await adapter.executeDryRunQuery(item.sql);
```
prokopetsv: 🟡 **Low** — `executeDryRunQuery` doesn't accept the abort signal (unlike `executeQueryAndFetchAll` after this PR), so an in-flight EXPLAIN survives an abort and the batch only stops afterwards. `executeQuery` already threads a signal; worth passing it through here too for symmetry. Fine to defer.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+            { SOURCE_TABLE: 'DEV.DLU.ORDERS', LAST_CHANGE: String(NANOS_AUG_1) },
+          ]),
+      })
+    );
+
+    expect(result.dataLastUpdatedAt).toBe('2026-08-01T10:00:00.000Z');
+    expect(result.coverage).toBe('complete');
+  });
+
+  it('reports a table with no recorded changes as a null source with a note', async () => {
```
prokopetsv: 🟡 **Low (test gap)** — the `partial` branch driven by `resolvedTimes.length < sources.length` (a resolved table alongside a zero-commit-time one) is untested; the fallback test reaches `partial` only via `anyFailed`. One spec with `ORDERS: NANOS_AUG_1` + `FRESH: 0` expecting `partial` and ORDERS' time would cover it.

.changeset/6795-data-last-updated-snowflake.md
```diff
@@ -0,0 +1,11 @@
+---
+'owox': minor
+---
+
+# Data Last Updated Date for Snowflake
+
+**Data Last Updated** now works for Snowflake Data Marts, on every surface where it already worked for other storages: the Data Mart page, the Data Marts list, the model canvas, and MCP `query_data_mart` responses. No setup is needed — the value appears as soon as someone checks a Data Mart or a run delivers data.
+
+OWOX asks Snowflake which tables the executed query reads — views and SQL Data Marts resolve through to their underlying base tables — and then reads each table's commit time of the last **data** change. Schema changes and Snowflake's own background maintenance do not move the value: a table that was only re-clustered or renamed keeps the timestamp of its last real write. The time is approximate to within seconds, which is how Snowflake reports it.
```
prokopetsv: 🟡 **Low (docs)** — "approximate to within seconds, which is how Snowflake reports it": Snowflake's docs don't quantify a bound — they say precision and skew "can vary" and recommend the value as a change indicator. Suggest "The time is approximate, as Snowflake reports it" to avoid promising a bound the vendor doesn't.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,330 @@
...
+    nanos = Number(value);
+  } else if (typeof value === 'string' && /^\d+$/.test(value.trim())) {
+    nanos = Number(value.trim());
+  } else {
+    return null;
+  }
+  if (!Number.isFinite(nanos) || nanos <= 0) {
+    return null;
+  }
+  return new Date(Math.floor(nanos / 1_000_000)).toISOString();
```
prokopetsv: 🔴 **Blocker** — `nanosEpochToIsoUtc()` converts `SYSTEM$LAST_CHANGE_COMMIT_TIME` into user-facing `dataLastUpdatedAt`. [Snowflake defines it as a change token](https://docs.snowflake.com/en/sql-reference/functions/system_last_change_commit_time), documents variable precision/skew, and strongly discourages treating it as a timestamp. With no documented skew bound, it can violate the schema’s lower-bound guarantee. Keep Snowflake unsupported until a timestamp source with documented ordering is used, or expose this only as an opaque token.

prokopetsv (changes requested):

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
laskevych: Accepted and reworked in 7f869f6d9. You're right — the vendor defines it as a change-detection token with unspecified precision/skew, so it cannot promise the lower bound the contract requires, no matter how small the skew usually is. The value now comes from `SNOWFLAKE.ACCOUNT_USAGE.TABLE_DML_HISTORY`: `MAX(START_TIME)` of the newest hourly window with recorded user DML is a documented lower bound (start-of-hour ≤ actual change), and background maintenance such as Automatic Clustering is documented as excluded from the view. The trade-offs are hourly granularity, the view's publishing delay of up to ~6 hours, and the connection role needing access to the SNOWFLAKE database — without it every source degrades to an honest unknown with a note. Docs, changeset, and the manual-test expectations updated accordingly. A more precise time via explicit approximation semantics in the schema (which would also unlock the token here and the Hive metadata time on Athena) stays a possible follow-up.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
laskevych: Fixed in 7f869f6d9 with exactly this predicate — `sources.length < tables.length` now caps coverage at partial when an abort lands between the EXPLAIN and the measurement.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
laskevych: Fixed in 7f869f6d9 — `executeDryRunQuery` now takes an optional signal and threads it into `executeQuery`, so an in-flight EXPLAIN cancels with the batch.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.spec.ts
laskevych: Fixed in 7f869f6d9 — the suite now covers a resolved table next to an unknown one reaching `partial` without any failure involved (the fixture became a table with no recorded DML after the rework, same branch).

.changeset/6795-data-last-updated-snowflake.md
laskevych: Fixed in 7f869f6d9 — and the sentence disappeared with the rework: the changeset now describes the DML-history semantics (start of the hour, may trail by several hours) instead of promising a seconds bound the vendor doesn't.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,330 @@
...
+        );
+      })
+      .join('\n   OR ');
+
+    return (
+      `SELECT DATABASE_NAME || '.' || SCHEMA_NAME || '.' || TABLE_NAME AS SOURCE_TABLE,\n` +
+      `       TO_CHAR(CONVERT_TIMEZONE('UTC', MAX(START_TIME)), ` +
+      `'YYYY-MM-DD"T"HH24:MI:SS.FF3"Z"') AS LAST_DML_AT\n` +
+      `FROM SNOWFLAKE.ACCOUNT_USAGE.TABLE_DML_HISTORY\n` +
+      `WHERE ${tuples}\n` +
```
amarchenk0: `TABLE_DML_HISTORY` retains historical records, and Snowflake explicitly provides ID columns because objects can be dropped and recreated under the same name. This query filters and groups only by database/schema/table names, so `DROP` + `CREATE` or `CREATE OR REPLACE` can combine an old table generation with the current one and report the old object's DML time with `coverage: complete`. Please resolve the active table's `TABLE_ID` and match history by ID, or degrade to Unknown when identity is ambiguous, and add a recreate regression test. [Snowflake Account Usage documentation](https://docs.snowflake.com/en/sql-reference/account-usage), [TABLE_DML_HISTORY columns](https://docs.snowflake.com/en/sql-reference/account-usage/table_dml_history).

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,330 @@
...
+  if (parts.length !== 3 || parts.some(part => part.length === 0)) {
+    return null;
+  }
+  return parts as [string, string, string];
+}
+
+/**
+ * Flattens the EXPLAIN plan into the deduplicated set of scanned objects. Every operation
+ * carries an `objects` list; table scans name the fully-qualified base tables the compiler
+ * resolved (views already expanded away).
```
amarchenk0, line 333: The assumption that views are always “expanded away” is not a Snowflake contract: `EXPLAIN.objects` may identify a materialized or secure view, while `TABLE_DML_HISTORY` explicitly excludes background materialized-view maintenance. A Data Mart reading—or optimizer-rewritten through—a materialized view can therefore report Unknown instead of the documented last-refresh time. Please distinguish scan object types and resolve materialized-view refresh/base-table history correctly, or explicitly exclude this scenario from Snowflake support and align the docs, with a regression test. [Snowflake EXPLAIN output contract](https://docs.snowflake.com/en/sql-reference/sql/explain), [TABLE_DML_HISTORY exclusions](https://docs.snowflake.com/en/sql-reference/account-usage/table_dml_history).

amarchenk0 (changes requested):
Two new data-freshness contract blockers remain at the current head.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
laskevych: Fixed in 9b0a808f2. The measurement now resolves the active generation through `ACCOUNT_USAGE.TABLES` (`DELETED IS NULL`) and joins `TABLE_DML_HISTORY` on `TABLE_ID`, never on names — a recreated table answers for its own generation only, and a fresh generation with no DML of its own honestly reports `no data changes recorded` instead of inheriting its predecessor's time. Regression test added for exactly that scenario (catalog row present, no history rows → null + note). A side effect worth noting: a just-created table may not be visible in ACCOUNT_USAGE.TABLES yet (catalog latency), which now surfaces as `table not found in account usage metadata` with partial coverage — documented in the manual-test expectations.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
laskevych: Fixed in 9b0a808f2, taking the explicit-exclusion route you allowed. The same ACCOUNT_USAGE.TABLES row that pins the generation also supplies TABLE_TYPE, so anything that is not a BASE TABLE — a materialized or secure view the engine did not expand, an external table — degrades to an unknown source with a `<kind> — modification time not measured` note and partial coverage, instead of a fabricated "no changes". Docs and the changeset now state the exclusion; regression test covers a materialized view arriving in the EXPLAIN objects next to a measurable table. Resolving MV refresh history properly stays a possible follow-up.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    for (const table of tables) {
+      const row = rowByTable.get(table);
+      let time: string | null = null;
+      let note: string | undefined;
+      if (row === undefined) {
+        // Not in the account-usage catalog of live objects: either the table is newer than
+        // the catalog's publishing delay, or it is not a catalog object at all.
+        note = 'table not found in account usage metadata';
+      } else if (String(row.TABLE_TYPE ?? '') !== 'BASE TABLE') {
```
prokopetsv, line 242: 🔴 **Blocker** — The query treats every `TABLE_TYPE` value `BASE TABLE` as measurable, but [`ACCOUNT_USAGE.TABLES` exposes `IS_ICEBERG` separately](https://docs.snowflake.com/en/sql-reference/account-usage/tables). External-catalog Iceberg data can change outside Snowflake, so DML history can retain an older Snowflake write and this resolver can return it with `complete` coverage. Select `IS_ICEBERG` and report Unknown, or resolve Iceberg snapshot/refresh metadata; add a regression test.

prokopetsv (changes requested):

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
laskevych: Fixed in 8d38945c2 — the measurement now selects `IS_ICEBERG` from the same ACCOUNT_USAGE.TABLES row and reports Iceberg tables as unknown sources with an `iceberg table — modification time not measured` note, even when a DML-history row exists (that row can only describe Snowflake-side writes, not external-catalog changes). Regression test covers exactly that: an Iceberg table with a retained history row next to a regular table resolves to the regular table's time with partial coverage. Docs and changeset state the exclusion; resolving Iceberg snapshot metadata properly stays a possible follow-up.

prokopetsv (approved):
**Approve** ✅

Re-reviewed at `8d38945` with focus on bugs, performance, security, and correctness. No critical issues.

The switch from `SYSTEM$LAST_CHANGE_COMMIT_TIME` to `SNOWFLAKE.ACCOUNT_USAGE.TABLE_DML_HISTORY` resolves the earlier lower-bound concern: `MAX(h.START_TIME)` is a documented lower bound of the last data change, matching the "at least as recent as" contract. The history join on `TABLE_ID` with `t.DELETED IS NULL` correctly isolates a recreated table from its predecessor's history. Iceberg tables, materialized views, and uncataloged objects degrade to honest unknowns with `partial` coverage. Injection into the batched history query is blocked (`splitFqn` rejects quoted/exotic identifiers before interpolation; remaining values are `''`-escaped), and connection cleanup, abort handling, and per-item failure isolation are all test-covered.

🟡 **Non-Blocker** — the PR description is stale: it still describes the abandoned `SYSTEM$LAST_CHANGE_COMMIT_TIME` design, per-table retry, and zero-commit-time handling that the code no longer contains. Please refresh it before merging so the squash-commit message documents the design that actually ships.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,357 @@
...
+          `(t.TABLE_CATALOG = ${quote(database)} AND t.TABLE_SCHEMA = ${quote(schema)} ` +
+          `AND t.TABLE_NAME = ${quote(name)})`
+        );
+      })
+      .join('\n   OR ');
+
+    return (
+      `SELECT t.TABLE_CATALOG || '.' || t.TABLE_SCHEMA || '.' || t.TABLE_NAME AS SOURCE_TABLE,\n` +
+      `       t.TABLE_TYPE AS TABLE_TYPE,\n` +
+      `       t.IS_ICEBERG AS IS_ICEBERG,\n` +
```
amarchenk0, line 301: 🟡 **Non-blocking** — Snowflake reports hybrid tables as `TABLE_TYPE = 'BASE TABLE'` and exposes them through the separate `IS_HYBRID` flag, while `TABLE_DML_HISTORY` explicitly excludes DML on hybrid tables. A changed hybrid table therefore passes the current checks, gets `LAST_DML_AT = null`, and is described as “no data changes recorded in the last year.” The value still degrades safely to Unknown, so this does not block approval, but please select `IS_HYBRID`, return an explicit “hybrid table — modification time not measured” note, and add a regression test.

amarchenk0 (approved):

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-source-data-last-updated.resolver.ts
laskevych: Fixed in 3def7e99e — the query now selects `t.IS_HYBRID`, and a hybrid table gets its own note, `hybrid table — modification time not measured`, before the null-history branch can claim "no data changes": as you said, TABLE_DML_HISTORY excluding hybrid DML makes a null row mean "not tracked here", not "unchanged". Regression test added: a hybrid table next to a healthy one reports partial with the healthy table's time.

PR #1534, opened 2026-08-14
feat(data-marts): measure Data Last Updated Date for Databricks storage

laskevych:
## What

Data Last Updated Date is now measured for Databricks Data Marts, on every surface where it already works for BigQuery, Redshift, and Athena: the Data Mart page tile, the Data Marts list, the canvas (badge + Actions sweep), MCP `query_data_mart` responses, and run history. No changes outside the storage layer — the new resolver registers into the existing `SourceDataLastUpdatedResolver` family and everything above it picks Databricks up automatically. This is the last storage of the family: all five storage types now have resolvers.

## How

- **Source discovery** — `EXPLAIN EXTENDED` over the composed SQL (the adapter's existing dry-run path). The analyzer expands views down to base-table relations; a tolerant parser collects `Relation` and `(Photon)(File)Scan` nodes from the textual plan, strips backticks, and normalises the legacy `spark_catalog` prefix. An unrecognised plan shape degrades to `unavailable` with the raw plan logged.
- **Per-table times** — `DESCRIBE HISTORY` on each Delta table: the newest commit whose operation is on an explicit whitelist of data-changing operations (WRITE, INSERT, UPDATE, DELETE, MERGE, TRUNCATE, RESTORE, COPY INTO, STREAMING UPDATE, CTAS/RTAS, CLONE). History entries document both the commit time and the operation name, so maintenance commits (OPTIMIZE, VACUUM, FSCK) never masquerade as data changes, and an operation the resolver does not recognise simply does not count — the answer can only err older, never newer, which is the direction the `partial` contract ("at least as recent as") requires. The two tempting shortcuts were rejected on documentation grounds: `DESCRIBE DETAIL`'s `lastModified` says nothing about whether maintenance moves it, and Unity Catalog's `information_schema.tables.last_altered` moves on DDL.
- History is read per table with a bounded LIMIT (100 commits) and rendered to ISO-8601 UTC in SQL (`to_utc_timestamp` against the session's own timezone), so the driver hands over deterministic strings regardless of warehouse configuration. History retention (`logRetentionDuration`, 30 days by default) bounds what can be answered: older data changes, non-Delta and external tables, and permission gaps all degrade to unknown sources with notes and partial coverage. Per-table conclusions are cached across a batch, and the adapter's session is closed when the batch ends.

## Tests

- 14 resolver specs mirroring the sibling suites: complete/MAX across tables, escaped identifier + bounded LIMIT in the history query, maintenance commits skipped in favour of the older write, maintenance-only history as unknown, non-Delta failure isolation next to a healthy table, unrecognised timestamp noted distinctly, empty plan, per-item failure isolation, batch-level caching, session closed on success and on failure, abort before and between items, non-Databricks config.
- 6 specs for the Spark-plan parser (Relation nodes, Photon/FileScan nodes, dedup, backticks + `spark_catalog` normalisation, unrelated plan text, empty output).
- All 12 Databricks suites pass (118 tests); ESLint, Prettier, and `tsc` clean against the baseline.

Docs updated (`data-last-updated.md`, `mcp.md`): Databricks is exact for Delta tables within the table's history retention; maintenance operations do not count; non-Delta and external tables show Unknown.


apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+   * The timestamp is rendered to ISO-8601 UTC in SQL — `to_utc_timestamp` with the session's
+   * own timezone makes the string deterministic regardless of warehouse configuration.
+   */
+  private buildHistoryQuery(table: string): string {
+    const escaped = escapeFullyQualifiedIdentifier(table.split('.'));
+    return (
+      `SELECT date_format(to_utc_timestamp(timestamp, current_timezone()), ` +
+      `"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'") AS COMMITTED_AT, operation AS OPERATION ` +
```
prokopetsv: 🔴 Blocker: This query discards `operationMetrics`, while `measureTable` treats every `STREAMING UPDATE` as a data change. Databricks [documents empty streaming commits](https://docs.databricks.com/aws/en/structured-streaming/delta-lake#use-delta-lake-tables-as-a-sink), including every `AvailableNow` run; they do not change table data, so this can report a falsely fresh timestamp. Select the metrics and skip commits with no row/file changes (or return unknown), then add an empty-commit regression test.

apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,274 @@
...
+      // Typically a non-Delta or external table (history only exists for Delta), a dropped
+      // table, or a permission gap. All are per-table conditions; the item stays alive.
+      this.logger.warn(`Failed to read table history for ${table}: ${errorText(error)}`);
+      return {
+        entry: { table, dataLastUpdatedAt: null, note: 'could not read table history' },
+        failed: true,
+      };
+    }
+
+    for (const row of rows) {
```
prokopetsv, line 240: 🟡 Non-Blocker — this loop takes the *first* data-changing row, per the comment "History is newest-first, so the first match IS the latest data change." That holds for `DESCRIBE HISTORY` itself, but the query wraps it in `SELECT ... FROM (DESCRIBE HISTORY t LIMIT 100)`, and Spark guarantees no row order without an `ORDER BY` — the outer projection may return rows in any order.

Impact: `dataLastUpdatedAt` can silently be an older commit from the window and flicker between refreshes, while coverage still says `complete` and the docs promise Databricks "answers exactly". It stays a lower bound, so the contract holds — but the value stops being deterministic.

Fix: `ORDER BY timestamp DESC LIMIT 1` in the outer query, or take the max over all matching rows client-side instead of the first.

apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,274 @@
...
+export class DatabricksSourceDataLastUpdatedResolver implements SourceDataLastUpdatedResolver {
+  readonly type: DataStorageType = DataStorageType.DATABRICKS;
+  private readonly logger = new Logger(DatabricksSourceDataLastUpdatedResolver.name);
+
+  /**
+   * How deep into a table's history to look for a data-changing commit. Aggressively
+   * maintained tables interleave OPTIMIZE/VACUUM commits with writes; a window of 100 covers
+   * that comfortably while keeping the read bounded for streaming tables with huge histories.
+   */
+  private static readonly HISTORY_READ_LIMIT = 100;
```
prokopetsv, line 61: 🟡 Non-Blocker — the operation whitelist is applied client-side, *after* this SQL `LIMIT 100`. A table whose newest 100 commits are all maintenance (auto-compaction / `OPTIMIZE` on a frequently streamed table) reports `no data changes in the recent table history` → Unknown, despite being actively written.

This gets more likely once the line-257 blocker lands: skipping empty `STREAMING UPDATE` commits pushes real writes further down the window.

Fix: push the operation filter into the SQL `WHERE` so `LIMIT` applies after filtering. That also resolves the ordering issue I flagged in `measureTable` and gives the line-257 `operationMetrics` check a natural home.

apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  private async resolveOne(
+    adapter: DatabricksApiAdapter,
+    item: ResolveSourceDataLastUpdatedItem,
+    cache: Map<string, CachedSource>,
+    signal?: AbortSignal
+  ): Promise<SourceDataLastUpdated> {
+    const computedAt = new Date().toISOString();
+
+    const explain = await adapter.executeDryRunQuery(item.sql);
```
prokopetsv: 🟡 Non-Blocker — the signal is never passed to the driver: this `executeDryRunQuery(item.sql)` and the `executeQueryAndFetchAll(...)` on line 210 both omit it, and `DatabricksApiAdapter` accepts no signal at all. The sibling resolvers thread it through (`AthenaSourceDataLastUpdatedResolver` / `RedshiftSourceDataLastUpdatedResolver` use `this.queryOptions(signal)`), and the Snowflake PR added the same plumbing to its adapter.

Consequence: the `if (signal?.aborted) break` guard in the `resolveForSqlBatch` catch block cannot fire from an abort, and an in-flight statement against a cold or paused SQL warehouse (start-up is minutes) runs to completion past the deadline. The between-items and between-tables checks bound this to one statement, so it is robustness rather than a hang.

Fix: add an optional signal parameter to the adapter, matching the siblings.

apps/backend/src/data-marts/data-storage-types/databricks/utils/databricks-plan-relations.util.ts
```diff
@@ -0,0 +1,50 @@
...
+  const segments = [...raw.matchAll(/`([^`]*)`|([^.`]+)/g)].map(match => match[1] ?? match[2]);
+  if (segments.length === 0 || segments.some(segment => segment.length === 0)) {
+    return null;
+  }
+  if (segments[0] === 'spark_catalog') {
+    segments.shift();
+  }
+  if (segments.length < 2) {
+    return null;
+  }
```
prokopetsv, line 56: 🟡 Non-Blocker — this join makes the segment boundaries unrecoverable. `normalizeRelationName` correctly parses `` `main`.`a.b`.`orders` `` into three segments, then joins them with `.`; `buildHistoryQuery` re-splits that string (`escapeFullyQualifiedIdentifier` joins the parts, and the escaper splits on dots again), producing a four-part identifier. `DESCRIBE HISTORY` then fails and the table degrades to `could not read table history` with partial coverage.

No wrong timestamp, and dotted schema names are rare — but note the Databricks docs bullet omits this limitation while the Redshift bullet documents the equivalent one.

Fix: carry the segment array through to `buildHistoryQuery` rather than a joined string.

prokopetsv (commented):
## Summary

The resolver follows the established `SourceDataLastUpdatedResolver` family shape closely, and the parts easiest to get wrong are right: `to_utc_timestamp(timestamp, current_timezone())` + `date_format` correctly cancels the session-timezone shift and yields true UTC; the maintenance-operation whitelist errs older rather than newer, which is the direction the `partial` contract needs; per-item failure isolation, batch-level caching, and session cleanup in `finally` all match the siblings. The plan parser tolerating both `Relation` and Photon/`FileScan` shapes with dedup is sensible.

My findings cluster on one theme: `measureTable`'s "first row wins" logic rests on assumptions that neither SQL nor the tests guarantee. Four inline 🟡 comments below; no new blockers beyond the existing one on line 257 (which I've deliberately not duplicated).

### 🟡 Non-Blocker — the history query shape is only asserted against mocks

`databricks-source-data-last-updated.resolver.spec.ts` mocks `executeQueryAndFetchAll` in every test, so `SELECT ... FROM (DESCRIBE HISTORY t LIMIT 100)` is verified only as a substring match. If a Databricks SQL warehouse rejects that subquery form, `measureTable`'s catch swallows it into a per-table note and every Databricks Data Mart degrades uniformly to Unknown — a silent failure no test would catch. Worth one manual run against a live warehouse before merge, since this path has no integration coverage.

prokopetsv (changes requested):

apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
laskevych: Fixed in aa1347c30 — the query now selects `to_json(operationMetrics)` and a whitelisted commit additionally has to show a change: metrics present with every counter at zero (an empty AvailableNow run) are skipped. Metrics-less commits still count on purpose — those are the plain CREATE/REPLACE boundary commits, and skipping them would reopen the predecessor-history leak closed in dad80483b. Regression tests cover both: an empty streaming commit on top falls through to the older WRITE, and an empty-commits-only history reports unknown with the existing note.

apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
laskevych: Fixed in aa1347c30 with both belts: `ORDER BY timestamp DESC` in the SQL and a client-side max over the matching rows, so the answer does not depend on row order at all. A shuffled-rows spec pins it.

apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
laskevych: Fixed in aa1347c30 — the operation whitelist moved into the SQL `WHERE upper(operation) IN (...)`, so the `LIMIT 100` window is spent on candidate commits rather than maintenance noise, exactly as suggested. The `operationMetrics` emptiness check stays client-side (expressing it over the map in SQL would hardcode metric keys), which can spend window slots on empty streaming commits — if a table's 100 newest candidate commits are ALL empty, it honestly reports unknown; that felt acceptable against the brittleness trade-off.

apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
laskevych: Fixed in aa1347c30 — the adapter now takes an optional `{signal}` through `executeQuery` / `executeQueryAndFetchAll` / `executeDryRunQuery`; on abort the cursor is closed (best-effort server-side cancel) and the fetch loop checks the flag between chunks, so an in-flight statement against a cold warehouse no longer runs past the deadline. The resolver threads its batch signal into both call sites, matching the siblings.

apps/backend/src/data-marts/data-storage-types/databricks/utils/databricks-plan-relations.util.ts
laskevych: Fixed in aa1347c30 — `parseRelationsFromSparkPlan` now returns `{name, segments}` and `buildHistoryQuery` quotes the segments directly (no re-split anywhere), so a backticked segment containing dots survives into `DESCRIBE HISTORY` intact. Specs cover the parser (```main`.`weird.schema`.`orders``` → three segments) and the query building; with the fix in place there is no limitation left to document.

apps/backend/src/data-marts/data-storage-types/databricks/services/databricks-source-data-last-updated.resolver.ts
```diff
@@ -0,0 +1,345 @@
...
+  if (metrics === null || typeof metrics !== 'object') {
+    return true;
+  }
+  const counters = Object.values(metrics as Record<string, unknown>)
+    .map(value => (typeof value === 'string' && /^\d+$/.test(value) ? Number(value) : null))
+    .filter((value): value is number => value !== null);
+  if (counters.length === 0) {
+    return true;
+  }
+  return counters.some(value => value > 0);
```
prokopetsv, line 332: 🟡 Non-Blocker — "any positive numeric metric" defeats the emptiness check for DML: Databricks documents `executionTimeMs`/`scanTimeMs`/`rewriteTimeMs` among `operationMetrics` for UPDATE, DELETE, and MERGE (TRUNCATE has `executionTimeMs`), and timings are essentially always positive. So a zero-row DELETE/UPDATE/MERGE (`numDeletedRows: "0"`, `executionTimeMs: "1523"`) still counts as a data change — the same falsely-fresh inversion the original blocker described, now only for zero-row DML. Fix: exclude the three documented timing keys (or count only `num*` counters), keeping the metrics-less fallback that protects CREATE/REPLACE boundaries. STREAMING UPDATE and WRITE have no timing metrics, so the fixed streaming case stays fixed.

apps/backend/src/data-marts/data-storage-types/databricks/adapters/databricks-api.adapter.ts
```diff
@@ -116,19 +116,39 @@ export class DatabricksApiAdapter {
...
     let queryError: Error | null = null;
+    const signal = options?.signal;
+    const onAbort = () => {
+      // Best-effort server-side cancel; the pending fetch rejects with the cancel error.
+      cursor.close().catch(() => undefined);
+    };
+    if (signal) {
+      if (signal.aborted) {
+        onAbort();
+        throw new Error(`Aborted before executing query: ${query}`);
```
prokopetsv, line 142: 🟡 Non-Blocker — this pre-abort branch is the one exit path where a failed cursor close leaves no trace: `onAbort()` fire-and-forgets `cursor.close().catch(() => undefined)` and the throw happens before the `try/finally`, so `closeCursorSafely`'s warning log never runs. A caller passing an already-aborted signal (any future caller — the resolver happens to pre-check) could leave the statement running server-side with nothing in the logs, against this method's stated goal. Fix: route this branch through `closeCursorSafely` (or await the close with a logged catch) before throwing.

prokopetsv (commented):
## Re-review of aa1347c30

All five prior findings are genuinely fixed, each with a regression spec: the operation filter and `ORDER BY timestamp DESC` moved into SQL before the `LIMIT` (window now holds candidates; row order deterministic, plus a client-side max as a second belt); `operationMetrics` gates whitelisted commits, with metrics-less CREATE/REPLACE commits deliberately still counting (correct — skipping them would reopen the predecessor-history leak); the adapter threads `AbortSignal` through all three query methods with idempotent cursor close (verified: `isClosed` is set synchronously, so the abort-listener close and `closeCursorSafely` cannot double-close); and the parser's `{name, segments}` shape ends the re-split — a dotted backticked segment survives into `DESCRIBE HISTORY` intact. `ORDER BY timestamp` referencing the non-projected column inside the SELECT wrapper is valid Spark SQL. Two new 🟡 inline comments on the fix commit itself; nothing blocking.

prokopetsv (approved):
**Approve** ✅ — head `19621d6f0` is `aa1347c30` plus a clean merge of `origin/main` (the only Databricks-side change is main's #1526 validation-code refactor flowing through; the resolver is untouched).

All originally blocking findings are resolved and regression-tested: SQL-side operation filter + `ORDER BY … LIMIT`, `operationMetrics` gate for empty streaming commits, abort plumbing, and the segments-based identifier round-trip.

The two 🟡 Non-Blockers from the `aa1347c30` re-review stay open as acknowledged follow-ups: timing metrics (`executionTimeMs` etc.) defeating the emptiness check for zero-row UPDATE/DELETE/MERGE, and the unlogged cursor close on the pre-abort path. Neither blocks merging.

🔴 Blocker: none
🟡 Non-Blocker: 2 (follow-ups, noted inline on the previous review)

PR #1535, opened 2026-08-14
fix(data-marts): escape leading + values when exporting to Google Sheets

laskevych:
## Problem

Google Sheets parses a cell value that starts with `+` as the beginning of a formula. Exported values like `+Dri7…` (e.g. session identifiers) rendered as `#ERROR! Formula parse error` instead of the data, because the destination writes values with `valueInputOption: 'USER_ENTERED'`.

## Solution

`SheetValuesFormatter` now prefixes string values that start with `+` with an apostrophe (`'`) — the standard Sheets escape symbol. The cell displays the original value exactly as stored; the apostrophe is not part of the cell content.

Only `+` is escaped, deliberately:
- values starting with `=` stay untouched, so formulas can still be inserted as data (unlike Connected Sheets, which escapes everything and blocks formulas);
- numbers, dates, nulls and all other values are exported unchanged.

The escape is applied in the single value-formatting layer used by every write path (manual report runs, scheduled runs, MCP), so all export scenarios are covered. It is a first-char check per cell — no measurable impact on write time.

## Testing

- New `sheet-values-formatter.spec.ts`: `+` values get the apostrophe (including columns without a registered type formatter), `=` formulas and non-string values stay untouched, timestamp formatting still applies.
- `google-sheets-report-writer.spec.ts` — 35/35 green.
- `nest build`, eslint, prettier — clean.

### DIFF
--- a/apps/backend/src/data-marts/data-destination-types/google-sheets/services/sheet-formatters/sheet-values-formatter.spec.ts
+++ b/apps/backend/src/data-marts/data-destination-types/google-sheets/services/sheet-formatters/sheet-values-formatter.spec.ts
@@ -1,0 +1,90 @@
+import { BigQueryFieldType } from '../../../../data-storage-types/bigquery/enums/bigquery-field-type.enum';
+import { ReportDataHeader } from '../../../../dto/domain/report-data-header.dto';
+import { SheetValuesFormatter } from './sheet-values-formatter';
+
+describe('SheetValuesFormatter', () => {
+  let formatter: SheetValuesFormatter;
+
+  beforeEach(() => {
+    formatter = new SheetValuesFormatter();
+  });
+
+  describe('formatRowsValuesByName', () => {
+    const headersByName = new Map<string, ReportDataHeader>([
+      [
+        'session_id',
+        new ReportDataHeader('session_id', undefined, undefined, BigQueryFieldType.STRING),
+      ],
+      ['clicks', new ReportDataHeader('clicks', undefined, undefined, BigQueryFieldType.INTEGER)],
+    ]);
+    const finalNames = ['session_id', 'clicks'];
+
+    it('prefixes values starting with + with an apostrophe', () => {
+      const rows = [['+Dri7CMMPMBcFIRWcd8RP3AAIi0eVSDWK0PB7RnFQIc=', 42]];
+
+      const result = formatter.formatRowsValuesByName(rows, finalNames, headersByName, 'UTC');
+
+      expect(result).toEqual([["'+Dri7CMMPMBcFIRWcd8RP3AAIi0eVSDWK0PB7RnFQIc=", 42]]);
+    });
+
+    it('escapes + values in columns without a registered type formatter', () => {
+      const rows = [['+380501234567', '+1']];
+      const names = ['phone', 'delta'];
+
+      const result = formatter.formatRowsValuesByName(rows, names, new Map(), 'UTC');
+
+      expect(result).toEqual([["'+380501234567", "'+1"]]);
+    });
+
+    it('leaves values starting with = untouched so formulas can be inserted as data', () => {
+      const rows = [['=SUM(A1:A2)', 1]];
+
+      const result = formatter.formatRowsValuesByName(rows, finalNames, headersByName, 'UTC');
+
+      expect(result).toEqual([['=SUM(A1:A2)', 1]]);
+    });
+
+    it('leaves non-string values and strings without leading + untouched', () => {
+      const rows = [
+        ['plain text', 0],
+        ['a+b', -5],
+        [null, undefined],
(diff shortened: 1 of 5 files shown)


apps/backend/src/data-marts/data-destination-types/google-sheets/services/sheet-formatters/sheet-values-formatter.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+      const result = formatter.formatRowsValuesByName(rows, names, new Map(), 'UTC');
+
+      expect(result).toEqual([["'+380501234567", "'+1"]]);
+    });
+
+    it('leaves values starting with = untouched so formulas can be inserted as data', () => {
+      const rows = [['=SUM(A1:A2)', 1]];
+
+      const result = formatter.formatRowsValuesByName(rows, finalNames, headersByName, 'UTC');
+
```
prokopetsv, line 43: 🟡 Non-Blocker — the formatter is well covered here, but nothing proves the escaped value actually reaches the sheet: `google-sheets-report-writer.spec.ts` stubs `formatRowsValuesByName` with an identity mock, so if a future refactor passed `orderedRows` instead of `formattedRows` to `adapter.updateValues` (`google-sheets-report-writer.ts:384`), every existing test would still pass while the fix silently stopped applying in production.

Suggest one writer-level test using the real `SheetValuesFormatter` that feeds a `+`-leading value through the batch write and asserts `adapter.updateValues` received the apostrophe-prefixed value.

prokopetsv (commented):
## Review summary

The fix is correct and well-scoped. Verified: the data write really uses `USER_ENTERED` (`google-sheets-api.adapter.ts:773`); only TIMESTAMP-family formatters are registered, and numeric driver values arrive as JS numbers, so the `typeof value === 'string'` guard cannot corrupt typed columns (no `"+5"` regression); retries cannot double-escape (`reorderRowsToFinalLayout` rebuilds rows fresh per batch, and an escaped `'+…` no longer starts with `+`); the in-place mutation never aliases back to the caller's original rows; comments, tests, and changeset claims all match the code. Two 🟡 findings.

### 🟡 Non-Blocker — header row still hits the same `#ERROR!`, one write path over

`writeHeaders()` (`google-sheets-report-writer.ts:1009–1021`, not in this diff) writes `header.alias || header.name` through the same `updateValues`/`USER_ENTERED` path without any escaping. A column alias starting with `+` (aliases are free text — `z.string().optional()`, no character restriction) renders `#ERROR! Formula parse error` in row 1: the exact defect class this PR fixes. It fails silently, and since refresh logic matches columns against row-1 text, a corrupted header cell also risks breaking column matching on later runs. Fine as a follow-up if you prefer — but a public `escapeRow`-style helper on `SheetValuesFormatter` applied to `headerRow` before the `updateValues` call would close the class completely.

### Optional

`formatRowsValues` (the positional variant) has no production caller — only `formatRowsValuesByName` is used (`google-sheets-report-writer.ts:372`). Pre-existing, and this PR had to widen the duplication by wiring the escape into both loops. A follow-up could delegate one to the other or drop the dead variant.

prokopetsv (approved):

apps/backend/src/data-marts/data-destination-types/google-sheets/services/sheet-formatters/sheet-values-formatter.spec.ts
laskevych: Added in 45eda867b: `buildWriter` now takes `useRealValuesFormatter`, and a writer-level test wires the real `SheetValuesFormatter` through `prepareToWriteReport` → `writeReportDataBatch` and asserts `adapter.updateValues` received the apostrophe-prefixed values (both a `+`-leading data cell and a `+`-leading header alias). A refactor that drops `formattedRows` in favour of `orderedRows` now fails this test.

laskevych:
Thanks for the review — both 🟡 findings are addressed in 45eda867b:

- **Header row**: the escape helper is now public (`SheetValuesFormatter.escapeRowValues`) and `writeHeaders()` runs the header row through it before `updateValues`, so a `+`-leading alias no longer hits the formula parser. Read-back of row 1 returns the value without the apostrophe, so column matching on later runs is unaffected.
- **Writer-level coverage**: a new test wires the real formatter into the writer and asserts the escaped values are what actually reaches `adapter.updateValues` (details in the inline thread).

The optional cleanup (dead positional `formatRowsValues`) is intentionally left out of this PR — will handle it as a separate follow-up after merge.

laskevych:
Update: the optional cleanup landed in this PR after all — 48afddce9 drops the unused positional `formatRowsValues` variant (only the by-name one has a caller), so no follow-up needed. Both specs green.

PR #1536, opened 2026-08-14
feat(data-marts): remove the Row Count column from aggregated reports

laskevych:
## What

The `Row Count` column is removed from aggregated reports entirely. Reports — and the MCP `query_data_mart` tool and the HTTP data endpoint — return only the columns the user selected.

## Why

A report should contain exactly what the user selected, and Unique Count already covers the counting need, so the Row Count concept is dead weight. (Note: Unique Count is `COUNT(DISTINCT pk)` — a replacement by purpose, counting entities, not a byte-for-byte `COUNT(*)` equivalent; per-group row counts remain reachable via an explicit **Count** aggregation on an always-filled column.)

## How

The branch evolved in two steps (kept as separate commits for reviewability):

1. **Stop auto-adding** (reviewed earlier): `shouldIncludeRowCount` became an explicit read-plan opt-in; validator un-reserved the name; UI hint removed; MCP/HTTP docs updated.
2. **Remove the concept** (new commits): the read-plan and query-option `rowCount` flags, the `COUNT(*)` rendering in every storage builder (flat, legacy and blended paths), the synthetic header, and the `ROW_COUNT_LABEL` constants (backend and the web mirror) are deleted, together with every test that exercised them — net −700 lines.

## Tests

- Backend: 66 storage/interface suites (952 tests) + services/use-cases suites (484 tests) pass after the removal; shared utils suites pass.
- Web: ReportColumnPicker + aggregation-labels suites pass (336 tests).
- Warehouse integration tests updated: no test opts into or asserts a Row Count column anymore.

## User-facing communication

One message (the changeset): the Row Count column is gone from aggregated reports; use **Count** on an always-filled column for per-group row counts, or Unique Count for unique entities.


apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -91,7 +91,7 @@ When building the query:
...
+A data mart can narrow a field's aggregations further ("only where enabled on the field") — get_data_mart_details_by_id returns each field's effective allowedAggregations; trust that over this table. Note COUNT/COUNT_DISTINCT are NOT available on number fields — to count rows per group, apply COUNT to a non-number field you already select (an id or text field). To count UNIQUE records of a JOINED data mart (e.g. "how many distinct orders per customer"), do not aggregate its id column — instead select that source's own Unique Count field like any other field; get_data_mart_details_by_id (with_joined_fields) lists it among that source's joined fields whenever one is available. Copy its "name" (e.g. "orders__unique_count"), never its human-readable "displayName". It can be selected in "fields" and ordered by in "sort" (using that same exact name), but it cannot be used in filters, slices, aggregations, or date_buckets.
```
prokopetsv: 🟡 Non-Blocker — `COUNT(field)` skips NULLs, while the removed automatic column was `COUNT(*)`. An agent following this guidance literally may pick a nullable text field and silently undercount group rows. The docs version (`report-aggregations.md`) already says “a column that is always filled” — mirror that caveat here, e.g. “apply COUNT to a non-number field that is always filled (an ID field works well)”.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -506,16 +505,16 @@ export class OutputControlsValidatorService {
...
-   * synthetic `Row Count` (when aggregated), `Unique Count` (when uniqueCount) and one
-   * `<source>__unique_count` per joined source. A real
+   * synthetic `Unique Count` (when uniqueCount) and one
+   * `<source>__unique_count` per joined source. Saved reports never project `Row Count`
+   * (it is opt-in on read plans only), so it is not part of this namespace. A real
    * column whose name equals a synthetic label — or any two projected names that coincide —
    * is a duplicate alias on BigQuery / silent clobber on name-keyed readers. Uses the SAME
    * label helpers the renderer/header-generator use so this can never drift from the SELECT.
    */
   validateOutputColumnNames(
```
prokopetsv, line 514: 🟡 Non-Blocker — latent guard gap: the renderer can still emit `COUNT(*) AS "Row Count"` for an opted-in read plan, but this validator no longer reserves the name. Unreachable today (only integration tests pass `rowCount: true`), but a future caller wiring the opt-in alongside a real "Row Count" column would get a duplicate alias with no validation net. Worth a one-line comment here — or removing the now production-dead opt-in plumbing in a follow-up.

prokopetsv (commented):
## Review summary

The opt-in flip is coherent end-to-end: `shouldIncludeRowCount` is the single predicate, the header resolver mirrors it (`=== true`), the explicit opt-in SQL plumbing keeps working, and no production path passes `rowCount: true` anymore (only specs and the warehouse integration tests). Sort configs never accepted "Row Count" as a target, so no saved report can break at run time, and the save-path validator change correctly un-reserves the name. Tests were updated at every affected layer. No blockers — one finding below plus two inline 🟡 comments.

### 🟡 Non-Blocker — stale public API description

`apps/backend/src/data-marts/controllers/spec/external/http-data.api.ts:35` (not in the diff, so no inline anchor) still promises: “…and a synthetic `"Row Count"` column is appended”. That is the Swagger description of the exact endpoint this PR changes — API consumers reading OpenAPI will expect a column that never arrives. Fix: drop the clause, mirroring the `docs/api/api-client.md` edit.

prokopetsv (changes requested):
**Summary**

The opt-in `rowCount` behavior is coherent, and existing review comments were accounted for. However, the PR leaves multiple tests relying on the removed implicit behavior, including failures in the required backend CI job.

**🔴 Blocker**

- `redshift-report-reader.service.spec.ts:50,82,203` still expects `Row Count` without passing `rowCount: true`, causing three failures in `Overall test @owox/backend`. Update these cases for the new default, or explicitly opt in when testing row-count binding.

- `query-data-mart-real.integration.ts:208-232` still asserts that `QueryDataMartCommand` automatically returns `Row Count`. This will fail when the live integration suite runs. Assert that the column is absent and remove its count checks, or explicitly opt in if intentional.

**🟡 Non-Blocker**

No additional substantive findings beyond existing PR feedback.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
laskevych: Good catch — fixed in 8ba9aa6fe: the guidance now says to use a non-number field that is always filled (an id field works well) and spells out that COUNT skips NULLs, mirroring the docs wording. Thanks!

apps/backend/src/data-marts/services/output-controls-validator.service.ts
laskevych: Agreed there's a latent gap. Added a warning comment in 8ba9aa6fe: a caller that ever wires the rowCount opt-in into a SAVED report must add the label back into this namespace. Left the opt-in plumbing in place for now — it's what the warehouse integration tests exercise, and it's the natural seam if Row Count ever returns as an explicit toggle. Thanks!

laskevych:
Thanks for the thorough review — all findings are addressed in 8ba9aa6fe:

- **Redshift reader specs** (`redshift-report-reader.service.spec.ts`): the three name-binding cases now pass `rowCount: true` explicitly — they exercise Row Count binding, so opting in keeps their intent, consistent with the neighbouring pagination cases.
- **Live MCP integration** (`query-data-mart-real.integration.ts`): now asserts `Row Count` is absent from an aggregated query's columns and drops the per-group count checks.
- **OpenAPI description** (`http-data.api.ts`): the appended-Row-Count clause is removed — thanks for spotting it outside the diff.
- **MCP guidance**: now warns that COUNT skips NULLs and points to an always-filled field.
- **Validator**: warning comment added about re-reserving the label if the opt-in ever reaches saved reports.

Also fixed the Prettier failure (long line in `stream-http-data.service.spec.ts`). The remaining `Audit all` failure is an unrelated repo-wide advisory in a transitive dependency (`nanoid < 3.3.18`) and will be handled separately.

prokopetsv (approved):
**Approve** ✅ — re-reviewed at `8ba9aa6fe`; all outstanding findings are resolved.

- 🔴 redshift reader specs: fixed by explicit `rowCount: true` opt-ins — the right choice, since those tests exercise row-count binding and this keeps the opt-in plumbing covered. `Overall test @owox/backend` is green on this head.
- 🔴 `query-data-mart-real.integration.ts`: now asserts `Row Count` is absent; per-group count checks removed.
- 🟡 MCP tool guidance: carries the "always filled (an id field works well; COUNT skips NULLs)" caveat.
- 🟡 validator namespace: warning comment added for any future saved-report opt-in wiring.
- Stale `http-data.api.ts` OpenAPI clause removed.

The only failing check is `Audit all` — the repo-wide nanoid advisory (GHSA-2v37-7h3g-55p8), unrelated to this PR and fixed by #1539.

🔴 Blocker: none
🟡 Non-Blocker: none

laskevych:
Heads-up: the scope of this PR has grown per a team decision — Row Count is not just no longer automatic, the concept is removed entirely (Unique Count covers the counting need). The new commit (`refactor(data-marts): remove the Row Count concept entirely`) is a mechanical deletion on top of the previously reviewed state: read-plan/query-option flags, the COUNT(*) rendering in every storage builder, the synthetic header, the labels, and their tests — net −700 lines, no user-visible change beyond what the updated changeset already announces. Reviewing just the diff since your last review should be enough. The PR title/description are updated accordingly.

zapolsky (commented):

Context for both findings: **head `55131c912` has no CI run at all** (`gh api repos/OWOX/owox-data-marts/commits/55131c912/check-runs` → `total_count: 0`). The existing ✅ approve was given at `8ba9aa6fe`, i.e. before the −700-line final commit. There is no green-CI safety net for the largest commit in this PR.

apps/backend/src/data-marts/data-storage-types/utils/sql-clause-renderer.spec.ts
```diff
@@ -263,14 +244,6 @@ describe('SqlClauseRenderer', () => {
       expect(out.aliasByColumn.get('x')).toBe('"x | SUM"');
     });
```
zapolsky: **[CI break — verified]** Stray blank line left where the last `it()` of the `describe` was deleted. Prettier strips blank lines at the end of a block, so the `Prettier` CI job will go red.

**Evidence chain:**
- `sql-clause-renderer.spec.ts:245` = `    });` (closes the last surviving `it`) → `:246` = empty → `:247` = `  });` (closes the `describe`). The final commit removed `it('appends an unqualified COUNT(*) Row Count even in qualified mode', …)` here and kept the leading blank context line.
- In scope for the check: `apps/backend/package.json:45` → `format:check: prettier --check "**/*.{ts,json}" --ignore-path ../../.prettierignore`; `.prettierignore` excludes only `node_modules`/`dist`/`build`/`coverage`/`playwright-report`/`test-results`/`package-lock.json`/`.astro` — spec files are checked.
- `.github/workflows/lint-owox.yml:32-34` runs `npm run format:check` as its **own matrix job**, separate from ESLint — so a green ESLint says nothing about it.
- **Empirical proof that Prettier rejects this shape:** a repo-wide scan for a blank line immediately preceding a closing `})` across every `apps/backend/**/*.ts` and `apps/web/**/*.{ts,tsx}` file returns **exactly one hit — this one**. A tree that CI keeps Prettier-clean contains no other instance, which is why this is the artifact of the deletion rather than an accepted style.

**Why CI has not caught it:** no workflow has run on this head (0 check-runs), and the last green `Prettier` job was on `8ba9aa6fe`, before this line existed.

**Fix:** delete line 246.

*(Not run locally, per this review's no-execution rule — settled by `npm --workspace @owox/backend run format:check`, or simply by the CI run this head never got.)*

.changeset/6838-no-automatic-row-count.md
```diff
@@ -0,0 +1,7 @@
+---
+'owox': minor
+---
+
+# The "Row Count" column is removed from aggregated reports
+
+Previously, every aggregated report automatically included a `Row Count` (`COUNT(*)`) column, even though it was never selected. The column has been removed: reports — and the MCP and HTTP data endpoints — now return only the columns you selected. If you need the number of underlying rows per group, apply the **Count** aggregation to a column that is always filled (an ID column works well); to count unique entities, use a Unique Count. A column of your own named "Row Count" is also no longer rejected in aggregated reports.
```
zapolsky, line 7: **[Breaking contract change, uncommunicated — verified, adversarially confirmed]** The release note says the column is gone from "reports — and the MCP and HTTP data endpoints". It never mentions **destinations** — but `Row Count` was a persisted **Looker Studio schema field**, and existing charts bound to it now fail with an opaque HTTP 400.

**Evidence chain (every hop verified against the checked-out head, and against `origin/main` for the pre-PR state):**
1. Pre-PR the column was **opt-out, defaulted ON** — not opt-in: `origin/main:apps/backend/src/data-marts/data-storage-types/utils/report-data-headers.utils.ts:115` — `const includeRowCount = options?.rowCount ?? (options?.aggregationConfig?.length ?? 0) > 0;`. So *every* aggregated report got the header.
2. The Looker Studio path supplies `aggregationConfig` and never sets `rowCount`: `report-data-cache.service.ts:72` (`aggregationConfig: decision.aggregations ?? report.aggregationConfig ?? undefined`), so it took the default branch → header present.
3. That header list **is** the Looker Studio schema: `looker-studio-connector-api-schema.service.ts:32-40` → `getSchemaFields(...)` → `looker-studio-type-mapper.service.ts` `buildSchemaField` sets `name: header.name`, i.e. literally `"Row Count"`. Its `aggregateFunction: 'COUNT'` maps to `conceptType: METRIC` (`looker-studio-aggregation-mapper.service.ts`) — a first-class LS metric, exactly what a scorecard is built on.
4. Looker Studio persists the data-source schema client-side. This repo documents that: `docs/destinations/supported-destinations/data-studio.md` tells users to click **Refresh fields / Reconnect** after a schema change, and `data-destination-type.enum.ts:57` states "Looker Studio matches its schema fields by `SchemaField.name`". Charts therefore keep requesting the old name.
5. `looker-studio-connector-api-data.service.ts:130-132` filters requested fields against the *current* headers; `:134-136` `if (filteredHeaders.length === 0) throw new BusinessViolationException('No valid fields found in the request')`. Same path for streaming (`prepareStreamingContext`, `:303-308`).
6. `common/exceptions/base-exception.filter.ts` renders that as a **400 whose message names neither the field nor the cause**, and the run is recorded as failed in Run History.

**Trigger:** any existing Looker Studio dashboard on an aggregated Data Mart. A scorecard whose only field is `Row Count` → hard 400. A table that also uses surviving fields → **silently narrowed**: the unknown name is dropped from `filteredHeaders` and from the response schema (`:139` → `createFieldIndexMap` `:255-268`, `buildResponseSchema` `:186-194`), so a fix that only addresses the 400 leaves the silent case.

**Rollout shape worth stating in the note:** the break is staggered, not instant. `report-data-cache.service.ts:147` returns the *persisted* `dataDescription` on a cache hit, so each report keeps working until its entry expires — 5 min default, up to 12 h from the UI (`LookerStudioReportEditForm.tsx:63-75`). Reports fail in a trickle over up to half a day, which makes it harder to correlate with the deploy.

**Also unmentioned:** Google Sheets destinations structurally **delete** the column on their next run (`column-plan-builder.ts` emits `StructuralColumnOp('delete', …)`, applied via `deleteDimension` in `google-sheets-report-writer.ts`), shifting anything to its right one column left — user formulas and pivots included.

**Fix:** extend the changeset with a short "What changes in existing reports" block covering Google Sheets and Looker Studio — the sibling `.changeset/6792-joined-unique-count.md` is the in-repo precedent for exactly this (it devotes a whole section to existing-report impact). Optionally, name the unresolvable fields in the 400 (`looker-studio-connector-api-data.service.ts:135`) — cheap, LS-local, and turns an opaque error into a self-diagnosing one for this migration.

*Adversarially verified: an independent skeptic tried to refute this by looking for a pre-PR opt-in the LS path never set (the finding would have collapsed) — the flag was the opposite, an opt-out defaulted on. Aggregated LS reports are also not an exotic corner: `data-studio.md` recommends aggregating as the remedy for the 1M-row/50MB limits.*

zapolsky:

Primary notes (criticality ≥ 80), highest first — none of these block merge. Two blocked issues are inline.

- **⚠️ No CI has run on this head at all.** `gh api repos/OWOX/owox-data-marts/commits/55131c912/check-runs` → `total_count: 0`, and `gh run list --branch feat/6838-no-auto-row-count` shows the newest runs are on `8ba9aa6fe`. The ✅ approve was given at `8ba9aa6fe`, i.e. **before** the `refactor: remove the Row Count concept entirely` commit (−700 lines, 50 files). Nothing mechanical has been verified on what will actually merge. Please push an empty commit or re-run the workflows before merging — the inline Prettier blocker is precisely what a CI run would have caught for free.

- `.changeset/6838-no-automatic-row-count.md:7` + `docs/getting-started/setup-guide/report-aggregations.md:127` — **the advertised migration path is impossible for a numeric ID, and the docs page now contradicts itself.** Both texts say "apply the **Count** aggregation to a column that is always filled (an **ID column** works well)" with no type qualifier. But `field-aggregation-governance.ts:35` — `number: ['SUM','AVG','MIN','MAX','ANY_VALUE','P25','P50','P75','P95']` — omits `COUNT` from the **supported** set entirely, and `:101-116` runs any per-field `allowedAggregations` override through `intersectWithSupported`, so a Data Mart owner cannot enable it either. Every warehouse type parser normalizes numeric types into that category before storage (`bigquery-field-type.enum.ts:45-51` `INT64→INTEGER`, `snowflake-field-type.enum.ts` `NUMBER→NUMERIC`, and the Athena/Redshift/Databricks equivalents), so there is no escape hatch. The web picker never even offers the checkbox (`aggregation-governance.ts` mirror; pinned by `aggregation-governance.test.ts:120-123`). Meanwhile **line 51 of the same docs page** already states "`COUNT` / `COUNT_DISTINCT` are **not offered for numeric fields**" — 76 lines above the new advice.
  Two further reachability limits the wording hides: on **date/time** columns `COUNT` is supported but **off by default** (`:57-58`, `['MIN','MAX']`), needing a Data-Mart-level edit first; and on a **joined** Data Mart `withoutCountBesideSleevedCountDistinct` (`field-aggregation-governance.ts:127-133`) strips `COUNT` whenever `COUNT_DISTINCT` is present — which is the default for a joined string ID. So on a `date` + numeric-metrics mart there is no eligible column at all.
  **Fix:** copy the wording your own MCP prompt already gets right — `query-data-mart.tool.ts:94` correctly says "a **non-number** field … that is always filled". Suggested: *"apply the **Count** aggregation to a **non-numeric** column that is always filled (a text ID works well — Count is not available on numeric columns, and it skips NULLs, so a nullable column undercounts); to count unique entities, use a Unique Count."* Consider promoting **Unique Count** to the primary answer: with a valid PK, `COUNT(DISTINCT pk)` per group *is* that group's row count, needs no per-field opt-in, and is already in the same sentence.
  *(Skeptic note: the MCP prompt itself is correct and is **not** part of this finding — an earlier draft claimed it contradicts itself; it does not. "Canonical case" is also overstated: OWOX's own connectors declare ID fields as STRING ~129:7, for which Count is on by default. The defect is real but is documentation precision, not a functional dead end.)*

- `.changeset/6838-no-automatic-row-count.md:7` — **"Previously, *every* aggregated report automatically included a `Row Count` column" is factually wrong.** The pre-PR predicate keyed on aggregations only: `git show 957598c04~1:…/report-like-read-plan.ts` — `explicit ?? (report.aggregationConfig?.length ?? 0) > 0`. A date-bucket-only report and a Unique-Count-only report are aggregated and never had the column — the paragraph this PR deleted said so explicitly ("`Row Count` follows the aggregate functions, not the grouping"). Owners of such reports will expect a change that never comes. Suggested: *"Previously, a report that applied any aggregate function automatically included…"*.

- `.changeset/6838-no-automatic-row-count.md` — **no impact/migration section, against this repo's own convention for an output-shape change.** The immediately preceding, structurally comparable `.changeset/6792-joined-unique-count.md` devotes a full `## Some existing reports will show different numbers` section to it and even states *"Cached report data is cleared on upgrade…"*. 6838 mentions neither Google Sheets losing a column on the next run, nor Looker Studio (inline blocker), nor the cache behaviour below.

- **Cross-repo: the Google Sheets extension still advertises the removed column.** The hint was correctly deleted from `apps/web` here, but its twin ships from `OWOX/owox-data-marts-spreadsheets` and still renders `A Row Count column is included automatically in aggregated reports.` (`src/features/reports/components/ReportColumnPicker/AggregationSettingsDropdown.tsx:112-116`, gated by `showsRowCountHint = value.aggregationConfig.length > 0` at `:100`), plus a stale rationale comment at `src/features/reports/components/ReportColumnPicker/index.tsx:575-576` ("Row Count is automatic for aggregated reports and has no opt-in toggle"). Extension users will read a promise the backend no longer keeps. Worth a paired PR there and a note on release coupling. *(Verified via the GitHub API against that repo's default branch, not statically inferrable from this one.)*

- `apps/backend/src/data-marts/data-destination-types/ee/email/services/email-report-writer.ts:286-298` — **Email / Slack / Teams / Google Chat templates that name the column by hand will hard-fail the run.** `buildPreloadedMainSource` feeds the report's `dataHeaders` into a persisted, user-authored Handlebars template; `table-tag.handler.ts:155-161` resolves `columns="…"` by exact name and **throws** `TagHandlerException` on `-1`, which `email-report-writer.ts:143-144` does not catch (the `processingError` branch at `:131` covers only the read phase). The softer half: `{{value column="Row Count"}}` renders a caution block **into the delivered message** (`value-tag.handler.ts:147-151, :236`). Applies to all four EE writers (`:423`, `:453`, `:483`, and `google-chat-report-writer.ts:268`). **Question, not a claim:** I can prove the mechanism and that `columns="…"` is a first-class persistable construct (`template-tag-renderer.service.ts:57-58, :89`), but not that such templates exist in the wild — and the AI generator was never shown the column (authoring-time sources come from raw `SELECT *`, `insight-template-source-data.service.ts:127-135`), so a hand-written template is the realistic case. Worth a `SELECT` over stored templates for `Row Count`, and possibly making `filterColumns` degrade like `ValueTagHandler` rather than throw.

- `apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts` … the **`Count` replacement silently changes the report's grain**, and neither the docs nor the MCP prompt say so. `sql-clause-renderer.ts:253-256` — a column carrying a function "is an aggregated metric — never a GROUP BY key" and is not pushed to `groupByParts`. The old Row Count was purely additive; applying `Count` to a column you also wanted as a dimension removes that dimension and merges its groups. Worth one clause in `report-aggregations.md:127` and in `query-data-mart.tool.ts:94`.

- `apps/backend/src/data-marts/services/report-data-cache.service.ts:147` — **stale Looker Studio cache: harmless on four storages, a real degradation on Databricks.** The cache is keyed on report id + expiry only, carries no config/SQL fingerprint, and a hit returns the *persisted* `dataDescription` (the freshly-resolved one computed at `:285` is discarded, then overwritten at `:287`). Invalidation fires only on a config edit. On BigQuery/Snowflake/Athena/Redshift the cached headers and the cached result set are both pre-upgrade, so it is a consistent, delayed transition. On **Databricks** it is not: `databricks-report-reader.service.ts:76-85` eagerly opens a cursor on the *newly composed* SQL while `:246` restores the *old* header list, so `mapRowsToHeaders` (`:143-168`) resolves `Row Count` to `null` and logs `Column 'Row Count' not found in query results` **once per row** until the entry expires. A purge migration in the style of `1786000000000-purge-report-data-cache-for-unique-count.ts` would close it. *(Explicitly downgraded from a proposed blocker by the skeptic pass: that precedent existed because cached **numbers were wrong** — "leaving them serves wrong numbers" — whereas here no value changes. The 400 after expiry is inevitable either way; a purge makes it happen sooner, not less.)*

<details>

- `apps/backend/src/data-marts/services/blended-report-data.service.spec.ts:2356-2374` — the two surviving tests in `describe('no rowCount flag reaches buildBlendedQuery')` are now unfalsifiable and indistinguishable from each other. `:2357` reads `(context as Record<string, unknown> | undefined)?.['rowCount']`; because `context` is optional-chained, both pass even if `buildBlendedQuery` were called with **no context argument at all** (the pre-PR `toBe(true)`/`toBe(false)` assertions could not). Their two setups differ only in `aggregationConfig` non-empty vs empty, which no longer influences the result. Suggest collapsing to one and asserting on the object: `expect(context).toBeDefined(); expect(context).not.toHaveProperty('rowCount');` — the sibling at `report-sql-composer.aggregation.spec.ts:118` already uses `not.toHaveProperty`.
- `apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.spec.ts:2577` — the renamed test no longer demonstrates its own title. With `columns: ['channel','spend__cost']` and one sleeve metric, SELECT order and header order are now byte-identical (`abstract-blended-query-builder.ts:516` vs `report-data-headers.utils.ts:63,85`), so "so SELECT order != header order" is not exercised by anything; the surviving assertions compare the sleeve pull against the **first** select item rather than the last non-sleeve one. Either rename it to what it checks ("sleeve pulls are the last select items"), or restore divergence, e.g. `columns: ['channel','spend__cost','clicks']` + a main-side `SUM(clicks)`.
- `apps/backend/src/data-marts/data-storage-types/utils/report-data-headers.utils.ts:53` — production comment still reads "the SELECT emits only the synthetic metric / **Row Count** / Unique Count columns", inside the very function whose Row Count block this commit deleted. It documents the `metricsOnly` predicate on `:57`, so it actively misdescribes a live branch.
- `apps/backend/src/data-marts/services/report-sql-composer.service.ts:86` — "the blended path below already carries any aggregation / date-trunc / **row-count**".
- `apps/backend/src/data-marts/services/report-totals.service.ts:62` — `// Unique Count is not part of the totals summary.` is orphaned: the `rowCount: false` line it annotated is gone, so it now sits above the unrelated `queryTimeoutMs` spread. Statement still true, just detached.
- `apps/backend/test/integration/query-data-mart-real.integration.ts:25, :31-32, :208` — the file docblock still lists `Row Count = 2` / `Row Count = 1` as the expected aggregated output and calls the concept "opt-in on read plans only", while `:211` asserts `not.toContain('Row Count')`. Two generations stale: `:208` described the state at `8ba9aa6fe`, which the final commit removed. The assertions were fixed; the prose explaining them was not — and this file never runs in CI, so nothing will catch it.
- `apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.spec.ts:2159-2164` — orphaned comment block ("with a sleeve metric, Row Count is NOT last … Pin the order here…") now heads the unrelated `describe('Totals restricted to the groups a metric filter keeps')`. Pre-existing on `main`, but this PR is the sweep that made its content false and rewrote every neighbouring Row Count comment.
- `apps/backend/src/data-marts/services/report-sql-composer.aggregation.spec.ts:361` and `:1098` — `expect(result!.sql).not.toContain('Row Count')` is now trivially true (nothing can emit the string), while the title at `:332` still sells "NO Row Count" as a checked guarantee. Pre-PR it was the only pin on `composeTotals`' explicit opt-out. Suggest dropping both the assertions and the clause in the title.
- `apps/backend/src/data-marts/dto/schemas/http-data-query.schema.ts:155` — the rewritten comment line is 105 columns against `printWidth: 100` (`packages/prettier-config/base.js:11`). Prettier does not reflow comments, so this will **not** fail CI; purely a consistency wart from editing the sentence in place.
- `docs/getting-started/setup-guide/report-aggregations.md:61, :84, :94` — three hosted screenshots predate the change and may still show the `Row Count` column or the deleted aggregation-panel hint. UNPROVEN: the images are remote CDN URLs, so pixel content is not statically readable; the alt texts do not mention Row Count, which is consistent but not proof. Worth a human eyeball.
- minor: `apps/web/.../AggregationSettingsDropdown.test.tsx:155` title claims it covers "or hint" but asserts nothing about the hint, and its `EMPTY_CONFIG` setup is the one case where the hint never rendered pre-PR — the next test (`:168`) is the one that actually pins it · `queryByLabelText('Add a Row Count metric')` assertions in `ReportColumnPicker.test.tsx:704,708`, `AggregationSettingsDropdown.test.tsx:164` and `OutputSettingsDropdown.test.tsx:251` probe an aria-label that never existed in this repo (pre-existing dead assertions, now guarding nothing) · `AggregationSettingsDropdown.tsx:88-105` is now a pure pass-through `div` around a single `AggregationSection` — the hint was its only other content · `http-data.api.ts:35-37` left an awkward two-fragment string concat where one line would do (not a Prettier failure — it will not merge adjacent literals) · `report-aggregations.md:125` `## Counting rows in a group` is the only sentence-case H2 on a page whose other H2s are Title Case (the rename breaks no anchor — `#auto-row-count` has zero inbound references) · the PR body and all three commit trailers carry AI attribution, against this repo's convention for GitHub content.

</details>

### What was checked and found clean

The removal itself is mechanically complete and symmetric — I could not break it. Hypotheses raised and **disproved** with the code open: a saved `sortConfig`/`filterConfig` on `"Row Count"` (the validator never let one persist — `ROW_COUNT_LABEL` appeared only in the collision-name list, never in a sort/filter allow-list, so removing it from the blended drop-list at `abstract-blended-query-builder.ts:281` is a net *fix* for a real column of that name); the `validateOutputColumnNames` positional-argument shift (both call sites at `:602` and `:929` updated correctly, and the dropped parameter was `boolean`-adjacent so a missed one would still compile — there are no others); header↔SELECT desync in any of the five branches (plain / aggregated / uniqueCount-only / totals / blended); a dangling import or removed export anywhere (`ROW_COUNT_LABEL`, `shouldIncludeRowCount`, `includeRowCount` → zero hits repo-wide); the aggregated-path predicate losing its only true disjunct (`rowCount === true ⟹ aggregations.length > 0` held in every reachable case, so the removed disjunct was already dead); an empty `SELECT` list (`AGGREGATION_REQUIRES_COLUMN_CONFIG` gates it at save *and* run); a needed DB migration for a persisted flag (it never lived on the `Report` entity); the `.passthrough()` HTTP query schema (`rowCount` was never a param); Google Sheets leaving a stale column (the writer emits `deleteDimension` for the shrunk set, and `capturedColumnFormats` is keyed by name, so surviving columns keep their formatting); `packages/api-client`, `apps/ctl`, notification templates' own copy, OpenAPI examples and e2e fixtures (no references); and the backend↔web `aggregation-labels` mirror (byte-identical after the edit).

Also worth stating plainly: the final commit `55131c912` is **pure dead-code deletion** — `shouldIncludeRowCount` was already `'rowCount' in report && report.rowCount === true` at `957598c04` with no producer setting it, so the user-visible behaviour changed in the *first*, already-reviewed commit. That materially lowers the risk of the missing CI run, though it does not remove it.

<sub>Automated OWOX review · worktree multi-agent passes (5 specialists + fresh-eyes, second wave on 2 confirmed blockers) · every blocker adversarially verified · fully static, no tests/builds run.</sub>

apps/backend/test/integration/query-data-mart-real.integration.ts
```diff
@@ -205,10 +205,10 @@ describeIfCredentials('QueryDataMartService — aggregated query on real BigQuer
     // --- columns must reflect the executed aggregated SQL, NOT the raw `fields` list ---
     // 'dim' is a GROUP BY key → stays as 'dim'.
     // 'revenue' is aggregated with SUM → becomes 'revenue | SUM'.
-    // Row Count is appended automatically for every aggregated query.
+    // No Row Count: it is opt-in on read plans only, and MCP queries never ask for it.
```
prokopetsv: 🟡 Non-blocker (Low): this comment is now inaccurate — `rowCount` is no longer "opt-in on read plans"; this PR removed the concept entirely. Suggest: `// No Row Count: the concept was removed — aggregated output has only selected columns.` The file header is stale too: lines 23–25 still list `Row Count` among the rewritten headers, and lines 31–32 still show `Row Count = 2` / `= 1` in the expected output.

apps/backend/src/data-marts/data-storage-types/utils/report-data-headers.utils.ts
```diff
@@ -110,22 +105,6 @@ export function resolveReportDataHeaders(
...
-        ROW_COUNT_LABEL,
-        undefined,
-        undefined,
-        integerTypeFor(storageType),
-        'COUNT'
-      ),
-    ];
-  }
-
   if (mainUniqueCount) {
```
prokopetsv, line 108: 🟡 Non-blocker (Low): just above this hunk, the comment at line 53 still reads "the SELECT emits only the synthetic metric / Row Count / Unique Count columns". Row Count no longer exists — drop it: "the synthetic metric / Unique Count columns".

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.spec.ts
```diff
@@ -2587,24 +2574,23 @@ describe('AbstractBlendedQueryBuilder — post-join aggregation', () => {
     });
   });
 
-  it('emits sleeve pulls AFTER Row Count, so SELECT order != header order', () => {
+  it('emits sleeve pulls AFTER the non-sleeve select items, so SELECT order != header order', () => {
```
prokopetsv, line 2571: 🟡 Non-blocker (Low): the orphaned comment fragment at lines 2159–2164 ("with a sleeve metric, Row Count is NOT last — the sleeve pulls are appended after it…") still describes the deleted Row Count ordering. This renamed test now carries that rationale in its own comment, so delete the stale fragment (the following Totals paragraph at 2165–2169 is still accurate and should stay).

docs/getting-started/setup-guide/report-aggregations.md
```diff
@@ -122,17 +122,15 @@ When a joined Data Mart cannot offer the metric, the row is still shown, disable
 | The primary key is a nested field (for example `user.id`) | Unique Count cannot key on a nested field — declare a top-level key instead. |
 | The primary key is nested **and** part of it is disconnected | Both need fixing — a top-level key, all of whose fields are connected. |
 
-## Auto Row Count
+## Counting rows in a group
 
-Whenever a report is aggregated, OWOX automatically adds a **`Row Count`** column (`COUNT(*)`) — the number of underlying rows in each group. There is no toggle; it is included because it is almost always useful when reading aggregated output.
-
-`Row Count` follows the **aggregate functions**, not the grouping. A report whose only metric is a Unique Count — its own Data Mart's or a joined one's — applies no aggregate function, so it gets no `Row Count` column even though it is grouped.
+An aggregated report contains only the columns you select — no `Row Count` column is added on its own. To see how many underlying rows each group stands for, apply the **Count** aggregate function to a column that is always filled (an ID column works well). To count unique entities instead, use a Unique Count.
```
prokopetsv: 🟡 Non-blocker (Low, docs style): the second sentence is 27 words; the docs rule caps sentences at 20. Split it: "To see how many underlying rows each group stands for, apply the **Count** aggregate function. Use a column that is always filled, such as an ID column."

prokopetsv (commented):
## Multi-perspective review (code quality, comment accuracy, test coverage, silent failures, type design, simplification)

**Summary:** This is a clean, complete removal of the Row Count concept. The `rowCount` flag, `COUNT(*)` rendering, synthetic header, and `ROW_COUNT_LABEL` constants were deleted identically across all five warehouse builders (flat, legacy, and blended paths), every call site was updated in lockstep (no dead parameters or orphaned branches), the backend/web label mirrors stay in sync, and test coverage is strong for a removal PR — explicit negative assertions (`not.toHaveProperty('rowCount')`, exact-SQL `toBe`, exact header arrays) at every layer, plus tests for the un-reserved "Row Count" column name. No persisted data carries the removed flag, and the HTTP schema's pre-existing `.passthrough()` means old clients are unaffected.

### 🔴 Blockers

None. No correctness, backward-compatibility, or architectural issues detected.

### 🟡 Non-blockers

Four inline comments, all comment/docs hygiene: three stale comments still describing the removed Row Count behavior (`query-data-mart-real.integration.ts`, `report-data-headers.utils.ts:53`, `abstract-blended-query-builder.spec.ts:2159`) and one docs sentence over the 20-word style cap.

**Informational (no code change needed):** existing Looker Studio dashboards that selected the `Row Count` field will silently lose it — the pre-existing filter in `looker-studio-connector-api-data.service.ts` quietly drops requested-but-missing fields rather than erroring. Worth a heads-up in release comms alongside the changeset.

apps/backend/src/data-marts/data-storage-types/utils/sql-clause-renderer.spec.ts
laskevych: Fixed in 488873c30 — and thanks for the precise evidence chain. Running `prettier --check` over the whole backend locally surfaced five more spec files with the same deletion artifacts (the storage clause-renderer specs and the headers spec); all six are formatted now and the workspace check passes clean.

.changeset/6838-no-automatic-row-count.md
laskevych: Great catch — addressed in 488873c30, both parts:

1. The changeset now has a "What changes in existing reports" section (following the 6792 precedent): Google Sheets structurally deletes the column on the next run and shifts formulas; Looker Studio scorecards on Row Count alone fail while mixed charts silently narrow, with Refresh fields as the remedy; and the cache-staggered rollout is called out so the trickle of failures can be correlated with the release.

2. The Looker Studio 400 now names the unresolvable requested fields and points at refreshing the data source fields (covered by a new spec). The silently-narrowed case is left as-is deliberately — dropping unknown requested fields is the connector's long-standing contract and changing it would break legitimate partial requests.

laskevych:
All four non-blocking suggestions applied in 3c4ef1399: the docs sentence is split per the 20-word rule, the orphaned sleeve-ordering comment fragment is removed (the Totals paragraph below it kept), the metrics-only comment in the header resolver no longer lists Row Count, and the live integration test's header/expectations/comment now describe the removed concept instead of an opt-in. Thanks!

PR #1537, opened 2026-08-14
Improve Google Chat setup guidance and delivery logs

AnastasiaKhukhra:
Google Chat destination settings now provide contextual instructions for choosing a delivery
method and obtaining an incoming webhook URL or space email address. Report run logs also identify
the delivery method and destination without exposing webhook credentials, with per-part entries
shown only when a report is split across multiple Google Chat messages.

### DIFF
--- a/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.spec.ts
+++ b/apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.spec.ts
@@ -31,6 +31,7 @@ describe('GoogleChatReportWriter', () => {
       },
       dataDestination: {
         id: 'destination-1',
+        title: 'Team Chat',
         type: DataDestinationType.GOOGLE_CHAT,
       },
       dataMart: {
@@ -103,13 +104,15 @@ describe('GoogleChatReportWriter', () => {
     expect(eventDispatcher.publishExternal.mock.calls[0][0].name).toBe(
       'google-chat.report.run.successfully'
     );
-    expect(executionLogger.log).toHaveBeenCalledWith({
-      type: 'google_chat_part_sent',
-      part: 1,
-      totalParts: 1,
-    });
+    expect(executionLogger.log).not.toHaveBeenCalledWith(
+      expect.objectContaining({ type: 'google_chat_part_sent' })
+    );
     expect(executionLogger.log).toHaveBeenCalledWith({
       type: 'google_chat_sent',
+      deliveryMethod: 'incoming_webhook',
+      destinationId: 'destination-1',
+      destinationTitle: 'Team Chat',
+      spaceId: 'space-1',
       messageCount: 1,
     });
   });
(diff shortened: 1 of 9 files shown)


apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.spec.ts
```diff
@@ -128,11 +131,19 @@ describe('GoogleChatReportWriter', () => {
 
     expect(executionLogger.log).toHaveBeenCalledWith({
       type: 'google_chat_part_sent',
+      deliveryMethod: 'incoming_webhook',
```
prokopetsv, line 134: 🟡 Non-Blocker — the only multi-part scenario tested is the failure path. No test covers 2+ parts all succeeding, so part 2's `google_chat_part_sent` and the final `google_chat_sent` with `messageCount: 2` are never exercised together — a regression there would pass today's suite. Suggest one test: split fixture, both sends resolve, assert `part_sent` for parts 1 and 2 plus the enriched `google_chat_sent`.

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
```diff
@@ -270,6 +270,7 @@ export class GoogleChatReportWriter extends BaseEmailReportWriter {
   readonly type = DataDestinationType.GOOGLE_CHAT;
 
   private googleChatCredentials?: GoogleChatCredentials;
+  private channelEmailRecipients?: string[];
```
prokopetsv, line 273: 🟡 Non-Blocker — delivery mode now spans three uncoordinated fields (`googleChatCredentials`, `channelEmailRecipients`, `usesEmailDelivery`). The webhook branch guards its field before use; the email branch reads `channelEmailRecipients` unguarded, so a future desync would silently log `recipients: undefined`. A single discriminated union (`{ mode: 'webhook'; credentials } | { mode: 'email'; recipients }`) makes that unrepresentable — and typing recipients as `EmailCredentials['to']` keeps the schema's non-empty guarantee instead of widening to `string[]`.

apps/backend/src/data-marts/data-destination-types/ee/google-chat/services/google-chat-report-writer.ts
```diff
@@ -340,6 +352,13 @@ export class GoogleChatReportWriter extends BaseEmailReportWriter {
...
+    const spaceId = new URL(this.googleChatCredentials.webhookUrl).pathname.split('/')[3];
+    const deliveryDetails = {
+      deliveryMethod: 'incoming_webhook',
+      destinationId: this.report.dataDestination.id,
+      destinationTitle: this.report.dataDestination.title,
+      spaceId,
+    } as const;
 
     // Incoming webhooks do not support transactions or rollback. If a later part fails,
     // earlier parts remain delivered, so record each outcome for troubleshooting.
```
prokopetsv, line 364: 🟡 Non-Blocker — this comment now overpromises: with the `messages.length > 1` gate below, a single-part send records no per-part outcome. Suggested wording: "For multi-part sends, if a later part fails, earlier parts remain delivered, so record each part's outcome for troubleshooting."

prokopetsv (commented):
## Review summary

The change does what the description says, and the risky parts check out: the `spaceId` parse cannot throw or mis-split (`GoogleChatCredentialsSchema` pins the URL shape), no key/token ever reaches the logs, the part-failure `catch` rethrows so `google_chat_sent` never logs after a failure (spec-asserted), the writer is `Scope.TRANSIENT` so no state leaks across runs, and `executionLogger` is always set before any send path. Backward compatibility holds: nothing consumes the `google_chat_*` log types, and the three other `EmailFields` callers are unaffected by the new optional prop. The new accordion components follow the established `FormDescriptions/` sibling pattern exactly. Three 🟡 inline comments.

### Optional

- Accordion tests assert trigger headings only; collapsed Radix content is unmounted in jsdom, so mis-wired content would pass. One `fireEvent.click` + content assertion would cover it.
- `EmailFields`' no-description default (Email/Slack/MS Teams) has no test file at all — pre-existing gap this PR touches.
- Changeset filename `google-chat-webhook-adjustments.md` misses the task-number prefix the project convention requires (`NNNN-slug.md`).

zapolsky (approved):

PR #1538, opened 2026-08-15
feat(data-marts): recover Google Sheets reports from a deleted sheet

prokopetsv:
# Problem

A Google Sheets report stores the destination sheet as a numeric Google gid, captured when the report is created. That gid dies the moment someone deletes the sheet, or an import re-creates the spreadsheet's sheets with fresh ids — a sheet with the same name is still a different sheet. From then on every run failed with `Failed to find sheet 0 in spreadsheet 11IQ…`, which names an internal id the reader has no way to act on, and offers no way forward.

The failure is also expensive and silent. Destination access is validated only when a report is created or edited, so nothing re-checks it before a scheduled run: the warehouse query executes first and the write fails afterwards, burning a full scan every morning. The only escape was manual — open the spreadsheet, create a tab, copy its URL with the correct `#gid=`, paste it into the report's Document Link field.

# Solution

Two halves: make the failure explain itself, then give it a one-click repair.

- **One source of truth for the message.** The text was duplicated across the run-time writer and the setup-time access validator. It now lives in `google-sheet-not-found.error.ts` as two builders: `sheetNotFoundMessage` for a failing run (names the repair action), and `sheetNotFoundSetupMessage` for the create/edit form (points at the document-link field, since the repair action doesn't exist there — on create, neither does the report). The gid stays, moved to the end for support.

- **Repair by name, not by id.** `POST /reports/:id/google-sheets/reconnect-sheet` binds the report to a sheet named after it: an existing sheet with that name is reused, otherwise one is created, and the new gid is persisted. Name is the only handle a user recognises; renaming the sheet afterwards stays safe because the report tracks the gid. Reuse-before-create matters — the common repair is "someone deleted `Sheet1` and made a new one by hand", where the wanted name already exists under a different gid, and Google rejects duplicate titles anyway.

- **Never rebind a healthy report.** Before rebinding, the service checks whether the report's current gid still resolves. If it does, nothing is written and the response carries `changed: false`. Without this, a report failing for an unrelated reason (quota, timeout) could be moved onto a same-named sheet a user maintains by hand, and the next run would overwrite it.

- **One press delivers data.** The row menu offers **Reconnect & run**, which repairs the destination and immediately runs the report. It appears only after a failed run, so a curiosity click on a healthy report can't reach it. An earlier iteration used a confirmation dialog and a two-step toast with a Run button; both were dropped once the server-side guard made the action safe — the dialog asked for a sheet name the user never needed to choose.

- **Persistence is scoped.** `ReportService.updateDestinationConfig` issues a targeted `UPDATE` rather than saving the whole entity, so it cannot roll back `lastRunStatus` / `runsCount` written by a concurrent run.

Trade-off worth flagging for review: the menu item gates on `lastRunStatus === ERROR`, not on the specific sheet-not-found failure, because the frontend only receives the error message string. Tightening it needs an error code carried end to end — `BusinessViolationException` has a `code` field, but `AccessValidationException` and run-error persistence drop it. The server-side guard makes the coarse gate safe in the meantime.

# Changes

- Added `sheetNotFoundMessage` and `sheetNotFoundSetupMessage` in `google-sheet-not-found.error.ts`, replacing three duplicated inline strings in the report writer and access validator
- Added `spreadsheetNotAccessibleMessage` for the distinct case where the spreadsheet itself cannot be opened
- Added `GoogleSheetNotFound` error details (`spreadsheetId`, `sheetId`) alongside the message
- Added `ReconnectGoogleSheetService`, its command and request/response DTOs, mapper method, OpenAPI spec, controller route `POST /reports/:id/google-sheets/reconnect-sheet`, and module registration
- Added a guard that skips rebinding when the report's current sheet still resolves, reported as `changed: false`
- Added sheet-title sanitization: trimmed, capped at Google's 100-character limit, apostrophes replaced (the writer builds unescaped `'title'!A1` ranges), blank falls back to `Report data`
- Added `GoogleSheetsApiAdapter.addSheet` and `findSheetByTitle`, both null-checking gid rather than testing falsiness so gid 0 is handled correctly
- Added `ReportService.getByIdAndProjectIdWithDestination` and `updateDestinationConfig`, keeping repository access below the use-case layer
- Added a **Reconnect & run** item to the Google Sheets report row menu, shown only when the last run failed and gated on `canEditConfig`, which reconnects, toasts the outcome, runs the report, and refreshes the row
- Added `reportService.reconnectSheet` and its response DTO type on the web side
- Changed the run-status tooltip to show the error message alone instead of a `Fail` heading above it
- Added tests: 8 cases for the reconnect use case (reuse, create, title fallback, sanitization, access check ordering, healthy-report guard, non-Sheets rejection), 3 for the row action, and one for the writer's missing-sheet error
- Added changeset `6790-reconnect-sheet-and-run.md`

# Screenshots

**1. The failure explains itself and names the fix**
Row status tooltip on a failed report reading "Can't find the sheet this report writes to. Someone deleted it, or an import replaced the sheets of this spreadsheet. Use 'Reconnect & run' on the report to repair it and refresh the data." followed by the spreadsheet and sheet IDs

**2. "Reconnect & run" appears in the row menu of a failed report**
Report row context menu open on a failed report, listing Run report, Edit report, Reconnect & run, and Delete report

**3. One click creates the sheet, reconnects, and starts the run**
Success toast reading "Created sheet 'New report'. Your data is loading into it now." above the report row, whose status now shows a spinner for the run in progress

**4. A healthy report's menu has no reconnect option**
Report row context menu on a successful report, listing only Run report, Edit report, and Delete report — Reconnect & run is absent


apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
```diff
@@ -0,0 +1,130 @@
...
+      return {
+        spreadsheetId,
+        sheetId: report.destinationConfig.sheetId,
+        sheetTitle: currentSheet.properties.title,
+        created: false,
+        changed: false,
+      };
+    }
+
+    const existingSheetId = adapter.findSheetByTitle(spreadsheet, title)?.properties?.sheetId;
```
laskevych, line 134: 🟡 **Important (blocking):** `toSheetTitle` rewrites `'` → `’` *before* this lookup, so reuse-by-title can never match a sheet whose name contains a straight apostrophe. Take the exact repair scenario the class doc describes: report titled `Bob's data`, user hand-recreated a tab literally named `Bob's data`. The lookup searches for `Bob’s data` (U+2019), misses, `addSheet` creates a second visually identical tab (Google's duplicate-title check doesn't fire — different codepoints), and the report binds to the new *empty* tab while the user's hand-made sheet is silently abandoned. This also breaks the request DTO's promise "An existing sheet with this title is reused."

Cheapest fix inside this PR: look up the raw title first, and only sanitize when creating. The deeper fix (escaping in the writer, see my comment there) would remove the need for the apostrophe swap entirely.

apps/backend/src/data-marts/data-destination-types/google-sheets/services/google-sheets-report-writer.ts
```diff
@@ -663,14 +667,16 @@ export class GoogleSheetsReportWriter implements DataDestinationReportWriter {
...
+            spreadsheetNotAccessibleMessage(this.destination.spreadsheetId, error.message),
+            { spreadsheetId: this.destination.spreadsheetId }
           );
         });
       const sheet = this.adapter.findSheetById(spreadsheet, this.destination.sheetId);
 
       if (!sheet) {
         throw new GoogleSheetNotFound(
-          `Failed to find sheet ${this.destination.sheetId} in spreadsheet ${this.destination.spreadsheetId}`
+          sheetNotFoundMessage(this.destination.spreadsheetId, this.destination.sheetId),
```
laskevych, line 678: 🟡 **Important (non-blocking, but the changeset claim needs adjusting):** the apostrophe swap in the reconnect service exists because this writer builds A1 ranges as `'${sheetTitle}'!A1:...` unescaped (lines ~337, ~386, ~1026, plus 4 sites in the adapter) — but sanitizing one entry path doesn't protect the others. The writer re-reads the tab title by gid every run, so a user renaming the tab to `Bob's data` in Google Sheets breaks every subsequent run with an unparseable-range 400 — while the changeset says "Renaming the sheet afterwards is safe: the report tracks the sheet by its ID." Worse, "Reconnect & run" is a dead end there: the gid still resolves, so it returns `changed: false` and the rerun fails identically.

The right-depth fix is escaping at range construction (A1 grammar: double embedded apostrophes, `title.replace(/'/g, "''")` in one shared helper covering all interpolation sites). That fixes every title source at once and lets `toSheetTitle` drop the swap, which also resolves the reuse-by-title issue. Fine as an immediate follow-up if out of scope here — but then the changeset sentence about renaming being safe should be softened.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
```diff
@@ -83,6 +89,51 @@ export function GoogleSheetsActionsCell({
...
+      const outcome = !result.changed
+        ? `The report is already connected to the sheet "${result.sheetTitle}"`
+        : result.created
+          ? `Created sheet "${result.sheetTitle}"`
+          : `Reconnected to the existing sheet "${result.sheetTitle}"`;
+      toast.success(`${outcome}. ${dataNote}`, { duration: RECONNECT_TOAST_DURATION_MS });
+      // runReport refreshes the report, starts status polling and toasts
+      // "Report run started"; it catches its own failures, so a failed start
+      // surfaces through the row status.
+      setIsRunning(true);
```
laskevych, line 126: 🟡 **Important (non-blocking — pre-existing pattern, but the new comment codifies a false premise):** if the run fails to *start* after a successful reconnect, the row is stuck on a disabled "Running report…" until page reload. `useReport.runReport` swallows its own errors (its catch logs only — no rethrow, no state change), so the `catch` below is unreachable for a run failure; `lastRunStatus` stays `'ERROR'`, so the sync effect keyed on it never re-fires and `isRunning` is never released. The comment "it catches its own failures, so a failed start surfaces through the row status" is not what the code does — a failed start changes no row state (the axios interceptor does toast HTTP errors, but a pure network failure is fully silent).

`handleRun` has the same latent bug, so consider fixing it at the source: have `runReport` return a success flag (or rethrow) and reset `isRunning` on failure. At minimum, please correct the comment.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
```diff
@@ -83,6 +89,51 @@ export function GoogleSheetsActionsCell({
     }
   }, [canRun, onRunSuccess, runReport, row.original.id]);
 
+  // Rebinds the report to a sheet named after it (reuse or create), then runs.
+  // Renaming the sheet in Google Sheets later is safe — the report stores the gid.
+  const handleReconnectSheet = useCallback(async () => {
+    if (!canEditConfig || isReconnecting) return;
```
laskevych: 🟡 **Important (non-blocking):** this gates on `canEditConfig` only, while `handleRun` gates on `canRun`. These capabilities genuinely diverge: a user with data-mart maintenance access but no USE on the destination gets `canEditConfig: true, canRun: false`. For them, reconnect succeeds (the backend checks mutate access only), the toast says "Your data is loading into it now.", then `/run` 403s — a persistent forbidden toast right after a success toast, plus the stuck "Running report…" state from the other comment, and no data ever loads.

Since the action is "Reconnect **& run**", it should probably require `canRun` as well, here and in the `disabled` condition.

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
```diff
@@ -0,0 +1,130 @@
...
+  ) {}
+
+  async run(command: ReconnectGoogleSheetCommand): Promise<ReconnectGoogleSheetResponseDto> {
+    const report = await this.reportService.getByIdAndProjectIdWithDestination(
+      command.reportId,
+      command.projectId
+    );
+
+    // Changing where a report writes is a config mutation, not an operation.
+    await this.reportAccessService.checkMutateAccess(
```
laskevych: 🔵 **Question:** reconnect requires only `checkMutateAccess`, while `/run` requires `checkOperateAccess` (destination USE). A user the run endpoint would 403 can still trigger `addSheet` — a structural write into the customer spreadsheet with the stored OAuth/SA credentials. I see the precedent (update-report allows config mutation with mutate-only access, and the access validator already makes read calls with stored credentials), but `addSheet` is the first *write* on the mutate-only path. Was operate-level access considered here, given the class doc frames the click as consent to write into the sheet?

apps/web/src/features/data-marts/reports/list/components/StatusIcon/StatusIcon.tsx
```diff
@@ -72,15 +69,16 @@ export function StatusIcon({ status, error, className }: StatusIconProps) {
           <Icon
             className={cn('h-5 w-5', color, className)}
             role='img'
-            aria-label={getAccessibleDescription()}
+            aria-label={errorMessage ?? label}
```
laskevych: 🔵 **Question (accessibility, non-blocking):** with an error message present, the word "Fail" is no longer conveyed anywhere textually — both the tooltip and `aria-label` become the raw error text, and ERROR/CANCELLED/RESTRICTED share the same `XCircle` glyph differentiated only by color (this component is also used by the Email/Insight/project-level tables, which render no visible status text). The visual tooltip change is an improvement, but could we keep the status in the accessible name, e.g. ``aria-label={errorMessage ? `${label}: ${errorMessage}` : label}``, so screen-reader users still hear that the run failed?

apps/backend/src/data-marts/dto/presentation/google-sheets/reconnect-google-sheet-request.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+  @ApiPropertyOptional({
+    example: 'Revenue by channel',
+    description:
+      'Title of the sheet (tab) to reconnect to. An existing sheet with this title is reused; ' +
+      'otherwise it is created. Falls back to the report title when omitted.',
+  })
+  @IsOptional()
+  @IsString()
+  @MaxLength(MAX_SHEET_TITLE_LENGTH)
+  title?: string;
```
laskevych: 🔵 **Question (non-blocking):** the optional `title` parameter has no caller — the UI always sends `{}` — and its two validation authorities diverge: this DTO rejects >100 chars with a 400 while `toSheetTitle` silently truncates, and the DTO accepts straight apostrophes the service rewrites. Is a caller planned? If not, a bodyless POST would delete this DTO, a command field, a mapper argument and the untested branch; if kept, the DTO validation and `toSheetTitle` should agree (and the `MAX_SHEET_TITLE_LENGTH` constant duplicated here and in the service could be one shared export).

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
```diff
@@ -0,0 +1,130 @@
...
+    const title = toSheetTitle(command.title?.trim() || report.title);
+
+    const adapter = await this.adapterFactory.createFromDestination(report.dataDestination);
+    if (!adapter) {
+      throw new BadRequestException(
+        'No authentication method available for Google Sheets: neither OAuth nor Service Account credentials found'
+      );
+    }
+
+    const spreadsheet = await adapter.getSpreadsheet(spreadsheetId).catch((error: Error) => {
```
laskevych, line 101: 💡 **Suggestion (non-blocking):** this blanket `.catch` turns *any* `getSpreadsheet` failure — including transient Google 429/5xx (note `executeWithRetry` only retries quota errors) — into "Check that it still exists, and that the connected Google account has Editor access" with HTTP 400. During a brief Sheets outage the user is sent to fix sharing that was never broken. The adapter already has an `extractStatus` helper — discriminating 404/403 (current message) from 429/5xx ("temporary problem, try again") would make the remediation trustworthy. The writer has the same pre-existing wrap, so fine as a follow-up.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
```diff
@@ -83,6 +89,51 @@ export function GoogleSheetsActionsCell({
...
+      // One shared tail so the variants can't drift apart. `changed: false` means
+      // the report's sheet was alive and the backend left it alone — the run below
+      // is then the whole point, so don't claim a repair that didn't happen.
+      const dataNote = 'Your data is loading into it now.';
+      const outcome = !result.changed
+        ? `The report is already connected to the sheet "${result.sheetTitle}"`
+        : result.created
+          ? `Created sheet "${result.sheetTitle}"`
+          : `Reconnected to the existing sheet "${result.sheetTitle}"`;
+      toast.success(`${outcome}. ${dataNote}`, { duration: RECONNECT_TOAST_DURATION_MS });
```
laskevych, line 121: 🧹 **Nit (non-blocking):** the toast fires before `runReport` is called and asserts "Your data is loading into it now." on every branch — including `changed: false` after a non-sheet failure (SQL error, quota, credentials), where the immediate rerun will most likely fail again with the same error. "Starting a run now" claims only what is known at that point. The `changed: false` toast text is also untested (tests cover `changed: true` only).

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
```diff
@@ -0,0 +1,130 @@
...
+const DEFAULT_SHEET_TITLE = 'Report data';
+
+/**
+ * Google accepts almost any character in a sheet name but caps it at 100 and
+ * rejects empty. Apostrophes are swapped for typographic ones because the report
+ * writer builds A1 ranges as `'${title}'!A1:...` without escaping quotes — a
+ * straight apostrophe in the name would corrupt every range on the next run.
+ */
+function toSheetTitle(raw: string): string {
+  const cleaned = raw.replace(/'/g, '’').trim().slice(0, MAX_SHEET_TITLE_LENGTH).trim();
```
laskevych: 🧹 **Nit (non-blocking):** `slice(0, 100)` counts UTF-16 code units and can bisect a surrogate pair (emoji at the boundary), producing a malformed title on the `report.title` fallback path — which has no length validation (`@Column()` + `@IsString()` only), so titles longer than 100 units are reachable. `Array.from(title).slice(0, MAX_SHEET_TITLE_LENGTH).join('')` is code-point safe. Narrow edge, but it would fail precisely on the one-click repair flow.

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
```diff
@@ -0,0 +1,130 @@
...
+  constructor(
+    private readonly reportService: ReportService,
+    private readonly reportAccessService: ReportAccessService,
+    private readonly adapterFactory: GoogleSheetsApiAdapterFactory
+  ) {}
+
+  async run(command: ReconnectGoogleSheetCommand): Promise<ReconnectGoogleSheetResponseDto> {
+    const report = await this.reportService.getByIdAndProjectIdWithDestination(
+      command.reportId,
+      command.projectId
```
laskevych, line 64: 💡 **Suggestion (non-blocking):** `getByIdAndProjectIdWithDestination` and `checkMutateAccess` run the identical joined `findOne` back to back — the access service internally re-loads the report with the same relations. `ReportAccessService` already has a private `evaluateMutateAccessForReport` documented as operating on an already-loaded report to avoid the extra round-trip; exposing a `checkMutateAccessForReport(userId, roles, report)` and passing the entity loaded here would drop the duplicate query.

laskevych (commented):
Reviewed the change end to end — backend reconnect flow, the writer, the frontend action, and the tests. Overall this is a careful, well-documented PR: the healthy-report guard, the targeted `UPDATE` for `destinationConfig`, and the gid-0 null checks all hold up under scrutiny.

There is **one blocking issue** (the apostrophe sanitization defeats the reuse-by-title contract — see the inline comment) plus a handful of non-blocking remarks and questions, so I'm holding off on approval until the blocking one is addressed. Everything else is marked non-blocking and can be taken here or as follow-ups.

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
prokopetsv: Fair question, and on reflection the precedent doesn't cover this: update-report's validator only reads, while `addSheet` is a structural write into the customer's spreadsheet with stored credentials. Reconnect now requires operate access in addition to mutate access — `checkOperateAccessForReport` runs right after the mutate check, so the server demands the same consent `/run` does, matching the UI's `canEditConfig && canRun` gating. Tested: an operate rejection blocks before any Google call. b19cd712c.

apps/web/src/features/data-marts/reports/list/components/StatusIcon/StatusIcon.tsx
prokopetsv: Agreed and implemented exactly as proposed: the accessible name is now `${label}: ${errorMessage}` while the visible tooltip stays message-only. Left a comment in the component explaining why the status word must stay in the aria path — the ERROR/CANCELLED/RESTRICTED glyphs differ only by color. b19cd712c.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
prokopetsv: Reworded to claim only what is known when the toast fires: every branch now ends with "Starting a run to load your data." instead of asserting the data is already loading. The `changed: false` text is also covered by a new test (*tells the truth when the sheet was already connected*). b19cd712c.

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
prokopetsv: Agreed — this was real, and fixed via the deeper route from your writer comment rather than the lookup-first workaround: apostrophes are no longer rewritten anywhere. `toSheetTitle` keeps them as typed, so reuse-by-title now matches a hand-recreated sheet literally named `Bob's data`, and no visually-identical duplicate can be created. Regression test added: *reuses a hand-made sheet whose name contains a straight apostrophe*. Fixed in b19cd712c.

apps/backend/src/data-marts/data-destination-types/google-sheets/services/google-sheets-report-writer.ts
prokopetsv: Implemented at the depth you suggested, in this PR rather than as a follow-up: `quoteA1SheetTitle` in the adapter doubles embedded apostrophes per the A1 grammar, and every interpolation site — 3 in the writer, 5 in the adapter — now goes through it. With escaping at construction, the apostrophe swap in `toSheetTitle` is gone (which also resolves the reuse-by-title issue), and the changeset's "renaming is safe" sentence is now actually true for any title a user can type. Helper unit-tested in the adapter spec. b19cd712c.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
prokopetsv: Confirmed — the comment codified a false premise and the state was genuinely stuck. Fixed at the source as you suggested: `runReport` now returns whether the run started (it still tracks/logs its own errors), and both `handleRun` and `handleReconnectSheet` release `isRunning` on a failed start instead of relying on the unreachable catch / never-firing sync effect. The misleading comment is gone. New test: *releases the running state when the run fails to start*. b19cd712c.

apps/web/src/features/data-marts/reports/list/components/GoogleSheetsReportsTable/GoogleSheetsActionsCell.tsx
prokopetsv: Agreed — the action's second half is a run, so it now requires `canRun` in both the handler guard and the `disabled` condition. The backend enforces the same consent server-side (operate access on reconnect — see the other thread), so the mutate-only user you describe gets a consistent 403 instead of a success toast followed by a forbidden one. b19cd712c.

apps/backend/src/data-marts/dto/presentation/google-sheets/reconnect-google-sheet-request.dto.ts
prokopetsv: No caller is planned, so I took the deletion branch: the endpoint is now a bodyless POST — request DTO deleted, command field, mapper argument and the fallback branch removed. The name always derives from the report title, so there is exactly one validation authority left (`toSheetTitle`), and the DTO/service divergence you list can no longer exist. b19cd712c.

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
prokopetsv: Implemented in this PR: the reconnect path discriminates via `GoogleSheetsApiAdapter.httpStatusOf` — 429/5xx now throws `GoogleApiException` ("Google Sheets is temporarily unavailable. Please try again in a few minutes.", 502) and only the remaining cases keep the exists/Editor-access remedy. Both branches tested. The writer's identical pre-existing wrap I left for the follow-up, as you suggested. b19cd712c.

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
prokopetsv: Fixed — the cap now counts code points: `Array.from(title).slice(0, MAX_SHEET_TITLE_LENGTH).join('')`. Tested with an emoji straddling the 100 boundary (99 chars + surrogate pair): the pair is kept whole, never bisected. b19cd712c.

apps/backend/src/data-marts/use-cases/google-sheets/reconnect-google-sheet.service.ts
prokopetsv: Done — exposed `checkMutateAccessForReport` (and the symmetric `checkOperateAccessForReport`, needed for the operate-access thread) as thin throwing wrappers over the existing private evaluators, and the use case passes the report it already loaded with both relations. One `findOne` total for the whole request. b19cd712c.

laskevych (approved):

PR #1539, opened 2026-08-15
fix(deps): bump nanoid to 3.3.18 (GHSA-2v37-7h3g-55p8)

laskevych:
## Why

The **Audit all** check fails on every branch (including `main`) because of a high-severity advisory: nanoid <3.3.18 — custom generators can loop indefinitely when size is zero ([GHSA-2v37-7h3g-55p8](https://github.com/advisories/GHSA-2v37-7h3g-55p8)). This blocks the check on every open PR.

## What

Bump the single transitive `nanoid` lockfile entry from 3.3.17 to 3.3.18. The dependency comes in via `@owox/connectors -> vite -> postcss`, and postcss's `^3.3.11` range already allows 3.3.18, so no `package.json` changes are needed — the diff is 3 lines in `package-lock.json` (version, resolved, integrity, as written by `npm audit fix`).

## Verification

- `npm audit --omit=dev` (the exact Audit all CI command): 0 vulnerabilities
- `npm audit --audit-level=high`: 0 vulnerabilities
- `npm ci --dry-run`: lockfile resolves cleanly
- `npm run lint:root` and `npm run format:root:check`: pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29721,9 +29721,9 @@
       "optional": true
     },
     "node_modules/nanoid": {
-      "version": "3.3.17",
-      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.17.tgz",
-      "integrity": "sha512-xQLf0A3HOMlgHq0n247/LRuAOYmB7dXJ/DvAxGvsSBij45XtBSmQycu+F8ODbHwns/XyFZagyL1+J0Offw1E0g==",
+      "version": "3.3.18",
+      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.18.tgz",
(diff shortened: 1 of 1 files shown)


prokopetsv (approved):
**Approve** ✅ — verified end to end:

- **Advisory correctly targeted** — GHSA-2v37-7h3g-55p8: infinite loop (CWE-835, CVSS 8.2) in nanoid's `customAlphabet`/`customRandom` with size 0; <3.3.18 affected, 3.3.18 is the exact fix version.
- **Supply chain clean** — one lockfile entry changed; `resolved` stays on `registry.npmjs.org`, and the new `integrity` hash matches the npm registry's published `sha512-DTg4…` for 3.3.18 byte-for-byte.
- **Correctness** — patch bump within the same 3.x major of a transitive dep (`vite → postcss → nanoid`); postcss's `^3.3.11` admits 3.3.18, so the lockfile-only change with no `package.json` edit is right.

🔴 Blocker: none
🟡 Non-Blocker: none

PR #1540, opened 2026-08-17
feat(data-marts): add an optional description to data mart relationships

laskevych:
## What

Each join between data marts can now carry an optional free-text description explaining the business meaning of the relationship (e.g. *"Visitors from the website sign up for the product and convert into users"*).

## Why

A join definition carries only data context: which fields match. An AI assistant working through MCP could see *which* fields a joined data mart contributes, but not *how and why* the two data marts relate — which limits reasoning like root-cause analysis across joined data. An analyst-written relationship description closes that gap, and is also useful documentation for humans reading the Data Setup page.

## How

**Backend**
- New nullable `description` (text) column on `DataMartRelationship` + migration.
- `POST`/`PATCH /data-marts/:id/relationships` accept an optional `description`; responses return it. `null` or a blank string clears it; values are trimmed before storing.

**Web (Data Setup)**
- A joined data mart's card gets a new **Description** tab next to Join Settings, with an autosaving textarea (same debounce/save pattern as SQL Alias / Join Fields).
- Inherited (transitive) joins render it read-only with a banner linking to the owning data mart.

**MCP**
- `get_data_mart_details_by_id` with `detail_level=with_joined_fields` now returns a `joins` array — one entry per join edge with the source/target data mart titles, the alias path, the join key fields, and the relationship description. Only sources the caller may report on are exposed, matching the existing joined-fields gate.
- System instructions tell the model to read `joins` before interpreting joined fields.

## Tests

- Mapper, relationship service (set/trim/clear/untouched), MCP facade (joins exposure + inaccessible-source filtering), and details tool specs updated/added — all passing.
- Web `RelationshipAccordionItem` tests passing; backend build typecheck and web typecheck clean.

### DIFF
--- a/apps/backend/src/data-marts/dto/presentation/create-relationship-request-api.dto.ts
+++ b/apps/backend/src/data-marts/dto/presentation/create-relationship-request-api.dto.ts
@@ -2,7 +2,7 @@ import {
   IsString,
   IsArray,
   IsNotEmpty,
-  IsOptional,
+  ValidateIf,
   ValidateNested,
   MinLength,
   MaxLength,
@@ -76,7 +76,10 @@ export class CreateRelationshipRequestApiDto {
     description: 'Business meaning of this relationship, shared with AI assistants',
     required: false,
   })
+  // Not @IsOptional(): that would skip validation for an explicit null too, smuggling a null
+  // into the `string | undefined` command field. Omitting the key is fine; null is a 400 —
+  // unlike the update DTO, which deliberately opts INTO null as its "clear" signal.
+  @ValidateIf(obj => obj.description !== undefined)
   @IsString()
-  @IsOptional()
   description?: string;
 }
--- a/apps/backend/src/data-marts/dto/presentation/create-relationship-request-api.dto.ts
+++ b/apps/backend/src/data-marts/dto/presentation/create-relationship-request-api.dto.ts
@@ -2,6 +2,7 @@ import {
   IsString,
   IsArray,
   IsNotEmpty,
+  IsOptional,
   ValidateNested,
   MinLength,
   MaxLength,
@@ -69,4 +70,13 @@ export class CreateRelationshipRequestApiDto {
   @ValidateNested({ each: true })
   @Type(() => JoinConditionApiDto)
   joinConditions: JoinConditionApiDto[];
+
+  @ApiProperty({
+    example: 'Visitors from the website sign up for the product and convert into users',
+    description: 'Business meaning of this relationship, shared with AI assistants',
+    required: false,
+  })
+  @IsString()
+  @IsOptional()
+  description?: string;
 }
--- a/apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
+++ b/apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
@@ -229,29 +229,39 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
    * one relationship (`relationshipId`), including transitive ones, so one lookup by ids covers
    * the whole tree. Gives the model the join keys and the analyst-written relationship
    * description (#6780) that `joinedFields` alone cannot carry.
+   *
+   * Failure here degrades only `joins` to [] — by this point the blendable schema has already
+   * resolved, and this extra lookup must not take the joined fields down with it.
    */
   private async resolveJoins(
     accessibleSources: Array<{ aliasPath: string; relationshipId: string }>
   ): Promise<McpJoinDto[]> {
-    const relationshipsById = new Map(
-      (await this.relationshipService.findByIds(accessibleSources.map(s => s.relationshipId))).map(
-        rel => [rel.id, rel]
-      )
-    );
+    try {
+      const relationshipsById = new Map(
+        (
+          await this.relationshipService.findByIds(accessibleSources.map(s => s.relationshipId))
+        ).map(rel => [rel.id, rel])
+      );
 
-    const joins: McpJoinDto[] = [];
-    for (const source of accessibleSources) {
-      const rel = relationshipsById.get(source.relationshipId);
-      if (!rel?.sourceDataMart || !rel.targetDataMart) continue;
-      joins.push({
-        aliasPath: source.aliasPath,
-        sourceDataMart: rel.sourceDataMart.title,
-        targetDataMart: rel.targetDataMart.title,
-        joinConditions: rel.joinConditions,
-        ...(rel.description ? { description: rel.description } : {}),
-      });
+      const joins: McpJoinDto[] = [];
+      for (const source of accessibleSources) {
+        const rel = relationshipsById.get(source.relationshipId);
+        if (!rel?.sourceDataMart || !rel.targetDataMart) continue;
+        joins.push({
+          aliasPath: source.aliasPath,
+          sourceDataMart: rel.sourceDataMart.title,
+          targetDataMart: rel.targetDataMart.title,
+          joinConditions: rel.joinConditions,
+          ...(rel.description ? { description: rel.description } : {}),
+        });
+      }
+      return joins;
+    } catch (err) {
+      this.logger.warn(
+        `resolveJoins failed; returning joined fields without joins: ${err instanceof Error ? err.message : String(err)}`
+      );
+      return [];
     }
-    return joins;
   }
 
   async queryDataMart(
--- a/apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
+++ b/apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
@@ -25,6 +25,7 @@ import {
   McpDataMartsFacade,
   McpDataMartDetailsResponse,
   McpGetDataMartDetailsRequest,
+  McpJoinDto,
   McpJoinedFieldDto,
   McpListDataMartsRequest,
   McpListDataMartsResponse,
@@ -93,9 +94,9 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
         } as DataMartSchema) as { fields: Array<Record<string, unknown>> })
       : undefined;
 
-    const { joinedFields, uniqueCountSources } = request.includeJoinedFields
+    const { joinedFields, joins, uniqueCountSources } = request.includeJoinedFields
       ? await this.resolveJoinedFields(request, this.topLevelFieldNames(schema?.fields ?? []))
-      : { joinedFields: [], uniqueCountSources: [] };
+      : { joinedFields: [], joins: [], uniqueCountSources: [] };
 
     return {
       id: dataMart.id,
@@ -103,6 +104,7 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
       description: dataMart.description ?? '',
       fields: this.withDisplayNames(schema?.fields ?? []),
       joinedFields,
+      joins,
       uniqueCountSources,
     };
   }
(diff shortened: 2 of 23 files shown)


apps/web/src/features/data-marts/edit/components/DataMartRelationships/JoinDescriptionForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+        dataMartId,
+        relationship.id,
+        // An all-whitespace description is a cleared one: send null so the backend stores NULL.
+        { description: value.trim() === '' ? null : value },
+        { skipErrorToast: true, skipLoadingIndicator: true }
+      )
+      .then(updated => {
+        lastSavedRef.current = updated.description ?? '';
+        isDirtyRef.current = false;
+        onSaved(updated);
```
prokopetsv, line 86: 🔴 Blocker — silent loss of typed text. Type → debounce fires save A (slow PATCH) → keep typing → the next debounced save early-returns on `isSaving`. When save A resolves, this line clears `isDirtyRef` unconditionally, `onSaved` updates the parent, and the `savedValue` effect above runs `setLocalValue(savedValue)` — visibly reverting the textarea to the stale server value with no error and no recovery (blur is guarded by the now-false dirty flag). Fix: clear the dirty flag only when `localValue` still equals the value that was sent, and skip the prop-sync reset while `isDirtyRef.current` is true.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/JoinDescriptionForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+        setIsSaving(false);
+      });
+  };
+
+  useEffect(() => {
+    if (!isDirtyRef.current) return;
+    save(debouncedValue);
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- run only when the debounced value settles
+  }, [debouncedValue]);
+
```
prokopetsv, line 137: 🟡 Non-Blocker — a debounced save skipped mid-flight is never retried: `save()` early-returns on `isSaving`, and this effect only re-runs when `debouncedValue` changes, so dirty text sits unsaved until blur. The sibling `JoinSettingsForm` includes `isSaving` in its analogous effect's deps precisely to retry once the prior save settles. Add `isSaving` to the deps (and drop the eslint-disable) to match.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/JoinDescriptionForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+  dataMartId,
+  readOnly = false,
+  inheritedFrom,
+  onSaved,
+}: JoinDescriptionFormProps) {
+  const { scope } = useProjectRoute();
+
+  const savedValue = relationship.description ?? '';
+  const [localValue, setLocalValue] = useState(savedValue);
+  const debouncedValue = useDebounce(localValue, 800);
```
prokopetsv, line 35: 🟡 Non-Blocker — an edit inside the 800ms window is discarded on unmount: the Description `TabsContent` unmounts when the user switches tabs or collapses the row, and `useDebounce`'s cleanup clears the pending timer. Saving then depends on the textarea happening to blur before Radix swaps panels — not a guaranteed ordering. Flush explicitly in an unmount cleanup (`if (isDirtyRef.current) save(latestValueRef.current)`).

apps/backend/src/data-marts/facades/mcp-data-marts.facade.impl.ts
```diff
@@ -149,6 +152,8 @@ export class McpDataMartsFacadeImpl implements McpDataMartsFacade {
       );
       const accessiblePaths = new Set(accessibleSources.map(s => s.aliasPath));
 
+      const joins = await this.resolveJoins(accessibleSources);
```
prokopetsv, line 155: 🟡 Non-Blocker — a failure in this new lookup takes down the old payload: `resolveJoins` → `findByIds` runs inside the method-wide catch, so a transient DB error in the joins query returns `{ joinedFields: [], joins: [], uniqueCountSources: [] }` even though the blendable schema already resolved — a regression path that didn't exist before this PR. Wrap `resolveJoins` in its own try/catch degrading only `joins: []`, and tag the warn log with which stage failed.

apps/backend/src/data-marts/dto/presentation/create-relationship-request-api.dto.ts
```diff
@@ -69,4 +70,13 @@ export class CreateRelationshipRequestApiDto {
...
   joinConditions: JoinConditionApiDto[];
+
+  @ApiProperty({
+    example: 'Visitors from the website sign up for the product and convert into users',
+    description: 'Business meaning of this relationship, shared with AI assistants',
+    required: false,
+  })
+  @IsString()
+  @IsOptional()
+  description?: string;
```
prokopetsv, line 84: 🟡 Non-Blocker — `@IsOptional()` skips validation for `null` as well as `undefined`, so `POST { description: null }` passes even though the declared type is `string?` — a null smuggled into `CreateRelationshipCommand.description: string | undefined`. Harmless today (`normalizeDescription` uses optional chaining), but future code trusting the signature can throw. Swap for `@ValidateIf(obj => obj.description !== undefined)` so an explicit null 400s, mirroring how the update DTO deliberately opts INTO null.

prokopetsv (changes requested):
## Review summary

The backend is in good shape: the tri-state description semantics (undefined = untouched, null/blank = clear, trimmed store) are enforced consistently end-to-end and well-tested; the migration matches the repo's idempotent pattern exactly; the MCP `joins` gating reuses the accessibility filter correctly (spec-pinned), and the eager-loaded relations make the `findByIds` edge resolution sound in production. The blocker is in the new web autosave: two independent review passes traced silent loss of typed text on ordinary latency. One 🔴 and five 🟡 inline.

### Test coverage (🟡)

- `JoinDescriptionForm` / the Description tab have zero web coverage — `RelationshipAccordionItem.test.tsx` never opens the tab, so a broken save flow or missing tab passes the suite. One test file with fake timers (type → advance 800ms → assert `updateRelationship` payload; whitespace → `description: null`; `inheritedFrom` → banner + disabled textarea) covers the advertised behavior.
- The class-validator layer for `PATCH { description: null }` is untested (works today — `@IsOptional()` skips null — but nothing pins it).

### Optional

- This is the third hand-rolled debounce/dirty/save machine in `DataMartRelationships/` (alias block in `RelationshipAccordionItem`, `JoinSettingsForm`, now this). Rule of three is met — a shared `useDebouncedFieldAutosave` hook would absorb two of them near line-for-line and would be the natural place to fix the 🔴 once.
- The PATCH DTO doc says "null or an empty string" clears — whitespace-only strings also clear (`normalizeDescription`); one clause in the API description closes the gap.
- `description` has no `@MaxLength` — consistent with the data mart description DTO, so follow-up material, not this PR.

laskevych:
Addressed in 6b2f479f3:

- 🔴 **Silent loss of typed text** — the form now tracks the latest typed value separately from the in-flight payload: the dirty flag is cleared only when the textarea still matches what was sent, and the server-sync effect skips resets while dirty. Pinned by a regression test (slow save + typing during flight → textarea keeps the newer text, newer text is saved next).
- 🟡 **Skipped save never retried** — the autosave effect now also fires when `isSaving` settles, retrying any edit typed during the request (same shape as `JoinSettingsForm`); the eslint-disable is gone.
- 🟡 **Unmount discards pending edit** — an unmount cleanup flushes the dirty value (bypassing the in-flight guard, last-write-wins on the server). Covered by a test that unmounts before the debounce fires.
- 🟡 **`resolveJoins` failure took down the whole payload** — it now catches internally and degrades only `joins` to `[]`, with a stage-tagged warn log; the outer catch keeps its original role.
- 🟡 **`POST { description: null }`** — create DTO now uses `@ValidateIf(obj => obj.description !== undefined)` so an explicit null 400s, with a comment contrasting the update DTO's deliberate null-clear. Both contracts are pinned by validation tests (update: string / null / omitted all pass; create: null rejected).
- 🟡 **Test coverage** — new `JoinDescriptionForm.test.tsx` (fake timers): debounce → payload, whitespace → `description: null`, slow-save regression, unmount flush, inherited banner + disabled textarea, read-only no-op.

Optional items: took the PATCH doc clarification (whitespace-only also clears). Leaving the shared `useDebouncedFieldAutosave` extraction and `@MaxLength` as follow-ups — the hook refactor touches `JoinSettingsForm` and the alias block, which I'd rather not fold into this PR.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/JoinDescriptionForm.tsx
```diff
@@ -0,0 +1,168 @@
...
+    if (!isDirtyRef.current || isSaving) return;
+    saveRef.current(latestValueRef.current);
+  }, [debouncedValue, isSaving]);
+
+  // Flush a pending edit when the tab unmounts (tab switch, row collapse) — the
+  // debounce timer dies with the component, and blur is not guaranteed to fire first.
+  useEffect(() => {
+    return () => {
+      if (isDirtyRef.current) {
+        saveRef.current(latestValueRef.current, { ignoreInFlight: true });
```
amarchenk0: The cleanup bypasses `isSaving` and starts a second PATCH while the first one is still in flight. The server has no version/CAS check, so the two PATCHes can be processed out of order: the user types B, switches tabs, B is saved, and the delayed PATCH A overwrites it with stale text. Queue the flush behind the active save (and cover slow A → type B → unmount) instead of bypassing the in-flight guard.

amarchenk0 (changes requested):
A blocking autosave race remains in the current head.

laskevych:
Fixed in ea4de5e5b: the unmount flush no longer bypasses the in-flight guard — it chains onto the active request's settlement (`inFlightRef.then(flushPending)`), re-checking the dirty flag afterwards since that save may already have covered the edit, so at most one PATCH is ever on the wire. The read-only guard also moved into the sender itself, so a flush can never fire for a read-only form.

Covered by the suggested test: slow save A → type B → unmount — asserts B's PATCH does not start while A is in flight and goes out with the latest text once A settles.

prokopetsv (approved):
**Approve** ✅ — re-reviewed at `ea4de5e5b`; all seven outstanding findings are resolved, each with a regression test, and CI is fully green (including `Audit all`).

- 🔴 autosave clobber: dirty flag now cleared only when the textarea still matches the sent value, and the prop-sync effect skips resets while dirty — pinned by the *"keeps text typed during a slow save and retries it afterwards"* spec.
- 🟡 retry-on-completion: `isSaving` added to the debounce effect's deps.
- 🟡 unmount flush: cleanup flushes pending edits, queued **behind** `inFlightRef` so a second PATCH can never race the first (also resolves @amarchenk0's out-of-order finding, with a dedicated test).
- 🟡 `resolveJoins` isolated: a joins-only failure now degrades only `joins: []` with a stage-tagged log; joined fields survive.
- 🟡 create DTO: `@ValidateIf(obj => obj.description !== undefined)` replaces `@IsOptional()`, so an explicit null 400s — verified through real `class-validator` in the new DTO tests.
- 🟡 coverage: `JoinDescriptionForm.test.tsx` (8 cases) plus DTO validation via `plainToInstance` + `validate()` for create and update.
- Doc: whitespace-only clearing now stated in the PATCH description.

Still fine as follow-ups: the shared `useDebouncedFieldAutosave` hook (the fixed form is the natural extraction template) and `@MaxLength` on description fields.

🔴 Blocker: none
🟡 Non-Blocker: none — remaining items are acknowledged follow-ups

amarchenk0 (approved):

PR #1542, opened 2026-08-17
fix(web): stop storage setup step from clearing fields while editing

prokopetsv:
# Problem

On the connector setup step "Choose where to store your data", editing the table name could wipe the already-filled dataset (or database) field. Once the table field was empty, filling the dataset became impossible — every keystroke disappeared immediately. The step's `useEffect` re-parsed `target.fullyQualifiedName`, a value the component itself had just emitted upstream. While the table name was empty, that FQN was partial (`mydataset.`), the parse produced no dataset part, and the effect overwrote the dataset state with an empty string — the dataset field was the only one without a "user edited" guard.

# Solution

Treat local field state as the source of truth once the user touches any field. The seeding `useEffect` now returns early after the first manual edit, so it only runs for its real purpose: filling defaults on first open and parsing an existing target when editing a saved connector. The change handlers already push the complete target upstream on every keystroke, so nothing is lost by skipping the echo. This fixes all five storage types (BigQuery, Athena, Snowflake, Redshift, Databricks) in one place; the previous per-field guards became dead code and were removed. A test harness feeds the emitted target back into the component's prop, reproducing the real parent loop — two of the three new tests fail without the fix.

# Changes

- Added a `datasetEditedByUser` ref and an early return in the seeding effect of `TargetSetupStep` once any field has been manually edited
- Removed the now-redundant per-field `*EditedByUser` guards and the edited-flag conditions in `shouldUpdate`
- Added `TargetSetupStep.test.tsx` with an echo harness covering the dataset-wipe regression, dataset input while the table is empty, and Snowflake default seeding
- Added changeset `6811-connector-storage-step-field-reset.md`

### DIFF
--- a/.changeset/6811-connector-storage-step-field-reset.md
+++ b/.changeset/6811-connector-storage-step-field-reset.md
@@ -1,0 +1,9 @@
+---
+'owox': minor
+---
+
+# Editing the table name no longer clears the other storage fields
+
+On the connector setup step "Choose where to store your data", clearing the table name also wiped the dataset (or database) name. Typing a dataset name was impossible while the table field stayed empty — each keystroke disappeared.
+
+The step now stops re-reading its own emitted value once you edit any field. Defaults still fill in on first open, and an existing target still loads when you edit a saved connector. The fix applies to every storage type: Google BigQuery, AWS Athena, Snowflake, AWS Redshift, and Databricks.
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/TargetSetupStep.test.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/TargetSetupStep.test.tsx
@@ -1,0 +1,62 @@
+import { TargetSetupStep } from './TargetSetupStep.tsx';
+import { fireEvent, render, screen } from '@testing-library/react';
+import { MemoryRouter } from 'react-router';
+import { useState } from 'react';
+import { DataStorageType } from '../../../../../data-storage';
+
+type Target = { fullyQualifiedName: string; isValid: boolean } | null;
+
+// Mirrors ConnectorEditForm: the target the step emits is fed straight back as a prop.
+// That echo is what used to wipe the fields, so the test harness must reproduce it.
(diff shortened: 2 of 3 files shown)


laskevych (approved):
Nice, surgical fix — the echo-loop diagnosis is spot on, and the test harness that feeds the emitted target back into the prop is exactly the right way to pin this regression. Verified locally: all 3 new tests pass on the branch, 2 of 3 fail against the previous component version, and lint + `tsc -b` are clean. No blocking remarks, approving. A few non-blocking notes:

🟡 **Important (non-blocking):** The wipe can still reproduce across step navigation. Repro: clear the table name (the emitted target becomes the partial FQN `tiktok_ads_owox.`), click **Back**, then **Next**. The step remounts, the `*EditedByUser` refs reset, and the seeding effect re-parses the echoed partial FQN — it splits into a single part, no parse branch matches, and since `target.fullyQualifiedName` is truthy the defaults branch is skipped, so every field ends up empty (dataset included; for Snowflake the schema too). I confirmed this with a small unmount/remount harness. This path pre-dates this PR, so not blocking — but since it's the same user-visible symptom, it may be worth a follow-up. One option: emit an empty `fullyQualifiedName` when a required part is missing (as the Redshift/Databricks branches of `updateTarget` already do), so a remount falls back to seeding defaults instead of rendering empty required fields.

💡 **Suggestion:** Now that the seeding effect bails on the first edit of *any* field, the per-field granularity of `datasetEditedByUser` / `catalogEditedByUser` / `schemaEditedByUser` / `tableEditedByUser` is unused — they are only ever read OR-ed together. A single `editedByUser` ref set in all four handlers would express the same invariant with less state to keep in sync. Fine to skip if you prefer the field symmetry.

💡 **Suggestion:** `TargetSetupStep.test.tsx` uses `describe` / `it` / `expect` without importing them from `vitest`. It runs (`globals: true`) and typechecks today, but every other test file in `apps/web` imports them explicitly, and the typecheck currently only succeeds via ambient types picked up from hoisted `@types` packages. Adding the explicit import keeps the suite consistent and removes that hidden dependency.

PR #1545, opened 2026-08-18
fix(microsoft-ads): prevent out-of-memory crash when importing keywords

prokopetsv:
# Problem

Connector runs for the Microsoft Ads `campaigns` node failed on large accounts with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`, killing the runner process mid-import (observed on an account with 4,085 campaigns and 32k ad groups). The root cause was in the Keywords bulk-download path: after downloading a zip for a batch of 50 campaigns, the connector held the full CSV text, a full parsed row array (doubled again by `concat`), a full record-object array, and a filtered copy in memory at the same time — roughly 5× the file size in simultaneous allocations, which exceeded Node's default ~2 GB heap.

# Solution

Stream keyword records to storage in small chunks instead of materializing whole batches:

- `Helper.js` gains `processCsvRecordsFromUrl()` / `processCsvTextInChunks()`, which parse the downloaded CSV line by line and flush record objects to a callback every 2,000 records. The full row array and full object array are never built, so peak memory drops to the CSV text plus one chunk.
- `Source.js` threads an optional `onRecordsChunk` callback through `_downloadEntity` → `_downloadEntityBatch` → `_fetchEntityByCampaigns`, so each chunk flows through `filterByFields` straight into the storage `onBatchReady`. Per-batch record counts are still logged after each batch completes.
- The existing 100 MB batch-halving retry, the Campaigns/AssetGroups/AdGroups path, and the reporting path are intentionally untouched — they keep their current return-array contract. A streaming unzip was considered and rejected as unnecessary given Microsoft's own 100 MB per-file cap.

Partial chunks saved before a mid-download failure are safe: the BigQuery MERGE is idempotent on `uniqueKeys`, so a re-run converges.

# Changes

- Added `processCsvRecordsFromUrl()` and `processCsvTextInChunks()` to the Microsoft Ads helper for chunked CSV-to-record conversion with identical parsing semantics to the legacy `csvRowsToObjects` pipeline (header sanitization, quote handling, `Format Version` row skipping)
- Added an optional `onRecordsChunk` streaming callback to `_downloadEntity` and `_downloadEntityBatch`
- Replaced whole-batch buffering in `_fetchEntityByCampaigns` (both the normal and the halved-batch retry path) with chunked streaming to `onBatchReady`
- Added `test/Sources/MicrosoftAds/Source.test.js` covering legacy-parity parsing, chunk-boundary flushing, header-only files, streaming through `onBatchReady`, and the 100 MB batch-halving retry (5 tests; full suite 135/135 green)

### DIFF
--- a/.changeset/6853-microsoft-ads-keywords-oom.md
+++ b/.changeset/6853-microsoft-ads-keywords-oom.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Memory-safe Keywords import for Microsoft Ads
+
+Previously, importing Keywords from a large Microsoft Ads account crashed the run with an out-of-memory error. The connector now streams records to storage in small chunks instead of loading the whole download at once. Imports of any account size now complete within normal memory limits.
--- a/packages/connectors/src/Sources/MicrosoftAds/Helper.js
+++ b/packages/connectors/src/Sources/MicrosoftAds/Helper.js
@@ -78,8 +78,9 @@ const MicrosoftAdsHelper = {
   /**
    * Download and unzip a CSV file, converting rows to objects in chunks that are
    * flushed to the callback as soon as they fill up. Unlike downloadCsvRows +
-   * csvRowsToObjects, this never holds the full row array or full object array in
-   * memory — only the CSV text of one file and one chunk at a time.
+   * csvRowsToObjects, this never builds the full cell-array or record-object
+   * arrays. Peak memory is one file's CSV text plus its split line strings
+   * (V8 slices sharing the parent string's memory) plus one chunk of records.
    * @param {string} url
    * @param {function(Array<Object>): Promise<void>} onRecordsChunk
    * @param {number} [chunkSize=2000]
@@ -99,10 +100,10 @@ const MicrosoftAdsHelper = {
    * the header (sanitized like csvRowsToObjects); 'Format Version' rows are skipped.
    * @param {string} csvText
    * @param {function(Array<Object>): Promise<void>} onRecordsChunk
-   * @param {number} chunkSize
+   * @param {number} [chunkSize=2000]
    * @returns {Promise<void>}
    */
-  async processCsvTextInChunks(csvText, onRecordsChunk, chunkSize) {
+  async processCsvTextInChunks(csvText, onRecordsChunk, chunkSize = 2000) {
     const lines = csvText.split('\n');
     let headerNames = null;
     let chunk = [];
--- a/packages/connectors/src/Sources/MicrosoftAds/Helper.js
+++ b/packages/connectors/src/Sources/MicrosoftAds/Helper.js
@@ -76,6 +76,65 @@ const MicrosoftAdsHelper = {
   },
 
   /**
+   * Download and unzip a CSV file, converting rows to objects in chunks that are
+   * flushed to the callback as soon as they fill up. Unlike downloadCsvRows +
+   * csvRowsToObjects, this never holds the full row array or full object array in
+   * memory — only the CSV text of one file and one chunk at a time.
+   * @param {string} url
+   * @param {function(Array<Object>): Promise<void>} onRecordsChunk
+   * @param {number} [chunkSize=2000]
+   * @returns {Promise<void>}
+   */
+  async processCsvRecordsFromUrl(url, onRecordsChunk, chunkSize = 2000) {
+    const response = await HttpUtils.fetch(url);
+    const blob = await response.getBlob();
+    const files = FileUtils.unzip(blob);
+    for (const file of files) {
+      await this.processCsvTextInChunks(file.getDataAsString(), onRecordsChunk, chunkSize);
+    }
+  },
+
+  /**
+   * Convert CSV text to record objects chunk by chunk. The first non-empty line is
+   * the header (sanitized like csvRowsToObjects); 'Format Version' rows are skipped.
+   * @param {string} csvText
+   * @param {function(Array<Object>): Promise<void>} onRecordsChunk
+   * @param {number} chunkSize
+   * @returns {Promise<void>}
+   */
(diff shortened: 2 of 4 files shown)


laskevych (approved):
Ran the new test file locally (5/5 green) and traced the streaming path end to end, including driving `_fetchEntityByCampaigns` with a stubbed `_downloadEntityBatch` to check the retry path empirically. No blocking issues — approving. Everything below is non-blocking.

🟢 **Praise:** Great root-cause writeup, and the optional-callback threading is a clean mechanism — the legacy return-array contract stays untouched for the paths that don't need streaming, and the legacy-parity test pins the parsing semantics.

🟡 **Important (non-blocking, pre-existing):** After a 100MB halving retry, the outer loop re-fetches part of the range the retry already covered. The retry loop processes the full failed range `[i, i + batchSize)` in halves, but `batchSize = newBatchSize` makes the outer `i += batchSize` advance by the *halved* size, so campaigns in `[i + newBatchSize, i + oldBatchSize)` are downloaded and streamed to storage a second time (with the default 50→25 halving that's 25 campaigns per halving event). I verified this with a stubbed `_downloadEntityBatch`: 20 campaigns with one halving event produced 21 fetches and 21 records delivered to `onBatchReady`. The MERGE on `uniqueKeys` makes the data converge, so the impact is duplicate downloads and inflated "Fetched N" logs rather than corruption — and the logic predates this PR — but since this function was rewritten here (and the new halving test can't see it because it collects IDs into a `Set`), it may be worth advancing `i` past the retried range while you're in it.

🔵 **Question (non-blocking):** `error.message.includes('100MB')` now guards a block in which `onBatchReady` (i.e. `storage.saveData`) also executes, because chunks stream from inside `_downloadEntityBatch`. If a storage-side error message ever contains "100MB" (a payload-size error, or `pollUntilStatus` appending the raw API response JSON to the message), the batch would be halved and re-downloaded, re-saving chunks that were already flushed. MERGE keeps it convergent, so this is mostly theoretical — did you consider distinguishing "download too large" errors from callback errors before entering the retry?

🔵 **Question (non-blocking):** Two intentional-looking parity divergences worth pinning down: (1) for a zip with multiple CSV entries, the legacy pipeline used only the first file's header and emitted each subsequent file's header row as a junk data record (`{Type:'Type', Id:'Id', …}`), while `processCsvRecordsFromUrl` correctly re-derives the header per file; (2) a header-only/empty file made the legacy `csvRowsToObjects` throw on `filteredRows[0].map`, while the new code completes with zero records. Both are improvements, but the parity test only covers the single-file case — worth a note and/or a multi-file test so the difference is documented as intentional?

🔵 **Question (non-blocking):** Has the originally affected large account been re-run end-to-end on this branch? The Campaigns/AssetGroups/AdGroups path still buffers all three entity types in `allRecords` plus a full `filterByFields` copy before its single `onBatchReady` call — for an account big enough to OOM on Keywords, that's the next candidate. I see the PR description scopes this out intentionally; just checking a follow-up is on the radar.

💡 **Suggestion (non-blocking):** Give `processCsvTextInChunks` its own `chunkSize = 2000` default. The parameter currently has no default, and `chunk.length >= undefined` is always false — a future caller using the natural two-argument shape would silently buffer the entire file into one giant chunk, reintroducing exactly the OOM this PR fixes, with all small-file tests still green.

💡 **Suggestion (non-blocking, tests):** Both `_fetchEntityByCampaigns` tests stub `_downloadEntityBatch`, so nothing exercises the actual streaming branch in `_downloadEntity` (the `if (onRecordsChunk)` early return) or the callback pass-through in `_downloadEntityBatch`. A small wiring test with mocked fetch/poll/unzip would catch a regression that silently falls back to the buffering path — which would restore the OOM without failing any test.

🧹 **Nitpick (non-blocking):** Two doc drifts: the `processCsvRecordsFromUrl` comment says only "the CSV text of one file and one chunk at a time" stay in memory, but `csvText.split('\n')` also materializes an array of all line strings for the duration of the loop (the genuine win — never building the full record-object array — is worth stating precisely so a future OOM hunt isn't misdirected); and `_downloadEntityBatch`'s `@returns` wasn't updated to mention it returns `[]` when `onRecordsChunk` is provided, unlike the sibling `_downloadEntity` doc.

prokopetsv:
**Fixed in code:**

1. **🟡 Double fetch after halving** (valid, reviewer-verified) — restructured `_fetchEntityByCampaigns` into a while-loop that advances `i` past the entire retried range (`i = retryRangeEnd`) instead of by the halved size. Regression-tested: the halving test now counts fetches per campaign with a Map (the old `Set` hid duplicates, as the reviewer noted) and asserts every campaign is fetched exactly once.
2. **🔵 Storage errors matching '100MB'** (valid edge) — `onBatchReady` failures are now tagged (`isStorageError`) inside the streaming callback, and the halving guard checks `!error.isStorageError` before matching `'100MB'`. New test proves a BigQuery error mentioning 100MB propagates as a failure instead of triggering a bogus re-download.
3. **💡 Missing `chunkSize` default** (valid — `chunk.length >= undefined` is always false, so a two-argument caller would have buffered the whole file) — `processCsvTextInChunks` now defaults to 2000 like its sibling.
4. **💡 Untested streaming branch** — added a wiring test that stubs `HttpUtils.fetch`/`FileUtils.unzip` and drives the real `_downloadEntity` → `pollUntilStatus` → `processCsvRecordsFromUrl` chain, asserting chunks stream and `[]` is returned. It uses a two-file zip, which also **documents the multi-file per-file-header divergence (🔵) as intentional**.
5. **🧹 Doc drifts** — the Helper comment now states peak memory precisely (CSV text + line-string slices + one chunk; the win is never building the full record arrays), and `_downloadEntityBatch`'s `@returns` mentions the empty-array-when-streaming contract.

amarchenk0 (approved):

PR #1548, opened 2026-08-19
feat(idp): add Microsoft NAA extension identity flow

amarchenk0:
## Summary

- add a feature-gated, host-neutral Microsoft NAA extension exchange and revoke API
- verify Entra access tokens for the configured OWOX delegated API audience and scope
- delegate verified Microsoft identity provisioning and project-scoped token issuing to Integrated Backend and the existing ROI `/Users/identity/` flow
- reuse the existing `/auth/api/projects` endpoint after the initial project token is issued
- keep Google Sheets and browser authentication contracts unchanged

## Behavior

- without `project_id`, ROI selects an existing provisioned project or creates the initial free project, matching the Google Sheets extension flow
- with `project_id`, IB provisions/touches the user and then validates current membership and roles for that exact project
- project switching repeats the NAA assertion exchange with an explicit `project_id`
- refresh and revoke accept only tokens carrying `authFlow=extension`
- an Entra identity without `email` plus `xms_edov=true` returns the recoverable `unknown_identity` result for the future dialog fallback

## Security and scope

- validate RS256 signature, Microsoft tenant issuer consistency, `aud`, `scp`, `exp`, `nbf`, `iat`, `oid`, and `tid`
- never use `preferred_username` as verified email for NAA provisioning
- apply exact-origin CORS only when the feature is enabled
- do not add project-neutral tokens, identity-session families, local identity/account linking, assertion replay tables, or process-local rate limiting
- do not add an internal-feature changeset; it will be included with the main feature

## Deployment dependencies

This is the ODM part of Fibery [6833 — Add-in Identity OAuth Flow](https://owox.fibery.io/Sprint_Work/Work_Item/Add-in-Identity-OAuth-Flow-6833).

Keep the feature disabled until the matching IB `/auth-flow/extension/microsoft-identity` C2C endpoint and Entra audience/scope/origin configuration are deployed. No Analytics change is required because IB reuses the existing ROI identity provisioning endpoint.

## Validation

- `npm test -w @owox/idp-owox-better-auth -- --runInBand` — 45 passed suites, 291 passed tests, 1 skipped suite/test
- `npm run typecheck -w @owox/idp-owox-better-auth`
- `npm run lint -w @owox/idp-owox-better-auth`
- `npm run format:check -w @owox/idp-owox-better-auth`
- `git diff --check`
- IB targeted controller/service/legacy contract tests — 43 passed
- IB Prettier and Checkstyle — passed with 0 violations


romandubovyi (commented):
Multi-agent review of the Microsoft NAA extension identity flow (Antigravity CLI, Kimi CLI, Grok CLI and Claude Code analysed the branch independently; findings were merged, deduplicated and re-validated against the code at `402d5aef`).

Two blocking points, both in the security properties the description claims: the revoke path dispatches on an upstream error contract instead of the locally-known token type, and the assertion replay record expires before the assertion itself stops verifying. The two non-blocking points are an error-classification concern in the Entra verifier and a question about the shape of IB project refresh tokens.

The overall structure — assertion consumed before identity lookup, durable `(oid, tid)` identity, email linking gated on `xms_edov` — reads well.

packages/idp-owox-better-auth/src/services/auth/extension-auth-service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  async refreshProjectToken(refreshToken: string): Promise<AuthResult> {
+    return this.tokenFacade.refreshExtensionProjectToken(refreshToken);
+  }
+
+  async revoke(refreshToken: string): Promise<void> {
+    try {
+      await this.tokenFacade.revokeExtensionSession(refreshToken);
+    } catch (error) {
+      if (!(error instanceof AuthenticationException)) throw error;
```
romandubovyi: 🔴 **Blocking:** `revoke()` dispatches on an exception rather than on the token type: it calls the IB identity-session revoke first and falls back to `revokeExtensionProjectToken` only for `AuthenticationException`, which `IdentityOwoxClient.handleAxiosError` raises for HTTP 401 only. RFC 7009 recommends a revocation endpoint answer 200 for a token it does not recognise — if the IB session-revoke route does that, a project refresh token is never revoked while this endpoint still returns `204`. A 400 from IB (`IdentityApiException`) bypasses the fallback too. The token type is determinable locally: `OwoxTokenFacade.assertExtensionProjectToken` already parses `payload.authFlow`. Please dispatch on the parsed token type instead of relying on an undefined upstream error contract.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Grok CLI (Grok 4.6) + Claude Code (Opus 5)

packages/idp-owox-better-auth/src/services/auth/microsoft-entra-access-token-verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+      const replayKey = `sha256:${createHash('sha256').update(replayMaterial).digest('hex')}`;
+      const verifiedEmail = claims.xms_edov === true ? normalizeEmail(claims.email) : undefined;
+
+      return {
+        accountId: `${claims.oid}:${claims.tid}`,
+        oid: claims.oid,
+        tid: claims.tid,
+        ...(verifiedEmail ? { verifiedEmail } : {}),
+        replayKey,
+        expiresAt: new Date(claims.exp * 1000),
```
romandubovyi: 🔴 **Blocking:** The replay row is persisted with `expiresAt = claims.exp`, but `jwtVerify` above still accepts the same assertion until `exp + clockTolerance` (`5s` by default). `consumeExtensionAssertion` deletes rows with `expiresAt <= now` before inserting (`mysql-database-store.ts:299`, `sqlite-database-store.ts:299`), so in the `(exp, exp + clockTolerance]` window the row is already gone, the `INSERT` succeeds, and a previously consumed assertion mints a second session. Please persist `exp + clockTolerance` so the replay record outlives the acceptance window that verification actually grants.

packages/idp-owox-better-auth/src/services/auth/microsoft-entra-access-token-verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+      };
+    } catch (cause) {
```
romandubovyi, line 117: ⚪ **Non-blocking:** The whole of `verify()` is wrapped in one catch that maps every failure to `AuthenticationException('invalid_assertion')`. That swallows key-resolution and transport failures too — `createRemoteJWKSet` fetch errors, `JWKSTimeout`, `JWKSNoMatchingKey` — so a Microsoft JWKS outage or an egress problem surfaces to every add-in user as an invalid token and drives a re-auth loop. Because the controller logs anything under 500 at `info`, there is also no error-level signal for it. Consider rethrowing key-resolution/transport failures as a 5xx (`IdpFailedException`) and keeping `invalid_assertion` for genuine token problems.

packages/idp-owox-better-auth/src/facades/owox-token-facade.ts
```diff
@@ -190,4 +237,36 @@ export class OwoxTokenFacade {
...
+      accessToken: response.accessToken,
+      refreshToken: response.refreshToken,
+      accessTokenExpiresIn: response.accessTokenExpiresIn,
+      refreshTokenExpiresIn: response.refreshTokenExpiresIn,
+    };
+  }
+
+  private async assertExtensionProjectToken(token: string): Promise<void> {
+    const payload = await this.tokenService.parse(token);
+    if (payload?.authFlow !== 'extension') {
```
romandubovyi, line 236: ⚪ **Non-blocking:** `assertExtensionProjectToken` runs a *refresh* token through `TokenService.parse`, which is documented as the access-token verifier and maps the payload through `IdpOwoxPayloadSchema` — where `projectId`, `userEmail`, `userFullName`, `projectTitle` and `roles` are all required. Neither `refreshToken()` nor the Google Sheets refresh path parses refresh tokens locally. If IB project refresh tokens are opaque or carry a different claim set, `parse` returns `null` and every `/project-token` refresh and scoped revoke will 401 with `invalid_project_refresh_token`. The unit test cannot catch this: `owox-token-facade.extensionAuth.test.ts:38` replaces `tokenService.parse` with a mock returning `{ authFlow: 'extension' }`, a shape that would never satisfy the real schema. Could you confirm the IB refresh token is this exact JWT, or gate `authFlow` via introspection instead?

packages/idp-owox-better-auth/src/services/auth/extension-auth-service.ts
romandubovyi: Verified as fixed at `2584e3aa`. `revokeProjectToken` now delegates straight to `revokeExtensionProjectToken`, which calls `assertExtensionProjectToken` first — so the dispatch is on the locally parsed `authFlow`, not on an upstream error contract. Thanks.

One consequence worth a look, raised separately: `IdentityOwoxClient.revokeToken` collapses every axios failure into `{ success: false }`, so the new `IdpFailedException` fires for an already-revoked or expired token too, and a repeat sign-out returns 500 instead of 204.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Grok CLI (Grok 4.6) + Claude Code (Opus 5) + Kimi CLI

packages/idp-owox-better-auth/src/services/auth/microsoft-entra-access-token-verifier.ts
romandubovyi: Accepting this as a deliberate scope decision — the replay table is gone entirely and the rewritten description lists "do not add ... assertion replay tables" explicitly, so the invariant I was checking against no longer exists. The rationale holds: the assertion is a bearer access token, so replaying it within its lifetime yields the same identity the first call already granted.

packages/idp-owox-better-auth/src/services/auth/microsoft-entra-access-token-verifier.ts
romandubovyi: Verified as fixed at `2584e3aa`. `isMicrosoftJwksFailure` now routes JWKS timeouts, invalid JWKS payloads, non-200 JWKS responses and fetch-level `TypeError`s to `IdpFailedException` (500), leaving `invalid_assertion` for genuine token problems.

I also see `JWKSNoMatchingKey` is deliberately kept as a 401 and pinned by its own test — that is a reasonable call, since `createRemoteJWKSet` re-fetches on an unknown `kid` after its cooldown, so an unknown key really does point at the token rather than at Microsoft being down. No further change needed from my side.

packages/idp-owox-better-auth/src/facades/owox-token-facade.ts
romandubovyi: Reopening: this one is unchanged at `2584e3aa`. `assertExtensionProjectToken` (`owox-token-facade.ts:234`) still runs the refresh token through `TokenService.parse`, and `owox-token-facade.extensionAuth.test.ts` still mocks `tokenService.parse` to return `{ authFlow: 'extension' }` — a shape `IdpOwoxPayloadSchema` would reject, since `projectId`, `userEmail`, `userFullName`, `projectTitle` and `roles` are all required there. So the assumption that an IB project refresh token parses as that JWT is still untested, and the whole `/auth/api/extension` refresh and revoke surface depends on it.

The rewrite raised the stakes slightly: `refreshExtensionProjectToken` and `revokeExtensionProjectToken` are now the only two paths through this check, so if the assumption is wrong both fail closed with `invalid_project_refresh_token`.

Related, same line: `TokenService.parse` swallows every failure into `null`, so an Identity OWOX JWKS outage also surfaces here as a 401 `invalid_project_refresh_token` — the same class of problem you just fixed on the Microsoft side in `2584e3aa`.

Happy to close this if you can confirm the IB refresh token really is that JWT shape.

🤖 Generated with Grok CLI (Grok 4.6) + Kimi CLI

apps/owox/src/web/cors.ts
```diff
@@ -42,14 +42,26 @@ export function buildCorsAllowedHeaders(
...
+
 export function buildCorsConfig(): CorsOptions {
-  const googleSheetsExtensionOrigin = process.env.GOOGLE_SHEETS_EXTENSION_ORIGIN;
-  const allowedOrigins = googleSheetsExtensionOrigin
-    ? googleSheetsExtensionOrigin
-        .split(',')
-        .map(origin => origin.trim())
-        .filter(origin => origin.length > 0)
-    : [];
+  const allowedOrigins = parseCorsAllowedOrigins(
```
romandubovyi, line 66: 🔴 **Blocking:** These origins are only trimmed, while the IDP config runs the same `IDP_OWOX_EXTENSION_ALLOWED_ORIGINS` value through `normalizeOrigin` → `new URL(v).origin` (`idp-owox-config.ts`), and `idp-owox-config.test.ts` relies on that by collapsing `https://addin.owox.test/` to the bare origin. Browsers send `Origin` with no path and no trailing slash, and the `cors` package compares the array by exact string equality. Because `cors(corsConfig)` is installed at `serve.ts:208`, before the IDP routes are registered, and terminates preflight itself, `ExtensionAuthController`'s `express.options` handlers never run in the real app — this raw list alone decides the preflight outcome. So `IDP_OWOX_EXTENSION_ALLOWED_ORIGINS=https://addin.example.com/` (or an add-in SourceLocation with a path, which is what operators tend to copy) passes config validation and the controller allowlist, but the browser preflight comes back without `Access-Control-Allow-Origin` — breaking `/auth/api/projects`, the endpoint this allowlist was lifted here to serve. Please normalize with `new URL(origin).origin` here too so both layers agree; none of the 8 cases in `test/web/cors.test.ts` currently covers a non-canonical origin.

apps/owox/src/web/cors.ts
```diff
@@ -42,14 +42,26 @@ export function buildCorsAllowedHeaders(
...
+    process.env.GOOGLE_SHEETS_EXTENSION_ORIGIN,
+    process.env.IDP_OWOX_EXTENSION_ALLOWED_ORIGINS
```
romandubovyi: ⚪ **Non-blocking:** `buildCorsConfig()` adds `IDP_OWOX_EXTENSION_ALLOWED_ORIGINS` to the app-wide allowlist unconditionally — `IDP_OWOX_EXTENSION_MICROSOFT_ENABLED` is never read anywhere in `apps/owox`. The description states that exact-origin CORS applies "only when the feature is enabled", and the deployment section recommends rolling out the Entra/origin configuration *before* enabling the flow — which is exactly the window where these origins get `credentials: true` access to every route while `/auth/api/extension*` is not even registered (`owox-better-auth-idp.ts`). The blast radius is limited by the refresh cookie being `sameSite: 'lax'`, but gating the origin parsing on the enabled flag would make the code match the property the description claims.

🤖 Generated with Claude Code (Opus 5) + Kimi CLI

packages/idp-owox-better-auth/src/facades/owox-token-facade.ts
```diff
@@ -125,9 +126,34 @@ export class OwoxTokenFacade {
...
+
+  async refreshExtensionProjectToken(refreshToken: string): Promise<AuthResult> {
+    await this.assertExtensionProjectToken(refreshToken);
+    return this.refreshToken(refreshToken);
+  }
+
+  async revokeExtensionProjectToken(refreshToken: string): Promise<void> {
+    await this.assertExtensionProjectToken(refreshToken);
+    const revoked = await this.requestTokenRevocation(refreshToken);
+    if (!revoked) {
```
romandubovyi, line 144: ⚪ **Non-blocking:** Failing loudly instead of silently returning 204 is the right direction, but `IdentityOwoxClient.revokeToken` collapses *every* axios failure into `{ success: false }` (`IdentityOwoxClient.ts:294-297`), including a 400/404 for a token IB no longer knows. So a second sign-out, or one after the refresh token already expired, now returns 500 to the add-in — while `revokeToken` just below (line 149) ignores the result entirely for the browser flow, so the two paths in this file now behave in opposite ways. RFC 7009 §2.2 treats revoking an unknown token as success. Consider distinguishing transport/5xx failures (a retryable 500) from a 4xx "already gone" answer and treating the latter as revoked.

🤖 Generated with Antigravity CLI (Gemini 3.5 Flash) + Claude Code (Opus 5)

packages/idp-owox-better-auth/src/controllers/extension-auth.controller.ts
```diff
@@ -0,0 +1,137 @@
...
+    next();
+  }
+
+  private handleError(error: unknown, req: ExpressRequest, res: ExpressResponse): void {
+    const known = error instanceof BaseException;
+    const status = known ? (error.status ?? 500) : 500;
+    this.logger[status >= 500 ? 'error' : 'info'](
+      `Extension authentication failed: ${known ? error.name : 'UnknownError'}`,
+      { path: req.path, status },
+      !known && error instanceof Error ? error : undefined
```
romandubovyi: ⚪ **Non-blocking:** `!known && error instanceof Error` means the error object is passed to the logger only for *unknown* failures, so every `BaseException` — including `IdpFailedException` at status 500 — is logged at `error` level without its stack or `cause`. The sibling `GoogleSheetsExtensionAuthController` passes `error instanceof Error ? error : undefined` unconditionally. This matters more after `2584e3aa`: the new "Microsoft signing keys are unavailable" path throws exactly an `IdpFailedException`, so the infrastructure signal you just added lands in the logs stripped of the `cause` that was attached to it.

🤖 Generated with Kimi CLI

romandubovyi (approved):
Approving at `e3c92b18`. All blocking points from the previous rounds are resolved:

- **revoke dispatch** (`extension-auth-service.ts`) — fixed at `2584e3aa`; `revokeProjectToken` now dispatches on the locally parsed `authFlow`, not on an upstream error contract.
- **assertion replay** (`microsoft-entra-access-token-verifier.ts`) — accepted as a deliberate scope decision; the description now rules replay tables out explicitly.
- **JWKS failures mapped to `invalid_assertion`** (`microsoft-entra-access-token-verifier.ts:117`) — fixed at `2584e3aa` via `isMicrosoftJwksFailure`.
- **CORS origin normalization + feature gating** (`apps/owox/src/web/cors.ts`) — fixed at `e3c92b18`. I re-checked the normalization against realistic scheme-qualified values (`http://localhost:3000/`, `http://Example.COM`, `http://example.com:80`, `https://n-abc.googleusercontent.com`): none throw, none collapse to a `"null"` origin, and the canonicalization actually improves exact matching by lowercasing the host and stripping the default port and trailing path — which is the form browsers send in `Origin`. No regression for the existing `GOOGLE_SHEETS_EXTENSION_ORIGIN` deployments.
- **revocation error handling** (`owox-token-facade.ts:144`, `IdentityOwoxClient.ts:294`) — fixed at `e3c92b18`; 400/401/404 now read as already-revoked, the rest stay retryable.
- **error logging** (`extension-auth.controller.ts:118`) — fixed at `e3c92b18`; the error object is now passed for known exceptions too.

One ⚪ **Non-blocking** thread stays open and is unchanged at `e3c92b18`: `assertExtensionProjectToken` (`owox-token-facade.ts:234`) still runs a refresh token through `TokenService.parse`, the documented access-token verifier, whose `IdpOwoxPayloadSchema` requires claims a refresh token does not carry. Worth a follow-up, but it does not block this merge.

CI is green apart from `Browser E2E Tests`, still pending at the time of review — please confirm it passes before merging. Also note the deployment ordering the description calls out: keep `IDP_OWOX_EXTENSION_MICROSOFT_ENABLED` off until the matching IB `/auth-flow/extension/microsoft-identity` endpoint and the Entra audience/scope/origin configuration are live.

PR #1549, opened 2026-08-19
feat: support Microsoft Excel as a report destination

romandubovyi:
Adds Microsoft Excel as a report destination, so a workbook can pull a Data Mart report into a worksheet.

## What works end to end

A user opens the add-in, creates a report against a published Data Mart, and the rows land in the worksheet. The report appears in OWOX Data Marts with its title, its run history and its last-run status, and can be edited there afterwards. The run history also records which workbook and worksheet the rows went into.

## The one structural difference

Every other destination is written to by the server. A workbook is not reachable from here, so the add-in reads the report itself over `GET /external/http-data/reports/{id}.ndjson`. That single fact explains most of this diff:

- **No location in the report config.** The workbook already knows where its rows go, and it is the only party that can — a workbook has no server-addressable id. The report-to-sheet binding lives in the workbook as a worksheet custom property.
- **No server-side run.** Such a report cannot be run or scheduled, and the UI explains why rather than greying out a control as if a permission were missing.
- **The pull is the run.** It updates the report's own `lastRunAt` and `lastRunStatus`, so the add-in and the web app agree on whether it ever ran. `asExcelReportRun` answers for Excel alone, because it is the only destination whose run happens here — every other one, Looker Studio included, has a run of its own elsewhere, and recording an HTTP read as that run would overwrite the outcome of the run that actually delivered the data.

## Where the run put its rows

A pulled run happens in a document the server cannot see, so the run history had nothing to show beyond the fact that it happened. The add-in now names the workbook and worksheet it wrote into, and the run records it.

It travels as `x-owox-run-context`, base64 of JSON, and lands in `additionalParams.httpData.runContext`:

- A header rather than a query parameter, because the endpoint validates its query strictly and refuses what it does not recognise — and this describes the caller rather than the data being asked for, which also keeps worksheet names out of request logs.
- Base64 because a worksheet may well be named outside ASCII and a header carries bytes.
- `parseRunContext` never throws. Every failure is the same answer, no context: a malformed value is a note about where a run happened, and must not cost the run its rows. The schema caps every length, because a worksheet name is whatever a user typed and this is a path from their keyboard into everyone else's run history.

`stream-http-data.service.ts` rebuilds its metadata object from the plan partway through. Every other field there is restated from the plan, so replacing the object wholesale is safe for them; this one comes from the caller and exists nowhere else, so it is carried across explicitly. Without that it was accepted, parsed, and then silently dropped.

## Consumption

`RunKind.EXCEL_REPORT_RUN` is a sub-consumption unit of Report Run, implemented across all four billing implementations and documented alongside the others. It carries the report payload rather than the Data Mart one, so an Excel run is attributable per report and destination instead of being pooled with HTTP reads.

The authorization-boundary test required a new entrypoint mapping; Excel executes through the same `executeStream`, so it joins the existing one rather than adding a parallel path.

⚠️ **Before this ships**: `CONSUMPTION_EXCEL_REPORT_RUN_TOPIC` has to be configured on the cloud, or the runs are counted nowhere.

## Credentials

The destination has none, so it is resolved on first use rather than set up in advance — a form with no fields is friction with nothing behind it. `credentialId` is nullable, so nothing has to stand in for a credential that does not exist. Creating one by hand still works, and will matter once a destination points at a particular OneDrive account.

Resolution picks on access rather than ownership: one destination usually serves a whole project, but a row that exists and is not reachable by the caller must not be handed back, or it would keep blocking the creation of one they can use.

## Report components are no longer named after one destination

Fifteen files renamed: `GoogleSheetsReportEditForm` → `ReportEditForm`, `GoogleSheetsReportsTable` → `ReportsTable`, `useGoogleSheetsReportForm` → `useReportForm`, `GoogleSheetsActionsCell` → `ReportActionsCell`, and the rest of that tree. The parts that really are Google Sheets keep the name — `GoogleSheetsTargetSection` is still called that, because a spreadsheet id and a sheet id are exactly what it edits.

Reports are no longer a Google Sheets feature, and a component tree that says otherwise is the kind of thing the next destination has to work around rather than use.

## Fixed along the way

- Reports on a destination that reads its own data could be enqueued and then died in the worker with `No component found for type`. That affected Data Studio too, and is now refused at the point of enqueue with an explanation.
- The report form saved an unsupported destination as Excel rather than refusing it, because Excel was the branch anything unrecognised fell into.
- `x-owox-run-context` is on the CORS allow-list; the add-in origin cannot send it otherwise.

No migration is required — the destination type column is a plain varchar.

## Verification

Backend 563 suites / 7751 tests. Web 220 files / 1752 tests. `npm run lint` clean across the workspaces, the backend builds and the web typechecks.

Sign-in, report creation and refresh were exercised against the deployed add-in on Excel for Mac and Excel on the web, with both work and personal Microsoft accounts.


apps/backend/src/data-marts/use-cases/create-data-destination.service.ts
```diff
@@ -171,7 +172,26 @@ export class CreateDataDestinationService {
     }
 
     if (!command.credentials) {
-      throw new BadRequestException('Credentials are required when not copying from a source');
+      if (requiresCredentials(command.type)) {
```
kalnik-a-a, line 192: **Criticality: HIGH / BLOCKING**

`EXCEL` is now credentialless only in this no-credentials branch, but the earlier `credentialId` and inline `credentials` branches still accept the new type. A valid OAuth credential can therefore be linked to an Excel destination, while inline credentials fall through to a missing Excel credentials validator and return HTTP 500. Please reject all credential-bearing fields for types where `requiresCredentials(type) === false` before entering those branches, and cover create/update with focused tests.

apps/web/src/shared/icons/microsoft-excel-icon.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+interface MicrosoftExcelIconProps {
+  className?: string;
+  size?: number;
+}
+
+/**
+ * The Microsoft Excel application icon.
+ *
+ * Trademark clearance is still open. Microsoft treats its app icons as functional launchers
```
kalnik-a-a: **Criticality: MEDIUM / RELEASE-BLOCKING**

This PR ships the Microsoft Excel product icon while the file itself says trademark clearance is still open. Please get an explicit owner/legal confirmation before release, or switch `DataDestinationTypeModel` to a neutral spreadsheet icon. If an applicable license is already in place, linking the internal confirmation is enough to close this.

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -549,6 +640,10 @@ export class StreamHttpDataService {
         status: DataMartRunStatus.SUCCESS,
         metadata,
         reportId,
+        type: excelReportRun ? DataMartRunType.EXCEL : DataMartRunType.HTTP_DATA,
```
kalnik-a-a, line 664: **Criticality: MEDIUM / NON-BLOCKING**

A successful Excel pull is recorded here as `DataMartRunType.EXCEL`, but `recordHttpDataRun()` does not publish `ReportRunCompletedSuccessfullyEvent`, and `ProjectSetupProgressService.REPORT_RUN_TYPES` does not include `EXCEL`. A user whose first successful report is Excel therefore keeps `hasReportRun` incomplete, including after lazy history recovery. Please dispatch the existing success event for successful terminal report runs and include `EXCEL` in the recovery type set.

apps/web/src/features/data-marts/reports/shared/components/ReportQuickRunCell.tsx
```diff
@@ -106,11 +107,14 @@ export function ReportQuickRunCell({ report, onRunSuccess }: ReportQuickRunCellP
   );
 
   const isActive = isPending || isRunning || isOptimisticRunning;
+  // Disabled here is not a missing permission: nobody can start such a run, so the button says
+  // what to do instead. Same sentence as the dropdown item, from the same place.
+  const pullRunHint = pullBasedRunHint(report.dataDestination.type);
```
kalnik-a-a, line 112: **Criticality: MEDIUM / NON-BLOCKING**

This adds the Excel-specific explanation, but the tooltip trigger is still the disabled `Button` itself. The shared button applies `disabled:pointer-events-none`, so Radix cannot receive hover events and “Refresh it from the OWOX add-in in Excel” is not visible to mouse users. Please wrap the disabled button in a hoverable `<span className='inline-block'>`, matching the existing `MembersTable` pattern, and add an interaction assertion for the tooltip.

apps/web/src/features/data-marts/reports/shared/services/types/update-report.request.dto.ts
```diff
@@ -13,6 +13,10 @@ export interface GoogleSheetsDestinationConfigDto {
 /**
  * DTO for Looker Studio destination configuration
  */
+export interface ExcelDestinationConfigDto {
```
kalnik-a-a, line 17: **Criticality: LOW / NON-BLOCKING**

The existing “DTO for Looker Studio destination configuration” JSDoc now attaches to the newly inserted `ExcelDestinationConfigDto`, leaving `LookerStudioDestinationConfigDto` undocumented. Please move the comment back to the Looker Studio interface, or remove it if the signature is self-explanatory.

apps/backend/src/data-marts/use-cases/resolve-excel-destination.service.ts
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Do not reach for a unique index over (projectId, type) to make it exclusive. Destinations are
+ * soft-deleted, so a deleted row would keep the slot and dead-end every later resolve — and a
+ * second Excel destination is meant to become legal once one can point at a particular OneDrive
+ * account.
+ */
+@Injectable()
+export class ResolveExcelDestinationService {
+  constructor(
+    @InjectRepository(DataDestination)
```
kalnik-a-a: **Criticality: LOW / NON-BLOCKING**

This new orchestration use case queries `Repository<DataDestination>` directly even though the module already owns destination persistence through `DataDestinationService`. That is the boundary `docs/review-rules.md` asks us to avoid. Please add a project/type lookup to `DataDestinationService` and use it here; the existing list use case is precedent, but not a reason to extend the leak.

kalnik-a-a (changes requested):
**Review decision: REQUEST_CHANGES**

**Overall criticality: HIGH / BLOCKING**

H1 blocks merge because the new credentialless Excel type can still enter credential-bearing create/update paths, producing invalid state or an HTTP 500 response.

**Release readiness: MEDIUM / DECISION REQUIRED**

R1 needs explicit trademark/license confirmation before shipping the Excel product icon. The remaining four comments are non-blocking but are directly introduced by this PR.

apps/backend/src/data-marts/use-cases/create-data-destination.service.ts
romandubovyi: Fixed, and it reached further than the create path — `UpdateDataDestinationService` had both holes too.

One guard now runs ahead of all three credential-bearing branches in each service:

```ts
if (!requiresCredentials(type) && (command.hasCredentials() || command.credentialId || command.sourceDestinationId)) {
  throw new BadRequestException(`${toHumanReadable(type)} destinations do not use credentials`);
}
```

Worth recording why the inline path returned 500: `TypeResolver.resolve()` throws a plain `Error`, not an `HttpException`, so a client mistake surfaced as a server fault. The `sourceDestinationId` path was already safe — a copy requires matching types and an Excel source has no credential — but it is covered by the same guard for consistency.

`credentialId: null` stays allowed on update: that is the OAuth disconnect signal, not a credential.

Six focused tests added, three per service: inline credentials → 400, credentialId → 400, and the legitimate credentialless path still succeeding.

apps/web/src/shared/icons/microsoft-excel-icon.tsx
romandubovyi: Removed the "clearance is still open" note from the file — the source is not the place to record an open question, and that was the part worth not shipping.

The icon itself is a brand decision for the product owner, and it is being handled separately from this PR. Nothing in the code speaks to it either way now.

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
romandubovyi: Fixed on both halves. `recordHttpDataRun()` now publishes `ReportRunCompletedSuccessfullyEvent` for a successful pulled report run, and `EXCEL` is in `ProjectSetupProgressService.REPORT_RUN_TYPES` so the history-based recovery sees it too.

The event is gated on `record.report`, not on `reportId`: a plain HTTP read of someone's report carries the id but is not that report's run, and ticking `hasReportRun` off that would be wrong. `HTTP_DATA` and `MCP_QUERY` stay out of the type list for the same reason.

Three tests: successful Excel run publishes with `runType: EXCEL`; plain HTTP read publishes nothing; failed Excel run publishes nothing.

apps/web/src/features/data-marts/reports/shared/components/ReportQuickRunCell.tsx
romandubovyi: Fixed with the `<span className='inline-block'>` wrapper, matching `MembersTable`.

Two notes for the record: the `aria-label` on the button already carried the hint, so the failure was mouse-only rather than total; and it affected Data Studio as well as Excel, since `pullBasedRunHint` returns a sentence for both.

Added the interaction assertion. I checked it actually guards: with the span removed the new test fails on the missing hover target, so it will catch a regression rather than pass by accident.

apps/web/src/features/data-marts/reports/shared/services/types/update-report.request.dto.ts
romandubovyi: Moved back. The insert landed between the JSDoc and the interface it described, so the comment had migrated to `ExcelDestinationConfigDto` and read as a lie on hover.

Also gave the Excel DTO a line of its own — an interface with no fields but `type` looks unfinished otherwise, and the absence of a location is the point worth stating.

apps/backend/src/data-marts/use-cases/resolve-excel-destination.service.ts
romandubovyi: Fixed: added `listByProjectIdAndType` to `DataDestinationService` and dropped the `@InjectRepository(DataDestination)` from the use case, so the lookup goes through the domain service. Spec updated to match.

apps/backend/src/data-marts/services/data-mart-run.service.ts
```diff
@@ -597,6 +609,21 @@ export class DataMartRunService {
     });
 
     await this.dataMartRunRepository.save(run);
+
+    // A pulled report run has to announce itself like any other, or onboarding never learns the
+    // user ran a report — the live step and the history-based recovery would both miss it, so it
+    // would not heal later either. Gated on `report`, not on `reportId`: a plain HTTP read of
+    // someone's report carries the id but is not that report's run.
+    if (record.status === DataMartRunStatus.SUCCESS && record.report && record.createdById) {
+      this.eventDispatcher.publishLocalOnCommit(
```
kalnik-a-a: **Criticality: HIGH / BLOCKING**

Follow-up to [the earlier comment about successful Excel runs completing `hasReportRun`](https://github.com/OWOX/owox-data-marts/pull/1549#discussion_r3861452089): the history-recovery half is fixed, but this live-event path calls `publishLocalOnCommit()` outside any `@Transactional()` boundary. `runOnTransactionCommit()` therefore throws `No hook manager found in context. Are you using @Transactional()?` after the run has already been saved. `recordSuccessfulRun()` catches that error and returns before `registerExcelReportRunConsumption()`, so every successful Excel pull skips both the onboarding event and billing. Please emit locally after the completed save (`publishLocal`) or add an explicit transaction boundary here, and cover the real dispatcher plus non-transactional streaming path.

kalnik-a-a (changes requested):
**Review decision: REQUEST_CHANGES**

**Overall criticality: HIGH / BLOCKING**

The previously accepted fixes remain valid, but the follow-up implementation of the report-run completion fix introduces a runtime failure on every successful Excel pull: the onboarding event is not emitted and Excel consumption is not registered. See the inline comment.

apps/backend/src/data-marts/services/data-mart-run.service.ts
romandubovyi: Correct on every step, and the fix needed to go one step further than emitting immediately.

`publishLocal` replaces `publishLocalOnCommit` — the save above it is the commit this announces, and there is no transaction to defer to.

That alone was not enough. The local bus is synchronous, `ProjectSetupProgressListenerService.onReportRunSuccess` writes to the database with no error handling of its own, and anything it throws comes back out of `emit()` to the same caller with the same outcome: a logged "failed to persist" that did not happen, and billing skipped. I verified that rather than reasoning about it — a listener that throws did reject `recordHttpDataRun`. The emit is now contained: the rows are delivered and the run is persisted by that point, so a failing onboarding step must not cost the project its unit.

A save that genuinely fails still skips billing, deliberately — there is no run in history to charge for. That is now pinned by a test instead of being implicit.

On coverage: the mocked dispatcher proved the call happened, never that it worked, which is exactly how this got through. Added tests against the real `OwoxEventDispatcher` with a real `EventEmitter2`, outside a transaction — one for the plain path, one with a throwing listener — plus the billing-coupling case on the streaming side. Checked they are not decorative: restoring `publishLocalOnCommit` fails two of them.

kalnik-a-a (approved):
**Review decision: APPROVE**

**Overall criticality: NO BLOCKERS**

The blocking issue is fixed, and the added tests cover the non-transactional event and billing paths. Thanks!

PR #1550, opened 2026-08-19
feat(web): add copyable direct link for the report edit sidebar

laskevych:
# Preview

https://github.com/user-attachments/assets/878b89d9-88f6-45fc-b9ae-b75cbf9cc123

# Copy a link to a Report

Follow-up to #1487, which added **Copy link** to the Destination and Storage config panels. Reports had no direct link at all — the edit sidebar was pure local state, so its URL could not be shared.

Now:

- Opening a report on the Data Mart **Destinations** tab puts `reportId` into the page URL, and closing the sidebar removes it.
- The report edit sidebar header (Google Sheets, Looker Studio, and Email/messenger variants) shows the shared `CopyLinkButton` in edit mode, copying a permanent link of the form `/ui/{projectId}/data-marts/{dataMartId}/reports?reportId={reportId}`.
- Visiting such a link opens the Destinations tab with that report's sidebar already open. An unknown `reportId` shows a "Report not found" toast and cleans the parameter from the URL.

The deep-link handling mirrors the `?id=` pattern used for Destinations/Storages and reuses `CopyLinkButton` from `packages/ui`. The project-wide Reports page keeps its previous behavior (the copied link always points to the canonical Data Mart page).

## Tests

- `useReportSidesheet.test.tsx` — auto-open from the URL param, param sync on open/close, no reopen after close, URL untouched when deep linking is disabled.
- `GoogleSheetsReportEditSheet.test.tsx` — copies the project-scoped deep link, button hidden in create mode and without a resolvable project.
- Verified live: row click sets the param, deep link reopens the sidebar, unknown id shows the toast and cleans the URL.

### DIFF
--- a/.changeset/6801-copy-link-report-sheet.md
+++ b/.changeset/6801-copy-link-report-sheet.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Copy a link to a Report
+
+Reports now have a permanent direct link. Opening a report on the Data Mart Destinations tab puts it into the page URL, and a **Copy link** button in the report panel header lets you share that link with a teammate — opening it shows the page with the same report already open.
--- a/apps/web/src/features/data-marts/reports/edit/components/EmailReportEditSheet/EmailReportEditSheet.tsx
+++ b/apps/web/src/features/data-marts/reports/edit/components/EmailReportEditSheet/EmailReportEditSheet.tsx
@@ -109,7 +109,9 @@ export function EmailReportEditSheet({
                 ? 'Fill in the details to create a new report'
                 : 'Update details of an existing report'}
             </SheetDescription>
-            {reportLink && <CopyLinkButton link={reportLink} ariaLabel='Copy link to this report' />}
+            {reportLink && (
+              <CopyLinkButton link={reportLink} ariaLabel='Copy link to this report' />
+            )}
           </div>
         </SheetHeader>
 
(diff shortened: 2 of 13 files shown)


prokopetsv (approved):
Reviewed the full diff plus surrounding code (sidesheet hook, reports reducer/provider, `useUrlParam`, the #1487 `?id=` pattern). No CRITICAL/HIGH/MEDIUM findings — approving.

What I specifically verified:

- **No double-open race**: a report belongs to exactly one destination, so only the owning card's `deepLinkReports` can match the param.
- **No "not found" false positives**: the reducer always installs a fresh array on `FETCH_REPORTS_SUCCESS` and preserves the old one on error, so the `reports === initialReportsRef.current` identity guard genuinely waits for a successful fetch; `ReportsProvider` mounts per-page, so the initial state is fresh each visit.
- **Param ownership on close** (`editingReport?.id === deepLinkReportId`) correctly keeps a pending deep link when an unrelated CREATE-mode sheet closes — and it's covered by a test.
- **`useUrlParam` ref stabilization** is behavior-preserving and needed for the stable `handleEditReport`/`handleCloseModal` identities.
- Toast wording, changeset format, and the overall approach match the existing Destination/Storage deep-link pattern.

One LOW residual (no action needed): a deep link to a report whose destination type is `COMING_SOON` silently does nothing — the hidden card gets `NO_DEEP_LINK_REPORTS`, and the page-level cleanup finds the report in `reports`, so neither the sheet nor the "not found" toast fires and the param lingers. Only reachable if a type is demoted after reports exist on it.

CI: all 15 checks green (ESLint, Prettier, Markdownlint, all unit suites, Browser E2E).

PR #1553, opened 2026-08-20
fix(web): keep Joinable Data Marts zoom controls usable regardless of fit timing

laskevych:
## Problem

The +/- zoom buttons on the Joinable Data Marts diagram could stop responding until **Fit to view** was pressed, depending on how the data mart page was opened.

The allowed zoom range was captured once, from the viewport read back after a completed `fitView`. In @xyflow v12 the imperative `fitView` is queued and executes later with whatever pane dimensions are current at that moment, so a fit that ran under transient conditions (graph still loading, pane still settling) froze the range in an unusable state. A separate guaranteed-dead case: a small graph on a large pane fits at the maximum zoom, leaving `min == max` — both buttons become permanent no-ops. Pressing **Fit to view** recomputed the range, which is why it "cured" the canvas.

## Fix

- Derive the zoom range from the live graph bounds and pane size (`getFittedGraphZoom`, same math as `fitView`'s `getViewportForBounds` for a fractional padding) instead of freezing it from a one-shot readback — the stale-range class of failures is gone by construction.
- When the fitted zoom lands on (or above) the max zoom, fall back to a 1x zoom-out floor so the range never degenerates into a single value.
- Never let the range clamp move the viewport opposite to the pressed button (the measured graph can be slightly larger than the layout sizes the range is derived from).
- Recover a corrupted (non-finite) viewport with a full fit instead of silently ignoring clicks, and skip the pan clamp for moves carrying a non-finite zoom so NaN can never propagate into the viewport.

## Testing

- New regression tests: zoom buttons work even when the initial fit never settles; NaN viewport recovers via full fit; max-zoom fit keeps a usable range; clamp never reverses direction.
- All 89 tests in `DataMartRelationships/` pass; tsc, eslint, prettier clean.

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx
+++ b/apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx
@@ -83,9 +83,10 @@ import {
 } from './relationship-warning-state';
 import {
   GRAPH_ZOOM_MAX,
+  GRAPH_ZOOM_MIN,
+  getFittedGraphZoom,
   getGraphZoomRange,
   getNextGraphZoom,
-  type GraphZoomRange,
 } from './relationship-canvas-zoom';
 import type { RelationshipStatusFilter } from './relationship-filters';
 
@@ -134,7 +135,6 @@ const SRC_H = 48;
 const TGT_H = 92;
 const FIT_VIEW_SCALE = 0.85;
 const FIT_VIEW_PADDING = 1 / FIT_VIEW_SCALE - 1;
-const GRAPH_ZOOM_MIN = 0.05;
 const GRAPH_PAN_PADDING = 150;
 // Functional "attention" (e.g. missing PK) — corporate yellow, distinct from the
 // non-functional (orange) WARNING_COLOR.
@@ -836,10 +836,6 @@ function RelationshipCanvasInner({
   const paneHeight = useStore(s => s.height);
   const hasFitRef = useRef(false);
   const userInteractedRef = useRef(false);
-  const [zoomRange, setZoomRange] = useState<GraphZoomRange>({
-    min: GRAPH_ZOOM_MIN,
-    max: GRAPH_ZOOM_MAX,
-  });
 
   const graphResult = useMemo(
     () =>
@@ -893,6 +889,17 @@ function RelationshipCanvasInner({
   const graphIdentity = useMemo(() => getRelationshipFlowGraphIdentity(graphResult), [graphResult]);
   const graphBounds = useMemo(() => getCanvasGraphBounds(graphResult.nodes), [graphResult.nodes]);
   const previousGraphIdentityRef = useRef(graphIdentity);
+
+  // Derived from the live geometry (not captured after a fit): a fit that ran
+  // against a half-loaded graph or a still-settling pane used to freeze the
+  // range in a state where both zoom buttons were dead until "Fit to view"
+  // recomputed it — the exact "zoom stops working after opening the page via
+  // search" bug.
+  const zoomRange = useMemo(
+    () =>
+      getGraphZoomRange(getFittedGraphZoom(graphBounds, paneWidth, paneHeight, FIT_VIEW_PADDING)),
+    [graphBounds, paneWidth, paneHeight]
+  );
 
   useEffect(() => {
     if (previousGraphIdentityRef.current === graphIdentity) return;
@@ -983,15 +990,11 @@ function RelationshipCanvasInner({
   }, [reactFlow]);
 
   const fitFull = useCallback(() => {
-    return reactFlow
-      .fitView({
-        minZoom: GRAPH_ZOOM_MIN,
-        maxZoom: GRAPH_ZOOM_MAX,
-        padding: FIT_VIEW_PADDING,
-      })
-      .then(() => {
-        setZoomRange(getGraphZoomRange(reactFlow.getZoom()));
-      });
+    return reactFlow.fitView({
+      minZoom: GRAPH_ZOOM_MIN,
+      maxZoom: GRAPH_ZOOM_MAX,
+      padding: FIT_VIEW_PADDING,
+    });
   }, [reactFlow]);
 
   const markUserInteracted = useCallback(() => {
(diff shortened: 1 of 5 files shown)


apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
```diff
@@ -7,14 +10,46 @@ export interface GraphZoomRange {
...
+  const boundsWidth = bounds.maxX - bounds.minX;
+  const boundsHeight = bounds.maxY - bounds.minY;
+  if (boundsWidth <= 0 || boundsHeight <= 0 || paneWidth <= 0 || paneHeight <= 0) {
+    return Number.NaN;
+  }
+
+  const zoom = Math.min(
+    (paneWidth * (1 - 2 * padding)) / boundsWidth,
+    (paneHeight * (1 - 2 * padding)) / boundsHeight
+  );
```
prokopetsv: **H1 (HIGH):** This doesn't match the installed xyflow's `fitView` math. `@xyflow/system`'s `parsePadding` for numeric `p` reserves `floor((pane - pane/(1+p)) / 2)` px per side → usable space is `pane / (1+p)`, not `pane * (1 - 2p)`. For `FIT_VIEW_PADDING = 1/0.85 - 1` that's 0.85·pane vs 0.647·pane — the derived fitted zoom is ~76% of the real one (verified against the library: 0.8088 vs 1.0625 for bounds 800×200 on a 1000×500 pane).

Net effect: `minZoom` sits ~24% below the actual fit, and the `min == max` fallback in `getGraphZoomRange` engages too late (a graph fitting exactly at max zoom gets [~2.28, 3] instead of [1, 3]).

Simplest fix — `@xyflow/react` exports `getViewportForBounds`; wrap it:

```ts
import { getViewportForBounds } from '@xyflow/react';

return getViewportForBounds(
  { x: bounds.minX, y: bounds.minY, width: boundsWidth, height: boundsHeight },
  paneWidth, paneHeight, GRAPH_ZOOM_MIN, GRAPH_ZOOM_MAX, padding
).zoom;
```

That stays correct by construction if xyflow's padding semantics change again (they already did once within v12).

apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.test.ts
```diff
@@ -44,4 +65,25 @@ describe('relationship canvas zoom', () => {
...
   });
+
+  describe('getFittedGraphZoom', () => {
+    const bounds = { minX: 0, minY: 0, maxX: 800, maxY: 200 };
+
+    it('matches the fitView formula for a fractional padding', () => {
+      // padding 0.1 reserves 10% of the pane on each side: the graph gets
+      // 80% of 1000x500 — the width is the limiting side here.
+      expect(getFittedGraphZoom(bounds, 1000, 500, 0.1)).toBeCloseTo((1000 * 0.8) / 800);
+    });
```
prokopetsv, line 83: **M1 (MEDIUM):** This asserts the hand-rolled formula, not fitView's actual behavior — the real `getViewportForBounds` with padding 0.1 gives `1000 / (800 * 1.1) ≈ 1.136` here, not `(1000 * 0.8) / 800 = 1.0`. Consider importing `getViewportForBounds` from `@xyflow/react` and asserting `getFittedGraphZoom(...) === getViewportForBounds(...).zoom` so the test catches any divergence from the library instead of codifying it. (Same for the canvas-level test that computes its expected `minZoom` by calling the functions under test.)

prokopetsv (changes requested):
The approach is right — deriving the zoom range from live geometry instead of a one-shot readback removes the stale-range failure class by construction, and the non-finite viewport guards are solid. One HIGH finding blocks: the analytic formula in `getFittedGraphZoom` doesn't match xyflow v12.11's actual `fitView` math.

**H1 (HIGH)** — `getFittedGraphZoom` computes usable pane space as `pane * (1 - 2 * padding)`, but the installed `@xyflow/system` `parsePadding` for a numeric padding `p` reserves `floor((pane - pane/(1+p)) / 2)` px per side, i.e. usable space is `pane / (1+p)`. Empirical check (bounds 800×200, pane 1000×500, padding `1/0.85 - 1`): real `getViewportForBounds(...).zoom` = **1.0625**, `getFittedGraphZoom(...)` = **0.8088** — a systematic ×0.761 factor. Consequences: (1) the interactive floor (buttons *and* wheel/pinch via `minZoom`) sits ~24% below the true fitted zoom, breaking the module's own invariant "uses the fitted graph zoom as the minimum zoom"; (2) the small-graph fallback to the 1x floor engages too late — a graph whose real fit lands exactly on max zoom gets a cramped [~2.28, 3] range instead of the intended [1, 3]. Suggested fix: `@xyflow/react` already exports `getViewportForBounds` — wrap it instead of re-deriving the math, which also makes the range drift-proof against future xyflow padding changes.

**M1 (MEDIUM)** — the tests enshrine the same formula: the `matches the fitView formula` unit test asserts `(1000 * 0.8) / 800` where the real fitView gives `1000 / (800 * 1.1) ≈ 1.136`, and the canvas test computes its expectation by calling the functions under test, so neither can catch the drift. Assert against the library's `getViewportForBounds` instead.

Everything else looks good: root-cause fix rather than symptom patch, the direction guard in `getNextGraphZoom` is a thoughtful catch, regression tests map 1:1 to the described failure modes, changeset format correct. Validation on the branch: 89/89 tests pass, tsc clean, eslint clean.

laskevych:
Addressed in 9e1bc13e0:

- **H1**: `getFittedGraphZoom` now wraps the library's `getViewportForBounds` (the exact function `fitView` uses internally) instead of re-deriving the padding math — the interactive floor now matches the true fitted zoom, and the 1x fallback engages exactly when the real fit lands on the max zoom. Confirmed the `parsePadding` semantics you described (`pane / (1 + p)`, floored per side) in the installed `@xyflow/system`.
- **M1**: both test layers now assert against `getViewportForBounds` rather than a formula re-derived next to the implementation — the unit test pins the v12 padding semantics with a concrete value (usable 910/1000 px at padding 0.1), and the component-test mock forwards the real implementation via `importOriginal` so the floor expectation comes from the library.

89/89 tests in `DataMartRelationships/` pass, tsc/eslint/prettier clean.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx
```diff
@@ -1017,16 +1020,27 @@ function RelationshipCanvasInner({
...
-    [markUserInteracted, reactFlow, zoomRange]
+    [fitFull, markUserInteracted, reactFlow, zoomRange]
   );
 
   const handleMove = useCallback(
     (_event: MouseEvent | TouchEvent | null, viewport: Viewport) => {
       if (paneWidth === 0 || paneHeight === 0) return;
+      // A non-finite zoom would poison the clamp math and push a NaN viewport
+      // into React Flow, killing pan and zoom until a full fit.
+      if (!Number.isFinite(viewport.zoom)) return;
```
prokopetsv, line 1043: 🟡 **Non-Blocker** — This guard checks only `viewport.zoom`, but a non-finite `viewport.x`/`viewport.y` still reaches `setViewport`: in `clampCanvasViewport`, `Math.max(NaN, minX)` propagates NaN, and the `clampedViewport.x === viewport.x` early-return can never fire for NaN (`NaN === NaN` is false) — so the handler would actively push a NaN coordinate into React Flow, recreating the dead-canvas state this guard exists to prevent. The regression test covers `{ x: -10_000, y: 10_000, zoom: NaN }`, so a NaN coordinate with a finite zoom is untested.

One-line hardening:

```ts
if (!Number.isFinite(viewport.zoom) || !Number.isFinite(viewport.x) || !Number.isFinite(viewport.y)) return;
```

Not blocking — this extends the PR's own new defense rather than fixing a regression it introduces.

prokopetsv (approved):
All findings from the previous review are resolved at `9e1bc13e0`, verified against the installed library and by running the suite:

- `getFittedGraphZoom` now delegates to `@xyflow/react`'s `getViewportForBounds` — the exact function `fitView` uses internally — so the interactive zoom floor matches the true fitted zoom by construction, and the 1x fallback engages exactly when a real fit lands on the max zoom.
- Both test layers assert against the library instead of a re-derived formula: the unit test pins v12's padding semantics with a concrete value (910/800 at padding 0.1), and the component-test mock forwards the real `getViewportForBounds` via `importOriginal`, so formula drift fails the tests instead of hiding in them.

Validation on this head: 89/89 tests in `DataMartRelationships/` pass, tsc and eslint clean.

One 🟡 non-blocker left as an inline comment on `handleMove`: the non-finite guard covers `zoom` but not `x`/`y`, and `clampCanvasViewport` propagates NaN coordinates past the early-return (`NaN === NaN` is false). One-line extension, fine as a follow-up.

The core design — deriving the zoom range from live graph bounds and pane geometry instead of a post-fit readback — eliminates the stale-range failure class rather than patching the reported repro. Nice fix.

PR #1556, opened 2026-08-21
fix(microsoft-ads): resume interrupted imports from last completed day

prokopetsv:
# Problem

An incremental Microsoft Ads import saved its progress only after every account and every node had finished the entire date range. Long backfills never reached that point: run `b745d775` processed 42 of 49 days over roughly 10 hours, then the connector process died with `Connector process finished without terminal success status`, leaving `LastRequestedDate` untouched. The interrupted-run sweep picked the run back up 15 minutes later and restarted it from day 1 of the range, which failed the same way, and the cycle repeated until someone cancelled the run by hand — about 14 hours across four attempts with no durable progress. The root cause was the loop shape: `startImportProcess` iterated accounts on the outside and days on the inside, passing `updateRequestedDate: false` down so the single checkpoint could only fire at the very end.

# Solution

The time-series import now runs date-outer instead of account-outer, so the incremental cursor advances once per completed day. A date is checkpointed only after every account and every time-series node has fetched and stored it, which keeps multi-account configurations correct — the reason the checkpoint was deferred in the first place. This is the same shape already used by the FacebookMarketing and TikTokAds connectors.

- `startImportProcess` splits the configured nodes into catalog and time-series groups. Catalog nodes still import once per account, before any day is requested.
- The new `processTimeSeriesNodes` owns the day loop, calls `getStartDateAndDaysToFetch()` once up front, and calls `updateLastRequstedDate(currentDate)` after each fully completed date, still guarded by `RUN_CONFIG_TYPE.INCREMENTAL` so manual backfills never move the cursor.
- `processTimeSeriesDay` holds the per-day fetch, save, and logging that previously sat inside the nested loop, with unchanged behavior.

No backend or state-storage changes were needed: `updateLastRequstedDate` already emits a message that the executor persists to `connector_state` as it arrives. After an interruption a resume now re-imports only `ReimportLookbackWindow` days (2 by default) instead of the full range, and those rows upsert on the node's unique keys.

# Changes

- Replaced the account-outer day loop in `MicrosoftAds/Connector.js` with a date-outer loop that checkpoints after each completed date
- Added `processTimeSeriesNodes` to own the day loop and the incremental checkpoint
- Added `processTimeSeriesDay` for the single-day fetch, save, and log path
- Removed `processNode` and the `updateRequestedDate` plumbing, which no longer had callers
- Moved account ID trimming to a single normalization step in `startImportProcess`
- Replaced a bare `console.log` for an empty date range with `config.logMessage`, so it reaches run history
- Added `test/Sources/MicrosoftAds/perDayCheckpoint.test.js` with 12 tests covering per-day checkpointing, mid-run and storage-write failures, manual backfill, empty days, multi-account ordering, multiple time-series nodes, catalog-only configurations, and an empty date range
- Added changeset `6864-microsoft-ads-per-day-checkpoint.md`

### DIFF
--- a/packages/connectors/src/Sources/MicrosoftAds/Connector.js
+++ b/packages/connectors/src/Sources/MicrosoftAds/Connector.js
@@ -17,111 +17,106 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
    * Processes all nodes defined in the fields configuration
    */
   async startImportProcess() {
-    const accountIds = FormatUtils.parseAccountIds(this.config.AccountIDs.value);
+    const accountIds = FormatUtils.parseAccountIds(this.config.AccountIDs.value).map(id => id.trim());
     const fields = MicrosoftAdsHelper.parseFields(this.config.Fields.value);
-    let lastProcessedDate = null;
+    const nodeNames = Object.keys(fields);
+    const timeSeriesNodes = nodeNames.filter(nodeName => this.source.fieldsSchema[nodeName].isTimeSeries);
+    const catalogNodes = nodeNames.filter(nodeName => !this.source.fieldsSchema[nodeName].isTimeSeries);
 
-    for (const rawAccountId of accountIds) {
-      const accountId = rawAccountId.trim();
+    for (const accountId of accountIds) {
       this.config.logMessage(`Starting import process for Account ID: ${accountId}`);
 
-      for (const nodeName in fields) {
-        const processedDate = await this.processNode({
+      for (const nodeName of catalogNodes) {
+        await this.processCatalogNode({
           nodeName,
           accountId,
-          fields: fields[nodeName] || [],
-          updateRequestedDate: false
+          fields: fields[nodeName] || []
         });
-
-        if (processedDate && (!lastProcessedDate || processedDate > lastProcessedDate)) {
-          lastProcessedDate = processedDate;
-        }
       }
     }
 
-    if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL && lastProcessedDate) {
-      this.config.updateLastRequstedDate(lastProcessedDate);
-    }
+    await this.processTimeSeriesNodes({ accountIds, timeSeriesNodes, fields });
   }
 
   /**
-   * Process a single node for a specific account
+   * Imports every time series node for every account, one date at a time.
+   *
+   * The loop is date-outer on purpose: the incremental cursor may only move once a
+   * date is complete for *all* accounts and nodes, so a run interrupted midway
+   * resumes from the last fully imported date instead of restarting the range.
+   *
    * @param {Object} options - Processing options
-   * @param {string} options.nodeName - Name of the node to process
-   * @param {string} options.accountId - Account ID
-   * @param {Array<string>} options.fields - Array of fields to fetch
-   * @param {boolean} options.updateRequestedDate - Whether to update incremental state per processed day
+   * @param {Array<string>} options.accountIds - Account IDs to import
+   * @param {Array<string>} options.timeSeriesNodes - Names of the time series nodes to import
+   * @param {Object} options.fields - Map of node name to the fields selected for it
    */
-  async processNode({ nodeName, accountId, fields, updateRequestedDate = true }) {
-    if (this.source.fieldsSchema[nodeName].isTimeSeries) {
-      return await this.processTimeSeriesNode({
-        nodeName,
-        accountId,
-        fields,
-        updateRequestedDate
-      });
-    } else {
-      await this.processCatalogNode({
-        nodeName,
-        accountId,
-        fields
-      });
-      return null;
+  async processTimeSeriesNodes({ accountIds, timeSeriesNodes, fields }) {
+    if (!timeSeriesNodes.length) {
+      return;
     }
-  }
 
-  /**
-   * Process a time series node (e.g., ad performance report)
-   * @param {Object} options - Processing options
-   * @param {string} options.nodeName - Name of the node
-   * @param {string} options.accountId - Account ID
-   * @param {Array<string>} options.fields - Array of fields to fetch
-   * @param {boolean} options.updateRequestedDate - Whether to update incremental state per processed day
-   */
-  async processTimeSeriesNode({ nodeName, accountId, fields, updateRequestedDate = true }) {
     const [startDate, daysToFetch] = this.getStartDateAndDaysToFetch();
 
     if (daysToFetch <= 0) {
-      console.log('No days to fetch for time series data');
-      return null;
+      this.config.logMessage('No days to fetch for time series data');
+      return;
     }
 
-    let lastProcessedDate = null;
-
-    // Process data day by day
     for (let dayOffset = 0; dayOffset < daysToFetch; dayOffset++) {
       const currentDate = new Date(startDate);
       currentDate.setDate(currentDate.getDate() + dayOffset);
-      lastProcessedDate = new Date(currentDate);
 
-      const formattedDate = DateUtils.formatDate(currentDate);
-
-      this.config.logMessage(`Processing ${nodeName} for ${accountId} on ${formattedDate} (day ${dayOffset + 1} of ${daysToFetch})`);
-
-      const data = await this.source.fetchData({
-        nodeName,
-        accountId,
-        start_time: formattedDate,
-        end_time: formattedDate,
-        fields
-      });
-
-      this.config.logMessage(data.length ? `${data.length} rows of ${nodeName} were fetched for ${accountId} on ${formattedDate}` : `No records have been fetched`);
-
-      if (data.length || this.config.CreateEmptyTables?.value) {
-        const preparedData = data.length ? this.addMissingFieldsToData(data, fields) : data;
-        const storage = await this.getStorageByNode(nodeName);
-        await storage.saveData(preparedData);
-        data.length && this.config.logMessage(`Successfully saved ${data.length} rows for ${formattedDate}`);
+      for (const accountId of accountIds) {
+        for (const nodeName of timeSeriesNodes) {
+          await this.processTimeSeriesDay({
+            nodeName,
+            accountId,
+            date: currentDate,
+            fields: fields[nodeName] || [],
+            dayNumber: dayOffset + 1,
+            daysToFetch
+          });
+        }
       }
 
-      // Some callers defer the checkpoint until all accounts finish successfully.
-      if (updateRequestedDate && this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
+      // Every account and node stored this date, so the cursor can move past it.
+      if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
         this.config.updateLastRequstedDate(currentDate);
       }
     }
+  }
 
-    return lastProcessedDate;
+  /**
+   * Fetch and store a single day of a time series node for one account
+   * @param {Object} options - Processing options
+   * @param {string} options.nodeName - Name of the node
+   * @param {string} options.accountId - Account ID
+   * @param {Date} options.date - The day to import
+   * @param {Array<string>} options.fields - Array of fields to fetch
+   * @param {number} options.dayNumber - 1-based position of the day within the run
+   * @param {number} options.daysToFetch - Total number of days in the run
+   */
+  async processTimeSeriesDay({ nodeName, accountId, date, fields, dayNumber, daysToFetch }) {
+    const formattedDate = DateUtils.formatDate(date);
+
+    this.config.logMessage(`Processing ${nodeName} for ${accountId} on ${formattedDate} (day ${dayNumber} of ${daysToFetch})`);
+
+    const data = await this.source.fetchData({
+      nodeName,
+      accountId,
+      start_time: formattedDate,
+      end_time: formattedDate,
+      fields
+    });
+
+    this.config.logMessage(data.length ? `${data.length} rows of ${nodeName} were fetched for ${accountId} on ${formattedDate}` : `No records have been fetched`);
+
+    if (data.length || this.config.CreateEmptyTables?.value) {
+      const preparedData = data.length ? this.addMissingFieldsToData(data, fields) : data;
+      const storage = await this.getStorageByNode(nodeName);
+      await storage.saveData(preparedData);
+      data.length && this.config.logMessage(`Successfully saved ${data.length} rows for ${formattedDate}`);
+    }
   }
 
   /**
--- a/packages/connectors/src/Sources/MicrosoftAds/Connector.js
+++ b/packages/connectors/src/Sources/MicrosoftAds/Connector.js
@@ -17,15 +17,29 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
    * Processes all nodes defined in the fields configuration
    */
   async startImportProcess() {
-    const accountIds = FormatUtils.parseAccountIds(this.config.AccountIDs.value).map(id => id.trim());
+    const accountIds = FormatUtils.parseAccountIds(this.config.AccountIDs.value);
+
+    // A blank-but-truthy AccountIDs (e.g. ",") passes required-field validation and parses
+    // to an empty list. Without this the day loop would still run, import nothing, and walk
+    // the incremental cursor to today — silently skipping every day once the config is fixed.
+    if (!accountIds.length) {
+      throw new Error('No valid Account IDs found in the AccountIDs parameter');
+    }
+
     const fields = MicrosoftAdsHelper.parseFields(this.config.Fields.value);
     const nodeNames = Object.keys(fields);
+    nodeNames.forEach(nodeName => this.assertKnownNode(nodeName));
+
     const timeSeriesNodes = nodeNames.filter(nodeName => this.source.fieldsSchema[nodeName].isTimeSeries);
     const catalogNodes = nodeNames.filter(nodeName => !this.source.fieldsSchema[nodeName].isTimeSeries);
 
-    for (const accountId of accountIds) {
-      this.config.logMessage(`Starting import process for Account ID: ${accountId}`);
+    // Resolve the date range up front so an invalid backfill config (missing StartDate,
+    // EndDate before StartDate) fails before any catalog import rather than after it.
+    const dateRange = timeSeriesNodes.length ? this.getStartDateAndDaysToFetch() : null;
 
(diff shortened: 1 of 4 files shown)


packages/connectors/src/Sources/MicrosoftAds/Connector.js
```diff
@@ -17,111 +17,106 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
...
+            daysToFetch
+          });
+        }
       }
 
-      // Some callers defer the checkpoint until all accounts finish successfully.
-      if (updateRequestedDate && this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
+      // Every account and node stored this date, so the cursor can move past it.
+      if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
         this.config.updateLastRequstedDate(currentDate);
```
laskevych, line 109: 🟡 **Important (blocking):** The old code only checkpointed when at least one time-series fetch actually happened (`if (... && lastProcessedDate)`). The new guards cover `!timeSeriesNodes.length` and `daysToFetch <= 0`, but not an empty `accountIds` list. `FormatUtils.parseAccountIds` filters out blank entries, so a value like `","` or `"; "` parses to `[]` — and it passes required-field validation, which only checks the raw string is truthy. In that state the day loop still runs, both inner loops no-op, and `updateLastRequstedDate(currentDate)` walks the cursor all the way to today with zero rows imported. Once the user fixes their `AccountIDs`, every day older than `ReimportLookbackWindow` is permanently skipped — a silent data gap.

Could you add an early return (or throw) when `accountIds.length === 0`, plus a test for it? (The current test stub `value => value.split(',')` can never produce an empty list, which is why this case slipped past the suite.)

laskevych (commented):
Nice fix — the date-outer shape matches FacebookMarketing/TikTokAds, the checkpoint-only-after-all-accounts invariant is well documented in code, and the test suite covers the failure modes that actually caused the incident.

One blocking edge case around empty-parsed `AccountIDs` (see the 🟡 inline comment); everything else below is non-blocking. Happy to approve once that guard is in.

📝 **Note (follow-up scope):** the exact pre-fix shape (account-outer loop + deferred checkpoint, including the same "Some callers defer the checkpoint…" comment) still exists verbatim in RedditAds, XAds, CriteoAds, and GoogleAds — they'll hit the identical "import never finishes" symptom, and this PR's test file is a ready template for fixing them.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
```diff
@@ -17,111 +17,106 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
...
+    const catalogNodes = nodeNames.filter(nodeName => !this.source.fieldsSchema[nodeName].isTimeSeries);
 
-    for (const rawAccountId of accountIds) {
-      const accountId = rawAccountId.trim();
+    for (const accountId of accountIds) {
       this.config.logMessage(`Starting import process for Account ID: ${accountId}`);
 
-      for (const nodeName in fields) {
-        const processedDate = await this.processNode({
+      for (const nodeName of catalogNodes) {
```
laskevych, line 43: 🔵 **Question (non-blocking):** With the catalog phase now running for all accounts before any time-series work, a single account whose catalog fetch consistently fails (e.g. revoked access) blocks report data for *every* account — the old account-outer loop at least landed the earlier accounts' rows before failing. I realize some of this is inherent to the date-outer design (a mid-range failure for one account must stop the cursor for all), but is the all-or-nothing catalog phase an accepted trade-off here? FacebookMarketing/TikTokAds handle this with skip-inaccessible-account logic; might be worth a follow-up if partial availability matters for multi-account setups.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
```diff
@@ -17,111 +17,106 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
    * Processes all nodes defined in the fields configuration
    */
   async startImportProcess() {
-    const accountIds = FormatUtils.parseAccountIds(this.config.AccountIDs.value);
+    const accountIds = FormatUtils.parseAccountIds(this.config.AccountIDs.value).map(id => id.trim());
```
laskevych: 💡 **Suggestion (non-blocking):** `FormatUtils.parseAccountIds`'s JSDoc already promises "Array of *trimmed* account IDs", but the implementation only trims inside the `filter` predicate — its split regex `/[,;]\s*/` eats whitespace only *after* a delimiter, so `" acc1 ,acc2"` still leaks spaces. This PR adds the third per-callsite copy of the trim (RedditAds and XAds helpers carry the same `.map(id => id.trim())`). Since MicrosoftAds is currently the only consumer of the shared helper, moving the trim into `parseAccountIds` itself (one line, and it makes the JSDoc true) and dropping the map here would fix the layer the next connector author will trust.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
```diff
@@ -17,111 +17,106 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
...
-  /**
-   * Process a time series node (e.g., ad performance report)
-   * @param {Object} options - Processing options
-   * @param {string} options.nodeName - Name of the node
-   * @param {string} options.accountId - Account ID
-   * @param {Array<string>} options.fields - Array of fields to fetch
-   * @param {boolean} options.updateRequestedDate - Whether to update incremental state per processed day
-   */
-  async processTimeSeriesNode({ nodeName, accountId, fields, updateRequestedDate = true }) {
     const [startDate, daysToFetch] = this.getStartDateAndDaysToFetch();
```
laskevych: 💡 **Suggestion (non-blocking):** The date-range resolution (and its validation errors — missing `StartDate` for manual backfill, `EndDate` before `StartDate`, etc.) now only runs after the full catalog import of every account. A backfill with a bad date config re-imports the whole catalog on every retry and then fails with the same error. Resolving the range up front in `startImportProcess` (when time-series nodes are selected) would surface config errors before any API/storage work.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
```diff
@@ -17,111 +17,106 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
    * Processes all nodes defined in the fields configuration
    */
   async startImportProcess() {
-    const accountIds = FormatUtils.parseAccountIds(this.config.AccountIDs.value);
+    const accountIds = FormatUtils.parseAccountIds(this.config.AccountIDs.value).map(id => id.trim());
     const fields = MicrosoftAdsHelper.parseFields(this.config.Fields.value);
-    let lastProcessedDate = null;
+    const nodeNames = Object.keys(fields);
+    const timeSeriesNodes = nodeNames.filter(nodeName => this.source.fieldsSchema[nodeName].isTimeSeries);
```
laskevych, line 33: 💡 **Suggestion (non-blocking):** If a saved `Fields` config references a node that's no longer in `fieldsSchema` (renamed/removed after an upgrade), this filter throws a bare `Cannot read properties of undefined (reading 'isTimeSeries')` before the first `logMessage`, leaving an empty run log. Pre-existing exposure (the old `processNode` had the same deref), but since the lookup moved anyway, a guarded check with an explicit `Unknown node 'X'` error would make such runs diagnosable from run history.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
```diff
@@ -17,111 +17,106 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
...
+   * @param {string} options.accountId - Account ID
+   * @param {Date} options.date - The day to import
+   * @param {Array<string>} options.fields - Array of fields to fetch
+   * @param {number} options.dayNumber - 1-based position of the day within the run
+   * @param {number} options.daysToFetch - Total number of days in the run
+   */
+  async processTimeSeriesDay({ nodeName, accountId, date, fields, dayNumber, daysToFetch }) {
+    const formattedDate = DateUtils.formatDate(date);
+
+    this.config.logMessage(`Processing ${nodeName} for ${accountId} on ${formattedDate} (day ${dayNumber} of ${daysToFetch})`);
```
laskevych: 💡 **Suggestion (non-blocking):** The `(day X of Y)` context is re-formatted and logged once per account×node — for 5 accounts × 2 report nodes over 30 days that's 300 duplicate day-context lines per run. TikTokAds logs the day-level context once per date in the outer loop; hoisting `formattedDate` + the day-position line into the `dayOffset` loop would also drop the `dayNumber`/`daysToFetch` pass-through params. Relatedly, `Starting import process for Account ID: X` now fires only for the catalog phase (and logs even when there are no catalog nodes), long before that account's report data is touched — worth adjusting so run history reflects the actual order.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
```diff
@@ -17,111 +17,106 @@ var MicrosoftAdsConnector = class MicrosoftAdsConnector extends AbstractConnecto
...
+      this.config.logMessage('No days to fetch for time series data');
+      return;
     }
 
-    let lastProcessedDate = null;
-
-    // Process data day by day
     for (let dayOffset = 0; dayOffset < daysToFetch; dayOffset++) {
       const currentDate = new Date(startDate);
       currentDate.setDate(currentDate.getDate() + dayOffset);
```
laskevych, line 91: 📝 **Note (pre-existing, follow-up):** `currentDate.setDate(...)` mutates in local time while `DateUtils.formatDate` formats in UTC, so a range crossing a DST transition can format the same date string twice (one day fetched twice, the last day of the range skipped that run) — and with this PR the skewed `Date` is now persisted per day as the checkpoint. These lines are carried over unchanged, so not for this PR — but `setUTCDate` would close it, ideally across the sibling connectors too.

packages/connectors/test/Sources/MicrosoftAds/perDayCheckpoint.test.js
```diff
@@ -0,0 +1,217 @@
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { describe, expect, it } from 'vitest';
+import { loadGasClass } from '../../support/loadGasClass.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+globalThis.DateUtils = { formatDate: date => date.toISOString().slice(0, 10) };
+globalThis.RUN_CONFIG_TYPE = { INCREMENTAL: 'INCREMENTAL', MANUAL_BACKFILL: 'MANUAL_BACKFILL' };
+globalThis.FormatUtils = { parseAccountIds: value => value.split(',') };
```
laskevych: 🧪 **Test (non-blocking):** The `parseAccountIds` mock (`value => value.split(',')`) diverges from the real helper, which splits on `/[,;]\s*/` and filters blank entries. The "trims whitespace" test feeds `'acc1, acc2'`, which the real parser would already return clean — so it exercises the mock rather than the production pipeline, and the load-bearing cases (leading space, `'acc1 ,acc2'`, semicolons, empty entries) are uncovered. `loadGasClass`-ing the real `FormatUtils.js` (the same way `Source.test.js` loads the real Helper) would make these tests exercise real parsing.

packages/connectors/test/Sources/MicrosoftAds/perDayCheckpoint.test.js
```diff
@@ -0,0 +1,217 @@
...
+      return fetchData(params);
+    },
+  };
+  self.getStorageByNode = async () => ({ saveData });
+  self.addMissingFieldsToData = data => data;
+  self.getStartDateAndDaysToFetch = () => [startDate, daysToFetch];
+  self.config = {
+    AccountIDs: { value: accountIds },
+    Fields: { value: JSON.stringify(nodes) },
+    CreateEmptyTables: { value: false },
```
laskevych, line 56: 🧪 **Test (non-blocking):** The `CreateEmptyTables` branch was relocated into `processTimeSeriesDay` by this refactor, but the fixture hardcodes `CreateEmptyTables: { value: false }` and no test flips it. One test with `fetchData: async () => []` + `CreateEmptyTables: true` asserting `saveData` is called per day would pin the moved behavior.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
prokopetsv: Confirmed and fixed in 623a60d5 — thank you, this was a real silent-gap bug.

Verified the premise: `parseAccountIds(",")` → `["",""]` → filtered → `[]`, and the day loop then checkpointed each date having imported nothing.

`startImportProcess` now throws `No valid Account IDs found in the AccountIDs parameter` before any work. You were also right about why the suite missed it — the stub `value => value.split(',')` could never return an empty list, so the test file now loads the real `FormatUtils` and the new case feeds `', ;'`, asserting nothing is fetched and the cursor never moves.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
prokopetsv: Fixed in 623a60d5. Confirmed the split regex only eats whitespace *after* a delimiter, so ` acc1 ,acc2` really did leak spaces past the filter — `parseAccountIds` now trims before filtering (`split(/[,;]/).map(trim).filter(length)`), and the `.map(id => id.trim())` at the callsite is gone. MicrosoftAds is the only consumer of the shared helper (RedditAds/XAds carry their own), so the blast radius is this PR. The parsing test now feeds ` acc1 ; acc2 ` through the real helper.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
prokopetsv: Fixed in 623a60d5. `getStartDateAndDaysToFetch()` now runs in `startImportProcess` before the catalog loop (only when time-series nodes are selected), and the resolved range is passed into `processTimeSeriesNodes`. A backfill with a bad date config now fails before any API or storage call instead of after a full catalog import on every retry.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
prokopetsv: Fixed in 623a60d5. Added `assertKnownNode`, run over every configured node before the isTimeSeries split, so an outdated `Fields` value fails with `Unknown node 'X'. Please update the Fields configuration` instead of a bare property-of-undefined. Covered by a new test. Agreed it was pre-existing, but as you say the lookup moved anyway, so it was the right moment.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
prokopetsv: Fixed in 623a60d5. The day-context line is now logged once per date in the `dayOffset` loop, and `dayNumber`/`daysToFetch` are gone from `processTimeSeriesDay` (which now takes the already-formatted date, so it is not re-formatted per account×node either). For your 5×2×30 example that is 300 lines down to 30.

On the account line: you are right that it was misplaced. Rather than move it, I replaced it with a single run-level `Importing N node(s) for account(s): ...`, since with dates outermost there is no longer a point in the run where one account 'starts'.

packages/connectors/test/Sources/MicrosoftAds/perDayCheckpoint.test.js
prokopetsv: Fixed in 623a60d5, and this one directly caused the blocking bug above — the stub could not produce the empty list, so the case went untested. The spec now `loadGasClass`-es the real `src/Core/Utils/FormatUtils.js` (it is `var FormatUtils`, so no vm read-back needed) and drops the stub entirely. New coverage: ` acc1 ; acc2 ` for leading/trailing space and semicolons, and `', ;'` for the all-blank case.

packages/connectors/test/Sources/MicrosoftAds/perDayCheckpoint.test.js
prokopetsv: Fixed in 623a60d5. Two tests now pin the relocated branch: with `fetchData` returning `[]` and `CreateEmptyTables: true`, `saveData` is called once per day with `[]`; with the flag off, storage is never touched. Both run over a 2-day range so the per-day placement is asserted, not just the single-call case.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
prokopetsv: Good catch, and yes — accepted trade-off for this PR, not something I want to smuggle into a hotfix.

You have the mechanism exactly right: the date-outer design already means one account's mid-range failure stops the cursor for all, so the catalog phase is not introducing a new class of coupling. What it does change is *ordering* — previously account 1's rows landed before account 2 failed; now nothing lands if an early catalog fetch throws.

The proper fix is the FacebookMarketing/TikTokAds skip-inaccessible-account handling (`_skipOrRethrow` + a run warning), which is a behavioural change deserving its own PR and its own tests, especially deciding what a run means when every account is skipped. I would rather not fold that in here. Happy to open a follow-up covering MicrosoftAds and the four sibling connectors in #1557 — say the word and I will.

packages/connectors/src/Sources/MicrosoftAds/Connector.js
prokopetsv: Agreed on all three points, and agreed it is out of scope here — thank you for the precise diagnosis.

Worth recording that this PR does make the consequence worse: the skewed `Date` is now what gets persisted as the checkpoint, so a DST-straddling range can checkpoint a date that was formatted twice. The mixing is also not confined to these lines — `getStartDateAndDaysToFetch` in `AbstractConnector` does its own local-time arithmetic, so a `setUTCDate` here alone would only move the seam rather than close it.

That is why I have left it: the fix wants to be UTC end to end across `AbstractConnector` and the sibling connectors (XAds already uses `setUTCDate`, the other four do not), with tests that actually cross a transition. Tracking it as a follow-up rather than a one-line change here.

laskevych (approved):
All review points addressed in 623a60d5 — verified the diff, not just the replies. The blocking empty-AccountIDs case is now rejected before any work with a named error, the trim lives in the shared parser where its JSDoc always claimed it did, the range resolves before the catalog phase, unknown nodes fail readably, the day-context log fires once per date, and the tests now run against the real FormatUtils with the empty-list, whitespace/semicolon, and CreateEmptyTables cases pinned. The two remaining threads (catalog-phase ordering, DST date math) are explicitly accepted as follow-ups. Nice turnaround — approving.

PR #1557, opened 2026-08-21
fix(connectors): resume interrupted Ads imports from last completed date

prokopetsv:
# Problem

Google Ads, Reddit Ads, Criteo and X Ads saved their incremental progress only after every account and every node had finished the entire date range. When a long import died before that point — as a Microsoft Ads run recently did after roughly 10 hours and 42 of 49 days — `LastRequestedDate` was never written, so the interrupted-run sweep restarted the run from day one of the range, which failed the same way, repeating until someone cancelled it by hand. The root cause was the loop shape shared by all four connectors: `startImportProcess` iterated accounts on the outside and days on the inside, passing `updateRequestedDate: false` down so the single checkpoint could only fire at the very end. This applies the fix already shipped for Microsoft Ads in #6864 to the remaining four connectors built on the same pattern.

# Solution

Each connector now runs its time-series import date-outer instead of account-outer, so the incremental cursor advances as the run progresses. A date is checkpointed only after every account and every node has fetched and stored it, which keeps multi-account configurations correct — the reason the checkpoint was deferred in the first place. Catalog nodes still import once per account, up front, and `getStartDateAndDaysToFetch()` is now called once per run instead of once per node.

Connector-specific behavior was preserved rather than flattened:

- **Google Ads** — the `isGlobalResource` rule moved into a `customerIdsForNode()` helper used by both the catalog and date paths, so global resources still fetch for the first customer only.
- **Criteo** — every node in its schema is time-series, so there is no catalog split; `getStorageByNode(nodeName, fields)` is unchanged.
- **X Ads** — the involved case. Async nodes (`stats_by_country`) submit one job per 15-day chunk, so both node kinds run under a single chunk loop: chunks are whole days when no async node is selected, and 15-day chunks when one is. Two independent loops would have advanced the cursor to the last date for one node kind and then dragged it back to day one for the other. `clearCache` moved to the end of the run because, with dates outermost, each account's promoted-tweet cache stays in use until the range finishes.

Two trade-offs worth a reviewer's attention. X Ads now holds every account's promoted-tweet cache simultaneously instead of one at a time — ID arrays, not row data, but it is a real change on wide multi-account configs. And checkpoint granularity for X Ads async nodes is a 15-day chunk rather than a single day; that is a limit of the jobs API, not a shortcut.

No backend or state-storage changes were needed: `updateLastRequstedDate` already emits a message the executor persists as it arrives. After an interruption a resume re-imports only `ReimportLookbackWindow` days, and those rows upsert on each node's unique keys.

# Changes

- Replaced the account-outer day loop with a date-outer loop in the Google Ads, Reddit Ads, Criteo and X Ads connectors, checkpointing after each completed date
- Added `processTimeSeriesNodes` and `processTimeSeriesDay` to each connector, and removed `processNode` along with the `updateRequestedDate` plumbing
- Extracted `customerIdsForNode()` in Google Ads so the global-resource rule is applied on both the catalog and time-series paths
- Unified sync and async X Ads nodes under one chunk loop, split into `importAsyncNodesForChunk` and `importSyncNodesForChunk`, so the cursor stays monotonic when both kinds are selected
- Moved the X Ads unique-key validation into `assertUniqueKeysSelected()` and ran it before any fetch, so a misconfigured node fails before a full catalog import
- Moved X Ads `clearCache` to the end of the run, in a `finally` so it also runs when the import fails
- Replaced bare `console.log` calls for an empty date range with `config.logMessage`, so the message reaches run history
- Added four `perDayCheckpoint.test.js` suites (48 tests) covering per-day checkpointing, mid-run and storage-write failures, manual backfill, empty days, multi-account ordering, catalog-only configurations, empty ranges, Google Ads global resources, and the X Ads chunk boundary, mixed-node cursor and cache-clearing paths
- Used the real `XAdsHelper` in the X Ads tests instead of a reimplemented chunker, deriving the chunk size from the production helper
- Documented the checkpointing rule under "Saving incremental progress" in `packages/connectors/CONTRIBUTING.md`
- Added changeset `6865-per-day-checkpoint-ads-connectors.md`

### DIFF
--- a/packages/connectors/src/Sources/CriteoAds/Connector.js
+++ b/packages/connectors/src/Sources/CriteoAds/Connector.js
@@ -19,7 +19,25 @@ var CriteoAdsConnector = class CriteoAdsConnector extends AbstractConnector {
     const fields = CriteoAdsHelper.parseFields(this.config.Fields?.value || "");
     const advertiserIds = CriteoAdsHelper.parseAdvertiserIds(this.config.AdvertiserIDs?.value || "");
 
-    await this.processTimeSeriesNodes({ advertiserIds, timeSeriesNodes: Object.keys(fields), fields });
+    // A blank-but-truthy AdvertiserIDs (e.g. ";") passes required-field validation and parses
+    // to an empty list. Without this the day loop would still run, fetch nothing, and walk the
+    // incremental cursor to today — marking days as imported that never were.
+    if (!advertiserIds.length) {
+      throw new Error('No valid Advertiser IDs found in the AdvertiserIDs parameter');
+    }
+
+    const nodeNames = Object.keys(fields);
+    // Every node in the Criteo schema is a time series, and this connector has no catalog
+    // path, so anything else would silently be re-fetched once per day by the loop below.
+    const unsupportedNodes = nodeNames.filter(
+      nodeName => !this.source.fieldsSchema[nodeName]?.isTimeSeries
+    );
+
+    if (unsupportedNodes.length) {
+      throw new Error(`Only time series nodes are supported. Unsupported: ${unsupportedNodes.join(', ')}`);
+    }
+
+    await this.processTimeSeriesNodes({ advertiserIds, timeSeriesNodes: nodeNames, fields });
   }
 
   /**
--- a/packages/connectors/src/Sources/CriteoAds/Connector.js
+++ b/packages/connectors/src/Sources/CriteoAds/Connector.js
@@ -18,92 +18,81 @@ var CriteoAdsConnector = class CriteoAdsConnector extends AbstractConnector {
   async startImportProcess() {
     const fields = CriteoAdsHelper.parseFields(this.config.Fields?.value || "");
     const advertiserIds = CriteoAdsHelper.parseAdvertiserIds(this.config.AdvertiserIDs?.value || "");
-    let lastProcessedDate = null;
 
-    for (const advertiserId of advertiserIds) {
-      for (const nodeName in fields) {
-        const processedDate = await this.processNode({
-          nodeName,
-          advertiserId,
-          fields: fields[nodeName] || [],
-          updateRequestedDate: false
-        });
-
-        if (processedDate && (!lastProcessedDate || processedDate > lastProcessedDate)) {
-          lastProcessedDate = processedDate;
-        }
-      }
-    }
-
-    if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL && lastProcessedDate) {
-      this.config.updateLastRequstedDate(lastProcessedDate);
-    }
+    await this.processTimeSeriesNodes({ advertiserIds, timeSeriesNodes: Object.keys(fields), fields });
   }
 
   /**
-   * Process a single node for a specific advertiser
+   * Imports every node for every advertiser, one date at a time.
+   *
+   * The loop is date-outer on purpose: the incremental cursor may only move once a
+   * date is complete for *all* advertisers and nodes, so a run interrupted midway
+   * resumes from the last fully imported date instead of restarting the range.
+   *
    * @param {Object} options - Processing options
-   * @param {string} options.nodeName - Name of the node to process
-   * @param {string} options.advertiserId - Advertiser ID
-   * @param {Array<string>} options.fields - Array of fields to fetch
-   * @param {boolean} options.updateRequestedDate - Whether to update incremental state per processed day
+   * @param {Array<string>} options.advertiserIds - Advertiser IDs to import
+   * @param {Array<string>} options.timeSeriesNodes - Names of the nodes to import
+   * @param {Object} options.fields - Map of node name to the fields selected for it
    */
-  async processNode({ nodeName, advertiserId, fields, updateRequestedDate = true }) {
-    return await this.processTimeSeriesNode({
-      nodeName,
-      advertiserId,
-      fields,
-      updateRequestedDate
-    });
-  }
+  async processTimeSeriesNodes({ advertiserIds, timeSeriesNodes, fields }) {
+    if (!timeSeriesNodes.length) {
+      return;
+    }
 
-  /**
-   * Process a time series node with date-based logic
-   * @param {Object} options - Processing options
-   * @param {string} options.nodeName - Name of the node
-   * @param {string} options.advertiserId - Advertiser ID
-   * @param {Array<string>} options.fields - Array of fields to fetch
-   * @param {boolean} options.updateRequestedDate - Whether to update incremental state per processed day
-   */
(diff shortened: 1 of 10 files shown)


packages/connectors/src/Sources/CriteoAds/Connector.js
```diff
@@ -18,92 +18,81 @@ var CriteoAdsConnector = class CriteoAdsConnector extends AbstractConnector {
...
+            fields: fields[nodeName] || []
+          });
+        }
       }
 
-      // Some callers defer the checkpoint until all advertisers finish successfully.
-      if (updateRequestedDate && this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
+      // Every advertiser and node stored this date, so the cursor can move past it.
+      if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
         this.config.updateLastRequstedDate(currentDate);
```
laskevych, line 84: 🔴 **Blocking:** When `advertiserIds` parses to an empty array, the date loop still runs and checkpoints every day: the inner advertiser loop is a no-op, but `updateLastRequstedDate(currentDate)` fires for each date, so the incremental cursor sweeps the whole range while nothing is fetched. Before this PR the cursor never moved in that case (`lastProcessedDate` stayed `null`), so this is a regression from "silent no-op" to "silent data loss" — days get marked as imported that never were.

This is reachable: `AdvertiserIDs` is `isRequired`, but that only rejects an empty string — an input like `";"` or `", ,"` passes validation and `parseAdvertiserIds` filters it to `[]`. Criteo is the only connector of the four affected, because its parser is the only one with `.filter(Boolean)`: Reddit/X return `[""]` and Google Ads returns `[NaN]`, so those fail the first fetch instead of advancing the cursor.

Suggestion: bail out before the day loop, e.g.
```js
if (!advertiserIds.length) {
  this.config.logMessage('No advertiser IDs configured, nothing to import');
  return;
}
```
plus a test next to the existing "empty configurations" describe block.

packages/connectors/src/Sources/CriteoAds/Connector.js
```diff
@@ -18,92 +18,81 @@ var CriteoAdsConnector = class CriteoAdsConnector extends AbstractConnector {
...
-        if (processedDate && (!lastProcessedDate || processedDate > lastProcessedDate)) {
-          lastProcessedDate = processedDate;
-        }
-      }
-    }
 
-    if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL && lastProcessedDate) {
-      this.config.updateLastRequstedDate(lastProcessedDate);
-    }
+    await this.processTimeSeriesNodes({ advertiserIds, timeSeriesNodes: Object.keys(fields), fields });
```
laskevych: 💡 **Suggestion (non-blocking):** Criteo passes `Object.keys(fields)` straight in as `timeSeriesNodes`, while the other three connectors filter by `fieldsSchema[nodeName].isTimeSeries`. Today that's correct — all four Criteo schema nodes are `isTimeSeries: true` — but if a catalog node is ever added to the schema, it would silently run through the per-day loop and be fetched once per day. Applying the same `isTimeSeries` filter here would keep all four connectors symmetric and make that future mistake impossible. Fine to skip if you consider it out of scope.

packages/connectors/src/Sources/CriteoAds/Connector.js
prokopetsv: Confirmed and fixed in c7eb3273 — thank you, and your scoping of it to Criteo alone is exactly right.

I checked all four parsers rather than take it on trust:

| Connector | Parser on `";"` | Empty list reachable |
|---|---|---|
| Criteo | `.filter(Boolean)` → `[]` | **yes** |
| Reddit | split + trim, no filter → `["",""]` | no |
| X Ads | split + trim, no filter → `["",""]` | no |
| Google Ads | `parseIds` → `formatId` → `[NaN]` | no |

So Criteo is the only one that reaches the day loop with nothing to iterate, and you are right that this is a regression in kind rather than degree — `lastProcessedDate` previously stayed `null`, so the cursor never moved.

One deliberate deviation from your snippet: I throw rather than log-and-return.

```js
if (!advertiserIds.length) {
  throw new Error('No valid Advertiser IDs found in the AdvertiserIDs parameter');
}
```

Two reasons. A log-and-return reproduces the old behaviour, which was itself a quiet failure — the run finishes SUCCESS having imported nothing, and the user has no signal their `AdvertiserIDs` is malformed. And #1556 landed the same guard as a throw for MicrosoftAds (whose `FormatUtils.parseAccountIds` also filters), so throwing keeps the two consistent. Happy to switch to your version if you would rather this stay non-fatal — it is a one-line change.

Test added in the `empty configurations` block as suggested, asserting nothing is fetched and the cursor never moves. It also exposed that the spec's `parseAdvertiserIds` stub could not produce an empty list, so the file now loads the real `Helper.js` — same class of gap that hid the MicrosoftAds bug in #1556.

packages/connectors/src/Sources/CriteoAds/Connector.js
prokopetsv: Agreed, and fixed in c7eb3273 — I had flagged this to myself during self-review and talked myself out of it, so I am glad you raised it.

Rather than filter silently I reject the node, because Criteo has no catalog path at all:

```js
const unsupportedNodes = nodeNames.filter(n => !this.source.fieldsSchema[n]?.isTimeSeries);
if (unsupportedNodes.length) {
  throw new Error(`Only time series nodes are supported. Unsupported: ${unsupportedNodes.join(', ')}`);
}
```

A plain `isTimeSeries` filter would keep the four connectors symmetric but trade one silent failure for another — a future catalog node would simply not be imported, with nothing in run history to say so. The other three connectors route non-time-series nodes to `processCatalogNode`; Criteo has nowhere to route them, so failing loudly is the honest equivalent.

The optional chaining also covers the adjacent case a reviewer raised on #1556: a `Fields` value naming a node the schema no longer defines now surfaces as `Unsupported: <name>` instead of a bare property-of-undefined. Test added.

laskevych (approved):
All review points are addressed:

- 🔴 Empty advertiser list: now rejected up front with a clear error instead of walking the cursor over never-imported days — and throwing beats the log-and-return I suggested, since a misconfigured required field should fail the run visibly. The regression test uses the real `CriteoAdsHelper` parser, so it exercises the actual `.filter(Boolean)` path that made Criteo the only affected connector.
- 💡 Non-time-series nodes: rejecting them explicitly is the right call for a connector with no catalog path.
- 🔵 #6864 wording: resolved by events — #1556 merged, so the description is accurate now.

Ran all four perDayCheckpoint suites locally at c7eb3273 — 44/44 pass. Approving.

PR #1558, opened 2026-08-21
fix(backend): report an interrupted connector run as a warning

prokopetsv:
# Problem

When a connector process exits without sending a terminal status, the backend records a fallback entry so the run is not mistaken for a success. That entry was logged at ERROR severity, which paged on a condition that resolves itself: the interrupted-run sweep picks the run back up within 15 minutes and, now that connectors checkpoint their progress per day, it continues from the last completed date rather than restarting the range. The result was ERROR-level noise for attempts that were about to be retried successfully, sitting alongside the genuine errors that do need attention. Run history showed the same entry as `[ERROR]`, making a resumable interruption look like a failed import.

# Solution

The fallback in `ConnectorExecutorService.executeInBackground` now emits `ConnectorMessageType.WARNING` instead of `ERROR`, with a matching `logger.warn` so the persisted message type and the monitoring severity stay in step — the same pairing the adjacent cancellation and expired-credentials branch already uses.

The message deliberately stays in `configErrors` rather than moving to `configLogs`. That array is what marks the run as unsuccessful, so relocating it would let a run that produced no terminal status finish as SUCCESS. Only the severity and the run-history label change; run status, the retry sweep, and the resume flow behave exactly as before.

An earlier version of this branch also added a five-attempt resume cap. That was dropped: with per-day checkpointing, a resumed run makes forward progress, so a cap would stop long backfills that were advancing normally. Whether to stop a run stays a user decision.

`truncateMessage` in `connector-message.utils.ts` already handles the `WARNING` shape, so long messages truncate correctly on this path with no change.

# Changes

- Changed the "Connector process finished without terminal success status" fallback from `ConnectorMessageType.ERROR` to `WARNING`, including its `error` → `warning` field and `[ERROR]` → `[WARNING]` label
- Changed the accompanying `logger.error` call to `logger.warn`
- Added a comment recording why a single interrupted attempt is not individually actionable
- Added a test asserting the fallback is persisted as a warning, carries the message in `warning`, and leaves `error` unset
- Added changeset `6857-terminal-status-warning.md`

### DIFF
--- a/.changeset/6857-terminal-status-warning.md
+++ b/.changeset/6857-terminal-status-warning.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Interrupted connector runs are reported as a warning
+
+A run that stops without reporting a final status is resumed automatically, and it continues from the last day it finished rather than starting over. Run history now shows this as a warning instead of an error, so a single interrupted attempt no longer looks like a failed import.
--- a/.changeset/6857-terminal-status-warning.md
+++ b/.changeset/6857-terminal-status-warning.md
@@ -2,6 +2,6 @@
 'owox': minor
 ---
 
-# Interrupted connector runs are reported as a warning
+# Connector runs interrupted by a restart are reported as a warning
 
-A run that stops without reporting a final status is resumed automatically, and it continues from the last day it finished rather than starting over. Run history now shows this as a warning instead of an error, so a single interrupted attempt no longer looks like a failed import.
+When a server restart stops a connector run, the run resumes by itself and continues from the last day it finished. Run history now shows this as a warning instead of an error, so a restart no longer looks like a failed import. Runs that stop for any other reason are still reported as errors.
--- a/apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
+++ b/apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
@@ -702,7 +702,23 @@ describe('ConnectorExecutorService', () => {
     expect(errors.join()).not.toContain('finished without terminal success status');
   });
 
+  // The fallback fires whenever the process resolves with no terminal success status and
+  // no captured error. Only a shutdown makes that resumable, so severity splits on it.
+  const findFallbackEntry = (dataMartRunRepository: Repository<DataMartRun>, status: string) => {
+    const finalUpdate = (dataMartRunRepository.update as jest.Mock).mock.calls.find(
+      call => call[1]?.status === status
+    );
+    const entry = (finalUpdate![1].errors as string[])
+      .map(raw => JSON.parse(raw))
+      .find(parsed => `${parsed.warning ?? parsed.error}`.includes('without terminal success'));
+
+    expect(entry).toBeDefined();
+    return entry;
+  };
+
   it('still reports a failure when the connector sends a status flag and no detail', async () => {
+    // status 5 is the connector explicitly reporting failure on a healthy backend: the run
+    // ends FAILED and is never swept, so this must stay an error rather than a warning.
     const { service, dataMartRunRepository, processSpawner, emitMessage } = createService();
     (processSpawner.spawnConnector as jest.Mock).mockImplementation(async () => {
       emitMessage({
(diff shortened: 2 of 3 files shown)


apps/backend/src/data-marts/services/connector/connector-executor.service.ts
```diff
@@ -508,13 +508,17 @@ export class ConnectorExecutorService {
           });
         } else if (configErrors.length === 0) {
           const errorMessage = 'Connector process finished without terminal success status';
+          // An interrupted attempt is transient: the retry sweep resumes the run, and with
+          // per-day checkpointing it carries on from the last completed date rather than
+          // starting over. Still recorded in configErrors, so the run is not reported as
+          // successful — it is just not paged as an ERROR on its own.
           addMessageToArray(configErrors, {
-            type: ConnectorMessageType.ERROR,
+            type: ConnectorMessageType.WARNING,
```
laskevych: 🟡 **Important (blocking):** The transience rationale only covers one of the two flavors that reach this branch. The fallback fires when `spawnConnector` *resolves* without a success status and with no captured errors — and `spawnConnector` resolves on exit 0, or on non-zero exit *only while `gracefulShutdownService.isInShutdownMode()`*. Only that shutdown flavor later becomes `INTERRUPTED` (`updateRunStatus`, the sole writer, checks `isInShutdownMode()`), and `executeInterruptedRuns` sweeps `INTERRUPTED` only. Every other path here — a connector that exits 0 without emitting `IMPORT_DONE` (silent bug, lost final STATUS line) or reports `STATUS: ERROR` with no error detail — lands terminal `FAILED` and is never resumed, yet its only persisted record is now a WARNING, and `logger.warn` replaces the error-level log. That also means a fleet-wide regression (e.g. a runner build that stops emitting the final STATUS line) would produce zero error-level signal while every run in the product fails.

The catch branch below already solves this exact problem by scoping the downgrade to known-transient causes (`signal?.aborted`, `CredentialsExpiredException`) and keeping unknown causes at ERROR. Could this branch do the same — WARNING only when `this.gracefulShutdownService.isInShutdownMode()` (already injected and consulted three lines away), ERROR otherwise?

apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
```diff
@@ -724,6 +724,34 @@ describe('ConnectorExecutorService', () => {
     );
   });
 
+  it('records the missing terminal status as a warning, not an error', async () => {
```
laskevych: 🧪 **Test (fix together with the blocking comment):** The scenario here doesn't match the name or the comment: `status: 5` is `Core.EXECUTION_STATUS.ERROR` — the connector *explicitly reporting failure* (routed to `configLogs`, which is why `configErrors` stays empty) — not an interrupted process. That's exactly the flavor that persists `FAILED` and is never swept, so the comment "The retry sweep resumes an interrupted run" documents a guarantee this scenario doesn't have. The motivating case from the changeset — a process that stops with *no* terminal message at all — has no test (`spawnConnector` mocked as emitting nothing). If the severity gets scoped by shutdown mode, this test naturally splits into the ERROR-stays-ERROR case and a shutdown-mode WARNING case.

apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
```diff
@@ -724,6 +724,34 @@ describe('ConnectorExecutorService', () => {
...
+    // The retry sweep resumes an interrupted run, and per-day checkpointing means it
+    // continues from the last completed date — so one such attempt is not individually
+    // actionable. It still counts as a run failure, hence errors rather than logs.
+    const { service, dataMartRunRepository, processSpawner, emitMessage } = createService();
+    (processSpawner.spawnConnector as jest.Mock).mockImplementation(async () => {
```
laskevych, line 757: 💡 **Suggestion (non-blocking):** This arrange/act is byte-identical to the existing `'still reports a failure when the connector sends a status flag and no detail'` test right above — same mock, same status 5, same FAILED-update lookup. Consider folding the three new assertions into that test instead of maintaining two copies of one scenario. Also minor: `fallback` comes from a `.find()` and is dereferenced unchecked, so a regression fails with `TypeError: Cannot read properties of undefined` instead of a readable diff — an `expect(fallback).toBeDefined()` first (or a string-level `toContain` like the sibling tests) keeps the failure output diagnosable.

.changeset/6857-terminal-status-warning.md
```diff
@@ -0,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Interrupted connector runs are reported as a warning
+
+A run that stops without reporting a final status is resumed automatically, and it continues from the last day it finished rather than starting over. Run history now shows this as a warning instead of an error, so a single interrupted attempt no longer looks like a failed import.
```
laskevych: 🟡 **Important (non-blocking, text-only):** "A run that stops without reporting a final status is resumed automatically" overpromises — auto-resume only happens when the backend itself was shutting down at the moment the run ended (the only writer of `INTERRUPTED`). For a healthy-backend case the run stays `FAILED` and only the *next scheduled run* continues from the checkpoint — and a one-off manual run doesn't get that either. Suggest wording along the lines of "…is safe to retry: the next run continues from the last completed day", which stays true in every case.

laskevych (commented):
The mechanical edit itself is verified clean: the `warning` field matches the WARNING message schema, `truncateMessage` already handles the WARNING shape, the web run-history views partition WARNING entries correctly (yellow, `addWarningToCurrentStatus`), and the failure notification still fires off run status, which this PR doesn't touch. The one blocking point is scoping: this branch is a shared exit for "shutdown killed the child (transient, resumed)" and "child ended without a terminal status for an unknown reason (persistent, never resumed)" — the downgrade should distinguish them the way the catch branch below already distinguishes its causes. Details in the inline comments.

💡 **Suggestion (non-blocking, adjacent):** while in this area — `connector-message.utils.ts:57-62`: the `addMessageToArray` overflow sentinel has the same type/label drift this PR fixes (typed `ConnectorMessageType.ERROR`, but its `toFormattedString` says `[WARNING]`), so the raw view buckets it as a red error labeled "[WARNING]". One-line alignment candidate.

apps/backend/src/data-marts/services/connector/connector-executor.service.ts
prokopetsv: You are right, and this was the wrong shape of fix — fixed in a99e31d0.

I traced each path rather than take it on trust:

| Path into this branch | Terminal status | Swept for resumption |
|---|---|---|
| exit 0, no `IMPORT_DONE` | FAILED | never |
| exit 0, `STATUS: ERROR` with no detail (routed to `configLogs`, so `configErrors` stays empty) | FAILED | never |
| non-zero exit during shutdown (`connector-process-spawner.service.ts:145-151` resolves) | INTERRUPTED (`:836`, sole writer) | yes |

So my rationale held for one of three, and I had downgraded all three. Your fleet-regression scenario is the one that convinced me: a runner build that stops emitting the final STATUS line would fail every run in the product while emitting nothing above WARNING.

Severity now splits on `isInShutdownMode()`, mirroring the `isWarning` shape in the catch branch directly below so the two read the same way.

apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
prokopetsv: Correct on both counts, fixed in a99e31d0 — and worse than mislabelled: under the new scoping that test asserted the exact opposite of the intended behaviour. `status: 5` on a healthy backend is the connector explicitly reporting failure, which must stay an ERROR.

The file now covers the split properly:

- `still reports a failure when the connector sends a status flag and no detail` — `status: 5`, asserts `type === ERROR` and `warning` unset
- `keeps a silent exit an error when the backend is healthy` — `spawnConnector` resolves emitting nothing, the changeset's motivating case, asserts ERROR
- `records a shutdown-interrupted run as a warning, not an error` — `isInShutdownMode()` true, asserts the run lands `INTERRUPTED` and the entry is a WARNING

The middle one is new coverage you correctly identified as missing.

apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
prokopetsv: Both done in a99e31d0. The duplicate is gone — its assertions are folded into the existing `status: 5` test, which now checks the persisted type rather than just the message text.

On the unchecked deref: extracted a `findFallbackEntry(repo, status)` helper shared by the three tests, which asserts `expect(entry).toBeDefined()` before returning. A regression now fails on the missing entry instead of a `TypeError`, and locating the entry by status is no longer copy-pasted three times.

.changeset/6857-terminal-status-warning.md
prokopetsv: Agreed — fixed in a99e31d0. "resumed automatically" was doing work the code does not do outside shutdown, and with the severity now scoped to that case the text had to follow.

New wording:

> When a server restart stops a connector run, the run resumes by itself and continues from the last day it finished. Run history now shows this as a warning instead of an error, so a restart no longer looks like a failed import. Runs that stop for any other reason are still reported as errors.

I went with naming the restart explicitly rather than your "safe to retry" phrasing, because the scoped fix means the changeset now describes a narrower behaviour than before: only the restart case changes severity. The last sentence makes the boundary explicit so nobody reads it as a blanket downgrade.

laskevych (approved):
Verified a99e31d0 against the diff, not just the replies. Severity now splits on isInShutdownMode() exactly mirroring the catch branch's isWarning shape, so the resumable shutdown case is a WARNING while a silent exit or detail-less STATUS: ERROR on a healthy backend stays an ERROR with an error-level log. The path table in your reply matches what I traced on my side. Tests now pin all three flavors — explicit ERROR, silent exit (the previously uncovered motivating case), and shutdown-interrupted asserting both the INTERRUPTED status and the WARNING entry — with the duplicate folded away and findFallbackEntry guarding the deref. The reworded changeset now claims exactly what the code does, and naming the restart explicitly is indeed clearer than my phrasing given the narrower scope. Approving — the E2E shards were still running as of this review, so give them a moment before merging.

PR #1560, opened 2026-08-24
feat: merge null/empty filter operators into a single is blank pair

laskevych:
## Why

A blank spreadsheet cell renders identically whether the warehouse holds `NULL`, an empty string, or whitespace-only text. Offering **is empty** and **is null** side by side therefore asks the user to choose between two operators whose difference is invisible in their own output — and the wrong pick silently returns 0 rows, which looks like a valid result. On top of that, `is_empty` was already implemented as `(col IS NULL OR col = '')` on every storage, so the two operators overlapped rather than complemented each other.

## What

- The filter picker now offers a single pair, **is blank** / **is not blank**, on every column type. Semantics follow what the cell shows, not warehouse state:
  - string columns: `col IS NULL OR TRIM(col) = ''` (NULL, `''`, and whitespace-only all count as blank);
  - number / date / time / boolean: `col IS NULL`.
- One shared `renderBlankFragment` in the base SQL clause renderer serves all five storages (BigQuery, Athena, Redshift, Snowflake, Databricks); the Redshift query builder now threads the column-type resolver so the fragment can branch on type.
- MCP `query_data_mart` and the report tools advertise only the new pair (field-type matrix, error messages, boolean menu).
- api-client types and `docs/api/api-client.md` updated.

## Backwards compatibility

Nothing saved breaks: `is_empty`, `is_not_empty`, `is_null`, `is_not_null` stay accepted by the schema, validator, renderers, and MCP mappers with their exact previous semantics. Saved rules keep rendering under their original labels — the filter editor appends a legacy operator to the menu when an existing rule carries one. Only the menus stop offering them.

## Tests

- Per-storage renderer specs: type-aware emission for the new pair (string TRIM form, NULL-only for other/unknown types).
- Schema, output-controls validator, MCP field-type matrix / details tool, and web picker/editor specs updated and extended (legacy operators asserted as accepted-but-not-advertised).
- Redshift emission e2e extended with an `is_blank` filter on a VARCHAR column to cover the composer → builder → renderer type-resolution seam.


apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-clause-renderer.ts
```diff
@@ -132,6 +132,10 @@ export class SnowflakeClauseRenderer extends SqlClauseRenderer {
           sql: `(${col} IS NULL OR REGEXP_INSTR(${col}, ${text(rule.value)}) = 0)`,
           params: [],
         };
+      case 'is_blank':
+      case 'is_not_blank':
+        return this.renderBlankFragment(rule.operator, col, columnType);
```
kalnik-a-a, line 174: **Criticality: HIGH / BLOCKING**

Snowflake's one-argument `TRIM(expr)` removes only the ordinary space character by default; tabs and line endings are not removed unless the trim character set is provided explicitly ([Snowflake documentation](https://docs.snowflake.com/en/sql-reference/functions/trim)). Because this branch delegates to the base `TRIM(...)` implementation, values consisting only of `\t`, `\r`, or `\n` are classified as not blank. That contradicts this PR's promised cross-storage semantics and silently returns the wrong rows. Please provide a Snowflake-specific `blankTrimmedExpression` for the explicit space/tab/CR/LF set and cover those characters with a dialect or warehouse-semantic test.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
```diff
@@ -37,6 +38,22 @@ export function collectSchemaFieldPathTypes(
...
+ * REPEATED field stores its ELEMENT type (`STRING` + mode `REPEATED`), but the column is
+ * an ARRAY<STRING>: string operators and TRIM() are type errors on it, and only the
+ * type-agnostic operators (is_blank / is_null pairs, rendered as bare `col IS NULL`) are
+ * valid SQL. Wrapping the collected type as `ARRAY<T>` files it under the `other`
+ * category everywhere downstream — validator gating, the MCP field-type matrix, and the
+ * renderers' blank/cast branches — so all three surfaces agree (#6779).
+ */
+function comparisonType(field: DataMartSchemaField): string {
+  const rawType = String(field.type);
+  return 'mode' in field && field.mode === BigQueryFieldMode.REPEATED
```
kalnik-a-a, line 57: **Criticality: HIGH / BLOCKING**

This effective BigQuery type never reaches the picker. `flattenNativeFields()` drops `mode`, so a REPEATED STRING remains `STRING` in the web type map and the UI offers string operators such as `contains`, which the backend now rejects as `INVALID_OPERATOR_FOR_TYPE`. The same UI helper still returns no operators for ARRAY/MAP/STRUCT/JSON/VARBINARY/GEOGRAPHY and hides the filter action, although the PR promises the blank pair on every column type. Please propagate the same effective comparison type to the UI and make the `other` fallback expose only `is_blank` / `is_not_blank`, with a picker-to-validator regression test for REPEATED and complex types.

apps/backend/src/ee/mcp/tools/query-data-mart.input.ts
```diff
@@ -172,16 +180,26 @@ export const queryDataMartInputSchema = z
...
+// Every ACCEPTED operator maps to the internal FilterRule.
 export const SUPPORTED_MCP_OPERATORS = McpOperatorEnum.options;
 
+/**
+ * The operators the tool descriptions and the field-type matrix advertise: everything
+ * accepted minus the legacy null/empty cluster, which only stays parseable for
+ * pre-merge callers (#6779). Error messages list these — steering a failed call to
+ * is_blank, never back to a legacy name.
+ */
+export const ADVERTISED_MCP_OPERATORS = SUPPORTED_MCP_OPERATORS.filter(
```
kalnik-a-a: **Criticality: HIGH / BLOCKING**

This filters only the descriptions and field-type matrix. The actual Zod input schema still uses `z.enum(MCP_OPERATORS)`, and that array still contains `is_empty`, `is_not_empty`, `is_null`, and `is_not_null`. Since `query_data_mart`, `add_report`, and `update_report` all use `makeMcpFilterSchema()`, their machine-readable JSON Schema continues to advertise the four compatibility-only names. Please separate the advertised input enum from compatibility parsing and add a contract test that inspects the generated JSON Schema for all three tools.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
```diff
@@ -153,6 +153,31 @@ export class ReportSqlComposerService {
       ? new Map(collectSchemaFieldPathTypes(schemaFields).map(f => [f.name, f.type]))
       : undefined;
 
+    // is_blank / is_not_blank branch on the column's type; with no actualized schema
+    // there is nothing to branch on, and silently degrading a string column to the
+    // NULL-only form would return wrong rows (#6779). Refuse loudly instead — the
+    // paths that can reach here without types (MCP queries, the generated-SQL
+    // preview) actualize the schema and retry; run/stream actualize beforehand.
+    if (!columnTypes) {
```
kalnik-a-a: **Criticality: HIGH / BLOCKING**

This guard checks only whether the map exists. If a persisted schema contains fields but every top-level field is hidden, `schemaFields.length` is non-zero while `collectSchemaFieldPathTypes()` produces an empty `Map`, so the guard does not run. The existing no-visible-schema validator path also skips post-join rules; the new renderer then receives an unknown type and silently renders a string `is_blank` as bare `IS NULL`, no longer matching `''` or whitespace. This wrong result is introduced by the new type-dependent operator; legacy `is_empty` did not depend on column type. Please require a resolved type for every blank-filter column and add an all-hidden STRING regression test.

docs/api/api-client.md
```diff
@@ -447,13 +447,15 @@ Each filter rule is `{ column, operator, value? }`. The shape of `value` depends
...
 
 `relative_date` presets: `{ kind }` with `today`, `yesterday`, `this_week`, `last_week`, `this_month`, `last_month`, `this_quarter`, `last_quarter`, or `this_year`, and `{ kind, n }` with `last_n_days`, `last_n_months`, or `next_n_days` (`n` is a positive integer, up to 3650). Weeks are ISO weeks (Monday start) on every storage, and `last_n_days` / `next_n_days` both include today.
 
-The negative operators `neq`, `not_contains`, `not_regex`, and `not_in` are NULL-inclusive: they keep rows where the column is `NULL`, treating a missing value as "not equal to". Add a separate `is_not_null` (or `is_not_empty`) rule on the same column to also drop missing values.
+`is_blank` / `is_not_blank` match by what a rendered cell shows: on string columns a value is blank when it is `NULL`, `''`, or whitespace-only; on every other type it is blank only when `NULL`. The pre-merge operators `is_empty`, `is_not_empty`, `is_null`, and `is_not_null` are still accepted for saved configs and older callers, but new rules should use the blank pair.
```
kalnik-a-a, line 456: **Criticality: MEDIUM / NON-BLOCKING**

This API reference now teaches the new pair, but the public setup documentation remains inconsistent: `docs/getting-started/setup-guide/output-controls.md` still lists and recommends the four old operators, while `docs/getting-started/setup-guide/mcp.md` still tells users to add `is not null`. Please update those user-facing tables and guidance with `is blank` / `is not blank` and the type-aware semantics so new users are not directed back to compatibility-only names.

kalnik-a-a (changes requested):
**Criticality: HIGH / BLOCKING**

Requesting changes because four blocking correctness/contract issues remain. One additional **MEDIUM / NON-BLOCKING** documentation consistency note is included inline.

apps/backend/src/data-marts/data-storage-types/snowflake/services/snowflake-clause-renderer.ts
laskevych: Fixed in dbeb19fa3 — you're right, Snowflake's one-argument `TRIM` strips only the blank space. `SnowflakeClauseRenderer` now overrides `blankTrimmedExpression` with `TRIM(col, ' \t\n\r')`, and the renderer spec asserts the explicit-set form for both `is_blank` and `is_not_blank` — tab/CR/LF-only cells now count as blank, same as the other dialects.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
laskevych: Fixed in dbeb19fa3. The effective comparison type now reaches the picker: `flattenNativeFields` marks a REPEATED field as `ARRAY<T>` (mirroring the backend collector; `NativeField` gained `mode`), and `operatorsForType` falls back to a blank-pair-only menu for every type outside the known category sets (ARRAY/MAP/STRUCT/JSON/VARBINARY/GEOGRAPHY) instead of hiding the filter action. Regression tests pin both sides: `ARRAY<STRING>` and the complex types offer exactly `is_blank` / `is_not_blank` — which the validator accepts type-agnostically — and string operators are no longer offered where the backend would reject them with `INVALID_OPERATOR_FOR_TYPE`.

apps/backend/src/ee/mcp/tools/query-data-mart.input.ts
laskevych: Fixed in dbeb19fa3. The operator field is now a fresh `z.enum(ADVERTISED).catch(...).superRefine(...)` per use: zod-to-json-schema renders ZodCatch/ZodEffects as their inner type, so the published JSON Schema of `query_data_mart`, `add_report`, and `update_report` advertises only the blank pair, while the catch keeps the legacy cluster parseable with unchanged semantics (a missing operator stays required; an unknown name gets the operator-guidance message). The new contract spec `mcp-operator-advertising.spec.ts` walks the generated JSON Schema of all three tools — asserting no legacy name appears anywhere and every operator enum equals the advertised list — and covers the accepted-but-not-advertised parsing behavior.

apps/backend/src/data-marts/services/report-sql-composer.service.ts
laskevych: Fixed in dbeb19fa3. The guard now requires a resolved type for every blank-filter column instead of checking that the map exists: any `is_blank` / `is_not_blank` rule whose column is absent from the collected type map is rejected with `BLANK_FILTER_COLUMN_TYPE_UNRESOLVED` (one error entry per column), covering both the never-actualized schema and the all-hidden schema (non-empty field list, empty map). Added the all-hidden STRING regression test alongside the no-schema one; legacy `is_null` still composes untyped, since it does not depend on the column type.

docs/api/api-client.md
laskevych: Updated in dbeb19fa3 — the operator tables in `docs/getting-started/setup-guide/output-controls.md` now list the blank pair per column type, with a note on its type-aware semantics and on saved legacy rules keeping their labels, and `mcp.md`'s NULL-inclusive-negatives guidance now points to `is_not_blank`.

kalnik-a-a (approved):
All five review findings have been addressed. The fixes and regression coverage look good. Approved.

PR #1561, opened 2026-08-24
test(web): pin the clock in formatDataLastUpdatedLabel specs

laskevych:
## Why

`formatDataLastUpdatedLabel` reads the real clock, so the spec's fixed fixture drifts one day further from "now" on every run. Thirty days after the fixture date the day step gives way to the month step, `Intl.RelativeTimeFormat` with `numeric: 'auto'` renders **"last month"** instead of "N days ago", and the `/ago$/` assertion fails. The time bomb went off today at 08:30 UTC — CI runs before that passed, runs after it fail, on **every** branch (`Overall test @owox/web`).

## What

Freeze the system time with vitest fake timers in the `formatDataLastUpdatedLabel` describe block, so the fixture is always three days old. The `formatRelativeTime` specs already inject `now` explicitly and are unaffected.


laskevych:
Folding this into #1560 instead — the fix ships there.

PR #1563, opened 2026-08-24
Connect Google Sheets source with Google OAuth

AnastasiaKhukhra:
Users can authorize a Google account and choose a spreadsheet with Google Picker when configuring a Google Sheets source. Service account authentication remains available.

### DIFF
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/impl/GoogleSheetsOauthRender.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/impl/GoogleSheetsOauthRender.tsx
@@ -1,0 +1,99 @@
+import {
+  GoogleSheetsLoginButton,
+  type GoogleSheetsLoginResponse,
+} from '../../../../../../../shared/components/GoogleSheetsLoginButton';
+import type { OauthRenderComponentProps } from '../OauthRenderFactory';
+import { useGoogleSheetsPicker } from '../../../../../../../../google-oauth/hooks/useGoogleDrivePicker';
+import { Button } from '@owox/ui/components/button';
+import { FileSpreadsheet } from 'lucide-react';
+import { useState } from 'react';
+
+export function GoogleSheetsOauthRender({
+  isLoading,
+  status,
+  settings,
+  onOAuthSuccess,
+  configuration,
+  onValueChange,
+}: OauthRenderComponentProps) {
+  const [isPickingSpreadsheet, setIsPickingSpreadsheet] = useState(false);
+  const [pickerError, setPickerError] = useState<string | null>(null);
+  const { openPicker } = useGoogleSheetsPicker();
+
+  const handleGoogleLogin = (response: GoogleSheetsLoginResponse) => {
+    void onOAuthSuccess({ code: response.code });
+  };
+
+  const clientId = settings?.vars.ClientId as string | undefined;
+  const pickerApiKey = settings?.vars.PickerApiKey as string | undefined;
+  const projectNumber = settings?.vars.ProjectNumber as string | undefined;
+  const spreadsheetSelected =
+    typeof configuration.SpreadsheetId === 'string' && configuration.SpreadsheetId.trim() !== '';
+  const connectedEmail = status?.user?.email ?? status?.user?.name;
+
+  const handlePickSpreadsheet = () => {
+    if (!clientId || !pickerApiKey || !projectNumber) {
+      setPickerError('Google Picker configuration is incomplete');
+      return;
+    }
+    if (!connectedEmail?.includes('@')) {
+      setPickerError('Reconnect Google Sheets before choosing a spreadsheet');
+      return;
+    }
+
+    setPickerError(null);
+    setIsPickingSpreadsheet(true);
+    void openPicker({
+      apiKey: pickerApiKey,
+      appId: projectNumber,
+      clientId,
+      hintEmail: connectedEmail,
+      onPicked: spreadsheet => {
+        onValueChange('SpreadsheetId', spreadsheet.url);
+      },
+      onError: setPickerError,
+    }).finally(() => {
+      setIsPickingSpreadsheet(false);
+    });
+  };
+
+  return (
+    <div className='mt-2 mb-2 space-y-4'>
+      <GoogleSheetsLoginButton
+        clientId={settings?.vars.ClientId as string}
+        redirectUri={settings?.vars.RedirectUri as string}
+        onSuccess={handleGoogleLogin}
+        disabled={isLoading}
+      >
+        {status?.user ? (
+          <>
+            Connected as <strong>{status.user.name ?? status.user.id}</strong>
+          </>
+        ) : (
+          'Sign in with Google'
+        )}
+      </GoogleSheetsLoginButton>
+      {status?.valid && (
+        <Button
+          type='button'
+          variant='outline'
+          onClick={handlePickSpreadsheet}
+          disabled={isLoading || isPickingSpreadsheet}
+          className='w-full'
+        >
+          <FileSpreadsheet className='h-4 w-4' />
+          {isPickingSpreadsheet
+            ? 'Opening Google Picker...'
+            : spreadsheetSelected
+              ? 'Change spreadsheet'
+              : 'Choose spreadsheet'}
+        </Button>
+      )}
+      {pickerError && (
+        <div role='alert' className='text-destructive text-sm'>
+          {pickerError}
+        </div>
+      )}
+    </div>
+  );
+}
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/impl/GoogleSheetsOauthRender.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/impl/GoogleSheetsOauthRender.tsx
@@ -3,7 +3,7 @@ import {
   type GoogleSheetsLoginResponse,
 } from '../../../../../../../shared/components/GoogleSheetsLoginButton';
 import type { OauthRenderComponentProps } from '../OauthRenderFactory';
-import { useGoogleSheetsPicker } from '../../../../../../../../google-oauth/hooks/useGoogleDrivePicker';
+import { useGoogleSheetsPicker } from '../../../../../../../../google-oauth';
 import { Button } from '@owox/ui/components/button';
 import { ExternalAnchor } from '@owox/ui/components/common/external-anchor';
 import { FileSpreadsheet } from 'lucide-react';
@@ -21,10 +21,23 @@ export function GoogleSheetsOauthRender({
   const [pickerError, setPickerError] = useState<string | null>(null);
   const { openPicker } = useGoogleSheetsPicker();
 
+  const connectedEmail = status?.user?.email ?? status?.user?.name;
+
   const handleGoogleLogin = async (response: GoogleSheetsLoginResponse) => {
-    const connected = await onOAuthSuccess({ code: response.code });
-    if (connected) {
+    const result = await onOAuthSuccess({ code: response.code });
+    const nextEmail = result?.user.email ?? result?.user.name;
+
+    if (
+      result &&
+      connectedEmail &&
+      nextEmail &&
+      nextEmail.toLowerCase() !== connectedEmail.toLowerCase() &&
+      configuration.SpreadsheetId
+    ) {
       onValueChange('SpreadsheetId', '');
+    }
+    if (result) {
+      setPickerError(null);
     }
   };
 
@@ -34,8 +47,6 @@ export function GoogleSheetsOauthRender({
   const spreadsheetUrl =
     typeof configuration.SpreadsheetId === 'string' ? configuration.SpreadsheetId.trim() : '';
   const spreadsheetSelected = spreadsheetUrl !== '';
-  const connectedEmail = status?.user?.email ?? status?.user?.name;
-
   const handlePickSpreadsheet = () => {
     if (!clientId || !pickerApiKey || !projectNumber) {
       setPickerError('Google Picker configuration is incomplete');
--- a/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/impl/GoogleSheetsOauthRender.tsx
+++ b/apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/impl/GoogleSheetsOauthRender.tsx
@@ -21,8 +21,11 @@ export function GoogleSheetsOauthRender({
   const [pickerError, setPickerError] = useState<string | null>(null);
   const { openPicker } = useGoogleSheetsPicker();
 
-  const handleGoogleLogin = (response: GoogleSheetsLoginResponse) => {
-    void onOAuthSuccess({ code: response.code });
+  const handleGoogleLogin = async (response: GoogleSheetsLoginResponse) => {
+    const connected = await onOAuthSuccess({ code: response.code });
+    if (connected) {
+      onValueChange('SpreadsheetId', '');
+    }
   };
 
   const clientId = settings?.vars.ClientId as string | undefined;
@@ -64,7 +67,7 @@ export function GoogleSheetsOauthRender({
       <GoogleSheetsLoginButton
         clientId={settings?.vars.ClientId as string}
         redirectUri={settings?.vars.RedirectUri as string}
-        onSuccess={handleGoogleLogin}
+        onSuccess={response => void handleGoogleLogin(response)}
         disabled={isLoading}
       >
         {status?.user ? (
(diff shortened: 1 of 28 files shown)


apps/web/src/features/connectors/edit/components/ConnectorEditForm/steps/ConfigurationStep/Oauth/impl/GoogleSheetsOauthRender.tsx
```diff
@@ -0,0 +1,111 @@
...
+  onValueChange,
+}: OauthRenderComponentProps) {
+  const [isPickingSpreadsheet, setIsPickingSpreadsheet] = useState(false);
+  const [pickerError, setPickerError] = useState<string | null>(null);
+  const { openPicker } = useGoogleSheetsPicker();
+
+  const handleGoogleLogin = async (response: GoogleSheetsLoginResponse) => {
+    const connected = await onOAuthSuccess({ code: response.code });
+    if (connected) {
+      onValueChange('SpreadsheetId', '');
```
zapolsky, line 37: **[Bug — verified]** Re-authorising with the **same** Google account wipes the spreadsheet the user already picked. The intent was to clear it only when the account *changes* — the test in this PR is literally named `clears the selected spreadsheet after connecting another account` — but the implementation clears on every successful exchange.

**Evidence chain:**
1. `GoogleSheetsOauthRender.tsx:25-28` — the only gate is the boolean from `onOAuthSuccess`; no account comparison exists anywhere in `features/connectors` or `features/google-oauth`.
2. `GoogleSheetsOauthRender.tsx:67-80` — the `Connected as <email>` chip **is** the `children` of `GoogleSheetsLoginButton`, not a separate status element.
3. `GoogleSheetsLoginButton.tsx:102-110` — `handleLogin` has no "already connected" guard; `isDisabled = disabled || isLoading || !clientId || !redirectUri`. One click on that chip reopens consent (`prompt=consent` is forced at `:87`).
4. `OauthRenderFactory.tsx:167` binds `exchanged`, `:184` returns a bare `true` — the identity needed for the intended condition is discarded one line before the caller needs it. It is genuinely available: `connector.service.ts:209-214` returns `user: exchanged.user`, and `Source.js:287` builds `{ id, name: <email> }`.
5. Recovery is constrained by this PR's own rule: `ConfigurationStep.tsx:353-355` filters `SpreadsheetId` out of `visibleSpecifications` in managed mode, so the field cannot be retyped. The `Manually` link (`OauthRenderFactory.tsx:361-376`) is not a recovery path — it does `onValueChange('AuthType', { oauth2: {} })`, destroying `_source_credential_id` as well.
6. Reachability is proven by the PR's own fixture: `ConfigurationStep.test.tsx` renders `initialConfiguration = { AuthType: { oauth2: { _source_credential_id: 'credential-1' } }, SpreadsheetId: <url> }` with `checkStatus → { valid: true, user }` — i.e. a saved Data Mart opened for editing shows the read-only link and the clickable connected chip at the same time.

**Scope of the damage (checked, so the severity is not overstated):** what is lost is unsaved form state, not a saved Data Mart — `SpreadsheetId` is `isRequired`, so `ConfigurationStep.tsx:257-258` marks the config invalid and Next/Finish stays disabled rather than persisting an empty value. Recovery costs ~3-4 interactions (re-open Picker, possible account chooser, find the sheet again) and the user must remember which spreadsheet it was.

**Fix** — the account is available upstream, so pass it through instead of a boolean:
1. `OauthRenderFactory.tsx:42` → `onOAuthSuccess: (credentials) => Promise<OAuthCallbackResponseDto | null>`
2. `OauthRenderFactory.tsx:184` → `return exchanged;` (and `:189` → `return null;`)
3. here → clear only on a real change:
```ts
const result = await onOAuthSuccess({ code: response.code });
const nextEmail = result?.user.email ?? result?.user.name;
if (result && nextEmail?.toLowerCase() !== connectedEmail?.toLowerCase()) {
  onValueChange('SpreadsheetId', '');
}
```
Then split the existing test into a same-account case (value kept) and a genuinely different-account case (value cleared), as its name already promises.

zapolsky (commented):
🔴 **Blocked issue** — see the inline comment. It should be addressed before merge, but this review does not formally block; a human owns the merge decision.

Scope: worktree checkout at `4665e16`, 5 parallel analysis agents + 3 adversarial skeptics. 3 blocker candidates were raised, **2 were refuted by the skeptic gate and dropped** (details in the top-level comment). All 20 CI checks are green. Posted by automated OWOX review.

zapolsky:

One blocked issue was posted inline. **Three blocker candidates went through the adversarial gate; two were refuted and dropped** — the refutations are recorded below because they are useful conclusions in their own right. All 20 CI checks are green on `4665e16`.

Primary notes (criticality ≥ 80), highest first — none of these block merge.

- `apps/web/.../steps/ConfigurationStep.tsx:353-355` — **managed OAuth hides the required `SpreadsheetId`, and the only control that brings it back destroys the OAuth connection.** The spec is filtered out of `visibleSpecifications`, so the Picker becomes the only way to satisfy a required field. The sole escape, `Manually` (`OauthRenderFactory.tsx:361-376`), does `onValueChange('AuthType', { oauth2: {} })` — wiping `_source_credential_id` — after which `ConfigurationStep.tsx:216-232` demands a raw `RefreshToken`/`ClientId`/`ClientSecret` the user does not have. To be fair to the PR: that wipe is **pre-existing** behaviour shared by every OAuth connector; what is new is that hiding `SpreadsheetId` promotes it to the only manual path. Aggravating sub-case: `isOAuthEnabled` (`connector.service.ts:161-173`) only inspects vars that are `required && SECRET && store==='env'`, and `RedirectUri` is declared `[UI, REQUIRED]` with no `SECRET` (`Source.js:57-63`) — so with `OAUTH_GOOGLE_SHEETS_REDIRECT_URI` unset the UI enters managed mode with a permanently disabled sign-in button (`GoogleSheetsLoginButton.tsx:110`), no spreadsheet field, and no error (the `OAuth not configured` label is unreachable because `children` is always supplied). Consider rendering `SpreadsheetId` read-only with the Picker as its editor instead of removing it, and gating the hide on the Picker actually being usable.
- `packages/connectors/src/Sources/GoogleSheets/Source.js:31-33` — **putting `oauth2` first in the `oneOf` changes the default auth tab for every new Google Sheets source.** `ConfigurationOneOfRender.tsx:28-37` falls back to `oneOf[0]` for an empty config. On any deployment without the five new env vars, `isOAuthEnabled` is false, so `OauthRenderFactory.tsx:248` renders the manual Refresh Token / Client ID / Client Secret form as the landing state — with the `Back to OAuth` link suppressed — where users previously landed on the working Service Account field. Either declare `service_account` first, or make the fallback skip an `OAUTH_FLOW` option that is not configured.
- `apps/web/src/features/google-oauth/hooks/useGoogleDrivePicker.ts:252-276` — **the Picker is never torn down and its promise can never settle.** The built picker is not kept in a ref and is never `dispose()`d/`setVisible(false)`d; the wrapping promise resolves only on `PICKED` or `CANCEL`, with no `default` branch. `loadPicker` has a 10s timeout (`:146-150`), but the visible-picker phase and `requestAccessToken` (`:168-201`) have none — so any other action leaves the promise pending, `GoogleSheetsOauthRender.tsx:60-62`'s `.finally` never runs, and the button stays disabled on `Opening Google Picker...` until a page reload. Unmounting the render (switching the AuthType tab — Radix `TabsContent` has no `forceMount`, `packages/ui/src/components/tabs.tsx:44-52`) leaves the overlay on screen and drops a subsequent selection silently.
- **Question (unproven):** `useGoogleDrivePicker.ts:90,182,200` passes the account hint as `login_hint`, while the already-shipped destination picker passes `hint` (`apps/web/src/features/data-destination/shared/hooks/useGoogleDrivePicker.ts:76,147,159`). Both files hand-roll the GIS interface and there are no Google Identity Services typings installed, so `tsc` cannot catch the divergence and no test pins it — at most one of the two can be right. If the hint is ignored, GIS may mint a token for whatever account the browser is signed into, and `verifyGooglePickerAccount` (`:205-224`) then hard-fails with `Open Google Picker with the connected account …`. Which key GIS honours is **not settleable from this repository** — it needs the GIS `TokenClientConfig` reference or a two-account browser test. Which one is intended?
- `packages/connectors/src/Sources/GoogleSheets/Source.js:599-606` — **the 403 hint tells OAuth users to share the sheet with a service account.** The branch never inspects `AuthType`, and `_getConfiguredServiceAccountEmail()` returns `null` under `oauth2` (the injected items are `RefreshToken`/`AccessToken`/`ClientId`/`ClientSecret`), so the generic service-account wording fires. 403 is the *expected* failure mode here, not an exotic one: `drive.file` is per-file, so any sheet not selected through this app's Picker returns it. The message reaches the user verbatim via `connector-fields-preview.service.ts:152-154`. Suggest branching on `this.config.AuthType?.value === 'oauth2'` and pointing at the Picker instead.
- `apps/web/.../Oauth/impl/GoogleSheetsOauthRender.test.tsx:42-45` and `ConfigurationStep.test.tsx:244-247` — **both new fixtures model a `status.user` shape the backend never emits.** `Source.js:287` stores `{ id, name: <email> }` with no `email` key, and `connector.mapper.ts:119-126` copies it verbatim, so production always takes the `?? status?.user?.name` fallback at `GoogleSheetsOauthRender.tsx:37`. Deleting that fallback would keep every new test green while `connectedEmail` becomes `undefined` in production, permanently failing the `connectedEmail?.includes('@')` gate and making the Picker unreachable. Worth changing at least one fixture to the production shape — and ideally emitting `email` alongside `name` in `Source.js:287`, since the entity and mapper both already carry the field.
- `apps/web/src/features/google-oauth/hooks/useGoogleDrivePicker.ts` — **near-duplicate hook, wrong feature slice, deep import.** It is a copy of `features/data-destination/shared/hooks/useGoogleDrivePicker.ts` (same `scriptPromises` cache, same `loadScript`/`loadPicker`/`requestAccessToken`, same hand-rolled typings), placed in `features/google-oauth` — the storage/destination OAuth slice, whose `index.ts` does not export it — so the connector reaches past the barrel with an 8-level relative path (`GoogleSheetsOauthRender.tsx:6`), duplicated again in the test mock. The filename also does not match the export (`useGoogleSheetsPicker`). Suggest `features/connectors/shared/hooks/useGoogleSheetsPicker.ts` next to the existing `useOAuthPopup`/`useOAuthCallback`, with the script-loading/GIS half factored out so the two copies cannot drift again — they already have (see the hint question above).
- `apps/backend/.../connector-preview-credentials.service.ts:73-78` — **the OAuth branch skips the per-Data-Mart `EDIT` check that the `_secrets_id` branch performs, and that is fine — but it reads like an oversight.** We raised this as a blocker and then refuted it, so recording the conclusion: the preview returns column names/types only (`Source.js:662-671` → `connector-fields-schema.mapper.ts:21-36`), never sample rows; `drive.file` limits the token to files its owner picked through the app; and the same capability already exists — and always has — through the save+run path, where `_source_credential_id` is validated **nowhere** (`update-data-mart-definition.service.ts:255-289` only inspects `_secrets_id`, `connector-secret.service.ts:337-341` persists OAuth items untouched, and `connector-credential-injector.service.ts:72` scopes by `projectId` alone). Tightening only the preview branch would give false assurance. A one-line comment stating that project-level OAuth credentials are shared by design, and that preview is deliberately no more permissive than the run path, would stop the next reader from "fixing" it.

<details>

- `packages/connectors/src/Sources/GoogleSheets/Source.js:296-299` — `withExpiresIn(data.refresh_token_expires_in ?? null)` reads a field Google's token endpoint does not return, so it is effectively a hardcoded `null` (your own test asserts `expiresIn === null`). The intent is right and has precedent (`TikTokAds/Source.js:243` uses a plain `withExpiresIn(null)`), but the `??` operand is dead. Side effect worth knowing: `status.valid` is then permanently `true`, so a grant revoked at myaccount.google.com still shows `Connected as …` until a run fails.
- `Source.js:64-88` — `PickerApiKey`/`ProjectNumber` carry `OAUTH_CONSTANTS.SECRET`, which does no masking (both also carry `UI` and ship to the browser with real values via `parseOAuthVars(vars, ['UI'])`); it only makes `isOAuthEnabled` require the env var. Using `SECRET` as an availability gate on a public GCP project number will mislead the next reader — the `Scopes` precedent (`LinkedInAds/Source.js:48-54`, `FacebookMarketing/Source.js:41-47`) is `[UI]`-only. Either drop `SECRET` from `ProjectNumber` or add a comment explaining the gate.
- `packages/connectors/test/GoogleSheets.test.js:26-30` — the `OauthFlowException` stub never sets `this.name`, but the real class does (`Core/Exceptions.js` via `AbstractException`) and the backend classifies on exactly that (`connector-oauth.service.ts:98-102` → `BadRequestException`). Replacing the real throws in `Source.js` with plain `Error` would keep all three new tests green while every carefully-worded OAuth message became an opaque 500. Set `this.name` in the stub and assert it.
- `apps/backend/test/connector-oauth.e2e-spec.ts:42-48` — GoogleSheets was added to `OAUTH_CONNECTORS`, but the suite only asserts `vars` is non-empty with `string|null` values, and env is unset in tests, so nothing pins the new `PickerApiKey`/`ProjectNumber` UI vars. Dropping their `UI` attribute would keep this green while the Picker died with `Google Picker configuration is incomplete`. `expect(res.body.vars).toHaveProperty('PickerApiKey')` would close it.
- `apps/backend/test/connector-fields.e2e-spec.ts:94-117` — the new 403 case is produced by the pre-existing `connectorName` check (`connector-preview-credentials.service.ts:66-70`), which runs *before* the new oauth branch; the fixture sets no `dataMartId`/`configId`. Deleting the new `if (credential.dataMartId || credential.configId) throw` line leaves the whole suite green. A case with a matching connector name and a Data-Mart-scoped record would pin the actual new rule.
- `apps/web/.../GoogleSheetsOauthRender.tsx:21` — `pickerError` is never cleared by `handleGoogleLogin`, so `Reconnect Google Sheets before choosing a spreadsheet` stays on screen after a successful reconnect until the Picker is clicked again.
- `apps/web/.../GoogleSheetsOauthRender.tsx:34-36` — `ExternalAnchor href={spreadsheetUrl}` gets the config string with only `.trim()`. We chased this as a blocker and refuted both consequences (React 19.2.8 sanitises `javascript:` hrefs in the production bundle, and the bare-ID path is unreachable because line 27 clears the value at the moment a credential appears). One narrow path survives: if `settings.isEnabled` goes false while `_source_credential_id` survives, the input reappears, a bare ID can be saved, and the read-only link later renders `href="1AbC…"` — a relative URL. The repo's own precedent for this exact UI validates first (`drive-folder-url.utils.ts:8-14`, used by `GoogleSheetsFields.tsx:157`); `apps/web/src/features/plugins/safeHttpsUrl.ts` is the ready-made helper.
- `packages/connectors/src/Sources/GoogleSheets/Source.js:333-343` — the oauth2 branch drops the optional chaining used one line earlier and reads `authConfig.ClientId.value` / `ClientSecret.value` / `RefreshToken.value` unguarded. Reachability is low (validation rejects missing credentials first), and the `service_account` branch has the same shape, so this is a pattern rather than a regression — but a `ConnectorConfigurationException` would turn a rare opaque 500 into a readable 400.
- `.changeset/google-sheets-source-oauth.md` — filename omits the Fibery task number that every current sibling uses (`6814-…`, `6827-…`, `6857-…`, `6864-…`, `6865-…`). Honest caveat: the convention is not documented and not CI-enforced, and this feature's own predecessor (`google-sheets-source-connector.md`) also omits it. Content (`'owox': minor`) is correct.
- minor: the new flow adds a second Picker API key env var (`OAUTH_GOOGLE_SHEETS_PICKER_API_KEY`) alongside the existing `GOOGLE_PICKER_API_KEY` used by the destination flow (`google-oauth-config.service.ts:190`), and the `OAUTH_GOOGLE_SHEETS_*` prefix now spans two subsystems next to the pre-existing `OAUTH_GOOGLE_SHEETS_SCOPE` — no key collision, but worth a line in the docs · `GoogleSheetsLoginButton.tsx` is the sixth verbatim copy of the same ~130 lines including the inline Google logo SVG · `ConfigurationStep.tsx:346-355` hardcodes `'AuthType'`, `'oauth2'` and `'SpreadsheetId'` while the file already imports from `shared/utils/google-sheets-fields.utils.ts`, where every other Google Sheets field-name constant lives · `GoogleSheetsOauthRender.test.tsx:126`'s `queryByRole('textbox')` assertion passes unconditionally (the component renders no input in any branch) · the `announces an incomplete Picker configuration` test pins a state that cannot occur, since `isEnabled` is false whenever those vars are unset · `managedOAuthModes` (`ConfigurationStep.tsx:96`) is never reset on unmount and is not keyed by connector — currently inert, but correctness by coincidence · the `Map`-based dedup in `collectReferences` now allows one credential id to be validated twice under contradictory rules (unreachable today; the old `Set` collapsed it) · first paint flashes the manual OAuth fields and the `SpreadsheetId` input before `settings` resolves, since there is no loading gate · `Source.js:242-248` attaches the whole token response as `payload`, which on that branch contains a live `access_token` — latent only, as `connector-oauth.service.ts:39-42` drops `payload` before the exception leaves the service.

</details>

**What is good here:** granted-scope verification is enforced on *both* sides (`Source.js:248-262` server-side and `useGoogleDrivePicker.ts:72-82` in the browser); `verifyGooglePickerAccount` catches the "Picker opened under a different Google account" case that most implementations ship without noticing; `ClientSecret` is correctly withheld from the browser (`[SECRET, REQUIRED]` with no `UI`); the preview path mirrors the production run path exactly rather than inventing a parallel resolution; the `expiresIn` decision carries a comment explaining *why*; and `CREDENTIALS.md` both warns against reusing `OAUTH_GOOGLE_REDIRECT_URI` and explains the `drive.file` least-privilege rationale.

<sub>Automated OWOX review · worktree checkout at `4665e16` · 5 parallel analysis passes + 3 adversarial skeptics · 3 blocker candidates → 1 confirmed, 2 refuted · nothing was executed locally, CI parity read from `gh pr checks` (20/20 green).</sub>

zapolsky (approved):

PR #1564, opened 2026-08-24
feat: set the relationship description per join with inherited fallback

laskevych:
## What

A relationship's description used to be one text inherited read-only by every join that reuses the relationship. But the same join can mean something different depending on the path it is reached through: `A → B` carries one business meaning on its own and another inside `C → D → A → B`.

This PR makes the description resolvable per join node, the same way per-node field aliases already work:

- **Storage** — an optional `description` override on `BlendedSource` inside the existing `blendedFieldsConfig` JSON (keyed by `aliasPath`), so no migration is needed and the override survives alias renames via the existing path cascade.
- **Resolution (backend)** — `BlendableSchemaService` publishes an effective `joinDescription` on each `availableSources[]` entry: the per-join override when set, otherwise the relationship-level description.
- **Web (Data Setup)** — the Description tab of an inherited join becomes editable: the inherited text shows as the placeholder, typing stores an override for that join only, and "Reset to inherited" (or clearing the field) falls back to the relationship's text. Direct joins keep editing the relationship description itself.
- **MCP** — `get_data_mart_details_by_id` join edges now carry the effective per-node description instead of re-reading the relationship's own text.
- **Report column picker** — the join-path tooltip shows the effective description under the path tree.

## Tests

- Backend: effective-resolution cases in `blendable-schema.service.spec.ts` (override wins / inherited / neither), MCP facade forwards the resolved value (not the relationship text), DTO validation for the new field (`min 1` / `max 10000`, empty override rejected). Suites: 74 + 18 + 33 + 7 passed.
- Web: override mode of `JoinDescriptionForm` (placeholder, debounce commit, blur/unmount flush, reset button), join description in the picker tooltip. Suites: 86 + 134 passed.
- `tsc` clean for the web app; backend error profile identical to `main` (pre-existing spec-only errors).


apps/web/src/features/data-marts/edit/components/DataMartRelationships/JoinDescriptionForm.tsx
```diff
@@ -18,16 +25,206 @@ interface JoinDescriptionFormProps {
...
+        </TooltipContent>
+      </Tooltip>
+    </label>
+  );
+}
+
+/**
+ * Editor for the per-join description override. The commit is a synchronous callback into the
+ * blended-fields-config save path (optimistic local state, whole-config PUT), so unlike the
+ * relationship PATCH below it needs no in-flight serialization — only debounce and flushes.
```
kalnik-a-a, line 65: Criticality: MEDIUM / NON-BLOCKING

`override.onChange()` eventually starts an asynchronous whole-config PUT, but this editor clears its dirty state and permits another debounce, blur, or reset commit without waiting for the first request. On a slow connection, two PUTs can complete out of order, so the older description—or an older full config—can win; a rejected request also leaves the optimistic value looking saved.

This is non-blocking for this PR because it is an uncommon race in descriptive metadata editing. Please follow up by serializing or coalescing saves, applying only the latest successful response, and covering reversed completion order and rejection in tests.

apps/backend/src/data-marts/dto/schemas/blended-fields-config.schema.ts
```diff
@@ -27,6 +27,9 @@ export const BlendedSourceSchema = z.object({
     ),
   alias: z.string().min(1).max(255),
   isExcluded: z.boolean().optional(),
+  // Per-join override of the relationship description. Absent → the join inherits the
+  // relationship-level description; an empty override is stored as an absent key, not ''.
+  description: z.string().min(1).max(10000).optional(),
```
kalnik-a-a: Criticality: LOW / NON-BLOCKING

`min(1)` still accepts whitespace-only strings such as `"   "`. A direct API client can persist a present override that suppresses the inherited description via `??`, while the web UI treats the same input as clearing the override.

Please trim or normalize at the backend boundary (or reject whitespace-only input) and add a small contract test. This does not need to block this PR.

kalnik-a-a (approved):
Review outcome: APPROVED

Criticality: two NON-BLOCKING follow-ups (one MEDIUM, one LOW).

I left inline comments for serializing or coalescing per-join description config saves and for normalizing whitespace-only description overrides. The implementation is otherwise acceptable at head `7beda365224bdb442701a63810045b5c0599b909`. Required CI still needs a fresh green run on the current integration before merge.

PR #1565, opened 2026-08-24
feat (web): Data Marts canvas and nav polish

artemfenoga:
# Model Canvas and Data Mart navigation polish

The Model Canvas page (formerly "Models") now opens straight into your data model: with a single storage, it's selected automatically instead of asking you to pick one first. When you do need to switch, the storage picker moved up into the page header as a clear "Model for [Storage]" title, and a "Data Marts" breadcrumb takes you back to the list in one click — the same breadcrumb now appears on the Reports, Triggers, Run History, and Insights pages too.

On the canvas itself, each Data Mart card now shows a full-height color stripe reflecting its Data Quality status, so you can spot cards with warnings or errors at a glance instead of opening each one. Downloading the canvas (as an image, JSON, or OKF Markdown) is now a dedicated button in the toolbar instead of being tucked inside the Actions menu, and bulk actions like Publish and Delete now state exactly how many Data Marts they'll affect.

## How it started

Frame 23137625

## How it's going

Frame 23137626

## What's changed

- Updated ModelCanvasView to improve storage handling and UI structure.
- Removed unused DataStorageProvider and adjusted component exports.
- Enhanced data quality status handling with new stripe classes.
- Consolidated DataMartBulkActions to remove export submenu and adjusted tests accordingly.
- Introduced ProjectDataMartSectionHeader for consistent page headers across Data Mart pages.
- Improved accessibility and usability in ProjectDataMartTitleLink with tooltips.
- Refined Combobox component to allow custom rendering based on context.
- Updated global CSS for improved success color definitions.


apps/web/src/features/data-marts/shared/components/DataMartBulkActions/DataMartBulkActions.tsx
```diff
@@ -186,24 +154,17 @@ export function DataMartBulkActions({
         <DropdownMenuTrigger asChild>
           <Button
             variant='outline'
-            size='sm'
+            data-testid='data-mart-bulk-actions-trigger'
```
romandubovyi, line 158: 🔴 **Blocking:** Removing the visible "Actions" label breaks the existing E2E test `e2e/specs/datamart-list.spec.ts:73`, which locates this trigger with `getByRole('button', { name: /Actions/ })`. The button's accessible name now comes only from `title='Bulk actions for selected data marts'` — lowercase "actions" — so the case-sensitive regex no longer matches and the click times out.

`Browser E2E Tests` fails on this head (`DLIST-DQ`, initial run and retry) while passing on `main`. Adding a short `aria-label` (e.g. `aria-label='Actions'`) fixes both this and the naming inconsistency with the icon-only menu triggers elsewhere in this codebase, which all use `aria-label='Open menu'`.

apps/web/src/features/data-marts/shared/components/DataMartBulkActions/DataMartBulkActions.tsx
```diff
@@ -212,7 +173,7 @@ export function DataMartBulkActions({
             }}
           >
             <CircleCheckBig aria-hidden='true' />
-            Publish
+            Publish {dataMarts.length} Data Mart{dataMarts.length === 1 ? '' : 's'}
```
romandubovyi: 🔴 **Blocking:** This label counts every selected Data Mart, but `handleBatchPublish` iterates `draftDataMarts` only. With 5 selected of which 2 are drafts, the menu reads "Publish 5 Data Marts" while the confirmation that opens immediately after says "You're about to publish all 2 draft data marts" — contradicting the description's claim that these labels state exactly how many Data Marts they'll affect. `Delete` is correct because `handleBatchDelete` really does iterate all of them.

Please use the draft count here (e.g. "Publish N draft Data Marts"). Note `DataMartTable.test.tsx:86` currently locks the wrong behaviour in: it asserts `'Publish 1 Data Mart'` for a *Published* mart, so the item is disabled and still claims it will publish one.

🤖 Generated with Kimi CLI + Grok CLI (Grok 4.6)

apps/web/src/features/data-marts/model-canvas/components/ModelCanvasFlowNode.tsx
```diff
@@ -105,6 +117,13 @@ export default function ModelCanvasFlowNode({
         transition: 'opacity 0.2s, filter 0.2s',
       }}
     >
+      {withSource && (
```
romandubovyi: ⚪ **Non-blocking:** The stripe now encodes Data Quality status, but it is still gated on `withSource`, so turning off the "Input source" object label also hides the quality cue this PR added.

Two follow-ups: the comment at line 76 above still says the stripe and the source badge "encode the same definition type", which is no longer true; and `canvas-settings-panel.tsx:22` — not touched by this PR — still tells the user that toggle controls "The source badge (VIEW / TABLE / SQL / PATTERN / CONNECTOR) and its accent stripe". Consider gating the stripe on the status/quality label instead (or giving it its own toggle) and updating that helper text.

🤖 Generated with Grok CLI (Grok 4.6) + Claude Code (Opus 5 (1M context))

apps/web/src/features/data-marts/model-canvas/model/use-model-canvas-filters.ts
```diff
@@ -24,8 +24,8 @@ export function useModelCanvasFilters() {
   }, [storageValue, setStorageParam, storageLsKey]);
 
   const statusFilter: CanvasStatusFilter =
-    status.value === 'draft' ? 'draft' : status.value === 'all' ? 'all' : 'published';
-  const relFilter: CanvasRelFilter = rel.value === 'all' ? 'all' : 'connected';
+    status.value === 'draft' ? 'draft' : status.value === 'published' ? 'published' : 'all';
```
romandubovyi, line 27: ⚪ **Non-blocking:** The omitted-parameter defaults flipped here: no `status` used to mean `published` and no `rel` used to mean `connected`; both now mean `all`, and `removeParam` fires for `all` instead of the old defaults. Anyone opening the canvas without query parameters will now see drafts and unconnected Data Marts by default.

The reordered `<Select>` options suggest this is intentional, but it is a user-visible behaviour change that neither the description nor the changeset mentions — worth adding a line to the changeset.

🤖 Generated with Grok CLI (Grok 4.6) + Kimi CLI + Claude Code (Opus 5 (1M context))

apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
```diff
@@ -98,6 +95,11 @@ function ModelCanvasViewContent({ onActiveQualityRunChange }: ModelCanvasViewPro
     };
   }, [loadDataStorages]);
 
+  useEffect(() => {
```
romandubovyi, line 98: ⚪ **Non-blocking:** This gates on `filters.storageId` rather than on whether that storage actually exists. If localStorage holds an id for a since-deleted storage, `useModelCanvasFilters` restores it into the URL param, so `filters.storageId` is truthy and this effect bails out even when `dataStorages.length === 1`. `storageKnown` is then false, so the user lands on "Select a storage to view its data model" with the toolbar hidden — the auto-select feature silently not applying in exactly the single-storage case it targets. Gating on `storageKnown` instead fixes it.

Separately, the dependency array takes the whole `filters` object, which `useModelCanvasFilters` rebuilds every render, so this effect re-runs on every render; narrowing the deps would not help because `setStorageId` is also recreated each render — it would need a `useCallback` in the hook.

apps/web/src/pages/data-marts/model-canvas/ModelCanvasPage.tsx
```diff
@@ -1,17 +1,41 @@
...
         <div className='flex flex-wrap items-center justify-between gap-4'>
-          <h1 className='dm-page-header-title'>Models</h1>
+          <div className='dm-page-header-title flex items-center gap-1.5'>
+            <Link
+              to={scope('/data-marts')}
+              className='text-muted-foreground/75 hover:text-foreground transition-colors'
+            >
+              Data Marts
+            </Link>
+            <ChevronRight className='text-muted-foreground mt-1 size-4' aria-hidden='true' />
```
romandubovyi: ⚪ **Non-blocking:** This PR adds `ProjectDataMartSectionHeader` to standardise the "Data Marts ›" breadcrumb and uses it on Reports, Triggers, Run History and Insights — but the same `Link` + `ChevronRight` markup is duplicated verbatim here rather than reusing it, presumably because the component only accepts `title: string` and this page needs to slot in `ModelCanvasStorageSelect`. Adding a `children`/`titleSlot` prop would let the canvas page use it too, so breadcrumb styling stays in one place.

Minor: lines 12-13 call `useProjectRoute()` twice — `navigate` and `scope` can come from a single call.

romandubovyi (changes requested):
Multi-agent review of the Model Canvas and navigation polish. Two blocking points: the bulk-actions trigger loses the accessible name the Data Mart list E2E test matches on (`DLIST-DQ` is currently red on this head and green on `main`), and the Publish menu item counts every selected Data Mart while only drafts are published. The rest are non-blocking.

🤖 Generated with Kimi CLI + Grok CLI (Grok 4.6) + Claude Code (Opus 5 (1M context))

apps/web/src/features/data-marts/shared/components/DataMartBulkActions/DataMartBulkActions.tsx
artemfenoga: Fixed. Added aria-label='Open menu' to match the existing icon-only menu triggers, and switched the E2E test to getByTestId(TESTIDS.dataMartBulkActionsTrigger) instead of the accessible-name regex so it's not coupled to button text/title. Thanks for catching this.

apps/web/src/features/data-marts/shared/components/DataMartBulkActions/DataMartBulkActions.tsx
artemfenoga: Good catch — fixed. The Publish label now reads the draft count (draftCount), same set handleBatchPublish actually iterates, worded to match the confirmation dialog ("Publish N draft Data Mart(s)"). Updated DataMartTable.test.tsx:84 to expect "Publish 0 draft Data Marts" for the published-mart case, since that item is disabled and none would actually publish.

apps/web/src/features/data-marts/model-canvas/components/ModelCanvasFlowNode.tsx
artemfenoga: Good catch. The stripe is now gated on !titleOnly instead of withSource, so it hides only together with the quality-icon row (all three object labels off), not when the source badge alone is toggled. Updated the stale comment claiming the stripe and source badge "encode the same definition type," and trimmed the outdated "and its accent stripe" clause from the Input source helper text in canvas-settings-panel.tsx.

apps/web/src/features/data-marts/model-canvas/model/use-model-canvas-filters.ts
artemfenoga: Confirmed intentional — matches the reordered <Select> options. Added a line to the changeset: "Opening the canvas with no filters applied now shows all Data Marts, published or draft, connected or not — previously it defaulted to published-only, connected-only." Thanks for flagging that it wasn't documented.

apps/web/src/features/data-marts/model-canvas/components/ModelCanvasView.tsx
artemfenoga: Good catch on both. Gating switched from filters.storageId to storageKnown, so a stale/deleted storage id in localStorage no longer blocks the single-storage auto-select. And per your note, I stabilized setStorageId/setStatus/setRel/setSearchQuery in useModelCanvasFilters with useCallback (the underlying setParam/removeParam from useUrlParam were already stable) so the effect's deps could actually be narrowed to [storageKnown, dataStorages, filters.setStorageId] instead of the whole filters object.

apps/web/src/pages/data-marts/model-canvas/ModelCanvasPage.tsx
artemfenoga: Good suggestion. Added titleSlot/actions props to ProjectDataMartSectionHeader so the breadcrumb markup lives in one place; Models page now reuses it, slotting in the storage picker via titleSlot and RunActivityIndicator via actions. Also collapsed the duplicate useProjectRoute() call.

packages/ui/src/styles/globals.css
```diff
@@ -26,8 +26,8 @@
   --destructive: oklch(0.577 0.245 27.325);
   --destructive-foreground: oklch(0.577 0.245 27.325);
   --destructive-bg: oklch(0.971 0.013 17.38);
-  --success: oklch(0.627 0.194 149.21);
-  --success-bg: oklch(0.982 0.018 155.83);
+  --success: oklch(0.7157 0.1515 140.73);
```
romandubovyi: ⚪ **Non-blocking:** This lightens the global success token from `oklch(0.627 0.194 149.21)` (#00a63e, 3.22:1 on white) to `oklch(0.7157 0.1515 140.73)` (#6aba5c, **2.39:1**) — below the 3:1 WCAG 1.4.11 floor for non-text graphics, and well below 4.5:1 for text.

The blast radius is wider than the canvas: `text-success` labels the "Passed" state in `DataQualitySummaryPanel.tsx:33` and `DataQualityResultCard.tsx:278`, and `bg-success` is exactly the new full-height Data Quality stripe this PR introduces (`data-quality-status.ts:67`) — the cue is meant to let users spot healthy vs. problematic cards at a glance, and at 2.39:1 the green barely separates from the card background. The repo already enforces this rule explicitly: `shared/canvas/constants.ts:4-6` rejects `gray[3]` for edges because "1.73:1 on white … fails WCAG 1.4.11".

Also note dark-mode `--success` (line 100) stays at hue 151.71 while light mode moves to 140.73, so the two themes now disagree on the brand green.

Suggestion: keep the light token at or below ~L 0.63, or introduce a separate lighter `--success-accent` for fills only and leave `--success` for text and icons.

apps/web/src/features/data-marts/shared/canvas/constants.ts
```diff
@@ -15,6 +15,8 @@ export const EDGE_STROKE_WIDTH = 1.5;
 export const EDGE_SELECTED_STROKE_WIDTH = 2.5;
 export const EDGE_WARNING_DASH = '8 4';
 export const DIMMED_OPACITY = 0.15;
+/** MiniMap node fill — flat, theme-aware; no per-node color coding. */
+export const MINIMAP_NODE_COLOR = ' var(--muted)';
```
romandubovyi: ⚪ **Non-blocking:** `--muted` is `oklch(0.97 0 0)` (≈ #f5f5f5) in light mode, and @xyflow's MiniMap background defaults to `#fff` (`--xy-minimap-background-color-default`). That leaves MiniMap nodes at roughly **1.09:1** against their own background — they read as blank in light theme, where they used to be colour-coded per definition type. Dark mode is fine (#444 on #141414).

`var(--muted-foreground)` would keep the "flat, theme-aware, no per-node colour coding" intent while staying visible — it is what this same PR picks for the neutral stripe in `RelationshipCanvas.tsx:282`.

Minor: the string has a leading space (`' var(--muted)'`).

apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx
```diff
@@ -281,11 +279,7 @@ export function RelationshipFlowNode({ id, data, selected }: NodeProps<Relations
       >
         <IndicatorLabel data={data} />
         {withSource && (
-          <span
-            className='h-4 w-1 shrink-0 rounded-sm'
-            style={{ background: accent }}
-            aria-hidden='true'
-          />
+          <span className='bg-muted-foreground/50 h-4 w-1 shrink-0 rounded-sm' aria-hidden='true' />
```
romandubovyi: ⚪ **Non-blocking:** With `definitionTypeAccent` gone, this stripe (and its twin at line 317) is now a constant `bg-muted-foreground/50` bar that encodes nothing, yet it is still shown/hidden by the `withSource` object label.

`canvas-settings-panel.tsx` is shared by both canvases, and this PR narrowed its helper text to "The source badge (VIEW / TABLE / SQL / PATTERN / CONNECTOR)" — accurate for the Models canvas, but on Joinable Data Marts that toggle still controls a decorative gray bar the helper no longer mentions. It also now disagrees with the Models canvas, where the stripe was moved off `withSource` onto `!titleOnly`.

Since it carries no information any more, dropping both `withSource &&` stripe blocks would keep the toggle's meaning honest and shrink the diff. The "mirrors the Models canvas: accent stripe + source badge" comment above line 317 is stale either way.

🤖 Generated with Claude Code (Opus 5 (1M context)) + Grok CLI (Grok 4.5)

apps/web/src/features/data-marts/shared/canvas/canvas-settings-panel.tsx
```diff
@@ -19,7 +19,7 @@ import { VIEW_MODE_OPTIONS, type CanvasViewMode } from './view-mode';
 const OBJECT_LABEL_META: Record<ObjectLabelPart, { label: string; helper: string }> = {
   source: {
     label: 'Input source',
-    helper: 'The source badge (VIEW / TABLE / SQL / PATTERN / CONNECTOR) and its accent stripe',
+    helper: 'The source badge (VIEW / TABLE / SQL / PATTERN / CONNECTOR)',
```
romandubovyi, line 22: ⚪ **Non-blocking:** *(the point is the `status` entry a few lines below, at lines 29-30 — outside this diff hunk)*

The `source` helper was corrected here, but the `status` entry right below still reads `label: 'Status dot'` / `helper: 'The published/draft status dot in the card header'`. Models cards now show Draft/Published as an `ErdStatusBadge` pill in the meta row — the header dot is gone from `ModelCanvasFlowNode` entirely. Relationship canvas still uses the dot, so this panel is shared between two different renderings.

Worth rephrasing to something true for both, e.g. `label: 'Status'` / `helper: 'The published/draft indicator'`.

apps/web/src/pages/data-marts/model-canvas/ModelCanvasPage.tsx
```diff
@@ -1,29 +1,55 @@
...
+import { DataStorageProvider } from '../../../features/data-storage/shared/model/context';
+import { useDataStorage } from '../../../features/data-storage/shared/model/hooks/useDataStorage';
 import { useProjectRoute } from '../../../shared/hooks';
+import { ProjectDataMartSectionHeader } from '../shared/ProjectDataMartSectionHeader';
 
-export default function ModelCanvasPage() {
+function ModelCanvasPageContent() {
   const [hasActiveQualityRun, setHasActiveQualityRun] = useState(false);
   const { navigate } = useProjectRoute();
+  const { dataStorages } = useDataStorage();
```
romandubovyi, line 14: ⚪ **Non-blocking:** `DataStorageProvider` is a bare `useReducer` — it never fetches. The only `fetchDataStorages()` call is `loadDataStorages` inside `ModelCanvasView`, so this header picker is populated purely as a side effect of a descendant loading data for itself (before this PR both lived in the same component).

It works today, but the coupling is invisible: if `ModelCanvasView` is ever lazily loaded or short-circuited, this picker silently sits on "Select storage" forever, with no loading or error state of its own. Lifting the fetch (and its `loading` / `error`) into `ModelCanvasPageContent`, or into the provider itself, would make the header self-sufficient.

apps/web/src/pages/data-marts/shared/ProjectDataMartSectionHeader.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <div className='flex flex-wrap items-center justify-between gap-4'>
+        <div className='dm-page-header-title flex items-center gap-1.5'>
+          <Link
+            to={scope('/data-marts')}
+            className='text-muted-foreground/75 hover:text-foreground transition-colors'
+          >
+            Data Marts
+          </Link>
+          <ChevronRight className='text-muted-foreground mt-1 size-4' aria-hidden='true' />
+          {titleSlot ?? <h1 className='inline'>{title}</h1>}
```
romandubovyi: ⚪ **Non-blocking:** `title` is a required prop but is completely ignored once `titleSlot` is supplied. `ModelCanvasPage.tsx:20,23` therefore has to spell "Model for" twice — once as the dead `title`, once inside the slot — and the two can drift apart on the next edit.

Either make the props a discriminated union (`{ title: string } | { titleSlot: ReactNode }`), or keep `title` mandatory and add a narrower `titleAfter?: ReactNode` slot rendered after the `<h1>`, so the component always owns the heading element.

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
```diff
@@ -358,6 +346,19 @@ export default function DataMartReportsPage() {
           return <ProjectReportTitleCell report={row.original} />;
         },
       },
+      {
+        id: 'dataMart',
```
romandubovyi, line 350: ⚪ **Non-blocking:** This moves the `dataMart` column from first to second, behind `title` (the test rename at `DataMartReportsPage.test.tsx:199` confirms it is deliberate). That is a user-visible table change, but neither the PR description's "What's changed" list nor `.changeset/6835-datamarts-canvas-and-navigation-polish.md` mentions it — the changeset only covers the truncation change.

Worth one line in the changeset so the release notes match what ships.

romandubovyi (approved):
Approving — all findings from this round are non-blocking.

Multi-agent review pass over `a294165f7` (Claude Code + Grok CLI; Kimi CLI hit a provider quota, Antigravity CLI timed out twice). Every point raised in my previous round was verified as actually fixed in the code, not just claimed:

- bulk-actions trigger — `aria-label='Open menu'` added and the E2E selector switched to `TESTIDS.dataMartBulkActionsTrigger`;
- Publish label — now reads `draftCount`, the same set `handleBatchPublish` iterates;
- quality stripe — regated on `!titleOnly` instead of `withSource`;
- storage auto-select — gated on `storageKnown`, so a stale localStorage id no longer blocks it;
- breadcrumb — `ProjectDataMartSectionHeader` now reused by the Models page via `titleSlot`/`actions`;
- filter defaults — confirmed intentional and documented in the changeset.

Affected suites pass locally: 25 files / 197 tests (`vitest run --maxWorkers=3` over `model-canvas`, `DataMartBulkActions`, `DataMartTable`, `pages/data-marts`).

7 non-blocking notes left inline. The one worth a look before merge is the global `--success` token: it drops from 3.22:1 to 2.39:1 on white, under the 3:1 WCAG 1.4.11 floor, and it is the same token the new Data Quality stripe paints with.

🤖 Generated with Claude Code (Opus 5 (1M context)) + Grok CLI (Grok 4.5)

packages/ui/src/styles/globals.css
artemfenoga: Good catch, thanks — confirmed the numbers: #00a63e was 3.22:1, #6aba5c was 2.39:1.

Went with your second option: reverted --success back to oklch(0.627 0.194 149.21) (3.22:1, text/icons — text-success in the summary panel and result card), and added a separate --success-accent: oklch(0.7157 0.1515 140.73) used only by the canvas stripe (bg-success-accent in data-quality-status.ts). Also fixes the light/dark hue mismatch you flagged — light --success is now 149.21, close to dark's 151.71.

apps/web/src/features/data-marts/shared/canvas/constants.ts
artemfenoga: Good catch — --muted at 0.97 lightness is basically invisible on the minimap's white background (~1.09:1). Switched to var(--muted-foreground) (5.64:1), same token already used for the neutral stripe in RelationshipCanvas.tsx:282, and fixed the leading-space typo.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/RelationshipCanvas.tsx
artemfenoga: Good catch, agreed — dropped both withSource && <span className='bg-muted-foreground/50 .../> blocks (source card + ERD card). withSource still gates the real ErdDefinitionBadge (line 357), so the toggle's meaning now matches the settings-panel helper text again. Also cleaned up the two stale "mirrors the Models canvas: accent stripe" comments.

apps/web/src/features/data-marts/shared/canvas/canvas-settings-panel.tsx
artemfenoga: Good catch — confirmed, ModelCanvasFlowNode now renders ErdStatusBadge (pill in the meta row), the header dot is gone from Models cards entirely, only RelationshipCanvas still uses ErdStatusDot. Applied your suggested wording: label: 'Status' / helper: 'The published/draft indicator'.

apps/web/src/pages/data-marts/model-canvas/ModelCanvasPage.tsx
artemfenoga: Fair point, and accurate — DataStorageProvider is indeed just a bare useReducer, fetchDataStorages today only runs inside ModelCanvasView's race-safe effect (storageLoadGenerationRef/mountedRef), and the header picker rides on that as a side effect.

Leaving as-is for now: fixing this properly means either duplicating ModelCanvasView's fetch/loading/error handling in the page component, or moving the fetch into the provider — which is shared by 4 other call sites (DataMartCreateForm, BulkCreateFromStorageDialog, DataStorageListPage, CreateDataMartPage) and would need checking each doesn't double-fetch. Not worth the blast radius for a non-blocking nit right now; agree it's real coupling and worth a follow-up if ModelCanvasView ever gets lazy-loaded or conditionally rendered.

apps/web/src/pages/data-marts/shared/ProjectDataMartSectionHeader.tsx
artemfenoga: Good catch — "Model for" was duplicated exactly as you described. Went with your titleAfter suggestion since titleSlot is only used by this one page (ModelCanvasPage); the other 4 (Insights/Reports/Runs/Triggers) just pass title. ProjectDataMartSectionHeader now always owns the  h1 , titleAfter renders right after it, and ModelCanvasPage passes only the storage picker — no more repeated "Model for" string.

apps/web/src/pages/data-marts/reports/DataMartReportsPage.tsx
artemfenoga: Good catch, missed that. Added a line to the changeset: "In the Reports table, the Report column now leads, with Data Mart moved to second place."

PR #1568, opened 2026-08-25
feat(plugin-host): surface release rejections and gate collection changes on SemVer

laskevych:
## Problem

When a plugin's GitHub Release is rejected during sync, the reason is communicated nowhere: not in application logs, not in the audit trail, not in the UI. From the outside it looks like "Check now did nothing" — the plugin stays on an old version with no explanation, while the reason sits only in `plugin.lastSyncReport` in the database.

Separately, the collection-compatibility safeguard was temporarily disabled in #1541, with the plan to bring it back gated on SemVer.

## What changed

**Backend**
- Publisher-fixable rejections are logged as warnings from `RunPluginUpdateCheckService.run()`, covering both the daily cron and manual **Check now** in one place: `owner/plugin: v0.1.2 rejected (COLLECTIONS_INCOMPATIBLE) — Collection "dashboards" cannot change entity binding`. Rejections that are permanent by design (`DRAFT`, `PRERELEASE_FLAG`, `PRERELEASE_TAG`, `BUILD_METADATA_TAG`, `INVALID_TAG`) stay silent so a repository full of drafts cannot bury the signal. A throttled check re-serves a stored report and is not re-logged.
- The same actionable subset rides on the `UPDATE_CHECK` audit record's `afterState`, which previously carried detail only for thrown transport errors.
- The collection-compatibility gate is restored, now scoped by SemVer: a minor or patch release cannot remove, rescope or rebind a released collection; a major version bump waives the check, so a breaking collection change ships by declaring itself as one. The rejection detail names the way out ("publish a new major version").
- `COLLECTIONS_INCOMPATIBLE` gets its missing doc comment; the `ponytail:` markers are gone.

**Web**
- New **Release issues** card on the plugin details page renders `publisherDiagnostics.rejections` (tag, code, reason, time of the check). It is built from management-only diagnostics, so non-publishers never see it. **Check now** invalidates publications so the card refreshes with the new report.

**Docs**
- `docs/plugins/authoring-guide.md` and `packages/plugin-sdk/README.md` now state the SemVer-scoped collection rule (the SDK README previously did not document it at all).

## Tests

- `sync-plugin-releases.service.spec.ts`: the #1541-inverted test is reworked — a patch release that rebinds a collection is rejected with `COLLECTIONS_INCOMPATIBLE`; a new test verifies a major release with the same change is accepted.
- `run-plugin-update-check.service.spec.ts`: warn logging for actionable rejections, silence for by-design codes, audit `afterState.rejections`, and no re-logging from a stale throttled report.
- Web: unit tests for the rejection filter and page-level tests for the card (shown to a publisher with rejections, absent for by-design-only rejections and for non-publishers).


apps/backend/src/plugin-host/use-cases/sync-plugin-releases.service.ts
```diff
@@ -211,11 +214,35 @@ export class SyncPluginReleasesService {
...
+    const currentVersion = (await this.versionService.findAllByPluginId(pluginId)).reduce<
+      { semver: string; collections: PluginVersionCollections } | undefined
+    >(
+      (highest, version) =>
+        !highest || compareSemver(version.semver, highest.semver) > 0
+          ? { semver: version.semver, collections: version.collections ?? [] }
+          : highest,
+      undefined
+    );
+    if (currentVersion && majorOf(semver) <= majorOf(currentVersion.semver)) {
```
romandubovyi: ⚪ **Non-blocking:** This compares the candidate against the globally highest stored version (`majorOf(semver) <= majorOf(currentVersion.semver)`), so a 1.x release is checked against 2.x collections whenever a higher major already exists and did not settle this run — for example a recreated Release rejected with `VERSION_CONFLICT`, or a deleted one that never reaches the candidate list. `falls through to the next release down when the newest is rejected` shows that path is reachable. That does not match the "within the released major line" rule stated in the comment just above. Baseline the highest version with the same major, and skip the check when that line has no prior version.

apps/web/src/features/plugins/hooks/usePlugins.ts
```diff
@@ -114,7 +116,13 @@ export function usePluginActions() {
 
   const updateMutation = useMutation({
     mutationFn: (pluginId: string) => pluginsService.checkNow(pluginId),
-    onSuccess: invalidate,
+    // A check rewrites the sync report that the publisher's Release issues card reads
+    // from, so publications refresh along with the member-facing lists.
+    onSuccess: () =>
+      Promise.all([
+        invalidate(),
+        queryClient.invalidateQueries({ queryKey: [PUBLICATIONS_KEY, projectId] }),
```
romandubovyi, line 125: ⚪ **Non-blocking:** The new `PUBLICATIONS_KEY` invalidation is what refreshes the Release issues card after Check now, but `usePlugins.test.tsx` still only asserts toasts across its eight `checkNow` cases. A regression that drops this `invalidateQueries` call would restore the original "Check now did nothing" UI with nothing failing in CI. Since the spec for this exact hook already exists, please add a case there asserting the publications query is invalidated on a successful check.

apps/web/src/features/plugins/hooks/usePlugins.ts
```diff
@@ -114,7 +116,13 @@ export function usePluginActions() {
 
   const updateMutation = useMutation({
     mutationFn: (pluginId: string) => pluginsService.checkNow(pluginId),
-    onSuccess: invalidate,
+    // A check rewrites the sync report that the publisher's Release issues card reads
+    // from, so publications refresh along with the member-facing lists.
+    onSuccess: () =>
```
romandubovyi, line 122: ⚪ **Non-blocking:** The invalidation refreshes the card, but the immediate feedback is unchanged: `reportOutcome` still shows "You're up to date — v1.0.0" for `up_to_date`, including when this very check rejected a newer release with `COLLECTIONS_INCOMPATIBLE`. For a publisher that is still the "Check now did nothing" experience the PR description opens with — the card sits further down the page and can be collapsed. The data is already on the wire: `toUpdateResultResponse()` returns `diagnostics` (with `rejections`) to a publisher and `null` to an ordinary member; only the web `PluginUpdateResult` type omits the field. Worth pointing the toast at the rejections rather than asserting everything is fine.

apps/web/src/features/plugins/components/PluginReleaseIssuesCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+            ))}
+          </ul>
+        </div>
+      </CollapsibleCardContent>
+
+      <CollapsibleCardFooter
+        left={
+          issues.syncedAt ? (
+            <span className='text-muted-foreground text-sm'>
+              As of the check on {formatDateOnly(issues.syncedAt)}
```
romandubovyi: ⚪ **Non-blocking:** `formatDateOnly` drops the time, but the PR description promises the time of the check. With a daily cron behind it, this stamp only changes once a day, so a publisher who just fixed a release and pressed Check now cannot tell from the card whether it reflects the fresh check. `formatDateShort` in the same module already renders date + time ("Aug 17, 2026, 02:30 PM") and is the existing choice for this kind of stamp.

apps/backend/src/plugin-host/enums/release-rejection-code.enum.ts
```diff
@@ -28,3 +33,27 @@ export enum ReleaseRejectionCode {
...
+ * The rejections a publisher can and should act on: a release wanted to become current
+ * and could not.
+ *
+ * The candidacy codes (DRAFT, PRERELEASE_FLAG, PRERELEASE_TAG, BUILD_METADATA_TAG,
+ * INVALID_TAG) are deliberately absent. They mark releases that are ineligible by design
+ * and permanently -- a repository full of drafts and release candidates would repeat
+ * them on every daily check while nothing is actually wrong, burying the one line that
+ * matters.
+ */
+export const PUBLISHER_ACTIONABLE_REJECTION_CODES: ReadonlySet<ReleaseRejectionCode> = new Set([
```
romandubovyi: ⚪ **Non-blocking:** This is a hand-maintained allowlist, so a rejection code added to the enum later silently disappears from both the log line and `afterState.rejections` — the exact silence this PR exists to remove. The web side (`rejections.ts`) deliberately went the other way with a blocklist so an unknown code is shown, and documents that as intentional; the two halves therefore degrade in opposite directions on the next enum member. A `Record<ReleaseRejectionCode, boolean>` (or `satisfies` over the full enum) would make the compiler force a decision for every new code, at no runtime cost.

apps/backend/src/plugin-host/use-cases/sync-plugin-releases.service.ts
romandubovyi: ⚪ **Non-blocking:** `majorOf()` returns 0 for every `0.y.z`, so for a pre-1.0 plugin this condition is always true and the waiver is unreachable. The case this PR's own spec encodes at line 400 — 0.1.0 → 0.1.1 rebinding `dashboards`, i.e. the incident #1541 hotfixed — is rejected again, and the only escape for a plugin still in initial development is to jump straight to 1.0.0, while SemVer §4 treats `0.y.z` as exactly the range where anything may change. If that is intended, the authoring guide is worth a sentence: it currently says only that a major bump waives the check, without spelling out that for a 0.x plugin this means leaving 0.x entirely. Otherwise, treating a minor bump inside `0.y` as the declared break (compare minor as well when `major === 0`) would keep the escape hatch proportionate.

romandubovyi (approved):
Review complete — approving: nothing found rises to blocking.

The restored gate is strictly more protective than current `main`, where #1541 disabled it entirely, and the diagnostics stay on management-only responses. The changeset follows policy (`6839-`, `owox: minor`).

Verified on `15e42eb11`: `apps/backend` plugin-host update-check and sync suites pass (44/44), and the affected `apps/web` suites pass (33/33).

Six non-blocking points below — the two most substantive are the collection gate baselining the globally highest version rather than the same major line, and the new `PUBLICATIONS_KEY` invalidation shipping without a test in an existing hook spec.

🤖 Generated with Grok CLI (Grok 4.6) + Claude Code (Opus 5)

apps/backend/src/plugin-host/use-cases/sync-plugin-releases.service.ts
laskevych: Fixed in 7bb1695f4: the baseline is now the highest version in the candidate's own compatibility line (`sameCompatibilityLine` in semver.util), and the check is skipped when that line has no prior version. Added a spec where a 1.1.0 release is checked against 1.0.0 rather than an existing 2.0.0.

apps/web/src/features/plugins/hooks/usePlugins.ts
laskevych: Added in 7bb1695f4: 'refreshes publications after a check' asserts the invalidateQueries call with the publications key.

apps/web/src/features/plugins/hooks/usePlugins.ts
laskevych: Done in 7bb1695f4. One correction to the premise: the update response gated diagnostics on `isDeploymentPublisher`, so a browser member-scope publisher received null — the toast would never fire for them. Extended the gate to anyone managing a publication of this plugin (GET /publications already serves them the same block), added `diagnostics` to the web `PluginUpdateResult`, and the toast now reports the newest actionable rejection instead of "up to date". Private-repo name masking stays deployment-publisher-only.

apps/web/src/features/plugins/components/PluginReleaseIssuesCard.tsx
laskevych: Fixed in 7bb1695f4: the card footer now uses formatDateShort (date + time).

apps/backend/src/plugin-host/enums/release-rejection-code.enum.ts
laskevych: Fixed in 7bb1695f4: replaced with a `satisfies Record<ReleaseRejectionCode, boolean>` map, so a new enum member is a compile error until its visibility is decided. The web blocklist keeps its show-unknown direction.

apps/backend/src/plugin-host/use-cases/sync-plugin-releases.service.ts
laskevych: Fixed in 7bb1695f4: below 1.0.0 the compatibility line narrows to the minor, so a 0.x plugin declares the break with a minor bump (the 0.1.0→0.1.1 spec now expects a 'bump the minor version' hint, and 0.1.0→0.2.0 is accepted). Authoring guide and SDK README spell out both cases.

PR #1569, opened 2026-08-25
fix(linkedin-ads): fetch adAnalytics per day to avoid the 15,000-element truncation

zapolsky:
## Problem

An `adAnalytics` export over a large date range could silently lose data: campaigns from the beginning of the period were missing from the result. Re-exporting a single day recovered them — the data was not gone, it was truncated.

The LinkedIn `adAnalytics` endpoint does not support pagination and caps its response at 15,000 elements (https://learn.microsoft.com/en-us/linkedin/marketing/integrations/ads-reporting/ads-reporting, *Restrictions*). A wide date range × campaigns × creatives easily exceeds the cap, and LinkedIn silently drops the excess.

## Fix

- `fetchAdAnalytics` now requests analytics **one day at a time** (day × field chunks), so a single day cannot exceed the 15,000-element limit.
- If a daily response still reaches 15,000 elements, the import finishes with a **Warning** status and a message about possible truncation (element count + example day) instead of losing data silently.

## Verification

- New tests `packages/connectors/test/Sources/LinkedInAds/Source.test.js` (3 cases): per-day URLs, field-chunk merging, truncation warning.
- Full `packages/connectors` suite: 200/200 passing.
- ESLint clean.

## Docs

- FAQ entry added to `GETTING_STARTED.md` explaining the Warning status case.

### DIFF
--- a/packages/connectors/src/Sources/LinkedInAds/Connector.js
+++ b/packages/connectors/src/Sources/LinkedInAds/Connector.js
@@ -63,8 +63,10 @@ var LinkedInAdsConnector = class LinkedInAdsConnector extends AbstractConnector 
     }
 
     for (let dayOffset = 0; dayOffset < daysToFetch; dayOffset++) {
+      // UTC arithmetic: start dates are UTC midnight, so this never drifts across DST
+      // and the day fetched, logged and checkpointed is the same calendar day.
       const date = new Date(startDate);
-      date.setDate(date.getDate() + dayOffset);
+      date.setUTCDate(date.getUTCDate() + dayOffset);
 
       for (const nodeName of timeSeriesNodes) {
         for (const urn of urns) {
@@ -81,6 +83,19 @@ var LinkedInAdsConnector = class LinkedInAdsConnector extends AbstractConnector 
       if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
         this.config.updateLastRequstedDate(date);
       }
+    }
+
+    this.reportTruncatedAnalytics();
+  }
+
+  /**
+   * Emit one warning per account for the days whose adAnalytics response hit LinkedIn's
+   * element cap. The Source records them per call; reporting here keeps a saturated
+   * account to a single warning per run instead of one warning per day.
+   */
+  reportTruncatedAnalytics() {
+    for (const [urn, days] of Object.entries(this.source.truncatedAnalyticsDays)) {
+      this.config.addWarningToCurrentStatus(this.source.buildTruncationWarning(urn, days));
     }
   }
 
--- a/packages/connectors/src/Sources/LinkedInAds/Connector.js
+++ b/packages/connectors/src/Sources/LinkedInAds/Connector.js
@@ -18,71 +18,122 @@ var LinkedInAdsConnector = class LinkedInAdsConnector extends AbstractConnector 
    */
   async startImportProcess() {
     const urns = FormatUtils.parseIds(this.config.AccountURNs.value, {prefix: 'urn:li:sponsoredAccount:'});
-    const dataSources = FormatUtils.parseFields(this.config.Fields.value);
+    const fields = FormatUtils.parseFields(this.config.Fields.value);
+    const nodeNames = Object.keys(fields);
+    const isTimeSeries = nodeName => ConnectorUtils.isTimeSeriesNode(this.source.fieldsSchema[nodeName]);
+    const timeSeriesNodes = nodeNames.filter(isTimeSeries);
+    const catalogNodes = nodeNames.filter(nodeName => !isTimeSeries(nodeName));
 
-    for (const nodeName in dataSources) {
-      await this.processNode({
-        nodeName,
-        urns,
-        fields: dataSources[nodeName] || []
-      });
-    }
-  }
-
-  /**
-   * Process a specific node (data entity)
-   * @param {Object} options - Processing options
-   * @param {string} options.nodeName - Name of the node to process
-   * @param {Array} options.urns - URNs to process
-   * @param {Array} options.fields - Fields to fetch
-   */
-  async processNode({ nodeName, urns, fields }) {
-    const isTimeSeriesNode = ConnectorUtils.isTimeSeriesNode(this.source.fieldsSchema[nodeName]);
-    const dateInfo = this.prepareDateRangeIfNeeded(nodeName, isTimeSeriesNode);
-
-    if (isTimeSeriesNode && !dateInfo) {
-      return; // Skip processing if date range preparation failed
-    }
-
-    await this.fetchAndSaveData({
-      nodeName,
-      urns,
-      fields,
-      isTimeSeriesNode,
-      ...dateInfo
-    });
-
-    // Update LastRequestedDate only for time series data and incremental runs
-    if (isTimeSeriesNode && this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
-      this.config.updateLastRequstedDate(dateInfo.endDate);
-    }
-  }
-
-  /**
-   * Fetch data from source and save to storage
-   * @param {Object} options - Fetching options
-   * @param {string} options.nodeName - Name of the node to process
-   * @param {Array} options.urns - URNs to process
-   * @param {Array} options.fields - Fields to fetch
-   * @param {boolean} options.isTimeSeriesNode - Whether node is time series
-   * @param {string} [options.startDate] - Start date for time series data
-   * @param {string} [options.endDate] - End date for time series data
-   */
-  async fetchAndSaveData({ nodeName, urns, fields, isTimeSeriesNode, startDate, endDate }) {
-    for (const urn of urns) {
-      console.log(`Processing ${nodeName} for ${urn}${isTimeSeriesNode ? ` from ${startDate} to ${endDate}` : ''}`);
-
-      const params = { fields, ...(isTimeSeriesNode && { startDate, endDate }) };
-      const data = await this.source.fetchData(nodeName, urn, params);
-
-      this.config.logMessage(data.length ? `${data.length} rows of ${nodeName} were fetched for ${urn}${endDate ? ` from ${startDate} to ${endDate}` : ''}` : `No records have been fetched`);
-
-      if (data.length || this.config.CreateEmptyTables?.value) {
-        const preparedData = data.length ? this.addMissingFieldsToData(data, fields) : data;
-        const storage = await this.getStorageByNode(nodeName);
-        await storage.saveData(preparedData);
+    for (const nodeName of catalogNodes) {
+      for (const urn of urns) {
+        await this.processCatalogNode({
+          nodeName,
+          urn,
+          fields: fields[nodeName] || []
+        });
       }
     }
+
+    await this.processTimeSeriesNodes({ urns, timeSeriesNodes, fields });
+  }
+
+  /**
+   * Imports every time series node for every account, one date at a time.
+   *
+   * The loop is date-outer on purpose: the incremental cursor may only move once a
+   * date is complete for *all* accounts and nodes, so a run interrupted midway
+   * resumes from the last fully imported date instead of restarting the range.
+   * It also keeps memory bounded to one day of analytics instead of the whole range.
+   *
+   * @param {Object} options - Processing options
+   * @param {Array<string>} options.urns - Account URNs to import
+   * @param {Array<string>} options.timeSeriesNodes - Names of the time series nodes to import
+   * @param {Object} options.fields - Map of node name to the fields selected for it
+   */
+  async processTimeSeriesNodes({ urns, timeSeriesNodes, fields }) {
+    if (!timeSeriesNodes.length) {
+      return;
+    }
+
+    const [startDate, daysToFetch] = this.getStartDateAndDaysToFetch();
+
+    if (daysToFetch <= 0) {
+      this.config.logMessage('No days to fetch for time series data');
+      return;
+    }
+
+    for (let dayOffset = 0; dayOffset < daysToFetch; dayOffset++) {
+      const date = new Date(startDate);
+      date.setDate(date.getDate() + dayOffset);
+
+      for (const nodeName of timeSeriesNodes) {
+        for (const urn of urns) {
+          await this.processTimeSeriesDay({
+            nodeName,
+            urn,
+            date,
+            fields: fields[nodeName] || []
+          });
+        }
+      }
+
+      // Every account and node stored this date, so the cursor can move past it.
+      if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
+        this.config.updateLastRequstedDate(date);
+      }
+    }
+  }
+
+  /**
+   * Fetch and store a single day of a time series node for one account
+   * @param {Object} options - Processing options
+   * @param {string} options.nodeName - Name of the node
+   * @param {string} options.urn - Account URN
+   * @param {Date} options.date - The day to import
+   * @param {Array<string>} options.fields - Array of fields to fetch
+   */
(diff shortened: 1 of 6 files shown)


prokopetsv (commented):
The core fix is right and well-tested: per-day fetching genuinely eliminates the silent 15,000-element truncation, the detection heuristic (`elements.length >= MAX_RESPONSE_ELEMENTS`) is sound, and the new tests are behavioral and cover the key paths. The date loop is safe for both run modes (the connector derives `endDate` from `startDate`, so time-of-day components match and the `<=` bound is inclusive), and the changeset follows project conventions.

One 🔴 Blocker: the merge-scope regression (see inline) converts the silent-truncation bug into a hang for the same large accounts this PR targets. The 🟡 Non-Blockers (request/token-refresh amplification without retry, warning formatting, docs style) are worth addressing in this PR but don't block the approach.

packages/connectors/src/Sources/LinkedInAds/Source.js
```diff
@@ -341,20 +342,43 @@ var LinkedInAdsSource = class LinkedInAdsSource extends AbstractSource {
...
+        const res = await this.makeRequest(url);
+        const elements = res.elements || [];
+
+        if (elements.length >= this.MAX_RESPONSE_ELEMENTS) {
+          dayTruncated = true;
+        }
+
+        // Merge results from different chunks into a single dataset
+        // Each chunk contains the same rows but different fields
+        allResults = this.mergeAnalyticsResults(allResults, elements);
```
prokopetsv: 🔴 **Blocker — cross-day merge is quadratic over the whole date range.**

`mergeAnalyticsResults(allResults, elements)` runs a `findIndex` with two `JSON.stringify` calls per comparison against the *entire multi-day accumulation*, but rows from different days can never match (their `dateRange` always differs). Before this PR the scan set was capped at one response (≤15,000 rows); now it grows by up to 15,000 rows per day. For exactly the accounts this fix targets — near the cap, long backfills — day N's merge scans N × rows_per_day elements per incoming element. A 90-day backfill at 10k rows/day is on the order of 10^10 stringify comparisons; the run will effectively hang.

Fix: merge field chunks into a per-day buffer, then append the day to the total:

```js
let dayResults = [];
for (const fieldChunk of fieldChunks) {
  // ...
  dayResults = this.mergeAnalyticsResults(dayResults, elements);
}
allResults.push(...dayResults);
```

packages/connectors/src/Sources/LinkedInAds/Source.js
```diff
@@ -341,20 +342,43 @@ var LinkedInAdsSource = class LinkedInAdsSource extends AbstractSource {
...
-        fields: fieldChunk
-      });
-      const res = await this.makeRequest(url);
-      const elements = res.elements || [];
+    // The adAnalytics endpoint does not support pagination and caps each response at
+    // MAX_RESPONSE_ELEMENTS elements, so the requested range is fetched one day at a time
+    // to keep every request under the limit.
+    const truncatedDays = [];
+
+    for (let day = new Date(startDate); day <= endDate; day.setDate(day.getDate() + 1)) {
```
prokopetsv: 🟡 **Non-Blocker — request volume multiplies with no throttling protection.**

Request count goes from `fieldChunks` to `days × fieldChunks`, sequential. On top of that, `makeRequest` calls `OAuthUtils.getAccessToken` on *every* request with no caching, so a 365-day backfill with 2 field chunks now issues ~730 token-refresh POSTs plus ~730 API calls. `LinkedInAdsSource` does not override `isValidToRetry`, so the `AbstractSource` default (never retry) applies — a single 429 anywhere in a long range aborts the whole run.

Suggested mitigations, in order of value:
1. Refresh the token once per run instead of per request.
2. Add a LinkedIn `isValidToRetry` override for 429/5xx.
3. Optionally fetch the full range first and only fall back to per-day (or bisect) when a response hits the cap — keeps small accounts at today's request count.

packages/connectors/src/Sources/LinkedInAds/Source.js
```diff
@@ -341,20 +342,43 @@ var LinkedInAdsSource = class LinkedInAdsSource extends AbstractSource {
...
-      // Merge results from different chunks into a single dataset
-      // Each chunk contains the same rows but different fields
-      allResults = this.mergeAnalyticsResults(allResults, elements);
+      if (dayTruncated) {
+        truncatedDays.push(this.formatDateForUrl(day));
+      }
+    }
+
+    if (truncatedDays.length > 0) {
+      this.config.addWarningToCurrentStatus(
```
prokopetsv: 🟡 **Non-Blocker — warning shows an internal date format and understates coverage.**

The user-facing warning renders the example day as `(year:2026,month:8,day:1)` (the URL format from `formatDateForUrl`), and names only one example day, while the new FAQ promises the run "names the affected days in the log."

Since the warning is the user-facing deliverable of this PR: format the day as `YYYY-MM-DD` (a `formatDateFromLinkedInObject`-style helper already exists) and either list the truncated days (bounded, e.g. first 10) or align the FAQ wording.

packages/connectors/src/Sources/LinkedInAds/GETTING_STARTED.md
```diff
@@ -72,6 +72,12 @@ The process is complete when the **Run history** tab shows the message:
 
 ![LinkedIn Ads Success](res/linkedin_ads_successrun.png)
 
+## FAQ
+
+### Why does my Ad Analytics run finish with a Warning status?
+
+LinkedIn's `adAnalytics` endpoint does not support pagination and caps each response at **15,000 rows**. To avoid silently losing data on long date ranges, the connector fetches analytics **one day at a time**. If a single day's response still reaches that limit (for example, an account with more than 15,000 active creatives), the run finishes with a **Warning** status and names the affected days in the log — data for those days may be incomplete.
```
prokopetsv: 🟡 **Non-Blocker — violates the repo's docs style rule.**

This sentence is ~40 words; the repo's documented style for connector docs caps sentences at 20 words. Split it, e.g.:

> A single day can still exceed the limit — for example, more than 15,000 active creatives. The run then finishes with a Warning status and logs the affected days. Data for those days may be incomplete.

packages/connectors/src/Sources/LinkedInAds/Source.js
prokopetsv: Fixed in `8f798559b`.

I applied the per-day buffer as described, and also rewrote `mergeAnalyticsResults` itself. The buffer alone fixes the cross-day scan, but it still leaves the within-day cost: a 15,000-row day with the old `findIndex` + two `JSON.stringify` per comparison is on the order of 10^8 comparisons *per field chunk*. The merge now keys rows in a `Map` (`JSON.stringify([dateRange, pivotValues])`), so it is linear and the method got shorter.

Measured on a deliberately scaled-down shape — 20 days × 500 rows, versus the real 90 × 10,000:

```
old: 58555 ms    new: 22 ms    (10,000 rows, identical output)
```

Two tests added: one asserting rows with the same pivot on different days stay separate, one covering `mergeAnalyticsResults` field-combining and non-mutation of the input.

packages/connectors/src/Sources/LinkedInAds/Source.js
prokopetsv: Fixed points 1 and 2 in `8f798559b`; deliberately skipped point 3.

1. Token is now exchanged once per run via a `getAccessToken` method guarded by an instance flag. Confirmed on a real run: 204 analytics requests, `Successfully obtained access token` logged exactly once. Previously that was one token POST per request.
2. Added `isValidToRetry` for 429, 5xx, and network errors (no status code), matching the Criteo/Shopify connectors.

While doing this I hit a related problem worth recording: `makeRequest` called `HttpUtils.fetch` directly and never checked the status code, so a non-200 body was `JSON.parse`d and returned as data — a 429 did not abort the run, it silently produced an empty day. It now goes through `urlFetchWithRetry`, which validates the status before returning.

On point 3 (fetch full range first, fall back to per-day): not doing it for now. Scheduled runs are bounded by `MaxFetchingDays` (31), and the measured cost is acceptable — a 34-day run was 204 requests in 85s, about 0.42s each. A full-year manual backfill would be ~2,190 requests, roughly 15 minutes. Worth revisiting if backfill latency becomes a real complaint, but the adaptive/bisect logic is not earning its complexity yet.

packages/connectors/src/Sources/LinkedInAds/Source.js
prokopetsv: Fixed in `8f798559b`. The warning now formats days as `YYYY-MM-DD` and lists them rather than naming one example, capped at 10 days with an `and N more` suffix so it cannot blow up on a long backfill.

Extracted a small `toLinkedInDateObject` helper so `formatDateForUrl` and the warning share one source for the year/month/day split, instead of the warning reaching for the URL format. Aligned the FAQ wording to "lists the affected days". Tests cover both the date format and the 10-day bound.

packages/connectors/src/Sources/LinkedInAds/GETTING_STARTED.md
prokopetsv: Fixed in `8f798559b`, using your wording. The paragraph is now five sentences, each under 20 words. `markdownlint` passes.

prokopetsv (approved):

packages/connectors/src/Sources/LinkedInAds/Source.js
```diff
@@ -334,31 +336,88 @@ var LinkedInAdsSource = class LinkedInAdsSource extends AbstractSource {
...
     // LinkedIn API has a limitation - it allows a maximum of fields per request
     // To overcome this, split fields into chunks and make multiple requests
     const fieldChunks = this.prepareAnalyticsFieldChunks(uniqueApiFields);
 
+    // The adAnalytics endpoint does not support pagination and caps each response at
+    // MAX_RESPONSE_ELEMENTS elements, so the requested range is fetched one day at a time
+    // to keep every request under the limit.
+    const truncatedDays = [];
+
+    for (let day = new Date(startDate); day <= endDate; day.setDate(day.getDate() + 1)) {
```
laskevych: 💡 **Suggestion (non-blocking):** since `bb52155c8` the connector always calls `fetchData` with `startDate === endDate`, so the multi-day loop here, `truncatedDays`, `MAX_TRUNCATED_DAYS_IN_WARNING` and the `and N more` suffix are unreachable from production code, and the "lists at most 10 truncated days" test only covers that dead path.

The practical effect is that truncation is now reported as one warning per (account, day): a 31-day incremental run for a saturated account records 31 separate warnings per account (the backend stores WARNING messages without a cap).

Two ways to line this up, either is fine: aggregate truncated days in `processTimeSeriesNodes` and emit one warning per account per run, or reduce the Source to a single-day API and drop the cap. Happy to leave it as a follow-up.

packages/connectors/src/Sources/LinkedInAds/Connector.js
```diff
@@ -18,71 +18,122 @@ var LinkedInAdsConnector = class LinkedInAdsConnector extends AbstractConnector
...
+   * @param {string} options.nodeName - Name of the node
+   * @param {string} options.urn - Account URN
+   * @param {Date} options.date - The day to import
+   * @param {Array<string>} options.fields - Array of fields to fetch
    */
-  async fetchAndSaveData({ nodeName, urns, fields, isTimeSeriesNode, startDate, endDate }) {
-    for (const urn of urns) {
-      console.log(`Processing ${nodeName} for ${urn}${isTimeSeriesNode ? ` from ${startDate} to ${endDate}` : ''}`);
+  async processTimeSeriesDay({ nodeName, urn, date, fields }) {
+    const formattedDate = DateUtils.formatDate(date);
```
laskevych, line 111: 🔵 **Question (non-blocking):** `DateUtils.formatDate` is UTC-based, while `toLinkedInDateObject` in the Source builds the `dateRange` from local getters. The incremental start date is created at UTC midnight, so on a runner with a negative UTC offset the log says "Start importing data for 2026-08-01" but the URL asks for `day:31` of July, and the cursor is saved as 2026-08-01.

The URL behaviour predates this PR (Microsoft Ads has the same getters), and managed runs are UTC, so not blocking. Do we want the Source to derive the day from the same UTC formatting so the log, the request and the checkpoint always agree?

laskevych (approved):
✅ **Approve.** Reviewed the final state including `bb52155c8`, which moved the day loop into the connector after the earlier approval. The per-day fetch, save and checkpoint order mirrors the Criteo/Microsoft Ads connectors, the backend persists `LastRequestedDate` immediately on each message, and the Map-based merge, one-per-run token exchange and `isValidToRetry` all check out against `AbstractSource`.

Verified locally: 23/23 tests in `test/Sources/LinkedInAds`, ESLint and markdownlint clean.

Two non-blocking inline notes; neither needs to hold the merge.

packages/connectors/src/Sources/LinkedInAds/Source.js
prokopetsv: Fixed in `5a983c6c4`. Went with the first option, since the warning volume was the practical problem.

`fetchAdAnalytics` now fetches exactly the single day it is given — no range loop — and, if any field chunk hits the cap, records the day in `truncatedAnalyticsDays[urn]` instead of warning. `processTimeSeriesNodes` calls `reportTruncatedAnalytics()` once after the day loop and emits one warning per account listing its days. So a saturated account produces one warning per run instead of 31, and the 10-day cap / `and N more` are live code on the aggregated path rather than dead.

Tests: the Source records instead of warning and accumulates across calls; `buildTruncationWarning` is unit-tested for the 1-day and 12-day shapes; the connector test asserts one warning per account after the loop and none when nothing was truncated.

packages/connectors/src/Sources/LinkedInAds/Connector.js
prokopetsv: Yes — fixed in `5a983c6c4`. `toLinkedInDateObject` now reads the UTC parts, and `processTimeSeriesNodes` advances with `setUTCDate`, so the request, the `DateUtils.formatDate` log line and the checkpoint always name the same day.

Start dates are UTC midnight in both modes (`AbstractConfig` parses the string with `new Date`, the incremental start uses `Date.UTC`), so this is a no-op on the UTC managed runners and a real fix on negative-offset self-hosted ones.

`setUTCDate` also removes a DST drift the local `setDate` arithmetic had: on a Europe runner after the March switch, `startDate + k` local days lands at 23:00Z of the previous day, so `formatDate` would name day *k−1* and one day gets fetched twice. Tests: `formatDateForUrl(2026-07-31T23:30Z)` → `day:31`; the connector test runs three days across 2026-03-29 and asserts consecutive UTC days.

Criteo, Google Ads and Microsoft Ads use the same local `setDate` pattern, so the drift exists there too — worth a Core-level follow-up, out of scope here.

packages/connectors/src/Sources/LinkedInAds/Connector.js
```diff
@@ -18,71 +18,137 @@ var LinkedInAdsConnector = class LinkedInAdsConnector extends AbstractConnector
...
-      urns,
-      fields,
-      isTimeSeriesNode,
-      ...dateInfo
-    });
-
-    // Update LastRequestedDate only for time series data and incremental runs
-    if (isTimeSeriesNode && this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
-      this.config.updateLastRequstedDate(dateInfo.endDate);
+    this.reportTruncatedAnalytics();
```
prokopetsv: 🔴 Blocker: `reportTruncatedAnalytics()` runs only after the entire date loop. If an early day is truncated, subsequent days advance the incremental cursor, and a later request fails, execution never reaches this warning. On retry, the lookback window may exclude the truncated day, leaving incomplete data stored without ever reporting it. Emit or persist the warning before advancing beyond each affected day, and add a truncated-day-then-later-failure test.

PR #1572, opened 2026-08-27
feat(backend): add report output-schema endpoint

romandubovyi:
## Summary

Adds `GET /reports/:id/output-schema`: the columns a report's rows will carry (`name`, `title`, `description`, `type`, plus `aggregateFunction` and `calculatedFieldLevel`), including aggregated columns, Unique Count and calculated fields, which do not exist on the Data Mart schema. Resolved from the stored schema and report config, without opening a storage reader. Not warehouse-free, though: resolving the report's blending decision resolves its table references, which refreshes a SQL-defined Data Mart's view — the same `GET /reports/:id/generated-sql` already does.

`calculatedFieldLevel` travels because `type` alone cannot separate a non-additive calculated metric from an ordinary numeric column, and a consumer that re-aggregates the first is wrong at any grain. An absent value means a native column, not "unknown".

Excel joined-field labels now use the same form as Google Sheets: `Field name (Data Mart name)`.

## Reaching it from the clients

A backend route alone is not usable by the clients that need it. In pull mode the Excel add-in reads rows from the HTTP data stream and has no way to learn the real column names — that is what this schema is for.

- `@owox/api-client` — `ReportsApi.getOutputSchema`, beside the existing `traverseData`, with `OWOXReportOutputSchemaField` exported.
- MCP — `get_report_output_schema`, a read-only tool on `mcp:read`, over a new `McpReportsFacade.getReportOutputSchema`. Optional fields are published as explicit nulls: in JSON an absent key reads as "unknown" rather than "no alias".
- `@owox/plugin-sdk` — no code of its own. `connect()` already hands plugins a real `OWOXApiClient` over the iframe transport, and the host bridge forwards any root-relative `/api/` path. A test pins that chain end to end.

Access is unchanged and enforced where it already was: the report is looked up scoped to the caller's project, then gated on SEE of the source data mart — the same level `get_data_mart_reports` requires. The MCP tool takes only `report_id`; project, user and roles come from the verified token.

### DIFF
--- a/apps/backend/src/data-marts/controllers/spec/report.api.ts
+++ b/apps/backend/src/data-marts/controllers/spec/report.api.ts
@@ -118,7 +118,7 @@ export function GetReportOutputSchemaSpec() {
     ApiParam({ name: 'id', description: 'Report ID' }),
     ApiOkResponse({
       description:
-        "The columns the report's rows will carry, in the order they are projected. `name` is the key each row is keyed by; `title` is the alias configured for it, absent when there is none. Columns the report synthesises — aggregated (`<column> | <FN>`), Unique Count, calculated fields — appear here and nowhere in the Data Mart schema.",
+        "The columns the report's rows will carry, in the order they are projected, as of the Data Mart's last schema actualization. `name` is the key each row is keyed by; `title` is the alias configured for it, absent when there is none. Columns the report synthesises — aggregated (`<column> | <FN>`), Unique Count, calculated fields — appear here and nowhere in the Data Mart schema. This reads the stored schema rather than the warehouse, so a column added warehouse-side and not yet actualized is missing here until a run or a data read picks it up.",
       type: [ReportOutputSchemaFieldApiDto],
     })
   );
--- a/apps/backend/src/data-marts/controllers/spec/report.api.ts
+++ b/apps/backend/src/data-marts/controllers/spec/report.api.ts
@@ -8,6 +8,7 @@ import {
   ApiParam,
 } from '@nestjs/swagger';
 import { ReportResponseApiDto } from '../../dto/presentation/report-response-api.dto';
+import { ReportOutputSchemaFieldApiDto } from '../../dto/presentation/report-output-schema-field-api.dto';
 import { ReconnectGoogleSheetResponseDto } from '../../dto/presentation/google-sheets/reconnect-google-sheet-response.dto';
 import {
   createReportRequestBodySchema,
@@ -107,6 +108,18 @@ export function ListReportsByInsightTemplateSpec() {
     ApiOkResponse({
       description: 'List of email reports that use the insight template as their template source',
       type: [ReportResponseApiDto],
+    })
+  );
+}
+
+export function GetReportOutputSchemaSpec() {
+  return applyDecorators(
+    ApiOperation({ summary: "Get the resolved column schema of a report's output" }),
+    ApiParam({ name: 'id', description: 'Report ID' }),
+    ApiOkResponse({
+      description:
+        "The columns the report's rows will carry, in the order they are projected. `name` is the key each row is keyed by; `title` is the alias configured for it, absent when there is none. Columns the report synthesises — aggregated (`<column> | <FN>`), Unique Count, calculated fields — appear here and nowhere in the Data Mart schema.",
+      type: [ReportOutputSchemaFieldApiDto],
     })
   );
 }
--- a/apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts
+++ b/apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts
@@ -72,8 +72,16 @@ export function requiresCredentials(type: DataDestinationType): boolean {
  * on every column at once. The other destinations render the label with enough room that the
  * prefix reads fine, so they keep it.
  *
- * Excel qualifies on that same ground. Left out here, a run's recorded label disagreed with the
- * header written into the worksheet.
+ * Excel qualifies on that same ground, but reaches the label by a different road than Google
+ * Sheets: nothing pushes rows into a workbook, so its add-in pulls the worksheet header from
+ * `GET /reports/:id/output-schema`, which resolves the style from the Report. This opt-in is what
+ * that header follows, and it is the only surface it reaches.
+ *
+ * It deliberately does NOT reach the same report's stream metadata or run record. Those are built
+ * from a `ReportLikeReadPlan`, which drops `dataDestination` on purpose (see `streamReport` in
+ * `stream-http-data.service.ts`) so that NDJSON — where a prefix has room to read fine — does not
+ * change shape just because a report happens to write to a spreadsheet. So the two differ by
+ * design: suffix in the worksheet header, prefix in the recorded run. Nothing keys on either.
  *
  * What this pins is the POSITION of the Data Mart name, not a byte-exact label: every surface
  * still normalizes whitespace (see `formatBlendedFieldDisplayName`), and no consumer keys on the
--- a/apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts
+++ b/apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts
@@ -67,10 +67,13 @@ export function requiresCredentials(type: DataDestinationType): boolean {
  * Destinations that render a field from a joined Data Mart as `Field name (Data Mart name)`.
  * Everywhere else the Data Mart name stays a prefix: `Data Mart name Field name`.
  *
- * Google Sheets writes the label into a spreadsheet header cell, which is narrow and not resizable
- * per column by the reader — a leading Data Mart name pushes the field name out of view on every
- * column at once. The other destinations render the label with enough room that the prefix reads
- * fine, so they keep it.
+ * Both spreadsheet destinations write the label into a header cell, which is narrow and not
+ * resizable per column by the reader — a leading Data Mart name pushes the field name out of view
+ * on every column at once. The other destinations render the label with enough room that the
+ * prefix reads fine, so they keep it.
+ *
+ * Excel qualifies on that same ground. Left out here, a run's recorded label disagreed with the
+ * header written into the worksheet.
  *
  * What this pins is the POSITION of the Data Mart name, not a byte-exact label: every surface
  * still normalizes whitespace (see `formatBlendedFieldDisplayName`), and no consumer keys on the
@@ -80,5 +83,5 @@ export function requiresCredentials(type: DataDestinationType): boolean {
  * New destination types default to the prefix; opt in here deliberately.
  */
 export function usesSuffixedJoinedFieldNames(type: DataDestinationType): boolean {
-  return type === DataDestinationType.GOOGLE_SHEETS;
+  return type === DataDestinationType.GOOGLE_SHEETS || type === DataDestinationType.EXCEL;
 }
--- a/apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
+++ b/apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
@@ -78,12 +78,16 @@ export class GetReportOutputSchemaService {
       );
     }
 
-    const accessor = { userId: command.userId, roles: command.roles };
-    const decision = await this.blendedReportDataService.resolveBlendingDecision(report, accessor);
-
+    // Ahead of the blending decision, not just ahead of the header build: without a stored schema
+    // this request can only end in the exception below, and the decision is the one step here that
+    // still writes — it refreshes a SQL-defined source's view on the joined path. Failing first
+    // keeps a doomed request from touching the warehouse at all.
     if (!report.dataMart.schema) {
       throw new BusinessViolationException('Data mart schema must be provided');
     }
+
+    const accessor = { userId: command.userId, roles: command.roles };
+    const decision = await this.blendedReportDataService.resolveBlendingDecision(report, accessor);
 
     // Calculated fields are named by their formula and exist only in the plan that built them.
     //
--- a/apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
+++ b/apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
@@ -1,0 +1,107 @@
+import {
+  ForbiddenException,
+  Injectable,
+  NotFoundException,
+  UnauthorizedException,
+} from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+import { BusinessViolationException } from '../../common/exceptions/business-violation.exception';
+import { columnFilterWithoutCalculatedFields } from '../calculated-fields/calculated-field.utils';
+import { ReportHeadersGeneratorFacade } from '../data-storage-types/facades/report-headers-generator.facade';
+import { resolveReportDataHeaders } from '../data-storage-types/utils/report-data-headers.utils';
+import { CalculatedFieldPlan } from '../data-storage-types/utils/sql-clause-renderer';
+import { GetReportOutputSchemaCommand } from '../dto/domain/get-report-output-schema.command';
+import { ReportDataHeader } from '../dto/domain/report-data-header.dto';
+import { hasOutputControls } from '../dto/domain/report-like-read-plan';
+import { hasMainUniqueCount } from '../dto/schemas/unique-count-sources';
+import { Report } from '../entities/report.entity';
+import { AccessDecisionService, Action, EntityType } from '../services/access-decision';
+import { BlendedReportDataService } from '../services/blended-report-data.service';
+import { ReportSqlComposerService } from '../services/report-sql-composer.service';
+
+/**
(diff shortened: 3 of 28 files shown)


laskevych (approved):
Traced the new endpoint's header derivation against the run and stream paths (options assembly, calculated-field plans, blended label styles, Unique Count gating) — the mirror is faithful, and bypassing compose() correctly removes the view refresh on the non-blended path. Architecture, DTO boundaries, OpenAPI spec and the changeset format all follow the module's conventions. No blocking issues; a few non-blocking comments and questions follow separately.

apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts
```diff
@@ -67,10 +67,13 @@ export function requiresCredentials(type: DataDestinationType): boolean {
...
- * Google Sheets writes the label into a spreadsheet header cell, which is narrow and not resizable
- * per column by the reader — a leading Data Mart name pushes the field name out of view on every
- * column at once. The other destinations render the label with enough room that the prefix reads
- * fine, so they keep it.
+ * Both spreadsheet destinations write the label into a header cell, which is narrow and not
+ * resizable per column by the reader — a leading Data Mart name pushes the field name out of view
+ * on every column at once. The other destinations render the label with enough room that the
+ * prefix reads fine, so they keep it.
+ *
+ * Excel qualifies on that same ground. Left out here, a run's recorded label disagreed with the
```
laskevych: 🟡 **Important (non-blocking):** The justification paragraph — *"Left out here, a run's recorded label disagreed with the header written into the worksheet"* — reads backwards relative to what this helper can affect. The run-recorded label for Excel comes from the HTTP stream's read plan, which deliberately drops `dataDestination` (`buildPlan` in `stream-http-data.service.ts:324-330`, pinned by the *"does not carry the report destination into the read plan"* spec), so it stays prefix-form no matter what this function returns. With EXCEL opted in, the run record (`Orders Amount`) now permanently disagrees with the worksheet header the add-in will write from output-schema (`Amount (Orders)`) — the exact class of mismatch this paragraph says the flip closes. Functionally harmless (rows are keyed by `name`, and this doc itself says no consumer keys on the label), so I'm fine shipping the flip — but please reword the comment: the suffix is for the worksheet header delivered via output-schema, and the prefix intentionally remains in stream metadata and run records.

apps/backend/src/ee/mcp/tools/get-report-output-schema.tool.ts
```diff
@@ -0,0 +1,90 @@
...
+        name: z.string().describe('The key each output row is keyed by.'),
+        title: z.string().nullable().describe('Alias configured for the column, if any.'),
+        description: z.string().nullable(),
+        type: z.string().nullable().describe('Storage field type, when it can be derived.'),
+        aggregate_function: z
+          .string()
+          .nullable()
+          .describe('The aggregate function the report applies to this column, if any.'),
+        calculated_field_level: z
+          .enum(['metric', 'column'])
```
laskevych: 🟡 **Important:** `z.enum(['metric', 'column'])` in the output schema is enforced at runtime: `mcp-sdk-server.factory.ts:57` hands `outputSchema` to the SDK's `registerTool`, and SDK 1.30 validates `structuredContent` against it, throwing `McpError` on mismatch. `formula-level.ts` exists precisely to prevent this copy — its doc says a hand-written `z.enum(['metric', 'column'])` beside it *"would be a second source that drifts the moment a level is added"* — and `data-mart-details.tool.ts:63-65` records the sibling decision to keep `level` a bare string in output schemas. As written, adding a third level turns an additive domain change into a runtime failure of this tool. Either the bare-string precedent or `z.enum(CALCULATED_FIELD_LEVELS)` works (`ee/mcp` already imports `formula-level.ts` elsewhere). Same hand-copied pair sits on `report-output-schema-field-api.dto.ts:42` — there it's only Swagger drift, but `enum: CALCULATED_FIELD_LEVELS` plus typing the field `CalculatedFieldLevel` instead of `string` is free.

apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
```diff
@@ -0,0 +1,131 @@
...
+    // Ahead of the blending decision, not just ahead of the header build: without a stored schema
+    // this request can only end in the exception below, and the decision is the one step here that
+    // still writes — it refreshes a SQL-defined source's view on the joined path. Failing first
+    // keeps a doomed request from touching the warehouse at all.
+    if (!report.dataMart.schema) {
+      throw new BusinessViolationException('Data mart schema must be provided');
+    }
+
+    const accessor = { userId: command.userId, roles: command.roles };
+    const decision = await this.blendedReportDataService.resolveBlendingDecision(report, accessor);
```
laskevych, line 86: 🔵 **Question:** The class comment names the remaining gap and its fix — a headers-only mode on `resolveBlendingDecision` — and the tool description discloses the caveat honestly. Still: the tool ships `readOnlyHint: true` under `mcp:read`, and agents treat that as safe-to-poll, while for a blended report every call resolves table references for the main and each joined SQL-defined mart (one `CREATE OR REPLACE VIEW` per source) and renders blended SQL that is then discarded. Is the headers-only mode planned as a near-term follow-up? `blendedDataHeaders`/`uniqueCountSources`/`calculatedFields` are derivable from stored schemas alone (`resolveUniqueCountSources` needs `cteName` and target PK fields, not table references), so the write path looks closable without touching header semantics. Worth doing before clients learn to poll this.

apps/backend/src/data-marts/controllers/spec/report.api.ts
```diff
@@ -111,6 +112,18 @@ export function ListReportsByInsightTemplateSpec() {
   );
 }
 
+export function GetReportOutputSchemaSpec() {
+  return applyDecorators(
+    ApiOperation({ summary: "Get the resolved column schema of a report's output" }),
+    ApiParam({ name: 'id', description: 'Report ID' }),
+    ApiOkResponse({
+      description:
+        "The columns the report's rows will carry, in the order they are projected. `name` is the key each row is keyed by; `title` is the alias configured for it, absent when there is none. Columns the report synthesises — aggregated (`<column> | <FN>`), Unique Count, calculated fields — appear here and nowhere in the Data Mart schema.",
```
laskevych: 💡 **Suggestion:** "The columns the report's rows will carry" slightly overpromises: a run actualizes the schema unconditionally (`run-report.service.ts:401`) and the stream on a 30-minute TTL, while this endpoint reads the stored schema as-is — consistent with `generated-sql`, deliberate, and self-healing within one delivery cycle, but a describe issued after a warehouse-side schema change and before any actualizing event can omit a column a run will deliver. A clause like *"as of the Data Mart's last schema actualization"* here (and in the MCP tool description) would make the contract exact without reintroducing warehouse access.

apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
```diff
@@ -0,0 +1,131 @@
...
+      );
+      calculatedFields = plans.length > 0 ? plans : undefined;
+    }
+
+    const nativeHeaders = await this.reportHeadersGeneratorFacade.generateHeadersFromSchema(
+      report.dataMart.storage.type,
+      report.dataMart.schema
+    );
+
+    return resolveReportDataHeaders(
```
laskevych, line 100: 💡 **Suggestion (follow-up scope):** This options block is now the fifth near-verbatim copy of the same 7-key assembly (`run-report.service.ts:318-329`, `stream-http-data.service.ts:458-468`, `report-data-cache.service.ts:52-87`, `query-data-mart.service.ts:299-313`). The mirror is faithful today, but the next header-affecting option added to run/stream — the paths a feature's own tests exercise — can silently miss this copy, and then the schema stops matching the row keys `traverseData` delivers, which is this endpoint's entire contract. Extracting a shared helper for the tail is mechanical for four of the five sites; the calculated-fields branch would stay per-site, since avoiding `compose()` here is load-bearing. Fine as a follow-up.

apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
```diff
@@ -0,0 +1,131 @@
...
+    // and for a SQL-defined Data Mart that runs `CREATE OR REPLACE VIEW` against the customer's
+    // warehouse — DDL a describe endpoint under `Role.viewer` must not issue. The plans are the
+    // same ones `compose` derives before it ever reaches that step, so nothing about the headers
+    // changes; what goes away is the warehouse write, a second `resolveBlendingDecision`, and the
+    // SQL-composition errors that used to fail a request that only ever needed to name columns.
+    let calculatedFields: CalculatedFieldPlan[] | undefined;
+    if (decision.needsBlending) {
+      calculatedFields = decision.calculatedFields;
+    } else if (hasOutputControls(report)) {
+      const plans = this.reportSqlComposerService.buildCalculatedFieldPlans(
```
laskevych: 💡 **Suggestion (optional):** On the non-blended output-controls path, `resolveBlendingDecision` has already built these exact plans (`blended-report-data.service.ts:268`) and drops them at its late `needsBlending: false` return (`:303-309`) — so every formula gets parsed twice per request. Returning `calculatedFields` from that late return would let both branches read `decision.calculatedFields`, deleting this rebuild and the need to make `buildCalculatedFieldPlans` public. Checked the other callers: safe — whenever plans are non-empty, `hasOutputControls` is true and stream/cache overwrite from `compose()` anyway. One invariant comment in `blending-decision.dto.ts:38-43` would need updating. Entirely optional; the current form is correct.

apps/backend/src/ee/mcp/tools/get-report-output-schema.tool.ts
```diff
@@ -0,0 +1,90 @@
...
+import { z } from 'zod';
+import type { McpScope } from '@owox/idp-protocol';
+import {
+  MCP_REPORTS_FACADE,
+  type McpReportsFacade,
+} from '../../../data-marts/facades/mcp-reports.facade';
+import type { McpAuthContext } from '../auth/mcp-auth-context';
+import { jsonToolResult, type McpToolDefinition, type McpToolResult } from './mcp-tool.definition';
+
+const inputSchema = z.object({ report_id: z.string().min(1) }).strict();
```
laskevych: 💡 **Suggestion (nit):** `report_id: z.string().min(1)` without `.trim()` joins the no-trim minority and turns the sibling split into 3-3 (`run_report`, `get_report_run_status`, `create_report_run_schedule` trim; `update_report`, `delete_report` don't). A padded id — common from LLM callers — 404s here while succeeding against `get_report_run_status`. One token: `.trim()`.

laskevych (approved):
Re-reviewed the two commits added after my approval (api-client/MCP/plugin-sdk surface and `calculatedFieldLevel`). The clients reach the endpoint through the proper boundaries, optional-vs-null semantics are handled deliberately on the MCP edge, and coverage pins the chain end to end. No blocking issues; a few non-blocking comments and questions inline.

apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
romandubovyi: Yes — near-term follow-up, and deliberately not in this PR: the change lands inside `resolveBlendingDecision` and `buildRelationshipChains`, which every blended run/stream/totals/cache path goes through, so it wants its own PR with those suites run rather than riding along with a describe endpoint.

Your read matches what I traced. Every header-relevant value is final before the SQL build: `blendedDataHeaders` at `:296`, `primaryKeyColumns` before the early return, `uniqueCountSources` at `:349` and never re-filtered afterwards, `normalizedAggregations` and `calculatedFields` as *inputs* to `buildBlendedQuery` — which returns only `blendedSql`/`params`. `mainTableReference` and each chain's `targetTableReference` feed the builder and nothing else. The amplification is worth naming too: it is main plus one refresh per joined source, per call.

One addition before it lands. The safe shape is a separate method returning a type without `blendedSql`/`params`, not a boolean flag on the existing one: a flag returns the same `BlendingDecision` with a hole in it, and `blendedSql` is already optional, so nothing would catch a misuse at compile time. `stream-http-data.service.ts:421` throws on `needsBlending && !blendedSql`, but `run-report.service.ts:278` has no such guard — it would silently fall back to a flat read of the main table and record it as a successful run. That guard is worth adding on its own, independent of this work.

apps/backend/src/ee/mcp/tools/get-report-output-schema.tool.ts
romandubovyi: Fixed in 38da0c55e — and you were right that this one could actually fire.

Followed the bare-string precedent rather than `z.enum(CALCULATED_FIELD_LEVELS)`: `data-mart-details.tool.ts` records the reason ("an OUTPUT schema over a value the facade forwards from persisted JSON"), and my case is the same shape, so consistency with the sibling wins over the extra precision. That also removes the failure class rather than keeping it synchronised.

Added a test that feeds the output schema a level the tool has never heard of and expects it to pass — it fails against the old `z.enum`, so the guard is real rather than decorative.

The DTO now takes `enum: CALCULATED_FIELD_LEVELS` and types the field `CalculatedFieldLevel` instead of `string`, as suggested.

apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts
romandubovyi: Reworded in 38da0c55e — the paragraph now states the split instead of the backwards claim.

Your trace matches mine: the run-recorded label is built from a `ReportLikeReadPlan` that drops `dataDestination` on purpose, so it stays prefix-form no matter what this helper returns. The comment now says Excel reaches the label by a different road — the add-in pulls the worksheet header from output-schema, which resolves the style from the `Report` — and that stream metadata and run records intentionally keep the prefix, where a label has room to read fine. Behaviour untouched.

apps/backend/src/data-marts/controllers/spec/report.api.ts
romandubovyi: Added in 38da0c55e, to both descriptions.

The Swagger description now says the columns are "as of the Data Mart's last schema actualization" and spells out the consequence: a column added warehouse-side and not yet actualized is missing here until a run or a data read picks it up. Same clause on the MCP tool description, where an agent is the one reading it.

Worth noting the window is narrow in the shape that motivated the endpoint: the add-in calls the schema and then opens the stream, which actualizes on its 30-minute TTL.

apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
romandubovyi: Taken in 38da0c55e — and it turned out better than "optional", so thanks for pushing on it.

Verified your claim before changing anything, since it contradicted my own earlier trace. The plans are built at `:268` and dropped by the `needsBlending: false` return at `:303-309`, and the two builders agree on every header-relevant field: `outputName`, `type`, `level` (both via `calculatedFieldLevelOf`), `alias`, `description`, and both partition through `partitionCalculatedPlans` with the same `aggregationConfig`. The blending one adds `formulaOwnership`, which headers never read.

Also confirmed the caller safety independently: `stream-http-data.service.ts` reads `decision.calculatedFields` *ungated*, but overwrites from `compose()` when `!needsBlending && hasOutputControls` — and non-empty plans mean a calculated field is selected, so `hasSelectedCalculatedField` makes that true. Same shape in cache.

Net effect is bigger than removing the double parse: both branches now read `decision.calculatedFields`, `buildCalculatedFieldPlans` is private again, and the describe service no longer injects `ReportSqlComposerService` at all. Invariant comment on `blending-decision.dto.ts` updated to cover both paths.

apps/backend/src/data-marts/use-cases/get-report-output-schema.service.ts
romandubovyi: Agreed, and tracked rather than done here: [6882](https://owox.fibery.io/Sprint_Work/Work_Item/6882), next sprint, alongside the headers-only mode as [6881](https://owox.fibery.io/Sprint_Work/Work_Item/6881) — both touch the same core, so they pair.

Your risk framing is the reason it is worth doing: the next header-affecting option lands with tests that exercise run and stream, and the describe copy is the one that silently misses it — after which the schema stops matching the row keys `traverseData` delivers, which is this endpoint's whole contract.

Keeping it out of this PR for the same reason as the headers-only work: it edits four hot delivery paths, so it deserves its own PR with their suites run rather than riding along with a describe endpoint. One note for whoever picks it up — the calculated-fields branch here just collapsed to `decision.calculatedFields`, so this copy is closer to the others than when you read it.

apps/backend/src/ee/mcp/tools/get-report-output-schema.tool.ts
romandubovyi: Fixed in 38da0c55e. Confirmed the 3-3 split you counted (`run_report`, `get_report_run_status`, `create_report_run_schedule` trim; `update_report`, `delete_report` did not), and for a tool whose only callers are agents the asymmetry is the wrong side to be on — a padded id 404s here while working against `get_report_run_status`. Added `.trim()`, plus a test pinning that `'  report-1  '` normalises and `'   '` is still rejected.

PR #1574, opened 2026-08-28
fix(web): open the Joinable Data Marts graph fitted to the whole diagram

laskevych:
## Problem

Switching the Joinable Data Marts card to the **Graph** view could open the canvas zoomed in on the root card (near maximum zoom) instead of showing the whole diagram. Pressing **Fit to view** fixed it, so users had to click it every time they opened the tab.

## Root cause

The automatic first fit called React Flow's imperative `fitView()`. In `@xyflow/react` v12, `fitView` only fits nodes whose **DOM-measured** dimensions are available (`getFitViewNodes` filters on `measured.width`), while nodes with declared `width`/`height` make the graph count as *initialized* before any measurement happens. The queued fit therefore executed against a half-measured subset of nodes — in practice just the root card — and centered on it at max zoom. The later "Fit to view" click worked because everything was measured by then.

## Fix

Derive the initial viewport analytically from the layout geometry (dagre positions + declared card sizes) with `getViewportForBounds` — the same inputs and math the zoom range already uses — and apply it with `setViewport`. The first fit no longer depends on DOM measurement timing, so the Graph view always opens fitted, exactly as the "Fit to view" button leaves it.

- `getFittedGraphViewport` added to `relationship-canvas-zoom.ts` (`getFittedGraphZoom` now delegates to it)
- the automatic fit in `RelationshipCanvas` applies that viewport (with a `fitView` fallback for degenerate geometry); the fitted viewport is read through a ref so cosmetic relayouts (view mode, direction, join labels) still keep the user's viewport
- the "Fit to view" button and the corrupted-viewport recovery are unchanged

## Testing

- new regression test: the first mount applies the layout-derived viewport and never calls `fitView`
- existing viewport suite updated to the new fit mechanism (36 tests passing)
- `tsc --noEmit`, ESLint and Prettier clean

### DIFF
--- a/apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
+++ b/apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
@@ -45,19 +45,6 @@ export function getFittedGraphViewport(
   );
 }
 
-/**
- * The zoom component of the fitted viewport, NaN for degenerate geometry so
- * the zoom range falls back — see getFittedGraphViewport.
- */
-export function getFittedGraphZoom(
-  bounds: CanvasGraphBounds,
-  paneWidth: number,
-  paneHeight: number,
-  padding: number
-): number {
-  return getFittedGraphViewport(bounds, paneWidth, paneHeight, padding)?.zoom ?? Number.NaN;
-}
-
 export function getGraphZoomRange(fittedZoom: number): GraphZoomRange {
   const safeFittedZoom =
     Number.isFinite(fittedZoom) && fittedZoom > 0
--- a/apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
+++ b/apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
@@ -1,4 +1,4 @@
-import { getViewportForBounds } from '@xyflow/react';
+import { getViewportForBounds, type Viewport } from '@xyflow/react';
 import type { CanvasGraphBounds } from '../../../shared/canvas/viewport';
 
 export const GRAPH_ZOOM_MAX = 3;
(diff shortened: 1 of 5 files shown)


apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
```diff
@@ -41,7 +42,20 @@ export function getFittedGraphZoom(
...
     padding
-  ).zoom;
+  );
+}
+
+/**
+ * The zoom component of the fitted viewport, NaN for degenerate geometry so
+ * the zoom range falls back — see getFittedGraphViewport.
+ */
+export function getFittedGraphZoom(
```
romandubovyi: ⚪ **Non-blocking:** After this refactor `getFittedGraphZoom` has no production caller left: the component derives the zoom inline with `getGraphZoomRange(fittedViewport?.zoom ?? Number.NaN)` (`RelationshipCanvas.tsx:903`), and the only remaining users are the tests in `relationship-canvas-zoom.test.ts`. That also duplicates the NaN fallback in two places. Either drop the wrapper and assert on `getFittedGraphViewport(...)?.zoom` in the test, or call it from the component so the fallback lives in one place.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
```diff
@@ -12,26 +12,27 @@ export interface GraphZoomRange {
...
- * — a mid-load fit, a pane that was still settling, or a layout that changed
- * after the fit ran.
+ * Derived analytically (instead of asking fitView) because an imperative
+ * fitView only fits nodes whose DOM dimensions are already measured, while
+ * declared node sizes make the graph count as initialized before measurement —
+ * a fit that runs at that point centers on whatever subset happens to be
+ * measured. The layout bounds are complete from the first render.
  */
-export function getFittedGraphZoom(
+export function getFittedGraphViewport(
```
romandubovyi, line 26: ⚪ **Non-blocking:** The sibling Models canvas still has the automatic first fit this PR replaces — `ModelCanvas.tsx:428` calls `reactFlow.fitView(...)` over nodes that likewise get their `width`/`height` from the dagre layout, and its only protection is a single `requestAnimationFrame`, which does not guarantee the `ResizeObserver` has delivered measurements yet. Did you check whether that canvas is exposed to the same regression? If so, `getFittedGraphViewport` may belong next to the already-shared `getCanvasGraphBounds`/`clampCanvasViewport` in `shared/canvas/viewport.ts` rather than staying relationship-only.

romandubovyi (approved):
Reviewed the layout-derived initial viewport. The mechanism checks out: `fitView` internally calls the same `getViewportForBounds` with identical math, and `setViewport` returns `Promise<boolean>` so the `.then()` chain is safe. Two non-blocking follow-ups on leftovers around the refactor; no blockers.

🤖 Generated with Claude Code (Opus 5) + Kimi CLI + Grok (Grok 4.6)

apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
laskevych: Done in a5c41c784 — dropped the wrapper; the tests now assert on `getFittedGraphViewport(...)?.zoom` and the NaN fallback lives only in the component.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/relationship-canvas-zoom.ts
laskevych: Good catch — the mechanism exists there too (imperative `fitView` under a single rAF plus declared node sizes). The difference is that ModelCanvas also passes the declarative `fitView` prop, which the library resolves via `updateNodeInternals` after measurements arrive, and that appears to mask the race on first mount — we have not observed the regression there. Spun off a follow-up to verify it live and move `getFittedGraphViewport` next to the shared viewport helpers if it reproduces; keeping this PR relationship-only for now.

PR #1566, opened 2026-08-25
feat(data-marts): calculated fields in the output schema

zapolsky:
## Summary

An analyst can define a **Calculated Field** directly in a Data Mart's Output Schema — a formula written in that Data Mart's own warehouse SQL, e.g. `SUM(clicks) * 1.0 / NULLIF(SUM(impressions), 0)`. Unlike a ratio hard-coded into a Data Mart's SQL, it recomputes at whatever grain the request asks for, so a report broken down by day, a Google Sheet, a Looker Studio page, an MCP query, an HTTP Data stream and a report's Totals row all get the true ratio of the sums rather than an average of per-row ratios.

Shipped on all five storages (BigQuery, Athena, Snowflake, Redshift, Databricks). The matching report-picker guards for the Google Sheets add-on are ready in the `google-sheets-extension` repo and will follow as a separate PR.

## What's included

- **Level derived from the formula, never chosen** — one that aggregates is a metric, one that does not is a dimension, and the two behave differently at grouping, in Totals, and in what Looker Studio is told. The derivation is transitive: a formula reading another formula inherits its level.
- **Formula editor** — autocomplete over the Data Mart's own fields, every joined Data Mart's fields, the dialect's aggregate and scalar functions, `CASE`, and a guarded-division snippet. Resolved references render as atomic chips; live validation reports problems a moment after typing without touching the warehouse, and never blocks Apply.
- **Save-time validation** — parse, resolve references against the schema being saved, then dry-run the composed query against the warehouse once per save. Every refusal names the field it belongs to.
- **A formula is a single expression** — `;`, a top-level `,`, and the two comment markers the warehouses contradict each other about (`#`, `//`) are refused rather than interpreted.
- **Reports** — a calculated field can be selected, filtered, sorted, aggregated (row-level only) and date-bucketed, on plain, joined and blended reports, including the Totals row and the kept-groups restriction.
- **Joined Data Marts** — a formula may read a joined mart's field inside an aggregation. A *joined* Data Mart's own calculated field is refused on every report surface, and the pickers no longer offer one.
- **Surfaces** — REST, Looker Studio, Google Sheets, MCP (`get_data_mart_details`, `query_data_mart`) and HTTP Data all carry the field and its header.
- **Docs** — `docs/getting-started/setup-guide/calculated-fields.md`, including per-dialect aggregate functions and the limitations below.

## Migrations

[...]

PR #1575, opened 2026-08-28
fix(data-marts): calculated field improvements

zapolsky:
## Summary

Six fixes to the **Calculated Field** editor, from feedback gathered in the first days of use of #6732. One branch, one commit, one changeset.

Fibery: [#6876 Calculated field improvments](https://owox.fibery.io/Sprint_Work/Work_Item/Calculated-field-improvments-6876)

## What's included

- **Unsaved edits survive a background refresh.** The Output Schema editor reset itself whenever `initialSchema` arrived as a new object, whether or not its content had changed — and every re-read of the Data Mart produces such an object. A formula typed but not yet saved went with it, while the failed save's errors stayed on screen naming a field no longer in the table. It compares by content now.
- **A row-level field keeps its "Σ available" control while its formula is edited.** Applying a formula replaced `calculated` wholesale, dropping the level the previous save derived; an absent level reads as aggregate-level, and an aggregate has no aggregations of its own. The level rides along now. The warehouse stamp is still dropped — it must not outlive the formula it judged.
- **A calculated field no longer stretches its row.** A formula authored over several lines made its row that many lines tall. The row clamps to two lines; hovering shows the whole formula with the author's own line breaks.
- **"Save & leave" names the rejected formula** instead of reporting `Request failed with status code 400`, which named neither the field nor the fix while the field-grouped detail sat on screen behind the dialog.
- **The unguarded-division warning is about the formula in front of you.** It quoted `NULLIF(SUM(impressions), 0)` whether or not `impressions` appeared anywhere in the formula, and read like a refusal although it refuses nothing. It quotes the denominator this formula actually divides by, underlines it, and says it is advice.
- **`COALESCE` no longer counts as a guard against that division.** What fails a query is a zero denominator, not a null one, and `COALESCE(SUM(x), 0)` produces exactly that zero.

Docs gain a worked example for a **rate over a condition** (activation rate, share of paid accounts, fill rate), asked for so agents have one to follow.

## The report that started it

An analyst reported a calculated field disappearing after a failed save. It had not been lost. The load balancer's own log shows exactly **six** `PUT /schema` calls on that Data Mart in 48 hours — five 400s and then a 200 — and every request body is within ~20 bytes of the others, so no payload ever lost a field. The field is in the Data Mart today with the formula that last save carried.

The defect was in the editor, not in the data, and that is the first item above.

## Measured, not assumed

- **Two lines are free.** A calculated row is **49px** whether its formula occupies one 16px line or two — the height comes from the controls beside it. The third line takes the row to 64px.
- **Not `whitespace-nowrap`.** It also bounds the height, but an unbreakable formula grew its column from **549px to 1042px** at the other columns' expense on this table's own widths.
- **Zero, not null.** On BigQuery `1 / CAST(NULL AS INT64)` returns NULL and raises nothing, while `1 / 0` and `1 / COALESCE(CAST(NULL AS INT64), 0)` both fail.

## What the warning quotes, and why it matters

The subject is now the **whole** denominator or nothing at all. Naming a fragment was worse than naming none: an analyst who wrapped what the message quoted got `revenue / (NULLIF(cost, 0) + tax)` — still fatal when `cost + tax` is zero, and now **silent**, because the guard sits exactly where the check looks.

| Formula | Before | After |
| --- | --- | --- |
| `revenue / (cost + tax)` | `cost` | `(cost + tax)` |
| `SUM(a) / CASE … END` | `CASE` | warns, names nothing |
| `SUM(a) / -1` | `-` | silent (a constant) |
| `SUM(a) / SUM(b` | `SUM` | warns, names nothing |

## Shared components

`EditableText` reports when its editor **closes**, so a cell can stand its hover card down while the editor's own popover is open. `HoverCardContent` gets the `data-slot` its sibling parts already carry.

## Testing

| | |
| --- | --- |
| `nest build` · `tsc -b` (web) | pass |
| backend `jest src/data-marts` | 434 suites, 7727 tests |
| web `vitest` | 245 files, 2295 tests |
| eslint (web, e2e, backend) · markdownlint | clean |

## For the reviewer

- The cell no longer carries the formula as its `title` — a native tooltip beside the hover card is a second, worse telling of the same thing. Nine Playwright locators moved to the slot the card puts there (`e2e/helpers/formula-cell.ts`), and the DSET-10 wrapping spec was rewritten against the new contract. **Playwright was not run locally**, so its first CI run is the real check on that migration.
- Both formula triggers take focus, so the card is reachable without a pointer. The clamp is visual only: the whole formula stays in the DOM, which is what a screen reader reads.
- Deliberately **not** in this PR, each worth its own change: `SAFE_DIVIDE` / `IFF` / `DIV0` still sit in the guard set although none of them guards a denominator (pre-existing, and a behaviour change with its own changeset); the warning's subject is whitespace-collapsed, which can stop the editor placing a marker on a multi-line denominator; and the hover card can be left open behind a suppressed state.

### DIFF
--- a/apps/backend/src/data-marts/calculated-fields/formula-analyzer.ts
+++ b/apps/backend/src/data-marts/calculated-fields/formula-analyzer.ts
@@ -291,8 +291,9 @@ export function analyzeFormula(input: AnalyzeFormulaInput): FormulaAnalysis {
     }
   }
 
-  if (hasUnguardedDivision(tokens, calls)) {
-    warnings.push(FormulaViolations.unguardedDivision(fieldName));
+  const division = unguardedDenominator(formula, tokens, calls, references);
+  if (division) {
+    warnings.push(FormulaViolations.unguardedDivision(fieldName, division.subject));
   }
 
   return {
@@ -365,32 +366,111 @@ function dedupeViolations(violations: FormulaViolation[]): FormulaViolation[] {
 }
 
 /**
- * A `/` whose right-hand side is neither a numeric literal nor already wrapped in a null-guarding
- * call. Deliberately shallow — an advisory, not a correctness gate. Known gaps, left for a human:
- * a guard around the RESULT of the division still warns, because it does not guard the denominator;
- * IF and CASE guards are not recognised; the guard-function set is fixed rather than per dialect.
+ * A `/` whose right-hand side is neither a numeric literal nor already wrapped in a call that turns
+ * a ZERO denominator into something harmless.
+ *
+ * Zero, not null. Measured on BigQuery: `1 / CAST(NULL AS INT64)` is NULL and raises nothing, while
+ * `1 / 0` fails the statement. So COALESCE is NOT in the guard set — it replaces a null, which was
+ * never the danger, and `COALESCE(SUM(x), 0)` actively PRODUCES the zero that fails the query
+ * (verified: `1 / COALESCE(CAST(NULL AS INT64), 0)` raises `division by zero`). Accepting it meant
+ * agreeing with an analyst who believed they had guarded.
+ *
+ * Deliberately shallow — an advisory, not a correctness gate. Known gaps, all erring toward warning
+ * rather than staying silent: a guard around the RESULT of the division still warns, because it does
+ * not guard the denominator; a guard NESTED inside the denominator (`COALESCE(NULLIF(x, 0), 1)`)
+ * warns too; IF and CASE guards are not recognised; the guard set is fixed rather than per dialect.
  */
-function hasUnguardedDivision(
+interface UnguardedDivision {
+  /** The span to quote, or absent when nothing in the formula stands for the whole denominator. */
+  subject?: string;
+}
+
+function unguardedDenominator(
+  formula: string,
   tokens: readonly SqlToken[],
-  calls: readonly SqlFunctionCall[]
-): boolean {
-  const GUARDS = new Set(['NULLIF', 'SAFE_DIVIDE', 'IFF', 'COALESCE', 'NULLIFZERO', 'DIV0']);
+  calls: readonly SqlFunctionCall[],
+  references: readonly FormulaReference[]
+): UnguardedDivision | undefined {
+  const GUARDS = new Set(['NULLIF', 'SAFE_DIVIDE', 'IFF', 'NULLIFZERO', 'DIV0']);
   const code = tokens.filter(t => t.kind !== 'comment');
-  return code.some((token, i) => {
-    if (!(token.kind === 'punct' && token.value === '/')) return false;
+  for (const [i, token] of code.entries()) {
+    if (!(token.kind === 'punct' && token.value === '/')) continue;
     // Past any parens the analyst wrapped the denominator in, so `a / (NULLIF(b, 0))` reads the
-    // same as `a / NULLIF(b, 0)`.
+    // same as `a / NULLIF(b, 0)` — and past a unary sign, so `a / -1` reads like `a / 1` and is
+    // dropped as the constant it is.
     let j = i + 1;
-    while (code[j]?.kind === 'punct' && code[j].value === '(') j++;
+    let group: SqlToken | undefined;
+    while (code[j]?.kind === 'punct' && ['(', '-', '+'].includes(code[j].value)) {
+      if (code[j].value === '(') group ??= code[j];
+      j++;
+    }
     const denominator = code[j];
-    if (!denominator) return false;
-    if (denominator.kind === 'number') return false;
+    if (!denominator) continue;
+    if (denominator.kind === 'number') continue;
     // The denominator must BE the guard call. Asking only that it fall somewhere INSIDE one also
     // accepts a guard wrapped around the whole quotient — the case the policy above says still
     // warns, because a guard on the result does nothing about dividing by zero.
-    const guarded = calls.some(
-      c => GUARDS.has(c.name.toUpperCase()) && c.nameStart === denominator.start
-    );
-    return !guarded;
-  });
+    const call = calls.find(c => c.nameStart === denominator.start);
+    if (call && GUARDS.has(call.name.toUpperCase())) continue;
+
+    // What to NAME is a separate question from whether to warn, and naming the wrong thing is
+    // worse than naming nothing: an analyst who wraps what the message quotes in `NULLIF` ends up
+    // with a formula that still divides by zero AND no longer warns, because the guard now sits
+    // where this scan looks. So the subject is only ever a span that stands for the whole
+    // denominator.
+    if (group) {
+      // A parenthesised denominator is that whole group: `a / (cost + tax)` is guarded by wrapping
+      // the SUM, never by wrapping `cost`.
+      const close = matchingParen(code, group);
+      return { subject: close && authoringText(formula, references, group.start, close.end) };
+    }
+    // A `{{ref}}` tag is punctuation to the scanner, so a bare reference as the denominator starts
+    // at its `{` and would be quoted as one brace. Its own span is the thing to name.
+    const ref = references.find(r => r.start <= denominator.start && denominator.start < r.end);
+    if (ref) return { subject: authoringText(formula, references, ref.start, ref.end) };
+    if (call?.closed) {
+      return { subject: authoringText(formula, references, denominator.start, call.argEnd + 1) };
+    }
+    if (denominator.kind === 'quotedIdentifier') {
+      return { subject: authoringText(formula, references, denominator.start, denominator.end) };
+    }
+    // A bare word (`CASE`), an unclosed call, a stray operator: the token is a fragment of the
+    // denominator rather than the denominator, so the warning stands without pointing at it.
+    return {};
+  }
+  return undefined;
+}
+
+/** The `)` closing `open`, or undefined if the formula never closes it. */
+function matchingParen(code: readonly SqlToken[], open: SqlToken): SqlToken | undefined {
+  let depth = 0;
+  for (const token of code) {
+    if (token.start < open.start || token.kind !== 'punct') continue;
+    if (token.value === '(') depth++;
+    else if (token.value === ')' && --depth === 0) return token;
+  }
+  return undefined;
+}
+
+/**
+ * A slice of the STORED formula written the way the analyst sees it — every `{{ref}}` tag inside it
+ * replaced by the name it stands for.
+ *
+ * The analyzer only ever holds the stored form, and a message quoting that verbatim would show a tag
+ * the editor deliberately never displays.
+ */
+function authoringText(
+  formula: string,
+  references: readonly FormulaReference[],
+  start: number,
+  end: number
+): string {
+  const inside = references
+    .filter(r => start <= r.start && r.end <= end)
+    .sort((a, b) => b.start - a.start);
+  let text = formula.slice(start, end);
+  for (const ref of inside) {
+    text = text.slice(0, ref.start - start) + refLabel(ref) + text.slice(ref.end - start);
+  }
+  return text.replace(/\s+/g, ' ').trim();
 }
--- a/apps/backend/src/data-marts/calculated-fields/formula-analyzer.ts
+++ b/apps/backend/src/data-marts/calculated-fields/formula-analyzer.ts
@@ -1,6 +1,7 @@
 import {
   FormulaReference,
   parseFormulaReferences,
+  renderFormula,
   FormulaReferenceSyntaxError,
 } from './formula-reference';
 import { scanSql, SqlToken } from './sql-token-scanner';
@@ -393,11 +394,42 @@ function unguardedDenominator(
 ): UnguardedDivision | undefined {
   const GUARDS = new Set(['NULLIF', 'SAFE_DIVIDE', 'IFF', 'NULLIFZERO', 'DIV0']);
   const code = tokens.filter(t => t.kind !== 'comment');
+
+  /**
+   * The span, written as the analyst sees it — unless it cannot be quoted honestly.
+   *
+   * A comment inside it survives the slice and the whitespace collapse folds the newline that ends
+   * it, so `NULLIF((b -- note + 1), 0)` is what the advice would say: broken SQL when pasted. A
+   * quoted identifier arrives with its own delimiters, which the message's backticks would then
+   * double — unreadable to the web's leading-backtick parser, and unmarkable in an editor that
+   * masks quoted spans before searching. Both take the unnamed message, which reads fine.
+   */
+  const named = (from: number, to: number): UnguardedDivision => {
+    // A `{{ref}}` tag's own `"field"` scans as a quoted identifier, so tokens inside a tag are
+    // excluded — they are the tag's syntax, not text the analyst wrote.
+    const insideTag = (t: SqlToken) => references.some(r => r.start <= t.start && t.end <= r.end);
+    const unquotable = tokens.some(
+      t =>
+        t.start >= from &&
+        t.end <= to &&
+        !insideTag(t) &&
+        (t.kind === 'comment' || t.kind === 'quotedIdentifier')
+    );
+    if (unquotable) return {};
+    try {
+      return { subject: authoringText(formula, from, to) };
+    } catch {
+      // A span holding half a `{{ref}}` tag cannot be rendered. Not reachable from the spans below,
+      // which are all tag-aligned, but this runs inside the editor's live check: a throw here would
+      // silence every diagnostic rather than lose one subject.
+      return {};
+    }
+  };
+
   for (const [i, token] of code.entries()) {
     if (!(token.kind === 'punct' && token.value === '/')) continue;
     // Past any parens the analyst wrapped the denominator in, so `a / (NULLIF(b, 0))` reads the
-    // same as `a / NULLIF(b, 0)` — and past a unary sign, so `a / -1` reads like `a / 1` and is
-    // dropped as the constant it is.
+    // same as `a / NULLIF(b, 0)` — and past a unary sign, so `a / -1` reads like `a / 1`.
     let j = i + 1;
     let group: SqlToken | undefined;
     while (code[j]?.kind === 'punct' && ['(', '-', '+'].includes(code[j].value)) {
@@ -406,7 +438,13 @@ function unguardedDenominator(
     }
     const denominator = code[j];
     if (!denominator) continue;
-    if (denominator.kind === 'number') continue;
+    const close = group && matchingParen(code, group);
+    // A constant cannot come out zero unless it IS zero. Inside a group the whole group has to be
+    // constant to say that: `(2)` is, `(1 - rate)` is not, and testing only the first token called
+    // both of them constants.
+    if (group ? close && isConstantGroup(code, group, close) : denominator.kind === 'number') {
+      continue;
+    }
     // The denominator must BE the guard call. Asking only that it fall somewhere INSIDE one also
     // accepts a guard wrapped around the whole quotient — the case the policy above says still
     // warns, because a guard on the result does nothing about dividing by zero.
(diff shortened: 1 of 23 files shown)


apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/hooks/useSchemaState.ts
```diff
@@ -63,6 +78,9 @@ export function useSchemaState(initialSchema: DataMartSchema | null | undefined)
    */
   const markSchemaSaved = useCallback((savedSchema: DataMartSchema) => {
     skipNextInitialSchemaResetRef.current = true;
+    // What is on screen from here on, so the next initialSchema that DIFFERS from it still resets
+    // — including the server's own copy of this save, which carries the level it derived.
+    appliedInitialSchemaRef.current = fingerprint(savedSchema);
```
laskevych, line 87: 🟡 **Important (non-blocking):** after a guarded save this pins a fingerprint the server will never reproduce — the client copy lacks the derived `level`, the `warehouseValidation` stamp, and the canonicalized formula text, all of which the response carries. With the skipped republish deliberately unrecorded, the *first* later republish always fingerprint-differs and hard-resets, discarding anything typed since the guarded save; until it lands, the saved row also misleads (a level-less field reads as aggregate, and the reference popover still claims the schema isn't saved). The `keepUnsavedEdits` path shares this skip branch and records nothing at all, so an edit applied over an in-flight save survives republish #1 and is wiped by republish #2.

`updateDataMartSchema` in `DataMartContext` already maps the full response and returns only `{ warnings }` — returning the mapped schema and feeding `markSchemaSaved` the **server** copy would land the level immediately, make the fingerprint match every later republish, and let the skip-ref dance shrink. Fine as a follow-up (the `keepUnsavedEdits` case needs the same routing), but worth capturing before the two-ref exception hardens.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/calculated/CalculatedFieldFormulaCell.tsx
```diff
@@ -105,20 +106,77 @@ function LiveCheckedFormulaEditor({
...
+  children,
+}: {
+  text: string;
+  references: readonly ResolvedReference[];
+  describeReference: (reference: ResolvedReference) => string | undefined;
+  suppressed: boolean;
+  children: ReactElement;
+}) {
+  const [open, setOpen] = useState(false);
+  if (text.trim() === '') return children;
```
laskevych: 🟡 **Important (non-blocking):** applying the *first* formula of a newly added field flips this component's output from bare `children` to the `HoverCard` wrapper in the same batched commit that closes the editor. React replaces the subtree (different element type at the same position), the old `EditableText` unmounts before its `isEditing` edge-effect can run, and `onEditEnd` is never delivered — so the cell's `isEditing` sticks at `true` and the hover card stays suppressed for that row until the editor is opened and closed once more. It fires exactly once per new field (an empty Apply is refused, so the reverse flip can't happen).

Rendering the `HoverCard` wrapper unconditionally and gating only `open` avoids the branch flip; reporting edit-end synchronously from the Apply/Cancel handlers instead of the unmount-prone effect fixes it at the source (both close routes are named handlers in `editable-text.tsx`, so the effect-based edge detector could go entirely).

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/calculated/CalculatedFieldFormulaCell.tsx
```diff
@@ -105,20 +106,77 @@ function LiveCheckedFormulaEditor({
...
+  return (
+    <HoverCard open={open && !suppressed} onOpenChange={setOpen} openDelay={300} closeDelay={100}>
```
laskevych: 🟡 **Important (non-blocking):** closing the editor re-opens this card uninvited on mouse flows. The popover's default close-autofocus returns focus to the cell's `role='button'` trigger on Apply/Cancel/Escape (outside-click is the one close path that doesn't), the focus bubbles up to `HoverCardTrigger`, and the installed `@radix-ui/react-hover-card@1.1.23` opens on *any* focus — `onFocus: composeEventHandlers(props.onFocus, context.onOpen)`, no focus-visible gate — so ~300 ms after every mouse-driven Apply the 520px card pops over the rows below with no visible cause (programmatic focus draws no ring).

This is a different mechanism from the acknowledged “card left open behind a suppressed state” — and that acknowledged case is stickier than it looks: while `open && !suppressed` renders false, Radix drops close attempts (next value === prop), so a stale `open=true` reappears instantly on unsuppress, no focus needed. Holding `suppressed` until the trigger's next blur/pointerenter covers both; an `onCloseAutoFocus` override would too, at the cost of the keyboard return-to-cell behavior.

apps/backend/src/data-marts/calculated-fields/formula-analyzer.ts
```diff
@@ -365,32 +366,111 @@ function dedupeViolations(violations: FormulaViolation[]): FormulaViolation[] {
...
+    let group: SqlToken | undefined;
+    while (code[j]?.kind === 'punct' && ['(', '-', '+'].includes(code[j].value)) {
+      if (code[j].value === '(') group ??= code[j];
+      j++;
+    }
     const denominator = code[j];
-    if (!denominator) return false;
-    if (denominator.kind === 'number') return false;
+    if (!denominator) continue;
+    if (denominator.kind === 'number') continue;
```
laskevych: 🟡 **Important (non-blocking):** once the loop above has stepped inside a parenthesized denominator, this numeric-literal test is applied to the group's *first token* rather than to the group. `SUM({{ref a}}) / (-1 + {{ref b}})` is now silent — the pre-PR scan warned here (its `-` token was neither a number nor a guard) — and `SUM({{ref a}}) / (1 - {{ref rate}})` stays silent too (pre-existing, but it now contradicts this function's own “all erring toward warning” policy, and `x / (1 - rate)` is a common shape).

Suggest skipping the number short-circuit whenever `group` is set unless every code token inside the group is a number or sign — `a / (2)` then stays silent, while `a / (1 - rate)` warns and names `(1 - rate)`.

apps/backend/src/data-marts/calculated-fields/formula-analyzer.ts
```diff
@@ -365,32 +366,111 @@ function dedupeViolations(violations: FormulaViolation[]): FormulaViolation[] {
...
+      return { subject: close && authoringText(formula, references, group.start, close.end) };
+    }
+    // A `{{ref}}` tag is punctuation to the scanner, so a bare reference as the denominator starts
+    // at its `{` and would be quoted as one brace. Its own span is the thing to name.
+    const ref = references.find(r => r.start <= denominator.start && denominator.start < r.end);
+    if (ref) return { subject: authoringText(formula, references, ref.start, ref.end) };
+    if (call?.closed) {
+      return { subject: authoringText(formula, references, denominator.start, call.argEnd + 1) };
+    }
+    if (denominator.kind === 'quotedIdentifier') {
```
laskevych: 💡 **Suggestion (non-blocking):** two subject-integrity edges in these branches. (1) A backtick-quoted identifier denominator reaches this branch with its delimiters included, so the message wraps an already-backticked subject in a second pair — the leading-backtick fallback on the web can't parse that, and no marker is possible either (the editor masks quoted spans before searching). (2) A `--` line comment inside a parenthesized denominator survives the slice, and the whitespace collapse folds its terminating newline: `a / ({{ref b}} -- note` + newline + `+ 1)` yields the advice `NULLIF((b -- note + 1), 0)`, which is broken SQL when pasted.

Both fit the file's own “naming the wrong thing is worse than naming nothing” rule: returning no subject when the span contains a comment or quoted-identifier token keeps the warning and drops the misquote — the unnamed message already reads well.

apps/backend/src/data-marts/calculated-fields/formula-analyzer.ts
```diff
@@ -365,32 +366,111 @@ function dedupeViolations(violations: FormulaViolation[]): FormulaViolation[] {
...
+}
+
+/**
+ * A slice of the STORED formula written the way the analyst sees it — every `{{ref}}` tag inside it
+ * replaced by the name it stands for.
+ *
+ * The analyzer only ever holds the stored form, and a message quoting that verbatim would show a tag
+ * the editor deliberately never displays.
+ */
+function authoringText(
```
laskevych: 💡 **Suggestion (non-blocking):** this re-implements the stored→authoring substitution that `renderFormula` (formula-reference.ts) already owns — it re-parses tags itself, and every span chosen above contains only whole tags, so `renderFormula(formula.slice(start, end), refLabel)` plus the whitespace collapse does the same job without a second splice loop for the two renderers to drift apart on.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/tables/BaseSchemaTable.tsx
```diff
@@ -258,10 +258,15 @@ export function BaseSchemaTable<T extends BaseSchemaField>({
...
-                  calculated: { formula },
+                  // `calculated` is replaced wholesale, so `warehouseValidation` goes: a stamp
+                  // must not outlive the formula it judged.
+                  //
+                  // The LEVEL is carried instead. It decides this row's shape, and an absent one
+                  // reads as aggregate — which took the "Σ available" control off a dimension the
+                  // moment its formula was touched, and put it back only after a save. Safe to
+                  // keep, because the save overwrites `calculated.level` from its own analysis
+                  // (CalculatedFieldValidatorService) rather than trusting what it was sent.
+                  calculated: { formula, level: row.original.calculated?.level },
```
laskevych, line 269: 💡 **Suggestion (non-blocking):** the tradeoff comment is accurate and the persisted side is safe (the validator always overwrites `level`), but two UI gaps stay visible. A dimension edited *into* an aggregate keeps a live Σ control whose selections the save silently orphans — nothing cross-checks them against the re-derived level, and they're refused only at report time. And a field authored this session still shows no Σ control until a full save round-trip — the mirror of the bug fixed here.

The live check already computes the draft's real level (`CalculatedFieldValidatorService` derives it for every probe) and drops it before the response. Returning it and driving the row's shape from the draft's own level — stored level as the fallback when the live channel is off — would close both. Feels like the natural follow-up to this fix rather than more of this PR.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/calculated/useCalculatedFieldSave.ts
```diff
@@ -57,6 +57,19 @@ function groupByField(violations: ApiFormulaViolation[] | undefined): Violations
...
+ * The one-line summary of a rejected save, for a surface that can only render a string — today the
+ * unsaved-changes dialog, which otherwise shows axios's own "Request failed with status code 400".
+ * Returns undefined when the rejection carries no formula violations, so the caller keeps its own
+ * wording for an unrelated failure.
+ */
+export function rejectedFormulaMessage(error: unknown): string | undefined {
+  const violations = extractApiError(error).errorDetails?.errors;
+  const fields = Object.keys(groupByField(violations));
+  if (fields.length === 0) return undefined;
+  return `The warehouse rejected ${fields.join(', ')}. Close this dialog to see why.`;
```
laskevych: 🟢 **Nitpick (non-blocking):** “The warehouse rejected …” also covers violations the static analyzer refused before any warehouse was called — the fixture in the new test attributes `FORMULA_UNKNOWN_REFERENCE` to the warehouse, and the same wording would show when the storage isn't configured at all. Something like “The save was rejected for …” stays truthful for both.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/DataMartSchemaSettings.tsx
```diff
@@ -371,7 +372,14 @@ export function DataMartSchemaSettings({ definitionType }: DataMartSchemaSetting
...
     const current = schemaRef.current;
     if (dataMartId && current) {
-      await saveCalculatedFields(current);
+      try {
+        await saveCalculatedFields(current);
+      } catch (error) {
+        // The dialog can only render a string, and an axios rejection's own `message` is
+        // "Request failed with status code 400" — which names neither the field nor the fix. The
+        // field-grouped detail is already on screen behind the dialog; point at it.
+        throw new Error(rejectedFormulaMessage(error) ?? 'Failed to save changes');
```
laskevych: 💡 **Suggestion (non-blocking):** for a rejection carrying no formula violations this falls straight to the hardcoded string, skipping the `extractApiError(error).message ?? fallback` idiom that helper's own doc describes as the established chain — a non-formula 400 whose body carries a meaningful message would show “Failed to save changes”. `rejectedFormulaMessage(error) ?? extractApiError(error).message ?? 'Failed to save changes'` keeps it.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/hooks/useSchemaState.ts
```diff
@@ -44,13 +48,24 @@ export function useSchemaState(initialSchema: DataMartSchema | null | undefined)
   const [schema, setSchema] = useState<DataMartSchema | null | undefined>(clonedInitialSchema);
   const [isDirty, setIsDirty] = useState(false);
   const skipNextInitialSchemaResetRef = useRef(false);
+  const appliedInitialSchemaRef = useRef(fingerprint(clonedInitialSchema));
```
laskevych: 💡 **Suggestion (non-blocking):** `useRef`'s initializer argument isn't lazy, so this stringifies the whole schema on every render and discards the result after the first — on top of the per-render deep clone above. Lazy-init (`useRef<string | null>(null)`, fill on first use) keeps it to one; and since the clone and the fingerprint are the same string, the reset effect and `markSchemaSaved` could `JSON.parse` the fingerprint they just computed instead of serializing the schema twice.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/calculated/CalculatedFieldFormulaCell.tsx
```diff
@@ -105,20 +106,77 @@ function LiveCheckedFormulaEditor({
 }
 
 /**
- * Plain text on the row's own background, wrapping rather than truncating.
+ * Plain text on the row's own background, ONE line, truncated.
```
laskevych: 🟢 **Nitpick:** “ONE line, truncated” — the constant wraps and clamps to *two* lines (`line-clamp-2`, `whitespace-normal`, nothing truncated), as the rest of this very comment says. Leftover from the earlier design.

apps/web/src/features/data-marts/edit/components/DataMartSchemaSettings/calculated/CalculatedFieldFormulaCell.test.tsx
```diff
@@ -158,9 +158,27 @@ function renderCell({
  * its own `<span>` chip now, so the formula is several text nodes and `getByText` matches none.
  */
 function formulaRow(text: string): HTMLElement {
```
laskevych, line 180: 🟢 **Nitpick (non-blocking):** `formulaRow(text)` no longer uses `text` (the first trigger in the document wins), yet the call sites still pass formulas and the JSDoc above still says “Found by its title”. A parameterless `formulaRow()` that throws when no trigger exists says what it does — and `hoverRow` can reuse it instead of re-querying the selector.

apps/web/e2e/specs/datamart-calculated-field-formula.spec.ts
```diff
@@ -90,7 +91,7 @@ test.describe('Data Setup - Calculated field formula autocomplete', () => {
     await expect(page.getByTestId(TESTIDS.datamartTabDataSetup)).toBeVisible();
 
     // The formula cell of the `roas` row — the whole cell carries the formula as its title.
```
laskevych: 🟢 **Nitpick:** this comment still says the cell “carries the formula as its title” — the contract this PR removed; the next line already locates by the hover-card slot.

docs/getting-started/setup-guide/calculated-fields.md
```diff
@@ -100,6 +100,28 @@ SUM(clicks) * 1.0 / NULLIF(SUM(impressions), 0)
...
+```text
+SUM(CASE WHEN firstUsageDateTime IS NOT NULL THEN 1 ELSE 0 END) * 1.0
+  / NULLIF(SUM(CASE WHEN firstLogInDateTime IS NOT NULL THEN 1 ELSE 0 END), 0)
+```
+
+Three things this shape depends on:
+
+- **`NULLIF(..., 0)`** — a period where nobody logged in gives an empty cell instead of failing the whole report with a division by zero.
+- **`* 1.0`** — both counts are integers, and on Redshift and Athena an integer divided by an integer truncates to `0`. See [Dividing one integer by another](#dividing-one-integer-by-another).
+- **Two arguments to `IFNULL`.** On BigQuery `IFNULL(x, 0, 1)` is not a shorter `IF` — `IFNULL` accepts exactly two. The editor's live check does not run the query, so a wrong-arity call is caught by the save's warehouse test run, not while you type. On BigQuery, `SAFE_DIVIDE(a, b)` is the guarded division in one call and needs no `NULLIF`.
```
laskevych: 🔵 **Question:** the example above uses `NULLIF`/`COUNTIF`, but this third “thing the shape depends on” is about `IFNULL` arity — a function the shape never uses. Is it guarding against the NULLIF↔IFNULL mix-up (`IFNULL(x, 0)` being precisely a zero-producer, like `COALESCE`)? If so, one connecting clause would land it; as written it reads like it wandered in from a different section.

laskevych (approved):
Reviewed with a deep pass over the diff and the surrounding flows (schema state lifecycle, the analyzer/scanner pair, the Radix primitives as installed, and the e2e migration). The six fixes do what the description says, the measurements in the description held up against the code, and the changeset and docs are accurate.

**Approving — nothing below blocks.** The two Important edit-loss items are pre-existing side channels this PR narrows (before it, they were the *only* behavior), the hover-card items are contained UX defects in the new affordance with cheap escapes, and the rest are edges in advisory paths or polish. Inline comments carry the details; two findings land here because their lines are outside the diff.

---

🟡 **Important (non-blocking, pre-existing path): the `'success'` effect can still wipe edits after the first successful save.** `DataMartSchemaSettings.tsx:298-302` — `operationStatus` is sticky (`'success'` until the next save starts; `resetOperationStatus` currently has zero callers), and `resetSchema` gets a new identity on every republish of the Data Mart, so after the first successful save every later republish re-fires the effect and `resetSchema()` clones `initialSchema` over the live state — bypassing the new fingerprint compare one hook over. The save itself opens the window: `handleSave` chains `runSchemaActualization()`, whose completion republishes the mart, so a formula typed while that sync is still running disappears silently with `isDirty` cleared. Before this PR the unconditional identity-reset made this effect redundant; now it is the one remaining wipe path, so the headline “unsaved edits survive a background refresh” holds only until the first successful save of the session. Consuming the status after the reset (the unused `resetOperationStatus()`) or keying the effect on a one-shot transition closes it — worth doing while this code is warm.

🟡 **Important (non-blocking): a bare-reference subject underlines the numerator too.** `formula-violation-ranges.ts:103` (not in this diff, but it places the new `subject`): for a subject that is a bare reference, the reference branch marks *every* same-named ref — right for name-validity violations, wrong for this positional one. `({{ref revenue}} - {{ref cost}}) / {{ref revenue}}` underlines both `revenue`s, and an analyst who wraps the underlined numerator is left wondering why the warning persists. The text-search branch below already refuses ambiguous multi-occurrence subjects; the reference branch could apply the same single-occurrence rule when the violation is the division warning.

📝 **CI:** the Browser E2E failure is `datamart-calculated-field-chips.spec.ts:482` (“keeps a chip readable under the error the backend blames it for”) — a test this PR does not touch, timing out on diagnostics after `page.keyboard.type(' + clicks')`. That is the known editor keystroke-loss issue deferred from #1566, with the same signature as on that PR's run. The nine migrated locators all passed before that point, so the migration this run was meant to check looks validated; the other ✘ (sheet-confirmation) passed on retry.

💡 **Minor:** the new `formulaCell` unit-test helper is pasted verbatim into `BaseSchemaTable.test.tsx` and `BigQuerySchemaTable.test.tsx` (plus a near-variant inline at the ×1-count assertion). This PR just paid for exactly that kind of drift once, with `title`; one shared helper — like the e2e one — keeps the next migration to one edit.

PR #1576, opened 2026-08-28
feat(data-marts): add explicit UTC timezone to scheduled triggers

max-voloshyn:
## Summary

- Add one explicit `UTC` option to scheduled-trigger timezone selectors.
- Label UTC as `UTC (+00:00, no DST)` and guarantee a fixed zero offset.
- Filter equivalent UTC/GMT aliases from the picker to avoid duplicate choices.
- Display the compatible `Europe/Kiev` identifier as `Europe/Kyiv` without changing its stored or serialized value.
- Preserve existing API, persistence, cron parsing, and scheduler contracts.

## Testing

- Added frontend coverage for:
  - UTC selection
  - canonical API serialization
  - schedule display and preview text
  - `Europe/Kiev` → `Europe/Kyiv` presentation
- Added backend coverage confirming UTC next-run calculations remain fixed in winter and summer.
- Full web suite: 2,309 tests passed.
- Full backend suite: 9,140 tests passed.
- Web/backend builds and lint passed.
- Web type-check, Prettier, and `git diff --check` passed.

## Compatibility

No database migration or backend production change is required. Scheduled triggers already accept and persist the canonical `UTC` value. Existing timezone identifiers remain unchanged in stored schedules and API payloads.

## Release note

- Added a non-Fibery changeset for the user-visible scheduled-trigger behavior.

🤖 Codex (GPT-5.6 Sol High)

### DIFF
--- a/.changeset/explicit-utc-scheduled-triggers.md
+++ b/.changeset/explicit-utc-scheduled-triggers.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Schedule Data Mart triggers in UTC
+
+Scheduled triggers now offer UTC as an explicit fixed-offset timezone with no daylight saving time changes. The timezone picker also displays Europe/Kiev as Europe/Kyiv while preserving existing saved schedules.
--- a/.changeset/explicit-utc-scheduled-triggers.md
+++ b/.changeset/explicit-utc-scheduled-triggers.md
@@ -4,4 +4,4 @@
 
 # Schedule Data Mart triggers in UTC
 
-Scheduled triggers now offer UTC as an explicit fixed-offset timezone with no daylight saving time changes. The timezone picker also displays Europe/Kiev as Europe/Kyiv while preserving existing saved schedules.
+Scheduled triggers now offer UTC as an explicit fixed-offset timezone with no daylight saving time changes. Existing UTC aliases remain compatible, and legacy Europe/Kiev identifiers are shown as Europe/Kyiv without rewriting saved schedules.
--- a/apps/backend/src/common/scheduler/shared/entities/scheduled-trigger.entity.spec.ts
+++ b/apps/backend/src/common/scheduler/shared/entities/scheduled-trigger.entity.spec.ts
@@ -68,6 +68,20 @@ describe('ScheduledTrigger', () => {
       expect(utcTrigger.nextRunTimestamp?.toISOString()).toBe(expectedNextRun);
     });
 
+    it.each([
+      ['winter', '2026-01-15T10:00:00.000Z', '2026-01-16T09:00:00.000Z'],
+      ['summer', '2026-07-15T10:00:00.000Z', '2026-07-16T08:00:00.000Z'],
+    ])(
+      'applies Europe/London daylight saving behavior in %s',
+      (_season, startFrom, expectedNextRun) => {
+        const londonTrigger = new TestScheduledTrigger('0 9 * * *', 'Europe/London');
+
+        londonTrigger.scheduleNextRun(new Date(startFrom));
+
+        expect(londonTrigger.nextRunTimestamp?.toISOString()).toBe(expectedNextRun);
+      }
+    );
+
     it('should handle hourly cron expression correctly', () => {
       // Arrange - Every hour
       const hourlyTrigger = new TestScheduledTrigger('0 0 * * * *');
--- a/apps/backend/src/common/scheduler/shared/entities/scheduled-trigger.entity.spec.ts
+++ b/apps/backend/src/common/scheduler/shared/entities/scheduled-trigger.entity.spec.ts
@@ -3,10 +3,10 @@ import { TriggerStatus } from './trigger-status';
 
 // Create a concrete implementation of the abstract ScheduledTrigger class for testing
 class TestScheduledTrigger extends ScheduledTrigger {
-  constructor(cronExpression: string) {
+  constructor(cronExpression: string, timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone) {
     super();
     this.cronExpression = cronExpression;
-    this.timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+    this.timeZone = timeZone;
     this.id = 'test-id';
     this.nextRunTimestamp = null;
     this.lastRunTimestamp = null;
@@ -57,6 +57,17 @@ describe('ScheduledTrigger', () => {
   });
 
   describe('scheduleNextRun', () => {
+    it.each([
+      ['winter', '2026-01-15T10:00:00.000Z', '2026-01-16T09:00:00.000Z'],
+      ['summer', '2026-07-15T10:00:00.000Z', '2026-07-16T09:00:00.000Z'],
+    ])('keeps a daily UTC schedule fixed through %s', (_season, startFrom, expectedNextRun) => {
+      const utcTrigger = new TestScheduledTrigger('0 9 * * *', 'UTC');
+
+      utcTrigger.scheduleNextRun(new Date(startFrom));
+
+      expect(utcTrigger.nextRunTimestamp?.toISOString()).toBe(expectedNextRun);
+    });
+
     it('should handle hourly cron expression correctly', () => {
       // Arrange - Every hour
       const hourlyTrigger = new TestScheduledTrigger('0 0 * * * *');
--- a/apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleConfig/ScheduleConfig.test.tsx
+++ b/apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleConfig/ScheduleConfig.test.tsx
@@ -1,0 +1,56 @@
+import { fireEvent, render, screen, waitFor, within } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { ScheduleConfig } from './ScheduleConfig';
+
+describe('ScheduleConfig timezone selection', () => {
+  beforeEach(() => {
+    const resolvedOptions = Intl.DateTimeFormat().resolvedOptions();
+    vi.spyOn(Intl, 'supportedValuesOf').mockReturnValue(['Europe/Kiev', 'Europe/London']);
+    vi.spyOn(Intl.DateTimeFormat.prototype, 'resolvedOptions').mockReturnValue({
+      ...resolvedOptions,
+      timeZone: 'Europe/London',
+    });
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('shows the modern Kyiv spelling for the compatible Europe/Kiev value', () => {
+    render(<ScheduleConfig timezone='Europe/Kiev' />);
+
+    expect(screen.getByText(/^Europe\/Kyiv \(/)).toBeInTheDocument();
+
+    fireEvent.click(screen.getByText('Preview'));
+
+    expect(screen.getByText('Europe/Kyiv')).toBeInTheDocument();
+    expect(screen.queryByText('Europe/Kiev')).not.toBeInTheDocument();
+    expect(
+      screen.getByText(
+        'Schedule runs in Europe/Kyiv (not your local Europe/London time). Execution time may differ from expected.'
+      )
+    ).toBeInTheDocument();
+  });
+
+  it('lets the user select UTC and emits the canonical UTC value', async () => {
+    const onChange = vi.fn();
+    render(<ScheduleConfig timezone='Europe/London' onChange={onChange} />);
+
+    const timezoneCombobox = screen.getAllByRole('combobox')[1];
+    expect(timezoneCombobox).toHaveTextContent(/^Europe\/London \(/);
+    fireEvent.click(timezoneCombobox);
+    fireEvent.click(
+      await within(document.body).findByRole('option', {
+        name: 'UTC (+00:00, no DST)',
+      })
+    );
+
+    await waitFor(() => {
+      expect(onChange).toHaveBeenLastCalledWith({
+        cron: '0 9 * * *',
+        timezone: 'UTC',
+        enabled: true,
+      });
+    });
+  });
+});
--- a/apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleConfig/ScheduleConfig.test.tsx
+++ b/apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleConfig/ScheduleConfig.test.tsx
@@ -32,16 +32,91 @@ describe('ScheduleConfig timezone selection', () => {
     ).toBeInTheDocument();
   });
 
+  it.each(['Etc/UTC', 'GMT', 'Etc/GMT'])(
+    'represents the stored %s alias as UTC without changing the emitted value',
+    async timezone => {
+      const onChange = vi.fn();
+      render(<ScheduleConfig timezone={timezone} onChange={onChange} />);
+
+      expect(screen.getByRole('combobox', { name: 'Timezone' })).toHaveTextContent('UTC (+00:00)');
+
+      await waitFor(() => {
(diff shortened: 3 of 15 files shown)


.changeset/explicit-utc-scheduled-triggers.md
```diff
@@ -0,0 +1,7 @@
+---
```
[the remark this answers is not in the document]
max-voloshyn: it's outside Fibery Work Item

laskevych (commented):
Reviewed the timezone handling end-to-end (picker, display, create paths, backend cron parsing). One blocking remark (the self-contradictory timezone alert, see inline comment on `ScheduleConfig.tsx`); everything else is non-blocking. Not repeating the two points already raised by the Codex review (changeset filename ID, stored UTC aliases hidden by the picker filter) — both still stand.

apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleConfig/ScheduleConfig.tsx
```diff
@@ -657,13 +660,13 @@ export function ScheduleConfig({
                     </div>
                   </div>
 
-                  {isEnabled && needsTimezone && currentTimezone !== getBrowserTimezone() && (
+                  {isEnabled && needsTimezone && currentTimezone !== browserTimezone && (
```
laskevych: 🟡 **Important (blocking):** The mismatch check still compares raw identifiers (`currentTimezone !== browserTimezone`) while the alert body now renders both sides through `getTimezoneDisplayName`. For a trigger stored as `Europe/Kiev` viewed in a modern browser (where `resolvedOptions().timeZone` is the canonical `Europe/Kyiv`), the condition is true and the alert renders: *"Schedule runs in Europe/Kyiv (not your local Europe/Kyiv time). Execution time may differ from expected."* — a warning that names the same zone twice. The same happens for stored `UTC` vs a browser resolving `Etc/UTC`, which is the exact flow this PR adds. Comparing canonicalized values (or at least the display names) would fix it; the new test pins the browser to `Europe/London`, so this path is currently unreachable in the suite.

apps/web/src/services/timezone.service.ts
```diff
@@ -28,6 +28,10 @@ interface TimezoneData {
   isDST: boolean;
 }
 
+const UTC_TIMEZONE = 'UTC';
+const UTC_TIMEZONE_ALIASES = new Set(['UTC', 'Etc/UTC', 'GMT', 'Etc/GMT']);
+const TIMEZONE_DISPLAY_NAME_OVERRIDES = new Map([['Europe/Kiev', 'Europe/Kyiv']]);
```
laskevych: 🔵 **Question (non-blocking):** The `Europe/Kiev → Europe/Kyiv` override is display-only and one-directional, so behavior stays runtime-dependent: on modern-ICU runtimes `Intl.supportedValuesOf` returns only `Europe/Kyiv`, so a schedule stored as `Europe/Kiev` matches no option (`Combobox` resolves by exact `value`) and the editor shows the "Select timezone" placeholder; on older-ICU runtimes (e.g. current Node lists `Europe/Kiev`, not `Europe/Kyiv`) the option labeled `Europe/Kyiv (+03:00)` persists `Europe/Kiev`, so the same visible choice stores different identifiers per browser. This is pre-existing, but the changeset's "preserving existing saved schedules" claim only holds on legacy-ICU runtimes, and `ScheduleConfig.test.tsx` mocks `supportedValuesOf` to the legacy spelling, so only the favorable half is tested. Was the modern-ICU case considered? If yes, could the changeset wording be softened (the rename also shows in schedule tables via `ScheduleDisplay`, not just the picker)?

apps/web/src/services/timezone.service.ts
```diff
@@ -41,7 +45,11 @@ class TimezoneService {
   getTimezones(): string[] {
     // Currently using the browser's Intl API to get supported timezones
     // This could be replaced with an API call in the future
-    return Intl.supportedValuesOf('timeZone');
+    const runtimeTimezones = Intl.supportedValuesOf('timeZone').filter(
```
laskevych: 🟡 **Important (non-blocking):** Both trigger create paths — `use-scheduled-trigger-form.ts:25` and `ReportSchedulesInlineList.tsx:72` (`defaultTimezone`) — seed new triggers with the raw `Intl.DateTimeFormat().resolvedOptions().timeZone`, which can itself be one of the aliases this filter strips (`Etc/UTC`, `GMT` on some hosts/containers). The picker then shows "Select timezone" on a trigger the user just created, while the form still submits the alias — i.e. the app keeps generating, post-PR, the very values the explicit `UTC` option was meant to replace. This extends the stored-alias case already flagged above to the create path.

apps/web/src/services/timezone.service.ts
```diff
@@ -28,6 +28,10 @@ interface TimezoneData {
   isDST: boolean;
 }
 
+const UTC_TIMEZONE = 'UTC';
+const UTC_TIMEZONE_ALIASES = new Set(['UTC', 'Etc/UTC', 'GMT', 'Etc/GMT']);
```
laskevych: 💡 **Suggestion (non-blocking):** The PR now encodes "same zone, different spelling" in two independent mechanisms (`UTC_TIMEZONE_ALIASES` as a filter + early-return guards, and `TIMEZONE_DISPLAY_NAME_OVERRIDES` at render), while no consumer canonicalizes identifiers where they enter. A single `canonicalizeTimezone()` applied at the boundaries (mapper / form defaults / picker value) would resolve the alert comparison, the unmatched-option cases, and the create-path defaults with one rule, and future tzdata renames (`Asia/Calcutta → Asia/Kolkata`, `America/Godthab → America/Nuuk`) become one table row instead of scattered patches.

apps/web/src/services/timezone.service.ts
```diff
@@ -157,11 +182,12 @@ class TimezoneService {
...
   private getDisplayName(timezone: string, offsetString: string): string {
-    // Replace underscores with spaces and format city names
-    // const parts = timezone.split('/');
-    // const city = parts[parts.length - 1].replace(/_/g, ' ');
+    if (timezone === UTC_TIMEZONE) {
+      return `${UTC_TIMEZONE} (${offsetString}, no DST)`;
+    }
 
-    return `${timezone} (${offsetString})`;
+    const displayTimezone = this.getTimezoneDisplayName(timezone);
```
laskevych, line 263: 💡 **Suggestion (non-blocking):** `Combobox` search matches only `keywords={[option.label]}`, so identifiers hidden by the relabeling become unsearchable: on legacy-ICU runtimes the only Ukrainian option (value `Europe/Kiev`, label `Europe/Kyiv (+03:00)`) returns "No timezones found" for the query "Kiev", and "Etc/UTC"/"GMT" no longer find the zero-offset entry. Passing the identifier alongside the label in `keywords` restores this.

apps/backend/src/common/scheduler/shared/entities/scheduled-trigger.entity.spec.ts
```diff
@@ -57,6 +57,17 @@ describe('ScheduledTrigger', () => {
   });
 
   describe('scheduleNextRun', () => {
+    it.each([
```
laskevych, line 60: 💡 **Suggestion (non-blocking):** The winter/summer UTC cases can't fail on a UTC CI runner even if the timezone plumbing regresses — a UTC cron in a UTC process yields the same instants whether `this.timeZone` is honored or not. A contrast case with a DST-observing zone (e.g. `Europe/London`, `0 9 * * *` → `09:00Z` in January but `08:00Z` in July) would actually pin the behavior the changeset describes.

apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleDisplay/ScheduleDisplay.tsx
```diff
@@ -1,4 +1,5 @@
 import { useMemo } from 'react';
+import { timezoneService } from '../../../../../services';
```
laskevych: 🧹 **Nitpick (non-blocking):** Importing from the `services` barrel pulls the `api-service`/`apiClient` module graph (axios instance, auth state, interceptors) into a pure presentational table cell — and the new `ScheduleDisplay.test.tsx` now evaluates that whole chain to assert two strings. Importing `services/timezone.service` directly avoids it.

apps/web/src/features/data-marts/scheduled-triggers/model/mappers/data-quality-run-trigger.mapper.test.ts
```diff
@@ -51,4 +51,15 @@ describe('DataQualityRunTriggerMapper', () => {
       triggerConfig: null,
     });
   });
+
+  it('serializes the canonical UTC value in an update request', () => {
```
laskevych: 🧹 **Nitpick (non-blocking):** The mapper is unchanged by this PR and copies `timeZone` verbatim, so this test passes identically with any string — the title "serializes the canonical UTC value" implies canonicalization that doesn't exist. Consider dropping it or asserting at a boundary that actually normalizes.

apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleConfig/ScheduleConfig.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        'Schedule runs in Europe/Kyiv (not your local Europe/London time). Execution time may differ from expected.'
+      )
+    ).toBeInTheDocument();
+  });
+
+  it('lets the user select UTC and emits the canonical UTC value', async () => {
+    const onChange = vi.fn();
+    render(<ScheduleConfig timezone='Europe/London' onChange={onChange} />);
+
+    const timezoneCombobox = screen.getAllByRole('combobox')[1];
```
laskevych: 🧹 **Nitpick (non-blocking):** `screen.getAllByRole('combobox')[1]` targets the timezone control by DOM position; any select added above it (or a changed default schedule type) silently retargets the click, and the failure will read as a timezone regression. The `Timezone` label isn't wired to the control, so an `aria-label` on the `Combobox` (or `htmlFor`/`id` pairing) would make this robust.

apps/web/src/services/timezone.service.ts
```diff
@@ -41,7 +45,11 @@ class TimezoneService {
   getTimezones(): string[] {
     // Currently using the browser's Intl API to get supported timezones
     // This could be replaced with an API call in the future
-    return Intl.supportedValuesOf('timeZone');
+    const runtimeTimezones = Intl.supportedValuesOf('timeZone').filter(
+      timezone => !UTC_TIMEZONE_ALIASES.has(timezone)
+    );
```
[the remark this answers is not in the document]
max-voloshyn: Fixed in `346a56a0b`. The picker canonicalizes stored UTC aliases to the single `UTC` option for presentation while keeping the original stored value in form state until the user explicitly changes the timezone. Added selection and preservation coverage.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleConfig/ScheduleConfig.tsx
max-voloshyn: Fixed in `346a56a0b`. The mismatch warning now compares canonical timezone identities, so equivalent UTC and Kyiv aliases do not produce a same-name warning. Added regression coverage for alias-equivalent browser and stored values.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/services/timezone.service.ts
max-voloshyn: Addressed in `346a56a0b`. Picker options now canonicalize both ICU spellings to `Europe/Kyiv`, use a compatible runtime identifier for offset and DST calculations, and preserve the saved `Europe/Kiev` value until explicit selection. Coverage includes both legacy- and modern-ICU lists.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/services/timezone.service.ts
max-voloshyn: Fixed in `346a56a0b`. Both scheduled-trigger creation defaults now canonicalize the browser timezone, so UTC aliases create `UTC` and `Europe/Kiev` creates `Europe/Kyiv`. Existing loaded schedules are still preserved verbatim unless explicitly changed.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/services/timezone.service.ts
max-voloshyn: Implemented in `346a56a0b`. UTC aliases and `Europe/Kiev` are handled through one canonicalization map used by picker options, defaults, equivalence checks, display, search keywords, and runtime calculation fallback.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/services/timezone.service.ts
max-voloshyn: Fixed in `346a56a0b`. Combobox options now include canonical and legacy identifiers as search keywords, so `Kiev`, `Etc/UTC`, `GMT`, and `Etc/GMT` find their canonical picker choices.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/common/scheduler/shared/entities/scheduled-trigger.entity.spec.ts
max-voloshyn: Added in `346a56a0b`. The scheduler test now includes `Europe/London` in winter and summer, asserting `09:00Z` and `08:00Z` respectively, so ignoring the configured timezone would fail.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleDisplay/ScheduleDisplay.tsx
max-voloshyn: Fixed in `346a56a0b`. `ScheduleDisplay` now imports `timezone.service` directly, avoiding the unrelated API and auth service barrel graph.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/data-marts/scheduled-triggers/model/mappers/data-quality-run-trigger.mapper.test.ts
max-voloshyn: Removed in `346a56a0b`. Canonicalization is now tested at the actual form-default and picker boundaries instead of the unchanged pass-through mapper.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/data-marts/scheduled-triggers/components/ScheduleConfig/ScheduleConfig.test.tsx
max-voloshyn: Fixed in `346a56a0b`. `Combobox` accepts an `ariaLabel`, the timezone control is labelled `Timezone`, and tests query it by accessible name rather than DOM position.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/services/timezone.service.ts
```diff
@@ -157,19 +256,22 @@ class TimezoneService {
    * @returns {string} Display name
    */
   private getDisplayName(timezone: string, offsetString: string): string {
-    // Replace underscores with spaces and format city names
-    // const parts = timezone.split('/');
-    // const city = parts[parts.length - 1].replace(/_/g, ' ');
+    if (this.canonicalizeTimezone(timezone) === UTC_TIMEZONE) {
+      return `${UTC_TIMEZONE} (${offsetString})`;
```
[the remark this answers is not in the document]
max-voloshyn: UTC (+00:00) is expected

laskevych (approved):

PR #1579, opened 2026-08-31
perf(connectors): prune partitions in BigQuery incremental MERGE runs

prokopetsv:
# Problem

Every incremental connector run into Google BigQuery merged new records with an `ON` clause containing only unique-key equality conditions (`target.date = source.date`). BigQuery cannot prune partitions from a join predicate against a subquery of literals — pruning requires a constant filter on the target column — so every MERGE scanned the entire destination table to write a few days of data. Scan cost therefore grew with table history rather than with imported volume, and long-lived time-series tables (ad performance reports) eventually exhausted per-user daily quotas: a client's scheduled RedditAds run failed with `Custom quota exceeded: QueryUsagePerUserPerDay`. Measured on a synthetic 600-partition table, the current query shape processes 1.16 GB where a pruned one processes 3.89 MB (~99.7% reduction).

# Solution

Derive a constant date-range filter from the batch being merged and append it to the MERGE `ON` clause: `AND target.date BETWEEN DATE '…' AND DATE '…'`. The addition is semantically neutral whenever the partition column is part of the unique key, because `target.date = source.date` already restricts matches to the batch's dates — the literal range only restates that fact in a form the query planner can act on.

The partition column comes from the **destination table itself**, not from the schema: `getAListOfExistingColumns()` now reads `is_partitioning_column` from `INFORMATION_SCHEMA.COLUMNS`, and `createTableIfItDoesntExist()` records which column it actually emitted `PARTITION BY` for. The `GoogleBigQueryPartitioned` schema flag only drives table creation (unchanged last-wins selection when several fields carry it). This matters in practice: LinkedIn Ads flags both `dateRangeStart` and `dateRangeEnd`, so schema inference would have filtered a different column than existing tables are partitioned by, producing a valid predicate that prunes nothing.

Safety is guard-based: the predicate is emitted only when every condition holds, and any ambiguity falls back to the existing full-scan behavior (slow but always correct). Guards cover:

- the destination table has no partitioning column — entity nodes like `ads` and `campaigns`, legacy unpartitioned tables — SQL unchanged
- the partitioning column is not in `uniqueKeyColumns` — without the matching equality condition, a bounded scan would hide matching target rows in out-of-range partitions and the MERGE would insert duplicates (an audit of all 15 sources found one live instance: OpenHolidays flags `date` but keys on `id`)
- the column type is not `DATE`/`DATETIME`/`TIMESTAMP`
- any record missing its partition value
- any value not matching one canonical fixed-width shape per type (fractions, `Z`, and zone offsets rejected; `T`/space separators normalized) — min/max is computed by string comparison, which equals chronological order only for uniform fixed-width values; this also blocks quote-shaped input from reaching the SQL literal
- any value that is shape-valid but not a real calendar date/time (`2026-02-31`, `25:00:00`) — a typed literal would fail the whole query, where `SAFE_CAST` in source rows degrades the same input to `NULL`

Value formatting is shared with the source-row builder via an extracted `formatColumnValue()`, so predicate literals can never diverge from the row values they bound. Chunked batches (`executeMergeQueryRecursively`) each compute their own, tighter range. Table-creation DDL is fixed alongside: `DATETIME`/`TIMESTAMP` partition columns now emit valid `_TRUNC(col, DAY)` expressions (the bare-column form is invalid DDL for those types; dormant today since every flagged field repo-wide is `DATE`-typed). Other MERGE storages (Snowflake, Redshift, Databricks) are out of scope — they have no partition metadata flag and different billing models.

# Changes

- Added `buildPartitionPredicate()` to `GoogleBigQueryStorage` and wired it into the MERGE `ON` clause
- Added `is_partitioning_column` to the `INFORMATION_SCHEMA.COLUMNS` query and both result-parsing branches; `createTableIfItDoesntExist()` records the emitted partition column the same way, so same-run create-then-merge flows also prune
- Added `buildPartitionByExpression()`: `DATE` partitions by bare column, `DATETIME`/`TIMESTAMP` by `_TRUNC(col, DAY)`, unsupported types skip `PARTITION BY` instead of emitting invalid SQL
- Added `isRealCalendarValue()` semantic validation on top of the shape regexes
- Extracted value normalization from `buildMergeQuery()` into `formatColumnValue()`, shared by row building and the predicate
- Normalized `T`/space datetime separators before validation and comparison
- Added 13 tests covering the emitted range, entity-node passthrough, guard suppressions (missing value, unpartitioned table, non-unique-key partition column, zone-bearing timestamps, injection-shaped values, impossible calendar dates, out-of-range times), TIMESTAMP literals, table-truth-vs-schema mismatch (LinkedIn case), and unchanged unique-key conditions
- Documented the `GoogleBigQueryPartitioned` contract in `CREATING_CONNECTOR.md`: date-like type required (`_TRUNC` daily granularity for `DATETIME`/`TIMESTAMP`), field must be in `uniqueKeys`, flag exactly one field per node
- Added changeset `6880-bq-merge-partition-pruning.md` (`'owox': minor`)

### DIFF
--- a/packages/connectors/CREATING_CONNECTOR.md
+++ b/packages/connectors/CREATING_CONNECTOR.md
@@ -338,12 +338,15 @@ The flag has two effects:
 
 Two rules make the filter work:
 
-- The field type must be `DATE`, `DATETIME`, or `TIMESTAMP`.
+- The field type must be `DATE`, `DATETIME`, or `TIMESTAMP`. `DATETIME`
+  and `TIMESTAMP` fields get daily partitions via `_TRUNC`.
 - The node's `uniqueKeys` must include the field. Without it, the MERGE
   skips the filter and scans the whole table on every run.
 
-Time-series nodes should set both the flag and the `uniqueKeys` entry.
-Entity nodes without a date field need neither.
+Flag exactly one field per node. When several fields carry the flag, the
+last one selected for the table wins. Time-series nodes should set both
+the flag and the `uniqueKeys` entry. Entity nodes without a date field
+need neither.
 
 ## 6. Testing Your Connector
 
--- a/packages/connectors/CREATING_CONNECTOR.md
+++ b/packages/connectors/CREATING_CONNECTOR.md
@@ -339,7 +339,9 @@ The flag has two effects:
 Two rules make the filter work:
 
 - The field type must be `DATE`, `DATETIME`, or `TIMESTAMP`. `DATETIME`
-  and `TIMESTAMP` fields get daily partitions via `_TRUNC`.
+  and `TIMESTAMP` fields get daily partitions via `_TRUNC`. Any other type
+  skips partitioning: the table is created unpartitioned and the run log
+  records the skipped column.
 - The node's `uniqueKeys` must include the field. Without it, the MERGE
   skips the filter and scans the whole table on every run.
 
--- a/packages/connectors/CREATING_CONNECTOR.md
+++ b/packages/connectors/CREATING_CONNECTOR.md
@@ -318,6 +318,33 @@ constructor(config) {
 }
 ```
 
+### Partitioning in Google BigQuery
+
+Mark one date-like field per node with `GoogleBigQueryPartitioned: true`:
+
+```javascript
+date: {
+  description: 'The date for this metric.',
+  type: DATA_TYPES.DATE,
+  GoogleBigQueryPartitioned: true
+}
+```
+
+The flag has two effects:
+
+- New destination tables get `PARTITION BY` on this field.
+- Incremental MERGE runs add a date-range filter on the target table.
+  BigQuery then scans only the partitions being written, not the full table.
+
+Two rules make the filter work:
+
+- The field type must be `DATE`, `DATETIME`, or `TIMESTAMP`.
+- The node's `uniqueKeys` must include the field. Without it, the MERGE
+  skips the filter and scans the whole table on every run.
+
+Time-series nodes should set both the flag and the `uniqueKeys` entry.
+Entity nodes without a date field need neither.
+
 ## 6. Testing Your Connector
 
 After creating your connector:
--- a/packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
+++ b/packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
@@ -160,7 +160,7 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
           WHERE schema_name = '${this.config.DestinationDatasetName.value}'
         );
         IF dataset_exists THEN 
-          SELECT column_name, data_type
+          SELECT column_name, data_type, is_partitioning_column
           FROM \`${this.config.DestinationDatasetID.value}.INFORMATION_SCHEMA.COLUMNS\`
           WHERE table_name = '${this.config.DestinationTableName.value}'
           ORDER BY ordinal_position;
@@ -176,11 +176,11 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
 
         if( queryResults.rows ) {
           queryResults.rows.map(row => {
-            columns[ row.f[0].v ]  = {"name": row.f[0].v, "type": row.f[1].v}
+            columns[ row.f[0].v ]  = {"name": row.f[0].v, "type": row.f[1].v, "isPartitioningColumn": row.f[2].v === "YES"}
           });
         } else if (Array.isArray(queryResults)) {
           queryResults.map(row => {
-            columns[ row.column_name ] = {"name": row.column_name, "type": row.data_type}
+            columns[ row.column_name ] = {"name": row.column_name, "type": row.data_type, "isPartitioningColumn": row.is_partitioning_column === "YES"}
           });
         }
 
@@ -202,24 +202,11 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
 
     }
 
-  //---- getPartitionColumn ------------------------------------------
-    /**
-     * Finds the schema field flagged as the BigQuery partition column
-     * @return {string|null} - the column name, or null when the schema declares none
-     */
-    getPartitionColumn() {
-      for (const columnName in this.schema) {
-        if( this.schema[ columnName ] && this.schema[ columnName ]["GoogleBigQueryPartitioned"] ) {
-          return columnName;
-        }
-      }
-      return null;
-    }
-
   //---- createTableIfItDoesntExist ----------------------------------
     async createTableIfItDoesntExist(quoteColumnNames = false) {
 
       let columns = [];
+      let columnPartitioned = null;
       let existingColumns = {};
 
       let selectedFields = this.getSelectedFields();
@@ -239,6 +226,13 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
           columnDescription = ` OPTIONS(description="${this.obfuscateSpecialCharacters(this.schema[ columnName ]["description"])}")`;
         }
 
+        // Last flagged column wins — LinkedIn flags two; existing tables
+        // were created under this rule, so it must not change
+        if( "GoogleBigQueryPartitioned" in this.schema[ columnName ]
+        && this.schema[ columnName ]["GoogleBigQueryPartitioned"] ) {
+          columnPartitioned = columnName;
+        }
+
         const sqlColumnName = quoteColumnNames ? quoteBigQueryIdentifier(columnName) : columnName;
         columns.push(`${sqlColumnName} ${columnType}${columnDescription}`);
         
@@ -256,11 +250,14 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
       let query = `---- Creating table if it not exists -----\n`;
       query += `CREATE TABLE IF NOT EXISTS \`${this.config.DestinationDatasetID.value}.${this.config.DestinationTableName.value}\` (\n${columns})`
 
-      // The partition column must be among the columns actually created:
-      // a schema may flag a field the user did not select for this table.
-      const columnPartitioned = this.getPartitionColumn();
-      if( columnPartitioned && tableColumns.includes(columnPartitioned) ) {
-        query += `\nPARTITION BY ${quoteColumnNames ? quoteBigQueryIdentifier(columnPartitioned) : columnPartitioned}`;
+      if( columnPartitioned ) {
+        const sqlName = quoteColumnNames ? quoteBigQueryIdentifier(columnPartitioned) : columnPartitioned;
+        const partitionExpression = this.buildPartitionByExpression(sqlName, this.getColumnType(columnPartitioned));
+        if( partitionExpression ) {
+          query += `\nPARTITION BY ${partitionExpression}`;
+          // Record table truth for buildPartitionPredicate in same-run flows
+          existingColumns[ columnPartitioned ]["isPartitioningColumn"] = true;
+        }
       }
 
       if( this.description ) {
@@ -614,6 +611,43 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
       return this.obfuscateSpecialCharacters( rawValue );
     }
 
+  //---- buildPartitionByExpression ----------------------------------
+    /**
+     * DDL expression for PARTITION BY, or null for types BigQuery cannot
+     * partition by. DATETIME/TIMESTAMP columns need daily truncation.
+     */
+    buildPartitionByExpression(sqlColumnName, columnType) {
+      switch( String(columnType).toUpperCase() ) {
+        case "DATE": return sqlColumnName;
+        case "DATETIME": return `DATETIME_TRUNC(${sqlColumnName}, DAY)`;
+        case "TIMESTAMP": return `TIMESTAMP_TRUNC(${sqlColumnName}, DAY)`;
+        default: return null;
+      }
+    }
+
+  //---- isRealCalendarValue -----------------------------------------
+    /**
+     * True when a canonical 'YYYY-MM-DD[ HH:MM:SS]' string denotes a real
+     * calendar date and time. The shape regex accepts impossible values like
+     * 2026-02-31; SAFE_CAST turns those into NULL inside source rows, but a
+     * typed literal in the predicate would fail the whole query.
+     */
+    isRealCalendarValue(value) {
+      const [datePart, timePart] = value.split(" ");
+      const [year, month, day] = datePart.split("-").map(Number);
+      const date = new Date(Date.UTC(year, month - 1, day));
+      if( date.getUTCFullYear() !== year || date.getUTCMonth() !== month - 1 || date.getUTCDate() !== day ) {
+        return false;
+      }
+      if( timePart ) {
+        const [hours, minutes, seconds] = timePart.split(":").map(Number);
+        if( hours > 23 || minutes > 59 || seconds > 59 ) {
+          return false;
+        }
+      }
+      return true;
+    }
+
   //---- buildPartitionPredicate -------------------------------------
     /**
      * Builds a constant partition filter for the MERGE ON clause, so BigQuery
@@ -621,10 +655,16 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
      * `target.date = source.date` alone does not prune: the source is a
      * subquery of literals, and pruning needs a constant filter on the target.
      *
-     * Returns null (no predicate, today's full-scan behavior) whenever the
-     * range cannot be derived safely: no partition column in the schema, the
-     * column is missing from the destination table, any record lacks a value,
-     * or a value does not look like a date/datetime literal.
+     * The partition column is taken from the destination table itself
+     * (INFORMATION_SCHEMA, or recorded at CREATE TABLE time), not from the
+     * schema flag: a legacy table can be partitioned by a different flagged
+     * field, or by nothing at all.
+     *
+     * Returns null (no predicate, the MERGE falls back to a full table scan)
+     * whenever the range cannot be derived safely: the destination table has
+     * no partitioning column, that column is not part of the unique key, its
+     * type is not date-like, any record lacks a value, or a value is not a
+     * real calendar date/time in canonical shape.
      *
      * @param {Array} recordKeys - Record keys of the batch being merged
      * @return {string|null} - SQL predicate for the target table, or null
@@ -643,7 +683,13 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
         "TIMESTAMP": /^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$/
       };
 
-      const partitionColumn = this.getPartitionColumn();
+      // The destination table's real partitioning column, as reported by
+      // INFORMATION_SCHEMA or recorded at CREATE TABLE time. The schema flag
+      // is not trusted here: a legacy table can be partitioned by a different
+      // flagged field (LinkedIn flags two), or by nothing at all.
+      const partitionColumn = Object.keys(this.existingColumns).find(
+        name => this.existingColumns[ name ] && this.existingColumns[ name ]["isPartitioningColumn"]
+      );
       if( !partitionColumn ) {
         return null;
       }
@@ -657,11 +703,7 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
         return null;
       }
 
-      // A table created before the schema gained the flag has no such column
       const columnInfo = this.existingColumns[ partitionColumn ];
-      if( !columnInfo ) {
-        return null;
-      }
 
       const columnType = String(columnInfo.type).toUpperCase();
       const literalPattern = DATE_LIKE_LITERALS[ columnType ];
@@ -679,9 +721,10 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
         // batch mixing them still has one uniform, comparable shape
         const value = typeof formatted === "string" ? formatted.replace("T", " ") : formatted;
 
-        // One unusable value and the whole predicate is off: a predicate that
-        // misses a record's partition would silently drop that record's update
-        if( typeof value !== "string" || !literalPattern.test(value) ) {
+        // One unusable value and the whole predicate is off: a range that
+        // misses a record's partition would hide its target row from the
+        // MERGE and insert a duplicate instead of updating it
+        if( typeof value !== "string" || !literalPattern.test(value) || !this.isRealCalendarValue(value) ) {
           return null;
         }
 
--- a/packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
+++ b/packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
@@ -257,6 +257,10 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
           query += `\nPARTITION BY ${partitionExpression}`;
           // Record table truth for buildPartitionPredicate in same-run flows
           existingColumns[ columnPartitioned ]["isPartitioningColumn"] = true;
+        } else {
+          // A silent skip here would ship a dead flag: the table is created
+          // unpartitioned and every MERGE full-scans with no symptom but cost
+          this.config.logMessage(`Column '${columnPartitioned}' has type '${this.getColumnType(columnPartitioned)}' which cannot be a partition column; creating the table without partitioning`);
         }
       }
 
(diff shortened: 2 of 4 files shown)


packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -575,6 +583,163 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+    /**
+     * DDL expression for PARTITION BY, or null for types BigQuery cannot
+     * partition by. DATETIME/TIMESTAMP columns need daily truncation.
+     */
+    buildPartitionByExpression(sqlColumnName, columnType) {
+      switch( String(columnType).toUpperCase() ) {
+        case "DATE": return sqlColumnName;
+        case "DATETIME": return `DATETIME_TRUNC(${sqlColumnName}, DAY)`;
+        case "TIMESTAMP": return `TIMESTAMP_TRUNC(${sqlColumnName}, DAY)`;
+        default: return null;
```
laskevych, line 643: 🟡 **Important (non-blocking):** this `default: return null` converts a mis-flagged partition column from a loud failure into a silent one.

Before this PR, `createTableIfItDoesntExist` emitted `PARTITION BY <col>` unconditionally, so flagging a `STRING`- or `TIME`-typed field produced DDL BigQuery rejects and the run failed on the first attempt — the author found out immediately. Now the expression is `null`, the clause is dropped, and the table is created unpartitioned with nothing in the log.

I confirmed the behaviour change by generating the DDL on both branches:

```
DATE       main: PARTITION BY date   pr: PARTITION BY date                      (identical)
DATETIME   main: PARTITION BY date   pr: PARTITION BY DATETIME_TRUNC(date, DAY)
TIMESTAMP  main: PARTITION BY date   pr: PARTITION BY TIMESTAMP_TRUNC(date, DAY)
STRING     main: PARTITION BY date   pr: (no PARTITION BY)
TIME       main: PARTITION BY date   pr: (no PARTITION BY)
```

The first three rows are fine — `DATE` is untouched, and the two `_TRUNC` rows replace DDL that BigQuery rejects. The last two are the concern: the connector ships with a dead flag, every MERGE keeps full-scanning, and the only symptom is an unchanged bill. Since `CREATING_CONNECTOR.md` now advertises the flag, new authors are the ones most likely to hit it.

A `this.config.logMessage(...)` in the `default` branch naming the column and its type would be enough.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -575,6 +583,163 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+      };
+
+      // The destination table's real partitioning column, as reported by
+      // INFORMATION_SCHEMA or recorded at CREATE TABLE time. The schema flag
+      // is not trusted here: a legacy table can be partitioned by a different
+      // flagged field (LinkedIn flags two), or by nothing at all.
+      const partitionColumn = Object.keys(this.existingColumns).find(
+        name => this.existingColumns[ name ] && this.existingColumns[ name ]["isPartitioningColumn"]
+      );
+      if( !partitionColumn ) {
```
laskevych, line 733: 🟡 **Important (non-blocking):** all four guards return `null` silently, so "pruning is on" and "pruning was suppressed" look identical from the outside.

This one and the three below it (`uniqueKeyColumns` at :703, type at :711, value shape at :728) each fall back to correct-but-slow SQL without recording why. The PR description already names live cases that land here: OpenHolidays flags `date` but keys on `id`, and legacy unpartitioned tables exist in the field. A customer on one of those upgrades, keeps the same scan volume, keeps hitting `QueryUsagePerUserPerDay`, and the run log offers nothing to distinguish that from the feature working as intended — support has to read this source and re-derive the guard chain from the connector's schema.

A single `logMessage` when a partition column exists but the predicate was suppressed, naming the reason, turns that into a one-line answer. Worth guarding so it fires once per run rather than once per batch, since `buildPartitionPredicate` is called per MERGE.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -176,11 +176,11 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
 
         if( queryResults.rows ) {
           queryResults.rows.map(row => {
-            columns[ row.f[0].v ]  = {"name": row.f[0].v, "type": row.f[1].v}
+            columns[ row.f[0].v ]  = {"name": row.f[0].v, "type": row.f[1].v, "isPartitioningColumn": row.f[2].v === "YES"}
```
laskevych, line 179: 🟡 **Important (non-blocking):** this line and its sibling at :183 are the only production source of `isPartitioningColumn`, and neither is covered.

All 13 new tests hand-inject `isPartitioningColumn: true` into `existingColumns`, so they prove `buildPartitionPredicate` *given* the flag — never that the flag arrives. The only producers are `row.f[2].v === "YES"` here, `row.is_partitioning_column === "YES"` at :183, and the `createTableIfItDoesntExist` assignment at :259.

`is_partitioning_column` is documented as a `STRING` `"YES"`/`"NO"` column, so the comparison looks right. The concern is not that it is wrong today — it is that if it ever stops matching (a boolean, different casing, a shifted field index after another column is added to the `SELECT`), `partitionColumn` is `undefined`, every guard returns `null`, and the feature becomes a complete no-op that still emits correct SQL and still passes the entire suite. That is the one failure mode here with no symptom at all.

Two small tests over `getAListOfExistingColumns` with a stubbed `executeQuery` — one per result shape — would close it. The same gap covers `buildPartitionByExpression` and the `createTableIfItDoesntExist` flag assignment, neither of which has a test.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -160,7 +160,7 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
           WHERE schema_name = '${this.config.DestinationDatasetName.value}'
         );
         IF dataset_exists THEN 
-          SELECT column_name, data_type
+          SELECT column_name, data_type, is_partitioning_column
```
laskevych, line 163: 🔵 **Question (non-blocking):** was this verified end-to-end against a real BigQuery destination, or measured as a comparison of two query shapes?

The 1.16 GB → 3.89 MB figure reads like a hand-written before/after on a synthetic table, which validates the *idea* but not the plumbing. The step unit tests cannot reach is whether `is_partitioning_column` actually arrives through `executeQuery`: this is a multi-statement script (`DECLARE` / `SET` / `IF`), and `job.getQueryResults()` on a script job returns the last statement's rows — that row shape decides which of the two parsing branches runs.

If a real connector run against a partitioned destination produced a MERGE containing the `BETWEEN` clause, that settles it and the coverage gap above is much lower-stakes. Asking rather than assuming, because a "yes" here is the difference between a shipped optimisation and a silent no-op.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -575,6 +583,163 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+        const record = this.updatedRecordsBuffer[ key ];
+        const formatted = this.formatColumnValue( record ? record[ partitionColumn ] : null, columnType );
+        // 'T' and ' ' separators are equivalent to BigQuery; normalize so a
+        // batch mixing them still has one uniform, comparable shape
+        const value = typeof formatted === "string" ? formatted.replace("T", " ") : formatted;
+
+        // One unusable value and the whole predicate is off: a range that
+        // misses a record's partition would hide its target row from the
+        // MERGE and insert a duplicate instead of updating it
+        if( typeof value !== "string" || !literalPattern.test(value) || !this.isRealCalendarValue(value) ) {
```
laskevych, line 773: 💡 **Suggestion (optional):** a record missing its partition value could be skipped when computing `min`/`max`, instead of disabling pruning for the entire batch.

The reasoning: such a record emits `SAFE_CAST(NULL AS <TYPE>)`, so `target.<col> = source.<col>` evaluates to `NULL`, the row never matches a target row, and it is always an `INSERT`. Adding `AND target.<col> BETWEEN ...` cannot change that outcome — the row is outside the matched set either way. So excluding it from the range is semantically identical to today's behaviour, and the all-null case stays covered by the existing `minValue === null` check below.

As written, one sparse null anywhere in a 250-record buffer disables pruning for that whole MERGE. On a connector whose date field is occasionally absent, that can mean pruning effectively never engages.

Flagging this as optional deliberately — the current guard is correct, and "reject the whole batch on any doubt" is a defensible default for a first release. Worth it only if you have a connector where nulls are realistic.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -575,6 +583,163 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+
+      if (rawValue === undefined || rawValue === null) {
+
+        return null;
+
+      } else if( ( columnType.toUpperCase() == "DATE") && (rawValue instanceof Date) ) {
+
+        return DateUtils.formatDate( rawValue );
+
+      } else if( (columnType.toUpperCase() == "DATETIME") && (rawValue instanceof Date) ) {
```
laskevych: 💡 **Suggestion (non-blocking, pre-existing):** there is no `TIMESTAMP` branch here, so a `Date` in a `TIMESTAMP` column falls through to `obfuscateSpecialCharacters`.

`formatColumnValue(new Date('2026-08-30T10:00:00Z'), 'TIMESTAMP')` matches neither branch and returns `String(date)` — `"Sun Aug 30 2026 13:00:00 GMT+0300 (Eastern European Summer Time)"`. `buildMergeQuery` then emits `SAFE_CAST("Sun Aug 30 2026 ..." AS TIMESTAMP)`, which BigQuery degrades to `NULL`: the row is written with a null timestamp and no error. The predicate itself is safe — the shape regex rejects the string, so pruning is correctly suppressed — this is a row-level data loss, not a pruning bug.

`SnowflakeStorage.buildMergeQuery` already handles exactly this case (`columnType.includes("TIMESTAMP") || == "DATETIME"` with a `Date`), so the two storages disagree on the same input.

This predates the PR and no shipped connector currently declares a `TIMESTAMP` field, so it is not urgent. Raising it here because this PR turns the block into a named shared helper and promotes `TIMESTAMP` to a first-class partition type in both the docs and the tests — which makes the missing branch look intentional to the next reader.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -575,6 +583,163 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+     * @param {string} columnType - the BigQuery column type
+     * @return {string|null} - the formatted value, or null for missing values
+     */
+    formatColumnValue(rawValue, columnType) {
+
+      if (rawValue === undefined || rawValue === null) {
+
+        return null;
+
+      } else if( ( columnType.toUpperCase() == "DATE") && (rawValue instanceof Date) ) {
```
laskevych: 💡 **Suggestion (non-blocking, pre-existing):** `instanceof Date` alone is narrower than the two sibling helpers this function sits next to.

`AbstractStorage.getUniqueKeyByRecordFields` and `AbstractStorage.stringifyNeastedFields` both guard with `!(value instanceof Date) && !(value.constructor.name == "Date")`. That second clause is not redundant padding — it exists because `instanceof` fails across realms, which is the Apps Script target this package also builds for.

A cross-realm `Date` reaching this line takes the `obfuscateSpecialCharacters` path, produces a locale string, `SAFE_CAST`s to `NULL` in the source row, and simultaneously fails the shape regex so pruning switches off — the two symptoms reinforce each other and neither raises an error.

Same single check existed in the inlined code on `main`, so nothing regressed. Mentioning it because extracting the block into a shared helper is the natural moment to bring it in line with its two neighbours.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -226,7 +226,9 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
           columnDescription = ` OPTIONS(description="${this.obfuscateSpecialCharacters(this.schema[ columnName ]["description"])}")`;
         }
 
-        if( "GoogleBigQueryPartitioned" in this.schema[ columnName ] 
+        // Last flagged column wins — LinkedIn flags two; existing tables
+        // were created under this rule, so it must not change
+        if( "GoogleBigQueryPartitioned" in this.schema[ columnName ]
```
laskevych, line 231: 🔵 **Question (non-blocking):** the comment pins the last-wins rule, which is right — but for LinkedIn it makes the partition column depend on the user's configuration.

`columnPartitioned` is the last flagged column while iterating `tableColumns`, which is `getSelectedFields()` — parsed directly from the user's `Fields` config string. `LinkedInAdsAPIReference/adAnalyticsFields.js` flags both `dateRangeStart` (:72) and `dateRangeEnd` (:77), so two customers who list those fields in a different order get new tables partitioned by different columns, and the predicate then targets a different column per deployment.

Preserving last-wins for existing tables is clearly correct and I am not suggesting changing it. The question is about the schema: the new doc says "flag exactly one field per node", but `adAnalyticsFields.js` still carries two, so newly created LinkedIn tables keep inheriting the ambiguity. Is a follow-up planned to drop one flag — which would affect new tables only — or is the ambiguity accepted here? The choice of which one to keep looks like a product call rather than a review call.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -575,6 +583,163 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+
+      let minValue = null;
+      let maxValue = null;
+
+      for (const key of recordKeys) {
+        const record = this.updatedRecordsBuffer[ key ];
+        const formatted = this.formatColumnValue( record ? record[ partitionColumn ] : null, columnType );
+        // 'T' and ' ' separators are equivalent to BigQuery; normalize so a
+        // batch mixing them still has one uniform, comparable shape
+        const value = typeof formatted === "string" ? formatted.replace("T", " ") : formatted;
```
laskevych: 💡 **Suggestion (non-blocking):** this normalisation is the one part of the shared-formatting invariant that only one of the two callers applies.

The PR's safety argument is that predicate literals cannot diverge from the row values they bound, because both go through `formatColumnValue`. That holds today — but only because BigQuery's canonical `DATETIME`/`TIMESTAMP` grammar accepts `( |T)` as the separator. For a batch mixing `'2026-08-30T10:00:00'` and `'2026-08-31 09:30:00'`, `buildMergeQuery` emits `SAFE_CAST("2026-08-30T10:00:00" AS DATETIME)` while the predicate emits `DATETIME '2026-08-30 10:00:00'`: equal values, different strings.

So this is not a bug, and I am not proposing it for tidiness — the point is that the single line making the two comparable lives outside the shared helper, so the guarantee is currently coincidental rather than structural. Moving `.replace("T", " ")` into `formatColumnValue` would make both callers share it by construction, at no cost to the emitted SQL.

packages/connectors/CREATING_CONNECTOR.md
```diff
@@ -318,6 +318,36 @@ constructor(config) {
...
+
+The flag has two effects:
+
+- New destination tables get `PARTITION BY` on this field.
+- Incremental MERGE runs add a date-range filter on the target table.
+  BigQuery then scans only the partitions being written, not the full table.
+
+Two rules make the filter work:
+
+- The field type must be `DATE`, `DATETIME`, or `TIMESTAMP`. `DATETIME`
```
laskevych, line 341: 💡 **Suggestion (non-blocking):** this bullet gives the rule but not the consequence of breaking it.

With the new `default: return null` in `buildPartitionByExpression`, flagging a field of any other type no longer fails `CREATE TABLE` — it silently produces an unpartitioned table, so the author's only feedback is an unchanged scan bill later. One clause here (something like "any other type is skipped and the table is created unpartitioned") keeps the doc aligned with the behaviour this PR introduces.

Pairs with the comment on `buildPartitionByExpression`: if a log line is added there, this sentence is the natural place to name it, so an author who hits it knows what to search for.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -575,6 +583,163 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+    /**
+     * True when a canonical 'YYYY-MM-DD[ HH:MM:SS]' string denotes a real
+     * calendar date and time. The shape regex accepts impossible values like
+     * 2026-02-31; SAFE_CAST turns those into NULL inside source rows, but a
+     * typed literal in the predicate would fail the whole query.
+     */
+    isRealCalendarValue(value) {
+      const [datePart, timePart] = value.split(" ");
+      const [year, month, day] = datePart.split("-").map(Number);
+      const date = new Date(Date.UTC(year, month - 1, day));
```
laskevych, line 657: 💡 **Suggestion (nit, non-blocking):** this round-trip check also rejects years `0001`–`0099`, which is worth a comment rather than a change.

`new Date(Date.UTC(50, 0, 1)).getUTCFullYear()` returns `1950` — the two-digit-year legacy mapping — so a shape-valid `'0050-01-01'` fails the comparison and suppresses the predicate. BigQuery's `DATE` range does start at `0001-01-01`, so these are technically legal values.

The direction is fail-safe and such values do not occur in ad-report data, so no change is needed. The reason to note it: the mechanism is non-obvious, and the same `Date.UTC` behaviour is what makes this check reject genuine garbage like `'0000-00-00'`. A future reader who spots the two-digit-year quirk could "fix" it with `setUTCFullYear` and quietly weaken a guard that is currently doing useful work. One line of comment prevents that.

laskevych (approved):
Reviewed the MERGE query shape, the guard chain, the DDL change, and the new tests. No blocking findings — approving.

The core safety argument holds: the predicate is a provable no-op on the matched set, because the partition column is required to be in `uniqueKeyColumns`, so the `ON` clause already carries `target.<col> = source.<col>`, and `min`/`max` are derived from the same `recordKeys` batch through the same `formatColumnValue`. Every guard fails toward the old full-scan behaviour, so ambiguity costs money, never correctness.

I also checked this is not a regression of existing behaviour, by generating SQL from both `main` and this branch on identical fixtures:

- predicate suppressed (unpartitioned table / entity node / one record with a null date) — the generated MERGE is byte-identical to `main`;
- predicate emitted — the source rows and every other clause are byte-identical; the only delta is the appended `AND target.date BETWEEN ...` line;
- `CREATE TABLE` DDL for a `DATE`-typed flagged column is byte-identical to `main`. Since all 32 `GoogleBigQueryPartitioned` fields in the repo are `DATA_TYPES.DATE`, no existing connector changes its DDL. The `DATETIME`/`TIMESTAMP` change moves those from DDL BigQuery rejects to DDL it accepts, so it can only improve on the current state.

Verified separately: `uniqueKeyColumns` is normalised to an array in the `AbstractStorage` constructor, so the `.includes()` guard cannot do a substring match; `DateUtils.formatDate` goes through `toISOString()`, so the new tests are timezone-independent; and the extra `isPartitioningColumn` key does not disturb `hasSameSchema`, which compares only names and types.

The comments below are all non-blocking. The three worth acting on before this ships are the silent `PARTITION BY` skip, the absence of any log line when pruning is suppressed, and the untested `is_partitioning_column` parsing — together they describe the same risk: when this feature does nothing, nothing says so.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Agreed — the old loud failure was accidentally load-bearing. Added a `logMessage` at the call site naming the column and its type when `PARTITION BY` is skipped, plus a test asserting both the log line and the absence of the clause. Doc updated to name the behavior too (see the CREATING_CONNECTOR.md thread).

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Implemented as `suppressPartitionPredicate(reason)` — fires once per run (instance flag), and only when a partition column exists, so entity nodes stay quiet. Reasons name the column and the failed guard: unique-key membership, non-date type, malformed value, all-null batch. Test covers the once-per-run behavior. Your support scenario (an OpenHolidays user staring at an unchanged bill) is now a one-line log answer.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Right — and this was the sharpest finding of the review: a shifted field index would have been a symptomless no-op. Added six tests: both `INFORMATION_SCHEMA` result shapes (`row.f[i].v` and array-of-objects), `createTableIfItDoesntExist` recording the flag as table truth, `TIMESTAMP_TRUNC` DDL emission, and the unsupported-type skip path. The `"YES"` comparisons are now pinned.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Honest answer: shape comparison only. The 1.16 GB -> 3.89 MB figure is from BigQuery's dry-run estimator on a synthetic 600-partition table — it validates the SQL shape, not the plumbing. Your specific concern (script job -> last statement's rows -> which parse branch runs) is exactly the untested seam; the new producer tests pin both branches, but agreed that a live connector run against a partitioned destination is the real proof. Plan: run a RedditAds report node against a test project before merge and confirm the emitted MERGE contains the `BETWEEN` clause and `total_bytes_billed` tracks the lookback window. Will post results in this thread.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Implemented, with one refinement to the proposal: only genuinely missing values (`formatColumnValue` -> `null`) are skipped, because their source rows emit `SAFE_CAST(NULL ...)` and are INSERT-only exactly as you argued. Malformed *strings* still suppress the whole predicate — a value the shape regex rejects (e.g. `'2026-08-30 10:00'`, no seconds) may still `SAFE_CAST` to a real value in the source row, so it can match a target row and must be covered by the range. Tests cover skip-one, all-null, and the malformed case.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Added — a `Date` in a TIMESTAMP column now formats as `YYYY-MM-DD HH:MM:SS` via `toISOString()` (UTC, matching what a zone-less TIMESTAMP literal denotes), aligned with the Snowflake storage. Test pins it. Good catch on the timing: pre-existing, but this PR made the omission look intentional.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Added the `constructor.name == "Date"` fallback matching the two `AbstractStorage` siblings, with a test using a duck-typed cross-realm stand-in. Agreed the extraction moment was the right time to bring it in line.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Agreed on all points: last-wins over `getSelectedFields()` makes the partition column config-order-dependent for new LinkedIn tables, and existing tables are unaffected because the predicate reads table truth. Which flag to drop is a product call — I'd propose keeping `dateRangeEnd` (matches tables created under schema-order iteration) and dropping `dateRangeStart` in a follow-up work item, since it changes new-table behavior for one source and deserves its own changeset. Filing it separately — not bundling into this PR.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Moved. `formatColumnValue` now owns `replace("T", " ")` for DATETIME/TIMESTAMP-typed columns only (STRING columns holding ISO text are untouched), so both callers share it by construction. Your framing — coincidental rather than structural — was exactly right; source rows now also emit the space form, which BigQuery parses identically.

packages/connectors/CREATING_CONNECTOR.md
prokopetsv: Added: "Any other type skips partitioning: the table is created unpartitioned and the run log records the skipped column." Pairs with the log line added in the `buildPartitionByExpression` thread.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Comment added, framed as you suggested — naming the 0001-0099 rejection as deliberate and warning that "fixing" it with `setUTCFullYear` would weaken the `'0000-00-00'` rejection.

PR #1580, opened 2026-09-01
feat: add reusable project credentials

amarchenk0:
## Summary

- add project-scoped reusable Credentials with built-in and public GitHub definitions, owners, sharing, contexts, lifecycle, validation, and dependency visibility
- integrate plugin credential requirements, explicit installation bindings, host-mediated fetch, and logical AI model handles without exposing raw secrets to plugin iframes
- add the Credentials management UI and inline credential creation in the plugin installation flow
- add database migrations, plugin SDK contracts, product and authoring documentation, and a release changeset

## Validation

- backend, web, and plugin SDK production builds pass on Node 22.22.3
- backend, web, and plugin SDK lint pass
- web type-check and plugin SDK contract type-check pass
- backend unit result: 605 suites and 9,275 tests passed
- backend E2E result: 53 suites and 686 tests passed
- web result: 257 files and 2,392 tests passed
- plugin SDK result: 5 files and 48 tests passed
- formatting, Markdown lint, diff checks, and secret-pattern scan pass

Local full Jest runs still emit the existing Snowflake AWS telemetry callback after environment teardown; the generated Jest results contain zero failed suites and zero failed tests.

## Work item

https://owox.fibery.io/Sprint_Work/Work_Item/Credentials-(Plugins)-6862


docs/plugins/authoring-guide.md
```diff
@@ -126,6 +127,58 @@ The generic does not validate the response at runtime, so validate returned data
...
+    { "id": "ai", "models": ["fast", "reasoning"] },
+    { "id": "openai", "optional": true }
+  ]
+}
+```
+
+`github`, `openai`, `anthropic`, `gemini`, and `openrouter` are built-in exact definitions. An
+object may set `optional: true`; otherwise installation requires an explicit Credential selection.
+The logical `ai` requirement may declare one or more unique capabilities from `fast`, `reasoning`,
+and `embedding`. The shorthand string `"ai"` means required `fast` AI.
```
max-voloshyn:
```
The logical `ai` requirement may declare one or more unique capabilities from `fast`, `reasoning`,
and `embedding`. The shorthand string `"ai"` means required `fast` AI.
```
BTW this is property of any credentials that implements ai protocol, not solely logical `ai`

docs/plugins/authoring-guide.md
```diff
@@ -126,6 +127,58 @@ The generic does not validate the response at runtime, so validate returned data
...
+
+An external exact requirement uses `@owner/repository`. Credential definition repositories must be
+public in v1; private plugin repositories remain supported, but private Credential definition
+repositories are rejected.
```
max-voloshyn:
```
Credential definition repositories must be
public in v1; private plugin repositories remain supported, but private Credential definition
repositories are rejected.
```
Why is so? I believe we have everything in place to support private Credential definition
repositories while re-using mechanics we already have for regular plugins

docs/plugins/credential-definitions.md
```diff
@@ -0,0 +1,96 @@
+# Credential definitions
+
+An external Credential definition describes how OWOX Data Marts authenticates requests to one
+service. It is declarative: the repository cannot provide executable adapter code, callbacks,
+expressions, environment-variable references, or a way to read the stored secret.
+
+Credential definition repositories must be public in v1. Private repositories are rejected before
+OWOX reads their releases or manifest. Supporting private definitions requires an explicit project
+visibility and authorization model and is outside v1.
```
max-voloshyn: See https://github.com/OWOX/owox-data-marts/pull/1580/changes#r3902950956

Also we should avoid "v1" in global logic

docs/plugins/credential-definitions.md
```diff
@@ -0,0 +1,96 @@
+# Credential definitions
```
max-voloshyn, line 1: Should this doc be mentioned in authoring-guide.md that is served as entry point for plugin author's agents?

docs/plugins/credential-definitions.md
```diff
@@ -0,0 +1,96 @@
...
+}
+```
+
+`credential.name` is the stable JavaScript-safe property exposed to an exact consumer, such as
+`ctx.credentials.acme`. Built-in names and `ai` are reserved. The only authentication primitive in
+v1 is one opaque secret placed in an HTTP header. Origins must use HTTPS and resolve to public
+networks; every initial request and redirect remains inside the declared origin set.
+
+`credential.documentationUrl` is optional. When present, OWOX can show it as the
+**How to get this key** link next to the write-only secret field. It must be a parseable absolute
```
max-voloshyn: key -> secret?

docs/plugins/credential-definitions.md
```diff
@@ -0,0 +1,96 @@
...
+version from `1.0.0` onward and the minor version before `1.0.0`.
+
+Compatible releases in the accepted line apply automatically. Changing display metadata, model
+catalogs, recommendations, or header-placement details can stay within a line. Adding, removing, or
+replacing origins, changing `credential.name`, removing an existing AI interface, or changing its
+adapter requires a new compatibility line. Existing Credentials pause until a maintainer accepts
+that new line; their secrets and explicit model overrides remain stored.
+
+OWOX checks known external definitions daily. A temporary GitHub failure or invalid release leaves
+the last accepted definition active.
```
max-voloshyn: Is there "Check and Update" alternative for credentials definition similarly to regular plugins?

docs/project/credentials.md
```diff
@@ -0,0 +1,57 @@
...
+A Credential belongs to the project, not to the member who created it. The creator is recorded for
+audit and becomes the first owner, but removing that member from the project does not disable the
+Credential or stop consumers that already use it. At least one owner must remain.
+
+Owners and Project Admins can manage the Credential. The **Shared for use** and **Shared for
+maintenance** settings add access for other eligible project members, subject to their role scope
+and assigned contexts. Consumers re-check access, status, and definition compatibility whenever
+they use a Credential.
+
+## Plugin use
```
max-voloshyn, line 23: Within this section should be backlinks in both directions

.changeset/6862-reusable-credentials.md
```diff
@@ -0,0 +1,10 @@
+---
+'owox': minor
+---
+
+# Reuse project Credentials across plugins
```
max-voloshyn, line 5: Leave links to added documentation within this section

packages/plugin-sdk/src/credentials.ts
```diff
@@ -0,0 +1,417 @@
...
+
+export type PluginCredentials = Readonly<
+  Record<string, CredentialHandle | AiCredentialHandle | undefined>
+> & {
+  readonly ai?: AiCredentialHandle;
+  readonly openai?: CredentialHandle;
+  readonly anthropic?: CredentialHandle;
+  readonly gemini?: CredentialHandle;
+  readonly openrouter?: CredentialHandle;
+  readonly github?: CredentialHandle;
```
max-voloshyn: Should we really list built in here? Since we have ability to works with 3rd party defined credentials may we use that resolution logic to avoid duplication built-in credentials definitions list?

max-voloshyn (changes requested):
I will add automatic review a bit later also

apps/backend/src/data-marts/credentials/use-cases/update-credential.service.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (new Set(input.ownerIds).size === 0) {
+        throw new BadRequestException('A Credential must have at least one owner');
+      }
+    }
+    if (input.availableForUse !== undefined || input.availableForMaintenance !== undefined) {
+      await this.assertAccess(id, context, Action.CONFIGURE_SHARING);
+    }
+
+    const credential = await this.credentials.getByIdAndProjectId(id, context.projectId);
+    const definition = await this.definitions.getForView(credential);
```
max-voloshyn: 🔴 Blocking: Secret replacement resolves the current external definition with `getForView`, even when its compatibility line has not been accepted. The validation probe can therefore send the replacement secret to a newly changed validation origin/header before a maintainer consents to that outbound contract; normal validation correctly uses `getForCredential` and rejects this state. Resolve the accepted runtime contract here as well, or reject secret rotation until consent, and add a regression test covering an external definition whose compatibility line changed.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/pages/plugins/gallery/PluginsGalleryPage.tsx
```diff
@@ -95,8 +95,11 @@ export default function PluginsGalleryPage() {
...
    */
-  const installPlugin = async (plugin: PluginGalleryEntry) => {
-    const stale = await install(plugin.pluginId, plugin.currentVersionId);
+  const installPlugin = async (
+    plugin: PluginGalleryEntry,
+    credentialSelections: Readonly<Record<string, string | null>>
+  ) => {
+    const stale = await install(plugin.pluginId, plugin.currentVersionId, credentialSelections);
     if (stale) {
       setCandidate({
```
max-voloshyn: 🔴 Blocking: On a stale-version response this reopens the dialog by copying the old `plugin` and replacing only the version id/SemVer, so `credentialRequirements` still describe the version the server just rejected. The user can then confirm a new version against stale grants (and, for compatible credential IDs, the backend may accept a capability the dialog never showed). Refetch the complete current gallery entry—including credential requirements—before reopening the dialog, and apply the same fix to the details/history retry paths.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/plugins/services/plugins.service.ts
```diff
@@ -52,8 +52,15 @@ class PluginsService extends ApiService {
...
-  async install(pluginId: string, expectedVersionId: string | null): Promise<PluginInstallation> {
-    return this.post<PluginInstallation>(`/${pluginId}/installation`, { expectedVersionId });
+  async install(
+    pluginId: string,
+    expectedVersionId: string | null,
+    credentialSelections: Readonly<Record<string, string | null>> = {}
+  ): Promise<PluginInstallation> {
+    return this.post<PluginInstallation>(`/${pluginId}/installation`, {
+      expectedVersionId,
+      credentialSelections,
```
max-voloshyn: 🔴 Blocking: This now sends `credentialSelections: {}` for every install, including plugins with no credential requirements. During a web-first rolling deployment, the previous backend's global `ValidationPipe` has `forbidNonWhitelisted: true`, so it rejects this new field and all installs/reinstalls/restores fail. Omit the property when the selection map is empty (preserving compatibility for existing plugins), and add a mixed-version request test.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/facades/credential-consumer-binding.facade.impl.ts
```diff
@@ -0,0 +1,323 @@
...
+    const ready: NormalizedCredentialRequirement[] = [];
+    for (const [index, requirement] of requirements.entries()) {
+      const binding = await this.bindings.findOneBy({
+        consumerType: request.consumerType,
+        consumerId: request.consumerId,
+        projectId: request.projectId,
+        requirementKey: requirement.key,
+        active: true,
+      });
+      if (!binding && requirement.optional) continue;
```
max-voloshyn, line 215: 🔴 Blocking: Optional requirements are skipped only when no binding exists. If a user selected an optional credential and it later becomes disabled, inaccessible, or pending definition consent, `resolveBinding` throws and the plugin cannot open at all—even though the SDK contract says optional handles may be absent until usable. Treat expected unusable states for optional bindings as an omitted handle while preserving errors for required credentials and unexpected failures, and add lifecycle tests for disabled/access-revoked/consent-pending optional credentials.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/services/credential-external-definition-registry.service.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  @Transactional()
+  private async persist(
+    definitionId: string,
+    input: RegisterExternalCredentialDefinitionInput
+  ): Promise<ResolvedCredentialDefinition> {
+    const query = this.definitions
+      .createQueryBuilder('definition')
+      .where('definition.id = :definitionId', { definitionId });
+    if (this.dataSource.options.type !== 'sqlite') query.setLock('pessimistic_write');
```
max-voloshyn: 🔴 Blocking: The default SQLite configuration uses TypeORM's `better-sqlite3` driver, so this condition calls `setLock('pessimistic_write')` on a driver that does not support pessimistic locks. The unit test masks the problem by supplying the non-production type string `sqlite`, which means registering/syncing an external definition fails in the default deployment. Skip the lock for `better-sqlite3` (or positively gate it to supported drivers) and cover this path with a real `better-sqlite3` integration test.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/credentials/components/CredentialConfigSheet/CredentialForm.tsx
```diff
@@ -0,0 +1,739 @@
...
+    }
+
+    if (!isEditMode && values.secret.length === 0) {
+      form.setError('secret', { type: 'required', message: 'Secret is required' });
+      return;
+    }
+
+    const { mappings, modes } = serializeAiMappings(values);
+    if (credential) {
+      await onSubmit({
```
max-voloshyn: 🔴 Blocking: Every edit submits owner and sharing fields, but the backend requires `MANAGE_OWNERS` whenever `ownerIds` is present and `CONFIGURE_SHARING` whenever either sharing field is present. A maintenance-shared non-owner intentionally has `EDIT` without those capabilities, so even a title-only save from this form gets a 403. Submit only dirty fields and disable/hide controls the caller cannot manage; add a title-only maintenance-user test that verifies the restricted fields are omitted.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/plugins/runtime/pluginHostBridge.ts
```diff
@@ -264,15 +368,27 @@ export function createPluginHostBridge(options: PluginHostBridgeOptions): Plugin
...
       // retry. This both bounds validation work and prevents a second serialization from
       // failing only after a credential has been minted.
       const serializedBody = serializeJsonBody(request);
-      const response = await forward(request, serializedBody);
+      const response = await forward(request, serializedBody, signal);
       reply(response, 'stream' in response ? [response.stream] : []);
     } catch (caught) {
       reply({ id, ok: false, error: asErrorPayload(caught) });
     } finally {
+      if (requestController) requestControllers.delete(id);
```
max-voloshyn, line 392: 🔴 Blocking: The request controller is removed and `inFlight` is decremented immediately after the stream is transferred, not when that stream closes, errors, or is cancelled. The SDK also removes its abort listener once the response envelope arrives, so aborting after headers cannot reach the host/provider, and long-lived paid AI streams no longer count toward `MAX_IN_FLIGHT`. Keep the correlation/controller and admission slot for the transferred stream's lifetime, propagate post-header cancellation, and test abort plus the concurrency cap with a never-ending stream.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/plugin-host/controllers/plugin-credential-runtime.controller.ts
```diff
@@ -0,0 +1,257 @@
...
+  }
+}
+
+function requestAbortSignal(request: Request): {
+  readonly signal: AbortSignal;
+  dispose(): void;
+} {
+  const controller = new AbortController();
+  const abort = () => controller.abort();
+  request.once('aborted', abort);
```
max-voloshyn, line 249: 🔴 Blocking: Runtime cancellation listens only to the inbound request's `aborted` event. After Nest consumes the POST body, a browser disconnect while waiting for the provider normally closes/errors the response instead, so credential-authenticated work continues; the stream loop can also wait forever for `drain` after that close. Abort on response close/error as well, race backpressure waits against close/error/the abort signal, and add disconnect tests for both pre-stream provider work and a backpressured stream.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/use-cases/delete-credential.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    ) {
+      throw new ForbiddenException('You do not have permission to delete this Credential');
+    }
+
+    const credential = await this.credentials.lockActiveByIdAndProjectId(id, context.projectId);
+    if (!credential) {
+      throw new NotFoundException(`Credential ${id} was not found`);
+    }
+
+    const references = await this.credentials.countActiveBindings(id);
```
max-voloshyn, line 36: 🔴 Blocking: Active bindings are counted here, but removing a project member only clears context bindings; it does not uninstall that member's plugin installations or deactivate their new credential bindings. The departed member can no longer authenticate to uninstall, and an admin cannot uninstall another member's installation, so these references can permanently prevent credential deletion. Deactivate the removed member's plugin consumers/bindings during member cleanup (or provide an authorized admin cleanup path) and add a member-removal regression test.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/mappers/credential.mapper.ts
```diff
@@ -0,0 +1,84 @@
...
+      projectId: credential.projectId,
+      title: credential.title,
+      definition: mapCredentialDefinitionToApiDto(definition),
+      definitionConsentRequired:
+        definition.source === 'external' &&
+        definition.compatibilityLine !== credential.acceptedCompatibilityLine,
+      enabled: credential.enabled,
+      availableForUse: credential.availableForUse,
+      availableForMaintenance: credential.availableForMaintenance,
+      validationState: validation?.state ?? 'unknown',
```
max-voloshyn: 🟡 Non-blocking: Validation state exists only in the immediate mutation response; there are no persisted validation fields, and normal list/get mapping always falls back to `unknown`. The web invalidates and refetches immediately after validation, so verified/rejected state disappears and the new validation filters cannot return durable results. Persist state/message/timestamp and map them on normal reads, or remove the persistent-looking fields/filter if validation is intentionally transient; add a reload test.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/plugin-host/services/external-credential-definition-sync.service.ts
```diff
@@ -0,0 +1,130 @@
...
+@Injectable()
+export class ExternalCredentialDefinitionSyncService {
+  constructor(
+    private readonly github: GithubApiService,
+    private readonly registry: CredentialExternalDefinitionRegistryService
+  ) {}
+
+  async syncLocator(locator: string): Promise<ResolvedCredentialDefinition> {
+    const ref = parseExternalLocator(locator);
+    const repo = await this.github.getRepo(ref, GithubReadPolicy.PUBLIC_ONLY);
```
max-voloshyn: 🔴 Blocking: Credential-definition discovery forces `PUBLIC_ONLY` access even though ordinary plugin discovery uses the configured GitHub App path for private repositories. A private plugin can therefore be discovered successfully but fail publication as soon as it declares a Credential definition from a private repository. Use the configured repository access policy here while preserving the existing private-repository disclosure rules, and cover a private plugin whose requirement resolves through the GitHub App.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/facades/credential-consumer-binding.facade.impl.ts
```diff
@@ -0,0 +1,323 @@
...
+    const definition = await this.definitions.getForCredential(credential);
+    if (requirement.definitionId && credential.definitionId !== requirement.definitionId) {
+      throw new BadRequestException(
+        `Credential ${credential.id} does not satisfy ${requirement.key}`
+      );
+    }
+    if (!requirement.definitionId && !definition.contract.ai) {
+      throw new BadRequestException(`Credential ${credential.id} is not an AI Credential`);
+    }
+    const aiModelMappings = resolveCredentialAiModelMappings(
```
max-voloshyn, line 282: 🔴 Blocking: Readiness treats any non-empty stored override as valid without checking whether that model still exists in the current definition catalog or has the required language/embedding kind. After a compatible catalog update removes an overridden model, consumers remain marked ready and fail only at provider runtime. Preserve the override for repair, but make affected handles require setup until a maintainer chooses a valid catalog model or returns to recommended.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/services/external-credential-manifest.ts
```diff
@@ -0,0 +1,133 @@
...
+        rejectedStatuses: z.array(z.number().int().min(400).max(499)).min(1).max(20).optional(),
+      })
+      .optional(),
+    ai: z
+      .object({
+        adapter: z.object({
+          type: z.enum(['openai', 'anthropic', 'google', 'openrouter', 'openai-compatible']),
+          baseUrl: z.string().url().startsWith('https://'),
+        }),
+        models: z
```
max-voloshyn: 🔴 Blocking: An external definition can declare `ai` while omitting both model catalogs and all recommended mappings; even `fast` and `reasoning` are optional strings and are not checked against the declared language catalog. Such a release passes validation without the portable baseline needed to initialize a usable AI Credential. Require language catalogs plus valid `fast` and `reasoning` recommendations, and validate embedding against the embedding catalog when supplied.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/plugin-host/services/external-credential-definition-sync.service.ts
```diff
@@ -0,0 +1,130 @@
...
+          contract: parsed.contract,
+        });
+      } catch (error) {
+        rejections.push(
+          `${candidate.semver}: ${error instanceof Error ? error.message : 'definition is invalid'}`
+        );
+      }
+    }
+
+    const current = await this.registry.getCurrentByGithubRepoId(repo.githubRepoId);
```
max-voloshyn, line 80: 🟡 Non-blocking: When the newest same-line release is detectably incompatible, this loop records the registry error, falls back to an older/current release, and returns success, so the required guidance to publish a new compatibility line never reaches the publisher or scheduled-sync outcome. Preserve the current accepted version, but surface the rejected release and actionable compatibility-line guidance instead of treating fallback as an unqualified success.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/features/credentials/CredentialsTable.tsx
```diff
@@ -0,0 +1,357 @@
...
+          <Button
+            variant='ghost'
+            className={`dm-card-table-body-row-actionbtn opacity-0 transition-opacity ${isOpen ? 'opacity-100' : 'group-hover:opacity-100'}`}
+            aria-label={`Open actions for ${credential.title}`}
+          >
+            <MoreHorizontal className='dm-card-table-body-row-actionbtn-icon' />
+          </Button>
+        </DropdownMenuTrigger>
+        <DropdownMenuContent align='end'>
+          <DropdownMenuItem
```
max-voloshyn: 🟡 Non-blocking: Every viewer gets Validate, Replace secret, Edit, Enable/Disable, Accept update, and Delete actions, including members who intentionally have only `SEE`/`USE`. The backend denies these mutations, but the UI repeatedly offers workflows that cannot succeed. Return or derive action capabilities and hide or disable each action the viewer cannot perform.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/services/external-credential-manifest.ts
```diff
@@ -0,0 +1,133 @@
+import { z } from 'zod';
+import {
+  CredentialDocumentationUrlSchema,
+  type CredentialDefinitionContract,
+} from '../credential.types';
+
+const RESERVED_NAMES = new Set(['ai', 'openai', 'anthropic', 'gemini', 'openrouter', 'github']);
```
max-voloshyn: 🔴 Blocking: External definitions accept `then` as `credential.name`, but the SDK proxy always returns `undefined` for that property to avoid Promise assimilation. The release is therefore considered eligible even though its declared exact handle can never be used. Reserve `then` and any other proxy-conflicting runtime names during definition validation, and add a manifest-to-SDK regression test.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/services/builtin-credential-definitions.ts
```diff
@@ -0,0 +1,131 @@
...
+    documentationUrl: 'https://platform.openai.com/docs/quickstart',
+    auth: { type: 'header', label: 'API key', headerName: 'authorization', prefix: 'Bearer ' },
+    origins: ['https://api.openai.com'],
+    validation: { method: 'GET', path: '/v1/models' },
+    ai: {
+      adapter: 'openai',
+      baseUrl: 'https://api.openai.com/v1',
+      models: {
+        language: [
+          { id: 'gpt-4.1-mini', name: 'GPT-4.1 mini' },
```
max-voloshyn: 🔴 Blocking: The built-in OpenAI catalog initializes `fast` and `reasoning` with `gpt-4.1-mini` and `o4-mini`, while the current Credential product contract defines the GPT-5.6 catalog and mappings. New OpenAI Credentials will therefore start with model selections different from the agreed defaults. Align the built-in catalog and recommendations with the current contract, and cover the initialized mappings in the built-in-definition test.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/services/external-credential-manifest.ts
```diff
@@ -0,0 +1,133 @@
...
+  'transfer-encoding',
+]);
+const JS_IDENTIFIER = /^[A-Za-z_$][A-Za-z0-9_$]*$/;
+
+const ModelSchema = z.object({
+  id: z.string().trim().min(1).max(255),
+  name: z.string().trim().min(1).max(255),
+});
+
+const ExternalCredentialManifestSchema = z.object({
```
max-voloshyn: 🟡 Non-blocking: These nested Zod objects are not strict, so unsupported definition fields are silently stripped and the release remains eligible. That makes a publisher’s unsupported behavior disappear without an actionable error. Reject unknown fields throughout the external definition schema and report the offending field path.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/data-marts/credentials/use-cases/create-credential.service.ts
```diff
@@ -0,0 +1,183 @@
...
+  ai: CredentialDefinitionContract['ai'] | undefined,
+  current?: {
+    readonly mappings: CredentialAiModelMappings | null;
+    readonly modes: CredentialAiModelMappingModes | null;
+  }
+): {
+  readonly mappings: CredentialAiModelMappings | null;
+  readonly modes: CredentialAiModelMappingModes | null;
+} {
+  if (mappings === null) {
```
max-voloshyn: 🟡 Non-blocking: Passing `aiModelMappings: null` clears every mapping, while arbitrary non-empty keys are accepted and inferred as overrides. AI Credentials require exact `fast` and `reasoning` mappings plus optional `embedding`, initialized from recommendations. Validate the authoritative API shape and model kind, require the two language mappings for AI definitions, and reserve `null` for non-AI Credentials.

🤖 Codex (GPT-5.6 Sol High)

apps/web/src/pages/plugins/detail/PluginDetailsPage.tsx
```diff
@@ -172,12 +175,7 @@ export default function PluginDetailsPage() {
                 if (isInstalling) {
                   return;
                 }
-                // Live install: reinstall in place without re-confirming.
-                // not_installed / uninstalled: same Install + dialog path (backend reactivates).
-                if (showReinstall) {
-                  void installPlugin(plugin);
-                  return;
-                }
+                // Every install/restore/reconfigure uses the same explicit Credential grant flow.
```
max-voloshyn, line 175: 🔴 Blocking: Every installed-plugin Credential reconfiguration now enters this confirmation flow, but the only action remains labeled `Reinstall` and the dialog describes reinstall behavior. A member who only needs to select another eligible Credential is told a reinstall is required even though replacing the installation bindings restores readiness. Separate Credential configuration from a genuine reinstall, or label the action and confirmation copy according to the operation being performed.

🤖 Codex (GPT-5.6 Sol High)

apps/backend/src/plugin-host/controllers/credential-definitions-github.controller.ts
```diff
@@ -0,0 +1,34 @@
...
+  repository: string;
+}
+
+@ApiTags('Credentials')
+@Controller('credentials/definitions')
+@RejectPluginAuth()
+export class CredentialDefinitionsGithubController {
+  constructor(private readonly sync: ExternalCredentialDefinitionSyncService) {}
+
+  @Auth(Role.viewer(Strategy.INTROSPECT))
```
romandubovyi, line 24: 🔴 **Blocking:** This is the only mutating endpoint in `plugin-host` that takes no `@AuthContext()` and performs no capability check. `Role.viewer(Strategy.INTROSPECT)` matches the project convention — `plugin-installations`, `plugin-publications`, `plugin-admin` and every route in `credentials.controller.ts` use it too — but each of those passes the context into a use case that authorizes the operation. Here any project member reaches `sync.syncLocator()` directly, which writes to `credential_external_definition`: a table with no `projectId`, so the row is deployment-wide. The registry exposes no delete path, and every registered definition is scheduled into the daily `CredentialDefinitionUpdateCheckProcessor` refresh permanently, consuming the shared anonymous GitHub quota. Please authorize this the way the sibling endpoints do before syncing.

🤖 Generated with Kimi CLI

apps/backend/src/plugin-host/use-cases/sync-plugin-releases.service.ts
```diff
@@ -257,6 +263,28 @@ export class SyncPluginReleasesService {
...
+        }
+        credentialRequirements = await this.externalCredentialDefinitions.resolveRequirements(
+          manifest.manifest.credentials
+        );
+      }
+    } catch (error) {
+      into.rejections.push(
+        this.rejection(
+          release,
+          ReleaseRejectionCode.MANIFEST_SCHEMA,
```
romandubovyi: ⚪ **Non-blocking:** This `catch` maps every failure of `resolveRequirements` to `ReleaseRejectionCode.MANIFEST_SCHEMA`, including `GithubRateLimitedError`, network failures, and a 404 on the third-party definition repository. `MANIFEST_SCHEMA` is marked publisher-actionable in `REJECTION_IS_PUBLISHER_ACTIONABLE`, so a transient GitHub outage is surfaced to the plugin publisher as "your manifest is invalid" while an otherwise valid release is rejected for this sync run. The enum's own comment draws exactly this line ("nothing on our side failed — the publisher has something to fix"). Please separate genuine contract errors (unknown requirement id, duplicate resolved handle) from transient external-sync failures — either let rate-limit/transport errors propagate the way the rest of the sync handles them, or add a non-actionable rejection code for them.

apps/backend/src/data-marts/credentials/services/credential-fetch.service.ts
```diff
@@ -0,0 +1,281 @@
...
+      if (
+        name.length === 0 ||
+        name.length > 255 ||
+        typeof value !== 'string' ||
+        value.length > 8192 ||
+        /[\0\r\n]/.test(value)
+      ) {
+        throw new BadRequestException('Credential request contains an invalid header');
+      }
+      headers.set(name, value);
```
romandubovyi, line 183: ⚪ **Non-blocking:** Header names are checked for length only, while values are checked for `\0\r\n` and the HTTP method is checked against the full RFC token charset a few lines above in this same file. `Headers.set()` throws a `TypeError` for a name containing a space or a colon, and this call is not wrapped, so a plugin sending `headers: { "bad name": "x" }` gets a 500 rather than a 400. `pluginHostBridge.ts` validates header names by length only as well, so such a name reaches the backend. Reusing the same token regex for header names here would close it.

packages/plugin-sdk/src/iframe-transport.ts
```diff
@@ -53,33 +59,67 @@ export function createIframeTransport(port: MessagePort): OWOXTransportWithLowLe
...
       const isStream = 'stream' in request && request.stream === true;
 
+      const cancel = () => {
+        port.postMessage({
+          id: crypto.randomUUID(),
+          kind: 'cancel',
+          targetId: id,
+        } satisfies PluginRequest);
+      };
       const timer = isStream
```
romandubovyi, line 99: ⚪ **Non-blocking:** Non-stream `credentialAi` calls take the 30s `REQUEST_TIMEOUT_MS` used for ordinary API requests, while `CredentialAiService` gives the provider `AI_TIMEOUT_MS = 120_000`. Only `doStream` sets `stream: true`; `doGenerate` and `doEmbed` do not, so a `generateText` on a slower reasoning model that the backend would have answered in 45s is rejected client-side as `TIMEOUT` at 30s and a `cancel` is posted for nothing. The comment above `REQUEST_TIMEOUT_MS` ties it to the web app's axios timeout, which is a different budget from the AI path. Consider exempting `credentialAi` from this timer or raising it to the backend AI limit, with a test that a generate still pending at 30s is not rejected.

apps/backend/package.json
```diff
@@ -64,6 +64,10 @@
     "dump:apply": "ts-node commands/apply-dump-command.ts"
   },
   "dependencies": {
+    "@ai-sdk/anthropic": "4.0.44",
```
romandubovyi, line 67: ⚪ **Non-blocking:** `credential-ai.service.ts` imports `LanguageModelV4`, `EmbeddingModelV4`, `LanguageModelV4CallOptions` and friends from `@ai-sdk/provider`, but that package is not in `apps/backend`'s `dependencies` — only the four provider packages are. It currently resolves through hoisting from `packages/plugin-sdk` (pinned to `4.0.8`) and from the provider packages' own ranges. The import is type-only so runtime is unaffected, but `tsc` still has to resolve it: a range bump in any of those, or a different install layout, either breaks the backend build or silently resolves a different V4 contract version than the SDK compiles against. Declaring `@ai-sdk/provider` here at the same pinned version as the plugin SDK would make it explicit.

romandubovyi (commented):
Multi-agent review of the reusable Credentials feature. Three CLI agents analysed the PR head independently; every finding below was re-verified against the code before posting, and points already covered by the existing review comments were dropped.

Five points: one blocking authorization gap on the external definition registration endpoint, and four non-blocking items (rejection-code classification, header validation, an SDK/backend AI timeout mismatch, and an undeclared backend dependency).

🤖 Generated with Kimi CLI + Grok CLI (Grok 4.6) + Claude Code (Opus 5)

apps/backend/src/plugin-host/controllers/plugin-credential-runtime.controller.ts
```diff
@@ -0,0 +1,257 @@
...
+  @Post(':handle/fetch')
+  @ApiOkResponse({ type: CredentialFetchResponseApiDto })
+  async fetch(
+    @Param('handle') handle: string,
+    @Body() input: CredentialFetchRequestApiDto,
+    @AuthContext() context: AuthorizationContext,
+    @Req() request: Request
+  ): Promise<CredentialFetchResponseApiDto> {
+    const binding = await this.resolveBinding(handle, context, 'fetch');
+    const disconnected = new AbortController();
```
romandubovyi: ⚪ **Non-blocking:** `fetch()` manually instantiates an `AbortController` and registers a `request.once('aborted')` handler, duplicating the `requestAbortSignal(request)` helper already used by `generate()`, `embed()`, and `stream()` on lines 86, 113, and 141. Furthermore, the 'aborted' event on `http.IncomingMessage` is deprecated in Node.js in favor of 'close\. Reusing `requestAbortSignal(request)` here (and listening on 'close\) would keep request-cancellation handling consistent across all runtime endpoints.

🤖 Generated with Antigravity (Gemini 3.7 Flash)

packages/plugin-sdk/src/credentials.ts
```diff
@@ -0,0 +1,417 @@
...
+      return false;
+    },
+  });
+}
+
+function createExactHandle(requester: IframeRequester, handle: string): CredentialHandle {
+  return {
+    asFetch: () => async (input, init) => {
+      const request = new Request(input, init);
+      if (request.signal.aborted) throw request.signal.reason;
```
romandubovyi: ⚪ **Non-blocking:** If `request.signal.aborted` is true, `asFetch()` throws `request.signal.reason` directly. When aborted via a plain `AbortController.abort()` call without a specified reason, `request.signal.reason` can be `undefined` in standard DOM/fetch implementations, which causes the promise to reject with `undefined` rather than an `AbortError` `DOMException`. Falling back to `request.signal.reason ?? new DOMException('The request was aborted', 'AbortError')` (matching `iframe-transport.ts:74`) ensures rejected promises always carry an Error instance.

🤖 Generated with Antigravity (Gemini 3.7 Flash)

apps/backend/src/data-marts/credentials/use-cases/validate-credential.service.ts
```diff
@@ -0,0 +1,39 @@
...
+        EntityType.CREDENTIAL,
+        id,
+        Action.EDIT,
+        context.projectId
+      ))
+    ) {
+      throw new ForbiddenException('You do not have permission to validate this Credential');
+    }
+
+    const definition = await this.definitions.getForCredential(credential);
```
romandubovyi, line 35: ⚪ **Non-blocking:** `ValidateCredentialService.run` calls `this.definitions.getForCredential(credential)`, which throws a `ConflictException` ('Credential definition update requires maintainer consent') if `external.currentCompatibilityLine !== credential.acceptedCompatibilityLine`. As a result, when an upstream definition releases a new compatibility line, triggering the on-demand probe via the "Validate" action immediately 409s rather than verifying the secret against the provider. Using `getForView(credential)` (the way `UpdateCredentialService:52` does) would allow validating the credential even when an upstream definition update is pending consent.

🤖 Generated with Antigravity (Gemini 3.7 Flash)

apps/backend/src/common/typeorm/query-error.utils.ts
```diff
@@ -1,6 +1,13 @@
 import { QueryFailedError } from 'typeorm';
 
-const UNIQUE_CONSTRAINT_CODES = new Set(['23505', 'ER_DUP_ENTRY', 'ER_DUP_KEY', '1062']);
+const UNIQUE_CONSTRAINT_CODES = new Set([
+  '23505',
+  'ER_DUP_ENTRY',
+  'ER_DUP_KEY',
+  '1062',
+  'SQLITE_CONSTRAINT',
```
romandubovyi: ⚪ **Non-blocking:** `SQLITE_CONSTRAINT` is SQLite's generic constraint code — it covers FOREIGN KEY, NOT NULL and CHECK failures, not just unique violations, which `SQLITE_CONSTRAINT_UNIQUE` on the next line already handles. Since this is a shared helper, the blast radius is outside this feature: `insertVersion` maps a unique violation to `PluginVersionConflictError` and `ai-source-apply.service.ts` swallows one and reloads, so both would now treat a foreign-key or NOT NULL failure as "the row already exists". The message fallback below (`duplicate` / `unique constraint` / `violates unique`) still catches drivers that only report a generic code, so dropping the generic entry shouldn't lose real unique detection.

apps/backend/src/data-marts/credentials/services/credential-external-definition-registry.service.ts
```diff
@@ -0,0 +1,251 @@
...
+  return major === 0 ? `0.${minor}` : String(major);
+}
+
+function isCompatibleWithinLine(
+  previous: CredentialDefinitionContract,
+  next: CredentialDefinitionContract
+): boolean {
+  return (
+    previous.id === next.id &&
+    sameSet(previous.origins, next.origins) &&
```
romandubovyi, line 234: ⚪ **Non-blocking:** `isCompatibleWithinLine` treats a same-line release as compatible when `id`, `origins`, `auth.type` and the AI adapter are unchanged, but `auth.headerName`, `auth.prefix` and `ai.baseUrl` are not part of the check. A 1.x patch can therefore change which header carries the stored secret, its prefix, or the AI provider path, and become current with no `definitionConsentRequired` prompt. The origin allowlist is unchanged so the secret does not reach a new host, but these are exactly the fields that describe how the secret is attached — the part of the outbound contract a maintainer would want to consent to. Consider including them in the same-line check, or treating a change to them as a compatibility-line bump.

apps/backend/src/data-marts/credentials/use-cases/create-credential.service.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+    return { mappings: null, modes: null };
+  }
+
+  const resolvedMappings: CredentialAiModelMappings = {
+    ...(mappings ?? current?.mappings ?? ai?.recommended ?? {}),
+  };
+  validateAiMappings(resolvedMappings, ai);
+  const resolvedModes: CredentialAiModelMappingModes = {
+    ...(modes ?? current?.modes ?? inferAiMappingModes(resolvedMappings, ai?.recommended)),
```
romandubovyi: ⚪ **Non-blocking:** When only `aiModelMappings` is sent, `resolvedModes` falls back to `current?.modes`, and the loop below then overwrites `resolvedMappings[key]` with `ai.recommended[key]` for every key still in `recommended` mode. `UpdateCredentialService` explicitly allows either field on its own (`input.aiModelMappings !== undefined || input.aiModelMappingModes !== undefined`), so `PUT /credentials/{id}` with `{"aiModelMappings": {"fast": "my-model"}}` on a Credential whose `fast` mapping is `recommended` returns 200 while persisting the old recommended model — the caller's override is discarded with no error. The web form only hides this because it always submits both fields. Either flip a key to `override` when a new mapping value arrives without a mode, or reject a mappings-only update for keys currently in `recommended` mode.

apps/backend/src/plugin-host/dto/presentation/plugin-installation-api.dto.ts
```diff
@@ -14,6 +14,15 @@ export class InstallPluginApiDto {
...
   expectedVersionId: string | null;
+
+  @ApiPropertyOptional({
+    type: 'object',
+    additionalProperties: { type: 'string', nullable: true },
+    description: 'Credential requirement handle to selected project Credential id.',
+  })
+  @IsOptional()
+  @IsObject()
+  credentialSelections?: Record<string, string | null>;
```
romandubovyi, line 31: ⚪ **Non-blocking:** `credentialSelections` is validated only as `@IsObject()`, so the values are unconstrained — `whitelist: true` strips unknown top-level DTO properties but not the contents of this map. `replaceBindings` rejects undeclared keys and falsy values, but a non-string truthy value under a declared key flows into `CredentialService.lockActiveByIdAndProjectId` and then into `.where('id = :id', { id })`. `POST /plugins/{id}/installation` with `{"credentialSelections": {"github": ["a", "b"]}}` makes TypeORM expand the parameter into a value list and the query fails at the driver — a 500 where the client should get a 400. Validating the map shape at the boundary (each value `string | null`, plus an entry cap) would match how `CredentialFetchRequestApiDto.headers` values are re-checked in `buildRequestHeaders`.

apps/web/src/features/credentials/CredentialsTable.tsx
```diff
@@ -0,0 +1,357 @@
...
+  ];
+}
+
+function CredentialAvailability({ credential }: { credential: Credential }) {
+  return (
+    <div className='flex flex-wrap gap-1'>
+      <Badge variant={credential.enabled ? 'secondary' : 'outline'}>
+        {credential.enabled ? 'Enabled' : 'Disabled'}
+      </Badge>
+      {credential.availableForUse && <Badge variant='outline'>Shared for use</Badge>}
```
romandubovyi: ⚪ **Non-blocking:** The availability cell renders a badge for `availableForUse` but none for `availableForMaintenance`, which is the more privileged of the two: in `access-matrix.config.ts`, `credentialNonOwnerRules` grants `SEE` + `USE` for `SHARED_FOR_USE` but `SEE` + `USE` + `EDIT` + `DELETE` for `SHARED_FOR_MAINTENANCE` and `SHARED_FOR_BOTH`. `CredentialsTableFilters.ts` already derives a `sharing` filter from `classifyAvailability(availableForUse, availableForMaintenance)`, so a member can filter for a state no row ever displays. A badge for it (and for the both-flags case) would make the riskier sharing state visible at a glance.

packages/plugin-sdk/src/credentials.ts
```diff
@@ -0,0 +1,417 @@
...
+  declaredHandles: readonly (string | PluginCredentialHandleDescriptor)[]
+): PluginCredentials {
+  const descriptors = new Map(
+    declaredHandles.map(handle => [
+      typeof handle === 'string' ? handle : handle.name,
+      typeof handle === 'string' ? ({ name: handle, kind: 'exact' } as const) : handle,
+    ])
+  );
+  const created = new Map<string, CredentialHandle | AiCredentialHandle>();
+  return new Proxy(Object.create(null) as Record<string, CredentialHandle | AiCredentialHandle>, {
```
romandubovyi, line 92: ⚪ **Non-blocking:** The proxy defines only `get` and `set`; with `Object.create(null)` as the target and no `has`/`ownKeys` traps, `'github' in ctx.credentials` is always `false`, `Object.keys(ctx.credentials)` is always `[]`, and `JSON.stringify(ctx.credentials)` is `{}` — even for handles that are declared and configured. The exported type is `Readonly<Record<string, CredentialHandle | AiCredentialHandle | undefined>>`, which invites exactly those checks. The property-access form (`if (ctx.credentials.github)`) works fine, so this is about making discovery match the declared type: `has` and `ownKeys`/`getOwnPropertyDescriptor` traps backed by `descriptors` would do it.

docs/plugins/authoring-guide.md
amarchenk0: Keeping this public-only in the current scope is an explicit product decision. Plugin repositories may still be private, but reusable Credential definition repositories are intentionally limited to public repositories for now. The restriction is isolated behind `GithubReadPolicy.PUBLIC_ONLY`, so private-definition support can be added later without changing the stored definition or binding contracts.

packages/plugin-sdk/src/credentials.ts
amarchenk0: I tested removing the explicit built-in properties. Runtime resolution still works, but TypeScript widens `credentials.github` through the index signature to `CredentialHandle | AiCredentialHandle | undefined`, so `credentials.github.asFetch()` no longer compiles. These properties are type-only ergonomic aliases; runtime resolution remains descriptor-driven and custom definitions do not use a duplicated runtime list. I added a comment explaining this.

apps/backend/src/data-marts/credentials/use-cases/delete-credential.service.ts
amarchenk0: We should not deactivate consumers solely because the Credential creator or owner leaves the project. The agreed product model is that the Credential and its bindings belong to the project; owner and `createdById` are maintenance and audit metadata, and existing consumers must continue working. The orphaned personal-installation concern is valid, but it needs an explicit project-admin cleanup or transfer capability. Automatically deleting bindings during member removal would violate the accepted lifecycle rule, so I am leaving this thread open for that follow-up decision.

apps/backend/src/plugin-host/services/external-credential-definition-sync.service.ts
amarchenk0: This is intentional for the current scope, not a missing GitHub App path. Private plugin repositories remain supported, while third-party Credential definition repositories are public-only by product decision. The service rejects private definitions explicitly. Private support can later switch the read policy and add the required disclosure and authorization coverage without changing Credential Core.

apps/backend/src/data-marts/credentials/use-cases/validate-credential.service.ts
amarchenk0: We intentionally keep `getForCredential()` here. A pending compatibility line has not been accepted, so using `getForView()` would send the stored secret according to an unaccepted outbound contract. The UI now hides Validate and Replace secret while definition consent is pending; after consent, validation uses the accepted contract. Metadata-only edits remain available.

apps/backend/src/data-marts/credentials/services/builtin-credential-definitions.ts
max-voloshyn: @amarchenk0 please make sure models lists contains currently actual models and recomendations (not only for openai but for other providers also)

PR #1581, opened 2026-09-02
fix(web): open OWOX BI from the project menu in the current project

romandubovyi:
## Problem

The project menu **OWOX BI** item always opened the BI root (`https://bi.owox.com/`), even though ODM and OWOX BI share the same project id.

## Expected

When the user is in an ODM project, the menu item should open that same project in BI.

Example:

- ODM: https://app.owox.com/ui/2efa3556b1d5bec2f99b6fca4c0010db
- OWOX BI: https://bi.owox.com/ui/app/2efa3556b1d5bec2f99b6fca4c0010db

If the current project id is unknown, the link still opens `https://bi.owox.com/`.

## Fix

- Add optional `buildHref` on project menu items
- Resolve the OWOX BI href in `useProjectMenu` from `useProject().id`
- Keep the existing static href as the fallback

## Testing

- Hook tests cover the project-scoped URL and the root fallback
- `npm test -w @owox/web -- src/components/AppSidebar/ProjectMenu`
- ESLint, Prettier, and `tsc --noEmit` for `@owox/web`

### DIFF
--- a/apps/web/src/components/AppSidebar/ProjectMenu/items.ts
+++ b/apps/web/src/components/AppSidebar/ProjectMenu/items.ts
@@ -55,6 +55,7 @@ export const projectMenuItems: ProjectMenuItem[] = [
     type: 'menu-item',
     title: 'OWOX BI',
     href: 'https://bi.owox.com/',
+    buildHref: projectId => `https://bi.owox.com/ui/app/${projectId}`,
     icon: OWOXBIIcon,
     visible: { flagKey: 'MENU_OWOX_BI_VISIBLE', expectedValue: 'true' },
     group: 'external',
--- a/apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.test.ts
+++ b/apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.test.ts
@@ -15,7 +15,7 @@ const flagsState = vi.hoisted(() => ({
     flags: {
       MENU_OWOX_BI_VISIBLE: 'true',
     } as Record<string, unknown>,
-    callState: 'success' as RequestStatus,
+    callState: 'loaded' as RequestStatus,
   },
 }));
 
--- a/apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.test.ts
+++ b/apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.test.ts
@@ -1,0 +1,64 @@
+import { renderHook } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { RequestStatus } from '../../../shared/types/request-status.ts';
+import { useProjectMenu } from './useProjectMenu';
+
+const projectState = vi.hoisted(() => ({
+  value: {
+    id: '2efa3556b1d5bec2f99b6fca4c0010db' as string | null,
+    title: 'Demo project' as string | null,
+  },
+}));
+
+const flagsState = vi.hoisted(() => ({
+  value: {
+    flags: {
+      MENU_OWOX_BI_VISIBLE: 'true',
+    } as Record<string, unknown>,
+    callState: 'success' as RequestStatus,
+  },
+}));
+
+vi.mock('../../../app/store/hooks', () => ({
+  useProject: () => projectState.value,
+  useFlags: () => flagsState.value,
+}));
+
+vi.mock('../../../features/idp/hooks/useProjects.ts', () => ({
+  useProjects: () => ({
+    projects: [],
+    loadProjects: vi.fn(),
+    callState: 'idle',
+    error: null,
+    isLoading: false,
+    reset: vi.fn(),
+  }),
+}));
+
+function getOwoxBiHref() {
+  const { result } = renderHook(() => useProjectMenu());
+  const owoxBiItem = result.current.visibleMenuItems.find(
+    item => item.type === 'menu-item' && item.title === 'OWOX BI'
+  );
+
+  return owoxBiItem && 'href' in owoxBiItem ? owoxBiItem.href : undefined;
+}
+
+describe('useProjectMenu', () => {
+  beforeEach(() => {
+    projectState.value = {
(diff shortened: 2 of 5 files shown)


apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.ts
```diff
@@ -32,6 +32,10 @@ export function useProjectMenu() {
     });
 
     const mappedItems = filteredItems.map(item => {
+      if (id && item.buildHref) {
```
laskevych, line 35: 🟡 **Important (non-blocking):** `buildHref` is applied whenever a project id exists, regardless of the active identity provider — but the "ODM project id == BI project id" assumption only holds for the OWOX IDP.

With the null provider the id is the literal `'0'` (`packages/idp-protocol/src/providers/null-provider.ts:47`), and with the generic better-auth provider ids are generated locally. `MENU_OWOX_BI_VISIBLE` is whitelisted for any deployment (`apps/owox/src/web/flags-route.ts:25`), so a self-hosted instance that turns the item on now gets `https://bi.owox.com/ui/app/0` instead of the BI home page that used to work — a small regression for that setup.

The hook already computes `isOwoxIdpProvider` on line 15; gating this branch on it would keep the root-URL fallback for every other provider. If the intent is that this flag is only ever set in OWOX Cloud, that is fine too — worth stating so the assumption is explicit.

apps/web/src/components/AppSidebar/ProjectMenu/items.ts
```diff
@@ -55,6 +55,7 @@ export const projectMenuItems: ProjectMenuItem[] = [
     type: 'menu-item',
     title: 'OWOX BI',
     href: 'https://bi.owox.com/',
+    buildHref: projectId => `https://bi.owox.com/ui/app/${projectId}`,
```
laskevych, line 58: 🔵 **Question:** the BI origin stays hardcoded to production `bi.owox.com`, but the path now carries an environment-specific project id.

On a staging/dev ODM environment the item would link to `https://bi.owox.com/ui/app/<staging-project-id>`, i.e. a project that does not exist on production BI — with the old root URL that was harmless. Do non-production environments keep this menu item hidden, or should the BI origin follow the same flag mechanism as the other `MENU_*` variables?

apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.ts
```diff
@@ -32,6 +32,10 @@ export function useProjectMenu() {
...
+        return { ...item, href: item.buildHref(id) } as ProjectMenuItem;
```
laskevych, line 36: 💡 **Suggestion (optional follow-up):** after this change the same `map` carries two conventions for injecting the project id into an external URL: the `/p/none/` placeholder substitution (documented as the shared pattern in `apps/web/src/pages/project-settings/ExternalPlatformTab.tsx:19` and reused by the Credit Consumption / Subscription tabs) and the new `buildHref` callback.

Not worth reworking here, but the next project-scoped external link will have to choose between them, and picking the wrong one fails silently (e.g. adding a `/p/none/` template for a BI-shaped path). A single placeholder convention covering both would be worth considering when the third link appears.

apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    title: 'Demo project' as string | null,
+  },
+}));
+
+const flagsState = vi.hoisted(() => ({
+  value: {
+    flags: {
+      MENU_OWOX_BI_VISIBLE: 'true',
+    } as Record<string, unknown>,
+    callState: 'success' as RequestStatus,
```
laskevych: 💡 **Nit:** `callState: 'success' as RequestStatus` — `RequestStatus` only defines `idle`/`loading`/`loaded`/`error`, so the cast is what makes this compile.

It is harmless today because `useProjectMenu` never reads `callState`, but if the hook later gates on `callState === RequestStatus.LOADED`, both tests would quietly fall into the not-loaded branch and assert the fallback href while still appearing to cover the project-scoped case. `RequestStatus.LOADED` states the intent and keeps the type check meaningful.

laskevych (approved):
✅ **Approved.** Verified locally on the branch: `npm test -w @owox/web -- src/components/AppSidebar/ProjectMenu` passes (3 files / 14 tests), ESLint is clean on the touched files, and the only `tsc --noEmit` failure in `@owox/web` is a pre-existing one in `CalculatedFieldFormulaCell.tsx` that is already on `main` and unrelated to this change. The changeset matches the repository policy (`owox`: minor, Work Item ID prefix), and the fallback to the BI root when the project id is unknown is covered by tests.

All comments below are non-blocking.

apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.ts
romandubovyi: The OWOX BI item is only meant for OWOX Cloud, where the project id is shared with BI. Leaving the current behavior as-is.

apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.ts
romandubovyi: Agreed — not worth unifying in this PR. We can pick one convention when a third project-scoped link appears.

apps/web/src/components/AppSidebar/ProjectMenu/items.ts
romandubovyi: This is not a problem.

apps/web/src/components/AppSidebar/ProjectMenu/useProjectMenu.test.ts
romandubovyi: Agreed, I'll switch this to `RequestStatus.LOADED`.

PR #1583, opened 2026-09-02
feat(shopify): add checkoutToken field to orders

prokopetsv:
# Problem

The Shopify connector had no way to import the checkout token of an order. Analysts who collect Shopify Web Pixel events (`checkout_completed`) could not join those events to imported orders, because the pixel identifies a purchase by `checkout.token` while the connector only exposed order IDs, names, and totals. The GraphQL Admin API `Order` object in the version the connector already pins (`2026-07`) provides this value as `checkoutToken`, but it was missing from the connector's `orders` field schema.

# Solution

Expose `checkoutToken` as an opt-in field on the `orders` node. The connector's query builder, value extractor, and storage schema are all driven by the per-node `*Fields.js` maps, so the change is a single schema entry with no logic changes in `Source.js` or `Connector.js`.

- Added the field to `ordersFields.js` with a plain `graphqlPath` of `checkoutToken`; the existing `_buildQueryFields` and `_normalizeFromSchema` handle it like any other scalar.
- Left `defaultFields` untouched so existing data marts keep the same query and table shape until a user selects the new field.
- Verified against the Shopify docs that `Order.checkoutToken` exists in `2026-07`, is not deprecated, and requires no additional scope, so no API version bump is needed. `AbandonedCheckout` has no token field, so that node is unchanged.
- Added the first unit test for the Shopify source, loading the real `DataTypes`, `ordersFields`, `AbstractSource`, and `ShopifySource` via the existing `loadGasClass` helper, covering query generation, value mapping, and the null case.

Historical orders will show `NULL` for this column until re-imported via a manual backfill, because incremental runs only revisit the reimport lookback window.

# Changes

- Added `checkoutToken` (STRING) to `packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js`
- Added `packages/connectors/test/Sources/Shopify/Source.test.js` with three cases: query field emission, node-to-row mapping, and null handling
- Added changeset `.changeset/6883-shopify-order-checkout-token.md` (`owox: minor`)

### DIFF
--- a/.changeset/6883-shopify-order-checkout-token.md
+++ b/.changeset/6883-shopify-order-checkout-token.md
@@ -2,6 +2,6 @@
 'owox': minor
 ---
 
-# Checkout token for Shopify orders
+# Checkout and cart tokens for Shopify orders
 
-The Shopify connector now offers a `checkoutToken` field on the `orders` node. Add `orders checkoutToken` to the Fields list to import it. The value matches `checkout.token` in Shopify Web Pixel events, so you can join orders with `checkout_completed` events. Orders that did not originate from a checkout, such as orders created through the API, have no token and return NULL.
+The Shopify connector now offers `checkoutToken` and `cartToken` fields on the `orders` node. Select them under `orders` in the connector's field picker to import them. `checkoutToken` matches `checkout.token` in Shopify Web Pixel events, so you can join orders with `checkout_completed` events; `cartToken` identifies the cart that produced the order. Orders with no associated checkout or cart return NULL. Orders imported before you enable the fields stay NULL until you re-import them with a backfill run.
--- a/.changeset/6883-shopify-order-checkout-token.md
+++ b/.changeset/6883-shopify-order-checkout-token.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Checkout token for Shopify orders
+
+The Shopify connector now offers a `checkoutToken` field on the `orders` node. Add `orders checkoutToken` to the Fields list to import it. The value matches `checkout.token` in Shopify Web Pixel events, so you can join orders with `checkout_completed` events. Orders that did not originate from a checkout, such as orders created through the API, have no token and return NULL.
--- a/packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js
+++ b/packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js
@@ -194,9 +194,14 @@ var ordersFields = {
     'graphqlPath': 'sourceName'
   },
   'checkoutToken': {
-    'description': 'Token of the checkout that created the order. Matches checkout.token in Web Pixel events. Null when the order did not originate from a checkout, for example orders created through the API.',
+    'description': 'Token of the checkout that created the order. Useful for correlating the order with Web Pixel checkout events. NULL when no checkout is associated.',
     'type': DATA_TYPES.STRING,
     'graphqlPath': 'checkoutToken'
+  },
+  'cartToken': {
+    'description': 'Token of the cart that was used to create the order. NULL when no cart is associated.',
+    'type': DATA_TYPES.STRING,
+    'graphqlPath': 'cartToken'
   },
   'processedAt': {
     'description': 'The date and time when the order was processed.',
--- a/packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js
+++ b/packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js
@@ -193,6 +193,11 @@ var ordersFields = {
     'type': DATA_TYPES.STRING,
     'graphqlPath': 'sourceName'
   },
+  'checkoutToken': {
+    'description': 'Token of the checkout that created the order. Matches checkout.token in Web Pixel events. Null when the order did not originate from a checkout, for example orders created through the API.',
+    'type': DATA_TYPES.STRING,
+    'graphqlPath': 'checkoutToken'
+  },
   'processedAt': {
     'description': 'The date and time when the order was processed.',
     'type': DATA_TYPES.TIMESTAMP,
--- a/packages/connectors/test/Sources/Shopify/Source.test.js
+++ b/packages/connectors/test/Sources/Shopify/Source.test.js
@@ -16,22 +16,27 @@ loadGasClass(src('Sources/Shopify/Source.js'));
 const proto = globalThis.ShopifySource.prototype;
 const schema = { fields: globalThis.ordersFields };
 
-describe('orders checkoutToken', () => {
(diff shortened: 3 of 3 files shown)


packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js
```diff
@@ -193,6 +193,11 @@ var ordersFields = {
     'type': DATA_TYPES.STRING,
     'graphqlPath': 'sourceName'
   },
+  'checkoutToken': {
+    'description': 'Token of the checkout that created the order. Matches checkout.token in Web Pixel events. Null when the order did not originate from a checkout, for example orders created through the API.',
```
laskevych: 🟡 Important (non-blocking): The last sentence of the description, "Null when the order did not originate from a checkout, for example orders created through the API", is not something Shopify documents. The 2026-07 reference only says the token is "associated with the checkout that was used to create the order" and is useful for correlating with web pixels.

This description is not just a UI hint: BigQuery, Databricks, and Redshift storages write it into the column description, the web field picker renders it, and the BigQuery schema provider reads it back into the Data Mart schema. An analyst who filters `checkoutToken IS NULL` expecting "API-created orders" may get a different population (POS orders, draft orders paid via invoice checkout).

I'd trim it to what Shopify states, e.g. "Token of the checkout that created the order. Useful for correlating the order with Web Pixel checkout events. NULL when no checkout is associated."

packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js
```diff
@@ -193,6 +193,11 @@ var ordersFields = {
     'type': DATA_TYPES.STRING,
     'graphqlPath': 'sourceName'
   },
+  'checkoutToken': {
```
laskevych, line 196: 🔵 Question: The 2026-07 `Order` object exposes one more token scalar, `cartToken`, right next to this one. Is joining orders to cart-level Web Pixel events out of scope for this task, or is it worth adding in the same PR to avoid a near-identical follow-up?

.changeset/6883-shopify-order-checkout-token.md
```diff
@@ -0,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Checkout token for Shopify orders
+
+The Shopify connector now offers a `checkoutToken` field on the `orders` node. Add `orders checkoutToken` to the Fields list to import it. The value matches `checkout.token` in Shopify Web Pixel events, so you can join orders with `checkout_completed` events. Orders that did not originate from a checkout, such as orders created through the API, have no token and return NULL.
```
laskevych: 💡 Suggestion (non-blocking): The PR description mentions that orders imported before enabling the field stay `NULL` until a backfill run re-imports them, because incremental runs only revisit the `updated_at` lookback window. That is exactly what a user joining with `checkout_completed` events will hit first, so it belongs in the release note too. One sentence is enough: "Orders imported before you enable the field stay NULL until you re-import them with a backfill run."

💡 Suggestion (nit, non-blocking): "Add `orders checkoutToken` to the Fields list" describes the internal `node field` string that the backend builds from the selected fields. Users never type it; the `Fields` parameter is hidden from the config form and fields are picked by checkbox in the field-selection step. Suggest: "Select `checkoutToken` under `orders` in the connector's field picker."

packages/connectors/test/Sources/Shopify/Source.test.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('maps checkoutToken from the order node', () => {
+    const node = { id: 'gid://shopify/Order/1', checkoutToken: 'a1b2c3d4' };
+    expect(
+      proto._normalizeFromSchema.call(proto, { node, schema, fields: ['id', 'checkoutToken'] })
+    ).toEqual({ id: 'gid://shopify/Order/1', checkoutToken: 'a1b2c3d4' });
+  });
+
+  it('yields null when the order has no checkout token', () => {
```
laskevych: 🧪 Test (non-blocking): This case passes even if the `checkoutToken` entry is removed from `ordersFields`, because `_normalizeFromSchema` returns `null` for any field without a `graphqlPath`. The first two cases do guard the entry (the query string would drop to `'id'` and the mapped value would become `null`).

If you want the null case to carry weight, use a node where the key is absent, which exercises the `undefined` → `_formatValue` → `null` path that is currently untested, or assert `_buildQueryFields` in the same case.

laskevych (approved):
✅ Approving. Verified `Order.checkoutToken` against the Shopify 2026-07 reference (present, not deprecated, no extra scope). The schema-only change is the right depth: `defaultFields` untouched, storages add the column lazily, and no docs enumerate order fields. Test, ESLint, and Prettier pass locally. The inline comments are all non-blocking.

📝 Note (out of scope for this PR): `packages/connectors` is not in the workspace matrix of `.github/workflows/test-owox.yml`, so this new test and the existing connector test files only run locally via `npm test`. Worth a separate follow-up so connector unit tests actually gate merges.

packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js
prokopetsv: Agreed. Replaced with your wording so the column description and field picker only state what Shopify documents: "Token of the checkout that created the order. Useful for correlating the order with Web Pixel checkout events. NULL when no checkout is associated."

packages/connectors/src/Sources/Shopify/ShopifyAPIReference/ordersFields.js
prokopetsv: Good call, added in the same PR. `cartToken` is documented in 2026-07 as "the token associated with the cart that was used to create the order", not deprecated, no extra scope. Same one-entry pattern, opt-in, description kept to Shopify's wording; the test and changeset cover both tokens.

.changeset/6883-shopify-order-checkout-token.md
prokopetsv: Both taken. The release note now says "Select `checkoutToken` under `orders` in the connector's field picker" and ends with "Orders imported before you enable the field stay NULL until you re-import them with a backfill run."

packages/connectors/test/Sources/Shopify/Source.test.js
prokopetsv: Good catch. The case now uses a node without the `checkoutToken` key, so it covers the `undefined -> _formatValue -> null` path that had no test, and no longer passes trivially if the schema entry is removed. The first two cases keep guarding the entry itself.

prokopetsv:
Re the CI note: agreed that `packages/connectors` belongs in the `test-owox.yml` matrix so these tests gate merges. I'll open a separate ticket rather than widen this PR.

PR #1585, opened 2026-09-02
feat(mcp): update an existing report instead of creating duplicates

laskevych:
## Why

When a report is created through the assistant and the next request changes it ("add a filter", "sort by revenue"), the assistant tends to create a second report instead of changing the first one, and several related exports in one conversation end up as separate Google Sheets files. In some setups the assistant also copies query rows into a file or Drive document through its own integration instead of using the OWOX report tools. This PR reworks the MCP report tools and instructions around those three cases.

## What changes

**add_report**
- Refuses a report whose fields duplicate one the same user already created on the same data mart and destination, before any side effect (no Google Sheet is created). The refusal is a structured `similar_report_exists` error carrying the existing report's definition, so the assistant switches to `update_report`. `allow_similar: true` creates a separate report when the user explicitly wants one. Looker Studio keeps its own one-report-per-pair rule.
- Accepts `spreadsheet_id` (Google Sheets only) and adds the report as a new sheet of that spreadsheet instead of creating a new file. New use case `AddGoogleSheetToSpreadsheetService`: opens the spreadsheet with the destination's credentials, never reuses a same-named sheet (it may hold hand-maintained data), and explains an inaccessible spreadsheet with the remedy. Results now carry `spreadsheet_id`; `placed_in_root` / `shared_with_requester` are reported for new files only.

**update_report**
- Accepts `run_immediately`. By default a push-destination report is re-run when the export changed (fields or any output control), not for a name-only or message-only change; `true` is rejected for pull-based destinations, mirroring `add_report`.
- Returns the report as it is after the update (name, fields, filters, slices, aggregations, date buckets, sort, limit, spreadsheet/sheet for Google Sheets) and the `run` outcome, sharing the outcome shape and polling rule with `add_report`'s `initial_run`.

**get_data_mart_reports**
- Each item now includes the report's fields and output controls in the `add_report` / `update_report` vocabulary (a reverse mapper of the query-input mapper, round-trip tested), `post_aggregation_filters` for UI-only HAVING rules, `spreadsheet_id` / `sheet_url` for Google Sheets reports, `created_at`, and `created_by_current_user`.

**Instructions and descriptions**
- New "Exporting and delivering data" rules: use the report tools for any export or delivery, never copy `query_data_mart` rows into a CSV, file, or document through another integration; one report per request, then `update_report`; related Google Sheets exports go into one spreadsheet via `spreadsheet_id`; polling rules extended to `update_report`'s run.

**Docs and changeset**
- `docs/getting-started/setup-guide/mcp.md`: updated tool sections, example prompts, and troubleshooting entries; the tool count and `mcp:read` list now include `get_report_output_schema`, which was missing.
- One `'owox': minor` changeset.

## Verification

- `npm run build -w @owox/backend` passes.
- Targeted specs pass (191 tests across the touched suites): facade, both Google Sheets use cases, the three report tools, instructions, operator-advertising contract, and the MCP boundary spec. New specs cover the reverse output-controls mapper, the similar-report guard (same user / fields / destination, newest match, `['*']` vs no projection, `allow_similar`), the existing-spreadsheet path, `update_report`'s run defaults, and the new use case.
- Prettier, ESLint, and markdownlint are clean on the changed files.


apps/backend/src/ee/mcp/tools/add-report.tool.ts
```diff
@@ -298,38 +314,27 @@ export class AddReportTool implements McpToolDefinition<AddReportInput> {
...
+      error_code: SIMILAR_REPORT_EXISTS_ERROR_CODE,
+      message:
+        `${err.message} The existing report's current fields and output controls are in existing_report: ` +
+        'to change what it exports, call update_report with its report_id, sending only the controls that change ' +
+        '(update_report re-runs a push report by default, so the destination refreshes). Tell the user which report you reused.',
+      existing_report: { ...err.existingReport, report_url: reportUrl },
+    };
+    return {
+      isError: true,
+      structuredContent: payload,
```
prokopetsv: 🟡 Non-Blocker

This error result carries a `structuredContent` payload (`error_code`, `message`, `existing_report`) that does not match the tool's declared `outputSchema` (`report_id`, `status: 'created'`, `initial_run`, …).

The bundled MCP SDK client (1.30.0, `client/index.js` → `callTool`) skips validation only when `structuredContent` is *absent*; when it is present it validates it even on `isError` results and throws `InvalidParams`. An SDK-based client would therefore never see `existing_report`, which is the mechanism this PR relies on to redirect the agent to `update_report`.

This mirrors the pre-existing `toStructuredToolError` convention in `query_data_mart`, so it is not new — but here the payload is load-bearing. Cheapest fix for this path: keep the JSON in the `content` text and drop `structuredContent` from the error result, or confirm the target clients ignore output validation on errors.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -273,7 +290,26 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
-    return { report_id: request.reportId, status: 'updated' };
+    // A refresh run is queued when the export changed, so the destination does
+    // not keep showing rows the report no longer defines; a rename or a message
+    // edit alone delivers nothing new and is not worth a billed run (or a
+    // repeated Slack/email message) unless explicitly requested.
+    const run = await this.queueRunAfterWrite(
+      updated.id,
+      request,
+      destinationType,
+      request.runImmediately ?? exportChanged
```
prokopetsv: 🟡 Non-Blocker

With this default, any output-control change on an email / Slack / Teams / Google Chat report queues a run, which sends a message to every configured recipient or channel. Before this PR `update_report` never delivered.

The behavior is documented in the tool description and instructions, and `run_immediately=false` opts out — but a "sort by revenue" tweak re-broadcasting to third parties is an externally visible side effect worth a deliberate product call. If intended, no change needed.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
```diff
@@ -1188,7 +1454,80 @@ describe('McpReportsFacadeImpl.updateReport', () => {
...
+      expect.objectContaining({
+        destination_type: 'looker_studio',
+        run: { status: 'not_applicable' },
+      })
+    );
+    await expect(
+      built.facade.updateReport({ ...updateRequest, fields: ['channel'], runImmediately: true })
+    ).rejects.toThrow('run_immediately is not applicable to Data Studio reports');
+    expect(built.runReportService.run).not.toHaveBeenCalled();
   });
```
prokopetsv, line 1673: 🟡 Non-Blocker (test coverage)

This only asserts `runReportService.run` was not called. It does not assert `updateReportService.run` was *not called again* for the `runImmediately: true` case, so the pull-based guard could silently move after the write and the test would still pass.

The equivalent `add_report` Looker Studio test asserts the create service was never called; mirroring that here (e.g. `expect(built.updateReportService.run).toHaveBeenCalledTimes(1)`) closes the gap.

prokopetsv (commented):
Inline comments below. No blockers; approval to follow.

prokopetsv (approved):
Approved ✅

Reviewed the full change against docs/review-rules.md and the backend architecture conventions.

What I verified:
- The similar-report guard runs before any side effect (no Google Sheet is created on refusal), scopes to the caller's own reports through the access-checked list service, and skips Looker Studio, which keeps its own one-report-per-pair rule. Field comparison is set-based and handles `null`, `['*']`, and `[]` correctly.
- `AddGoogleSheetToSpreadsheetService` follows the Command → use case → DTO flow with no repository access, uses `castError`, and separates transient Google faults (429/5xx) from access errors with a clear remedy.
- `update_report` rejects `run_immediately=true` for pull-based destinations before the write, and the shared `queueRunAfterWrite` keeps every outcome non-fatal to the committed change.
- The reverse output-controls mapper is round-trip tested; `toSheetTitle` extraction removes a duplicate.
- Changeset carries Fibery ID 6847 with `'owox': minor` and describes user-visible outcomes. Docs tool count (19) matches the registered tools.
- CI is green; the `Audit all` failure is the unrelated `qs` advisory fixed in #1586.

Three 🟡 Non-Blocker inline comments left: the `similar_report_exists` error's `structuredContent` vs the SDK client's output-schema validation (pre-existing convention, but load-bearing here), the default re-run on email-family reports as a deliberate product call, and one missing ordering assertion in the pull-based `update_report` test. None blocks merge.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
laskevych: Confirmed against the bundled SDK 1.30.0 client (`callTool` validates `structuredContent` whenever it is present, error or not) — thanks, this would have hidden `existing_report` from every SDK-based client. Fixed in dbbfca80a: the refusal now travels as text only (`isError: true`, JSON in `content`), with a comment explaining why, and the spec asserts the result has no `structuredContent`. The pre-existing `toStructuredToolError` in `query_data_mart` has the same exposure; left untouched here as it is out of scope, worth a small follow-up.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Agreed that a re-broadcast on a "sort by revenue" tweak is the wrong default. Changed in dbbfca80a: `update_report` re-runs by default only a Google Sheets report whose export changed; email, Slack, Teams, and Google Chat reports are never re-sent by an update unless `run_immediately: true` is passed explicitly. The tool description, the `not_requested` message for email-family reports, the system instructions ("ask the user before re-sending"), docs, and the changeset now say so, and a facade spec covers both the default and the explicit opt-in.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.spec.ts
laskevych: Added in dbbfca80a: `expect(built.updateReportService.run).toHaveBeenCalledTimes(1)` after the rejected `runImmediately: true` call, mirroring the Looker Studio create test.

apps/backend/src/ee/mcp/tools/update-report.tool.ts
```diff
@@ -101,6 +122,12 @@ const baseInputSchema = z
       .describe(
         'Message changes. Applies only to reports with an email, slack, teams, or google_chat destination; rejected for other types. Provide at least one of subject/body inside. The send condition and recipients are not editable here.'
       ),
+    run_immediately: z
+      .boolean()
+      .optional()
+      .describe(
+        'Whether to run the report after the update (one billed Report Run). Omit for the default: a Google Sheets report re-runs when fields or any output control changed, so the sheet reflects the change; it does not run for a name-only or message-only change. Email, Slack, Microsoft Teams, and Google Chat reports are NOT re-sent by default, because a run delivers the message to every configured recipient or channel — set true only when the user explicitly wants it sent now. Set false to update a Google Sheets report without refreshing it. Looker Studio is pull-based: omit or set false; true is rejected.'
+      ),
```
prokopetsv, line 130: 🔴 Blocker

This new option can write to Google Sheets and deliver email, Slack, Teams, or Google Chat messages, but the tool still declares `openWorldHint: false` below. MCP clients may rely on that annotation when deciding whether external-side-effect warnings or confirmations are required. Set it to `true`, consistent with `run_report`, and update the annotation test.

apps/backend/src/data-marts/facades/mcp-report-output-controls.ts
```diff
@@ -0,0 +1,194 @@
...
+  date_buckets: McpReportDateBucket[];
+  sort: McpReportSort[];
+  /** Max rows per run, or `null` when uncapped. */
+  limit: number | null;
+}
+
+/** The domain "all fields" marker, as add_report/update_report spell it. */
+export const MCP_ALL_FIELDS = ['*'] as const;
+
+export function toMcpFields(columnConfig: ReportColumnConfig | undefined): string[] {
```
prokopetsv, line 95: 🟡 Non-Blocker

`columnConfig: []` is a valid explicit "project no dimensions" configuration for metrics-only Unique Count reports, but this maps it to `["*"]` and the returned definition omits `uniqueCountConfig`. `get_data_mart_reports` therefore describes such a report as exporting all fields, and `sameFieldSelection()` can incorrectly reject a new all-fields report as a duplicate. Preserve this distinction and expose the Unique Count selection, with coverage for the metrics-only case.

apps/backend/src/data-marts/facades/mcp-report-output-controls.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+  return {
+    filters,
+    slices,
+    ...(having.length > 0 && { post_aggregation_filters: having }),
+  };
+}
+
+export function toMcpAggregations(config: AggregationConfig | undefined): McpReportAggregation[] {
```
prokopetsv, line 227: 🟡 Non-Blocker

This emits every persisted `ReportAggregateFunction`, including `STRING_AGG` and `ANY_VALUE`, while the MCP input schema intentionally rejects those two. Because aggregation updates replace the whole list, an agent cannot round-trip or modify a valid UI-created report containing either function. Align accepted and returned aggregation vocabularies, or provide a preservation path for unsupported entries.

apps/backend/src/data-marts/use-cases/google-sheets/add-google-sheet-to-spreadsheet.service.ts
```diff
@@ -0,0 +1,89 @@
...
+    const title = toSheetTitle(command.title);
+    if (adapter.findSheetByTitle(spreadsheet, title)) {
+      throw new BusinessViolationException(
+        `Spreadsheet ${spreadsheetId} already has a sheet named "${title}". ` +
+          'Use a different report name, or omit spreadsheet_id to create a new file.',
+        { spreadsheetId, sheetTitle: title }
+      );
+    }
+
+    const sheetId = await adapter.addSheet(spreadsheetId, title);
```
prokopetsv: 🟡 Non-Blocker

Only `getSpreadsheet()` errors above are translated. A connected account with Viewer access can read metadata successfully and then fail here with a raw Google 403, so the advertised Editor-access remedy is skipped in the most relevant permission failure. Translate write-time permission/not-found errors consistently, keep transient 429/5xx handling, and add a test for successful read followed by `addSheet()` rejection.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -195,19 +206,17 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
   }
 
   async updateReport(request: McpUpdateReportRequest): Promise<McpUpdateReportResult> {
+    const exportChanged =
```
prokopetsv: 🟡 Non-Blocker

`exportChanged` records parameter presence, not an actual value change. Re-sending the current Google Sheets fields or filters therefore queues a run even though `UpdateReportService` detects the persisted definition is unchanged and does not invalidate its cache. That creates an unnecessary external write and potentially billed report run. Base the default refresh on the actual before/after comparison and cover same-value updates.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
```diff
@@ -298,38 +314,32 @@ export class AddReportTool implements McpToolDefinition<AddReportInput> {
...
+  private toSimilarReportError(
+    err: McpSimilarReportExistsException,
+    reportUrl: string
+  ): McpToolResult {
+    const payload = {
+      error_code: SIMILAR_REPORT_EXISTS_ERROR_CODE,
+      message:
+        `${err.message} The existing report's current fields and output controls are in existing_report: ` +
+        'to change what it exports, call update_report with its report_id, sending only the controls that change ' +
+        '(update_report re-runs a push report by default, so the destination refreshes). Tell the user which report you reused.',
```
prokopetsv: 🟡 Non-Blocker

This guidance is now stale: `update_report` reruns only Google Sheets by default; email and chat destinations require explicit `run_immediately: true`. Since this error is returned for duplicate reports on those destinations too, it can lead the assistant to claim a delivery was refreshed when no message was sent. Make the instruction destination-specific and test both destination families.

prokopetsv (changes requested):
Requested changes based on the blocking MCP safety-annotation issue; five additional non-blocking correctness and contract findings are inline.

apps/backend/src/ee/mcp/tools/update-report.tool.ts
laskevych: Fixed in 536086fac: `update_report` now declares `openWorldHint: true` (comment explains the refresh run reaches Google Sheets / email / chat, same as `run_report`), and the annotation assertion in the tool spec is updated.

apps/backend/src/data-marts/facades/mcp-report-output-controls.ts
laskevych: Fixed in 536086fac. `toMcpFields` keeps an explicit `[]` (only `null`/`undefined` become `['*']`), so a metrics-only Unique Count report is described as projecting no dimension column and `sameFieldSelection([], null)` is false — the guard no longer rejects an all-fields report as a duplicate of it. The definition now also carries `unique_count_sources` (`unique_count` for the report's own data mart, `<source>__unique_count` for joined ones — the names `get_data_mart_details_by_id` lists), with the legacy boolean shape normalised. Covered in the mapper spec (metrics-only + legacy `true`) and a facade guard test; docs updated.

apps/backend/src/data-marts/facades/mcp-report-output-controls.ts
laskevych: Aligned in 536086fac: `add_report` / `update_report` accept the full persisted vocabulary via a new `makeMcpReportAggregationSchema` (`REPORT_AGGREGATE_FUNCTIONS`, i.e. including `STRING_AGG` and `ANY_VALUE`), and the report mapper validates against the same list, so a UI-created report round-trips through `get_data_mart_reports` → `update_report`. `query_data_mart` keeps its narrower menu — those two are not analytical answers, and the field-type matrix stays as is. Tool descriptions and docs name the two extra functions; a tool spec pins that they parse.

apps/backend/src/data-marts/use-cases/google-sheets/add-google-sheet-to-spreadsheet.service.ts
laskevych: Fixed in 536086fac: both calls go through one `translateGoogleError` — 429/5xx stay `GoogleApiException`, anything else becomes a `BusinessViolationException` with the remedy; the `addSheet` variant says explicitly that the connected account needs Editor access (Viewer is not enough) or to omit `spreadsheet_id`. Two new tests: successful read followed by a 403 on `addSheet`, and a 503 on `addSheet`.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Fixed in 536086fac. The facade now builds the merged definition first and compares each control with the stored one serialized — the same rule `UpdateReportService` uses for cache invalidation — and the default refresh keys off that `definitionChanged`, not on parameter presence (presence is still what the "nothing to update" guard checks). Echoing the stored fields/filters therefore updates without queueing a run; new facade test covers the same-value case, and the `not_requested` wording mentions re-sent identical controls.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
laskevych: Fixed in 536086fac: the guidance is derived from `existing_report.destination_type` — Google Sheets says the update re-runs by default so the sheet refreshes; email / Slack / Teams / Google Chat say the report is NOT re-sent unless `run_immediately=true` is passed on the user's explicit request. The tool spec now asserts both families (and that neither text leaks into the other).

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -424,35 +500,84 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+        roles: context.roles,
+        reportId,
       });
-      return {
-        ...created,
-        initial_run: { status: 'queued', run_id: run.runId },
-      };
+      return { status: 'queued', run_id: run.runId };
     } catch (error) {
       const message = error instanceof Error ? error.message : 'Report run could not be queued';
```
amarchenk0: 🟡 Non-Blocker

This new helper normalizes an unknown error with `instanceof Error` and `String(error)`, while the repository rules require `castError` from `@owox/internal-helpers`. A thrown object currently loses useful details in the MCP response and becomes `[object Object]` in the log. Please normalize once with `const cause = castError(error)` and use `cause.message` / `cause.stack`.

amarchenk0 (approved):

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -424,35 +500,84 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+        request.userId,
+        request.roles
+      )
+    );
+    const similar = reports
+      .filter(
+        report =>
+          report.dataDestinationAccess.id === request.destinationId &&
+          report.createdByUser?.userId === request.userId &&
+          sameFieldSelection(report.columnConfig, wantedColumns)
```
prokopetsv, line 599: 🟡 Non-Blocker

Comparing only `columnConfig` treats an existing UI report with the same dimensions plus `uniqueCountConfig` as a duplicate, although it exports an extra metric that MCP cannot remove. The refusal directs the agent to `update_report`, which preserves Unique Count, so the requested plain report cannot be produced without bypassing the guard via `allow_similar`. Include immutable Unique Count output in the equivalence check or exclude such reports, and add coverage.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -424,35 +500,84 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+        request.projectId,
+        request.userId,
+        request.roles
+      )
+    );
+    const similar = reports
+      .filter(
+        report =>
+          report.dataDestinationAccess.id === request.destinationId &&
+          report.createdByUser?.userId === request.userId &&
```
prokopetsv, line 596: 🟡 Non-Blocker

Creator identity is not current edit access. Report owners can be changed, and `ReportDto` already carries `canEditConfig`. A creator removed from owners still matches here, so `add_report` refuses and directs the agent to `update_report`, which then returns 403. Restrict candidates to reports the caller can currently edit, or verify mutate access, and cover the former-creator case.

apps/backend/src/data-marts/use-cases/google-sheets/add-google-sheet-to-spreadsheet.service.ts
```diff
@@ -0,0 +1,109 @@
...
+@Injectable()
+export class AddGoogleSheetToSpreadsheetService {
+  private readonly logger = new Logger(AddGoogleSheetToSpreadsheetService.name);
+
+  constructor(
+    private readonly dataDestinationService: DataDestinationService,
+    private readonly adapterFactory: GoogleSheetsApiAdapterFactory
+  ) {}
+
+  async run(command: AddGoogleSheetToSpreadsheetCommand): Promise<AddedGoogleSheetDto> {
```
prokopetsv, line 48: 🟡 Non-Blocker

This path has no requester identity, but `add_report` explicitly accepts IDs from `get_data_mart_reports`, which lists other users’ reports. Destination credentials can add the tab even when the current requester cannot open that spreadsheet; unlike the new-file path, this service neither shares it nor returns `shared_with_requester`, so a successful result can contain an unusable `sheet_url`. Restrict reuse to a spreadsheet known to be accessible to the requester, or carry requester identity and establish or report access.

apps/backend/src/data-marts/facades/mcp-report-output-controls.ts
```diff
@@ -0,0 +1,227 @@
...
+    } else if (rule.placement === 'pre-join') {
+      slices.push(toMcpFilter(rule));
+    } else {
+      filters.push(toMcpFilter(rule));
+    }
+  }
+  return {
+    filters,
+    slices,
+    ...(having.length > 0 && { post_aggregation_filters: having }),
```
prokopetsv: 🔴 Blocker

This output cannot be preserved by the update path. `get_data_mart_reports` also returns UI-only `regex` and `relative_date` rules, but `update_report` rejects those operators and has no `function` field for these `post_aggregation_filters`; meanwhile `McpReportsFacadeImpl.mergeFilterConfig()` replaces every post-join rule, including HAVING, whenever `filters` is supplied. In the core “add a filter” flow, the agent must either omit the requested change or silently remove or reinterpret existing rules, potentially broadening delivered data. Provide a lossless preservation or update path and cover adding a filter to a report that already has UI-only and HAVING rules.

laskevych: Fixed in 9ddedfa69 — thanks, this was the real gap in the "add a filter" flow. `mergeFilterConfig` now replaces, per placement, only the rules MCP can express; a rule it cannot — a HAVING constraint, `regex`/`not_regex`, or a calendar preset outside the vocabulary (today, yesterday, last month, last N months) — is kept as stored (`isUiOnlyFilterRule` lives next to the reverse mapper so the two cannot drift). The read path lists those under `ui_only_filters` (with `placement` and, for HAVING, `function`) instead of mixing them into `filters`, so what `filters` shows is exactly what `update_report filters` replaces. Tool descriptions, instructions ("never re-send them"), docs and changeset follow. Facade spec covers adding a filter to a report with HAVING + regex slice + "today", clearing with `[]`, and replacing slices.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Fixed in 9ddedfa69: the guard skips any report whose normalised `uniqueCountConfig` is non-empty (legacy `true` included) — a plain report is genuinely different from one carrying a metric `add_report` cannot produce and `update_report` cannot remove. Covered in the guard's `it.each` (array and legacy shapes).

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Fixed in 9ddedfa69: candidates must also satisfy `report.canEditConfig` (the capability `ListReportsByDataMartService` already computes), so a creator removed from the owners is no longer sent to an `update_report` that would 403. Former-creator case added to the guard's `it.each`.

apps/backend/src/data-marts/use-cases/google-sheets/add-google-sheet-to-spreadsheet.service.ts
laskevych: Addressed in 9ddedfa69 by checking — deliberately never granting — the requester's access: sharing an existing spreadsheet on the way would hand any member with USE on the destination writer access to any file that account can edit. The service now takes the requester (id/email, IDP fallback like the new-file path), gets a Drive-capable client from a new `GoogleSheetsApiAdapterFactory.createWithDriveScope` (SA → Drive-scoped JWT; OAuth → `driveCapable` from the token's scopes), and reads `permissions.list`: a direct user, domain or anyone grant confirms; a confirmed miss is refused before any tab is written; a group grant, no Drive scope, unknown email or a failed lookup degrade to `shared_with_requester: false` so the agent warns about the link. Spec covers each branch; docs describe the refusal.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
laskevych: Fixed in 9ddedfa69: `queueRunAfterWrite` normalises once with `castError` and uses `cause.message` / `cause.stack`.

apps/backend/src/data-marts/facades/mcp-reports.facade.impl.ts
```diff
@@ -243,46 +260,104 @@ export class McpReportsFacadeImpl implements McpReportsFacade {
...
+      dateTruncConfig:
         request.dateTruncConfig !== undefined
           ? request.dateTruncConfig
           : (current.dateTruncConfig ?? null),
+    };
+    // Compared the way UpdateReportService decides whether to invalidate its
+    // cache — serialized, so a re-sent identical control is NOT a change. An
+    // agent that echoes the stored fields back must not trigger an external
+    // write and a billed run for a definition that did not move.
+    const definitionChanged = (Object.keys(next) as Array<keyof typeof next>).some(
```
prokopetsv, line 298: 🟡 Non-Blocker

The serialized comparison produces false "changed" verdicts in exactly the echo scenario it protects (the round-2 fix), through two paths introduced by the merge logic in 9ddedfa69:

1. **Placement injection.** A UI-created rule is stored without a `placement` key (`PlacementExtrasSchema` makes it optional), but the tool layer's `mapOne` always sets `placement` on incoming rules. An agent echoing that rule back verbatim produces a merged rule with `placement: 'post-join'` added — `JSON.stringify` differs — `definitionChanged` is true. The facade spec's echo test passes only because it hand-crafts a byte-identical rule and bypasses the tool mapper.
2. **Reordering.** `mergeFilterConfig` prepends preserved UI-only rules ahead of the replacement within each placement, and regroups pre-join before post-join. The spec's own "keeps UI-only rules" test shows stored `[channel, having, regexSlice, today]` becoming `[regexSlice, having, today, channel, country]`. Any filter update on a report with UI-only rules — including a pure echo — therefore reads as a definition change.

Impact: a Google Sheets report queues a default refresh run (external write + billed run) for an update that changed nothing, and the persisted config is rewritten with reordered/augmented rules.

Suggested fix: canonicalize both sides before comparing (apply the default placement, compare filter rules order-insensitively), or make `mergeFilterConfig` order-preserving (substitute expressible rules in place). Add an echo test that routes through the tool mapper against a placement-less stored rule.

apps/backend/src/data-marts/data-destination-types/google-sheets/adapters/google-sheets-api.adapter.ts
```diff
@@ -295,6 +295,23 @@ export class GoogleSheetsApiAdapter {
...
 
+  /**
+   * Lists who a file is shared with (Drive `permissions.list`), so a caller can
+   * tell whether a given person can open it WITHOUT granting anything. Requires a
+   * Drive scope on the underlying auth; errors propagate for the caller to classify.
+   */
+  public async listFilePermissions(fileId: string): Promise<drive_v3.Schema$Permission[]> {
+    const drive = this.getDriveService();
+    const res = await this.executeWithRetry(() =>
+      drive.permissions.list({
```
prokopetsv, line 306: 🟡 Non-Blocker

`drive.permissions.list` is paginated: for shared-drive files it returns at most 100 permissions per page plus a `nextPageToken`. This reads only the first page, so a requester whose grant sits beyond page 1 gets a truncated ACL that names neither them nor a group → `checkRequesterAccess` returns `'denied'` → `AddGoogleSheetToSpreadsheetService` hard-refuses a spreadsheet the requester can actually open. That breaks the documented invariant that "denied is claimed only when the ACL is fully resolvable" (enterprise shared drives with >100 members are the realistic trigger).

Cheapest correct fix: when `nextPageToken` is present and no grant matched, return `'unknown'` instead of `'denied'`; or follow the token. Either way, add a test with a paginated response.

prokopetsv (approved):
Approved ✅

Re-review of head 9ddedfa69 after the three prior rounds. All 15 previously raised findings are genuinely fixed at head, verified against the code, not the reply comments:

- update_report declares `openWorldHint: true`; the `similar_report_exists` refusal travels text-only, with guidance derived from `existing_report.destination_type` per destination family.
- `isUiOnlyFilterRule` exactly mirrors what `toMcpFilter` cannot translate — I cross-checked the stored `FilterRuleSchema` vocabulary against `MCP_OPERATORS`: `regex`/`not_regex` and the four calendar presets are precisely the gap; everything else round-trips, including the legacy `is_null`-family operators.
- The similar-report guard requires `canEditConfig` and skips Unique Count reports; `queueRunAfterWrite` normalises with `castError`.
- The requester-access check runs before the tab write, refuses only on a confirmed miss, and degrades to `shared_with_requester: false` otherwise; the `driveCapable` OAuth sniffing works (the `credential` relation is eager-loaded, default OAuth scopes include `drive.file`); `AddGoogleSheetToSpreadsheetCommand` positional args match the call site.
- The merge-preservation design — UI-only rules split from expressible ones by a single classifier shared by the read and write paths — is sound, and side effects stay behind validation throughout.

Two 🟡 Non-Blocker inline comments on 9ddedfa69 itself: a spurious default refresh run when echoed filters serialize differently than stored (placement injection and UI-only reordering), and a possible false access refusal from an unpaginated `permissions.list` on large shared-drive ACLs. Both are quality-of-behavior issues, not data-integrity ones; neither blocks merge.

PR #1586, opened 2026-09-02
chore(deps): bump qs to 6.16.0

laskevych:
## Why

The `Audit all` workflow started failing on every pull request after new advisories were published for `qs` (array-limit bypass via bracket-key comma parsing and a denial of service via attacker-controlled `isBuffer`). Nothing in the repository changed; the lockfile simply pins a version that is now flagged.

## What changes

`package-lock.json` only: `qs` 6.15.3 → 6.16.0 (a transitive dependency of `express` / `body-parser`). Its dependency ranges are unchanged, so the rest of the lockfile is untouched.

## Verification

- `npm audit --omit=dev` reports 0 vulnerabilities.
- `npm ls --package-lock-only` passes (lockfile consistent).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32699,9 +32699,9 @@
       }
     },
     "node_modules/qs": {
-      "version": "6.15.3",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.15.3.tgz",
-      "integrity": "sha512-O9gl3zCl5h5blw1KGUzQKhA5oUXSl8rwUIM5o0S3nCXMliSvy5Dzx7/DJcI+SwgICv+IneSZwhBh1oSyEHA71A==",
+      "version": "6.16.0",
+      "resolved": "https://registry.npmjs.org/qs/-/qs-6.16.0.tgz",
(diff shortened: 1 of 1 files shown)


prokopetsv (approved):
Approved ✅

Verified the lockfile change end to end:

- Single hoisted `qs` entry moves 6.15.3 → 6.16.0; `resolved` URL and `integrity` hash match the npm registry record exactly.
- 6.16.0 keeps the same dependency ranges as 6.15.3, so no other lock entries need to move. All four in-tree consumers (`^6.7.0` … `^6.15.2`) are satisfied.
- Both advisories are confirmed: GHSA-x5fp-wj9c-mxmx (`>=6.14.2 <=6.15.3`) and GHSA-4mjr-xmp4-gh2g (`<6.16.0`). 6.16.0 is the first version clear of both.
- `npm audit --package-lock-only --omit=dev` against this branch: 0 vulnerabilities. CI (Audit all, ESLint, Markdownlint, Prettier) is green.
- `main` has not touched `package-lock.json` since the branch point, so this merges cleanly.

No changeset needed: the published `owox` package ships no shrinkwrap or bundled deps, so consumers already resolve `qs@6.16.0` from the existing ranges. No observable release impact.

PR #1587, opened 2026-09-02
feat(mcp): guide users whose project has no published data mart

laskevych:
## Why

A user who connects the MCP server before their project has a published data mart gets nothing useful. `list_data_marts` and `get_relevant_data_marts_by_prompt` return a bare `data_marts: []`, `summarize_data_catalog` returns `data_mart_count: 0` together with an instruction to "summarize the business areas and suggest example prompts", and the system instructions tell the assistant to rephrase and search again. The assistant ends up saying "no data found" after a few futile searches, and the user never learns what a data mart is, where to create one, or that a draft has to be published first. A data mart cannot be created or published through MCP, so the tools have to point the user to the web app.

## What changes

**`getting_started` in the discovery tools** (new `mcp-getting-started.util.ts`)
- `list_data_marts`, `get_relevant_data_marts_by_prompt`, and `summarize_data_catalog` attach `getting_started` when the user sees no published data mart: `can_create_data_marts` (derived from the token roles: admin and editor may create), `create_data_mart_url` and `data_marts_url` built on `PUBLIC_ORIGIN`, `guides` (core concepts, connector-based, SQL-based), `draft_data_marts` the user can see with their pages, and `instructions` the assistant relays. The instructions are role-aware: creators are walked through create → set up → publish (and told to publish listed drafts); business users are told to ask a Project Admin or Technical User to create, publish, and share one.
- `get_relevant_data_marts_by_prompt` only guides when the published catalog is actually empty — an ordinary miss still asks the assistant to rephrase. `list_data_marts` with `status=draft` guides only when no published data mart exists either.
- `summarize_data_catalog` replaces its summary `_instruction` with an empty-catalog one instead of asking for business areas and example prompts.
- The draft lookup is best-effort: if it fails, the links and steps are still returned (warning logged), mirroring how the optional project summary is handled.
- `getting_started` is declared in each tool's `outputSchema`, since the SDK validates `structuredContent` against it.

**System instructions**
- Step 2 of the analytical flow now stops rephrasing when `getting_started` is present, and a new "Empty project" section tells the assistant to relay the guidance and not call `get_data_mart_details_by_id`, `query_data_mart`, or any report or schedule tool until a published data mart exists.

**Docs and changeset**
- `docs/getting-started/setup-guide/mcp.md`: new troubleshooting entry "The assistant says there are no data marts" with the `getting_started` fields and role-specific next steps; the three tool sections and the prerequisites link to it.
- One `'owox': minor` changeset.

Out of scope, on purpose: creating data marts through MCP, changes to `get_project_context`, and the not-found handling of `get_data_mart_details_by_id`.

## Verification

- `nest build` (CI-equivalent) passes.
- Targeted specs pass (35 tests across 5 suites): the new util spec (creator vs business user, admin, drafts and their wording, failed draft lookup, `resolveGettingStarted` both ways), the three tool specs (empty published list with drafts, empty draft list with a published data mart, search miss with a non-empty catalog, search hit never touches the catalog, empty-catalog summary instruction), and the instructions spec.
- Prettier, ESLint, and markdownlint are clean on the changed files.

### DIFF
--- a/apps/backend/src/ee/mcp/tools/data-mart-catalog.tool.spec.ts
+++ b/apps/backend/src/ee/mcp/tools/data-mart-catalog.tool.spec.ts
@@ -156,7 +156,77 @@ describe('ListDataMartsTool', () => {
 
     const result = await tool.handler({}, context);
 
-    expect(result.structuredContent).toEqual({ data_marts: [] });
+    expect(result.structuredContent).toEqual({
+      data_marts: [],
+      getting_started: expect.objectContaining({ reason: 'no_published_data_marts' }),
+    });
+  });
+
+  it('guides the user when the published catalog is empty, naming visible drafts', async () => {
+    const facade = {
+      listDataMarts: jest.fn(async ({ status }: { status: string }) => ({
+        dataMarts:
+          status === 'draft'
+            ? [
+                {
+                  id: 'dm_draft',
+                  title: 'Draft Orders',
+                  description: null,
+                  status: DataMartStatus.DRAFT,
+                  updatedAt: '2026-06-11T10:00:00.000Z',
+                },
+              ]
+            : [],
+      })),
+    } as unknown as jest.Mocked<McpDataMartsFacade>;
+    const tool = new ListDataMartsTool(facade, publicOrigin, projectContext as never);
+
+    const result = await tool.handler({}, context);
+
+    expect(result.structuredContent).toEqual({
+      project: { id: 'project-1', title: 'Analytics' },
+      data_marts: [],
+      getting_started: expect.objectContaining({
+        reason: 'no_published_data_marts',
+        can_create_data_marts: false,
+        create_data_mart_url: 'https://app.owox.com/ui/project-1/data-marts/create',
+        draft_data_marts: [
+          {
+            id: 'dm_draft',
+            title: 'Draft Orders',
+            url: 'https://app.owox.com/ui/project-1/data-marts/dm_draft/data-setup',
+          },
+        ],
+      }),
+    });
+    expect(facade.listDataMarts).toHaveBeenCalledTimes(2);
+  });
+
+  it('does not guide on an empty draft list while a published data mart exists', async () => {
+    const facade = {
+      listDataMarts: jest.fn(async ({ status }: { status: string }) => ({
+        dataMarts:
+          status === 'published'
+            ? [
+                {
+                  id: 'dm_1',
+                  title: 'Orders',
+                  description: null,
+                  status: DataMartStatus.PUBLISHED,
+                  updatedAt: '2026-06-10T10:00:00.000Z',
+                },
+              ]
+            : [],
+      })),
+    } as unknown as jest.Mocked<McpDataMartsFacade>;
+    const tool = new ListDataMartsTool(facade, publicOrigin, projectContext as never);
+
+    const result = await tool.handler({ status: 'draft' }, context);
+
+    expect(result.structuredContent).toEqual({
+      project: { id: 'project-1', title: 'Analytics' },
+      data_marts: [],
+    });
   });
 
   // Note: MCP_TOOL_PROVIDER_CLASSES assertion checks the LOCAL test registry snapshot, not
@@ -175,6 +245,7 @@ describe('ListDataMartsTool', () => {
       outputSchema: expect.objectContaining({
         project: expect.any(Object),
         data_marts: expect.any(Object),
+        getting_started: expect.any(Object),
       }),
       annotations: {
         title: 'List Data Marts',
--- a/apps/backend/src/ee/mcp/tools/data-mart-catalog.tool.spec.ts
+++ b/apps/backend/src/ee/mcp/tools/data-mart-catalog.tool.spec.ts
@@ -1,3 +1,4 @@
+import { z } from 'zod';
 import type { PublicOriginService } from '../../../common/config/public-origin.service';
 import { DataMartStatus } from '../../../data-marts/enums/data-mart-status.enum';
 import type { McpDataMartsFacade } from '../../../data-marts/facades/mcp-data-marts.facade';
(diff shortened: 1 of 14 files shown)


apps/backend/src/ee/mcp/tools/mcp-getting-started.util.ts
```diff
@@ -0,0 +1,167 @@
...
+} from './mcp-docs-urls';
+import { joinPublicOrigin } from './mcp-public-url.util';
+
+const logger = new Logger('McpGettingStarted');
+
+/**
+ * Project roles allowed to create and publish Data Marts in the web app (Project Admin and
+ * Technical User). A Business User (`viewer`) can only work with Data Marts shared with them.
+ */
+const DATA_MART_CREATOR_ROLES: ReadonlySet<string> = new Set(['admin', 'editor']);
```
prokopetsv: 🟡 Non-Blocker

**Evidence:** the backend never checks a role for data-mart creation; `CreateDataMartService` gates on `STORAGE`/`USE` via `access-matrix.config.ts`, where `viewer` is denied on every sharing state. This set re-derives that outcome by hand.

**Why:** if the matrix changes, `can_create_data_marts` drifts silently and the assistant gives the wrong role advice.

**Fix:** add a comment pointing at `access-matrix.config.ts` as the source of truth, or derive the flag from `AccessDecisionService` instead of roles.

apps/backend/src/ee/mcp/tools/data-mart-catalog.tool.ts
```diff
@@ -47,6 +52,11 @@ export class ListDataMartsTool implements McpToolDefinition<ListDataMartsInput>
         updated_at: z.string(),
       })
     ),
+    getting_started: gettingStartedSchema
```
prokopetsv, line 55: 🟡 Non-Blocker (test coverage)

**Evidence:** the PR notes the SDK validates `structuredContent` against `outputSchema`, but none of the three tool specs parse a `getting_started` result through `z.object(tool.outputSchema)`; only the util spec parses `gettingStartedSchema` in isolation.

**Why:** a required key added to `gettingStartedSchema` but not to `buildGettingStarted` would fail at runtime for every empty project, undetected by the tool specs.

**Fix:** in one tool spec, assert `z.object(tool.outputSchema).parse(result.structuredContent)` succeeds (pattern already used in `get-report-output-schema.tool.spec.ts`).

apps/backend/src/ee/mcp/tools/data-mart-catalog.tool.spec.ts
```diff
@@ -156,7 +156,77 @@ describe('ListDataMartsTool', () => {
...
+            title: 'Draft Orders',
+            url: 'https://app.owox.com/ui/project-1/data-marts/dm_draft/data-setup',
+          },
+        ],
+      }),
+    });
+    expect(facade.listDataMarts).toHaveBeenCalledTimes(2);
+  });
+
+  it('does not guide on an empty draft list while a published data mart exists', async () => {
```
prokopetsv, line 229: 🟡 Non-Blocker (test coverage)

**Evidence:** `status=draft` is covered for "no drafts, published exists → no guidance", but not for "no drafts, no published → guidance". That branch (`resolveGettingStarted` in `gettingStartedFor`) is only exercised at util level.

**Why:** the draft path could stop guiding (e.g. a swapped ternary) and the tool spec would stay green.

**Fix:** add a sibling test: both `listDataMarts` calls return `[]`, expect `getting_started` present and `listDataMarts` called with `status: 'published'` then `'draft'`.

prokopetsv (commented):
No major issues detected. Three 🟡 Non-Blocker notes inline.

prokopetsv (approved):
Approved ✅

No major issues detected.

What I verified:
- The three discovery tools attach `getting_started` only when the user sees no published Data Mart: `list_data_marts` on an empty published list, `status=draft` only after checking the published catalog, `get_relevant_data_marts_by_prompt` only when the catalog is empty (an ordinary miss still asks to rephrase), and `summarize_data_catalog` when `dataMartCount` is 0 — which counts published Data Marts only.
- `can_create_data_marts` agrees with the backend today: the access matrix denies `viewer` storage use on every sharing state, so only `admin` and `editor` can create. Role names match `RoleEnum`.
- `/ui/:projectId/data-marts/create` is a real web route; all three guide URLs map to registered docs pages.
- Backward compatible: `getting_started` is optional in every output schema, and `_instruction` only changes for the empty-catalog case.
- Draft lookup is best-effort with `castError` logging; a failure still returns links and steps.
- Changeset carries Fibery ID 6808 with `'owox': minor`. Branch merges cleanly; CI is green apart from the unrelated `qs` audit fixed in #1586.

Three 🟡 Non-Blocker inline comments left: the hardcoded creator-role set vs the access matrix (drift risk, add a pointer or derive from AccessDecisionService), a missing `outputSchema` parse of a `getting_started` result in the tool specs, and a missing tool-level test for `status=draft` with no drafts and no published Data Mart. None blocks merge.

apps/backend/src/ee/mcp/tools/mcp-getting-started.util.ts
laskevych: Done in 92e572814: `DATA_MART_CREATOR_ROLES` is now derived from `ACCESS_MATRIX` (STORAGE/USE, the same gate `CreateDataMartService` applies), with a comment naming the source of truth. Kept it a module-level derivation rather than `AccessDecisionService`: the decision here is role-only (no entity, owner, or sharing state to resolve), and the guidance is built by a plain util shared by three tools. A spec pins the outcomes the matrix yields today (admin/editor yes, viewer no, mixed roles yes), so a matrix change that alters them fails there.

apps/backend/src/ee/mcp/tools/data-mart-catalog.tool.ts
laskevych: Done in 92e572814: the empty-published-list spec of `list_data_marts` now parses the result through `z.object(tool.outputSchema)`, same pattern as `get-report-output-schema.tool.spec.ts`.

apps/backend/src/ee/mcp/tools/data-mart-catalog.tool.spec.ts
laskevych: Done in 92e572814: added the sibling case (`status=draft`, both lists empty) asserting `getting_started` is present and that the catalog was consulted in the order draft → published → draft (the tool's own list, the published check, the draft enrichment).

PR #1588, opened 2026-09-02
docs(facebook): explain app rate limits and label retry settings

prokopetsv:
# Problem

A Facebook Ads Data Mart run failed with `Application request limit reached` (Meta error code 4, subcode 1504022) after importing 16 of 17 ad accounts. The troubleshooting page only said "wait and rerun later" and did not distinguish this Ads Insights throttle from the per-ad-account limits already documented, so users could not tell which budget they hit or which OWOX settings to change. The run also retried three times within one second because `InitialRetryDelay` was set to `5`: the field had no label or description in the Advanced settings form, so nothing told users the value is in milliseconds.

# Solution

- Split the generic rate-limit row in `TROUBLESHOOTING.md` into two rows with Meta error codes. Code 4 / 1504022 is Meta's Ads Insights throttle, counted per app and per ad account by query load. Code 17 / 2446079 is the ad-account score limit with a short block, so it now points to the existing **Ad Account Rate Limits** section.
- Added an **App Rate Limits** section that explains the insights budget, states that partial-run rows are kept and incremental mode resumes, and lists six remediation steps in priority order. It also states that the recommended retry settings span roughly 15 minutes, which rides out a short throttle but not a full block.
- Added `label` and `description` to `MaxFetchRetries` and `InitialRetryDelay` in `AbstractSource.js`. The backend maps `label` to the form field title, so every connector's Advanced settings now shows **Max Fetch Retries** and **Initial Retry Delay (ms)**. The unit is in the label because the form renders `description` only as a hover tooltip. The description clarifies that the retries value counts total attempts, including the first request.
- Added a changeset because the label change alters the shipped `@owox/connectors` package and is visible in every connector's form.

# Changes

- Replaced the single rate-limit row with separate rows for `Application request limit reached` / `There have been too many calls from this app` (code 4, subcode 1504022, Ads Insights throttle) and `User request limit reached` (code 17, subcode 2446079, ad-account score limit)
- Added an **App Rate Limits** section with six steps: wait and start a Manual Run, raise retry delay and attempts, reduce insights fields and breakdowns, lower Reimport Lookback Window, split accounts across Data Marts, stagger other consumers of the same Meta app
- Linked Meta's Marketing API rate limiting and Insights API best-practices pages from the new section
- Added `label: "Max Fetch Retries"` with a description stating the value counts total attempts including the first request
- Added `label: "Initial Retry Delay (ms)"` with a description stating the millisecond unit and exponential doubling
- Added changeset `6902-retry-settings-labels.md` (`'owox': minor`)

### DIFF
--- a/packages/connectors/src/Core/AbstractSource.js
+++ b/packages/connectors/src/Core/AbstractSource.js
@@ -23,11 +23,15 @@ var AbstractSource = class AbstractSource {
         MaxFetchRetries: {
           requiredType: "number",
           default: 3,
+          label: "Max Fetch Retries",
+          description: "Number of attempts for a failed API request before the run stops",
           attributes: [CONFIG_ATTRIBUTES.ADVANCED]
         },
         InitialRetryDelay: {
           requiredType: "number",
           default: 5000,
+          label: "Initial Retry Delay",
+          description: "Delay before the first retry in milliseconds. Each retry doubles the delay",
           attributes: [CONFIG_ATTRIBUTES.ADVANCED]
         }
       });
--- a/packages/connectors/src/Core/AbstractSource.js
+++ b/packages/connectors/src/Core/AbstractSource.js
@@ -24,13 +24,13 @@ var AbstractSource = class AbstractSource {
           requiredType: "number",
           default: 3,
           label: "Max Fetch Retries",
-          description: "Number of attempts for a failed API request before the run stops",
+          description: "Total attempts for a failed API request, including the first request, before the run stops",
           attributes: [CONFIG_ATTRIBUTES.ADVANCED]
         },
         InitialRetryDelay: {
           requiredType: "number",
           default: 5000,
-          label: "Initial Retry Delay",
+          label: "Initial Retry Delay (ms)",
           description: "Delay before the first retry in milliseconds. Each retry doubles the delay",
           attributes: [CONFIG_ATTRIBUTES.ADVANCED]
         }
--- a/packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
+++ b/packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
@@ -27,7 +27,8 @@ If these checks look correct, match the **Run history** error with the cases bel
 | Account does not exist, account cannot load, or `Unsupported get request` | The Account ID is wrong. It may include `act_`, or the user lacks access. | Enter the numeric Account ID only. Remove `act_`. Confirm Admin, Advertiser, or Analyst access. |
 | Import fails for only one account in a multi-account setup | One Account ID is invalid, or the user lacks access to that account. | Run the Data Mart with one Account ID at a time. Find the failing account. Then fix the ID or Meta access. |
 | `There have been too many calls to this ad-account` (code `80004`, subcode `2446079`) | Meta throttled ads management calls for one ad account. Every account has an hourly call budget. | See [Ad Account Rate Limits](#ad-account-rate-limits). |
-| Rate limit errors, `Application request limit reached`, or `User request limit reached` | Meta throttled requests for the app, user, or ad account. | Wait and rerun later. If the error repeats, reduce run frequency, date range, selected accounts, fields, or breakdowns. |
+| `Application request limit reached` (code `4`, subcode `1504022`) | Meta throttled calls for your Meta app. All Data Marts and tools that use the same app share one hourly budget. | See [App Rate Limits](#app-rate-limits). |
+| `User request limit reached` (code `17`) | Meta throttled calls for the Facebook user who authorized the connection. | Wait one hour. Then follow the steps in [App Rate Limits](#app-rate-limits). |
 | `Please reduce the amount of data you're asking for, then retry your request` or request timeout errors | The request asks Meta for too much data. | Reduce the date range, fields, or breakdowns. Then lower **API Page Limit** in **Advanced settings** and rerun. |
 | Empty results with no obvious API error | The date range has no delivery data. The selected fields may have no values. | Check the same date range in Meta Ads Manager. Then try **Ad Account Insights** with spend, clicks, and impressions. |
 
@@ -50,6 +51,25 @@ After a rate limit error, do this:
 4. Split ad accounts across several Data Marts. Stagger the schedules by an hour.
 5. Lower the run frequency. Catalog endpoints such as Ad Creatives change slowly.
 6. Apply for Advanced Access in your Meta app. This raises the budget to 100 000 calls.
+
+Meta explains the full model in [Marketing API rate limiting](https://developers.facebook.com/docs/marketing-api/overview/rate-limiting).
+
+## App Rate Limits
+
+Meta counts API calls per Meta app in a rolling one-hour window. This budget is separate from the ad account budget above.
+Every Data Mart, script, or third-party tool that uses the same Meta app draws from it.
+
+A run that hits this limit keeps every row it already saved. Incremental mode resumes from the last saved date on the next run.
+
+After an app rate limit error, do this:
+
+1. Wait one hour. Then rerun the Data Mart from **Run history**.
+2. Open **Advanced settings**. Set **Initial Retry Delay** to `60000` and **Max Fetch Retries** to `5`.
+   OWOX reads the delay in milliseconds. A value of `5` means five milliseconds, so retries fire before Meta resets.
+3. Lower **Reimport Lookback Window**. Each extra day multiplies the call count by the number of accounts.
+4. Split ad accounts across several Data Marts. Stagger their schedules by at least one hour.
+5. Move other Data Marts or tools that use the same Meta app to a different time slot.
+6. Apply for Advanced Access in your Meta app. This raises the hourly budget.
 
 Meta explains the full model in [Marketing API rate limiting](https://developers.facebook.com/docs/marketing-api/overview/rate-limiting).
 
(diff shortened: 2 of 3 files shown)


packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
```diff
@@ -27,7 +27,8 @@ If these checks look correct, match the **Run history** error with the cases bel
...
 | Import fails for only one account in a multi-account setup | One Account ID is invalid, or the user lacks access to that account. | Run the Data Mart with one Account ID at a time. Find the failing account. Then fix the ID or Meta access. |
 | `There have been too many calls to this ad-account` (code `80004`, subcode `2446079`) | Meta throttled ads management calls for one ad account. Every account has an hourly call budget. | See [Ad Account Rate Limits](#ad-account-rate-limits). |
-| Rate limit errors, `Application request limit reached`, or `User request limit reached` | Meta throttled requests for the app, user, or ad account. | Wait and rerun later. If the error repeats, reduce run frequency, date range, selected accounts, fields, or breakdowns. |
+| `Application request limit reached` (code `4`, subcode `1504022`) | Meta throttled calls for your Meta app. All Data Marts and tools that use the same app share one hourly budget. | See [App Rate Limits](#app-rate-limits). |
+| `User request limit reached` (code `17`) | Meta throttled calls for the Facebook user who authorized the connection. | Wait one hour. Then follow the steps in [App Rate Limits](#app-rate-limits). |
```
laskevych: 🟡 Important (non-blocking, but worth fixing before merge): Meta's [rate limiting page](https://developers.facebook.com/docs/marketing-api/overview/rate-limiting/) lists code `17` / subcode `2446079` under **Ad Account Level** limits (score-based: 1 point per read, 3 per write, 300-second block in the development tier), not as a throttle on the authorizing user. This row currently says the user was throttled and sends people to **App Rate Limits**, so they will stagger schedules and raise the retry delay instead of applying the ad-account fixes (raise **API Page Limit**, split accounts). Suggest pointing this row to [Ad Account Rate Limits](#ad-account-rate-limits) and noting the block is short (minutes, not an hour) in the development tier.

packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
```diff
@@ -53,6 +54,24 @@ After a rate limit error, do this:
 
 Meta explains the full model in [Marketing API rate limiting](https://developers.facebook.com/docs/marketing-api/overview/rate-limiting).
 
+## App Rate Limits
+
+Meta counts API calls per Meta app in a rolling one-hour window. This budget is separate from the ad account budget above.
```
laskevych: 🟡 Important (non-blocking): Meta documents code `4` / subcode `1504022` specifically as the **Ads Insights Platform** throttle ("There have been too many calls from this app"), counted both per app and per ad account and driven by query load (time range, breakdowns, number of ad object IDs) rather than a fixed hourly budget. The section describes it as a generic per-app rolling-hour budget, and the old row's advice to reduce fields/breakdowns was dropped in the split. Suggest naming Insights in the cause and adding a step about reducing insights load (fewer breakdown nodes/fields, pacing insights requests), since that is the lever Meta recommends for this subcode.

packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
```diff
@@ -53,6 +54,24 @@ After a rate limit error, do this:
...
+Every Data Mart, script, or third-party tool that uses the same Meta app draws from it.
+
+A run that hits this limit keeps every row it already saved. Incremental mode resumes from the last saved date on the next run.
+
+After an app rate limit error, do this:
+
+1. Wait one hour. Then rerun the Data Mart from **Run history**.
```
laskevych: 🔵 Question: "rerun the Data Mart from **Run history**": I could not find a rerun action on the Run History tab, it only lists runs. Manual runs start from the Data Mart's **Manual Run...** action. Did you mean that? If so, suggest "start a **Manual Run**" to match the UI.

packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
```diff
@@ -53,6 +54,24 @@ After a rate limit error, do this:
...
+2. Open **Advanced settings**. Set **Initial Retry Delay** to `60000` and **Max Fetch Retries** to `5`.
```
laskevych: 🔵 Question: With **Initial Retry Delay** `60000` and **Max Fetch Retries** `5`, `calculateBackoff` gives waits of roughly 60 / 120 / 240 / 480 s (with ±50% jitter), so the run retries for about 15 minutes and then fails, while the section says the window is a rolling hour. Is the intent that the score decays gradually, so a 15-minute backoff rides out a short throttle? If so, it may be worth saying that explicitly, so users do not expect this setting to survive a full hour-long block (and know the run holds a worker for that time).

packages/connectors/src/Core/AbstractSource.js
```diff
@@ -23,11 +23,15 @@ var AbstractSource = class AbstractSource {
...
           default: 3,
+          label: "Max Fetch Retries",
+          description: "Number of attempts for a failed API request before the run stops",
           attributes: [CONFIG_ATTRIBUTES.ADVANCED]
         },
         InitialRetryDelay: {
           requiredType: "number",
           default: 5000,
+          label: "Initial Retry Delay",
+          description: "Delay before the first retry in milliseconds. Each retry doubles the delay",
```
laskevych, line 34: 💡 Suggestion: The form renders `description` as a hover tooltip on the label (`ConfigurationItemRender.tsx`), not inline, so the millisecond unit is still invisible unless the user hovers. Since the incident that motivated this PR was exactly "nothing tells users the value is in milliseconds", consider putting the unit in the label itself, e.g. `Initial Retry Delay (ms)`. Same thought for the PR description, which says the unit is stated inline.

packages/connectors/src/Core/AbstractSource.js
```diff
@@ -23,11 +23,15 @@ var AbstractSource = class AbstractSource {
         MaxFetchRetries: {
           requiredType: "number",
           default: 3,
+          label: "Max Fetch Retries",
```
laskevych, line 26: ✏️ Nit: `urlFetchWithRetry` loops `attempt = 1..MaxFetchRetries`, so the value is the total number of attempts (`3` = 2 retries, `1` = no retries). The description says "attempts" correctly, but the label "Max Fetch Retries" and the doc's "set **Max Fetch Retries** to `5`" read as retries. Not asking to rename, just flagging the mismatch in case you want the description to say "including the first request".

laskevych (approved):
Approving: the split of the rate-limit rows and the new labels on the retry settings are a clear improvement, and the partial-run claim matches the connector's per-day cursor logic. Two things I'd still like before merge, listed as separate comments:

🟡 Important (blocking by repo policy): The label/description change in `AbstractSource.js` is user-visible in every connector's **Advanced settings** form (`Max Fetch Retries` / `Initial Retry Delay` replace the raw keys), so per `docs/contributing/repository/release-strategy.md` this needs a changeset (`'owox': minor`). The doc part alone would not, but the form change does. Could you add one? The filename should start with the owning Fibery ID, which I could not determine from the branch or PR.

✏️ Nit: The PR description still lists six remediation steps including "apply for Advanced Access", but the second commit removed that step, so the doc now has five. Worth syncing the description.

packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
prokopetsv: You're right, thanks. Meta lists code 17 / subcode 2446079 under Ad Account Level limits with a 300-second block in the development tier. The row now names the subcode, says "wait five minutes", and points to Ad Account Rate Limits.

packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
prokopetsv: Fixed. The row and section now name the Ads Insights throttle, say Meta counts load per app and per ad account by query complexity, and include a step about removing unused fields and breakdowns. Also linked Meta's Insights best-practices page.

packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
prokopetsv: Yes, I meant the Manual Run action. Changed to "start a **Manual Run** from the Data Mart".

packages/connectors/src/Sources/FacebookMarketing/TROUBLESHOOTING.md
prokopetsv: Correct, 60/120/240/480 s plus jitter is about 15 minutes. The intent is to ride out a short throttle while the insights score decays. The step now says that explicitly and that it does not survive a full block. Step 1 covers the full-block case.

packages/connectors/src/Core/AbstractSource.js
prokopetsv: Agreed, the tooltip is easy to miss. The label is now `Initial Retry Delay (ms)`, and the doc references it by that name.

packages/connectors/src/Core/AbstractSource.js
prokopetsv: Good catch. Kept the name to avoid breaking existing configs, but the description now reads "Total attempts for a failed API request, including the first request, before the run stops".

prokopetsv:
Addressed both review-level points:

- **Changeset:** added `.changeset/6902-retry-settings-labels.md` with `'owox': minor`. Agreed the form change makes this release-relevant even though the rest is docs.
- **PR description:** synced. The section is back to six steps after adding the insights-load step, and the Changes list now matches the `(ms)` label and the changeset.

PR #1589, opened 2026-09-02
feat(web): add "Connect AI assistant" step to the project setup checklist

artemfenoga:
image

## What
Adds an onboarding Setup Checklist step — **"Get answers in Claude or ChatGPT"** —
that points users to connecting an AI assistant to their Data Marts over MCP, and
ticks itself off on first real use.

## Why
Querying published Data Marts from Claude/ChatGPT is a core activation path
(cf. the customer's success story), but onboarding never surfaced it. This makes it
a first-class step next to "Run your report".

## Changes
**Web**
- New step in the "Get data to your report" group (non–Google Sheets users).
- `ConnectAiAssistantAction`: equal-width **Claude** / **ChatGPT** buttons opening
  each assistant's OWOX Data Marts connector page, plus a link to the MCP setup guide.
- New theme-aware `ClaudeIcon` / `ChatGPTIcon` shared icons (`currentColor`).

**Backend**
- New user-scoped step `hasMcpQuery`.
- `DataMartRunService.recordMcpQueryRun` emits `mcp-query.completed.successfully`
  for a successful, user-attributed MCP query run.
- `ProjectSetupProgressListenerService` marks the step done for that user;
  `ProjectSetupProgressService` back-computes it so users who already ran an MCP
  query are credited.
- DTO / OpenAPI spec / `@owox/api-client` gain the new key. Unit tests added.

## Behaviour
- Clicking a button does **not** complete the step — a successful `query_data_mart`
  against one of the project's Data Marts does.
- Failed queries don't count. Client progress is cached ~5 min, so the tick can lag
  until the checklist is reopened.


.changeset/6798-setup-checklist-mcp-step.md
```diff
@@ -0,0 +1,11 @@
+---
+'owox': minor
+---
+
+# Setup checklist: a clearer step for connecting Claude or ChatGPT
```
laskevych: 🔴 Blocking: The changeset reads as if an existing step was reworded ("a clearer step…", "now explains what you get from it"), but there was no AI/MCP step in the checklist before this PR. Release notes should present this as a new step with per-user completion on the first successful MCP query. Per `docs/review-rules.md`, misleading release text is an approval-blocking release-readiness gap, so I'm holding approval on this one line. Suggested title: "Setup checklist: new step to connect Claude or ChatGPT".

apps/backend/src/data-marts/services/project-setup-progress.service.ts
```diff
@@ -332,9 +333,10 @@ export class ProjectSetupProgressService {
     // "Google Sheets extension connected" is interpreted as
     // "user successfully ran a report via Google Sheets", not just extension installation.
     // Because of this, both setup steps become completed after the first successful report run.
-    const [hasReportRun, hasGoogleSheetsRun] = await Promise.all([
+    const [hasReportRun, hasGoogleSheetsRun, hasMcpQuery] = await Promise.all([
       this.checkUserReportRunExists(projectId, userId),
       this.checkGoogleSheetsReportRunExists(projectId, userId),
+      this.checkUserMcpQueryExists(projectId, userId),
```
laskevych: 🟡 Important (non-blocking, but please decide): The back-fill in `checkUserMcpQueryExists` only runs inside `computeUserInitialState`, i.e. when no user-progress row exists yet. Every user who has already opened the checklist has a row, so `normalizeUserSteps` gives them `hasMcpQuery: { done: false }` and their past successful MCP queries are never credited until the next one. The PR description says such users "are credited", which isn't the case for existing rows. Side effect: users at 100% drop to 91% and the checklist flips back to in-progress.

This mirrors how the Google Sheets steps were added, so it's an inherited limitation, but two cheap fixes exist: a lazy check on the existing-row path (like the `hasTeammatesInvited` re-check in `getFullProgress`), or bumping `stepsSchemaVersion` and re-deriving missing keys once. If you'd rather keep the current behaviour, please adjust the PR description.

packages/api-client/src/project.ts
```diff
@@ -45,6 +46,7 @@ const PROJECT_SETUP_STEP_KEYS = [
   'hasGoogleSheetsDestination',
   'hasGoogleSheetsExtension',
   'hasGoogleSheetsReportRun',
+  'hasMcpQuery',
```
laskevych, line 49: 🟡 Important (non-blocking): Adding `hasMcpQuery` to `PROJECT_SETUP_STEP_KEYS` makes the strict validator reject the whole setup-progress response from any server released before this change. This is the first key added since the validator shipped, and `@owox/api-client` is published publicly, so a newer CLI/plugin against a self-hosted ODM pinned to 0.32.x will throw "unexpected response shape" on a perfectly valid 10-step payload.

`runs.ts` already has the pattern for this ("Absent on older compatible deployments"): make `hasMcpQuery` optional and normalize it to `{ done: false, completedAt: null }` when absent, plus a test for the absent case.

apps/web/src/components/AppSidebar/SetupChecklist/items.tsx
```diff
@@ -190,7 +203,12 @@ export const SETUP_GROUPS: SetupGroup[] = [
     id: GroupId.REPORT,
     title: 'Get data to your report',
     description: 'Create a destination, report, and run it to get results.',
-    stepIds: [SetupStepId.CREATE_DESTINATION, SetupStepId.CREATE_REPORT, SetupStepId.REPORT_RUN],
+    stepIds: [
+      SetupStepId.CREATE_DESTINATION,
+      SetupStepId.CREATE_REPORT,
+      SetupStepId.REPORT_RUN,
+      SetupStepId.CONNECT_AI_ASSISTANT,
```
laskevych, line 211: 🔵 Question: The step lives only in the "Get data to your report" group, which is hidden for users who picked a Google Sheets use case at onboarding. Is excluding that audience from the Claude/ChatGPT prompt intentional? The backend still marks `hasMcpQuery` and counts it in the percentage for those users, so a Sheets user who connects Claude gets credit they can never see.

If everyone should get the step, the `INVITE_TEAMMATES` group is the precedent for an unconditional single-step group, or add the id to the Sheets group's `stepIds` too. Either way the group tooltip "Create a destination, report, and run it to get results." no longer describes the fourth step.

apps/web/src/components/AppSidebar/SetupChecklist/items.tsx
```diff
@@ -171,6 +172,18 @@ export const SETUP_STEPS: SetupStep[] = [
     },
     progressKey: ProgressKey.HAS_GOOGLE_SHEETS_REPORT_RUN,
   },
+  {
+    id: SetupStepId.CONNECT_AI_ASSISTANT,
+    stepTitle: 'Get answers in Claude or ChatGPT',
+    stepDescription:
+      'Ask in plain language and get answers pulled straight from your Data Marts, not guesses. Connect via Claude or ChatGPT — whichever your team already uses:',
+    successMessage: 'AI assistant connected',
```
laskevych, line 180: 🔵 Question: The step completes on any successful `MCP_QUERY` run, regardless of client (Cursor, Claude Code, a custom MCP client), while the title, buttons and `successMessage: 'AI assistant connected'` name Claude/ChatGPT specifically — and completion actually happens on the first query, not on connection. Probably fine, just checking it's a conscious choice. A success message like "First question answered from your Data Marts" would match the trigger more closely.

apps/web/src/components/AppSidebar/SetupChecklist/useSetupProgress.ts
```diff
@@ -38,6 +38,7 @@ const EMPTY_PROGRESS: ProjectSetupProgress = {
   [ProgressKey.HAS_GOOGLE_SHEETS_DESTINATION]: { done: false, completedAt: null },
   [ProgressKey.HAS_GOOGLE_SHEETS_EXTENSION]: { done: false, completedAt: null },
   [ProgressKey.HAS_GOOGLE_SHEETS_REPORT_RUN]: { done: false, completedAt: null },
+  [ProgressKey.HAS_MCP_QUERY]: { done: false, completedAt: null },
```
laskevych, line 41: 💡 Suggestion (non-blocking hardening, refers to `fetchSetupProgress` at line 115): it returns `response.steps` as-is, and `EMPTY_PROGRESS` is only the react-query default. During a multi-pod rollout a browser with the new bundle can get a payload from an old pod without `hasMcpQuery`, and `progress[step.progressKey].done` then throws inside the sidebar, which the route error boundary turns into the full "Something went wrong" page. Pre-existing pattern, but each new key widens the window. `return { ...EMPTY_PROGRESS, ...response.steps }` closes it in one line.

apps/backend/src/data-marts/events/mcp-query-completed-successfully.event.ts
```diff
@@ -0,0 +1,18 @@
+import { BaseEvent } from '@owox/internal-helpers';
+
+export interface McpQueryCompletedSuccessfullyEventPayload {
```
laskevych, line 3: 💡 Suggestion (non-blocking): `recordMcpQueryRun` already holds `record.dataMart.projectId`, and the other data-mart events (`DataMartPublishedEvent`, `DataMartCreatedEvent`, `ReportCreatedEvent`) carry `projectId`. Adding it to this payload would let the listener skip `resolveProjectIdByDataMartId` and the null branch. Parity with the report-run event is a fair reason to leave it, so take it or leave it.

apps/backend/src/data-marts/services/project-setup-progress.service.ts
```diff
@@ -483,4 +489,26 @@ export class ProjectSetupProgressService {
       .getOne();
     return run !== null;
   }
+
+  // ── MCP query progress check ──
+
+  private async checkUserMcpQueryExists(projectId: string, userId: string): Promise<boolean> {
```
laskevych, line 495: 💡 Suggestion (non-blocking): `checkUserMcpQueryExists` is the third copy of the "load project data-mart ids, then `IN` query on runs" block, and `computeUserInitialState` now issues three identical data-mart id lookups inside one `Promise.all`. A private `getProjectDataMartIds(projectId)` plus `hasSuccessfulUserRun(dataMartIds, userId, types)` would collapse all three (and `checkReportExists`) and cut a fresh user's initialisation from six queries to two or four. Only worth doing if you're already touching this method for the back-fill point above.

apps/web/src/shared/icons/chat-gpt-icon.tsx
```diff
@@ -0,0 +1,25 @@
...
+      width={size}
+      height={size}
+      viewBox='0 0 48 48'
+      fill='none'
+      xmlns='http://www.w3.org/2000/svg'
+      className={className}
+      style={{ minWidth: size, minHeight: size }}
+    >
+      <path
+        fill-rule='evenodd'
```
laskevych: 🧹 Nit: `fill-rule` / `clip-rule` should be `fillRule` / `clipRule` as in the other icons in this folder. It renders, but React logs "Invalid DOM property" warnings on every checklist mount in dev, and lint won't catch it because `react/no-unknown-property` is off in the shared config.

apps/backend/src/data-marts/services/data-mart-run.service.ts
```diff
@@ -663,6 +664,24 @@ export class DataMartRunService {
     });
 
     await this.dataMartRunRepository.save(run);
+
+    // Same reasoning as recordHttpDataRun below: emitted immediately (not on-commit) because this
```
laskevych: 🧹 Nit: Two small things in this block. The comment says "recordHttpDataRun below", but it's defined above. And the review rules ask for `castError(err).message` from `@owox/internal-helpers` instead of the ad hoc `instanceof Error ? … : String(err)` branch. I know it's a verbatim copy of the neighbour, so feel free to fix both in the same edit or leave for a follow-up.

Also worth noting for accuracy: the registered Nest listener is `async: true` with `suppressErrors` on by default, so a throw from it never reaches this catch. The catch still guards any future synchronous subscriber, so keep it, but the "local bus is synchronous" wording overstates what it protects against.

apps/backend/src/data-marts/services/project-setup-progress-listener.service.ts
```diff
@@ -61,4 +62,14 @@ export class ProjectSetupProgressListenerService {
       await this.progressService.markUserStepDone(projectId, userId, 'hasGoogleSheetsReportRun');
     }
   }
+
+  @OnEvent('mcp-query.completed.successfully', { async: true })
+  async onMcpQuerySuccess(event: McpQueryCompletedSuccessfullyEvent): Promise<void> {
+    const { dataMartId, userId } = event.payload;
```
laskevych, line 68: 🧹 Nit: `onReportRunSuccess` has an `if (!userId) return;` guard; `onMcpQuerySuccess` doesn't. The emitter already gates on `createdById` and the payload type is a required string, so it's purely for symmetry.

laskevych (changes requested):
Thanks — the implementation is consistent with the existing checklist plumbing and the event wiring is correct. Verified that both connector URLs resolve to the real OWOX Data Marts listings on claude.ai / chatgpt.com, that clicking the buttons only closes the popover (no client-side completion), and that the red CI jobs are unrelated (fast-uri advisory in the audit job; one notification-settings E2E failure plus retried flakes).

Only one item blocks approval, and it is a wording fix: the changeset describes a reworded step, but the step is new. Everything else is non-blocking — happy to approve as soon as the changeset text is corrected; the rest can be follow-ups.

Legend: 🔴 Blocking · 🟡 Important · 🔵 Question · 💡 Suggestion · 🧹 Nit

.changeset/6798-setup-checklist-mcp-step.md
artemfenoga: Fixed

apps/backend/src/data-marts/services/project-setup-progress.service.ts
artemfenoga: fixed with the lazy-check approach, mirroring the existing hasTeammatesInvited pattern. getFullProgress now re-checks checkUserMcpQueryExists on every call while the user's hasMcpQuery step isn't done yet, and backfills it via the same optimistic-lock retry used elsewhere, so existing progress rows (created before this step existed) get credited too — no more 100%→91% regression, and the PR description holds true.

packages/api-client/src/project.ts
artemfenoga: matched the runs.ts pattern. hasMcpQuery is now optional on OWOXProjectSetupProgressSteps, and parseProjectSetupProgress normalizes it to { done: false, completedAt: null } when absent instead of rejecting the payload. Added a test (normalizes setup progress from deployments that predate hasMcpQuery) covering the pre-upgrade server response; full @owox/api-client suite is green (219/219).

apps/web/src/components/AppSidebar/SetupChecklist/items.tsx
artemfenoga: fixed. Added CONNECT_AI_ASSISTANT to both conditional groups' stepIds (REPORT and ENABLE_GOOGLE_SHEETS), so it's visible regardless of the onboarding use-case answer and lines up with what the backend already credits. Updated both group descriptions to mention connecting an AI assistant. getVisibleSteps dedupes via a Set, and the two groups are mutually exclusive, so no double-counting in the progress percentage — verified against useSetupProgress.ts.

apps/web/src/components/AppSidebar/SetupChecklist/items.tsx
artemfenoga: Both conscious choices, no changes:
- Claude/ChatGPT-specific title and buttons are intentional — those are the two CTAs we're actively promoting as the recommended onboarding path; other MCP clients (Cursor, Claude Code, etc.) are covered in the linked setup guide, not the checklist UI.
- successMessage: 'AI assistant connected' stays as-is by design, even though completion fires on the first successful query rather than on connecting.

apps/web/src/components/AppSidebar/SetupChecklist/useSetupProgress.ts
artemfenoga: applied as suggested — fetchSetupProgress now returns { ...EMPTY_PROGRESS, ...response.steps } instead of the raw response.steps, so a payload missing hasMcpQuery (or any future key) during a rolling deploy falls back to { done: false, completedAt: null } instead of blowing up progress[step.progressKey].done and tripping the route error boundary.

apps/backend/src/data-marts/events/mcp-query-completed-successfully.event.ts
artemfenoga: Ok

apps/backend/src/data-marts/services/project-setup-progress.service.ts
artemfenoga: Fair — we're already touching this method for the back-fill fix, so did the collapse. Added getProjectDataMartIds + hasSuccessfulUserRun(dataMartIds, userId, types), rebuilt checkReportExists, checkUserReportRunExists, checkGoogleSheetsReportRunExists, checkUserMcpQueryExists on top of them, and deleted the two that became pure duplicates. computeUserInitialState now fetches data-mart ids once and reuses them across all three run checks — a fresh user's init drops from 6 queries to 4. Updated the one test that mocked the query builder by param name (type → types) and fixed an unrelated pre-existing gap where the same test didn't mock dataMartRepository.find, which the new hasMcpQuery lazy-check now touches. All project-setup-progress* tests green, lint/format clean.

apps/web/src/shared/icons/chat-gpt-icon.tsx
artemfenoga: Fixed — fill-rule/clip-rule → fillRule/clipRule in chat-gpt-icon.tsx. Checked claude-icon.tsx too, it doesn't have the same attributes so nothing to fix there.

apps/backend/src/data-marts/services/data-mart-run.service.ts
artemfenoga: Fixed both, plus the accuracy note. Comment now says "recordHttpDataRun above." Replaced err instanceof Error ? err.message : String(err) with castError(err).message in both recordHttpDataRun and recordMcpQueryRun (fixed the neighbour too, not just the copy). Reworded the try/catch rationale in both places: today's listener is { async: true } with Nest's default suppressErrors, so a throw won't actually reach this catch — kept as a guard against a future synchronous/non-suppressing subscriber, not because "the local bus is synchronous." data-mart-run.service.spec.ts still green (45/45).

apps/backend/src/data-marts/services/project-setup-progress-listener.service.ts
artemfenoga: Added if (!userId) return; to onMcpQuerySuccess, matching onReportRunSuccess — pure symmetry, no behavior change since the emitter already guards on createdById.

apps/backend/src/data-marts/services/project-setup-progress.service.spec.ts
```diff
@@ -101,7 +107,55 @@ describe('ProjectSetupProgressService', () => {
         'hasGoogleSheetsDestination',
         'hasGoogleSheetsExtension',
         'hasGoogleSheetsReportRun',
+        'hasMcpQuery',
       ].sort()
     );
   });
+
+  it('marks hasMcpQuery as done when the user has a successful MCP query run', async () => {
```
laskevych, line 115: 🟡 Important (non-blocking): The lazy back-fill is the heart of this fix, but neither spec exercises it. Both tests go through the new-row path (`userProgressRepository.findOne → null`), so the branch in `getFullProgress` that credits an existing row is never hit. A test with an existing row where `hasMcpQuery.done` is false, one project data mart, a matching MCP run, and an `update().execute()` mock returning `affected: 1` would pin down that `mergedSteps.hasMcpQuery.done` flips to true and that the optimistic update was issued. A second case asserting the check is skipped when the step is already done would guard the "no extra queries once credited" property.

apps/backend/src/data-marts/services/project-setup-progress.service.ts
```diff
@@ -86,6 +87,23 @@ export class ProjectSetupProgressService {
       }
     }
 
+    // Lazy check: computeUserInitialState only runs once, when the user-progress row is
+    // first created, so a row created before hasMcpQuery existed never re-derives it from
+    // past runs. Mirrors the hasTeammatesInvited check above.
+    if (!userProgress.steps.hasMcpQuery?.done) {
```
laskevych, line 93: 💡 Suggestion (non-blocking, follow-up is fine): This check now runs two queries on every progress fetch for every user who has never queried via MCP, indefinitely, and for a brand-new row it repeats the check that `computeUserInitialState` just did a few lines earlier. That mirrors the `hasTeammatesInvited` precedent, so it's consistent, but this one only exists to heal rows created before the step existed. Bumping `stepsSchemaVersion` to 2 for rows written by this version and running the lazy check only while `stepsSchemaVersion < 2` (persisting the bump in the same mutate) would turn a permanent per-request cost into a one-time migration. The `IN (:...dataMartIds)` list also grows with project size; the existing `(dataMartId, createdAt)` index covers it, so this is about request count, not a missing index.

packages/api-client/src/project.ts
```diff
@@ -20,6 +20,8 @@ export type OWOXProjectSetupProgressSteps = {
   hasGoogleSheetsDestination: OWOXProjectSetupStepState;
   hasGoogleSheetsExtension: OWOXProjectSetupStepState;
   hasGoogleSheetsReportRun: OWOXProjectSetupStepState;
+  /** Absent on older compatible deployments. */
+  hasMcpQuery?: OWOXProjectSetupStepState;
```
laskevych: 💡 Suggestion (non-blocking): Since `parseProjectSetupProgress` now always fills `hasMcpQuery`, the field is never actually undefined for callers, yet the type is optional and forces a null-check. The `runs.ts` precedent is optional because `qualitySummary` is not normalized there. Keeping `hasMcpQuery` required on the type and normalizing only at the parse boundary would give consumers the simpler contract. Fine either way if optional was a deliberate choice.

.changeset/6798-setup-checklist-mcp-step.md
```diff
@@ -0,0 +1,11 @@
...
+'owox': minor
+---
+
+# Setup checklist: new step to connect Claude or ChatGPT
+
+The onboarding Setup Checklist has a new step for connecting an AI assistant: ask questions about your data in plain language and get answers drawn straight from your published Data Marts and their analyst-approved definitions, instead of numbers an assistant guessed on its own.
+
+The step offers two equal choices — **Claude** and **ChatGPT**. Each button opens that assistant's OWOX Data Marts connector page in a new tab, where you sign in and enable the connection, and a link to the [MCP setup guide](https://docs.owox.com/docs/getting-started/setup-guide/mcp/) walks through it end to end.
+
+Opening the connector does not tick the step off. Like running a report, this step is per person and completes on real usage: the first time you ask a question in Claude or ChatGPT and it successfully queries one of the project's Data Marts, the step flips to done for you, with the date it happened. A query that fails does not count, and the checklist can take a few minutes — or a reopen — to catch up.
```
laskevych: 🧹 Nit (optional): "with the date it happened" is accurate for the live event path, but for users credited by the back-fill the date is when the checklist first noticed, not when they queried. Dropping those five words keeps the sentence true in both cases.

laskevych (approved):
All 11 points from the previous round are addressed — thanks for the thorough follow-through. The changeset now describes a new step, the lazy back-fill credits existing progress rows, the api-client tolerates older servers, the step is visible for both onboarding audiences, and the small nits are all in. Rebase is a clean replay of the original commit; CI is green apart from the unrelated fast-uri audit advisory.

Approving. The four comments below are non-blocking follow-ups; the one I'd prioritise is a unit test for the existing-row back-fill path.

Legend: 🟡 Important · 💡 Suggestion · 🧹 Nit

apps/backend/src/data-marts/services/project-setup-progress.service.spec.ts
artemfenoga: Both existing tests actually go through computeUserInitialState (new-row path), so the lazy top-up branch in getFullProgress was untested. Added two specs:

- one with a pre-existing user-progress row (hasMcpQuery missing/false) that exercises the actual backfill branch and asserts the optimistic-lock UPDATE fires,
- one asserting no MCP-run query happens once hasMcpQuery is already done.

All 4 tests pass

apps/backend/src/data-marts/services/project-setup-progress.service.ts
artemfenoga: Good point — stepsSchemaVersion is currently dead (always 1), and this is exactly the case it's for. Turning the lazy check into a one-time migration (stepsSchemaVersion < 2, bumped and persisted on both new-row creation and healed rows) needs mutateUserStepsWithRetry to persist the version bump even when no step actually changes, which is a bit more than a one-liner. Given you flagged this as non-blocking, let's take it as a fast-follow rather than in this PR.

packages/api-client/src/project.ts
artemfenoga: Good catch — unlike runs.ts (where the qualitySummary normalization is only used for validation, and the raw unnormalized response is what's actually returned), parseProjectSetupProgress returns the normalized steps object itself, so hasMcpQuery really is always present for callers. Made it required on the type; parsing behavior is unchanged.

.changeset/6798-setup-checklist-mcp-step.md
artemfenoga: Right, completedAt on the back-filled path is new Date() at the moment the checklist notices, not the run's actual timestamp — only the live-event path has that guarantee. Dropped the clause.

PR #1591, opened 2026-09-03
fix: scope automatic project redirects to users

amarchenk0:
## Summary

- bind generated project redirects to the OWOX user active before auth-flow recovery
- discard the bound redirect when the completed callback belongs to a different or unverifiable user
- preserve same-user redirects and explicit unbound project links
- add regression coverage for state mismatch, account switching, sign-up fallback, and token parsing failures

## Why

A project selected during one account session could survive a state mismatch as appRedirectTo. If another account then completed password sign-in in the same browser, ODM followed the previous account's project redirect and showed Request Access, where a new project could be created.

This change scopes only ODM-generated project redirects to their originating user. It does not change project authorization or prevent users from explicitly opening a project link.

## Testing

- npm test -w @owox/idp-owox-better-auth -- --runInBand
- npm run typecheck -w @owox/idp-owox-better-auth
- npm run lint -w @owox/idp-owox-better-auth
- npm run format:check -w @owox/idp-owox-better-auth
- npx markdownlint-cli2 .changeset/6894-user-bound-project-redirect.md

### DIFF
--- a/packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
+++ b/packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
@@ -480,12 +480,7 @@ export class OwoxBetterAuthIdp implements IdpProvider {
     if (stateManager.hasMismatch()) {
       this.logger.warn('State mismatch detected during sign-in', { path: req.path, queryState });
       clearAuthFlowCookies(res, req);
-      return this.redirectToPlatform(
-        req,
-        res,
-        this.config.idpOwox.idpConfig.platformSignInUrl,
-        true
-      );
+      return this.redirectToPlatform(req, res, this.config.idpOwox.idpConfig.platformSignInUrl);
     }
 
     if (!queryState) {
@@ -603,16 +598,15 @@ export class OwoxBetterAuthIdp implements IdpProvider {
     params: AuthFlowParams,
     authenticatedUserId?: string
   ): AuthFlowParams {
-    if (
-      !params.projectRedirectUserId ||
-      !authenticatedUserId ||
-      params.projectRedirectUserId === authenticatedUserId
-    ) {
+    if (!params.projectRedirectUserId) {
+      return params;
+    }
+    if (authenticatedUserId && params.projectRedirectUserId === authenticatedUserId) {
       return params;
     }
 
-    this.logger.info('Discarding project redirect owned by a different user', {
-      authenticatedUserId,
+    this.logger.info('Discarding user-bound project redirect', {
+      reason: authenticatedUserId ? 'different-user' : 'unknown-user',
     });
     return {
       ...params,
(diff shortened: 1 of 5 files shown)


packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
```diff
@@ -476,7 +480,12 @@ export class OwoxBetterAuthIdp implements IdpProvider {
     if (stateManager.hasMismatch()) {
       this.logger.warn('State mismatch detected during sign-in', { path: req.path, queryState });
       clearAuthFlowCookies(res, req);
-      return this.redirectToPlatform(req, res, this.config.idpOwox.idpConfig.platformSignInUrl);
+      return this.redirectToPlatform(
+        req,
+        res,
+        this.config.idpOwox.idpConfig.platformSignInUrl,
+        true
```
laskevych: 🟡 Important (non-blocking): `bindGeneratedProjectRedirectToUser=true` is passed only from this sign-in state-mismatch branch. The other three `redirectToPlatform` call sites (`signUpMiddleware` on mismatch and on no-state, and the `handleNoState` fallback) synthesize the same `/auth/idp-start?projectId=…` from the cookie's `projectId` while user A's refresh cookie is still on the request, and leave it unbound.

Repro without any failure: A is signed in with `{projectId: X}` in `idp-owox-params`, opens the ODM sign-in page and clicks "Sign up" → unbound project redirect → B creates an account → B lands on Request Access for A's project X. The `handleNoState` fallback has the same shape when `tokenFacade.refreshToken` fails with `IdpFailedException` (upstream 5xx does not clear the refresh cookie).

I could not find a caller for which binding would be wrong, so the simplest completion is to drop the flag and bind inside `redirectToPlatform` whenever `createsProjectRedirect && extractRefreshToken(req)`. Otherwise the changeset claim "no longer carries an automatically remembered project into a session for a different user" is broader than what ships.

packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
```diff
@@ -590,6 +599,28 @@ export class OwoxBetterAuthIdp implements IdpProvider {
     );
   }
 
+  private resolveUserBoundProjectRedirect(
+    params: AuthFlowParams,
+    authenticatedUserId?: string
+  ): AuthFlowParams {
+    if (
+      !params.projectRedirectUserId ||
+      !authenticatedUserId ||
```
laskevych: 🟡 Important (non-blocking): the `!authenticatedUserId` short-circuit fails open. If `parseToken(response.accessToken)` returns `null` in the callback (cold JWKS cache right after a deploy, key rotation, clock skew — `changeAuthCode` does no local verification, so this is the first check), a redirect explicitly bound to user A is applied to whoever completed the callback, which is exactly the bug this PR fixes, with no log.

When `params.projectRedirectUserId` is set and the authenticated user is unknown, discarding looks like the safer default; unbound redirects are unaffected either way.

packages/idp-owox-better-auth/src/owox-better-auth-idp.ts
```diff
@@ -844,15 +875,34 @@ export class OwoxBetterAuthIdp implements IdpProvider {
...
+    bindGeneratedProjectRedirectToUser = false
   ): Promise<void | e.Response> {
     const params = extractAuthFlowParams(req);
+    const createsProjectRedirect = Boolean(params.projectId && !params.appRedirectTo);
+    let projectRedirectUserId = params.projectRedirectUserId;
+    if (createsProjectRedirect && bindGeneratedProjectRedirectToUser) {
+      const refreshToken = extractRefreshToken(req);
+      if (refreshToken) {
+        try {
+          const payload = await this.tokenFacade.parseToken(refreshToken);
```
laskevych: 🟡 Important (non-blocking): `TokenService.parse` swallows every error and returns `null`, so this `try/catch` and its "Failed to bind…" warning can never fire. The real failure mode is `payload === null` (JWKS cold cache / transient fetch failure at the moment of the mismatch, or a refresh JWT missing a claim required by `IdpOwoxPayloadSchema`): `projectRedirectUserId` stays `undefined`, the generated redirect is persisted unbound, nothing is logged, and the flow silently degrades to the pre-PR behaviour.

Suggest replacing the `try/catch` with an explicit `if (!payload) { this.logger.warn(...) }`, and consider not generating the project redirect at all when binding was requested but the user could not be established (fail closed).

laskevych (approved):
Approving: the reported sign-in state-mismatch → account-switch scenario is closed, `projectRedirectUserId` never reaches platform URLs and round-trips correctly through the params cookie and `auth_states`, and the package tests pass locally.

Three non-blocking Important comments inline: the binding is applied on only one of the four `redirectToPlatform` call sites, and both sides of the flow fail open (silently) when `parseToken` returns `null`. Worth completing here or as a quick follow-up.

laskevych:
📝 Note: please drop the Fibery link from the PR description. The repository is public and internal tracker links should not appear in PR bodies or commits; the work item ID is already carried by the changeset filename (`6894-…`).

PR #1593, opened 2026-09-04
fix(bigquery): retry transient query failures instead of failing the run

prokopetsv:
# Problem

BigQuery occasionally rejects a query with a temporary fault — `backendError`, "Error encountered during execution. Retrying may solve the problem." — where the query itself is fine and the service simply could not run it at that moment. The storage sent every MERGE batch through `executeQuery`, which had no error handling at all, so one rejected batch propagated straight up and ended the whole import. A production run hit this after storing roughly 260,000 rows across about a thousand successful batches: the next batch came back with a `backendError` and the run went to FAILED, discarding two and a half hours of work. The connector's `Max Fetch Retries` and `Initial Retry Delay` settings did not help, because they live in `AbstractSource` and only cover fetching from the upstream API, never the save into BigQuery.

# Solution

`executeQuery` now retries a failed query with exponential backoff before giving up. The retry sits there rather than in the MERGE loop because both failure points are inside it, and every job the storage runs — MERGE, table creation, column addition, schema reads — shares the same entry point and the same transient failure modes.

- **A retry reuses the job it already submitted.** Once `createQueryJob` resolves, the statement is with BigQuery; resubmitting would run the MERGE a second time, paying for the scan twice and letting two mutations of the same rows overlap. Only a failure that happened before a job existed sends a new one — which is the common case, since BigQuery reports a rejected job through `createQueryJob` itself.
- **What gets retried is deliberately narrow**: a reason allowlist (`backendError`, `internalError`, `jobBackendError`, `jobInternalError`, `rateLimitExceeded`) or a transport-level failure (5xx, `ECONNRESET`, `ETIMEDOUT`, `EAI_AGAIN`). A reason outside the list does not grant a retry but does not veto one either, so a transient fault whose reason we did not enumerate is still recovered. `invalidQuery`, `accessDenied` and `quotaExceeded` arrive with a 4xx and surface immediately.
- **The existing batch-halving path is untouched.** An oversized query is `invalidQuery` with status 400, so it is not retried and still reaches `executeMergeQueryRecursively` on the first attempt to halve the batch. Pinned by a test.
- **Settings are reused, not re-declared**, read with defaults so a storage built without them still works, and sanitised: config validation only checks `typeof value === 'number'`, and `NaN` passes that, which would have left the retry loop with an exit condition that is never true.
- **Backoff was extracted** to `AsyncUtils.backoffDelay`, since `AbstractSource.calculateBackoff` already implemented the identical formula. It is clamped to `setTimeout`'s maximum, because doubling passes that ceiling around the twentieth attempt and `setTimeout` then falls back to firing after 1ms — turning a long retry chain into a tight loop against the service it is backing off from. The clamp sits far above any usable configuration, so a delay a user configured is never shortened.
- Retry messages go through `config.logMessage`, so they appear in Run History, unlike the surrounding MERGE progress lines which use `console.log` and land as untyped output.

# Changes

- Added an attempt loop with exponential backoff to `GoogleBigQueryStorage.executeQuery`, holding the submitted job across attempts and rethrowing the original error once attempts are exhausted
- Added `_isRetryableBigQueryError` and `_bigQueryErrorReasons`, classifying failures by BigQuery reason code and by HTTP or network status
- Added `retrySettingOrDefault`, which falls back unless the configured value is a finite number, so a `NaN` retry count cannot produce an unbounded loop
- Added `AsyncUtils.backoffDelay`, clamped to `setTimeout`'s maximum, and pointed `AbstractSource.calculateBackoff` at it, removing the duplicated formula
- Added `test/Core/AsyncUtils.test.js` covering doubling, jitter, the clamp, overflow to `Infinity`, and that configured delays above a minute are honoured
- Added 17 storage tests covering retried and non-retried reasons, attempt limits, deterministic backoff, config fallbacks, `NaN` and sub-one retry counts, job reuse versus resubmission, and an end-to-end MERGE batch chain that recovers from a transient failure mid-chain
- Documented the retry behaviour in the Google BigQuery storage README
- Added changeset `6887-bigquery-retry-transient-merge-errors.md`

### DIFF
--- a/packages/connectors/src/Core/AbstractSource.js
+++ b/packages/connectors/src/Core/AbstractSource.js
@@ -197,7 +197,7 @@ var AbstractSource = class AbstractSource {
      * @return {number} Delay in milliseconds
      */
     calculateBackoff(attemptNumber) {
-      return this.config.InitialRetryDelay.value * Math.pow(2, attemptNumber - 1) * (0.5 + Math.random());
+      return AsyncUtils.backoffDelay(this.config.InitialRetryDelay.value, attemptNumber);
     }
     //----------------------------------------------------------------
 
--- a/packages/connectors/src/Core/AbstractSource.js
+++ b/packages/connectors/src/Core/AbstractSource.js
@@ -24,14 +24,14 @@ var AbstractSource = class AbstractSource {
           requiredType: "number",
           default: 3,
           label: "Max Fetch Retries",
-          description: "Total attempts for a failed API request, including the first request, before the run stops",
+          description: "Total attempts for a failed request, including the first one, before the run stops. Applies both to reading from the source and to saving into the destination storage",
           attributes: [CONFIG_ATTRIBUTES.ADVANCED]
         },
         InitialRetryDelay: {
           requiredType: "number",
           default: 5000,
           label: "Initial Retry Delay (ms)",
-          description: "Delay before the first retry in milliseconds. Each retry doubles the delay",
+          description: "Delay before the first retry in milliseconds. Each retry doubles the delay. Applies both to reading from the source and to saving into the destination storage",
           attributes: [CONFIG_ATTRIBUTES.ADVANCED]
         }
       });
--- a/packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
+++ b/packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
@@ -9,6 +9,48 @@ function quoteBigQueryIdentifier(identifier) {
   return `\`${String(identifier).replace(/`/g, '``')}\``;
 }
 
+/**
+ * BigQuery error reasons that Google documents as worth retrying: the query was fine,
+ * the service could not run it right now.
+ *
+ * A reason outside this list is not by itself grounds for a retry, but it does not veto
+ * one either — the status code is still consulted, so a transient fault whose reason we
+ * did not enumerate is not abandoned. In practice the permanent reasons (invalidQuery,
+ * notFound, accessDenied, quotaExceeded) arrive with a 4xx, so they surface immediately.
+ * That includes 'invalidQuery' for an oversized query, which the MERGE loop then handles
+ * by halving the batch.
+ */
+const RETRYABLE_BIGQUERY_REASONS = [
+  'backendError',
+  'internalError',
+  'jobBackendError',
+  'jobInternalError',
+  'rateLimitExceeded',
+];
+
+/**
+ * Transport-level failures with no BigQuery reason attached: the request never reached
+ * the service, or its answer never came back.
+ */
+const RETRYABLE_BIGQUERY_ERROR_CODES = [500, 502, 503, 504, 'ECONNRESET', 'ETIMEDOUT', 'EAI_AGAIN'];
+
+const DEFAULT_MAX_QUERY_ATTEMPTS = 3;
+const DEFAULT_INITIAL_RETRY_DELAY_MS = 5000;
+
+/**
+ * Reads a retry setting that comes from user configuration. Config validation only checks
+ * `typeof value === 'number'`, and NaN passes that — which would leave the retry loop with
+ * a comparison that is never true. Anything not a real number falls back to the default.
+ *
+ * @param {*} value - The configured value
+ * @param {number} fallback - Used when the value is not a finite number
+ * @param {number} minimum - Lower bound applied to a finite value
+ * @return {number} A usable setting
+ */
+function retrySettingOrDefault(value, fallback, minimum) {
+  return Number.isFinite(value) ? Math.max(minimum, value) : fallback;
+}
+
 function normalizeBigQueryType(type) {
   const normalized = String(type || '').toUpperCase();
   const aliases = {
@@ -928,9 +970,92 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage 
         useLegacySql: false,
       };
 
-      const [job] = await bigqueryClient.createQueryJob(options);
-      const [rows] = await job.getQueryResults();
-      return rows;
+      // Both are guaranteed finite here, so the exit condition below always becomes true.
+      const maxAttempts = retrySettingOrDefault(
+        this.config.MaxFetchRetries?.value,
+        DEFAULT_MAX_QUERY_ATTEMPTS,
+        1
+      );
+      const initialRetryDelay = retrySettingOrDefault(
+        this.config.InitialRetryDelay?.value,
+        DEFAULT_INITIAL_RETRY_DELAY_MS,
+        0
+      );
+
+      // A retry reuses the job it already submitted rather than sending a new one: the
+      // statement has already been handed to BigQuery, so resubmitting would run it a
+      // second time — paying for the scan twice and letting two mutations of the same
+      // rows overlap. Only a failure that happened before a job existed submits one.
+      //
+      // A job that fails on BigQuery's side is reported by createQueryJob, which inspects
+      // the job's status, so those attempts leave `job` unset and do resubmit. What gets
+      // polled again is the case this guards: the job was accepted and only the wait for
+      // its results failed.
+      let job = null;
+
+      for (let attempt = 1; ; attempt++) {
+        try {
+          if (!job) {
+            const [createdJob] = await bigqueryClient.createQueryJob(options);
+            job = createdJob;
+          }
+
+          const [rows] = await job.getQueryResults();
+          return rows;
+        } catch (error) {
+          if (attempt >= maxAttempts || !this._isRetryableBigQueryError(error)) {
+            throw error;
+          }
+
+          const delay = AsyncUtils.backoffDelay(initialRetryDelay, attempt);
+          const reason = this._bigQueryErrorReasons(error)[0] || error.code || 'unknown error';
+
+          this.config.logMessage(
+            `BigQuery query failed (${reason}), retrying in ${Math.round(delay / 1000)}s (attempt ${attempt + 1} of ${maxAttempts})`
+          );
+
+          await AsyncUtils.delay(delay);
+        }
+      }
+    }
+
+  //---- _bigQueryErrorReasons ---------------------------------------
+    /**
+     * Reason codes BigQuery attached to a failure, e.g. ['backendError'].
+     * Empty for transport-level errors, which carry no reason.
+     *
+     * @param {Error} error - The error thrown by the BigQuery client
+     * @return {Array<string>} The reason codes, in the order BigQuery reported them
+     */
+    _bigQueryErrorReasons(error) {
+      if (!error || !Array.isArray(error.errors)) {
+        return [];
+      }
+
+      return error.errors.map(item => item && item.reason).filter(Boolean);
+    }
+
+  //---- _isRetryableBigQueryError -----------------------------------
+    /**
+     * Whether a failed query is worth running again. A failure matching neither a known
+     * reason nor a transient status code is treated as permanent, so a new failure mode
+     * surfaces to the user instead of silently consuming the configured attempts.
+     *
+     * @param {Error} error - The error thrown by the BigQuery client
+     * @return {boolean} True when the same query may succeed on another attempt
+     */
+    _isRetryableBigQueryError(error) {
+      if (!error) {
+        return false;
+      }
+
+      const reasons = this._bigQueryErrorReasons(error);
+
+      if (reasons.some(reason => RETRYABLE_BIGQUERY_REASONS.includes(reason))) {
+        return true;
+      }
+
+      return RETRYABLE_BIGQUERY_ERROR_CODES.includes(error.code);
     }
 
   //---- obfuscateSpecialCharacters ----------------------------------
--- a/packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
+++ b/packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
@@ -10,15 +10,18 @@ function quoteBigQueryIdentifier(identifier) {
 }
 
 /**
- * BigQuery error reasons that Google documents as worth retrying: the query was fine,
+ * BigQuery error reasons that Google documents as worth retrying: the statement was fine,
  * the service could not run it right now.
  *
- * A reason outside this list is not by itself grounds for a retry, but it does not veto
- * one either — the status code is still consulted, so a transient fault whose reason we
- * did not enumerate is not abandoned. In practice the permanent reasons (invalidQuery,
- * notFound, accessDenied, quotaExceeded) arrive with a 4xx, so they surface immediately.
- * That includes 'invalidQuery' for an oversized query, which the MERGE loop then handles
- * by halving the batch.
+ * These are the reasons the client library never retries on its own. Its `shouldRetryRequest`
+ * acts on the HTTP status of a single request, so it cannot see a job that was accepted and
+ * then failed while running — which is the case this list exists for.
+ *
+ * A reason outside this list is not by itself grounds for a retry, but it does not veto one
+ * either: a reason-less transport failure is still classified by its code. The permanent
+ * reasons (invalidQuery, notFound, accessDenied, quotaExceeded) arrive with a 4xx, so they
+ * surface immediately. That includes 'invalidQuery' for an oversized query, which the MERGE
+ * loop then handles by halving the batch.
  */
 const RETRYABLE_BIGQUERY_REASONS = [
   'backendError',
@@ -26,16 +29,28 @@ const RETRYABLE_BIGQUERY_REASONS = [
   'jobBackendError',
   'jobInternalError',
   'rateLimitExceeded',
+  'jobRateLimitExceeded',
 ];
 
 /**
- * Transport-level failures with no BigQuery reason attached: the request never reached
- * the service, or its answer never came back.
+ * Transport-level failures with no BigQuery reason attached: the answer to a request never
+ * came back, so whatever it was asking about may still be fine.
+ *
+ * HTTP statuses are deliberately absent. @google-cloud/common already retries 408/429/5xx
+ * through `shouldRetryRequest`, so listing them here would only stack a second ladder on
+ * top of the client's own and report the total as one attempt.
  */
-const RETRYABLE_BIGQUERY_ERROR_CODES = [500, 502, 503, 504, 'ECONNRESET', 'ETIMEDOUT', 'EAI_AGAIN'];
+const RETRYABLE_BIGQUERY_ERROR_CODES = ['ECONNRESET', 'ETIMEDOUT', 'EAI_AGAIN'];
 
 const DEFAULT_MAX_QUERY_ATTEMPTS = 3;
 const DEFAULT_INITIAL_RETRY_DELAY_MS = 5000;
+
+/**
+ * Ceiling for a single backoff wait. `MaxFetchRetries` has no maximum, and a run holds one
+ * of the project's concurrent slots while it sleeps, so uncapped doubling lets one BigQuery
+ * outage park a run for over an hour before it reports the failure.
+ */
+const MAX_RETRY_DELAY_MS = 60000;
 
 /**
  * Reads a retry setting that comes from user configuration. Config validation only checks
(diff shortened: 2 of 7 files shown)


packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -928,9 +970,92 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+      //
+      // A job that fails on BigQuery's side is reported by createQueryJob, which inspects
+      // the job's status, so those attempts leave `job` unset and do resubmit. What gets
+      // polled again is the case this guards: the job was accepted and only the wait for
+      // its results failed.
+      let job = null;
+
+      for (let attempt = 1; ; attempt++) {
+        try {
+          if (!job) {
```
laskevych, line 1014: 🔴 **Blocking:** The retry never resubmits the MERGE for the production case this PR is about.

In `@google-cloud/bigquery` 8.3.1, `createQueryJob` only inspects `status.errors` on the *insert* response; it does not wait for the job to finish. A MERGE that BigQuery accepts and then fails during execution with `backendError` ("Error encountered during execution. Retrying may solve the problem.") comes back as RUNNING from the insert, so `createQueryJob` resolves and `job` is set. The failure then surfaces from `job.getQueryResults()`. At that point the loop calls `getQueryResults()` on the same job again, and a job that is DONE with `errorResult` returns the identical error on every poll. Net effect: the run sleeps 5s + 10s, logs "retrying" twice, and fails with the same error as before the PR.

Suggested fix: when the error from `getQueryResults` carries a BigQuery reason (`errors[].reason` non-empty), the job itself failed, so set `job = null` and resubmit. Keep re-polling only for reason-less transport errors (`ECONNRESET` etc.), where the job may still be running. Resubmitting a failed DML job is safe: a failed MERGE committed nothing, so the "two overlapping mutations" concern does not apply there.

The tests pass because every transient failure is thrown from `createQueryJob`; the only `getQueryResults` failure covered is `ECONNRESET`. Please add a case where `createQueryJob` resolves and `getQueryResults` rejects once with `{ code: 500, errors: [{ reason: 'backendError' }] }`, asserting that a second job is created. The changeset line "saves that batch again" becomes accurate once this is fixed.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -9,6 +9,48 @@ function quoteBigQueryIdentifier(identifier) {
...
+  'jobBackendError',
+  'jobInternalError',
+  'rateLimitExceeded',
+];
+
+/**
+ * Transport-level failures with no BigQuery reason attached: the request never reached
+ * the service, or its answer never came back.
+ */
+const RETRYABLE_BIGQUERY_ERROR_CODES = [500, 502, 503, 504, 'ECONNRESET', 'ETIMEDOUT', 'EAI_AGAIN'];
```
laskevych: 🟡 **Important (non-blocking):** The BigQuery client already retries these. `@google-cloud/common` defaults to `autoRetry: true`, `maxRetries: 3`, `noResponseRetries: 3`, and its `shouldRetryRequest` covers 408/429/500/502/503/504, `rateLimitExceeded`, `userRateLimitExceeded`, `EAI_AGAIN`, plus socket errors via `retry-request`. So every transport-level entry here (and `rateLimitExceeded`) only reaches `executeQuery` after four client requests, and the outer loop then repeats the whole ladder: a sustained 503 costs 12 requests reported as "attempt 3 of 3".

If you want transport retries tuned, `retryOptions` on the `BigQuery` constructor is the single place for it. The hand-rolled loop is genuinely needed only for job-level reasons the client never retries (`backendError`, `internalError`, `jobBackendError`, `jobInternalError`), which is also the path from the blocking comment.

packages/connectors/src/Core/AbstractSource.js
```diff
@@ -197,7 +197,7 @@ var AbstractSource = class AbstractSource {
      * @return {number} Delay in milliseconds
      */
     calculateBackoff(attemptNumber) {
-      return this.config.InitialRetryDelay.value * Math.pow(2, attemptNumber - 1) * (0.5 + Math.random());
+      return AsyncUtils.backoffDelay(this.config.InitialRetryDelay.value, attemptNumber);
```
laskevych, line 200: 💡 **Suggestion:** `Max Fetch Retries` and `Initial Retry Delay` (declared above, lines 23–36) now also govern BigQuery saves, but their label and description shown under Advanced settings still say "Total attempts for a failed API request". A user who sets `Max Fetch Retries = 1` to fail fast on a flaky source API silently disables BigQuery retries too (the test "makes a single attempt when retries are configured off" pins that). A one-line update to the description would make the coupling visible where users actually see it; the README only documents it on the storage page.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -928,9 +970,92 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+          }
+
+          const [rows] = await job.getQueryResults();
+          return rows;
+        } catch (error) {
+          if (attempt >= maxAttempts || !this._isRetryableBigQueryError(error)) {
+            throw error;
+          }
+
+          const delay = AsyncUtils.backoffDelay(initialRetryDelay, attempt);
```
laskevych: 🔵 **Question:** There is no cap on cumulative sleep. `MaxFetchRetries` has no maximum in the spec and the Node runner has no run timeout, so with `MaxFetchRetries = 10` and the default 5s delay a real BigQuery outage makes each batch sleep about 85 minutes before failing, while the run holds one of the project's concurrent run slots. Before this PR the run failed within seconds and the next schedule picked up.

Is that acceptable, or should the storage bound the delay (e.g. a max per-attempt delay, as the client's `retryOptions.maxRetryDelay` does)? The 2³¹−1 clamp only engages around attempt 20, so it does not help here.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -928,9 +970,92 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+      // A job that fails on BigQuery's side is reported by createQueryJob, which inspects
+      // the job's status, so those attempts leave `job` unset and do resubmit. What gets
+      // polled again is the case this guards: the job was accepted and only the wait for
+      // its results failed.
+      let job = null;
+
+      for (let attempt = 1; ; attempt++) {
+        try {
+          if (!job) {
+            const [createdJob] = await bigqueryClient.createQueryJob(options);
```
laskevych, line 1015: 🔵 **Question:** Resubmission mints a new job id. On a lost `createQueryJob` response, the outer loop calls `createQueryJob` again, which generates a fresh `randomUUID`; the client's 409 dedup only applies to its own internal retries. If the first insert actually landed, the statement runs twice. MERGE is idempotent (only the scan is paid twice), but the snapshot publish script (`BEGIN TRANSACTION; TRUNCATE; INSERT; COMMIT`) is not, and the second copy fails with a concurrent-update 400 that is not retried.

Narrow window, since the client's three no-response retries must fail first. Passing an explicit `jobId` per `executeQuery` call and tolerating 409 would close it. Fine as a follow-up if you consider the window too small to matter.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -9,6 +9,48 @@ function quoteBigQueryIdentifier(identifier) {
...
+ * the service could not run it right now.
+ *
+ * A reason outside this list is not by itself grounds for a retry, but it does not veto
+ * one either — the status code is still consulted, so a transient fault whose reason we
+ * did not enumerate is not abandoned. In practice the permanent reasons (invalidQuery,
+ * notFound, accessDenied, quotaExceeded) arrive with a 4xx, so they surface immediately.
+ * That includes 'invalidQuery' for an oversized query, which the MERGE loop then handles
+ * by halving the batch.
+ */
+const RETRYABLE_BIGQUERY_REASONS = [
```
laskevych, line 26: 💡 **Suggestion:** Two documented transient faults arrive with 4xx and are therefore never retried: `jobRateLimitExceeded` (400, concurrent job-rate limit) and the DML conflict "Could not serialize access to table … due to concurrent update" (reason `invalid`, 400), which the cleanup DELETE or a second importer into the same table can trigger mid-run. Both are retry-with-backoff conditions in Google's error table. Worth adding if you want the allowlist to cover the MERGE-specific cases.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -9,6 +9,48 @@ function quoteBigQueryIdentifier(identifier) {
...
+ * `typeof value === 'number'`, and NaN passes that — which would leave the retry loop with
+ * a comparison that is never true. Anything not a real number falls back to the default.
+ *
+ * @param {*} value - The configured value
+ * @param {number} fallback - Used when the value is not a finite number
+ * @param {number} minimum - Lower bound applied to a finite value
+ * @return {number} A usable setting
+ */
+function retrySettingOrDefault(value, fallback, minimum) {
+  return Number.isFinite(value) ? Math.max(minimum, value) : fallback;
```
laskevych, line 66: 🔹 **Nit:** The `NaN` branch and the local 3/5000 fallbacks guard states that cannot occur at runtime. Config reaches the runner through `JSON.stringify`/`JSON.parse`, and `JSON.stringify(NaN)` is `null`, which `validate()` replaces with the default. The storage always shares the config the source already merged, so `MaxFetchRetries?.value` is only undefined in unit tests.

If `NaN` is worth guarding, the identical `attempt <= NaN` shape exists in `AbstractSource.urlFetchWithRetry`, so a `Number.isFinite` check in `AbstractConfig` validation would cover both once. Not asking for a change here, just flagging the duplicated defaults that can drift.

laskevych (changes requested):
Thanks for the thorough write-up and the test coverage. One blocking issue: for the exact production case described here (a MERGE that BigQuery accepts and then fails during execution), the retry re-polls the already-failed job instead of resubmitting the statement, so the run still fails. Details inline, plus six non-blocking notes. Both touched test files pass locally; the red CI checks (Audit, web unit tests, Browser E2E) fail identically on `main`, so they are unrelated to this PR.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Fixed in 66d1e373f — you were right, and the failure mode was exactly as you described.

A retry now drops the job and resubmits whenever the error carries a BigQuery reason, since that means the job reached a terminal state. Re-polling is kept only for reason-less transport faults, where the job may still be running:

```js
const reasons = this._bigQueryErrorReasons(error);

// BigQuery answered, so the job is finished and failed. Drop it and resubmit.
if (reasons.length > 0) {
  job = null;
}
```

I also rewrote the comment block above the loop, since it asserted the opposite and would have misled the next reader.

Added the test you asked for, plus one for the exhaustion path:

- `resubmits the statement when the job failed after BigQuery accepted it` — `createQueryJob` resolves, `getQueryResults` rejects once with `{ code: 500, errors: [{ reason: 'backendError' }] }`, asserts `jobsCreated === 2`.
- `gives up on a job that keeps failing during execution, without polling it twice` — asserts three separate jobs and exactly one poll each, so no attempt is ever spent re-asking a dead job.

Both fail against the previous commit and pass now. The changeset line is accurate as of this push, and I extended it to name both cases explicitly.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Agreed on the HTTP statuses, changed. `RETRYABLE_BIGQUERY_ERROR_CODES` is now `['ECONNRESET', 'ETIMEDOUT', 'EAI_AGAIN']` — I confirmed `shouldRetryRequest` in `@google-cloud/common` covers 408/429/500/502/503/504, so listing them was stacking a second ladder and reporting 12 requests as "attempt 3 of 3".

I kept the socket-level codes rather than moving everything to `retryOptions`, because after the blocking fix they do different jobs. A reason-less fault is now the only case that re-polls an existing job, and that is genuinely useful here: a MERGE poll is a long-lived request (the production job ran 17 minutes before it failed), and `retry-request`'s `noResponseRetries: 2` only covers the immediate request. Tuning `retryOptions` would not give us that, since the client has no notion of "the job may still be running, ask again".

So the split is now: the client handles transport-level status retries, and this loop handles the two things it structurally cannot see — a job that failed after acceptance, and a poll worth repeating.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Added `jobRateLimitExceeded`, with a test (`retries a job-level rate limit, which the client library never retries itself`). It is a good fit: 400 so `shouldRetryRequest` ignores it, documented as retry-with-backoff, and directly relevant to a run issuing thousands of MERGEs against one table.

I left the concurrent-update case out. Its reason is `invalid`, which is also what a genuinely malformed statement returns, so matching it means matching on the message text `Could not serialize access to table`. That is fragile — it is not a stable API surface, and it is localised nowhere but still free to change. Given the failure resolves by retrying the whole batch anyway, I would rather handle it deliberately in a follow-up (with a test against a captured real error body) than add a substring match here. Happy to be overruled if you think the risk is the other way round.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Not acceptable, you are right — bounded it. A single wait is now capped at 60s:

```js
const delay = Math.min(
  AsyncUtils.backoffDelay(initialRetryDelay, attempt),
  MAX_RETRY_DELAY_MS
);
```

With your `MaxFetchRetries = 10` example and the default 5s that takes the worst case from about 85 minutes to about 6, while leaving the default configuration (3 attempts, 5s + 10s) untouched. Pinned by `caps a single backoff wait so an outage cannot park the run for an hour`, which asserts the exact sequence `[1000, 2000, 4000, 8000, 16000, 32000, 60000]`.

I capped per-attempt rather than cumulative, matching `retryOptions.maxRetryDelay` as you suggested. A cumulative budget would be the stricter guard, but it makes the attempt count no longer mean what the setting says, and the per-attempt cap already removes the pathological case.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: This one the blocking fix largely closes, which I had not expected.

The double-execution window needs a resubmission of a statement that may already be running. After the fix, a resubmission happens only when the error carried a BigQuery reason — which means BigQuery answered and the job is terminal, so there is nothing live to duplicate. The lost-`createQueryJob`-response case you describe carries no reason (it is a transport fault), so it now takes the re-poll path and never mints a second job id.

The residual window is a transport fault whose error nonetheless carries a reason, which I could not construct: `parseHttpRespBody` only attaches `errors` when the service returned a parseable error body, and that implies it answered.

So I have not added explicit `jobId` + 409 handling. If you can see a path I have missed I will file it as a follow-up rather than widen this PR — the snapshot publish script is the part that would actually be harmed, and it deserves its own test.

packages/connectors/src/Core/AbstractSource.js
prokopetsv: Good catch — updated both descriptions to name the coupling where users actually see it:

- **Max Fetch Retries**: "Total attempts for a failed request, including the first one, before the run stops. Applies both to reading from the source and to saving into the destination storage"
- **Initial Retry Delay (ms)**: same trailing sentence.

That makes the `Max Fetch Retries = 1` trade-off visible in Advanced settings rather than only on the storage README page.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Fair, and the analysis is right — `JSON.stringify(NaN)` is `null` and `validate()` substitutes the default, so `NaN` cannot reach the storage through the runner.

I have left the guard in place for now, on the grounds that `retrySettingOrDefault` is what makes "the exit condition always becomes true" a local property of this function rather than something that depends on a serialisation step three layers away. It is cheap and it is what the unit tests construct against.

But your point about drift is the real one, and I agree with the direction: `Number.isFinite` in `AbstractConfig.validate()` would cover this and the identical `attempt <= NaN` shape in `AbstractSource.urlFetchWithRetry` in one place, and then these local fallbacks could go. That touches config validation for every connector, so I would rather not fold it into this PR. Filing it as a follow-up.

prokopetsv:
Pushed 66d1e373f addressing all seven comments. Summary:

| Comment | Action |
|---|---|
| 🔴 Retry never resubmits the MERGE | **Fixed.** Drop the job and resubmit when the error carries a BigQuery reason; keep re-polling for reason-less transport faults. Two new tests, both red on the previous commit. |
| 🟡 Client already retries these | **Changed.** Dropped 408/429/5xx from the transport allowlist; kept socket codes, which now serve the distinct re-poll path. |
| 💡 `jobRateLimitExceeded` / concurrent-update | **Half.** Added `jobRateLimitExceeded` with a test. Left the concurrent-update case out — its reason is `invalid`, so matching it means matching message text. Follow-up. |
| 🔵 No cap on cumulative sleep | **Bounded.** Single wait capped at 60s. Worst case for `MaxFetchRetries = 10` goes from ~85 min to ~6. |
| 🔵 Resubmission mints a new job id | **Largely closed by the blocking fix** — resubmission now happens only after BigQuery reported a terminal job, so no live statement is duplicated. Details inline. |
| 💡 Setting descriptions | **Updated** both to say they govern storage saves as well as source reads. |
| 🔹 `NaN` guard | **Kept**, with agreement on the direction. `Number.isFinite` in `AbstractConfig.validate()` as a follow-up, since it touches every connector. |

The correction that matters beyond this PR: `createQueryJob` resolving does **not** mean the statement succeeded, and the old comment block said it did. That is rewritten.

`packages/connectors`: 282 tests pass, eslint and prettier clean.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
```diff
@@ -928,9 +985,106 @@ var GoogleBigQueryStorage = class GoogleBigQueryStorage extends AbstractStorage
...
+          return rows;
+        } catch (error) {
+          if (attempt >= maxAttempts || !this._isRetryableBigQueryError(error)) {
+            throw error;
+          }
+
+          const reasons = this._bigQueryErrorReasons(error);
+
+          // BigQuery answered, so the job is finished and failed. Drop it and resubmit.
+          if (reasons.length > 0) {
```
laskevych: 💡 **Suggestion (non-blocking, follow-up is fine):** On the residual window you asked about: "the error carries a reason" does not strictly imply "the job is terminal". The poll request itself can fail with a reasoned error while the job is still running: a transient 5xx on `GET /queries/{id}` carries `errors[].reason: 'backendError'` in its body, and a per-user API `rateLimitExceeded` (403) on the poll is about the request, not the job. The client retries both three times first, so it takes four consecutive poll failures, but when it happens this branch drops a live job and resubmits the MERGE alongside it.

The outcome is bounded (MERGE is idempotent, and BigQuery serialises DML on one table, so the worst case is the old pre-PR failure for that batch), which is why this is not blocking. The precise discriminator would be `job.getMetadata()` and checking `status.state === 'DONE' && status.errorResult` before clearing `job`; that is one extra request per retry and removes the assumption entirely.

laskevych (approved):
Re-reviewed 66d1e373f. The blocker is fixed: a job-level failure now drops the job and resubmits, and the two new tests (`resubmits the statement when the job failed after BigQuery accepted it`, `gives up on a job that keeps failing during execution, without polling it twice`) pin exactly the path that was broken. The 60s per-attempt cap, the trimmed transport list, `jobRateLimitExceeded`, and the setting descriptions all look right, and the changeset text now matches the behaviour.

Regression check: against `main` the PR removes only six lines. The two setting descriptions, the backoff formula (moved verbatim into `AsyncUtils.backoffDelay`, clamped only past the `setTimeout` ceiling), and the three-line body of `executeQuery`, which is reproduced unchanged as the first attempt. The success path is byte-for-byte the old call sequence, a non-retryable error is rethrown as the original object on attempt one with no delay, and the `query is too large` halving path still receives it. Full `packages/connectors` suite passes locally (282 tests); `Audit all` fails identically on `main`.

One non-blocking note inline on the resubmit discriminator, fine as a follow-up. Approving.

packages/connectors/src/Storages/GoogleBigQuery/GoogleBigQueryStorage.js
prokopetsv: Valid, and fixed in 14dd1f467 rather than deferred — the assumption was load-bearing enough that I would rather not leave it in.

You are right that a reasoned error does not imply a terminal job. I have dropped the reason-based inference entirely and now ask the job, which is the same check `poll_` makes in the client library:

```js
// Nothing left to poll: the statement has to be sent again.
if (job && (await this._hasJobFinishedWithError(job))) {
  job = null;
}
```

```js
async _hasJobFinishedWithError(job) {
  try {
    const [metadata] = await job.getMetadata();

    return metadata?.status?.state === 'DONE' && Boolean(metadata.status.errorResult);
  } catch {
    return false;
  }
}
```

Two notes on the shape of it:

- **The check is not gated on `reasons.length > 0`.** Once it is the job being asked rather than the error being read, the reason adds nothing — and the reason-less transport case has the same ambiguity in the other direction (a dropped socket does not prove the job survived). Now a single question decides it, whatever shape the fault arrived in. `reasons` survives only to name the failure in the run log.
- **An unreadable status counts as not failed.** The two errors are not symmetric: re-polling a finished job costs one attempt, while resubmitting beside a running one costs a duplicate MERGE. So the fallback favours the cheaper mistake, and the worst case degrades to the pre-PR behaviour for that batch.

Cost is one extra request per retry that has a job, which only happens on a path that is already sleeping for seconds.

Two tests, both red against 66d1e373f:

- `keeps polling when the poll failed but the job behind it is still running` — `getQueryResults` rejects with `backendError`, `getMetadata` reports RUNNING, asserts one job created.
- `keeps the job when its status cannot be read, rather than running the statement twice` — `getMetadata` throws, asserts one job created.

I also gave the existing job mocks a realistic `getMetadata`, so the terminal-versus-running distinction is now explicit in every test that exercises it. 284 tests pass.

PR #1594, opened 2026-09-07
fix(canvas): prevent full-page errors in Model Canvas for larger data models

artemfenoga:
image

**Problem.** Opening the Model Canvas — or the Joinable Data Marts diagram — for a
storage with many Data Marts and relationships could replace the whole page with a
"Something went wrong" error screen right after the loading skeleton. Storages with
only a few Data Marts were unaffected, so the failure was intermittent and hard to
reproduce.

**Cause.** The automatic layout step throws on some otherwise-valid relationship
graphs — typically when a model has several relationships between the same pair of
Data Marts, a Data Mart joined to itself, or circular joins. That error was not
handled, so it propagated to the route error boundary and took down the entire page
instead of just the canvas.

**Fix.** The canvas now recovers instead of crashing. When the layout step fails it
retries with a simplified view of the relationships, which resolves the vast
majority of these cases and still produces a proper layout. If that also fails, it
falls back to a plain grid you can rearrange by dragging. Node positions you set are
still saved per storage.

### DIFF
--- a/.changeset/6835-fix-canvas.md
+++ b/.changeset/6835-fix-canvas.md
@@ -1,0 +1,23 @@
+---
+'owox': minor
+---
+
+# Fix the Model Canvas failing with a full-page error on larger data models
+
+**Problem.** Opening the Model Canvas — or the Joinable Data Marts diagram — for a
+storage with many Data Marts and relationships could replace the whole page with a
+"Something went wrong" error screen right after the loading skeleton. Storages with
+only a few Data Marts were unaffected, so the failure was intermittent and hard to
+reproduce.
+
+**Cause.** The automatic layout step throws on some otherwise-valid relationship
+graphs — typically when a model has several relationships between the same pair of
+Data Marts, a Data Mart joined to itself, or circular joins. That error was not
+handled, so it propagated to the route error boundary and took down the entire page
+instead of just the canvas.
+
+**Fix.** The canvas now recovers instead of crashing. When the layout step fails it
+retries with a simplified view of the relationships, which resolves the vast
+majority of these cases and still produces a proper layout. If that also fails, it
+falls back to a plain grid you can rearrange by dragging. Node positions you set are
+still saved per storage.
--- a/apps/web/src/features/data-marts/shared/canvas/dagre-layout.test.ts
+++ b/apps/web/src/features/data-marts/shared/canvas/dagre-layout.test.ts
@@ -127,6 +127,40 @@ describe.each([
   });
 });
 
+it.each(['horizontal', 'vertical'] as const)(
+  'recovers a layout instead of throwing when Dagre cannot rank the raw graph (%s)',
+  direction => {
+    // Reduced from a real Models canvas that crashed the page: dense parallel
+    // edges (n3 -> n6 three times) converging with n2/n7 make Dagre place two
+    // connected nodes on the same point, and `dagre.layout` throws
+    // "Not possible to find intersection inside of the rectangle". The retry on
+    // the collapsed topology clears it.
+    const ids = ['n1', 'n2', 'n3', 'n6', 'n7'];
+    const nodes: DagreLayoutNode[] = ids.map(id => ({ id, width: NODE_W, height: NODE_H }));
+    const edges: DagreLayoutEdge[] = [
+      { id: 'e4', sourceId: 'n3', targetId: 'n6' },
+      { id: 'e10', sourceId: 'n7', targetId: 'n3' },
+      { id: 'e11', sourceId: 'n3', targetId: 'n6' },
+      { id: 'e13', sourceId: 'n1', targetId: 'n7' },
+      { id: 'e14', sourceId: 'n2', targetId: 'n6' },
+      { id: 'e15', sourceId: 'n3', targetId: 'n6' },
+      { id: 'e17', sourceId: 'n2', targetId: 'n7' },
+    ];
+
+    const result = runDagreLayout(nodes, edges, direction);
+
+    expect(result.positions.size).toBe(nodes.length);
+    for (const node of nodes) {
+      expect(result.positions.has(node.id)).toBe(true);
+    }
+    expect(uniquePositions(result.positions)).toBe(true);
+    for (const position of result.positions.values()) {
+      expect(Number.isFinite(position.x)).toBe(true);
+      expect(Number.isFinite(position.y)).toBe(true);
+    }
+  }
+);
+
 it('keeps a deep graph renderable when Dagre exhausts the call stack', () => {
   const nodes = Array.from({ length: 2_000 }, (_, index) => ({
     id: `node-${index}`,
--- a/apps/web/src/features/data-marts/shared/canvas/dagre-layout.ts
+++ b/apps/web/src/features/data-marts/shared/canvas/dagre-layout.ts
@@ -70,7 +70,26 @@ function buildFallbackPositions(
   return positions;
 }
 
-export function runDagreLayout(
+/**
+ * Only the edge topology drives dagre's ranking: parallel edges between the same
+ * ordered pair and self-loops add nothing to it, but they make dagre route extra
+ * dummy nodes that can land on a real node's centre and throw "Not possible to
+ * find intersection inside of the rectangle". Collapsing them is a safe retry.
+ */
+function simplifyEdgesForLayout(edges: DagreLayoutEdge[]): DagreLayoutEdge[] {
+  const seenPairs = new Set<string>();
+  const simplified: DagreLayoutEdge[] = [];
+  for (const edge of edges) {
+    if (edge.sourceId === edge.targetId) continue;
+    const pairKey = `${edge.sourceId} -> ${edge.targetId}`;
+    if (seenPairs.has(pairKey)) continue;
+    seenPairs.add(pairKey);
+    simplified.push(edge);
+  }
+  return simplified;
+}
+
+function layoutWithDagre(
   nodes: DagreLayoutNode[],
   edges: DagreLayoutEdge[],
   direction: CanvasDirection
@@ -78,8 +97,6 @@ export function runDagreLayout(
   const positions = new Map<string, PathPoint>();
   const routes = new Map<string, PathPoint[]>();
   const labelPositions = new Map<string, PathPoint>();
-
-  if (nodes.length === 0) return { positions, routes, labelPositions };
 
   const g = new dagre.graphlib.Graph<GraphLabel, NodeLabel, EdgeLabel>({ multigraph: true });
   g.setGraph({ rankdir: direction === 'horizontal' ? 'LR' : 'TB' });
@@ -96,16 +113,7 @@ export function runDagreLayout(
     g.setEdge(edge.sourceId, edge.targetId, edgeLabel, edge.id);
   }
 
-  try {
-    dagre.layout(g);
-  } catch (error) {
-    if (!(error instanceof RangeError)) throw error;
-    return {
-      positions: buildFallbackPositions(nodes, direction),
-      routes,
-      labelPositions,
-    };
-  }
+  dagre.layout(g);
 
   for (const nodeId of g.nodes()) {
     const n = g.node(nodeId);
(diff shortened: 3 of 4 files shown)


apps/web/src/features/data-marts/shared/canvas/dagre-layout.ts
```diff
@@ -126,3 +137,38 @@ export function runDagreLayout(
...
+  edges: DagreLayoutEdge[],
+  direction: CanvasDirection
+): DagreLayoutResult {
+  if (nodes.length === 0) {
+    return { positions: new Map(), routes: new Map(), labelPositions: new Map() };
+  }
+
+  try {
+    return layoutWithDagre(nodes, edges, direction);
+  } catch {
```
laskevych, line 152: 🟡 Important (non-blocking): this now swallows *every* exception with no signal, where the previous code deliberately rethrew anything that was not a `RangeError`.

Two concrete costs:
1. Our own loop at line 119 can throw a `TypeError`: graphlib auto-creates an endpoint that is not in `nodes`, `dagre.layout` succeeds, `g.node(id)` returns `undefined`, and `n.x` throws. Today's callers pre-filter edges so it is latent, but this is a shared helper, and the retry hits the same error, so the whole canvas would silently become a grid.
2. There is no `console.warn` on either fallback tier. A future dagre regression or a refactor bug would ship as "the canvas is a grid sometimes" with nothing in the console to bisect from. The file next door already does `console.error('Canvas export failed:', caught)` for the export path.

Suggestion: `if (!n) continue;` (or iterate `nodes` instead of `g.nodes()`), and a one-line `console.warn('Auto-layout failed, using fallback', error)` before the retry/grid.

apps/web/src/features/data-marts/shared/canvas/dagre-layout.ts
```diff
@@ -126,3 +137,38 @@ export function runDagreLayout(
...
+    // Dagre throws on graphs it cannot rank cleanly: a RangeError when a deep
+    // graph exhausts its recursion, or a plain Error ("Not possible to find
+    // intersection inside of the rectangle") when dense parallel edges or
+    // cycles collapse two connected nodes onto one point. Retry once on the
+    // ranking-only topology, which clears most of these, before falling back to
+    // a deterministic grid: anything is better than the exception taking down
+    // the whole page.
+    const simplified = simplifyEdgesForLayout(edges);
```
laskevych, line 160: 💡 Suggestion: the old `RangeError` short-circuit went straight to the grid. Now a depth-caused `RangeError` on a graph that also has one parallel edge or self-loop runs the simplified retry, which has the same depth and overflows again, synchronously on the main thread. Measured on the existing 2,000-node chain plus one duplicated edge: ~178 ms to grid on `main` vs ~355 ms here.

`catch (error) { if (!(error instanceof RangeError)) { ...retry } ...grid }` keeps the old behaviour for free. Not blocking; this only bites on very deep chains.

apps/web/src/features/data-marts/shared/canvas/dagre-layout.ts
```diff
@@ -70,7 +70,26 @@ function buildFallbackPositions(
   return positions;
 }
 
-export function runDagreLayout(
+/**
+ * Only the edge topology drives dagre's ranking: parallel edges between the same
+ * ordered pair and self-loops add nothing to it, but they make dagre route extra
+ * dummy nodes that can land on a real node's centre and throw "Not possible to
```
laskevych, line 76: 🔵 Question / docs: I could not reproduce the mechanism described here ("dummy nodes that land on a real node's centre"), and I think the comment will mislead the next maintainer.

What actually happens in dagre 3.1.1 (verified on this fixture): n3 and n6 are placed at distinct points. The failing edge has a dummy with `points = [{x: undefined, y: undefined}]`. In `sortSubgraph`, same-barycenter dummies of the same `(v, w)` are paired via `reversedPairs[name] = entry`, an object keyed by node name, so with **three or more** parallel edges the third dummy is overwritten, dropped from the order, never positioned, and `intersectRect` throws on its undefined point. The trigger is "3+ edges between the same unordered pair, join labels hidden" (`a→b ×2 + b→a ×1` throws as well). Cycles and self-loops alone lay out fine, and node sizes / `nodesep` / `ranksep` do not affect it. Labelled edges take a different code path, which is why this only shows up with join labels off (the default).

Suggest rewording to the real trigger so nobody later tries to "fix" it via spacing or node sizes. Same for the `Cause` sentence in the changeset (see there).

apps/web/src/features/data-marts/shared/canvas/dagre-layout.ts
```diff
@@ -126,3 +137,38 @@ export function runDagreLayout(
 
   return { positions, routes, labelPositions };
 }
+
+export function runDagreLayout(
```
laskevych, line 141: 💡 Suggestion (design, not asking for it in this PR): both callers destructure only `positions`, and `ModelCanvas` already draws parallel edges itself via `computeParallelEdgeOffsets`, so `routes`/`labelPositions` have no consumer. That means the parallel-edge collapse could be done deterministically *before* the first layout (keeping `max(width/height)` of the group as the surviving edge's label), which makes layout single-pass and removes the trigger instead of catching it. Two near-identical models would then lay out the same way rather than one going through the retried path.

If you prefer to keep the catch-and-retry shape, `dagre.layout(g, { disableOptimalOrderHeuristic: true })` is an alternative retry that keeps every edge and label; it also clears this fixture and all 167 fuzz cases. Either way, fine as a follow-up.

apps/web/src/features/data-marts/shared/canvas/dagre-layout.ts
```diff
@@ -114,6 +122,9 @@ export function runDagreLayout(
   }
 
   for (const edge of edges) {
+    // A collapsed retry graph (see `simplifyEdgesForLayout`) drops some edges,
+    // so the lookup can miss even though the type says otherwise.
+    if (!g.hasEdge(edge.sourceId, edge.targetId, edge.id)) continue;
```
laskevych, line 127: 🧹 Nit: this guard can never be false. `layoutWithDagre` iterates the same `edges` array it just inserted into `g` (on the retry it receives the already-simplified array), and `dagre.layout` never removes edges. The comment describes a path that does not exist, so I would drop both lines rather than leave a reader hunting for the missing-edge case.

apps/web/src/features/data-marts/shared/canvas/dagre-layout.ts
```diff
@@ -70,7 +70,26 @@ function buildFallbackPositions(
...
 }
 
-export function runDagreLayout(
+/**
+ * Only the edge topology drives dagre's ranking: parallel edges between the same
+ * ordered pair and self-loops add nothing to it, but they make dagre route extra
+ * dummy nodes that can land on a real node's centre and throw "Not possible to
+ * find intersection inside of the rectangle". Collapsing them is a safe retry.
+ */
+function simplifyEdgesForLayout(edges: DagreLayoutEdge[]): DagreLayoutEdge[] {
```
laskevych, line 79: 💡 Suggestion (cosmetic): the surviving edge keeps only the *first* edge's label dimensions, which may be `undefined` if that relationship has no join label while a later one has a wide one. With join labels shown, dagre then reserves no room while `ModelCanvasFlowEdge` still draws all N labels bowed ±56px, so the outer labels can overlap neighbouring cards. Taking the max width/height across the collapsed group avoids it. Only matters when labels are on, which is not the default.

apps/web/src/features/data-marts/shared/canvas/dagre-layout.test.ts
```diff
@@ -127,6 +127,40 @@ describe.each([
...
+      { id: 'e17', sourceId: 'n2', targetId: 'n7' },
+    ];
+
+    const result = runDagreLayout(nodes, edges, direction);
+
+    expect(result.positions.size).toBe(nodes.length);
+    for (const node of nodes) {
+      expect(result.positions.has(node.id)).toBe(true);
+    }
+    expect(uniquePositions(result.positions)).toBe(true);
```
laskevych, line 156: 🟡 Important (non-blocking): these assertions (size, `has`, unique, finite) are all satisfied by `buildFallbackPositions` too, so the test cannot tell "retry succeeded" from "fell through to the grid". I confirmed that deleting the retry block still leaves this test green.

A single discriminating assertion fixes it, e.g. `expect(result.routes.size).toBeGreaterThan(0)` (the grid returns empty maps), or the rank order for `horizontal`: `x(n1) < x(n7) < x(n3) < x(n6)`. Worth also adding the opposite-direction shape (`n3→n6 ×2 + n6→n3 ×1`), which throws on the raw graph as well.

.changeset/6835-fix-canvas.md
```diff
@@ -0,0 +1,23 @@
...
+**Cause.** The automatic layout step throws on some otherwise-valid relationship
+graphs — typically when a model has several relationships between the same pair of
+Data Marts, a Data Mart joined to itself, or circular joins. That error was not
+handled, so it propagated to the route error boundary and took down the entire page
+instead of just the canvas.
+
+**Fix.** The canvas now recovers instead of crashing. When the layout step fails it
+retries with a simplified view of the relationships, which resolves the vast
+majority of these cases and still produces a proper layout. If that also fails, it
+falls back to a plain grid you can rearrange by dragging. Node positions you set are
```
laskevych, line 22: 🔵 Question: does the Joinable Data Marts diagram actually reach the retry? In `RelationshipCanvas.tsx` node keys are unique alias paths with exactly one parent edge each, so `simplifyEdgesForLayout` never drops anything, `simplified.length < edges.length` is false, and any throw goes straight to the grid. Those nodes are also `draggable: false`, so the changeset's "a plain grid you can rearrange by dragging" does not hold for this surface.

If this diagram has been seen crashing, it might be worth scoping that sentence to the Model Canvas, or giving that caller a tree-aware fallback (`depth` is already on `NodeInfo`). If it has not, the grid is just an unreachable safety net here and nothing to do.

.changeset/6835-fix-canvas.md
```diff
@@ -0,0 +1,23 @@
...
+**Problem.** Opening the Model Canvas — or the Joinable Data Marts diagram — for a
+storage with many Data Marts and relationships could replace the whole page with a
+"Something went wrong" error screen right after the loading skeleton. Storages with
+only a few Data Marts were unaffected, so the failure was intermittent and hard to
+reproduce.
+
+**Cause.** The automatic layout step throws on some otherwise-valid relationship
+graphs — typically when a model has several relationships between the same pair of
+Data Marts, a Data Mart joined to itself, or circular joins. That error was not
+handled, so it propagated to the route error boundary and took down the entire page
```
laskevych, line 16: 🔵 Question / 📝 Docs, two small things on the release note:

1. "propagated to the route error boundary" is implementation detail in user-facing text; the "Something went wrong" screen sentence already conveys the outcome.
2. The `Cause` paragraph names "a Data Mart joined to itself, or circular joins", but neither triggers this error on its own (verified with dagre 3.1.1); the trigger is three or more relationships between the same two Data Marts, in either direction. Users reading the note will otherwise look for cycles that are not the problem.

Also, can you confirm the ID in the filename owns this fix? `dagre-layout.ts` predates the canvas polish work that already has a changeset with the same prefix on `main`, so it is not derivable from the repo. Fine if it is the same work item, just checking.

laskevych (approved):
Approving: no blocking findings. The recovery works and I could not construct an input where the retry fails on a graph that main handled.

What I verified locally on this branch:
- `dagre-layout.test.ts`: 11/11 pass; the two new cases fail on `main` with the exact `Not possible to find intersection inside of the rectangle` error, so the fixture is a real regression test.
- `tsc -b` and `eslint` for `apps/web` are clean. The `setStorageId` destructure in `ModelCanvasView.tsx` removes a real `react-hooks/exhaustive-deps` warning that `main` currently emits, so it is a lint fix, not just a tidy.
- With dagre 3.1.1: the raw fixture throws; deduplicating parallel edges clears it (a fuzz over 4,000 random graphs found 167 throws, all cleared by the dedup); a plain 2-cycle, a 3-cycle with 2 parallel edges, and a self-loop all lay out fine without any retry.

The inline comments are all non-blocking: one on observability (silent catch), one on the test not pinning the retry path, and a few on the mechanism description, which does not match what dagre actually does. Details inline.

PR #1595, opened 2026-09-07
fix: prevent truncated Data Studio responses

kalnik-a-a:
## Summary

Prevents large Data Studio report responses from being cut off as invalid JSON near the Apps Script URL Fetch response-size limit. The backend now returns only complete rows that fit within the configured response budget and finishes processing the report only after the compressed HTTP response completes.

## Problem

Large reports could fail in the Data Studio connector with an error such as:

```text
SyntaxError: Unterminated string in JSON at position 52401192
```

At the same time, the backend could log the report as successfully completed. The connector and backend therefore disagreed about the outcome of the same request.

### Example

If a response was already close to the 50 MB Apps Script limit, the backend could enqueue another batch of up to 5,000 rows. Apps Script could then truncate the response in the middle of a value:

```json
{"values":["an incomplete value
```

The connector could not parse that response, although the backend had already continued through its success flow.

## Root cause

The streaming implementation used `gzip.bytesWritten` as an immediate response-size counter after calling `gzip.write()`. That value reflects data processed asynchronously by the compression stream, so it could lag behind the data already queued for writing.

The size guard also ran only after an entire batch had been written and did not reserve space for the closing JSON structure. As a result, a batch could cross the limit before the backend detected it.

Finally, `streamData()` returned immediately after `gzip.end()` instead of waiting for the gzip and HTTP streams to finish. This allowed report success handling to run before delivery was actually complete.

## Solution

- Track the exact UTF-8 size of the JSON response independently of gzip state.
- Check each serialized row before writing it, including its separator and the reserved closing JSON structure.
- Stop at the last complete row that fits, then close the JSON normally.
- Await gzip writes and the complete HTTP pipeline so stream failures propagate to report-run handling.

This approach enforces the limit before data is written. Simply reducing the batch size or lowering the threshold would only reduce the likelihood of an overrun because row and batch sizes are variable. Using Node's stream pipeline also provides a single completion and error boundary for both gzip and the HTTP response.

## Test plan

- [x] Verify exact response-size accounting for UTF-8 JSON.
- [x] Verify that a row exceeding the remaining budget is not partially written and the response remains valid JSON.
- [x] Verify that `streamData()` does not resolve before the compressed response finishes.
- [x] Verify that a closed response rejects the streaming operation.
- [x] Backend unit suite: 9,433 tests passed.
- [x] Data Studio API E2E suite: 23 tests passed with streaming enabled.
- [x] Backend build, ESLint, Prettier, and changeset validation.

### DIFF
--- a/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.spec.ts
+++ b/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.spec.ts
@@ -1,5 +1,5 @@
 import { Response } from 'express';
-import { PassThrough } from 'stream';
+import { PassThrough, Writable } from 'stream';
 import * as zlib from 'zlib';
 import { Test } from '@nestjs/testing';
 import { BigQueryFieldType } from '../../../data-storage-types/bigquery/enums/bigquery-field-type.enum';
@@ -185,6 +185,123 @@ describe('LookerStudioConnectorApiDataService', () => {
   });
 
   describe('streamData', () => {
+    it('reports the exact UTF-8 byte size of the JSON body', async () => {
+      const { res, chunks } = createMockResponse();
+
+      const context = {
+        schema: [{ name: 'emoji 🚀', dataType: FieldDataType.STRING }],
+        reader: {
+          readReportDataBatch: jest.fn().mockResolvedValue({
+            dataRows: [['Hello 😀']],
+            nextDataBatchId: null,
+          }),
+        },
+        fieldIndexMap: [0],
+        rowLimit: 1_000_000,
+      };
+
+      const streamResultPromise = service.streamData(res as Response, context as any);
+      await (res as any).waitForFinish();
+      const result = await streamResultPromise;
+      const body = zlib.gunzipSync(Buffer.concat(chunks));
+
+      expect(result.bytesWritten).toBe(body.byteLength);
+    });
+
+    it('does not write a UTF-8 row that would exceed the Apps Script response limit', async () => {
+      const { res, chunks } = createMockResponse();
+
+      const context = {
+        schema: [{ name: 'field1', dataType: FieldDataType.STRING }],
+        reader: {
+          readReportDataBatch: jest.fn().mockResolvedValue({
+            dataRows: [['kept'], ['🚀'.repeat(13_000_000)]],
+            nextDataBatchId: null,
+          }),
+        },
+        fieldIndexMap: [0],
+        rowLimit: 2,
+      };
+
+      const streamResultPromise = service.streamData(res as Response, context as any);
+      await (res as any).waitForFinish();
+      const result = await streamResultPromise;
+      const body = zlib.gunzipSync(Buffer.concat(chunks));
+
+      expect(result.limitExceeded).toBe(true);
+      expect(result.rowCount).toBe(1);
+      expect(body.byteLength).toBeLessThanOrEqual(49 * 1024 * 1024);
+      expect(JSON.parse(body.toString()).rows).toEqual([{ values: ['kept'] }]);
+    });
+
+    it('does not resolve before the compressed response finishes', async () => {
+      let releaseFirstWrite!: () => void;
+      let markFirstWriteStarted!: () => void;
+      const firstWriteStarted = new Promise<void>(resolve => {
+        markFirstWriteStarted = resolve;
+      });
+      let shouldBlockWrite = true;
+      const res = new Writable({
+        write(_chunk, _encoding, callback) {
+          if (shouldBlockWrite) {
+            shouldBlockWrite = false;
+            releaseFirstWrite = callback;
+            markFirstWriteStarted();
+            return;
+          }
+          callback();
+        },
+      }) as Writable & Partial<Response>;
+      res.setHeader = jest.fn().mockReturnThis();
+
+      const context = {
+        schema: [{ name: 'field1', dataType: FieldDataType.STRING }],
+        reader: {
+          readReportDataBatch: jest.fn().mockResolvedValue({
+            dataRows: [['value1']],
+            nextDataBatchId: null,
+          }),
+        },
+        fieldIndexMap: [0],
+        rowLimit: 1_000_000,
+      };
+
+      let resolved = false;
+      const resultPromise = service.streamData(res as Response, context as any).then(result => {
+        resolved = true;
+        return result;
+      });
+
+      await firstWriteStarted;
+      await Promise.resolve();
+
+      expect(resolved).toBe(false);
+      releaseFirstWrite();
+      await resultPromise;
+      expect(res.writableFinished).toBe(true);
+    });
+
+    it('rejects when the response closes while a batch is loading', async () => {
+      const res = new PassThrough() as PassThrough & Partial<Response>;
+      res.setHeader = jest.fn().mockReturnThis();
+      res.resume();
+
+      const context = {
+        schema: [{ name: 'field1', dataType: FieldDataType.STRING }],
+        reader: {
+          readReportDataBatch: jest.fn().mockImplementation(async () => {
+            res.destroy();
+            await new Promise(resolve => setImmediate(resolve));
+            return { dataRows: [], nextDataBatchId: null };
+          }),
+        },
+        fieldIndexMap: [0],
+        rowLimit: 1_000_000,
+      };
+
+      await expect(service.streamData(res as Response, context as any)).rejects.toThrow();
+    });
+
     it('should stream JSON response with correct structure', async () => {
       const { res, chunks, headers } = createMockResponse();
 
--- a/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.spec.ts
+++ b/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.spec.ts
@@ -281,6 +281,69 @@ describe('LookerStudioConnectorApiDataService', () => {
       expect(res.writableFinished).toBe(true);
     });
 
+    it('rejects when the response closes during a backpressured gzip write', async () => {
+      let destinationWrites = 0;
+      let markBackpressured!: () => void;
+      const backpressured = new Promise<void>(resolve => {
+        markBackpressured = resolve;
+      });
+      let gzip!: zlib.Gzip;
+      const res = new Writable({
+        highWaterMark: 1,
+        write(_chunk, _encoding, callback) {
+          destinationWrites += 1;
+          if (destinationWrites === 1) {
+            callback();
+            return;
+          }
+          markBackpressured();
+        },
+      }) as Writable & Partial<Response>;
+      res.setHeader = jest.fn().mockReturnThis();
+      res.once('pipe', source => {
+        gzip = source as zlib.Gzip;
+      });
+
+      const context = {
+        schema: [{ name: 'field1', dataType: FieldDataType.STRING }],
+        reader: {
+          readReportDataBatch: jest.fn().mockResolvedValue({
+            dataRows: [[Array.from({ length: 100_000 }, (_, i) => i.toString(36)).join(',')]],
+            nextDataBatchId: null,
+          }),
+        },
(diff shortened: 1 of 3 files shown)


apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.ts
```diff
@@ -347,96 +348,124 @@ export class LookerStudioConnectorApiDataService {
...
-      bytesWritten: gzip.bytesWritten,
+      bytesWritten,
     };
   }
+
+  private writeGzipChunk(gzip: zlib.Gzip, chunk: Buffer): Promise<void> {
+    return new Promise((resolve, reject) => {
+      gzip.write(chunk, error => (error ? reject(error) : resolve()));
+    });
+  }
```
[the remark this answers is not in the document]
kalnik-a-a: Addressed in `358dab48e`. Each gzip write now races its callback against the existing pipeline completion, so a downstream error or close rejects a pending backpressured write promptly. Added a regression test that closes the response while gzip is backpressured; it failed with `streamData remained pending` before the fix and now passes.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.ts
```diff
@@ -347,96 +348,135 @@ export class LookerStudioConnectorApiDataService {
...
 
-      // Write entire batch as single chunk (much faster than row-by-row)
-      if (formattedRows.length > 0) {
-        const batchJson = formattedRows.map(row => JSON.stringify(row)).join(',');
-        gzip.write(isFirstBatch ? batchJson : ',' + batchJson);
-        isFirstBatch = false;
-        totalRows += formattedRows.length;
-      }
+        const batch = await reader.readReportDataBatch(nextBatchId, batchSize);
+        if (res.closed) {
```
laskevych, line 393: 🟡 Important (non-blocking): This changes the outcome of a client disconnect from a successful run to a failed one. On `main`, `res.closed` was `warn + break`, the closing JSON was written and the run went through `handleSuccessfulReportRun` (including `registerLookerReportRunConsumption`). Now the loop throws, so `handleFailedReportRun` marks the run `ERROR`, logs at error level, publishes `LookerReportRunEvent(..., 'unsuccessfully')`, and the run is not billed.

Two details worth an explicit decision:
- The stored `lastRunError` text depends on timing: the `res.closed` checks yield `Streaming aborted: response closed`, but when the socket closes during a backpressured write, `pipeline` rejects first with Node's `ERR_STREAM_PREMATURE_CLOSE`, so Run History shows `Premature close`. One dedicated error (e.g. a `StreamCancelledError` like the one in `stream-http-data.service.ts`) thrown from a single helper and mapped to a human-readable message would make this deterministic.
- The HTTP-data streamer maps the same condition to `ReportRunStatus.CANCELLED`; the Looker run has no cancelled state, so a viewer closing a tab mid-refresh now counts as a failure in analytics and in the report card. Treating it as failed may be the intended semantics (the client never got the data), but it is a user-visible change that the changeset does not mention.

I'm fine with this landing as is if the "disconnect = failed run" semantics are intentional; the message normalization can be a follow-up.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.ts
```diff
@@ -347,96 +348,135 @@ export class LookerStudioConnectorApiDataService {
...
+    const prefix = Buffer.from(`{"schema":${JSON.stringify(schema)},"rows":[`, 'utf8');
+    const suffix = Buffer.from(`],"filtersApplied":[]}`, 'utf8');
+
+    if (prefix.byteLength + suffix.byteLength > MAX_BYTES_LIMIT) {
+      throw new BusinessViolationException(
+        `Looker Studio response schema exceeds the size limit (${MAX_BYTES_LIMIT} bytes)`
+      );
+    }
+
+    if (res.closed) {
```
laskevych, line 360: 🟡 Important (non-blocking): This pre-header throw is a plain `Error` while `res.headersSent` is still `false`, so the caller rethrows it (`looker-studio-connector-api.service.ts:231`) and `GlobalExceptionFilter` treats it as an HTTP 500: it logs `Unhandled exception caught by GlobalExceptionFilter` with a stack and writes `status(500).json()` to an already destroyed response. Realistic trigger: the client gives up while `getCachedReader` is still running the warehouse query. Net effect per occurrence is a FAILED run plus two error-level log lines for a client-side abort. Same root cause as the comment on the in-loop check: a dedicated cancellation error the caller can recognize (and skip the rethrow / log at warn) would fix both.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.ts
```diff
@@ -347,96 +348,135 @@ export class LookerStudioConnectorApiDataService {
...
+        }
 
-      // Check if we've reached the row limit
-      if (totalRows >= rowLimit) {
-        if (nextBatchId) {
-          const limitType = rowLimit < MAX_ROWS_LIMIT ? 'sample extraction' : 'full extraction';
-          this.logger.warn(
-            `Row limit reached during streaming (${limitType}): returned ${totalRows} rows (more data available)`
+        if (batchJson.length > 0) {
+          await this.writeGzipChunk(
```
laskevych, line 419: 💡 Suggestion (non-blocking): Awaiting every batch write means `readReportDataBatch(N+1)` does not start until batch N has been compressed and, once gzip's readable side is full, drained downstream. On `main` the fetch overlapped with compression and sending (at the cost of unbounded buffering inside gzip, which this PR rightly removes). In a simulation with a small downstream buffer the total time approached fetch + transfer instead of max(fetch, transfer); with kernel-sized socket buffers the difference is under 1%, so this is not a regression to block on. If you want the overlap back cheaply: keep one write in flight (start the write for batch N without awaiting, fetch N+1, then await the pending write before issuing the next one). Memory stays bounded to one batch and the error propagation through `writeGzipChunk` is unchanged.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.spec.ts
```diff
@@ -185,6 +185,179 @@ describe('LookerStudioConnectorApiDataService', () => {
...
+      let resolved = false;
+      const resultPromise = service.streamData(res as Response, context as any).then(result => {
+        resolved = true;
+        return result;
+      });
+
+      await firstWriteStarted;
+      await Promise.resolve();
+
+      expect(resolved).toBe(false);
```
laskevych, line 278: 🧪 Test note (non-blocking): `expect(resolved).toBe(false)` passes even when `await streamCompletion;` is removed from `streamData` (I ran that mutation): one microtask after the first destination write, zlib is still working, so `streamData` cannot have settled either way. The assertion that actually guards the PR's claim is `expect(res.writableFinished).toBe(true)` at the end. Consider dropping the first assertion or replacing it with something that distinguishes the two implementations (e.g. release the blocked write, `await` a macrotask, and only then assert `resolved === false` before letting the response finish), so the test's name matches what it proves.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.ts
```diff
@@ -347,96 +348,135 @@ export class LookerStudioConnectorApiDataService {
...
-
-      const batch = await reader.readReportDataBatch(nextBatchId, batchSize);
+    try {
+      await this.writeGzipChunk(gzip, prefix, streamCompletion);
 
-      // Format all rows in the batch
-      const rowsToWrite = Math.min(batch.dataRows.length, rowLimit - totalRows);
-      const formattedRows: DataRow[] = [];
+      do {
+        if (res.closed) {
```
laskevych, line 385: 🧹 Nit: This top-of-loop `res.closed` check is unreachable in practice. After an awaited write, a closed response already rejected the `Promise.race` in `writeGzipChunk`; after an empty batch there is no `await` between the after-read check and this one, so it re-tests the same value. A probe with 400 aborting clients hit it zero times. The pre-pipeline check (line 360) and the after-read check (line 393) are the two that matter.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.ts
```diff
@@ -347,96 +348,135 @@ export class LookerStudioConnectorApiDataService {
...
     res.setHeader('Transfer-Encoding', 'chunked');
     res.setHeader('X-Accel-Buffering', 'no');
 
-    // Pipe gzip stream to response
-    gzip.pipe(res);
-
-    // Write opening JSON structure with schema
-    gzip.write(`{"schema":${JSON.stringify(schema)},"rows":[`);
+    const streamCompletion = pipeline(gzip, res);
+    streamCompletion.catch(() => undefined);
```
laskevych, line 373: 🧹 Nit: The standalone `streamCompletion.catch(() => undefined)` is dead. `writeGzipChunk` attaches a handler via `Promise.race` in the same synchronous run as `pipeline()`, so there is no tick in which a rejection could go unhandled (verified with `res`/gzip destroyed before, on `nextTick`, microtask and `setImmediate`). The `await streamCompletion.catch(...)` inside the `catch` block is the intentional teardown wait and should stay; removing this one makes that distinction clearer.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api-data.service.ts
```diff
@@ -347,96 +348,135 @@ export class LookerStudioConnectorApiDataService {
...
-      nextBatchId = batch.nextDataBatchId;
+        const rowsToWrite = Math.min(batch.dataRows.length, rowLimit - totalRows);
+        const batchJson: string[] = [];
+
+        for (let i = 0; i < rowsToWrite; i++) {
+          const row: DataRow = {
+            values: fieldIndexMap.map(index => this.convertToFieldValue(batch.dataRows[i][index])),
+          };
+          const serializedRow = `${totalRows === 0 ? '' : ','}${JSON.stringify(row)}`;
+          const rowBytes = Buffer.byteLength(serializedRow, 'utf8');
```
laskevych, line 405: 🧹 Nit: Each row is UTF-8 encoded twice: `Buffer.byteLength(serializedRow)` for accounting, then `Buffer.from(batchJson.join(''))` for the write. Encoding once per row (`const buf = Buffer.from(serializedRow)`, count `buf.length`, `Buffer.concat` per batch) gives the exact byte count in a single pass. Measured at ~150 ms per 1M-row response, so purely optional.

laskevych (approved):
✅ Approving. The root-cause analysis checks out: I confirmed on Node 24 that `gzip.bytesWritten` is still 0 immediately after `write()`, so the old post-batch check could lag by whole batches, and the new per-row UTF-8 budget with the reserved suffix closes the overrun. Running the spec three times in a row was stable (16/16), and the `pipeline`-based completion correctly propagates a closed response (`ERR_STREAM_PREMATURE_CLOSE`) into the run's failure path.

None of my comments block the merge. The two 🟡 items (disconnect semantics and the pre-header 500) are about how a client abort is classified and logged, not about the truncation fix itself, and can be a follow-up if you agree with the direction.

📝 Note (non-blocking, on the untouched caller `looker-studio-connector-api.service.ts:231`): with this PR, `res.headersSent` is already `true` right after the prefix write (gzip emits its header on the first chunk), so every reader error inside the loop takes the "cannot send error response" branch, and `gzip.destroy()` makes `pipeline` destroy `res`, i.e. the client now sees a connection reset rather than an incomplete body. That is an improvement over the response hanging until the socket timeout, but the comment "If headers haven't been sent yet, throw to let error handler respond" now only describes the two pre-`setHeader` throws. Worth updating the wording so nobody expects storage errors to reach Looker Studio as a 400 with diagnostics.

PR #1596, opened 2026-09-08
fix: surface Data Studio schema query errors

kalnik-a-a:
# Summary

Makes query failures during Data Studio connector setup understandable and relevant to the user. Instead of reporting a generic deployment URL problem, Data Studio now shows guidance together with the original Data Mart query error.

This change does not suppress, recover from, or otherwise hide the underlying failure. The schema request still fails when the Data Mart query cannot be executed; the failure is returned through the connector's supported error contract so the user can act on the actual cause.

## Problem

When a user connects a Data Mart to Data Studio, the connector requests the report schema. To produce that schema, the backend creates or restores a cached report reader and executes the initial query batch.

If query preparation or execution failed in `ReportDataCacheService.createNewCachedReader`, Data Studio displayed a generic connection error suggesting that the deployment URL should be checked. That message was misleading when the URL and connector deployment were valid and the real problem was the Data Mart query or storage access.

## Root cause

Errors from the cached-reader path propagated as unhandled exceptions. The backend therefore returned HTTP 500 for a failed schema request.

The Data Studio connector only extracts and displays the backend's structured error message for HTTP 400 responses. For other HTTP statuses, including 500, it intentionally falls back to a generic message about being unable to connect to the server and checking the deployment URL.

The query error was therefore not lost inside the warehouse or fixed by retrying: it crossed the backend boundary with a status the connector treats as an infrastructure failure, so the connector replaced the relevant query message with unrelated deployment guidance.

## Solution

- Catch cached-reader failures at the shared Looker Studio API boundary.
- Preserve existing `BusinessViolationException` instances unchanged, including their message, code, and details.
- Convert other reader failures into an HTTP 400 business error with actionable guidance and the original error message:

  ```text
  Failed to read data from this Data Mart. Check the Data Mart query and storage access, or contact the Data Mart owner. Details: <original error message>
  ```

- Log the original failure server-side with its stack trace, `reportId`, `dataMartId`, and `projectId` for proactive diagnosis.

The backend does not return a fake schema, swallow the exception, or claim that the query succeeded. It keeps the failure visible while presenting it in a form that is meaningful to both Data Studio and the user.

### Before

> CleanShot 2026-09-08 at 15 35 30

### After

> CleanShot 2026-09-08 at 15 57 19

## Test plan

- [x] Verify that an unknown cached-reader failure returns HTTP 400 with guidance and the original query error.
- [x] Verify that cached-reader failures are logged with the original stack trace and report context.
- [x] Verify that an existing `BusinessViolationException` is rethrown unchanged with its code and details.
- [x] Looker Studio API E2E suite: 24 tests passed.
- [x] Backend unit suite: 9,436 tests passed across 613 suites.
- [x] Backend build, ESLint, Prettier, markdownlint, and `git diff --check` passed.

### DIFF
--- a/.changeset/6919-data-studio-schema-query-errors.md
+++ b/.changeset/6919-data-studio-schema-query-errors.md
@@ -4,4 +4,4 @@
 
 # Show Data Studio query errors during connection
 
-Data Studio now shows troubleshooting guidance and the underlying query error when loading a Data Mart schema fails instead of suggesting that the deployment URL is incorrect.
+Data Studio now shows troubleshooting guidance and the underlying query error when loading a Data Mart schema or its data fails instead of suggesting that the deployment URL is incorrect.
--- a/.changeset/6919-data-studio-schema-query-errors.md
+++ b/.changeset/6919-data-studio-schema-query-errors.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Show Data Studio query errors during connection
+
+Data Studio now shows troubleshooting guidance and the underlying query error when loading a Data Mart schema fails instead of suggesting that the deployment URL is incorrect.
--- a/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.spec.ts
+++ b/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.spec.ts
@@ -2,12 +2,14 @@ import { Response } from 'express';
 import { Report } from '../../../entities/report.entity';
 import { GetDataRequest } from '../schemas/get-data.schema';
 import { GetSchemaRequest } from '../schemas/get-schema.schema';
+import { BusinessViolationException } from '../../../../common/exceptions/business-violation.exception';
 import { ProjectOperationBlockedException } from '../../../../common/exceptions/project-operation-blocked.exception';
 import { ProjectBlockedReason } from '../../../enums/project-blocked-reason.enum';
 
 // Mock external modules before importing service
 jest.mock('@owox/internal-helpers', () => ({
   createProducer: jest.fn(),
+  castError: (error: unknown) => (error instanceof Error ? error : new Error(String(error))),
   BaseEvent: class {
     constructor(public payload: unknown) {}
   },
@@ -174,6 +176,40 @@ describe('LookerStudioConnectorApiService', () => {
 
       expect(projectBilling.verifyCanPerformOperations).not.toHaveBeenCalled();
       expect(schemaService.getSchema).toHaveBeenCalledWith(request, report, cachedReader);
+    });
+
+    it('logs cached reader failures with report context', async () => {
+      const request = createMockSchemaRequest();
+      const report = createMockReport();
+      const queryError = new Error('Query execution failed');
+      reportService.getByIdAndLookerStudioSecret.mockResolvedValue(report);
+      cacheService.getOrCreateCachedReader.mockRejectedValue(queryError);
+
+      await expect(service.getSchema(request)).rejects.toThrow('Query execution failed');
+
+      expect((service as any).logger.error).toHaveBeenCalledWith(
+        'Failed to read Data Mart data for Looker Studio: Query execution failed',
+        queryError.stack,
+        {
+          reportId: 'report-1',
+          dataMartId: 'datamart-1',
+          projectId: 'project-1',
+        }
+      );
+    });
+
+    it('preserves cached reader business errors with their code and details', async () => {
+      const request = createMockSchemaRequest();
+      const report = createMockReport();
+      const queryError = new BusinessViolationException(
+        'Query execution failed',
+        { reason: 'missing column' },
+        'QUERY_EXECUTION_FAILED'
+      );
+      reportService.getByIdAndLookerStudioSecret.mockResolvedValue(report);
+      cacheService.getOrCreateCachedReader.mockRejectedValue(queryError);
+
+      await expect(service.getSchema(request)).rejects.toBe(queryError);
     });
   });
 
--- a/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.spec.ts
+++ b/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.spec.ts
@@ -3,13 +3,13 @@ import { Report } from '../../../entities/report.entity';
 import { GetDataRequest } from '../schemas/get-data.schema';
 import { GetSchemaRequest } from '../schemas/get-schema.schema';
 import { BusinessViolationException } from '../../../../common/exceptions/business-violation.exception';
+import { DataMartReadFailedException } from '../../../errors/data-mart-read-failed.error';
 import { ProjectOperationBlockedException } from '../../../../common/exceptions/project-operation-blocked.exception';
 import { ProjectBlockedReason } from '../../../enums/project-blocked-reason.enum';
 
 // Mock external modules before importing service
 jest.mock('@owox/internal-helpers', () => ({
   createProducer: jest.fn(),
-  castError: (error: unknown) => (error instanceof Error ? error : new Error(String(error))),
   BaseEvent: class {
     constructor(public payload: unknown) {}
   },
@@ -178,24 +178,16 @@ describe('LookerStudioConnectorApiService', () => {
       expect(schemaService.getSchema).toHaveBeenCalledWith(request, report, cachedReader);
     });
 
-    it('logs cached reader failures with report context', async () => {
+    it('does not duplicate-log a cached reader failure already logged by the cache service', async () => {
       const request = createMockSchemaRequest();
       const report = createMockReport();
-      const queryError = new Error('Query execution failed');
+      const queryError = new DataMartReadFailedException(new Error('Query execution failed'));
       reportService.getByIdAndLookerStudioSecret.mockResolvedValue(report);
       cacheService.getOrCreateCachedReader.mockRejectedValue(queryError);
 
-      await expect(service.getSchema(request)).rejects.toThrow('Query execution failed');
+      await expect(service.getSchema(request)).rejects.toBe(queryError);
 
-      expect((service as any).logger.error).toHaveBeenCalledWith(
-        'Failed to read Data Mart data for Looker Studio: Query execution failed',
-        queryError.stack,
-        {
-          reportId: 'report-1',
-          dataMartId: 'datamart-1',
-          projectId: 'project-1',
-        }
-      );
+      expect((service as any).logger.error).not.toHaveBeenCalled();
     });
 
     it('preserves cached reader business errors with their code and details', async () => {
@@ -204,12 +196,35 @@ describe('LookerStudioConnectorApiService', () => {
       const queryError = new BusinessViolationException(
         'Query execution failed',
         { reason: 'missing column' },
-        'QUERY_EXECUTION_FAILED'
+        'ANY_CODE'
       );
       reportService.getByIdAndLookerStudioSecret.mockResolvedValue(report);
       cacheService.getOrCreateCachedReader.mockRejectedValue(queryError);
 
       await expect(service.getSchema(request)).rejects.toBe(queryError);
+      expect((service as any).logger.error).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('getData', () => {
+    it('stores the original reader failure while returning connector guidance', async () => {
+      const request = createMockRequest();
+      const report = createMockReport();
+      const originalError = new Error('Query execution failed');
+      const connectorError = new DataMartReadFailedException(originalError);
+      const reportRun = {
+        markAsUnsuccessful: jest.fn(),
+        getReport: jest.fn().mockReturnValue(report),
+        getReportId: jest.fn().mockReturnValue('report-1'),
+      };
+      reportService.getByIdAndLookerStudioSecret.mockResolvedValue(report);
+      reportRunService.create.mockResolvedValue(reportRun as never);
+      cacheService.getOrCreateCachedReader.mockRejectedValue(connectorError);
+
+      await expect(service.getData(request)).rejects.toBe(connectorError);
+
+      expect(reportRun.markAsUnsuccessful).toHaveBeenCalledWith(originalError);
+      expect((service as any).logger.error).not.toHaveBeenCalled();
     });
   });
 
--- a/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
+++ b/apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
@@ -1,5 +1,4 @@
 import { Injectable, InternalServerErrorException, Logger } from '@nestjs/common';
-import { castError } from '@owox/internal-helpers';
 import { Response } from 'express';
 import { BusinessViolationException } from '../../../../common/exceptions/business-violation.exception';
 import { ProjectOperationBlockedException } from '../../../../common/exceptions/project-operation-blocked.exception';
@@ -7,6 +6,7 @@ import { OwoxEventDispatcher } from '../../../../common/event-dispatcher/owox-ev
 import { SystemTimeService } from '../../../../common/scheduler/services/system-time.service';
 import { CachedReaderData } from '../../../dto/domain/cached-reader-data.dto';
 import { Report } from '../../../entities/report.entity';
+import { DataMartReadFailedException } from '../../../errors/data-mart-read-failed.error';
 import { LookerReportRunEvent } from '../../../events/looker-report-run.event';
 import { LookerStudioReportRun } from '../../../models/looker-studio-report-run.model';
 import { logBlendedSqlIfNeeded } from '../../../report-run-logging/log-blended-sql';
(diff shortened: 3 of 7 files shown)


apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
```diff
@@ -271,7 +272,28 @@ export class LookerStudioConnectorApiService {
...
+      this.logger.error(
+        `Failed to read Data Mart data for Looker Studio: ${cause.message}`,
+        cause.stack,
+        {
+          reportId: report.id,
+          dataMartId: report.dataMart.id,
+          projectId: report.dataMart.projectId,
+        }
+      );
+      if (error instanceof BusinessViolationException) {
```
laskevych: 🟡 Important (non-blocking): Nest `HttpException`s thrown inside the cached-reader path are re-wrapped here as well. `ReportSqlComposerService.compose()` (reached via `resolvePrepareOptions` for non-blended reports with output controls) throws `BadRequestException` with a structured body — e.g. the `is_blank` / `is_not_blank` type-resolution error (which is reachable on this route because `getCachedReader` runs *before* `actualizeSchemaInEntity`), `PRE_JOIN_FILTERS_REQUIRE_JOINED_DATA_MART`, `OUTPUT_CONTROLS_NOT_SUPPORTED`. Those were already HTTP 400 with `message` + `details.errors[].code` via `GlobalExceptionFilter`, so Looker Studio displayed them correctly before this PR. After it the user gets `Failed to read data from this Data Mart. Check the Data Mart query and storage access… Details: The "is blank" / "is not blank" filters need…` (misleading advice for a schema/config condition), the `details`/`code` payload is dropped, and an error-level log with a stack is emitted.

`calculated-field-validator.service.ts:508` already handles the same situation:

```ts
if (error instanceof BusinessViolationException || error instanceof HttpException) {
  throw error;
}
```

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
```diff
@@ -271,7 +272,28 @@ export class LookerStudioConnectorApiService {
...
+        throw error;
+      }
+      throw new BusinessViolationException(
```
laskevych: 🟡 Important (non-blocking): `getCachedReader` is shared by `getSchema`, `getSampleDataExtraction`, `getFullDataExtraction` and `getFullDataExtractionStreaming`, so this wrapper also changes the getData paths, while the PR description, changeset and tests only describe schema requests. Two observable side effects there:

1. In both full-extraction paths the wrapped exception is what reaches `handleFailedReportRun` → `markAsUnsuccessful`, so `report.lastRunError` and `dataMartRun.errors` now store `Failed to read data from this Data Mart. … or contact the Data Mart owner. Details: <cause>`. That sentence is addressed to a Looker Studio viewer but is rendered verbatim to the Data Mart owner in ODM Run History (`ReportHoverCard`, `LookerStudioReportCard`, the reports table columns), whereas the same failure on a Google Sheets run stores the bare provider message.
2. The second `logger.error` in `handleFailedReportRun` now receives the wrapper, whose stack points at `getCachedReader` instead of the SDK call.

Options: keep the raw cause for run persistence (e.g. carry it on the exception and unwrap it in `handleFailedReportRun`), or apply the guidance only on the getSchema path if that is the intended scope. If getData stays in scope, the changeset body should say so.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
```diff
@@ -271,7 +272,28 @@ export class LookerStudioConnectorApiService {
       report.createdById
     );
 
-    return this.cacheService.getOrCreateCachedReader(report, accessor);
+    try {
+      return await this.cacheService.getOrCreateCachedReader(report, accessor);
+    } catch (error) {
+      const cause = castError(error);
+      this.logger.error(
```
laskevych: 💡 Suggestion (non-blocking): the log runs before the `instanceof BusinessViolationException` rethrow, so expected business refusals raised inside the cache path (blending refusals in `blended-report-data.service.ts`, the Redshift reader's `BusinessViolationException`, …) are now logged at `error` with a stack — previously they produced no log line at all, since `BaseExceptionFilter` does not log. The sibling `run-report.service.ts:641-646` logs `BusinessViolationException` at `warn` and reserves `error` for unknown failures; `handleFailedReportRun` in this file applies the same split for `ProjectOperationBlockedException`.

Two multipliers worth knowing: on the getData paths this is now the third error-level entry for one failure (reader → here → `handleFailedReportRun` / `Failed to get sample data:`), and `ReportDataCacheService.pendingOperations` hands the same rejected promise to every concurrent caller, so a dashboard with N charts on a broken report logs N identical stacks per refresh.

Moving the rethrow above `castError`/log keeps the new entry where it adds value: unknown errors on `getSchema`, which otherwise lose the `GlobalExceptionFilter` 500 log.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
```diff
@@ -271,7 +272,28 @@ export class LookerStudioConnectorApiService {
...
+          projectId: report.dataMart.projectId,
+        }
+      );
+      if (error instanceof BusinessViolationException) {
+        throw error;
+      }
+      throw new BusinessViolationException(
+        'Failed to read data from this Data Mart. ' +
+          'Check the Data Mart query and storage access, or contact the Data Mart owner. ' +
+          `Details: ${cause.message}`
```
laskevych: 🔵 Question (non-blocking): `Details: ${cause.message}` forwards the raw provider / TypeORM text to whoever holds the report secret — Looker Studio report viewers, who need not be members of the ODM project. Before this PR a plain `Error` produced a 500 whose body carried no `message`. Examples of what now crosses the boundary: BigQuery `Access Denied: Project <gcp-project>: User <sa>@….iam.gserviceaccount.com does not have bigquery.jobs.create permission`, Athena `User: arn:aws:iam::<account>:… is not authorized …`, or MySQL `QueryFailedError` text from `cacheRepository.save` (`report-data-cache.service.ts:181`). The MCP query tool made the opposite call for the same class of errors (`query-data-mart.tool.ts:534`: "Never forward the raw message — it can carry SQL/identifiers/PII").

Is forwarding the full provider text to viewers the intended policy? If only the query error is wanted, the existing `DataStorageErrorMapper.toStorageReadError` (e.g. `bigquery-storage-error.mapper.ts`) already extracts the provider `reason`/`message` and a stable code (`STORAGE_PERMISSION_DENIED`, `STORAGE_UNAUTHORIZED`, `STORAGE_READ_FAILED`) and could feed a curated message here.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
```diff
@@ -271,7 +272,28 @@ export class LookerStudioConnectorApiService {
       report.createdById
     );
 
-    return this.cacheService.getOrCreateCachedReader(report, accessor);
+    try {
+      return await this.cacheService.getOrCreateCachedReader(report, accessor);
```
laskevych: 🔵 Question (non-blocking): this wraps only `getOrCreateCachedReader`, so several failures on the same two routes still surface as 500 and still trigger the generic "check the deployment URL" text:

- cache hit: `restoreReaderFromCache` performs no warehouse call (BigQuery `prepareReportData` only configures the reader, `initFromState` only builds a table reference), so the first real read happens in `dataService.getData` → `readReportDataBatch`, outside the try — revoked credentials or an expired result table fail there;
- `getSchema` continues with `dataMartService.actualizeSchemaInEntity` (a warehouse round trip) after the wrapped call;
- `resolveBlendableSchemaAccessor` sits before the try, so an IDP failure is still a 500.

Conversely, ODM-internal failures inside the try (MySQL down during `cacheRepository.findOne/save`, a plain bug) now become a 400 that tells the user to check their query and drop out of the 5xx bucket used for alerting.

Was a route-scoped filter on `LookerStudioConnectorController` considered? The repo already applies per-route filters for exactly this "adapt the error contract to the route's audience" reason (`PluginHostExceptionFilter` via `@UseFilters`, `McpAuthExceptionFilter`). It would cover every path above once, keep `requestId`/URL correlation in the log, and let unknown errors keep a distinct "unexpected error" text. Fine as a follow-up if the schema-setup case is the priority now.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
```diff
@@ -271,7 +272,28 @@ export class LookerStudioConnectorApiService {
...
+          reportId: report.id,
+          dataMartId: report.dataMart.id,
+          projectId: report.dataMart.projectId,
+        }
+      );
+      if (error instanceof BusinessViolationException) {
+        throw error;
+      }
+      throw new BusinessViolationException(
+        'Failed to read data from this Data Mart. ' +
```
laskevych: 💡 Suggestion (non-blocking): consider giving this exception a stable `code` (e.g. `DATA_MART_READ_FAILED`) and putting the original message into `errorDetails` rather than only into the sentence. `base.exception.ts` documents `code` as the thing clients branch on instead of matching human-readable messages, `BaseExceptionFilter` already emits it, and the unit spec even uses a `QUERY_EXECUTION_FAILED` code that does not exist anywhere in `src`. It would also let the e2e assert `code` + `toContain(<original message>)` instead of pinning the full 150-character sentence with `toBe`.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.spec.ts
```diff
@@ -175,6 +177,40 @@ describe('LookerStudioConnectorApiService', () => {
...
     });
+
+    it('logs cached reader failures with report context', async () => {
+      const request = createMockSchemaRequest();
+      const report = createMockReport();
+      const queryError = new Error('Query execution failed');
+      reportService.getByIdAndLookerStudioSecret.mockResolvedValue(report);
+      cacheService.getOrCreateCachedReader.mockRejectedValue(queryError);
+
+      await expect(service.getSchema(request)).rejects.toThrow('Query execution failed');
```
laskevych: 🧪 Test (non-blocking): `rejects.toThrow('Query execution failed')` is a substring match on `message`, so it also passes for the raw `Error` — this test stays green if the wrapping is removed. Asserting `rejects.toThrow(BusinessViolationException)` plus the guidance prefix (or `rejects.toMatchObject({ message: expect.stringContaining('Details: Query execution failed') })`) would pin the behaviour this PR introduces. None of the unit tests exercise a getData path, where the change applies as well.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.spec.ts
```diff
@@ -2,12 +2,14 @@ import { Response } from 'express';
...
 import { GetDataRequest } from '../schemas/get-data.schema';
 import { GetSchemaRequest } from '../schemas/get-schema.schema';
+import { BusinessViolationException } from '../../../../common/exceptions/business-violation.exception';
 import { ProjectOperationBlockedException } from '../../../../common/exceptions/project-operation-blocked.exception';
 import { ProjectBlockedReason } from '../../../enums/project-blocked-reason.enum';
 
 // Mock external modules before importing service
 jest.mock('@owox/internal-helpers', () => ({
   createProducer: jest.fn(),
+  castError: (error: unknown) => (error instanceof Error ? error : new Error(String(error))),
```
laskevych: 🧪 Nit (non-blocking): the hand-rolled `castError` differs from the real helper — `packages/internal-helpers/src/utils/castError.ts` JSON-stringifies non-string values, so a Google API plain-object rejection renders as `{"code":403,…}` in production but as `[object Object]` under this mock. `run-ai-assistant.service.spec.ts` spreads `jest.requireActual('@owox/internal-helpers')`; doing the same here lets the real, pure `castError` run and drops the duplicate.

.changeset/6919-data-studio-schema-query-errors.md
```diff
@@ -0,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Show Data Studio query errors during connection
+
+Data Studio now shows troubleshooting guidance and the underlying query error when loading a Data Mart schema fails instead of suggesting that the deployment URL is incorrect.
```
laskevych: 📝 Changeset (non-blocking): if the getData paths stay in scope (see the comment on `getCachedReader`), the body under-describes the effect — data requests also move from the generic connection error to the guidance + query error, and Run History text changes. Something like "when loading a Data Mart schema or its data fails" would cover it.

laskevych (approved):
✅ Approving — no blocking findings.

The root-cause analysis in the description matches the code: `BaseExceptionFilter` maps `BusinessViolationException` to 400, everything else went through `GlobalExceptionFilter` as 500, and a plain `Error` produced a 500 body with no `message` at all, so the connector had nothing to show. Verified along the way: `castError` is used per the review rules, `report.dataMart` is always loaded on this path (`innerJoinAndSelect` in `getByIdAndLookerStudioSecret`), the streaming path cannot emit the new 400 after headers are sent (`getCachedReader` runs before `prepareStreamingContext`), and the `ProjectOperationBlockedException` / RESTRICTED handling is untouched. Changeset ID, bump and format follow the policy.

All inline comments are non-blocking. Two I would still handle before merge if you agree:

- preserve Nest `HttpException`s next to `BusinessViolationException` in the rethrow, otherwise the structured 400s from `ReportSqlComposerService` lose their `code` and get misleading advice;
- the getData paths: the wrapped message is now what lands in Run History for the Data Mart owner, and the changeset only mentions schema loading.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
kalnik-a-a: Moved the wrapping into `ReportDataCacheService.createNewCachedReader`, around `prepareReportData` + `readReportDataBatch` only. `compose()` now runs outside the try, and the catch rethrows `BusinessViolationException` and `HttpException` untouched (same guard as `calculated-field-validator.service.ts`). Covered by the new `preserves a reader HTTP error` unit test with a structured `BadRequestException`.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
kalnik-a-a: The failure is now a `DataMartReadFailedException` carrying the original error as `cause`; `handleFailedReportRun` unwraps it, so `lastRunError` / `dataMartRun.errors` store the bare provider message exactly like the Google Sheets path, and the duplicate `execution failed` log is skipped for this case (the cache service already logged the original stack with report context). A unit test on `getData` asserts `markAsUnsuccessful(originalError)`. The changeset now mentions data requests as well.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
kalnik-a-a: Reordered: the rethrow for `BusinessViolationException` / `HttpException` now happens before `castError` and the log, so expected refusals stay unlogged as before. Unknown failures are logged once, inside `createNewCachedReader`, which runs once per pending operation — concurrent callers sharing the rejected promise no longer produce N stacks. The second and third log entries in the API service are suppressed for this error.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
kalnik-a-a: Intended: the destination secret already grants full read access to the report data, so the provider error text does not widen what a holder of the secret can learn. Report viewers who do not hold the secret see the same text through Looker Studio; we accept that for now — the alternative (generic text on getData, details only in Run History) is a follow-up if this turns out to be a problem. `toStorageReadError` was considered, but it yields a 424, which the connector does not parse (only 400 bodies reach the user), and it forwards the same provider message anyway.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
kalnik-a-a: Narrowed the try to the two reader calls, so `cacheRepository.save`, `readerResolver.resolve` and `resolvePrepareOptions` keep failing as 500 (unit test `does not present a cache persistence failure as a Data Mart read failure`). The cache-hit read path, `actualizeSchemaInEntity` and the IDP lookup are out of scope here — the reported case is the query failing while connecting. A route-scoped filter is the right follow-up if those paths show up in practice; it would also have to keep genuine bugs in the 5xx bucket, so I would rather do it separately.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.ts
kalnik-a-a: Introduced `DataMartReadFailedException` (`data-marts/errors/`) with `code: 'DATA_MART_READ_FAILED'` (asserted in the e2e) and the original error as a typed `cause` for internal unwrapping. Left `errorDetails` empty: the connector only reads `message`, and nothing else consumes this endpoint. The made-up `QUERY_EXECUTION_FAILED` code in the passthrough fixture is renamed to an obviously arbitrary one.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.spec.ts
kalnik-a-a: Rewritten: the API-service tests now assert `rejects.toBe(...)` for the passthrough cases, plus a `getData` test that checks `markAsUnsuccessful` receives the original error. The wrapping itself is pinned in `report-data-cache.service.spec.ts` via `toBeInstanceOf(DataMartReadFailedException)` + `toMatchObject({ code, message, cause })`.

apps/backend/src/data-marts/data-destination-types/looker-studio-connector/services/looker-studio-connector-api.service.spec.ts
kalnik-a-a: Removed — the API service no longer calls `castError`; the cache-service spec runs the real helper.

.changeset/6919-data-studio-schema-query-errors.md
kalnik-a-a: Updated to cover data requests. Run History text does not change anymore (the original message is stored), so it is not mentioned.

PR #1597, opened 2026-09-09
fix: keep report output controls consistent with the column selection and the schema

laskevych:
## Summary

Three related problems with report output controls shared one generic error:

- **Unchecking a column left its rules behind.** The aggregation, date bucket and sort set on the column stayed in the config, invisible on the row, and failed the next save and every scheduled run with `AGGREGATION_COLUMN_NOT_SELECTED` / `DATE_TRUNC_COLUMN_NOT_SELECTED` / `SORT_COLUMN_NOT_SELECTED`.
- **A column removed from the Data Mart schema produced misleading errors.** A sort or filter on it was reported as a disconnected column, but an aggregation or date bucket came back as "not selected" or as a type-unknown refusal, pointing at a column that no longer exists. On a scheduled run only the generic message reached run history.
- **Sort was limited to selected columns while filters were not**, even though an ungrouped query orders by any source column without projecting it.

## Changes

**Backend**
- Every output-control rule on a column missing from the schema now feeds the disconnected-columns diagnosis, which names the column and how to repair it. The advice text covers filter, sort, aggregation and date bucket rules.
- Sort validation follows the query shape the builders give the report: an ungrouped query with an explicit projection may sort by any known column; a grouped one (aggregations, date buckets, a Unique Count, or an aggregate-level calculated field selected or filtered on) keeps sorting limited to the selected columns. A calculated field is sortable only while selected, because it renders as a SELECT alias and only a selected one is planned.
- The run paths (blending decision and the flat composer) drop a sort on a column missing from the schema with a warning instead of failing the run, on the same reasoning the stale Unique Count sort is dropped by. Save paths still reject it. A new `known-output-columns.util.ts` is the one collector of known output columns for the validator and both run paths.
- The "Output controls validation failed" message now carries a summary of the failed rules and their columns, so run history, MCP clients and the Google Sheets extension show the reason. The MCP error mapper wording for not-selected fields reflects the sort rule.

**Web**
- Every way a column leaves the selection (row checkboxes, "Select all") goes through one path that removes the aggregation and date bucket on it, and the sort rule when the report still groups after the pruning or the column is a calculated field. Filters and slices stay. The change is a user edit, so the form dirties.
- The "Add sort by" menu and the disconnected-controls badge follow the same shape rule the backend validates.
- The error toast no longer repeats the validation summary the server message now carries.

**Docs / changeset**
- Output controls guide: sort rules by query shape, what unchecking a column removes, the disconnected-columns behaviour on save and on scheduled runs.
- Changeset: `owox` minor.

## Tests

- Backend: validator spec (sort by shape, disconnected aggregations and date buckets, calculated-field cases, message summary), new `known-output-columns.util.spec.ts`, run-path drop tests in the blended service and composer specs, error mapper and report use-case specs. `npm run build -w @owox/backend` passes.
- Web: new `output-controls-cleanup.test.ts`, picker tests for pruning on row and "Select all" unchecks (including a row-level calculated field), updated Unique Count sort expectations, output settings dropdown tests. Type-check and lint pass.
- E2E: the sort-on-unselected-column expectation is flipped for ungrouped reports, with new cases for a grouped report and an aggregation on a missing column (runs in CI).


apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -836,6 +837,17 @@ export function ReportColumnPicker({
     () => new Set(nativeFields.filter(f => f.calculated).map(f => f.name)),
     [nativeFields]
   );
+  // The AGGREGATE-level subset: selecting or filtering on one of these turns the query into a
+  // GROUP BY on its own, which decides what a sort may name (`isAggregatedShape`).
+  const aggregateCalculatedFieldNames = useMemo(
+    () =>
+      new Set(
+        nativeFields
+          .filter(f => f.calculated && !isRowLevelCalculatedField(f.calculated))
```
kalnik-a-a, line 853: **Criticality: MEDIUM / BLOCKING**

This derives the query shape from the persisted `calculated.level`, but that value is explicitly allowed to be stale: backend tests for `calculatedFieldLevelOf` cover a field stored as `column` that is effectively `metric` because its own formula or a dependency aggregates. In that state the picker treats the report as ungrouped and the new sort list below offers unselected native columns, while the backend recomputes the effective level, treats the report as grouped, and rejects the rule with `SORT_COLUMN_NOT_SELECTED`. Before this PR, the menu only exposed selected fields, so the invalid rule was not creatable through the UI. Please make the picker consume the backend-derived effective level rather than duplicating formula analysis in the browser, and cover the stale-level case.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -1156,41 +1189,84 @@ export class OutputControlsValidatorService {
...
+            // An ungrouped query with an explicit projection resolves ORDER BY against the source
+            // row — the flat builders qualify or quote the bare column, the blended one carries
+            // every sorted column into its CTE (`referencedColumns`) — so a sort on a column the
+            // report does not print is valid SQL, exactly like a filter on one. Two names stay
+            // out: a calculated field that is not selected (it renders as a SELECT alias, and only
+            // a selected one is planned, so the resolver would fall back to `src.<name>`, a column
+            // the warehouse does not have) and a Unique Count output name (the metric is off in
+            // this shape, so the alias does not exist; a real field owning the name IS that
+            // field). A name absent from the schema is still reported below as disconnected.
+            const sortableColumns = new Set(knownOutputColumns);
```
kalnik-a-a: **Criticality: MEDIUM / BLOCKING**

This new ungrouped branch permits every known unselected physical field. A real main-table column literally named `Unique Count` therefore remains sortable when the synthetic metric is disabled. On the blended path, however, `AbstractBlendedQueryBuilder` removes `Unique Count` from `referencedColumns` whenever it is not selected (`columnSet`), because it assumes the name is the synthetic outer alias. Save and validation succeed, then the emitted outer sort references `Unique Count` through the main CTE even though that column was omitted. The base validator rejected this configuration, so this path becomes reachable in this PR. Please preserve physical-column ownership when stripping synthetic aliases, or reject this specific unselected sort, and add a generated-SQL regression test.

docs/getting-started/setup-guide/output-controls.md
```diff
@@ -175,18 +177,22 @@ Output controls reference columns by name. Rename or remove a column in the Data
...
 
 ### Validation error on save
 
-A filter, slice, or sort rule may still point to a disconnected column. Save the report, and OWOX blocks it with **"Output controls validation failed"**. Open the report, remove the affected rules from the Filters, Slices, or Sort sections, then save again.
+A filter, slice, sort, aggregation, or date bucket rule may still point to a disconnected column. Save the report, and OWOX blocks it with a **Disconnected columns** error that names the columns. Open the report, remove the affected rules from the Filters, Slices, Sort, or Aggregations sections, then save again.
+
+A scheduled run of such a report drops a sort rule on a disconnected column and continues — the order of the rows changes, their values do not. An aggregation or date bucket on a disconnected column fails the run with the same error, since it would change the values.
```
kalnik-a-a: **Criticality: LOW / NON-BLOCKING**

With `LIMIT` configured, dropping the stale `ORDER BY` can change which rows are returned, not only their order—for example, a former top-10 becomes an arbitrary 10. Keeping the limit is the safer fallback because clearing it could trigger an unbounded export, so this is not a request to change runtime behavior. Please narrow the wording, and ideally pin it with a focused test, so the documented contract acknowledges the row-set change.

docs/getting-started/setup-guide/output-controls.md
```diff
@@ -175,18 +177,22 @@ Output controls reference columns by name. Rename or remove a column in the Data
 
 Open **Edit report**. The column picker groups the missing columns under a red **Disconnected columns** label with a ⚠ icon. Hover the icon to see:
 
-> _They are missing from the current Data Mart output schema. Uncheck them to remove them from the report, or contact your analyst to restore the schema._
+> _They are missing from the current Data Mart output schema. Uncheck them and remove any filter, sort, aggregation or date bucket rule that references them, or contact your analyst to restore the schema._
 
 You have two options:
 
-- **Uncheck the disconnected columns** — this removes them from the report selection only. It does not clear filter, slice, or sort rules that reference them. Open the Filters, Slices, and Sort sections, delete those rules too, then save.
+- **Uncheck the disconnected columns** — this removes them from the report selection together with the aggregation, date bucket, and sort rules set on them. Filter and slice rules that reference them stay: open the Filters and Slices sections, delete those rules too, then save.
```
kalnik-a-a, line 184: **Criticality: LOW / NON-BLOCKING**

This statement is unconditional, but `pruneRulesForDeselectedColumns` preserves the sort when deselecting a non-calculated column from a plain or ungrouped report. If that column is disconnected (`ghost` no longer exists in the schema), unchecking it leaves the stale sort and the next save still fails. This behavior already existed before the PR; the new issue is that this changed paragraph now promises a cleanup that does not occur. Please describe the actual conditional behavior, or, as a separate product choice, teach the helper to distinguish disconnected fields from still-known unselected fields.

kalnik-a-a (changes requested):
**Review decision: REQUEST CHANGES**

**Overall criticality: MEDIUM / BLOCKING**

I found two PR-introduced blocking regressions:

1. The new sort menu derives query shape from a persisted calculated-field level that the backend explicitly treats as a stale cache, so the UI can create a sort that save validation rejects.
2. The new unselected-sort path accepts a physical main-table column named `Unique Count`, but the blended builder strips that name from its CTE references and emits invalid SQL.

I also left two **LOW / NON-BLOCKING** notes about documenting the accepted `LIMIT` degradation accurately and aligning the disconnected-column cleanup documentation with the helper's actual conditional behavior. The inline comments include the concrete failure paths and minimal remediation options.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
laskevych: Fixed in 253087e2d. The picker no longer trusts the persisted level: `BlendableSchemaService.computeBlendableSchema` now hands out every calculated field with its effective level from `calculatedFieldLevelOf` (copied onto the DTO, the entity is left untouched), so the sort menu, the AGG menu and the validator all read the same derivation. Covered by a new spec: a field stored as `column` whose formula aggregates comes back as `metric`, and the stored value stays `column`.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
laskevych: Fixed in 253087e2d, taking the reject option: the ungrouped branch now keeps every Unique Count output name (the main label and each available source's `<alias>__unique_count`) out of the sortable set unless the column is selected, so this configuration is refused with `SORT_COLUMN_NOT_SELECTED` as before. The picker mirrors it (the column is neither offered in "Add sort by" nor counted as resolved). Validator spec covers both the main label and a joined-source name; since the rule can no longer be saved, there is no generated-SQL path to pin — the ownership-preserving alternative in the builder felt like a bigger change than this PR should carry.

docs/getting-started/setup-guide/output-controls.md
laskevych: Reworded in 253087e2d: the docs, the changeset and the code comments now say the limit is kept, so under a limit the delivered rows may differ from the ones the sort used to pick. The composer spec for the drop asserts the limit travels through untouched.

docs/getting-started/setup-guide/output-controls.md
laskevych: Good catch — made the helper match the sentence rather than the other way round (253087e2d): `pruneRulesForDeselectedColumns` now takes the known column names, and a sort on an unchecked column the schema no longer offers is removed whatever the report's shape. Unit test on the helper plus a picker test for unchecking a disconnected row on a plain report.

laskevych:
Follow-up round after a self-review of the branch (commit f5f0c2560):

- Stale-sort degradation is now opt-in (`degradeStaleSort`) and only the paths that execute a saved report pass it: a run, the Looker Studio cache fill, and the report stream. Ad-hoc MCP/HTTP queries, save dry runs, the Generated SQL preview and the output-schema describe get the validator's disconnected error, as on `main`. The blending decision carries the degraded sort; the composer consumes it instead of pruning again, the run record snapshots the applied sort, and the Totals validation no longer re-checks a sort the totals plan never renders.
- Validator: a null projection under a main Unique Count is treated as metrics-only for the sort check; the Unique Count output names come from one shared helper. The MCP error mapper stops prefixing the summary-carrying message and names the unknown filter columns itself.
- Picker: one sort-resolution context feeds the sort menu, the disconnected badge and every writer. A selected sort on an excluded source's column is no longer flagged; a joined calculated field never resolves. Writers prune only the sort rules their own edit broke (grouping the report, toggling a Unique Count off, unchecking a column) and leave already-orphaned rules visible. Removing an aggregation, from the column checkbox, the row menu or the panel, takes the HAVING filter bound to it. Tooltip wording matches the backend; the changeset has its title; docs describe which paths degrade.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
kalnik-a-a: **Criticality: MEDIUM / BLOCKING**

The non-empty projection case is fixed, but an explicit `columnConfig: []` still bypasses the new guard because `hasColumnConfig` is based on `.length`. `ReportColumnConfigSchema` explicitly accepts an empty array.

With an empty projection, a post-join filter, and a sort on a real main field named `Unique Count`, validation falls back to the implicit-native branch and accepts the rule. The blended builder then removes `Unique Count` from the main CTE while still emitting:

```sql
ORDER BY main.`Unique Count`
```

This fails at the warehouse.

Please distinguish an absent projection (`null`) from an explicit empty projection (`[]`) for this validation and add a focused regression test that forces the blended path. Once the validator rejects this state, a generated-SQL test is not necessary.

kalnik-a-a (changes requested):
**Review decision: REQUEST CHANGES**

**Remaining criticality: MEDIUM / BLOCKING**

Thank you for the follow-up changes. I re-reviewed the updates through `f5f0c256`. The effective calculated-field level, LIMIT documentation and coverage, and disconnected-sort cleanup are now addressed.

The physical `Unique Count` guard is still incomplete for an explicit empty projection (`columnConfig: []`). I added the concrete reproduction to the existing thread. This valid API state is accepted, but the resulting blended SQL orders by a column omitted from the main CTE.

I am keeping this review at **Request changes** until that remaining path is covered. I will re-check the final result after the fix and after the branch is updated against `main`.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
laskevych: Fixed in c2f2a3962: the sort check now keys the explicit-projection branch on `columnConfig != null` rather than on `.length`, so an explicit `[]` is treated as what the run path makes of it — an explicit projection that prints nothing. With nothing projected no source column resolves (the verdict the empty list had before this PR), so a sort on a real main field named `Unique Count` — or on any native column — under `columnConfig: []` with a post-join filter is refused with `SORT_COLUMN_NOT_SELECTED`. Focused spec added for both columns with a post-join filter on a joined column, i.e. the shape that forces the blended path. The branch is also updated against `main` (merge 9c7fee933).

kalnik-a-a (approved):
**Review decision: APPROVE**

**Remaining criticality: NONE**

Thank you for the thorough follow-up and for addressing the review feedback. I re-checked the latest changes, including the explicit empty-projection case, and I have no remaining blocking concerns.

PR #1599, opened 2026-09-09
feat(data-marts): allow editing draft joins

kalnik-a-a:
## Summary

- allow the relationship editor to include configured draft Data Mart targets
- expose draft output fields, aliases, and join-specific descriptions before publication
- keep report, formula, and runtime consumers on the published-only default
- add a release changeset for Fibery work item 6905

## Testing

- `npm test -w @owox/backend -- --runInBand`
- `npm test -w @owox/web`
- `npm run lint -w @owox/web`
- `npm exec -- prettier --check <changed files>`
- `git diff --check`

## Notes

The default parallel backend test run hit the existing wall-clock threshold in `formula-analyzer.spec.ts` under CPU contention. The test passed in isolation, and the full backend suite passed serially (613 suites, 9435 tests).

### DIFF
--- a/apps/backend/src/data-marts/controllers/data-mart.controller.ts
+++ b/apps/backend/src/data-marts/controllers/data-mart.controller.ts
@@ -1,4 +1,15 @@
-import { Body, Controller, Delete, Get, HttpCode, Param, Post, Put, Query } from '@nestjs/common';
+import {
+  Body,
+  Controller,
+  Delete,
+  Get,
+  HttpCode,
+  Param,
+  ParseBoolPipe,
+  Post,
+  Put,
+  Query,
+} from '@nestjs/common';
 import { ApiTags } from '@nestjs/swagger';
 import { Auth, AuthContext, AuthorizationContext, Role, Strategy, ViewOnlySafe } from '../../idp';
 import { BlendableSchemaDto } from '../dto/domain/blendable-schema.dto';
@@ -446,9 +457,15 @@ export class DataMartController {
   @GetBlendableSchemaSpec()
   async getBlendableSchema(
     @AuthContext() context: AuthorizationContext,
-    @Param('id') dataMartId: string
+    @Param('id') dataMartId: string,
+    @Query('includeDraftTargets', new ParseBoolPipe({ optional: true }))
+    includeDraftTargets = false
   ): Promise<BlendableSchemaDto> {
-    const command = this.mapper.toGetBlendableSchemaCommand(dataMartId, context);
+    const command = this.mapper.toGetBlendableSchemaCommand(
+      dataMartId,
+      context,
+      includeDraftTargets
+    );
     return this.getBlendableSchemaService.run(command);
   }
 }
--- a/apps/backend/src/data-marts/services/blendable-schema.service.ts
+++ b/apps/backend/src/data-marts/services/blendable-schema.service.ts
@@ -174,10 +174,34 @@ export class BlendableSchemaService {
 
     await this.applyReportingAccess(availableSources, projectId, accessor);
 
+    let issueSources = availableSources;
+    let issueFields = blendedFields;
+    if (options.includeDraftTargets === true) {
+      const publishedSources: AvailableSourceDto[] = [];
+      this.collectBlendedFields({
+        sourceId: dataMartId,
+        parentPath: '',
+        sourcesByPath,
+        relationshipsBySource,
+        result: [],
+        availableSources: publishedSources,
+        branchDmIds,
+        depth: 1,
+        storageType: dataMart.storage.type,
+        includeDraftTargets: false,
+      });
+      const publishedPaths = new Set(publishedSources.map(source => source.aliasPath));
+      issueSources = availableSources.filter(source => publishedPaths.has(source.aliasPath));
+      issueFields = blendedFields.filter(field => publishedPaths.has(field.aliasPath));
+    }
+
     const rawSchemaFields = dataMart.schema?.fields ?? [];
-    // The join tree this very call just walked — so a formula's joined reference is checked against
-    // the SAME tree the report builder will route it through, on the one payload the picker reads.
-    const joinedReferenceIndex = buildJoinedReferenceIndex({ availableSources, blendedFields });
+    // Draft fields may be returned for relationship editing, but formulas remain report-safe: the
+    // report builder cannot route through a draft target, so neither may this health verdict.
+    const joinedReferenceIndex = buildJoinedReferenceIndex({
+      availableSources: issueSources,
+      blendedFields: issueFields,
+    });
 
     return {
       nativeFields,
--- a/apps/backend/src/data-marts/services/blendable-schema.service.ts
+++ b/apps/backend/src/data-marts/services/blendable-schema.service.ts
@@ -119,6 +119,7 @@ interface CollectContext {
   branchDmIds: Set<string>;
   depth: number;
   storageType: DataStorageType;
+  includeDraftTargets: boolean;
 }
 
 @Injectable()
@@ -132,7 +133,8 @@ export class BlendableSchemaService {
   async computeBlendableSchema(
     dataMartId: string,
     projectId: string,
-    accessor: BlendableSchemaAccessor
+    accessor: BlendableSchemaAccessor,
+    options: { includeDraftTargets?: boolean } = {}
   ): Promise<BlendableSchemaDto> {
     const dataMart = await this.dataMartService.getByIdAndProjectId(dataMartId, projectId);
     const nativeFields = (dataMart.schema?.fields ?? []).filter(
@@ -167,6 +169,7 @@ export class BlendableSchemaService {
       branchDmIds,
       depth: 1,
       storageType: dataMart.storage.type,
+      includeDraftTargets: options.includeDraftTargets === true,
     });
 
     await this.applyReportingAccess(availableSources, projectId, accessor);
@@ -260,9 +263,10 @@ export class BlendableSchemaService {
         );
       }
 
-      // Reports cannot join against an unfinalized schema, so a draft target — and any
-      // descendants reachable only through it — must not surface in the picker.
-      if (rel.targetDataMart.status !== DataMartStatus.PUBLISHED) continue;
+      // Reports cannot join against an unfinalized schema. The relationship editor opts in so an
+      // analyst can configure a draft from its saved output schema before publishing it.
+      if (!ctx.includeDraftTargets && rel.targetDataMart.status !== DataMartStatus.PUBLISHED)
+        continue;
 
       if (ctx.branchDmIds.has(rel.targetDataMart.id)) continue;
 
--- a/apps/backend/src/data-marts/use-cases/get-blendable-schema.service.spec.ts
+++ b/apps/backend/src/data-marts/use-cases/get-blendable-schema.service.spec.ts
@@ -31,7 +31,7 @@ describe('GetBlendableSchemaService', () => {
   it('returns blendable schema when user has SEE access on DataMart', async () => {
     const { service, blendableSchemaService, accessDecisionService } = createService(true);
 
-    const command = new GetBlendableSchemaCommand('dm-1', 'proj-1', 'user-1', ['viewer']);
+    const command = new GetBlendableSchemaCommand('dm-1', 'proj-1', 'user-1', ['viewer'], true);
 
     const result = await service.run(command);
 
@@ -43,10 +43,15 @@ describe('GetBlendableSchemaService', () => {
       Action.SEE,
       'proj-1'
     );
-    expect(blendableSchemaService.computeBlendableSchema).toHaveBeenCalledWith('dm-1', 'proj-1', {
-      userId: 'user-1',
-      roles: ['viewer'],
-    });
+    expect(blendableSchemaService.computeBlendableSchema).toHaveBeenCalledWith(
+      'dm-1',
+      'proj-1',
+      {
+        userId: 'user-1',
+        roles: ['viewer'],
+      },
+      { includeDraftTargets: true }
+    );
     expect(result).toBe(blendableSchema);
   });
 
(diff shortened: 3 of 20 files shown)


apps/web/src/features/data-marts/shared/hooks/useBlendableSchema.ts
```diff
@@ -6,25 +6,29 @@ import { BLENDABLE_SCHEMA_QUERY_KEY } from './blendable-schema-query-key';
...
  * a request that is a detail of one card is noise.
  */
-export function useBlendableSchema(dataMartId: string) {
+export function useBlendableSchema(
+  dataMartId: string,
+  { includeDraftTargets = false }: { includeDraftTargets?: boolean } = {}
+) {
   return useQuery({
-    queryKey: [BLENDABLE_SCHEMA_QUERY_KEY, dataMartId],
+    queryKey: includeDraftTargets
```
laskevych, line 23: 🟡 Important (non-blocking): Partitioning the key means the two cards on the Data Setup page (`DataMartSchemaSettings` on the default key, `DataMartRelationshipsContent` on the draft key) no longer dedupe. With `staleTime: 0` every tab visit now issues two `blendable-schema` requests, and each of the 7 `invalidateBlendableSchema` call sites (alias/description/override saves, relationship create/update/delete, schema save) triggers two concurrent backend walks (`findByStorageId` over all storage relationships + `canAccessMany`). The old hook comment explicitly promised "one fetch, one copy".

I understand the isolation rationale (report readers must never see drafts, and `calculatedFieldIssues` differs between variants, so a pure client-side filter would not be equivalent). If the doubled request is an accepted price, please say so in the hook doc; a cheaper long-term option is stamping draft sources on the backend, building `calculatedFieldIssues` from the published subset, and deriving the report-safe view via React Query `select` from one entry.

apps/backend/src/data-marts/services/blendable-schema.service.ts
```diff
@@ -167,6 +169,7 @@ export class BlendableSchemaService {
       branchDmIds,
       depth: 1,
       storageType: dataMart.storage.type,
+      includeDraftTargets: options.includeDraftTargets === true,
```
laskevych, line 172: 💡 Suggestion: With `includeDraftTargets: true` the `joinedReferenceIndex` below (line 180) is built from the draft-inclusive tree, so a root formula referencing `draftAlias.field` is reported as healthy in this payload while `calculated-field-validator`, `output-controls-validator` and the SQL composer (all default) report it broken. Nothing reads `calculatedFieldIssues` from the draft variant today, so this is latent, but the comment above the index ("checked against the SAME tree the report builder will route it through") is now false for that variant.

Consider recording draft alias paths in the collect context and building the index from sources not under a draft path, or document that this field is report-view only.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
```diff
@@ -215,9 +215,12 @@ export function DataMartRelationshipsContent({
     void loadRelationships();
   }, [loadRelationships]);
 
-  // Shared with the other cards on this page (the schema editor's formula autocomplete reads the
-  // same entry): one fetch, one copy, and `invalidateBlendableSchema` below refreshes all of them.
-  const { data: blendableSchema } = useBlendableSchema(dataMartId);
+  // This editor needs draft targets so their saved output schemas and per-join overrides remain
+  // configurable before publish. The query key keeps this edit-only payload separate from the
+  // report/formula readers, while `invalidateBlendableSchema` still refreshes both variants.
+  const { data: blendableSchema } = useBlendableSchema(dataMartId, {
```
laskevych, line 221: 💡 Suggestion (affects `RelationshipAccordionItem.tsx:448`, outside the diff): A draft target whose schema was never actualized now arrives with `fieldCount: 0` and `fields: []`, so `source` is non-null and the Fields tab renders `SourceFieldsTable`, whose only empty state is "No fields match the current filter" (`SourceFieldsTable.tsx:402`) even with no filter active. Join conditions can be saved against such a target (a missing join field is only a warning in `create-data-mart-relationship.service.ts`).

A short empty-state line like "This draft has no saved output schema yet" would explain the "0 fields" badge. Related: the "Allow for reporting" switch now shows the real saved state next to the Draft badge, which is a correction, but a caption noting the join will not reach reports until publish would help.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
```diff
@@ -215,9 +215,12 @@ export function DataMartRelationshipsContent({
...
+    includeDraftTargets: true,
+  });
```
laskevych, line 223: 📝 Docs: `connectedFieldCounts` and `canvasFieldsByAliasPath` below now derive from the draft-inclusive payload, so the relationship canvas shows field-count badges and Detailed-ERD rows for draft nodes and for published nodes behind a draft (they previously showed "0 fields" and no rows). Draft/Blocked labels still come from the graph endpoint, so nothing breaks, but this visible change is not mentioned in the PR description or changeset. Worth a line there.

apps/backend/src/data-marts/controllers/data-mart.controller.ts
```diff
@@ -446,9 +457,15 @@ export class DataMartController {
   @GetBlendableSchemaSpec()
   async getBlendableSchema(
     @AuthContext() context: AuthorizationContext,
-    @Param('id') dataMartId: string
+    @Param('id') dataMartId: string,
+    @Query('includeDraftTargets', new ParseBoolPipe({ optional: true }))
```
laskevych, line 461: 🔵 Question: `includeDraftTargets` is caller-controlled on a `Role.viewer` endpoint that checks only `SEE` on the root, and `applyReportingAccess` only flags `isAccessibleForReporting` without stripping `blendedFields`. So a viewer with SEE on DM A and no access to draft DM B can now read B's full field list (names, types, descriptions, aliases).

This widens an existing exposure rather than creating a new one (unshared published targets are already returned the same way, and the MCP facade filters them while this HTTP endpoint does not), but it is the first HTTP surface that returns a draft's complete schema; the graph endpoint exposed only join-key names. Is that acceptable, or should the flag require EDIT on the root? Stripping `blendedFields` for inaccessible sources server-side would close both the old and the new case.

apps/backend/src/data-marts/controllers/data-mart.controller.ts
```diff
@@ -446,9 +457,15 @@ export class DataMartController {
...
+    includeDraftTargets = false
```
laskevych, line 462: 🔵 Question: This is the first `ParseBoolPipe` in `apps/backend`. The sibling flags `includeRevoked` and `includeUninstalled` use `@Query('x') x?: string` + `x === 'true'`, which silently treats anything else as false. `ParseBoolPipe({ optional: true })` only short-circuits on nil, so `?includeDraftTargets=` (empty), `=1` or `=TRUE` return 400. Not a problem for the web client (it sends literal `'true'`/`'false'`), just checking the stricter behavior on a public endpoint is intentional.

apps/backend/src/data-marts/use-cases/get-blendable-schema.service.spec.ts
```diff
@@ -31,7 +31,7 @@ describe('GetBlendableSchemaService', () => {
   it('returns blendable schema when user has SEE access on DataMart', async () => {
     const { service, blendableSchemaService, accessDecisionService } = createService(true);
 
-    const command = new GetBlendableSchemaCommand('dm-1', 'proj-1', 'user-1', ['viewer']);
+    const command = new GetBlendableSchemaCommand('dm-1', 'proj-1', 'user-1', ['viewer'], true);
```
laskevych, line 34: 🧪 Test: This was the only test asserting `computeBlendableSchema`'s arguments and it now passes `true`, so nothing asserts the default command forwards `{ includeDraftTargets: false }`. A regression that hard-codes `true` in `run()` would keep the suite green while the report picker and formula editor start receiving drafts. Suggest an `it.each([true, false])` forwarding case.

Likewise in `blendable-schema.service.spec.ts:190`, the new test covers a single A→B(draft) edge; reusing the existing A→B(draft)→C fixture from the skip test and asserting `availableSources` alias paths equal `['b', 'b.c']` would pin the transitive behavior the flag changes most.

apps/web/src/features/data-marts/shared/services/data-mart-relationship.service.ts
```diff
@@ -80,13 +80,15 @@ class DataMartRelationshipService extends ApiService {
...
+   * @param params Optional view controls; draft targets are excluded by default.
    * @param config Optional axios config (e.g. `skipLoadingIndicator` for silent refreshes)
    */
   async getBlendableSchema(
     dataMartId: string,
+    params?: { includeDraftTargets?: boolean },
     config?: AxiosRequestConfig
   ): Promise<BlendableSchema> {
-    return this.get<BlendableSchema>(`/${dataMartId}/blendable-schema`, undefined, config);
+    return this.get<BlendableSchema>(`/${dataMartId}/blendable-schema`, params, config);
```
laskevych, line 91: 🧹 Nit: The hook always passes `{ includeDraftTargets }`, so every default reader now sends `?includeDraftTargets=false`. Sibling optional-boolean services (`plugins.service` `includeUninstalled`, `api-keys.service` `includeRevoked`) omit the param when false and send `'true'` otherwise. Mapping to that shape here keeps the wire identical to today for the report and formula readers.

laskevych (approved):
✅ Approve. The change is opt-in and fail-closed: every report, validator, MCP and HTTP-data consumer still calls `computeBlendableSchema` with the published-only default, prefix invalidation covers both cache variants, and the changeset is correct. Verified locally: the three touched backend specs (106 tests) and the touched web tests (220) pass.

All inline comments are non-blocking. The two I would most like addressed before merge are the doubled request on the Data Setup page (either accepted and documented, or derived from one entry) and the test for default forwarding.

🧹 Nit (`apps/web/src/features/data-marts/shared/hooks/blendable-schema-query-key.ts:3`, outside the diff): the doc comment still says every component fetching blendable schema "uses the same React Query cache entry for a given data mart", which this PR makes false. Please update it to mention the draft-inclusive variant.

apps/web/src/features/data-marts/shared/hooks/useBlendableSchema.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — DOCUMENTED; NO BEHAVIOR CHANGE.**

The two cache entries are intentional: default readers require the report-safe payload, while relationship editing opts into draft targets. The hook now documents the resulting two requests and that prefix invalidation refreshes both variants. Consolidation should wait for profiling and an explicit backend status/filtering contract.

apps/backend/src/data-marts/services/blendable-schema.service.ts
kalnik-a-a: **Criticality: MEDIUM / NON-BLOCKING — FIXED.**

The draft-inclusive response still exposes draft sources and fields for relationship editing, while calculatedFieldIssues is now derived from the published-only relationship tree used by validation and SQL composition. A mixed-tree regression verifies that published references remain valid while references through a draft path remain unavailable for reporting.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — FIXED.**

The table now distinguishes a source with no available fields from a filtered-empty result and shows a source-specific empty message. The toolbar and Output Alias controls remain available, and a component regression test covers the zero-field case.

apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — NO CODE CHANGE.**

Showing saved draft output fields in field-count badges and Detailed ERD rows is intentional and is part of enabling relationship configuration before publication. The existing changeset already describes that user-visible outcome, so a second release-note entry is not needed.

apps/backend/src/data-marts/controllers/data-mart.controller.ts
kalnik-a-a: **Criticality: NONE / NO CHANGE.**

Confirmed product contract: fields from an already configured relationship remain visible in join settings even when the viewer lacks independent access to the target Data Mart, consistently for published and draft targets. isAccessibleForReporting restricts use in report flows; it is not a schema-visibility gate. The endpoint only traverses saved, configured relationships under the visible root Data Mart, so requiring EDIT or redacting these fields would break the intended behavior.

apps/backend/src/data-marts/controllers/data-mart.controller.ts
kalnik-a-a: **Criticality: NONE / NO CHANGE.**

Strict parsing is intentional for this new Boolean query contract: omitted, true, and false are accepted, while malformed values return 400. There is no legacy compatibility requirement for this new parameter, and the web client sends valid Boolean values.

apps/backend/src/data-marts/use-cases/get-blendable-schema.service.spec.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — FIXED.**

Added coverage for the default false forwarding path and extended the draft-target regression through A → B(draft) → C, asserting both alias paths and traversal depths. The existing service-level tests are the narrowest sufficient coverage; no additional E2E suite is needed.

apps/web/src/features/data-marts/shared/services/data-mart-relationship.service.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — FIXED.**

The report-safe default hook call now omits request params, while the opt-in path sends only { includeDraftTargets: true }. Hook tests cover both wire shapes, and the relationship editor true variant remains covered.

PR #1601, opened 2026-09-09
fix(data-marts): keep output schema field connection statuses server-owned

kalnik-a-a:
Changeset: `.changeset/6935-output-schema-field-connection-status.md` (`owox`, `minor`)

## Summary

A field's connection status in the output schema (`CONNECTED`, `CONNECTED_WITH_DEFINITION_MISMATCH`, `DISCONNECTED`) is derived by schema actualization against the storage. Until now `PUT /api/data-marts/:id/schema` accepted whatever status the client sent, so an API user or a plugin could save a column as `CONNECTED` before the storage had verified it and immediately use it in reports. The status is now read-only for the API: the save derives it on the server, older clients that still send it keep working, and the web app no longer sends it.

### Backend

- `PUT /api/data-marts/:id/schema` ignores the client status and applies server-owned statuses to the parsed schema: a native field keeps the last persisted status of its namesake, a new native field starts `DISCONNECTED` until actualization verifies it, a calculated field is always `CONNECTED`. Nested `RECORD` fields are handled recursively.
- The zod field schema gives `status` a `DISCONNECTED` default, so the field is optional on the wire for all five storage types. A legacy row without a status now loads instead of failing the entity transformer.
- `UpdateDataMartSchemaApiDto` and the command carry a status-less `DataMartSchemaUpdate` type derived from the zod input type; the Swagger description spells out the rule and points at `POST /data-marts/{id}/schema-actualize-triggers` for refreshing statuses.

### Web

- `dataMartService.updateDataMartSchema` strips `status` from every field, nested ones included, before the request; the request DTO type excludes it. The feature type keeps `status` for display.

### Tests and fixtures

- Unit: server-owned derivation (persisted carry-forward, nested carry-forward, new native and calculated fields, a native field submitted under a persisted calculated field's name), plus the web strip.
- API e2e: a client-provided `CONNECTED` is ignored (nested included) and a field without status is accepted.
- The e2e fixtures used `PUT /schema` to declare native columns as `CONNECTED` on Data Marts that have no warehouse behind them, which is exactly what the endpoint no longer does. They now go through `setDataMartSchema` (`@owox/test-utils`), which seeds the schema with the given statuses straight into the test database, the same bypass both e2e layers already use for storage config, and then saves it through the API so calculated fields still get the real validation. The Playwright `setSchema` fixture does the same against the sqlite test file. No product code changed for this.

## Compatibility

- Wire format: unchanged. Requests with `status` return 200; the value is ignored.
- Semantics: a client can no longer make a column usable by declaring it connected. After a manual save, run schema actualization to get real statuses; the web app already does this after Save.
- Rolling update: a browser that loaded the new web bundle can hit a pod on the previous release during the rollout of `owox-data-marts-webserver`. That pod still requires `status` and answers 400 on schema save until the rollout completes. Self-heals on retry, no data loss in either direction.
- Other clients checked: Google Sheets extension (reads only, responses still carry `status`), `owox-mcp`, `odm-looker-studio`, backend MCP tools, CLI. Nothing else writes a schema.

## Verification

| Check | Result |
| --- | --- |
| Backend unit, full | 613 suites, 9445 tests passed |
| Web unit, full | 261 files, 2422 tests passed |
| API e2e, full (`jest --config test/jest-e2e.json`) | 54 suites, 717 tests passed |
| Playwright: `datamart-calculated-field-chips`, `datamart-calculated-field-formula`, `datamart-row-level-calculated-field` | 25 passed |
| Red-green of the new unit tests | the two new carry-forward tests fail with the recursion and the calculated guard removed, pass with them restored |
| tsc (backend build config, web), eslint, prettier, markdownlint | clean |

Two earlier full API e2e runs each had one transport-level failure in an unrelated suite (503, `socket hang up`), green in isolation and in the other runs.

## Notes for reviewers

- The changeset filename uses Fibery Work Item `6935`; please confirm it is the owning item.
- Pre-existing full `tsc -p tsconfig.json` errors in unrelated `*.spec.ts` files are untouched (same count before and after).

### DIFF
--- a/apps/backend/src/data-marts/data-storage-types/data-mart-schema.type.ts
+++ b/apps/backend/src/data-marts/data-storage-types/data-mart-schema.type.ts
@@ -16,22 +16,4 @@ export const DataMartSchemaSchema = z.discriminatedUnion('type', [
 export type DataMartSchema = z.infer<typeof DataMartSchemaSchema>;
 export type DataMartSchemaField = DataMartSchema['fields'][number];
 
-type KnownKey<Key> = string extends Key
-  ? never
-  : number extends Key
-    ? never
-    : symbol extends Key
-      ? never
-      : Key;
-
-type WithoutConnectionStatus<T> = T extends readonly unknown[]
-  ? { [Index in keyof T]: WithoutConnectionStatus<T[Index]> }
-  : T extends object
-    ? {
-        [Key in keyof T as Key extends 'status' ? never : KnownKey<Key>]: WithoutConnectionStatus<
-          T[Key]
-        >;
-      }
-    : T;
-
-export type DataMartSchemaUpdate = WithoutConnectionStatus<z.input<typeof DataMartSchemaSchema>>;
+export type DataMartSchemaUpdate = z.input<typeof DataMartSchemaSchema>;
--- a/apps/backend/src/data-marts/data-storage-types/data-mart-schema.type.ts
+++ b/apps/backend/src/data-marts/data-storage-types/data-mart-schema.type.ts
@@ -15,3 +15,23 @@ export const DataMartSchemaSchema = z.discriminatedUnion('type', [
 
 export type DataMartSchema = z.infer<typeof DataMartSchemaSchema>;
 export type DataMartSchemaField = DataMartSchema['fields'][number];
+
+type KnownKey<Key> = string extends Key
+  ? never
+  : number extends Key
+    ? never
+    : symbol extends Key
+      ? never
+      : Key;
+
+type WithoutConnectionStatus<T> = T extends readonly unknown[]
+  ? { [Index in keyof T]: WithoutConnectionStatus<T[Index]> }
+  : T extends object
+    ? {
+        [Key in keyof T as Key extends 'status' ? never : KnownKey<Key>]: WithoutConnectionStatus<
+          T[Key]
+        >;
+      }
+    : T;
+
+export type DataMartSchemaUpdate = WithoutConnectionStatus<z.input<typeof DataMartSchemaSchema>>;
--- a/apps/backend/src/data-marts/dto/presentation/update-data-mart-schema-api.dto.ts
+++ b/apps/backend/src/data-marts/dto/presentation/update-data-mart-schema-api.dto.ts
@@ -8,7 +8,7 @@ export class UpdateDataMartSchemaApiDto {
     type: () => Object,
     required: true,
     description:
-      'Updated schema of the data mart. Field connection statuses are server-owned: omit them, and any value sent by an older client is ignored. The saved schema carries the last status the server derived; a new native field starts DISCONNECTED until schema actualization (POST /data-marts/{id}/schema-actualize-triggers) verifies it against the storage.',
+      'Updated schema of the data mart. Field connection statuses are server-owned: omit them, and any value sent by an older client is ignored. The saved schema carries the last status the server derived; a new native field starts DISCONNECTED until schema actualization (POST /api/data-marts/{dataMartId}/schema-actualize-triggers) verifies it against the storage.',
   })
   @IsNotEmptyObject()
   schema: DataMartSchemaUpdate;
--- a/apps/backend/src/data-marts/dto/presentation/update-data-mart-schema-api.dto.ts
+++ b/apps/backend/src/data-marts/dto/presentation/update-data-mart-schema-api.dto.ts
@@ -1,16 +1,17 @@
 import { ApiProperty } from '@nestjs/swagger';
 import { IsNotEmptyObject } from 'class-validator';
-import { DataMartSchema } from '../../data-storage-types/data-mart-schema.type';
+import type { DataMartSchemaUpdate } from '../../data-storage-types/data-mart-schema.type';
 import { DataMartResponseApiDto } from './data-mart-response-api.dto';
 
 export class UpdateDataMartSchemaApiDto {
   @ApiProperty({
     type: () => Object,
     required: true,
-    description: 'Updated schema of the data mart',
+    description:
+      'Updated schema of the data mart. Field connection statuses are server-owned: omit them, and any value sent by an older client is ignored. The saved schema carries the last status the server derived; a new native field starts DISCONNECTED until schema actualization (POST /data-marts/{id}/schema-actualize-triggers) verifies it against the storage.',
   })
   @IsNotEmptyObject()
-  schema: DataMartSchema;
+  schema: DataMartSchemaUpdate;
 }
 
 export class FormulaViolationApiDto {
--- a/apps/backend/src/data-marts/dto/presentation/update-data-mart-schema-api.dto.ts
+++ b/apps/backend/src/data-marts/dto/presentation/update-data-mart-schema-api.dto.ts
@@ -8,7 +8,7 @@ export class UpdateDataMartSchemaApiDto {
     type: () => Object,
     required: true,
     description:
-      'Updated schema of the data mart. Field connection statuses are server-owned: omit them, and any value sent by an older client is ignored. The saved schema carries the last status the server derived; a new native field starts DISCONNECTED until schema actualization (POST /api/data-marts/{dataMartId}/schema-actualize-triggers) verifies it against the storage.',
+      'Updated schema of the data mart. Field connection statuses are server-owned: omit them, and any value sent by an older client is ignored. The saved schema carries the last status derived by schema actualization; a new native field starts DISCONNECTED until schema actualization (POST /api/data-marts/{dataMartId}/schema-actualize-triggers) verifies it against the storage, and a formula may only reference native fields that are already connected.',
   })
   @IsNotEmptyObject()
   schema: DataMartSchemaUpdate;
--- a/apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
+++ b/apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
@@ -2,13 +2,18 @@ import { Injectable, Logger, ForbiddenException } from '@nestjs/common';
 import { Transactional } from 'typeorm-transactional';
 import { BusinessViolationException } from '../../common/exceptions/business-violation.exception';
 import { DataMart } from '../entities/data-mart.entity';
-import { calculatedFieldsOf } from '../calculated-fields/calculated-field.utils';
+import { calculatedFieldsOf, isCalculatedField } from '../calculated-fields/calculated-field.utils';
 import {
   CalculatedFieldValidatorService,
   DryRunContext,
 } from '../calculated-fields/calculated-field-validator.service';
 import { FormulaViolations } from '../calculated-fields/formula-violations';
 import { DataStorageCredentialsResolver } from '../data-storage-types/data-storage-credentials-resolver.service';
+import type {
+  DataMartSchema,
+  DataMartSchemaField,
+} from '../data-storage-types/data-mart-schema.type';
+import { DataMartSchemaFieldStatus } from '../data-storage-types/enums/data-mart-schema-field-status.enum';
 import { DataMartSchemaParserFacade } from '../data-storage-types/facades/data-mart-schema-parser-facade.service';
 import { UpdateDataMartSchemaCommand } from '../dto/domain/update-data-mart-schema.command';
 import { UpdateDataMartSchemaResult } from '../dto/domain/update-data-mart-schema-result.dto';
@@ -55,6 +60,7 @@ export class UpdateDataMartSchemaService {
       command.schema,
       dataMart.storage.type
     );
+    applyServerOwnedFieldStatuses(parsed, dataMart.schema);
 
     // Assigned BEFORE the dry run, not after: composeMetricsOnly (via CalculatedFieldValidatorService)
     // reads `ctx.dataMart.schema` to find each metric's formula, so the context below must carry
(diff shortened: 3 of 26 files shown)


apps/backend/src/data-marts/dto/presentation/update-data-mart-schema-api.dto.ts
```diff
@@ -1,16 +1,17 @@
...
+import type { DataMartSchemaUpdate } from '../../data-storage-types/data-mart-schema.type';
 import { DataMartResponseApiDto } from './data-mart-response-api.dto';
 
 export class UpdateDataMartSchemaApiDto {
   @ApiProperty({
     type: () => Object,
     required: true,
-    description: 'Updated schema of the data mart',
+    description:
+      'Updated schema of the data mart. Field connection statuses are server-owned: omit them, and any value sent by an older client is ignored. The saved schema carries the last status the server derived; a new native field starts DISCONNECTED until schema actualization (POST /data-marts/{id}/schema-actualize-triggers) verifies it against the storage.',
```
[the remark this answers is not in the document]
kalnik-a-a: Fixed in 37bc4952e: the description now points at `POST /api/data-marts/{dataMartId}/schema-actualize-triggers`, matching the `SchemaActualizeTriggerController` route (`data-marts/:dataMartId/schema-actualize-triggers`) and the `/api` prefix used in `docs/api/coverage.md`.

packages/test-utils/src/helpers/set-data-mart-schema.ts
```diff
@@ -0,0 +1,32 @@
...
+  agent: supertest.Agent,
+  app: INestApplication,
+  dataMartId: string,
+  schema: Record<string, unknown>
+): Promise<supertest.Response> {
+  const backendDir = require('path').dirname(require.resolve('@owox/backend/package.json'));
+  const { DataSource } = require(require.resolve('typeorm', { paths: [backendDir] }));
+  await app
+    .get(DataSource)
+    .query('UPDATE data_mart SET schema = ? WHERE id = ?', [JSON.stringify(schema), dataMartId]);
```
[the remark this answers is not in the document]
kalnik-a-a: Done in 37bc4952e: `backendDir` and the `DataSource` token are now resolved once at module scope. The backend-relative resolution itself stays, so `app.get(DataSource)` keeps matching the typeorm instance the backend uses.

apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
```diff
@@ -55,6 +60,7 @@ export class UpdateDataMartSchemaService {
       command.schema,
       dataMart.storage.type
     );
+    applyServerOwnedFieldStatuses(parsed, dataMart.schema);
```
laskevych: 🟡 Important (non-blocking): this changes the API contract in a way neither the DTO description nor the changeset mentions: a single `PUT /schema` that introduces a native field **and** a calculated field referencing it is now refused. `applyServerOwnedFieldStatuses` marks the new native field `DISCONNECTED` before `calculatedFieldValidator.validate` runs, and the validator builds its reference index through `collectFormulaReferenceableFields`, which prunes `!isConnected`, so the `{{ref}}` resolves to nothing and the save answers 400 `FORMULA_UNKNOWN_REFERENCE` (previously 200 with `warehouseValidation: 'skipped'`). The e2e rewrites in this PR are the evidence: the new seed before the `cm*` block in `output-controls.e2e-spec.ts` and the `formula-live-validation` fixture exist precisely so this path keeps passing. The web is unaffected because it already creates native fields as `DISCONNECTED`, so for the UI this is parity; for API and plugin clients it is a new refusal whose error code says 'unknown reference' about a field that is right there in the same payload.

The behaviour itself looks consistent with the task (a field the storage has not verified must not be referenceable), so I am not asking to change it. Two asks: (1) state it in the DTO description and the changeset ('a formula may only reference native fields that are already connected: add the field, run actualization, then add the formula'), and (2) consider a dedicated violation, or at least a hint in `FORMULA_UNKNOWN_REFERENCE`, when the referenced name is present in the submitted schema but disconnected, so the client is told what to do rather than that the field does not exist. It is also worth pinning at the API level in `validation-schema.e2e-spec.ts`: the unit test 'derives statuses for new native and calculated fields' asserts exactly this payload with `validate` mocked, so today the suite documents a save the real service refuses.

apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
```diff
@@ -170,3 +176,37 @@ export class UpdateDataMartSchemaService {
     await this.reportDataCacheService.invalidateByDataMartId(dataMart.id);
   }
 }
+
+function applyServerOwnedFieldStatuses(
```
laskevych, line 184: 💡 Suggestion (non-blocking): consider resetting `dataMart.schemaActualizedAt` when the save introduces a native field that has no persisted namesake. Three consumers actualize lazily through `actualizeSchemaIfExpired` (MCP `get_data_mart_details`, AI metadata generation, SQL-context prefetch) and all of them filter by `isConnected`. With a fresh `schemaActualizedAt` they skip actualization, so a real warehouse column added through the API stays hidden from them for up to the 30-minute window, whereas before the client-sent `CONNECTED` made it visible at once. Clearing the timestamp is a cheap way to let those paths self-heal on their next call without adding a warehouse round trip to the save. The web is fine either way because it runs actualization right after Save.

apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
```diff
@@ -170,3 +176,37 @@ export class UpdateDataMartSchemaService {
...
+
+function applyFieldStatuses(
+  fields: DataMartSchemaField[],
+  persistedFields: readonly DataMartSchemaField[]
+): void {
+  const persistedByName = new Map(persistedFields.map(field => [field.name, field]));
+
+  for (const field of fields) {
+    const persistedField = persistedByName.get(field.name);
+    field.status = isCalculatedField(field)
```
laskevych: 🔵 Question (non-blocking): the carry-forward is keyed on the name only, so a save that changes the definition of an existing native field (type, mode, scalar vs RECORD) keeps the persisted `CONNECTED`, while the actualization mergers (`mergeField` in `bigquery-schema-merger.ts`, `flat-data-mart-schema-merger.ts`) define exactly that as `CONNECTED_WITH_DEFINITION_MISMATCH`. Same for a RECORD parent that gains a nested field: `mergeRecordFields` would roll it up to a mismatch, and the nested-fields unit test now locks in `CONNECTED`. It is not a regression against the old client-owned value, and the web re-actualizes after Save, but for API clients the DTO now promises 'the last status the server derived' for a state the server never derived. Was leaving this to actualization a deliberate scope cut? If not, comparing `type`/`mode` with the persisted namesake and downgrading to `CONNECTED_WITH_DEFINITION_MISMATCH` keeps the rule server-side at the same cost (no warehouse call). Either answer is fine with me; I just want the intent recorded.

apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
```diff
@@ -170,3 +176,37 @@ export class UpdateDataMartSchemaService {
...
+      ? DataMartSchemaFieldStatus.CONNECTED
+      : persistedField && !isCalculatedField(persistedField)
+        ? persistedField.status
+        : DataMartSchemaFieldStatus.DISCONNECTED;
+
+    if ('fields' in field && field.fields) {
```
laskevych: 🟢 Minor (non-blocking): the base field schema is `.passthrough()` and the flat storages (Athena, Redshift, Databricks) declare no `fields` key, so a client value like `fields: {}` on one of their fields survives validation and reaches this loop, where `for (const field of {})` throws and the save answers 500 instead of 400 (before this PR the payload was accepted and stored). If such a value was persisted earlier, `persistedFields.map` at the top of this function fails on every later save of that Data Mart. An `Array.isArray` check here and on the persisted side closes it.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.type.ts
```diff
@@ -15,3 +15,23 @@ export const DataMartSchemaSchema = z.discriminatedUnion('type', [
...
+  ? { [Index in keyof T]: WithoutConnectionStatus<T[Index]> }
+  : T extends object
+    ? {
+        [Key in keyof T as Key extends 'status' ? never : KnownKey<Key>]: WithoutConnectionStatus<
+          T[Key]
+        >;
+      }
+    : T;
+
+export type DataMartSchemaUpdate = WithoutConnectionStatus<z.input<typeof DataMartSchemaSchema>>;
```
laskevych: 💡 Suggestion (non-blocking): this recursion does less than it looks like. `createRecursiveBigqueryFieldSchema` (and the Snowflake twin) declares nested `fields` through a self-referencing getter, which under the backend's `noImplicitAny: false` infers as `any`, so `WithoutConnectionStatus` strips `status` at the top level only: a RECORD whose nested field carries `status` compiles clean against `DataMartSchemaUpdate`. Nothing type-checks against the type either (`validateAndParse` takes `unknown`, Swagger uses `type: () => Object`), so it is documentation. Given `.default(DISCONNECTED)` already makes `status` optional on `z.input`, `export type DataMartSchemaUpdate = z.input<typeof DataMartSchemaSchema>` documents the real wire contract ('optional, ignored') in one line and drops `KnownKey`, which exists only to undo the `.passthrough()` index signature. Keep the mapped type if you specifically want 'absent' rather than 'optional' at the top level; a short comment that it is top-level only would then help the next reader.

apps/web/src/features/data-marts/shared/services/data-mart.service.ts
```diff
@@ -229,10 +230,13 @@ export class DataMartService extends ApiService {
    */
   async updateDataMartSchema(
     id: string,
-    data: UpdateDataMartSchemaRequestDto,
+    data: { schema: DataMartSchema },
     config?: AxiosRequestConfig
   ): Promise<UpdateDataMartSchemaResponseDto> {
-    return this.put<UpdateDataMartSchemaResponseDto>(`/${id}/schema`, data, {
+    const request: UpdateDataMartSchemaRequestDto = {
+      schema: withoutFieldConnectionStatuses(data.schema),
```
laskevych, line 237: 🔵 Question (non-blocking): on the new backend this strip changes nothing (the server overwrites `status` regardless), so its only observable effect today is the rolling-update window described in the PR: a browser that already loaded the new bundle and reaches a pod on the previous release, whose field schema still requires `status`, gets 400 on every Output Schema save until the rollout completes. Would it be acceptable to keep the request unchanged in this release and land the strip (with the `WithoutConnectionStatus` request type) in a follow-up once the lenient backend is everywhere? That removes the window entirely at no behavioural cost. If the answer is that the window is short enough, that is fine too.

apps/backend/test/output-controls.e2e-spec.ts
```diff
@@ -484,6 +475,14 @@ describe('Output controls API (e2e)', () => {
       cmDataMartId = prereqs.dataMartId;
       cmDataDestinationId = prereqs.dataDestinationId;
 
+      await setDataMartSchema(agent, app, cmDataMartId, {
```
laskevych, line 479: 🧹 Nit (non-blocking): this seed is now load-bearing for the rest of the block: it is what keeps `clicks`/`impressions` connected, because the PUTs below would persist them `DISCONNECTED` and `BlendableSchemaService` would drop them from the composition tests further down. The comment on the first test in the block ('the next test's save is the first one that actually persists') is also no longer true. A one-line note here would stop someone from removing the seed as redundant with the PUT.

laskevych (approved):
Approving: no blocking findings.

The server-owned derivation is at the right depth (an actualization round trip on every save would be the wrong trade, and the web already re-actualizes after Save), the zod default matches the existing `level: .optional()` precedent for the load-path transformer, and the test-utils seeding follows the same bypass `setupConnectorDataMart` and the Playwright storage seed already use.

Changeset: I confirmed 6935 is the owning work item, so the filename is right.

Inline comments are all non-blocking: one contract change I would like documented (a native field and a formula referencing it in one PUT is now refused), one suggestion about `schemaActualizedAt`, two questions on intent (definition changes keeping `CONNECTED`; the web strip versus the rollout window), a type simplification, and two small nits.

apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
kalnik-a-a: Agreed on the behaviour, kept as is. Documented and pinned in 03bb034bb: the DTO description now states that a formula may only reference native fields that are already connected; `validation-schema.e2e-spec.ts` pins the 400 (`FORMULA_UNKNOWN_REFERENCE`, subject `revenue`) for a formula referencing a native field introduced in the same save; the unit test `derives statuses for new native and calculated fields` now references a persisted field, so it no longer describes a save the service refuses. The dedicated hint is skipped on purpose: the web never produces this payload (new native fields start DISCONNECTED in every schema table and the editor's reference index excludes them), so only API clients that set CONNECTED themselves are affected, and we are fine with those requests failing. Changeset left as is.

apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
kalnik-a-a: Done in 03bb034bb: when the save introduces a native field with no persisted namesake (nested included), `schemaActualizedAt` is set to null, so `actualizeSchemaIfExpired` treats the schema as expired and the MCP, AI-metadata, SQL-context and HTTP-data paths re-check the warehouse on their next call. Unit tests cover a nested new field (clears) and a calculated-only save (keeps the timestamp).

apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
kalnik-a-a: Deliberate. A definition mismatch is a comparison against the warehouse, which only actualization can make; the save has no warehouse view, so it carries the last derived status forward and the next actualization re-derives it (the web triggers one right after Save). Downgrading on the server would duplicate the per-storage merger rules in a second place for a badge that does not change gating (`isConnected` treats both statuses as connected), and it could not handle the reverse case, a definition corrected back to what the warehouse has. The DTO description now says "the last status derived by schema actualization" to record that (03bb034bb).

apps/backend/src/data-marts/use-cases/update-data-mart-schema.service.ts
kalnik-a-a: Fixed in 03bb034bb: the recursion only descends into array `fields`, on both the submitted and the persisted side (`nestedFieldsOf`), with a unit test for `fields: {}` on an Athena field.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.type.ts
kalnik-a-a: Done in 03bb034bb: `DataMartSchemaUpdate = z.input<typeof DataMartSchemaSchema>`, `KnownKey` and `WithoutConnectionStatus` removed. Confirmed your reading with a tsc probe first: a nested `status` compiled against the mapped type (`fields?: any`), only the top level was stripped.

apps/web/src/features/data-marts/shared/services/data-mart.service.ts
kalnik-a-a: Keeping the strip in this release. The window is the rollout of `owox-data-marts-webserver`: only Output Schema saves from a browser on the new bundle that reach a pod on the previous release get a 400, a retry works, no data loss. The original ask was that the UI stop sending the status, so a two-phase rollout is not worth a follow-up PR here.

apps/backend/test/output-controls.e2e-spec.ts
kalnik-a-a: Done in 03bb034bb: a one-line note on the seed, and the stale sentence replaced with "Rejected, so it leaves the seeded schema untouched".

PR #1604, opened 2026-09-11
fix(deps): resolve production audit vulnerabilities

kalnik-a-a:
## Summary

- replace the vulnerable `adm-zip` archive reader in the connector runner with `fflate`
- override Snowflake SDK's vulnerable transitive `toml` dependency with `toml@4.3.0`
- add a real ZIP round-trip regression test for `FileUtils.unzip`, including UTF-8 text

## Why

The `Audit all` workflow on `main` reported three production dependency vulnerabilities:

- `adm-zip@0.6.0` — GHSA-vwc7-r8mq-g2x9
- `toml@3.0.0`, pulled through `snowflake-sdk` — GHSA-82x6-q7mm-w9cf and GHSA-v5mp-jgw5-2x6j

`fflate` was already a production dependency of the web workspace, where it is used for ZIP generation. Reusing it removes the vulnerable ZIP implementation instead of adding another library to the monorepo. The connectors workspace still declares `fflate` as its own direct runtime dependency because the published connector runner imports it directly; npm resolves both workspace declarations to the same deduplicated `fflate@0.8.3` installation.

Snowflake SDK still declares `toml@^3.0.0`, so the root npm override selects patched `toml@4.3.0`. The SDK only uses the stable `toml.parse()` API in this path, and a local `connections.toml` smoke test confirmed compatibility.

## Verification

- `npm audit --omit=dev` — 0 vulnerabilities
- `npm test -w @owox/connectors` — 285 tests passed
- `npm run build -w @owox/connectors` — passed
- `npm run lint -w @owox/connectors` — passed
- connector and root Prettier checks — passed
- `npm ls fflate toml` — one deduplicated `fflate@0.8.3`; `toml@4.3.0` applied to Snowflake SDK
- Snowflake `connections.toml` smoke test — passed

The full monorepo test run completed successfully for every workspace except the backend, where three unrelated existing tests hit their 5-second timeout under parallel load. Both affected backend spec files passed when rerun serially on the final branch (18/18 tests).

## Release notes

No changeset is needed because this is dependency and security maintenance with no user-visible behavior or public contract change.


package.json
```diff
@@ -116,6 +116,9 @@
     "googleapis-common": {
       "google-auth-library": "^10.1.0"
     },
+    "snowflake-sdk": {
```
laskevych, line 119: 🟡 **Important (non-blocking):** This override fixes `toml` only for installs driven by the monorepo root. The production image installs the published package from the registry (`Dockerfile:6` → `npm install -g owox@${version}`), and npm applies `overrides` only from the root project's `package.json`, so `snowflake-sdk` still resolves `toml@3.0.0` there. The `npm-shrinkwrap.json` step in `publish.yml` that would carry root overrides into the published package (`tools/generate-shrinkwrap.mjs` already copies them) is currently commented out.

The `Audit all` workflow goes green, which is the stated goal, but "production audit vulnerabilities" in the title overstates the outcome. Could we note this limitation in the PR description and either re-enable the shrinkwrap step or track it as a follow-up so the fix reaches the deployed runtime?

packages/connectors/package.json
```diff
@@ -51,7 +51,7 @@
     "@aws-sdk/lib-storage": "^3.983.0",
     "@databricks/sql": "^1.12.0",
     "@google-cloud/bigquery": "^8.0.0",
-    "adm-zip": "^0.6.0",
+    "fflate": "^0.8.3",
```
laskevych: 🔵 **Question:** `adm-zip@0.6.1` was published earlier today and `npm audit` now reports `fixAvailable: true` for it. The advisory (GHSA-vwc7-r8mq-g2x9) covers on-disk extraction following symlinks via `extractAllTo`, which `FileUtils.unzip` never calls: it only reads entries in memory. So the vulnerable path was not reachable here, and a version bump would have closed the audit item with zero runtime change.

I'm fine with keeping `fflate` (one less dependency, already used by `apps/web`), but it changes the only production ZIP path (Microsoft Ads bulk reports). Could the PR description state that the swap is a deliberate consolidation rather than the only way to clear the audit?

packages/connectors/src/Core/Utils/FileUtils.js
```diff
@@ -50,9 +50,8 @@ var FileUtils = class FileUtils {
      * @returns {Array<{getDataAsString: Function}>} Array of file-like objects with getDataAsString method
      */
     static unzip(data) {
-        const zip = new AdmZip(data);
-        return zip.getEntries().map(entry => ({
-            getDataAsString: () => entry.getData().toString('utf8')
+        return Object.values(fflate.unzipSync(data)).map(entry => ({
+            getDataAsString: () => Buffer.from(entry).toString('utf8')
```
laskevych: 💡 **Suggestion:** `Buffer.from(entry)` copies the whole already-inflated `Uint8Array` before decoding, and `fflate.unzipSync` inflates every entry eagerly, whereas `adm-zip`'s `getData()` was lazy. For large Microsoft Ads bulk reports this adds one full-size binary copy on top of the peak that `Helper.processCsvRecordsFromUrl` is explicitly budgeting for.

`fflate.strFromU8(entry)` decodes without the copy (the new test already uses its counterpart `strToU8`), and it also removes the `Buffer` dependency from this GAS-style file:
```js
getDataAsString: () => fflate.strFromU8(entry)
```

packages/connectors/src/Core/Utils/FileUtils.js
```diff
@@ -50,9 +50,8 @@ var FileUtils = class FileUtils {
      * @returns {Array<{getDataAsString: Function}>} Array of file-like objects with getDataAsString method
      */
     static unzip(data) {
-        const zip = new AdmZip(data);
-        return zip.getEntries().map(entry => ({
-            getDataAsString: () => entry.getData().toString('utf8')
+        return Object.values(fflate.unzipSync(data)).map(entry => ({
```
laskevych: 🟢 **Note (non-blocking):** Two semantic differences from `adm-zip` worth knowing about, neither likely for Microsoft Ads archives:
- `adm-zip` verified the CRC32 of each entry and threw `BAD_CRC`; `fflate.unzipSync` does no CRC check, so a corrupted-but-decodable entry now flows into `parseCsv` silently instead of failing the run.
- `fflate` keys entries by filename, so duplicate entry names collapse into the last one; `adm-zip` returned every central-directory entry.

No change requested, just documenting the trade-off for whoever debugs a Microsoft Ads load later.

packages/connectors/test/Core/FileUtils.test.js
```diff
@@ -0,0 +1,23 @@
...
+import { describe, expect, it } from 'vitest';
+import { loadGasClass } from '../support/loadGasClass.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+loadGasClass(path.join(__dirname, '../../src/Core/Utils/FileUtils.js'), { fflate });
+
+describe('FileUtils.unzip', () => {
+  it('returns UTF-8 text for every archive entry', () => {
+    const archive = fflate.zipSync({
```
laskevych: 💡 **Suggestion (optional):** The fixture is produced by `fflate.zipSync` and read back with `fflate.unzipSync`, so the test proves fflate's self-consistency and UTF-8 decoding but not that archives from another zipper are handled. A small base64 fixture generated by an external tool (the real regression class this swap introduces) would be a stronger guard at the same test size.

laskevych (approved):
✅ **Approve.** The `adm-zip` → `fflate` swap is correct for the actual call sites (Node `Buffer` input from `HttpUtils.getBlob`, directory entries yield empty payloads in both libraries, ZIP64 is supported), `toml@4.3.0` is API-compatible with `snowflake-sdk`'s single `toml.parse` call, and the lockfile is consistent with one deduplicated `fflate@0.8.3`. No changeset needed under the release policy.

Inline comments are non-blocking; the one I'd like addressed before or right after merge is the `overrides` reach (the published `owox` package and the Docker image do not pick it up).

🟢 **Note (pre-existing):** the CI test matrix in `.github/workflows/test-owox.yml` has no `@owox/connectors` entry, so the new `FileUtils` regression test only runs locally. Not introduced by this PR, but this PR makes it matter more since it is the only automated guard for the new decompression path. Worth a one-line follow-up.

kalnik-a-a:
## Follow-up update

The first version reused `fflate`, which was already present in `apps/web`, to avoid carrying another ZIP implementation. After review, the narrower solution is a better fit for this audit fix:

- restored the existing `adm-zip` connector implementation and bumped it from `^0.6.0` to `^0.6.1`
- removed the `fflate` connector migration and its new round-trip test
- retained the root `snowflake-sdk` override to `toml@^4.3.0`
- regenerated and verified the lockfile

The final PR diff against `main` is therefore dependency-only: the package manifests and lockfile. Microsoft Ads keeps its existing ZIP code path, while `fflate` remains used only by `apps/web`, where it was already present.

Known follow-up: the root `toml` override fixes repository-root installs and the `Audit all` workflow, but npm does not apply dependency-package overrides when installing the published `owox` package used by the Docker image. Carrying the override into that artifact needs separate shrinkwrap work; the existing publish step is disabled for snapshot-version handling and affects the whole published dependency tree. The missing `@owox/connectors` CI test-matrix entry is also pre-existing and should be handled separately.

Local verification on commit `aaf4a2f02`:

- `npm audit --omit=dev`: 0 vulnerabilities
- `npm test`: all workspace suites passed
- `npm test -w @owox/connectors`: 21 files / 284 tests passed
- connector build, lint, package formatting, and root formatting passed
- Snowflake adapter: 5 / 5 tests passed

GitHub checks are now running for the updated commit.

packages/connectors/test/Core/FileUtils.test.js
kalnik-a-a: **Criticality: NON-BLOCKING / SUPERSEDED**

The test was removed together with the `fflate` behavior change. The final patch only updates the existing ZIP dependency, so this cross-implementation fixture is no longer needed in this PR.

package.json
kalnik-a-a: **Criticality: NON-BLOCKING / FOLLOW-UP**

Confirmed. The root override fixes the monorepo installation and `Audit all`, but it is not applied when the published `owox` package is installed by the Dockerfile. I kept this PR scoped to the current audit failure and documented the distribution limitation in the follow-up summary comment. Re-enabling shrinkwrap is separate work because the existing step is disabled for snapshot-version handling and would affect the full published dependency tree, not only `toml`.

The current OWOX Snowflake call sites always pass connection options, so the SDK's `connections.toml` parser is not reached in those flows. The packaging gap should still be addressed separately with a packed-artifact installation check.

packages/connectors/src/Core/Utils/FileUtils.js
kalnik-a-a: **Criticality: NON-BLOCKING / SUPERSEDED**

The `fflate` implementation was removed in favor of the narrower `adm-zip@^0.6.1` update, so the eager inflation and extra `Buffer.from` copy are no longer present.

packages/connectors/package.json
kalnik-a-a: **Criticality: NON-BLOCKING / ADDRESSED**

Switched the connector back to its existing `adm-zip` path and raised the minimum version to `^0.6.1`. This clears the audit finding while preserving the production ZIP implementation; the `fflate` migration and its regression test were removed.

packages/connectors/src/Core/Utils/FileUtils.js
kalnik-a-a: **Criticality: NON-BLOCKING / SUPERSEDED**

The `fflate` implementation was removed. Keeping `adm-zip@^0.6.1` preserves CRC validation and avoids the silent filename-key collapsing introduced by this migration. `adm-zip@0.6.1` itself also rejects duplicate entry names.

PR #1605, opened 2026-09-11
feat(connectors): pick the Google Sheets tab from a list and tidy the setup form

laskevych:
## Summary

Follow-up to the Google Sheets source connector setup form after the OAuth feature demo:

- **Sheet Name is picked from the spreadsheet's tabs.** Once the spreadsheet is known (Google Picker for OAuth, ID/URL for a service account) the wizard lists its tabs; the value is stored as before. If the list cannot be loaded the field falls back to free-text input with the provider error and a Retry action, so the user is never blocked.
- **No "Spreadsheet ID or URL" input under managed OAuth.** The spreadsheet is chosen with Google Picker in that mode, so the manual input stays hidden while the OAuth settings load, before sign-in and after picking. It still appears for Service Account and for manual OAuth credentials.
- **Header Row and Range moved to Advanced Settings.** Most sheets keep their headers in row 1 and import the whole tab; the defaults are unchanged.

## How it works

The connector contract gains a generic way to declare **dynamic options** for a configuration field:

- `packages/connectors`: new `CONFIG_ATTRIBUTES.DYNAMIC_OPTIONS` plus `optionsDependsOn` on the field; a source implements `fetchFieldOptions(fieldName, signal)`. `GoogleSheetsSource` implements it for `SheetName` with `spreadsheets.get?fields=sheets.properties`, reusing the same authorized request path (token refresh on 401, size guard, error mapping) as the values request, which is now shared through `_fetchSheetsApiJson`.
- `apps/backend`: new `POST /api/connectors/:connectorName/options/preview` (`{ field, configuration }` → `[{ value, label }]`). The whole configuration is deliberately **not** validated up front, because the caller is still filling the form; the source reports what is missing (for example "Spreadsheet ID or URL is required"). Credential injection, the 15 s timeout and the provider→HTTP error mapping are shared with the fields preview through `connector-preview-support.ts`. The specification DTO exposes `optionsDependsOn`.
- `apps/web`: `ConfigurationDynamicOptionsField` + `useConnectorFieldOptions` render a combobox for `DYNAMIC_OPTIONS` fields, load the options once every `optionsDependsOn` field has a value (debounced, stale responses dropped), clear a value that no longer exists in the freshly loaded list, and fall back to a text input on error.

### Root cause of the visible "Spreadsheet ID or URL" input

Two things made the input show up on a fresh OAuth configuration:

1. `OauthRenderFactory` reported "not managed" (and rendered the manual credential fields) while `GET /oauth/settings` was still loading. It now treats the loading state as the managed flow, renders a short loading note instead of the manual fields, and reports manual mode when it unmounts (another auth option was selected).
2. `ConfigurationOneOfRender` seeds `AuthType: { oauth2: {} }` in a mount effect, but the parent `ConfigurationStep` re-seeds its state from `initialConfiguration` right after and dropped that write, so `AuthType` stayed unset until the user touched the form. `ConfigurationStep` now seeds the first `oneOf` option itself, and the hide rule also trusts an explicit "managed" report from the renderer.

## Tests

- `packages/connectors`: `GoogleSheets.test.js` — spec attributes, tab listing in sheet order, 401 refresh while listing, missing spreadsheet, unsupported field.
- `apps/backend`: `connector-field-options-preview.service.spec.ts` (success without full validation, unsupported connector, configuration errors, credential failures pass-through, 401→400, 5xx→502, malformed response, timeout) and e2e cases in `connector-fields.e2e-spec.ts` for the new endpoint.
- `apps/web`: `dynamic-options.utils.test.ts`, `useConnectorFieldOptions.test.tsx`, and `ConfigurationStep.test.tsx` cases for the hidden input (loading, signed-out, fresh configuration, manual mode) and the dynamic Sheet Name field (waiting, loaded, stale value cleared, error fallback).

Verified locally in the setup wizard: Google Sheets → OAuth2 tab shows only the sign-in block and the Sheet Name picker; Service Account tab shows the spreadsheet input; Header Row and Range sit under Advanced Settings; Sheet Name falls back to text input with the provider error when the tabs cannot be loaded.

## Docs and release

- Connector README / GETTING_STARTED updated.
- Changeset: `.changeset/6950-google-sheets-config-ux.md`.

### DIFF
--- a/.changeset/6950-google-sheets-config-ux.md
+++ b/.changeset/6950-google-sheets-config-ux.md
@@ -1,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Google Sheets source: pick the sheet tab from a list and a leaner setup form
+
+The Google Sheets source setup now lists the tabs of the selected spreadsheet so the Sheet Name can be picked instead of typed, for both Google OAuth and Service Account authentication. With Google OAuth the spreadsheet is chosen with Google Picker only, so the manual "Spreadsheet ID or URL" input no longer appears in that mode. Header Row and Range moved to Advanced Settings because most sheets keep their headers in the first row and import the whole tab.
--- a/apps/backend/src/data-marts/connector-types/connector-field-options.ts
+++ b/apps/backend/src/data-marts/connector-types/connector-field-options.ts
@@ -1,0 +1,14 @@
+import { z } from 'zod';
+
+/**
+ * Allowed values of a connector configuration field resolved from the source
+ * at configuration time (fields declared with the DYNAMIC_OPTIONS attribute).
+ */
+export const ConnectorFieldOptions = z.array(
+  z.object({
+    value: z.string(),
+    label: z.string(),
+  })
+);
+
+export type ConnectorFieldOptions = z.infer<typeof ConnectorFieldOptions>;
--- a/apps/backend/src/data-marts/connector-types/connector-specification.ts
+++ b/apps/backend/src/data-marts/connector-types/connector-specification.ts
@@ -15,6 +15,11 @@ export const ConnectorSpecificationItem = z.object({
   placeholder: z.string().optional(),
   minimum: z.number().optional(),
   attributes: z.array(z.string()).optional(),
+  /**
+   * Configuration fields whose values are needed before the dynamic options of
+   * this field can be resolved (only meaningful with the DYNAMIC_OPTIONS attribute).
+   */
+  optionsDependsOn: z.array(z.string()).optional(),
   oauthParams: z.record(z.string(), z.unknown()).optional(),
 });
 
--- a/apps/backend/src/data-marts/controllers/connector.controller.ts
+++ b/apps/backend/src/data-marts/controllers/connector.controller.ts
@@ -6,6 +6,7 @@ import {
   GetConnectorSpecificationSpec,
   GetConnectorFieldsSpec,
   PreviewConnectorFieldsSpec,
+  PreviewConnectorFieldOptionsSpec,
   ExchangeOAuthCredentialsSpec,
   GetConnectorOAuthStatusSpec,
   GetConnectorOAuthSettingsSpec,
@@ -31,6 +32,9 @@ import { ConnectorOAuthStatusResponseApiDto } from '../dto/presentation/connecto
 import { ConnectorOAuthSettingsResponseApiDto } from '../dto/presentation/connector-oauth-settings-response-api.dto';
 import { ConnectorFieldsPreviewService } from '../services/connector/connector-fields-preview.service';
 import { ConnectorFieldsPreviewRequestApiDto } from '../dto/presentation/connector-fields-preview-request-api.dto';
+import { ConnectorFieldOptionsPreviewService } from '../services/connector/connector-field-options-preview.service';
+import { ConnectorFieldOptionsPreviewRequestApiDto } from '../dto/presentation/connector-field-options-preview-request-api.dto';
+import { ConnectorFieldOptionResponseApiDto } from '../dto/presentation/connector-field-option-response-api.dto';
 
 @Controller('connectors')
 @ApiTags('Connectors')
@@ -40,6 +44,7 @@ export class ConnectorController {
     private readonly specificationConnectorService: SpecificationConnectorService,
     private readonly fieldsConnectorService: FieldsConnectorService,
     private readonly connectorFieldsPreviewService: ConnectorFieldsPreviewService,
+    private readonly connectorFieldOptionsPreviewService: ConnectorFieldOptionsPreviewService,
     private readonly mapper: ConnectorMapper,
     private readonly connectorOauthService: ConnectorOauthService
   ) {}
@@ -86,6 +91,23 @@ export class ConnectorController {
       body.configuration ?? {}
     );
     return this.mapper.toFieldsResponse(fields);
+  }
+
+  @Auth(Role.editor())
+  @Post(':connectorName/options/preview')
+  @PreviewConnectorFieldOptionsSpec()
+  async previewConnectorFieldOptions(
+    @AuthContext() context: AuthorizationContext,
+    @Param('connectorName') connectorName: string,
+    @Body() body: ConnectorFieldOptionsPreviewRequestApiDto
+  ): Promise<ConnectorFieldOptionResponseApiDto[]> {
+    const options = await this.connectorFieldOptionsPreviewService.run(
+      context,
+      connectorName,
+      body.field,
+      body.configuration ?? {}
+    );
+    return this.mapper.toFieldOptionsResponse(options);
   }
 
   @Auth(Role.viewer())
--- a/apps/backend/src/data-marts/controllers/oauth-flow-only.controller.spec.ts
+++ b/apps/backend/src/data-marts/controllers/oauth-flow-only.controller.spec.ts
@@ -40,6 +40,7 @@ jest.mock('../use-cases/connector/available-connector.service', () => ({}));
 jest.mock('../use-cases/connector/specification-connector.service', () => ({}));
 jest.mock('../use-cases/connector/fields-connector.service', () => ({}));
 jest.mock('../services/connector/connector-fields-preview.service', () => ({}));
+jest.mock('../services/connector/connector-field-options-preview.service', () => ({}));
 jest.mock('../mappers/connector.mapper', () => ({}));
 jest.mock('../services/connector/connector-oauth.service', () => ({}));
 jest.mock('../use-cases/create-data-destination.service', () => ({}));
--- a/apps/backend/src/data-marts/controllers/spec/connector.api.ts
+++ b/apps/backend/src/data-marts/controllers/spec/connector.api.ts
@@ -16,6 +16,8 @@ import { ConnectorOAuthStatusResponseApiDto } from '../../dto/presentation/conne
 import { ConnectorOAuthSettingsResponseApiDto } from '../../dto/presentation/connector-oauth-settings-response-api.dto';
 import { ExchangeOAuthCredentialsDto } from '../../dto/presentation/exchange-oauth-credentials.dto';
 import { ConnectorFieldsPreviewRequestApiDto } from '../../dto/presentation/connector-fields-preview-request-api.dto';
+import { ConnectorFieldOptionsPreviewRequestApiDto } from '../../dto/presentation/connector-field-options-preview-request-api.dto';
+import { ConnectorFieldOptionResponseApiDto } from '../../dto/presentation/connector-field-option-response-api.dto';
 
 export function GetAvailableConnectorsSpec() {
   return applyDecorators(
@@ -52,6 +54,23 @@ export function PreviewConnectorFieldsSpec() {
     ApiResponse({ status: 403, description: 'Connector credentials cannot access the resource' }),
     ApiResponse({ status: 502, description: 'Connector provider is unavailable' }),
     ApiResponse({ status: 504, description: 'Connector field preview timed out' })
+  );
+}
+
+export function PreviewConnectorFieldOptionsSpec() {
+  return applyDecorators(
+    ApiOperation({
+      summary: 'Preview the allowed values of a connector configuration field',
+      description:
+        'Resolves the options of a field declared with the DYNAMIC_OPTIONS attribute from the source, for example the sheet tabs of the selected spreadsheet.',
+    }),
+    ApiParam({ name: 'connectorName', description: 'Connector name' }),
+    ApiBody({ type: ConnectorFieldOptionsPreviewRequestApiDto }),
+    ApiOkResponse({ type: ConnectorFieldOptionResponseApiDto, isArray: true }),
+    ApiResponse({ status: 400, description: 'Unable to preview connector field options' }),
+    ApiResponse({ status: 403, description: 'Connector credentials cannot access the resource' }),
+    ApiResponse({ status: 502, description: 'Connector provider is unavailable' }),
+    ApiResponse({ status: 504, description: 'Connector field options preview timed out' })
   );
 }
 
--- a/apps/backend/src/data-marts/data-marts.module.ts
+++ b/apps/backend/src/data-marts/data-marts.module.ts
@@ -159,6 +159,7 @@ import { ConnectorMapper } from './mappers/connector.mapper';
 import { SpecificationConnectorService } from './use-cases/connector/specification-connector.service';
 import { FieldsConnectorService } from './use-cases/connector/fields-connector.service';
 import { ConnectorFieldsPreviewService } from './services/connector/connector-fields-preview.service';
+import { ConnectorFieldOptionsPreviewService } from './services/connector/connector-field-options-preview.service';
 import { RunDataMartService } from './use-cases/run-data-mart.service';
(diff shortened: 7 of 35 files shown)


zapolsky:

One candidate blocker was raised and **refuted** at the verification gate (details in note 1), so nothing here is posted inline.

### 1. Skipping `validate()` wholesale costs the defaults and the `oneOf` diagnostics, not just the required-field check

`source.config.validate()` does three things, and the options preview gives up all three to relax one of them:

1. applies spec defaults — `packages/connectors/src/Core/AbstractConfig.js:152-153`, the **only** place `default` becomes `value`;
2. enforces required fields — the part you deliberately want relaxed;
3. reports missing `oneOf` items precisely — `AbstractConfig._validateOneOf:207`, `Parameter 'AuthType.ClientSecret' is required but was not provided`.

The sibling calls it and wraps the error into a clean 400 (`connector-fields-preview.service.ts:61-66`); `connector-field-options-preview.service.ts:70-77` goes straight from `createConnectorPreviewSource` to `fetchFieldOptions`. Losing (1) and (3) produces two separate misleading 502s:

**(a) Losing the defaults.** `MaxFetchRetries` is declared with `default: 3` and no `value` (`AbstractSource.js:23-29`). Unset, `Source.js:494` evaluates `1 <= undefined` → `false`, the retry loop body never runs, and `Source.js:525` throws `'Google Sheets request retry loop ended unexpectedly'` — **without a single HTTP request**, after an OAuth token refresh has already been burned at `Source.js:447`. `_fetchSheetsApiJson` re-wraps it as `HttpRequestException` (`Source.js:480-486`), which `connector-preview-support.ts:140/170` maps to `502 Connector provider is temporarily unavailable`.

*This was raised as a blocker and refuted, so please read the reachability honestly:* the create wizard starts from `connectorConfiguration = {}` (`ConnectorEditForm.tsx:65,261`), so the second seeding effect is skipped, `ConfigurationStep.tsx:124-126` seeds `MaxFetchRetries: 3`, and it is persisted. There is also no historical window — the Google Sheets source landed well after that seeding existed. **The common path is safe.** What remains: clearing *Max Fetch Retries* in Advanced Settings writes `undefined` (`ConfigurationNumberField.tsx:30`), `JSON.stringify` drops the key from both the preview POST and the save, and the field is not `isRequired`, so nothing stops that being persisted — after which the picker permanently reports a Google outage while *Next* and the connector run still work (both call `validate()`). Same for a configuration authored through `PUT /api/data-marts/:id/definition`.

**(b) Losing the `oneOf` diagnostics.** `getAccessToken` dereferences `authConfig.ClientId.value` / `ClientSecret` / `RefreshToken` unguarded (`Source.js:349-351`); the spec declares those under `AuthType.oneOf[0].items`, so `mergeParameters` never populates `AuthType.items` and a missing one is genuinely `undefined`. The `TypeError` is thrown inside `_fetchSheetsApiJson`'s `try`, so it is laundered into `HttpRequestException` → the same `502`. The dereference is pre-existing; **the new path that reaches it is not**, because the fields preview turns the same input into a clean 400 via `validate()`. Reachable whenever `SpreadsheetId` is filled before the credentials — editing or duplicating a config, switching the AuthType tab, and on any self-hosted instance with no managed OAuth, where manual mode is the *only* mode (`OauthRenderFactory.tsx:275`) and `SpreadsheetId` is not hidden.

**Fix (one shaped change, not two):** apply defaults on the preview path and keep the `oneOf` item diagnostics, relaxing only the required-field check. A defensive `this.config.MaxFetchRetries?.value ?? 3` at `Source.js:494` — the defensive shape `GoogleBigQueryStorage.js:988-992` already uses (`MaxFetchRetries?.value` through a default helper) — closes (a) on its own, and a guard in `getAccessToken` throwing `ConnectorConfigurationException` closes (b), matching the guard this PR already added for `SpreadsheetId` at `Source.js:734-736`.

### 2. The combobox hides a value that is not in the loaded options, and disables itself while doing so

`combobox.tsx:87` resolves `selectedOption = options.find(o => o.value === value)` and `:112-122` renders the **placeholder** when there is no match. In `waiting` the hook clears the list (`useConnectorFieldOptions.ts:66`) and the field is `disabled` (`ConfigurationDynamicOptionsField.tsx:110`) — so a configuration still holding `SheetName: "Data"` shows *"Fill the fields above to load sheet names"* in a greyed box, and that invisible value is what gets saved. On every open of an existing data mart this is at least a transient blank (the initial `options` is `[]` through the first `loading`); it is permanent whenever a dependency goes unready. The old plain text input always rendered the value. Worth merging `currentValue` in as a synthetic option, or not disabling the trigger. (The `error` and empty-list branches are fine — they fall back to an `Input` that does show the value.)

### 3. Opening a data mart whose tab was renamed silently blanks the field *and* arms the unsaved-changes guard

`ConfigurationDynamicOptionsField.tsx:41-46` writes `onValueChange(name, '')` on mount, which flows through `ConfigurationStep` `handleValueChange` → the echo effect → `ConnectorEditForm` `setIsDirty(true)`. The user opened a page and now has an empty required field and unsaved changes they did not make. An inline *"the tab X no longer exists in this spreadsheet"* would be kinder than a silent wipe.

### 4. The new `oneOf` seeding is inert for any non-empty `initialConfiguration`

The seed lands in the first effect (`ConfigurationStep.tsx:128-135`), but the pre-existing second effect (`:150-168`) re-seeds from raw `initialConfiguration` in the same flush and applies no defaults, no seed and no `SECRET_MASK`. Both effects share dependencies and run in declaration order, and `initializedRef` is synchronous, so the later write wins. It works for the fresh-configuration case you targeted (`initialConfiguration` is `{}` there, so the guard at `:153` is false) — but the comment above the hunk describes a general fix that is not general. Effect B is not yours; folding the two would make the next default anyone adds actually apply. *(Side effect of the same ordering, pre-existing and unrelated: Effect A's `isEditingExisting && isSecret → SECRET_MASK` write is discarded on every edit too.)*

### 5. The membership guard in the stale-value logic is unprotected by any test

`ConfigurationStep.test.tsx:443-461` ("clears a sheet name that no longer exists in the loaded tabs") asserts only that `SheetName` becomes `''`. An implementation that clears on **every** load passes identically — deleting the `options.some(...)` guard at `ConfigurationDynamicOptionsField.tsx:43` leaves every suite green while wiping a valid sheet name on each reload, including the `Retry` path. No test renders the field with a `SheetName` that *is* in the list and asserts it survives. Given the effect also carries an `exhaustive-deps` disable, that guard is worth pinning.

<details>

- `connector-preview-support.ts:68-80` — `createConnectorPreviewSource` has no declared return type, so the `any` from the `@ts-expect-error` `@owox/connectors` import is now an **exported cross-file** value; it was contained in one private method before the extraction. The capability probe (`:61`, a magic string) is also now decoupled from the call site. `connector-types/interfaces/` is the repo's home for this.
- `Source.js:187` — `HeaderRow` is `isRequired: true, minimum: 1` and this PR moves it into Advanced Settings, which is `defaultOpen = false` and whose `CollapsibleContent` has no `forceMount` (`wizard.tsx:180,205`), while `requiredSpecs` (`ConfigurationStep.tsx:266`) derives from the **full** spec list. Clear Header Row, collapse the section, and *Next* is disabled with nothing on screen explaining why.
- `connector-field-options-preview.service.ts:44` — the backend never checks that the requested `field` actually carries `DYNAMIC_OPTIONS`; it trusts each source. `ConnectorService.getItemByFieldPath` already exists. Also no `@MaxLength` on `field`, which is echoed back in the 400.
- `connector-preview-support.ts:19` — one hardcoded `CONNECTOR_PREVIEW_TIMEOUT_MS = 15_000` now governs both a 400 ms-debounced interactive tab lookup and a preview that can pull 50 MB. You parameterised the messages but not this.
- `ConfigurationDynamicOptionsField.tsx:96-113` — `Combobox` takes no `id`, but `ConfigurationItemRender` renders `<AppWizardStepLabel htmlFor={specification.name}>`, so the label dangles and the required marker drops out of the accessible name. Both `Input` fallbacks in the same file do set `id={name}`, and the tests encode the inconsistency (`name: 'Sheet Name'` vs `'Sheet Name *'`).
- `optionsDependsOn` crosses `connector-specification.ts:22` → `connector.service.ts:469,490` → `connector.mapper.ts:87,105` with **zero** backend assertions, and a `z.object` strips an unknown key silently. One assertion in `connector-specification.e2e-spec.ts` would pin the whole chain — worth it, because `dynamic-options.utils.ts:43` treats "no dependencies" as *ready*, so a break fires an immediate doomed request instead of waiting.
- `connector-preview-support.ts:184,195` — the now-shared, now-"generic" error mapper hardcodes Google-Sheets message fragments. A future source that does not throw `Core.ConnectorConfigurationException` gets a 500 where a 400 belongs.
- `connector.service.ts:490` — `optionsDependsOn` is mapped for `oneOf.items` too, but `NestedConfigurationField` passes the option's **inner** object as `configuration`, so a nested `DYNAMIC_OPTIONS` field would resolve root-level dependency names against it and wait forever. No connector does this today; latent.
- `connector-field-options-preview.service.ts:70` sits outside the `try` that starts at `:72`, so a nested `null`/`{}`/`[]` in `configuration` makes `recursiveValueTransform` throw before `mapConnectorPreviewError` can classify it → an unmapped 500 plus an error-level stack. Pre-existing — `main`'s `fields/preview` has the identical exposure — but now that both share the helper, moving the call inside the `try` fixes both. No stack reaches the client.
- The 15 s timeout does not actually bound the whole lookup: `getAccessToken` (`Source.js:447`) and `OAuthUtils.getAccessToken` never receive the signal, so only the sheets request honours it.
- Every options request builds a fresh source and performs a fresh Google token exchange; the 400 ms debounce is the only throttle, with no backend caching or coalescing.
- `ConfigurationStep.test.tsx:533-569` ("brand-new configuration") does not discriminate the seeding hunk — all four assertions are satisfied by the hide-rule change alone. Similarly untested: the debounce itself (`DYNAMIC_OPTIONS_DEBOUNCE_MS` could be `0`), the abort signal *across* the seam (both connector tests call `fetchFieldOptions('SheetName')` with no signal), the `Retry` button's `onClick`, and the headline combination — the OAuth fixture still declares `SheetName` as a plain `STRING`, a shape `Source.js:164-169` no longer emits.
- `docs/api/coverage.md` gains no row for the new API-key-reachable `POST` endpoint (the sibling `fields/preview` is missing too); CI does not enforce it.
- `CommonConstants.js:30` points readers at `AbstractSource.fetchFieldOptions`, which does not exist — the method is on `GoogleSheetsSource` only (consistent with `fetchFieldsSchema`, so the code is fine; the comment is not). `CREATING_CONNECTOR.md:158-160`, the only doc listing `CONFIG_ATTRIBUTES`, does not mention `DYNAMIC_OPTIONS`.
- minor: `body.configuration ?? {}` (`connector.controller.ts:108`) is unreachable given `@IsObject()` without `@IsOptional()` — same on the sibling, so consistent rather than wrong; `Source.js:409` uses `?.value` where `:426` uses `.value` for the same parameter; four symbols exported from `connector-preview-support.ts` have no external importers; the `unexpected` log string drifted from `'Unexpected connector field preview failure'` for the fields path; "Fill the fields above to load sheet names" points at a field that managed OAuth deliberately hides; `getDynamicOptionsDependencyKey` is `JSON.stringify` key-**order** sensitive, so a same-content rebuild refetches; the 50 MB guard message still says "Narrow the Range and try again" on the tab listing, where `Range` is irrelevant; `POST :connectorName/options/preview` names a field-level concept at connector level while the field travels in the body.

</details>

**Verified and explicitly not an issue**, so it does not get re-litigated: the ~170-line extraction into `connector-preview-support.ts` is behaviour-preserving line for line (timeout, abort, the full status ladder, both fragment lists, `extractProviderStatus` recursion — only a log string changed, and `fields/preview` still calls `validate()`); `@Auth(Role.editor())` matches the sibling exactly; no credential material can reach the HTTP response (`looksLikeAuthenticationFailure` intercepts the `JSON.parse` echo first, and the 403 pass-through carries only the service-account e-mail); SSRF is closed by the `^[A-Za-z0-9_-]+$` check plus `encodeURIComponent` against a hardcoded origin; cross-project/cross-connector credential reuse is rejected by `validateReferences` + `AccessDecisionService` and now locked in by e2e; the preview writes nothing to persistence or run state; Radix unmounts the inactive `TabsContent`, so the managed-mode report cannot strand and hide the spreadsheet input; the stale-response/abort handling in `useConnectorFieldOptions` is correct and cannot loop; masked secrets resolve on both edit flows; and version skew degrades gracefully in both directions to the pre-PR text input.

**CI (read, not re-run):** 19 green. `Audit all` is red on `adm-zip` and `toml`/`snowflake-sdk` — also red on `main`, and this PR touches no manifest or lockfile, so **not attributable to this PR**.

zapolsky (approved):

laskevych:
Thanks for the thorough pass. Addressed in the latest commit:

**1. Skipping `validate()`** — fixed on the source side, as suggested, so the lookup no longer depends on validation side effects: `_fetchSheetResponse` falls back to the default retry budget when `MaxFetchRetries` has no value, and `getAccessToken` resolves the `AuthType` items through a guard that throws `ConnectorConfigurationException` (`Parameter 'AuthType.ClientSecret' is required but was not provided`), which the preview maps to a 400 through `cause`. Both have connector tests. The contract is now spelled out in `CREATING_CONNECTOR.md` and in the service doc comment.

**2 + 3 + 5. Combobox hiding the value / silent wipe / unpinned guard** — the field no longer changes the stored value on its own. The current value is always rendered (merged in as an option while the list is loading, waiting, or does not contain it), and a tab that no longer exists is flagged inline with `aria-invalid` instead of being cleared, so opening a data mart does not arm the unsaved-changes guard. Tests now cover both directions: a value that *is* in the list survives with no `onChange`, and a missing one is flagged with no `onChange`.

**4. Inert `oneOf` seeding** — both effects now go through one `seedConfiguration` function (migration, defaults, `oneOf` seed, secret mask), so the rule applies to a non-empty `initialConfiguration` as well; covered by a test with a non-empty configuration.

From the additional observations:

- the backend now verifies that the requested field carries `DYNAMIC_OPTIONS` in the specification before touching credentials or the source, and `field` has `@MaxLength(128)`;
- `createConnectorPreviewSource` is called inside the mapped `try` in both preview services, returns a typed `ConnectorPreviewSource`, and the helpers without external importers are no longer exported; the fields-preview log line is back to its original wording;
- the combobox takes an `id`, so the label (required marker included) names it — tests use the same accessible name as the text inputs;
- `optionsDependsOn` and the `ADVANCED` placement are pinned by an assertion in `connector-specification.e2e-spec.ts`;
- the oversize message on the tab listing no longer mentions `Range`; `SpreadsheetId` is read the same way in both requests; the `CommonConstants` comment and `CREATING_CONNECTOR.md` describe `DYNAMIC_OPTIONS` correctly; the waiting placeholder no longer points at a field that managed OAuth hides.

Left as follow-ups, since they are either pre-existing or a design decision beyond this change: a required advanced field (`HeaderRow`) cleared inside the collapsed section disables *Next* without an on-screen hint; a separate, shorter timeout for interactive lookups; threading the abort signal through the token exchange; caching/coalescing of option lookups; the Google-Sheets-specific fragments in the shared error mapper; dependency resolution for a `DYNAMIC_OPTIONS` field nested inside a `oneOf`; and the missing `docs/api/coverage.md` rows for both preview endpoints.

PR #1606, opened 2026-09-11
fix(reports): handle array fields in output controls

kalnik-a-a:
## Goal

Allow array and repeated fields to be selected as whole-value report columns, but not as filters, slices, sorting, aggregations or date buckets. Prevent invalid element paths from reaching report controls while preserving ordinary nested fields outside arrays.

## What changes

- Apply the same array-only rule in backend validation, the web picker and MCP guidance. Arrays remain selectable in the column picker; their descendants are not exposed to report controls or calculated-field references.
- Reject unsupported saved array controls with a readable error naming the field and control. This includes `COUNT`, `ANY_VALUE` and blank checks, even where an individual operation could previously execute successfully.
- Serialize joined array values as JSON arrays, preserving each matched source row rather than flattening its elements. Joined column labels identify the JSON array and source type; the effective SQL output type stays `STRING`.
- Normalize whole BigQuery `RECORD` and `ARRAY` values in the report reader, recursively unwrapping SDK scalar wrappers before emitting JSON text. BigQuery repeated headers expose `ARRAY<T>`, so Sheets and Looker Studio receive the same string-valued contract.
- Update the existing release note and report/joined-source guides. The matching spreadsheets extension change uses the same branch name.

## Scope decisions

- Use existing `type`, BigQuery `mode=REPEATED` and joined `sourceFieldType`. No new persisted schema properties, capability flags or storage enum types.
- Ordinary non-repeated `RECORD`/`STRUCT` descendants, such as `address.city`, remain available. Opaque `JSON`, `VARIANT` and `SUPER` retain their existing behavior; runtime values are not inspected to guess whether they contain arrays.
- Snowflake keeps its existing `ARRAY`/`OBJECT`/`VARIANT` to `VARIANT` mapping. Redshift types and parsing are unchanged. No Snowflake compatibility migration or special rollout restriction is introduced.
- No dedicated repair/removal UI, join-override reset or automatic cleanup for old array settings. Existing generic removal actions remain; affected reports must be reconfigured or recreated before saving/running.
- Native BigQuery array dimensions in grouped/sleeve joins and array join-key validation remain separate follow-ups; this PR does not claim to fix those paths.

## Joined JSON shape

Two matched values `["a","b"]` and `["c"]` produce `[["a","b"], ["c"]]`; each ancestor join adds another array level. An empty source array, a source SQL `NULL` and an absent descendant are distinct. Source element order is preserved. Aggregated rows are sorted by serialized JSON text, but separate columns are sorted independently, so sibling-column index alignment is not promised and collation can differ between storage engines.

The real BigQuery regression uses a logical view for the SQL `NULL` case: storing a null array in a BigQuery table normalizes it to `[]`.

## Verification

- Full backend: 614 suites, 9562 tests and 6 snapshots passed with `--maxWorkers=4`.
- Full web: 264 files and 2490 tests passed with `--maxWorkers=4`.
- Backend and web builds, lint, format checks, Markdownlint and `git diff --check` passed.
- A targeted real BigQuery integration test passed for whole-record serialization, nested SDK wrapper normalization and a literal `value` field. The transitive array transport remains covered by the real-DB integration suite.

Fibery Work Item 6748.


apps/backend/src/data-marts/data-destination-types/google-sheets/services/sheet-formatters/sheet-values-formatter.ts
```diff
@@ -59,6 +59,12 @@ export class SheetValuesFormatter {
       columnsToFormat.forEach(({ index, formatter }) => {
         row[index] = formatter!(row[index], sheetTimeZone);
       });
+      // Records and arrays need JSON text; Date already serializes to a scalar.
+      row.forEach((value, index) => {
+        if (value !== null && typeof value === 'object' && !(value instanceof Date)) {
+          row[index] = JSON.stringify(value);
```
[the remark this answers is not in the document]
kalnik-a-a: Thanks for flagging this. With our current BigQuery SDK configuration, nested `INT64` values arrive as `number`, not native `bigint`. We verified this with a real BigQuery query covering RECORDs, arrays, and INT64 boundaries, then successfully processed the results through our reader and Sheets formatter. No change is needed for this scenario.

laskevych: 🟡 Important (non-blocking): `JSON.stringify(value)` on a whole BigQuery RECORD serializes the SDK's nested wrapper objects, because `getReportCellValue` in `bigquery-report-reader.service.ts` unwraps only the top-level cell and returns a plain record as-is. For `order STRUCT<ts TIMESTAMP, amount NUMERIC, n INT64>` the sheet cell becomes `{"ts":{"value":"2024-01-01T00:00:00.000Z"},"amount":"12.5","n":7}` (nested DATE → `{"value":"..."}`, NUMERIC → quoted string via big.js `toJSON`, BYTES → `{"type":"Buffer","data":[...]}`). Separately, the reader's `else if (cell['value']) return cell['value']` heuristic collapses any record with a truthy `value` member (`STRUCT<value NUMERIC, currency STRING>` exports just `"5"`; a sibling row with `value = 0` exports the full JSON).

Since the Snowflake and Databricks readers already JSON-encode object cells at read time, the natural home for this is `getReportCellValue`: recursively unwrap `Big` / `BigQueryDate` / `BigQueryDatetime` / `BigQueryTime` / `BigQueryTimestamp` (the SDK exports these for an `instanceof` check) and emit JSON text for plain objects. That would also fix Looker Studio, which still renders the same record as `[object Object]` via `String(value)`, and would make this loop unnecessary. Not blocking: today's behaviour is still an improvement over the pre-PR run failure.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -1559,7 +1576,7 @@ export function ReportColumnPicker({
   const dropdownColumns = useMemo<DropdownColumn[]>(() => {
     const cols: DropdownColumn[] = [];
     for (const f of nativeFields) {
-      if (f.type) {
+      if (f.type && !isArrayFieldType(f.type)) {
```
laskevych, line 1579: 🟡 Important (non-blocking): the migration path described in the changeset ("remove those controls before saving or running") is hard to discover in the UI. `isSortResolvable` in `output-controls-cleanup.ts` is name-based, so a sort saved before this PR on a REPEATED column still renders as healthy (no badge, not pruned by `withoutUnresolvableSorts`) and only fails with the 400 toast on save. Stored filters / aggregations / date buckets on an array column do get the orphan badge (because `dropdownColumns` now excludes arrays here), but its text says "This column is no longer available in the data mart schema. Remove this rule or restore the column." while the column exists and is selected.

Suggest threading the array check into `SortResolutionContext` (so the sort row gets a badge like the others) and, if cheap, a dedicated "array column-only" wording for the other three rows. The picker test only asserts row icons are absent; it never opens Output settings with a stored array control.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
```diff
@@ -1168,6 +1170,30 @@ export abstract class AbstractBlendedQueryBuilder implements BlendedQueryBuilder
...
 
+  protected buildArrayJsonRollup(fieldName: string, isJsonFragment: boolean): string {
+    // Native SQL NULL is a matched row; a NULL fragment is an absent descendant.
+    const fragment = isJsonFragment
+      ? fieldName
+      : `COALESCE(${this.serializeArrayAsJson(fieldName)}, 'null')`;
+    return `CASE WHEN COUNT(${fragment}) = 0 THEN NULL ELSE CONCAT('[', CONCAT(${this.buildStringAgg(fragment)}, ']')) END`;
+  }
+
+  private serializeArrayAsJson(fieldName: string): string {
```
laskevych, line 1181: 💡 Suggestion: `serializeArrayAsJson` is the only `switch (this.type)` that spells SQL in this class; every other dialect difference is a subclass hook (`buildStringAgg` abstract, `buildAnyValue` / `buildRowSurrogate` / `quoteIdentifier` overrides, per the class doc at the top). The same five expressions already exist as per-dialect overrides in `data-quality-sql-dialect.ts`. A `protected abstract serializeArrayAsJson(fieldName)` next to `buildStringAgg` keeps the documented pattern and makes a new dialect fail at compile time instead of returning `undefined`.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
```diff
@@ -1168,6 +1170,30 @@ export abstract class AbstractBlendedQueryBuilder implements BlendedQueryBuilder
     return `ANY_VALUE(${fieldName})`;
   }
 
+  protected buildArrayJsonRollup(fieldName: string, isJsonFragment: boolean): string {
```
laskevych, line 1173: 💡 Suggestion: for `isJsonFragment = false` the `CASE WHEN COUNT(fragment) = 0 THEN NULL` arm is unreachable (`COALESCE(..., 'null')` is never NULL and `buildAggregationCte` always groups by join keys, so every group has ≥ 1 row), and the serializer is evaluated three times per row (COUNT, agg, ORDER BY) on BigQuery/Athena/Snowflake/Redshift. The CASE is load-bearing only on the passthrough path, where Athena `ARRAY_JOIN` / Databricks `CONCAT_WS` return `''` over an all-NULL group and would print `[]`. Either drop the CASE for the leaf arm or leave a one-line comment saying why it exists, so the next reader doesn't hunt for an empty-group case.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
```diff
@@ -1168,6 +1170,30 @@ export abstract class AbstractBlendedQueryBuilder implements BlendedQueryBuilder
...
+    return `CASE WHEN COUNT(${fragment}) = 0 THEN NULL ELSE CONCAT('[', CONCAT(${this.buildStringAgg(fragment)}, ']')) END`;
+  }
+
+  private serializeArrayAsJson(fieldName: string): string {
+    switch (this.type) {
+      case DataStorageType.GOOGLE_BIGQUERY:
+      case DataStorageType.LEGACY_GOOGLE_BIGQUERY:
+        return `TO_JSON_STRING(${fieldName})`;
+      case DataStorageType.AWS_ATHENA:
+        return `json_format(CAST(${fieldName} AS JSON))`;
```
laskevych, line 1187: 💡 Suggestion: Trino/Athena does not support `CAST(... AS JSON)` for arrays whose element (or nested ROW field) type is DATE / TIMESTAMP / TIME / VARBINARY, and `parseAthenaFieldType` flattens every array to bare `ARRAY`, so `array(timestamp)` joined columns will fail with "Cannot cast array(timestamp(3)) to json". Not a regression (the old `ARRAY_JOIN(ARRAY_AGG(CAST(arr AS VARCHAR)))` failed as well), but specs only cover `ARRAY(VARCHAR)`; either a `transform(col, x -> CAST(x AS VARCHAR))` fallback or a documented limitation would close it.

apps/backend/src/data-marts/services/blendable-schema.service.ts
```diff
@@ -373,8 +379,10 @@ export class BlendableSchemaService {
         dto.sourceDataMartTitle = rel.targetDataMart.title;
         dto.targetAlias = rel.targetAlias;
         dto.originalFieldName = field.name;
-        const dedupFunction =
-          fieldOverride?.aggregateFunction ?? getDefaultAggregateFunction(field.type);
+        // Arrays use the dialect's JSON rollup, regardless of stale scalar overrides.
+        const dedupFunction = isArray
```
laskevych, line 383: 💡 Suggestion: forcing the JSON rollup ignores stored dedup overrides that used to produce valid SQL: a `COUNT` override on a joined BigQuery REPEATED field returned an INTEGER with SUM offered post-join, and on Athena/Databricks `ANY_VALUE` / `COUNT` overrides on `ARRAY` columns worked too. Those now come back as STRING JSON with `postJoinAggregations = []`, so a saved report aggregating such a column fails validation, the override stays persisted, and `SourceFieldsTable` only shows "JSON array". Worth one sentence in the changeset ("existing dedup overrides on array fields are ignored"). (`ANY_VALUE`/`MIN`/`MAX` on BigQuery arrays already failed pre-PR because re-aggregation rewrote `ANY_VALUE` → `MAX`, so that part is not a regression.)

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
```diff
@@ -70,8 +67,9 @@ export function collectSchemaFieldPathDescriptors(
     if (field.isHiddenForReporting) continue;
     if (!isConnected(field)) continue;
     const fullName = prefix ? `${prefix}.${field.name}` : field.name;
-    result.push({ name: fullName, type: comparisonType(field), field });
-    if ('fields' in field && field.fields?.length) {
+    const reportType = getReportFieldType(field);
+    result.push({ name: fullName, type: reportType, field });
+    if ('fields' in field && field.fields?.length && !isArrayFieldType(reportType)) {
```
laskevych, line 73: 💡 Suggestion (follow-up): this walker and `flattenSchemaFields` now stop descending into array-typed parents, but `collectFormulaReferenceableFields` below (and the web `buildReferenceIndex` mirror) still recurse into a REPEATED RECORD, so a formula can reference `items.sku` that no picker / MCP / known-output-columns surface exposes any more; with warehouse validation skipped it saves and fails at run. Applying the same `!isArrayFieldType(getReportFieldType(field))` guard there keeps autocomplete and validation consistent. Fine as a follow-up.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
```diff
@@ -97,15 +100,15 @@ const JoinedFieldSchema = z
...
+        'Qualified field name (<alias>__<field>) — copy verbatim into query_data_mart fields, and into filters/slices only when the corresponding category is present.'
       ),
     displayName: z
       .string()
       .describe('Business-friendly label for presentation; use name for queries.'),
     type: z
       .string()
       .describe(
-        'Type in the blended result — use it to pick operators for filters and functions for aggregations. For a slice, use "sliceType" instead when it is present.'
+        'Type in the blended result — use it to pick operators for filters and functions for aggregations unless it is an array. For a slice, use "sliceType" instead when it is present.'
```
laskevych: 🧹 Nit: for a joined array the facade emits `type: 'STRING'` and `sliceType: 'ARRAY<...>'` (see the `costs__items` fixture), so the clause "use it to pick operators … unless it is an array" on `type` can never apply and nudges a model toward string operators the validator will reject. Suggest pointing at the absent `category` / ARRAY `sliceType` instead.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
```diff
@@ -454,11 +487,21 @@ export class OutputControlsValidatorService {
...
     const errors: ValidationError[] = [];
     for (const rule of sort) {
+      const type = resolveType?.(rule.column);
+      if (type !== undefined && isArrayFieldType(type)) {
+        errors.push(arrayControlError(rule.column, type, 'sort'));
+        continue;
+      }
       if (!selectedColumns.has(rule.column)) {
         errors.push({ code: 'SORT_COLUMN_NOT_SELECTED', column: rule.column });
+        continue;
```
laskevych, line 504: 🧹 Nit: this `continue;` is the last statement of the loop body, so it is a no-op (it looks mirrored from `validateAggregations`, where the same `continue` guards the pair-key logic that follows).

apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
```diff
@@ -276,17 +279,30 @@ export class GetDataMartDetailsTool implements McpToolDefinition<GetDataMartDeta
...
-      out.category = category;
+      const reportFieldType = getReportFieldType({
+        type: out.type,
+        mode: out.mode as BigQueryFieldMode | undefined,
+      });
+      isArray =
+        isArrayFieldType(reportFieldType) ||
+        (typeof out.sliceType === 'string' && isArrayFieldType(out.sliceType));
+      if (isArray) {
+        delete out.category;
```
laskevych, line 299: 🧹 Nit: `delete out.category; delete out.sliceCategory;` are dead — neither key exists on the input (the facade's joined-field DTO and `prepareSchema` produce neither, and recursion runs on raw `fields`), and `sliceCategory` is only assigned later under `out.category !== undefined`. The branch collapses to `if (!isArray) { … }`.

apps/backend/src/ee/mcp/tools/field-type-matrix.ts
```diff
@@ -117,6 +118,7 @@ export function effectiveMcpAggregations(
     allowedAggregations?: ReportAggregateFunction[];
   }
 ): ReportAggregateFunction[] {
+  if (isArrayFieldType(fieldType)) return [];
```
laskevych, line 121: 🧹 Nit: this guard is unreachable from production — the only caller (`enrichField`) already returns `[]` for `isArray` before calling `effectiveMcpAggregations`. Harmless as defence-in-depth, just noting it is not a live invariant.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
```diff
@@ -44,15 +45,11 @@ export function collectSchemaFieldPathTypes(
...
- * category everywhere downstream — validator gating, the MCP field-type matrix, and the
- * renderers' blank/cast branches — so all three surfaces agree (#6779).
+ * The field type report and blending machinery must see. A BigQuery REPEATED field stores
+ * its element type (`STRING` + mode `REPEATED`), but the column is an `ARRAY<STRING>`.
+ * Normalizing it here keeps arrays column-only across downstream output controls.
  */
-function comparisonType(field: DataMartSchemaField): string {
+export function getReportFieldType(field: { type: string; mode?: BigQueryFieldMode }): string {
   const rawType = String(field.type);
   return 'mode' in field && field.mode === BigQueryFieldMode.REPEATED
```
laskevych, line 55: 🧹 Nit: with the new structural parameter type, `'mode' in field &&` and `String(field.type)` are no-ops carried over from the old union-typed `comparisonType` (where the `in` check narrowed a 5-way union). Typing `mode?: string` would also let `blendable-schema.service.ts` and `data-mart-details.tool.ts` drop the `BigQueryFieldMode` imports added only for this (the string enum still compares with `=== BigQueryFieldMode.REPEATED`).

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/output-controls-operators.ts
```diff
@@ -181,12 +181,20 @@ export function isFilterableType(fieldType: string): boolean {
   return operatorsForType(fieldType).length > 0;
 }
 
+/** Mirror of the backend's normalized array type check. */
+export function isArrayFieldType(type: string | undefined): boolean {
+  const head = type
+    ?.trim()
+    .toUpperCase()
+    .match(/^([A-Z]+)(?:\s*[<(]|$)/)?.[1];
```
laskevych, line 189: 🧹 Nit: the "mirror" already differs from the backend regex (`[A-Z]+` here vs `[A-Z_]+` in `field-type-compatibility.ts`). Equivalent today (no input makes them disagree), but `/^ARRAY(?:\s*[<(]|$)/.test(type.trim().toUpperCase())` has nothing left to drift on either side.

apps/backend/src/data-marts/data-storage-types/blending/blended-query.types.ts
```diff
@@ -18,10 +18,11 @@ export interface BlendTreeNode {
 
 /**
  * A child chain's blended field as seen by an ANCESTOR chain: it is already aggregated by
- * the child's own CTE, so the ancestor re-rolls it (see `getReAggregateFunction`).
+ * the child's own CTE, so the ancestor re-rolls it (see `reAggregateFunctionFor`).
  */
 export interface PassthroughField {
   outputAlias: string;
+  targetFieldType?: string;
```
laskevych, line 25: 🧹 Nit: an undefined `targetFieldType` silently falls back to a flat `STRING_AGG` over the child's JSON text in `buildAggregationCte` (`["a"], ["b"]` instead of `[["a"],["b"]]`), with no error. The single producer today always sets it, so this is latent; a comment stating the invariant here (or an assertion in the builder) would guard it without touching the many spec fixtures that omit the field.

docs/getting-started/setup-guide/joinable-data-marts.md
```diff
@@ -175,6 +175,14 @@ Two paths to the same target Data Mart are also supported. If both **Campaigns 
 
 If a chain folds back on a Data Mart that already appears earlier in the same branch (e.g., `A → B → A`), OWOX Data Marts stops descending at that point and renders a collapsed **Loop** badge in the relationship list and the canvas. The loop branch contributes no fields to the column picker — this is expected behavior, not an error.
 
+## Joined array columns
+
+An array identified by the source schema uses JSON array rollup, preserving the array from each source row. For example, two matching rows containing `["a", "b"]` and `["c"]` produce `[["a", "b"], ["c"]]`. Each ancestor join adds one more array level; descendants are not flattened or unnested.
```
laskevych: 🧹 Nit: all five serializers emit compact JSON, so the actual output is `[["a","b"], ["c"]]` (the only space is the `', '` fragment separator). Also, every dialect's `buildStringAgg` sorts fragments by their JSON text, so the order of aggregated rows is deterministic (lexicographic) rather than "not guaranteed" — conservatively true, but not what the code does.

laskevych (approved):
Approving — no blocking findings.

Verified end-to-end: run-time `targetFieldType` for joined arrays is the normalized `ARRAY<T>` (`blended-report-data.service.ts:848`), the validator / web picker / MCP details tool agree on the array classification for native and joined fields, every reachable dialect (BigQuery, Athena, Databricks) emits valid rollup SQL, no `buildStringAgg` uses `DISTINCT`, and all run paths (scheduled runs, Looker cache, HTTP stream, MCP query) go through the same `validateForReport` chokepoint. The scope decisions stated in the description (sleeve/Totals `=` on native array dimensions, join-key validation, Snowflake ARRAY→VARIANT) are not re-raised here.

Inline comments are all non-blocking: two 🟡 Important items (whole-RECORD JSON export leaking BigQuery SDK wrappers — best fixed in the reader; stale array sort rendering as healthy in the web), a few 💡 suggestions and 🧹 nits.

One note on a file outside the diff — `bigquery-report-headers-generator.service.ts:52` (pre-existing): the header generator still emits the element type for REPEATED fields, so Looker `getSchema` advertises NUMBER and MCP/REST metadata say `INTEGER` for a column whose values are JSON-array text, while joined arrays are typed STRING. `getReportFieldType(field)` there yields `ARRAY<INTEGER>`, which the Looker mapper already maps to STRING. Fine as a follow-up.

apps/backend/src/data-marts/data-destination-types/google-sheets/services/sheet-formatters/sheet-values-formatter.ts
kalnik-a-a: Addressed in `1a1f03e94`: the BigQuery reader now recursively normalizes SDK scalar wrappers, serializes whole RECORD/ARRAY values at the reader boundary, and unwraps only known wrapper classes, so a literal `value` field is preserved. Unit coverage includes nested date/time, numeric, bytes and array values, and the targeted real BigQuery integration test passed. Repeated headers now also expose `ARRAY<T>` to downstream consumers.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
kalnik-a-a: Kept out of scope by design. This PR does not add array-specific repair/removal UX or change the existing saved-control discovery flow; affected reports must remove/reconfigure those controls or be recreated. The PR description and changeset now state that migration path explicitly, while backend validation remains the authoritative guard.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
kalnik-a-a: Not changed. The existing switch is the single serialization decision point for the five supported dialects and is covered by dialect SQL tests. Moving one expression into five subclass overrides would add boilerplate without changing the contract; an abstract hook becomes useful when another dialect is actually introduced.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
kalnik-a-a: Kept the shared `CASE` intentionally. It gives leaf and passthrough rollups one NULL contract, and the null/empty-array cases are covered by the integration suite. Splitting the paths solely to avoid repeated serialization is an optimization outside this correctness fix.

apps/backend/src/data-marts/data-storage-types/interfaces/abstract-blended-query-builder.ts
kalnik-a-a: Addressed in the guide: the limitations section now states that Athena/Trino joined-array rollup requires JSON-castable elements and nested fields, and that date, time, timestamp and binary values are not currently supported in this cast path. No fallback was added because the parsed type metadata is flattened; a safe transform would be a separate storage-specific change.

apps/backend/src/data-marts/services/blendable-schema.service.ts
kalnik-a-a: Addressed in the changeset: existing deduplication overrides on joined array fields are explicitly documented as ignored in favor of JSON-array rollup. No automatic reset or migration was added, consistent with the stated scope.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
kalnik-a-a: Addressed in `1a1f03e94`: both backend `collectFormulaReferenceableFields` and the web `buildReferenceIndex` mirror now stop recursion under ARRAY/REPEATED parents, while ordinary non-repeated RECORD descendants remain referenceable. Unit tests cover both sides.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
kalnik-a-a: Addressed: the MCP description now permits filter and aggregation operations only when `category` or `allowedAggregations` is present, uses `sliceType` for slices, and explicitly identifies a field without a category whose `sliceType` is ARRAY as column-only.

apps/backend/src/data-marts/services/output-controls-validator.service.ts
kalnik-a-a: Not changed. The trailing `continue` is harmless and keeps the early-rejection shape consistent with the adjacent validators; removing it would be cosmetic and unrelated to the behavior fixed here.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
kalnik-a-a: Not changed. Current producers omit these properties, but deleting them at the array boundary is a small defensive invariant on the cloned `RawField`: an array cannot advertise a category if an upstream shape later carries one. No user-visible behavior depends on this branch today.

apps/backend/src/ee/mcp/tools/field-type-matrix.ts
kalnik-a-a: Kept intentionally as defense in depth in the exported governance helper. The current caller short-circuits arrays, but returning `[]` locally prevents a future or direct caller from advertising array aggregations.

apps/backend/src/data-marts/data-storage-types/data-mart-schema.utils.ts
kalnik-a-a: Partially simplified in `1a1f03e94`: `getReportFieldType` now preserves the input type in its return type instead of coercing it through `String`. I kept the BigQuery mode enum at storage-aware call sites and the structural guard so non-BigQuery field shapes remain accepted without widening every caller; no behavioral change was needed.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/output-controls-operators.ts
kalnik-a-a: Not changed. Both predicates recognize every ARRAY spelling currently produced by supported schemas, so there is no behavioral discrepancy for the present input contract. Unifying the mirror is a valid small cleanup, but is not required for this fix.

apps/backend/src/data-marts/data-storage-types/blending/blended-query.types.ts
kalnik-a-a: Not changed. The production assembly path always supplies `targetFieldType`; optionality remains for existing internal call shapes and fixtures. The real producer path and nested array levels are covered by the integration suite. Making the field required across those fixtures can be handled as a separate type-tightening cleanup.

docs/getting-started/setup-guide/joinable-data-marts.md
kalnik-a-a: Addressed in the guide: the example now matches compact per-row JSON plus the `, ` fragment separator (`[["a","b"], ["c"]]`). The text also states that rows are sorted by serialized JSON text, while separate columns sort independently and collation can vary by storage engine.

PR #1607, opened 2026-09-11
fix(ci): prevent Release PR lockfile update failures

kalnik-a-a:
## Summary

- update `package-lock.json` in the Release PR workflow with `npm install --package-lock-only`
- preserve the dependency graph already recorded in the committed lockfile
- avoid the npm Arborist crash triggered by deleting the lockfile and resolving the complete dependency tree from scratch

## Context

The Release PR workflow installs the repository from the committed lockfile with `npm ci`, then runs Changesets to update package versions and changelogs. The following `Update package-lock.json` step previously called `npm run clean:reinstall`.

`clean:reinstall` removes both `node_modules` and `package-lock.json` before running an unrestricted `npm install`. Recent runs consistently fail during that fresh dependency resolution with:

```text
Cannot read properties of null (reading 'edgesOut')
```

The exception comes from npm Arborist while it builds a peer dependency set. The latest failing workflow run is [Release PR #1326](https://github.com/OWOX/owox-data-marts/actions/runs/34610299957).

## Why this preserves the release workflow behavior

The purpose of this step is to synchronize `package-lock.json` with the package manifests generated by Changesets. `npm install --package-lock-only` still reads the updated `package.json` files, resolves any required lockfile changes, and writes `package-lock.json`. The flag only prevents npm from modifying `node_modules`.

For the current Version Packages state, Changesets updates 12 workspace package versions. The new command updates all 12 corresponding lockfile entries. A clean `npm ci` using that generated lockfile then succeeds, confirming that the lockfile remains complete and installable.

The workflow no longer deletes the existing lockfile or incidentally refreshes unrelated transitive dependencies that still satisfy semver ranges. That full graph refresh was not required to synchronize Changesets version metadata and was the operation triggering the npm crash. Dependency updates already present on `main` remain preserved because the command starts from the committed lockfile.

The Changesets action, changelog formatting, package formatting, generated lockfile commit, and push steps remain unchanged. No application code, package API, runtime behavior, or publishing contract is modified.

## Verification

- parsed `.github/workflows/release-pr.yml` successfully as YAML
- `git diff --check`
- ran the new lockfile command on a clean archive of the current `changeset-release/main` branch
- confirmed that the generated lockfile contains all 12 expected workspace version updates and no unrelated dependency changes
- ran a clean `npm ci --ignore-scripts --no-audit --no-fund --loglevel error` against the generated lockfile; it installed 2,590 packages successfully

## Release notes

No changeset is included because this is a CI-only correction with no observable product or operator-facing release impact.

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Update package-lock.json
         if: steps.changesets.outputs.hasChangesets == 'true'
-        run: npm run clean:reinstall
+        run: npm install --package-lock-only --ignore-scripts --no-audit --no-fund --loglevel error
 
       - name: Commit fixed changelogs
         if: steps.changesets.outputs.hasChangesets == 'true'


.github/workflows/release-pr.yml
```diff
@@ -43,7 +43,7 @@ jobs:
 
       - name: Update package-lock.json
         if: steps.changesets.outputs.hasChangesets == 'true'
-        run: npm run clean:reinstall
+        run: npm install --package-lock-only --ignore-scripts --no-audit --no-fund --loglevel error
```
laskevych, line 46: 🟡 **Important (non-blocking):** `--package-lock-only` still executes the Arborist code path that crashes, so this reduces exposure rather than removing it. Could you confirm the verification ran on npm 10.9.8 — the version this runner actually has?

The failing run `34610299957` logs `node: v22.23.2` / `npm: 10.9.8`, i.e. the npm bundled with Node 22. The stack in [npm/cli#9787](https://github.com/npm/cli/issues/9787) is `Arborist.reify` → `buildIdealTree` → `#buildDepStep` → `#loadPeerSet`, and `--package-lock-only` does not skip `buildIdealTree`: it only swaps `loadActual` for `loadVirtual` and skips writing `node_modules`. The same `#loadPeerSet` path still runs — what changes is that seeding from the committed lockfile leaves far less peer-set resolution to do.

That makes the fix plausible and probably effective, but empirical rather than immunity, and the upstream issue is still open and labelled `Cannot Reproduce`, so there is no upstream fix to lean on. If the local trial used a separately installed npm 11.x, it does not cover the code path CI takes. Stating the npm version in the Verification section would settle it.

laskevych, line 46: 🟡 **Important (non-blocking):** this step no longer proves the generated lockfile is installable, and the Version Packages PR's own checks do not run automatically. Consider adding a validation step right after this one:

```yaml
      - name: Verify package-lock.json installs
        if: steps.changesets.outputs.hasChangesets == 'true'
        run: npm ci --dry-run --ignore-scripts --no-audit --no-fund --loglevel error
```

`clean:reinstall` ended in a real `npm install`, so a lockfile that could not be resolved failed this step before anything was committed. `--package-lock-only` writes the lockfile without reifying it, so that guard is gone.

The natural fallback would be the Version Packages PR's own CI, but every check on `changeset-release/main` sits at `action_required` and never executes — run `34610427495` (Test OWOX, `pull_request`) is `completed / action_required` with 0 jobs, and the same holds for Code Quality root, Audit all and the E2E workflows. A lockfile that `npm ci` would reject therefore reaches `main` unchallenged, where `publish.yml` runs `npm ci` before `changeset publish`.

The dry-run install encodes in the workflow exactly the check that was done manually for this PR, and costs seconds since `node_modules` is already populated.

laskevych, line 46: 🔵 **Question (non-blocking):** with the from-scratch resolution gone, is anything left that refreshes transitive dependencies automatically? I don't see a Dependabot or Renovate config in the repository.

The description frames dropping the incidental transitive refresh as a benefit, and for the purpose of this step that is right. The part worth confirming is the wider consequence: deleting `package-lock.json` in `clean:reinstall` was the only automated path that re-resolved the whole graph, there is no `.github/dependabot.yml` and no renovate config, and `audit-all.yml` runs `npm audit --omit=dev` against the committed lockfile on every PR and on `main`.

After this change, picking up patched transitive versions depends entirely on manual PRs such as #1604. That may well be the intended state — just worth making explicit, or scheduling a separate periodic refresh job so the audit gate is not the only thing that notices.

laskevych (approved):
Approving — no blocking findings. The diagnosis holds up: run `34610299957` fails exactly on `Update package-lock.json` with `npm error Cannot read properties of null (reading 'edgesOut')`, and this step has failed on every run of the workflow for the past week. The change is also low-risk for this monorepo specifically: internal dependencies are pinned as `"*"` (#419), so `changeset version` only rewrites `version` fields and the lockfile delta stays minimal and predictable. `changesets/action@v1` does not touch the lockfile itself, so the step is not redundant. Omitting a changeset is correct — `docs/contributing/repository/release-strategy.md` puts "CI, local developer tooling" changes in the no-release-impact category.

Everything below is non-blocking. Two inline comments are worth a reply before merge; the rest are optional.

---

### 🔵 Question (non-blocking): should this workflow pin npm the way `publish.yml` does?

The release lockfile is now written by npm 10.9.8 here, but installed by npm 11.19.0 at publish time.

`publish.yml` (lines 41-44) explicitly runs `npm install --global npm@11.19.0` before `npm ci` and `changeset publish`, while `release-pr.yml` uses whatever npm ships with Node 22 — 10.9.8 in run `34610299957`. This workflow also pinned `npm@11` itself until #1113 removed that step.

So the npm that generates the release lockfile and the npm that consumes it to publish are two different majors. Pinning the same version here would remove that skew, keep lockfile generation deterministic against future Node 22 patch bumps, and give a second independent shot at the Arborist bug — without changing the command this PR introduces.

---

### 💡 Suggestion (non-blocking): add a `workflow_dispatch` trigger

The workflow only triggers on `push: branches: [main]`, so this fix is first exercised after merge — and this step has failed on every run from `34106111503` through `34610299957`. If the new command still hits the Arborist crash, or the lockfile needs another attempt, the only way to retry is to land another commit on `main`.

Adding `workflow_dispatch` under `on:` makes the job re-runnable on demand. The job body is unaffected by the trigger, so a manual run does exactly what a push does.

---

### 💡 Suggestion (non-blocking): `clean:reinstall` keeps the failing behaviour for contributors

After this PR, `clean:reinstall` is referenced nowhere except its own definition in `package.json` (line 28), so its only remaining consumers are contributors running it locally. It still does `rm -f package-lock.json && npm install`, which is precisely the from-scratch resolution that crashes on npm 10.9.x.

The CI symptom is fixed while the command that triggers it stays in the repo unchanged and undocumented. A line in the description or a follow-up pointing at [npm/cli#9787](https://github.com/npm/cli/issues/9787) would save the next person the same investigation.

Worth referencing too: `tools/generate-shrinkwrap.mjs` (line 337) already runs `npm install --package-lock-only --ignore-scripts`, so the pattern this PR adopts is the one the repo already relies on elsewhere.

---

### 💡 Suggestion (optional, out of scope): `release-pr.yml` has no `concurrency` group

`lint-root.yml`, `test-owox.yml`, `audit-all.yml`, `test-integration.yml` and the rest all declare a `concurrency` block; `release-pr.yml` does not. Two pushes to `main` close together therefore start two jobs that both run `changeset version` against `changeset-release/main` and both end in `git push origin HEAD` (line 56), so the later push can be rejected or silently clobber the earlier one's changelog and lockfile commit.

A top-level block serialises them without dropping a release PR update:

```yaml
concurrency:
  group: ${{ github.workflow }}
  cancel-in-progress: false
```

Entirely unrelated to the one-line fix — listed only because this file is open.

PR #1608, opened 2026-09-13
feat(destinations): implement auto-advancing promo tabs for Google Sheets, Excel, and MCP

artemfenoga:
# Destinations tab now promotes Google Sheets, Excel, and MCP together

image

A published Data Mart with no destinations yet now offers three ways to get
value from your data as switchable tabs — **Google Sheets**, **Microsoft
Excel**, and connecting an AI assistant (**Claude** or **ChatGPT**) via MCP —
instead of only Google Sheets.

The active tab cycles automatically every few seconds until you hover, focus,
or pick one yourself, so all three options stay visible without cluttering
the page with three promo blocks at once.


laskevych (approved):
Amazing 🤩

PR #1609, opened 2026-09-14
feat(data-marts): allow configuring disconnected joined fields

kalnik-a-a:
## Summary

- keep disconnected target Output Schema fields configurable in Joined Data Mart Report Fields
- preserve the report-safe default that excludes disconnected fields
- document the behavior and add the #6951 changeset

## Deliberate trade-off

Disconnected fields are intentionally shown in the relationship editor without a separate status indicator in this iteration. In the Detailed ERD view, disconnected fields that retain a known type therefore render as ordinary field rows. As a result, configuration field counts may include them and differ from reporting surfaces. If this becomes confusing, explicit status highlighting can be added in a follow-up, potentially matching the Output Schema UI.

## Testing

- `AWS_EC2_METADATA_DISABLED=true npm test -w @owox/backend -- --maxWorkers=4`
- focused ESLint, Prettier, Markdownlint, and diff checks
- connectors, IDP protocol, and backend builds


laskevych (approved):
✅ Approve. Verified the report-safe path end to end: every report / MCP / HTTP-data consumer calls `computeBlendableSchema` without options, the blended-fields-config save does no field-name validation (so overrides on disconnected fields persist and apply again on reconnect), and the second collect pass keeps `calculatedFieldIssues` report-safe. One non-blocking note below.

🟡 Important (non-blocking) — `apps/web/src/features/data-marts/edit/components/DataMartRelationships/DataMartRelationshipsContent.tsx:327` (also line 238)

The Detailed ERD view and the canvas field-count badge are built from the same `includeDraftTargets: true` payload, and their only "not connected" proxy is `field.type === 'UNKNOWN'`. Both schema mergers keep the original `type` when marking a field `DISCONNECTED` (they spread the existing field and override only `status`), so after this change a dropped warehouse column renders on the ERD card as an ordinary typed row and is counted by `connectedFieldCounts`. The PR body covers the field-count difference, but the ERD rows are a separate surface, and the comment there ("UNKNOWN-typed fields are not connected") now describes an invariant that no longer holds. If the missing indicator stays a deliberate trade-off for this iteration, it would be good to at least update that comment and the PR description so the diagram behaviour is a documented consequence rather than a surprise.

kalnik-a-a:
@laskevych Addressed in `1daf81ec0`:

- documented that disconnected fields retaining a known type render as ordinary rows in the Detailed ERD and may be included in configuration field counts;
- updated the `connectedFieldCounts`/ERD comments to describe that intentional behavior;
- clarified the guide that only connected joined fields are available for reporting and that report exclusion applies to disconnected native fields, not calculated fields.

No runtime behavior was changed; the missing status indicator remains the deliberate trade-off for this iteration.

PR #1610, opened 2026-09-14
feat(search): add reports search

kalnik-a-a:
## What & why

Reports can now be found directly from project search alongside Data Marts, storages and destinations. Results show the report's Data Mart and destination and open the report sidesheet on the Data Mart's Destinations tab. The same search is available in the web UI, REST API, API client and MCP.

## Demo

https://github.com/user-attachments/assets/3b393570-96e8-4133-b533-610ee6489f93

## Behavior and implementation

- **Search and ranking:** the report title is the strongest keyword signal; the Data Mart and destination titles/type provide lower-weight context. Exact words and word prefixes are supported. Only the `REPORT` branch omits `report`/`reports` from a query when other searchable words remain. A sole `report` query stays searchable, and existing entity types keep their matching behavior. SQL candidate selection and keyword scoring use the same entity-specific tokens.
- **Visibility:** a report is searchable only when both its Data Mart and destination are visible, matching the Destinations tab. The existing visibility helpers are combined before candidate limits, including owner/admin access and context restrictions. Both parents must be live and in the indexed project. Report ownership neither adds a search restriction nor bypasses these checks; existing permissions for report actions remain unchanged.
- **Index lifecycle:** adds the report index, project reindex triggers and migrations. Existing reports are backfilled by reconcile; report create/update/delete keep search in sync. Deletion uses the stable command ID after TypeORM removes the report, and the shared queue rejects absent or empty IDs before repository access, preserving other reports' pending triggers. Missing destinations are skipped, and an empty hydrated page follows its cursor instead of ending the scan. Report hydration selects only indexed fields; empty Looker report titles fall back to the destination title in the backend, consistently for UI, REST and MCP.
- **Parent renames and concurrent indexing:** only actual title changes enqueue report work. One `REINDEX_REPORTS` job per parent runs through the fast entity queue in bounded pages, with persisted progress and supporting parent indexes. It yields between pages and keeps concurrent rename requests separate. REPORT writes compare the expected index state atomically and verify fresh source data after writing, including the rename-back case; conflicting IDs get at most three attempts. Unchanged READY documents skip embedding generation. Successful vectors from a partially failed page are retained, later pages continue, and cumulative failures keep the final job in ERROR for visibility; periodic reconcile remains the recovery path.
- **Embedding outages:** a missing vector still stores the current report text with `MISSING` status, keeping keyword search available for new and renamed reports. Later syncs retry the vector, while project statistics distinguish embedding failures from other errors. Partially failed pages retain successful work and report the actual failure count; a delayed failure cannot overwrite an embedding already recovered by another worker.
- **REST and SDK:** `/api/search` accepts `entityTypes=REPORT`. Report results include the report ID/title, `report.dataMart`, `report.dataDestination` with its type, relevance scores and a canonical direct `url`. The SDK exposes a discriminated `REPORT` result with runtime validation. **Omitting `entityTypes` searches all supported types, including reports.** Explicit filters and an explicitly empty SDK filter retain their existing behavior. OpenAPI and [API client search documentation](https://github.com/OWOX/owox-data-marts/blob/feat/add-reports-search/docs/api/api-client.md#search-project-entities) are updated.
- **Web and MCP:** the web search displays report context with monochrome Data Mart and Destination icons matching the existing search results, and a fallback title for untitled reports. Long names truncate without displacing the icons; screen readers retain explicit entity labels. Short tooltips identify the entity type in every result and the Data Mart and Destination in report context. `get_relevant_reports_by_prompt` uses the shared search facade with the MCP user's access scope and `mcp:read`. Existing `get_data_mart_reports` remains available for Data Mart-scoped listing. See the [MCP report search documentation](https://github.com/OWOX/owox-data-marts/blob/feat/add-reports-search/docs/getting-started/setup-guide/mcp.md#get_relevant_reports_by_prompt).
- **Runtime and release:** `ADVANCED_SEARCH_REPORT_PROJECT_PROCESSING_CRON` defaults to `15,45 * * * * *`; `.env.example`, MCP documentation and the user-facing `owox: minor` changeset are included. Managed deployment configuration does not need an override for the default.

## Compatibility and scope

**Upgrade and rollback:** the new `REINDEX_REPORTS` operation is not compatible with older API producers or search consumers sharing the queue. Stop old API replicas and direct/Pub/Sub consumers and drain their in-flight work before applying migrations and starting the new version. Older consumers can silently complete these jobs without indexing reports, while older producers can overwrite them during deduplication. Before rollback, stop new producers and let the new consumers finish pending report jobs before starting old replicas; retain the additive schema until the queue is drained. Coordinate managed rollout in the deployment repository.

Older SDK releases reject responses containing the new `REPORT` type. Clients making unfiltered searches must update to a Report-aware SDK, or explicitly request their supported entity types, before using the new backend. The backend retains the established search-all default.

Typo tolerance, arbitrary substring matching and relaxed matching of missing query words are deferred. The `report` token rule above applies only to Report searches. Bulk deletion through a Data Mart can still leave unreachable report index rows; search excludes these rows, and cleanup is deferred.

## Validation

Latest backend follow-up verified locally on the exact code committed as `9332151e7f84f56386b32405b290d2d302cdd93e`:

| Surface | Result |
| --- | --- |
| Search unit and SQLite tests, including provider outages and concurrent recovery | 24 suites, 471 tests passed; 4 workers |
| Isolated real MySQL 8.1.0 integration suite | 1 suite passed, Jest reports 9 passed; 1 worker. Cloud SQL-only `COSINE_DISTANCE` assertion was not executed on vanilla MySQL. |
| Backend production build and dependencies | Passed |
| Scoped backend ESLint/Prettier, Markdownlint and `git diff --check` | Passed |

Outage regressions cover keyword discoverability of a newly indexed report, refreshed search text after a rename, `MISSING` → `READY` recovery, correct partial-failure counts, and a delayed failed embedding response racing with a successful worker.

The live MySQL run verified REPORT migrations, FULLTEXT search/project isolation, parent pagination indexes and the conditional batch UPDATE, including rejection of stale writes while an independent row is updated. No SQL fix was needed. Live Cloud SQL vector execution and Pub/Sub validation remain outstanding; the local run does not establish those results. The CI workflow still has no dedicated MySQL job.

The broader backend baseline was verified on `79d316b72dda617ccad53d38cd6c64c1b182be50`: 622 suites / 9601 tests, including 30 focused suites / 528 tests, plus 4 search HTTP parameter/validation tests using real NestJS/SQLite migrations and a mocked search facade. Concurrency coverage includes either completion order, rename-back (`A → B → A`), simultaneous SQLite writers, bounded retries, concurrent parent jobs, parent/project isolation, and continuation after a failed page with cumulative ERROR status. The migration test preserves existing trigger rows through upgrade and rollback. The first HTTP E2E run timed out in application setup during the full unit run; an isolated rerun passed without code or timeout changes. Full-suite output included Snowflake/AWS teardown diagnostics; the suite exited successfully.

Previously verified on `9f38e60454ed26eb56402e6d14aeaaa66eb3e182`, for surfaces unchanged by this follow-up: API client search and typecheck; web search/report sidesheet tests; Chromium report layout at 900 px and 375 px; web production build. The release-note preview verified Cloudflare playback, responsive 16:9 layout and both documentation anchors with a `/pr1610` base. Links and media are unchanged; the new upgrade warning is prose only.

The release-note preview used a temporary changelog fixture and extended its existing workflow-file link exclusion to `/pr1610/.github/**`. No docs publishing code or generated changelog is changed in the PR. CI results must be checked against the newly pushed commit.


apps/backend/src/data-marts/search/sources/report.source.ts
```diff
@@ -0,0 +1,154 @@
...
+    const pageIds = pageRows.map(report => report.id);
+    const reports = await this.reportRepo.find({
+      where: { id: In(pageIds), dataMart: { projectId, deletedAt: IsNull() } },
+      relations: { dataMart: true, dataDestination: true },
+      loadEagerRelations: false,
+    });
+    const reportById = new Map(reports.map(report => [report.id, report]));
+    const descriptors = pageIds
+      .map(id => reportById.get(id))
+      .filter((report): report is Report => report?.dataDestination !== undefined)
```
laskevych: 🟡 **Important (non-blocking):** change the guard to `report?.dataDestination != null`.

**Why:** TypeORM sets a missing `ManyToOne` to `null`, not `undefined` (`RawSqlResultsToEntityTransformer`). A soft-deleted destination is excluded by the automatic `dd.deletedAt IS NULL` join condition. So `dataDestination === null` passes this filter, and `toDescriptor` throws on `dataDestination.type`. `syncTypeProject` does not catch it, so one row stops report indexing for the whole project.

**Reachability:** narrow. `DeleteDataDestinationService` counts referencing reports through an inner join on `report.dataMart`, so only reports on a soft-deleted Data Mart slip through.

**Fix with it:** `syncTypeProject` has `if (page.descriptors.length === 0) break;`. A fully filtered page ends the sync early instead of following `nextCursor`.

**Test:** "report with a missing destination" covers both.

apps/backend/src/data-marts/search/search.facade.impl.ts
```diff
@@ -1,20 +1,40 @@
...
 import {
   SEARCH_SEMANTIC_ENGINE,
+  SearchableEntityType,
   SearchEngine,
   SearchFacade,
   SearchOptions,
   SearchResult,
 } from '../../common/search/search.facade';
 
+export function buildReportUiPath(projectId: string, dataMartId: string, reportId: string): string {
```
laskevych, line 12: 🟡 **Important (non-blocking):** replace `buildReportUiPath` + `new URL(...)` with `buildReportUrl()` from `common/helpers/data-mart-url.helper.ts`.

**Why:** this is the third copy of the same deep link:
1. `buildReportUrl()` produces the same URL and is used by the Google Sheets, Google Chat and email writers.
2. `joinPublicOrigin()` + `buildReportsUiPath()` in `ee/mcp/tools` use an identical join.
3. This one.

The helper documents `REPORT_ID_URL_PARAM` as kept in sync with `apps/web`. The new copy is outside that contract. `common/helpers` is importable from here without a layering inversion.

apps/backend/src/data-marts/search/sources/report.source.ts
```diff
@@ -0,0 +1,154 @@
...
+      where,
+      select: { id: true, createdAt: true },
+      loadEagerRelations: false,
+      order: { createdAt: 'ASC', id: 'ASC' },
+      take: limit,
+    });
+    if (pageRows.length === 0) return { descriptors: [], nextCursor: null };
+
+    const pageIds = pageRows.map(report => report.id);
+    const reports = await this.reportRepo.find({
```
laskevych, line 145: 💡 **Suggestion:** add an explicit `select` to this hydration query.

**Why:** without it the query loads every Report column (7 JSON configs plus `lastRunError`), the full `DataMart` row (`definition` and `schema` JSON) and the full `DataDestination` row, all for 6 scalars. It runs on every project sync, 50 reports per page. The sibling `data-destination.source.ts` already uses `select`.

```ts
select: {
  id: true, title: true, modifiedAt: true,
  dataMart: { id: true, title: true, projectId: true },
  dataDestination: { id: true, title: true, type: true },
},
```

apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
```diff
@@ -233,6 +233,10 @@ export class UpdateDataDestinationService {
         updatedEntity.id,
         command.projectId
       );
+      await this.advancedSearchIndexSync?.scheduleTypeProjectSync(
+        SearchableEntityType.REPORT,
```
laskevych: 💡 **Suggestion:** schedule the REPORT project sync only when `title` changed.

**Why:** both return paths fire it on every destination update: credential rotation, Drive folder, availability toggles, contexts. Only `dataDestination.title` is in the report document, so those passes cannot change any `docHash`. The same applies to `update-data-mart-title.service.ts` when the submitted title equals the stored one.

packages/api-client/src/search.ts
```diff
@@ -26,23 +34,42 @@ const SEARCH_ENTITY_TYPES = new Set<OWOXSearchEntityType>([
...
+    typeof value.title !== 'string' ||
+    !(typeof value.description === 'string' || value.description === null) ||
+    typeof value.finalScore !== 'number' ||
+    typeof value.kwScore !== 'number' ||
+    !(typeof value.vecScore === 'number' || value.vecScore === null)
+  ) {
+    return false;
+  }
+
+  return (
```
laskevych, line 71: 💡 **Suggestion:** in `parseSearchResults`, drop unrecognised or malformed entries and return the rest.

**Why:** today one bad element throws for the whole response. That is why adding `REPORT` breaks pinned SDK versions, as the changeset says. The same cost repeats for every future entity type, and for a newer SDK against an older backend.

**If strict is deliberate:** add a one-line comment saying so. The closed `entityType` union reads as the opposite intent.

apps/backend/src/data-marts/search/engine/tokenizer.ts
```diff
@@ -131,6 +133,13 @@ export function tokenize(text: string): Set<string> {
   );
 }
 
+export function tokenizePrompt(prompt: string, entityType: SearchableEntityType): string[] {
```
laskevych, line 136: 💡 **Suggestion:** move the `'report'` stop word to the source, as an optional `promptStopTokens` next to `scoringConfig` on `IndexableSource`.

**Why:** the tokenizer is a generic text utility, and it now imports the entity enum and has a per-type branch. Each new type with its own noun ("storage", "dashboard") adds another branch. `search-query-builder.ts` and `in-memory-paginated.search.ts` thread `entityType` through only to reach it.

apps/backend/src/data-marts/search/sources/report.source.ts
```diff
@@ -0,0 +1,154 @@
...
+  const richTextSlots: RichTextSlot[] = [
+    { kind: 'title', text: report.title },
+    ...contextTexts.map((text): RichTextSlot => ({ kind: 'context', text })),
+  ];
+
+  return {
+    entityType: SearchableEntityType.REPORT,
+    entityId: report.id,
+    projectId: dataMart.projectId,
+    title: report.title,
```
laskevych: 💡 **Suggestion:** apply the empty-title fallback once on the backend (on the descriptor or in `SearchFacadeImpl`).

**Why:** `Report.generateId` sets `title = ''` for Looker Studio reports. Only `SearchPage.tsx` falls back to the destination title. The web UI shows "Looker", while `/api/search` and `get_relevant_reports_by_prompt` return `"title": ""`. The descriptor also stores a `{ kind: 'title', text: '' }` slot that matches nothing.

apps/web/src/pages/search/SearchPage.tsx
```diff
@@ -1,40 +1,78 @@
...
     icon: ArchiveRestore,
-    to: entityId => `/data-destinations?id=${entityId}`,
+    to: result => `/data-destinations?id=${result.entityId}`,
+  },
+  REPORT: {
+    label: 'Report',
+    icon: FileText,
+    to: result =>
+      result.report
+        ? `/data-marts/${result.report.dataMart.id}/reports?reportId=${result.entityId}`
```
laskevych, line 47: 💡 **Suggestion:** import `REPORT_ID_URL_PARAM` from `features/data-marts/reports/list/model/hooks` instead of hardcoding `?reportId=`.

**Why:** `useReportSidesheet` and `DataMartDestinationsContent` read that constant. Importing it keeps the link and its consumer renamed together.

apps/backend/src/migrations/1787788800004-create-report-search-index-tables.ts
```diff
@@ -0,0 +1,95 @@
...
+      );
+    }
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await softDropTable(queryRunner, TRIGGER_TABLE);
+    await softDropTable(queryRunner, INDEX_TABLE);
+  }
+
+  private async ensureMysqlFullTextIndex(queryRunner: QueryRunner): Promise<void> {
```
laskevych, line 78: 💡 **Nit:** move `ensureMysqlFullTextIndex` to `migration-utils.ts`.

**Why:** it is a verbatim copy of the private method in `1782131671353-create-search-index-tables.ts`. `softDropTable` is already shared from that file. This prevents a third copy with the next index table.

apps/backend/src/data-marts/services/advanced-search-index-sync.service.ts
```diff
@@ -88,6 +91,10 @@ export class AdvancedSearchIndexSyncService {
     operation: ReindexOperation,
     projectId: string
   ): Promise<void> {
+    if (!entityId?.trim()) {
+      throw new Error('entityId must be a non-empty string');
```
laskevych, line 122: 💡 **Nit (optional):** delete `if (existing.status !== TriggerStatus.IDLE) return;` a few lines below.

**Why:** `findOne` already filters on `status: IDLE`, so the branch is unreachable. It reads as if a `PROCESSING` trigger can be returned and skipped.

laskevych (commented):
Reviewed the backend search source, index lifecycle, access predicate, migration, REST/OpenAPI/SDK contract, MCP tool and web rendering. No blocking findings; everything below is non-blocking.

Verified in particular:

- the access predicate composes the Data Mart and destination visibility filters on one query builder — both use `andWhere` with identical parameter values, so there is no clause reset or parameter collision, and `toRawBindableParams` correctly leaves `:projectId` to the outer query;
- the `report` word rule is applied consistently to SQL candidate selection (`buildDbSearchQuery`) and keyword scoring (`tokenizePrompt`), and indexed documents keep the word;
- `deleteOrphans` needs no `report.deletedAt` clause — `Report` is a hard-delete entity;
- the delete path uses `command.id` rather than `report.id`, which TypeORM clears on `remove()`;
- `ReportIndexableSource` is picked up by `SearchIndexDriftProcessor` through `INDEXABLE_SOURCES`, so existing reports are backfilled without a data migration, and the new entity file matches the `**/*.entity.{ts,js}` glob in `data-source-options.config.ts`;
- the migration timestamp follows the current head (`…800003`) and `report_search_index` mirrors the existing index tables including the MySQL full-text index;
- `PublicOriginService` is exported by the already-imported `CommonModule`, so the new facade dependency resolves.

The comments below are one latent `null` vs `undefined` guard, one duplicated URL builder, and a few efficiency/consistency/reuse notes.

apps/backend/src/data-marts/use-cases/update-data-mart-title.service.ts
```diff
@@ -53,6 +53,10 @@ export class UpdateDataMartTitleService {
       dataMart.id,
       command.projectId
     );
+    await this.advancedSearchIndexSync?.scheduleTypeProjectSync(
```
laskevych: 🟡 **Important (non-blocking):** on a title change, reindex the affected reports with `scheduleReindexMany(REPORT, ids)` instead of `scheduleTypeProjectSync`.

**Why:** on `main` the project queue had one caller, the hourly drift. For renames it has two problems:
1. **Dropped updates.** `enqueueTypeProjectSync` skips when a trigger is `IDLE`, `READY` or `PROCESSING`. A rename during a running sync that already passed the report's page enqueues nothing. The old title stays until the next drift tick.
2. **Latency.** `PROJECT_PROCESSING_BATCH_LIMIT = 1` at `15,45 * * * * *` is about 120 projects per hour per instance. Drift fills the same queue at `:00`. With N projects that have reports, a rename waits about N × 30 s. With N > 120 the queue never drains. The initial backfill follows the same maths.

**Fix:** select report ids by `dataMartId` / `dataDestinationId` and call `scheduleReindexMany` (entity queue, 2 per 2 s). This also resolves my earlier note about reindexing on non-title updates.

apps/backend/src/data-marts/use-cases/update-report.service.ts
```diff
@@ -187,6 +190,11 @@ export class UpdateReportService {
     report.uniqueCountConfig = nextUniqueCountConfig;
 
     const updatedReport = await this.reportRepository.save(report);
+    await this.advancedSearchIndexSync?.scheduleReindex(
```
laskevych, line 193: 💡 **Suggestion:** schedule this reindex only when `title` or the destination changed.

**Why:** the index document has 4 values: report title, Data Mart title, destination title, destination type. This call fires on every save (columns, filters, sort, owners, MCP `update_report`). `upsertDescriptor` calls `provider.embed()` before comparing `docHash`, so each save costs 1 embedding request for an identical document. If the provider fails on such a no-op, the row is rewritten with `embedding: null` and loses a valid vector.

**Alternative fix:** skip in `reindexEntity` when the stored `docHash` matches and the embedding is `READY`. That helps every entity type.

apps/backend/src/data-marts/search/schema/search-index.repository.ts
```diff
@@ -84,24 +85,40 @@ const ENTITY_CLASS_BY_TYPE: Record<SearchableEntityType, EntityTarget<SearchInde
...
-};
+function liveEntityExistsSql(entityType: SearchableEntityType, indexTable: string): string {
+  if (entityType === SearchableEntityType.REPORT) {
+    return `SELECT 1 FROM report r
+            JOIN data_mart dm ON dm.id = r.dataMartId
+            WHERE r.id = ${indexTable}.entity_id
+              AND dm.projectId = ${indexTable}.project_id
+              AND dm.deletedAt IS NULL`;
+  }
+  const entityTable = {
```
laskevych, line 106: 💡 **Suggestion:** move this map back to a typed module-level const: `Record<Exclude<SearchableEntityType, SearchableEntityType.REPORT>, string>`.

**Why:** the backend compiles with `noImplicitAny: false`, so this inline literal accepts a missing enum key. A fifth entity type breaks the build in the 3 sibling maps but not here. The result is `SELECT 1 FROM undefined e …` in `deleteOrphans` at runtime.

**Also:** add `JOIN data_destination dd … AND dd.deletedAt IS NULL` to the REPORT branch. The access predicate and `loadSearchableOne` require a live destination; this SQL does not. A report with a soft-deleted destination is never returned and never pruned.

apps/backend/src/data-marts/use-cases/delete-report.service.ts
```diff
@@ -47,6 +50,11 @@ export class DeleteReportService {
     const deletedEvent = new ReportDeletedEvent(report);
 
     await this.reportService.deleteReport(report);
+    await this.advancedSearchIndexSync?.scheduleDelete(
```
laskevych, line 53: 💡 **Suggestion (follow-up is fine):** call `scheduleTypeProjectSync(REPORT, projectId)` in `DeleteDataMartService` after `deleteAllByDataMartIdAndProjectId`.

**Why:** the PR description says bulk-delete cleanup is deferred. In one case it never happens:
1. `deleteOrphans(REPORT, projectId)` runs only inside a REPORT project sync.
2. Drift schedules that sync only for projects from `listProjectIds()`: projects with at least 1 report under a live Data Mart.
3. Delete the last such Data Mart, and the project leaves the list. Its `report_search_index` rows and embeddings stay forever.

**Impact:** storage only. Search never returns these rows (the predicate inner-joins `report`). An empty page still reaches `deleteOrphans`, so one line is enough.

apps/backend/src/migrations/1787788800004-create-report-search-index-tables.ts
```diff
@@ -0,0 +1,95 @@
...
+            new TableIndex({
+              name: `idx_${INDEX_TABLE}_project_entity`,
+              columnNames: ['project_id', 'entity_id'],
+            }),
+          ],
+        })
+      );
+    }
+
+    if (queryRunner.connection.options.type === 'mysql') {
```
laskevych, line 41: 🧪 **Test coverage (optional):** add this migration and `report_search_index` to `test/integration/advanced-search-mysql.integration.ts` (`migrations` list, table list, FULLTEXT assertion).

**Why:** this MySQL-only branch runs in no test. The new migration spec uses `better-sqlite3`. The MySQL suite registers only `CreateSearchIndexTables1782131671353`. Risk is low: the branch copies a covered one.

laskevych (approved):
Second pass on the same head — approving. No blocking findings. My earlier comments still stand as non-blocking. The additions below are about report-index freshness after parent renames (the one I'd prioritise), embedding cost on report saves, an exhaustiveness check that was lost in `liveEntityExistsSql`, and two small follow-ups.

apps/backend/src/data-marts/search/sources/report.source.ts
kalnik-a-a: **Criticality: MEDIUM / NON-BLOCKING — Fixed.**

Changed the guard to `!= null`. Empty hydrated pages now follow `nextCursor`, and repeated cursors fail explicitly instead of looping. SQLite regression tests cover the missing-destination case and continued pagination. Already-deleted Data Marts are excluded by the source query, so the missing-destination test deliberately exercises an inconsistent/live-parent case rather than assuming the ordinary deletion endpoint produces it.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/search/search.facade.impl.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Deferred.**

The duplication is real, but the helpers differ in encoding: search encodes the project and Data Mart path segments, whereas the shared helper currently interpolates them. I kept the working deep-link behavior in this fix. Consolidating the builders is a separate cleanup that should preserve that encoding; no current navigation failure was reproduced.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/search/sources/report.source.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Fixed.**

Added the explicit projection to the shared report hydration query: report ID/title/modifiedAt and the required parent IDs, titles, project and destination type. Both paginated loading and fresh-state reads use it. SQLite coverage checks the descriptor fields, and the inspected SQL no longer loads the full report configs, Data Mart definition/schema or destination credentials.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/use-cases/update-data-destination.service.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Fixed.**

Both destination update paths and the Data Mart title update now compare the submitted title with the stored value before scheduling report work. Credential/config/access-only changes and equal-title updates do not enqueue it. Actual renames now use a parent-scoped paginated job in the fast entity queue rather than a full-project sync; tests cover the changed and unchanged cases.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

packages/api-client/src/search.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Not changed.**

Strict response validation is intentional for this PR. Silently dropping malformed entries would turn a contract error into an apparently successful, incomplete result set. Skipping only unknown future entity types is a reasonable separate compatibility decision, but it would not repair SDK versions already installed by callers. The Report-aware SDK accepts existing entity types from an older backend; the breaking direction is an older pinned SDK receiving REPORT. The upgrade/explicit-filter warning remains in the changeset and PR description. I left the parser and optional explanatory code comment unchanged in this scoped fix.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/search/engine/tokenizer.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Not changed.**

Kept the current entity-specific prompt handling. Candidate selection and scoring use the same REPORT-only rule, and existing entity types retain the word `report`. Moving the configuration into sources would be an architectural cleanup rather than a fix for a reproduced behavior problem; another entity-specific rule would be a good point to revisit it.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/search/sources/report.source.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Fixed.**

Applied the destination-title fallback in the backend report descriptor. It now feeds the result title, title scoring slot and embedding text, so REST and MCP receive the same usable title as the UI. SQLite coverage checks the generated empty-title Looker report and consistency between single and batch loading.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/web/src/pages/search/SearchPage.tsx
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Deferred.**

Left the literal unchanged in this pass. It matches the current sidesheet parameter, and the existing direct-link tests cover opening the report. Sharing the constant is valid cleanup, but no behavior defect was found here; this follow-up stays focused on indexing correctness and unnecessary work.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/migrations/1787788800004-create-report-search-index-tables.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Deferred.**

The migration helper is duplicated, but I left the extraction out of this fix because it does not change current migration behavior. The MySQL suite has instead been extended to cover the report migration and its FULLTEXT index; actual MySQL execution remains outstanding, as noted in the validation section.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/services/advanced-search-index-sync.service.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Deferred.**

Agreed that this check is redundant after the IDLE predicate. It remains unchanged as an optional cleanup, and we do not rely on it for concurrency protection. Parent report jobs are now kept separate from ordinary REINDEX/DELETE jobs, and the actual report-index race is covered by conditional writes plus fresh-source verification.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/use-cases/update-data-mart-title.service.ts
kalnik-a-a: **Criticality: MEDIUM / NON-BLOCKING — Fixed with bounded parent jobs.**

Confirmed the dropped-update and rename-latency problems. Actual parent renames now create one REINDEX_REPORTS job in the fast entity queue, selecting only that parent's reports by an indexed keyset cursor. Each run handles one configured batch and persists its cursor, yielding between pages. Concurrent renames retain separate jobs; ordinary parent REINDEX/DELETE cannot overwrite them. This avoids loading every affected ID or creating one trigger per report, while removing rename dependence on the slow project queue. Reconciliation/backfill scheduling is unchanged.

REPORT writes also check the expected index state atomically and verify source data after writing. Regression tests cover both completion orders and A → B → A; retries are limited to three attempts. Partial failures preserve successful vectors, continue later pages and leave the final job in ERROR.

The new operation requires draining old API producers and search consumers before enabling the new version; the upgrade/rollback sequence is recorded in the PR description.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/use-cases/update-report.service.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Fixed in the indexer.**

Used the shared guard alternative: a matching project/document hash with a READY embedding returns before calling the provider. This covers all entity types and avoids adding field-change decisions to each caller. Changed documents and MISSING embeddings still reindex. Tests cover the no-op skip and reindex cases. For REPORT batches, a partial provider failure also preserves successfully generated vectors. An actual provider exception does not itself overwrite an existing row; the previous vector-loss case was a returned null embedding.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/search/schema/search-index.repository.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Deferred.**

Both are valid follow-ups. The currently supported entity types are covered, so the typed-map change protects future additions rather than fixing a present missing key. The missing live-destination condition can retain unused index rows, but candidate access filtering excludes those reports before search results are returned. I left both the map cleanup and orphan-retention improvement out of this correctness-focused follow-up; the storage limitation remains documented.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/use-cases/delete-report.service.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Deferred.**

Confirmed: after deleting the last live Data Mart with reports, the project can disappear from drift discovery and leave unreachable index rows indefinitely. This is a storage-retention issue; the live-report access predicate keeps those rows out of results. We explicitly kept bulk-delete cleanup deferred for this PR, and the description continues to state that limitation. No change was made to DeleteDataMartService.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/migrations/1787788800004-create-report-search-index-tables.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — Coverage added; MySQL run outstanding.**

Extended the existing MySQL integration suite with the REPORT migration/table and FULLTEXT assertion, a report-title prefix search with project isolation, the new progress/parent-index migration, and conditional batch-write conflicts. SQLite migration and repository tests pass. I could not execute the MySQL suite locally: the installed server crashed while initializing a separate empty test instance. The new assertions are committed, but live MySQL validation is still required; SQLite success is not being treated as proof for the MySQL-only branch.

Follow-up commit: [`79d316b72`](https://github.com/OWOX/owox-data-marts/commit/79d316b72dda617ccad53d38cd6c64c1b182be50).

apps/backend/src/data-marts/search/schema/search-index.repository.ts
```diff
@@ -177,6 +194,84 @@ export class SearchIndexRepository {
     );
   }
 
+  async upsertReportsIfUnchanged(
```
laskevych, line 197: 🟡 **Important (non-blocking):** run `test/integration/advanced-search-mysql.integration.ts` against a real MySQL before merge and paste the result here.

**Why:**
1. This batched `CASE … WHEN` conditional UPDATE is hand-built SQL, and production runs MySQL.
2. No CI job executes MySQL. The `test-integration.yml` matrix has no MySQL pattern, and the suite is `describe.skip` without a database.
3. Your reply says the MySQL run is outstanding.

**Impact if it fails on MySQL:** inserts still work, but every update of an existing `report_search_index` row throws. Renamed reports keep stale rows, and `REINDEX_REPORTS` triggers end in `ERROR`.

apps/backend/src/data-marts/search/indexing/search-indexer.service.ts
```diff
@@ -178,12 +228,138 @@ export class SearchIndexerService {
...
+            stale.map(({ descriptor }) => embeddingText(descriptor)),
+            {
+              inputType: 'search_document',
+            }
+          )
+        : [];
+    const rows: SearchIndexRow[] = [];
+    stale.forEach(({ descriptor, document, hash }, i) => {
+      const vector = vectors[i];
+      if (!vector) return;
```
laskevych: 💡 **Suggestion:** when the vector is `null`, still write the row with `embedding: null`, as `upsertDescriptor` does.

**Why:** for the other three types, an entity reindex during an embedding outage writes a `MISSING` row, so keyword search finds the new entity. For reports this line skips the row. A report created during an outage is unsearchable until a later sync succeeds. Report failures also count as `errors`, not `embedFailed`.

apps/backend/src/data-marts/search/indexing/search-indexer.service.ts
```diff
@@ -26,14 +27,25 @@ export class SearchIndexerService {
     private readonly registry: IndexableSourceRegistry,
     @Inject(EMBEDDING_PROVIDER) private readonly provider: EmbeddingProvider,
     private readonly repository: SearchIndexRepository,
-    @Inject(ADVANCED_SEARCH_CONFIG) private readonly config: AdvancedSearchConfig
+    @Inject(ADVANCED_SEARCH_CONFIG) private readonly config: AdvancedSearchConfig,
+    private readonly reportSource: ReportIndexableSource
```
laskevych, line 31: 💡 **Suggestion (follow-up is fine):** make the compare-and-set write path generic instead of a REPORT branch.

**Why:** the generic indexer now injects a concrete source and has two write paths:
1. Reports: conditional UPDATE plus post-write verification.
2. Data Marts, storages, destinations: read, check, then `upsertMany`. The read/write race described in the new code comment still exists there.

**Fix:** add an optional `loadSearchableByIds` to `IndexableSource` and route all types through `indexReports`-style logic.

laskevych (approved):
Re-checked head `79d316b72`. The fixes match the replies and I found no regression in code. One request before merge: a real MySQL run of the new batch UPDATE. Note for the managed rollout: during a rolling deploy, old pods treat `REINDEX_REPORTS` as a plain `REINDEX` of the parent and mark it `SUCCESS`. A rename in that window stays stale until the next hourly drift.

apps/backend/src/data-marts/search/schema/search-index.repository.ts
kalnik-a-a: **Criticality: MEDIUM / NON-BLOCKING — validation completed.**

Ran `test/integration/advanced-search-mysql.integration.ts` against a fresh, isolated MySQL 8.1.0 instance with one worker. The suite passed: Jest reports **1 suite / 9 tests passed**. One test returns early because vanilla MySQL has no `cloudsql_vector`, so the Cloud SQL-only `COSINE_DISTANCE` assertion was **not** executed.

The run exercised REPORT migrations, FULLTEXT/project isolation, parent pagination indexes, and the conditional batch UPDATE. The existing-row test verified that a stale write is rejected while an independent row in the same batch is updated. No SQL changes were needed. The PR description now records the live result; live Cloud SQL vector and Pub/Sub checks remain outstanding, and this does not add MySQL to CI.

apps/backend/src/data-marts/search/indexing/search-indexer.service.ts
kalnik-a-a: **Criticality: MEDIUM — fixed in `9332151e7f84`.**

A null vector now persists the current report document with `embedding: null` / `MISSING`, so new and renamed reports remain searchable by keywords. Later syncs retry the missing vector; project statistics count these failures in `embedFailed`, and parent batches report the actual failed-report count instead of marking the whole page as failed. Trigger failure signaling is preserved after the keyword row is saved.

The same conditional-write and freshness guards apply to missing vectors. A regression test verifies that a delayed null response cannot overwrite a vector already recovered by another worker. Tests also cover keyword discoverability, rename updates, recovery to `READY`, and skipping unchanged ready reports. The complete search suite passed: **24 suites / 471 tests**; backend build and scoped lint/format checks also passed.

apps/backend/src/data-marts/search/indexing/search-indexer.service.ts
kalnik-a-a: **Criticality: LOW / NON-BLOCKING — deferred from this PR.**

The observation is valid: Data Marts, storages and destinations still use the read/check/upsert path, which does not close the read/write race. This follow-up keeps the conditional write and post-write verification scoped to REPORT.

Generalizing that path would change the shared source contract and three additional entity implementations, each needing its own concurrency coverage. I have left that broader change for a separate follow-up to keep this PR focused on report search. No generic source API or write-path refactor is included in this commit.

PR #1611, opened 2026-09-15
feat(facebook): select conversions and conversion values by default

prokopetsv:
# Problem

New Data Marts built on Facebook Marketing insights endpoints did not include conversion data unless the user manually added the `conversions` and `conversion_values` fields. The default field set covered impressions, clicks, spend, and actions, but skipped conversions, which most advertisers need to measure campaign performance. The gap existed because the `defaultFields` arrays in the connector schema were never extended after these two fields were added to the insights field definitions.

Both fields also carried placeholder descriptions: the description of `conversions` was the literal string `conversions`. Making them user-facing defaults would have pushed that placeholder into the field picker, the Output Schema, and the metadata exposed to MCP and AI clients, where it gives an assistant nothing to distinguish them from `actions`.

# Solution

Add `conversions` and `conversion_values` to the `defaultFields` of the insights endpoints where Meta actually supports them, and fill in real descriptions.

- Edited `FieldsSchema.js`, the single source of default field selection for the connector.
- **Excluded `insights-by-link-url-asset`.** Meta's [Breakdowns reference](https://developers.facebook.com/docs/marketing-api/insights/breakdowns/) states that Dynamic Creative asset breakdowns support only `impressions`, `clicks`, `spend`, `reach`, `actions`, and `action_values`. Both fields stay in the schema and remain selectable by hand.
- Kept the alphabetical, normalized `defaultFields` ordering that #1247 introduced for `insights-by-country`.
- Replaced the placeholder descriptions with Meta's official definitions, plus a clause that distinguishes conversions from actions.
- Documented the breakdowns where Meta returns conversion data that is degraded rather than absent, so the limits are discoverable before someone reports on them.
- Left non-insights endpoints untouched, as they do not expose these fields.

Two known limits are documented rather than coded around, because both affect fields the connector already ships. On `insights-by-product-id`, Meta returns web conversions without the product breakdown and drops mobile conversions, so the value repeats per row and must not be summed. On `insights-by-region`, Meta no longer splits off-Meta conversions by region, which affects the existing `actions`/`action_values` defaults identically.

# Changes

- Added `conversions` and `conversion_values` to the defaults of nine insights endpoints: the base ad account insights endpoint, the breakdowns by age and gender, country, device platform, product ID, publisher platform and position, and region, and the ad set and campaign endpoints
- Left `insights-by-link-url-asset` defaults unchanged, since Meta's asset breakdowns do not support conversion metrics
- Restored the alphabetical `defaultFields` ordering for `insights-by-country`
- Replaced the placeholder descriptions for both fields across the ten active insights field files and in `ENDPOINTS_AND_FIELDS.md`
- Added a "Conversion Metrics and Breakdowns" section to `ENDPOINTS_AND_FIELDS.md` covering the link URL asset, product ID, and region limits
- Named both fields in `TROUBLESHOOTING.md` as the first to clear when Meta asks for less data
- Added changeset `6903-facebook-marketing-default-conversion-fields.md` with a minor version bump

# Test plan

- [x] `eslint` clean on `@owox/connectors`
- [x] 284 tests pass across 21 test files
- [x] markdownlint clean on all changed Markdown
- [x] Scripted check that every entry in every `defaultFields` array resolves against its endpoint's field schema, since `_prepareFields()` drops unknown fields silently
- [ ] Confirm live Meta responses for the product ID and region breakdowns against a real ad account


packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
```diff
@@ -122,7 +122,7 @@ var FacebookMarketingFieldsSchema = {
         "fields": adAccountInsightsFieldsByLinkUrlAsset,
         "breakdowns": ["link_url_asset"],
         'uniqueKeys': ["ad_id", "date_start", "date_stop"],
-        'defaultFields': ["account_id", "account_name", "campaign_id", "campaign_name", "adset_id", "adset_name", "ad_name", "impressions", "reach", "clicks", "spend", "cpc", "cpm", "ctr", "frequency", "actions", "action_values"],
+        'defaultFields': ["account_id", "account_name", "campaign_id", "campaign_name", "adset_id", "adset_name", "ad_name", "impressions", "reach", "clicks", "spend", "cpc", "cpm", "ctr", "frequency", "actions", "action_values", "conversions", "conversion_values"],
```
laskevych: 🟡 **Important (blocking until verified):** `ad-account/insights-by-link-url-asset` uses a Dynamic Creative asset breakdown, and Meta's [Breakdowns reference](https://developers.facebook.com/docs/marketing-api/insights/breakdowns/) states that asset breakdowns *"only support a limited set of metrics: impressions, clicks, spend, reach, actions, and action_values"* — `conversions` and `conversion_values` are not in that list.

The per-breakdown field files are copies of the base insights list (the only difference between `ad-account-insights-fields.js` and `ad-account-insights-fields-by-link-url-asset.js` is the `link_url_asset` key itself), so the presence of these two fields in the schema file does not prove API support. `_prepareFields()` in `Source.js` only filters out fields missing from the schema, so both will be sent in `&fields=` alongside `&breakdowns=link_url_asset`. Depending on Meta's behaviour this either fails the first run of the default configuration with error 100, or returns two permanently empty columns.

Could you run this endpoint once against a real ad account with the new defaults and confirm the response? If the fields are rejected, please exclude them from this endpoint's `defaultFields`.

laskevych (commented):
Small, focused change. I verified that all ten insights field files really define `conversions` and `conversion_values` (both `DATA_TYPES.ARRAY`), that `defaultFields` only drives the pre-selection in the connector setup UI (so existing Data Marts are untouched and no migration is needed), and that the changeset follows the repository convention.

One item should be cleared before merge — see the comment on `insights-by-link-url-asset`. The rest are non-blocking.

---

**💡 Suggestion (all `ad-account-insights-fields*.js` files):** both new default fields carry placeholder descriptions — `'description': 'conversions'` and `'description': 'conversion_values'` — unlike the neighbouring `actions`/`action_values`, which have real explanations. Now that these two ship selected by default in every insights Data Mart, the placeholder text propagates to the field-picker tooltip, the Output Schema, and the metadata exposed to MCP/AI clients, where "conversions" as the description of `conversions` gives an assistant nothing to distinguish it from `actions`. Since this PR is what makes them user-facing defaults, it seems like a good moment to fill in proper descriptions.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
```diff
@@ -100,7 +100,7 @@ var FacebookMarketingFieldsSchema = {
         "fields": adAccountInsightsFieldsByCountry,
         "breakdowns": ["country"],
         'uniqueKeys': ["ad_id", "date_start", "date_stop", "country"],
-        'defaultFields': ["account_currency", "account_id", "account_name", "ad_name", "adset_id", "adset_name", "campaign_id", "campaign_name", "clicks", "impressions", "inline_link_clicks", "reach", "spend"],
+        'defaultFields': ["account_currency", "account_id", "account_name", "ad_name", "adset_id", "adset_name", "campaign_id", "campaign_name", "clicks", "impressions", "inline_link_clicks", "reach", "spend", "conversions", "conversion_values"],
```
laskevych: 🟡 **Important (non-blocking):** in #1247 ("Normalize defaultFields across ad connectors") the by-country defaults were deliberately changed from the shared ad-level set to an alphabetically ordered, array-free list, and `actions`/`action_values` were dropped. This PR appends two ARRAY fields to that list without restoring the action fields or keeping the alphabetical order.

The result is a default table whose only nested columns are `conversions`/`conversion_values`, with no `actions` context to reconcile them against. Was the normalization intentionally reverted for this endpoint? If conversions belong here, `actions`/`action_values` arguably do too; if not, it may be better to leave by-country as normalized.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
```diff
@@ -177,7 +177,7 @@ var FacebookMarketingFieldsSchema = {
         "fields": adAccountInsightsFieldsByCampaign,
         "level": "campaign",
         "uniqueKeys": ["campaign_id", "date_start", "date_stop"],
-        "defaultFields": ["account_id", "campaign_id", "campaign_name", "impressions", "reach", "frequency", "clicks", "spend", "cpm", "cpp"],
+        "defaultFields": ["account_id", "campaign_id", "campaign_name", "impressions", "reach", "frequency", "clicks", "spend", "cpm", "cpp", "conversions", "conversion_values"],
```
laskevych, line 180: 💡 **Suggestion:** `insights-by-campaign` was introduced in #1265 with a deliberately lean, scalar-only default set, while its sibling `insights-by-adset` (added in the same commit) included `actions`/`action_values`. After this change the campaign endpoint has `conversions`/`conversion_values` but still no action fields, so ad-set and campaign Data Marts created side by side expose different metric shapes and cannot be compared on the same columns. Worth either adding `actions`/`action_values` here as well, or keeping the campaign defaults scalar-only.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
```diff
@@ -133,7 +133,7 @@ var FacebookMarketingFieldsSchema = {
         "fields": adAccountInsightsFieldsByProductId,
         "breakdowns": ["product_id"],
         'uniqueKeys': ["ad_id", "date_start", "date_stop", "product_id"],
-        'defaultFields': ["account_id", "account_name", "campaign_id", "campaign_name", "adset_id", "adset_name", "ad_name", "impressions", "reach", "clicks", "spend", "cpc", "cpm", "ctr", "frequency", "actions", "action_values"],
+        'defaultFields': ["account_id", "account_name", "campaign_id", "campaign_name", "adset_id", "adset_name", "ad_name", "impressions", "reach", "clicks", "spend", "cpc", "cpm", "ctr", "frequency", "actions", "action_values", "conversions", "conversion_values"],
```
laskevych, line 136: ⚠️ **Data quality (non-blocking):** Meta classifies `product_id` as a Type 2 breakdown restriction — off-Meta web metrics are still returned but *without* the breakdown value, and mobile metrics are not returned at all. `conversions`/`conversion_values` are pixel/off-Meta metrics, so on this endpoint they will not actually be split per product: the same value repeats on every product row, and summing across rows multiplies the total. Nothing in the output schema signals that the column is not product-scoped. If these fields stay in the defaults here, this is worth calling out in `ENDPOINTS_AND_FIELDS.md`.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
```diff
@@ -78,7 +78,7 @@ var FacebookMarketingFieldsSchema = {
         "documentation": "https://developers.facebook.com/docs/marketing-api/reference/ad-account/insights",
         "fields": adAccountInsightsFields,
         'uniqueKeys': ["ad_id", "date_start", "date_stop"],
-        'defaultFields': ["account_id", "account_name", "campaign_id", "campaign_name", "adset_id", "adset_name", "ad_name", "impressions", "reach", "clicks", "spend", "cpc", "cpm", "ctr", "frequency", "actions", "action_values"],
+        'defaultFields': ["account_id", "account_name", "campaign_id", "campaign_name", "adset_id", "adset_name", "ad_name", "impressions", "reach", "clicks", "spend", "cpc", "cpm", "ctr", "frequency", "actions", "action_values", "conversions", "conversion_values"],
```
laskevych, line 81: 📝 **Note (non-blocking):** conversion metrics are among the heaviest insights fields for Meta to compute, and this PR adds two of them to the defaults of eight ad-level endpoints. The connector's own `TROUBLESHOOTING.md` advises removing fields when Meta answers *"Please reduce the amount of data you're asking for"* — so for large accounts doing a wide backfill, the out-of-the-box configuration becomes more likely to hit that limit, and the documented remedy is to undo the new defaults by hand. Fine to accept as a deliberate trade-off; just worth a line in the troubleshooting table.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
prokopetsv: **Fixed in e5a7ebad1 — both fields removed from this endpoint's `defaultFields`.**

Being upfront about how I resolved it: **I did not run the endpoint against a real ad account.** I have no Meta credentials in this environment, so I took the fallback you specified rather than claiming a verification I did not perform.

What I did verify:

1. Meta's [Breakdowns reference](https://developers.facebook.com/docs/marketing-api/insights/breakdowns/) states verbatim: *"All Dynamic Creative asset breakdowns only support a limited set of metrics: impressions, clicks, spend, reach, actions, action_values."* It applies the rule to `ad_format_asset`, `body_asset`, `call_to_action_asset`, `description_asset`, `image_asset`, `link_url_asset`, `title_asset`, and `video_asset`. Neither conversion field is in the supported list.
2. Your point about the field files is exactly right. Diffing the key sets, the only difference between `ad-account-insights-fields.js` and `ad-account-insights-fields-by-link-url-asset.js` is the `link_url_asset` key itself, so the schema proves nothing about API support.
3. `_prepareFields()` in `Source.js` filters only on schema membership and breakdown exclusion, so both fields would have gone out in `&fields=` next to `&breakdowns=link_url_asset`, as you described.

Since the documented metric list is explicit and this ships as the out-of-the-box default, shipping it unverified was the wrong risk to take. Both fields stay in the schema and remain selectable by hand, so nobody loses access to them.

I also documented the limit in `ENDPOINTS_AND_FIELDS.md` under a new "Conversion Metrics and Breakdowns" section, so the next person does not re-add them.

If you do have an account handy to confirm the live behaviour, I am happy to follow up. Meta's docs alone do not tell us whether it hard-fails with error 100 or silently returns empty columns.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
prokopetsv: **Verified, and fixed in e5a7ebad1 — alphabetical order restored.**

I checked `353810d2b` ("Normalize defaultFields across ad connectors", #1247) and it is exactly as you describe. It deliberately replaced the shared ad-level set with an alphabetically ordered, array-free list and dropped `actions`/`action_values`.

To answer your question directly: **no, the normalization was not intentionally reverted.** I appended the two fields to the end of every insights endpoint mechanically and did not notice that by-country had a different convention. That is my oversight.

The list is now back in alphabetical order, with the two fields in their correct position:

```js
["account_currency", "account_id", "account_name", "ad_name", "adset_id", "adset_name",
 "campaign_id", "campaign_name", "clicks", "conversion_values", "conversions",
 "impressions", "inline_link_clicks", "reach", "spend"]
```

On the array-free property, I kept the conversion fields here deliberately. The task behind this PR is to ship conversion data by default across the insights endpoints, and country is one of the more common places people report conversions. Meta lists neither Type 1 nor Type 2 restrictions for the `country` breakdown, so the data should be genuine there.

On restoring `actions`/`action_values`: I deliberately did **not** do that. #1247 dropped them on purpose, and re-adding them is a product decision about the default table shape rather than something this PR should decide quietly. Happy to open a follow-up if you think the reconciliation context is worth the two extra ARRAY columns.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
prokopetsv: **Verified. No code change, and I want to explain why rather than just decline.**

I checked `4f3848677` (#1265) and your reading is correct. `insights-by-adset` and `insights-by-campaign` landed in the same commit, with ad set getting `actions`/`action_values` and campaign deliberately scalar-only.

I kept `conversions`/`conversion_values` on the campaign endpoint, because shipping conversion data by default across insights endpoints is the point of this PR, and omitting the campaign level would make it the one endpoint where the feature does not apply.

I did not add `actions`/`action_values`. That reopens the deliberate #1265 decision about campaign default shape, and it changes the default table for a reason unrelated to conversions. It felt wrong to fold that into a PR whose stated scope is two fields.

So I have narrowed the asymmetry rather than removed it. Ad set and campaign now share the conversion columns, and still differ on the action columns.

If you would rather they match completely, I am glad to add `actions`/`action_values` here, or to open a separate PR that makes the campaign and ad set defaults converge on purpose. Your call. I just did not want to make it silently.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
prokopetsv: **Confirmed against Meta's docs, and documented in e5a7ebad1.**

Meta classifies `product_id` as a Type 2 breakdown restriction and states: *"Off-Meta web metrics will continue to be returned from the API, however will not contain the breakdown value. The mobile metrics will not be returned anymore when queried with these breakdowns."* That matches your description, including the double-counting hazard when someone sums across product rows.

I added a "Conversion Metrics and Breakdowns" section to `ENDPOINTS_AND_FIELDS.md` that spells this out, with the explicit warning not to sum the column across rows.

**One thing your review did not cover, which I found while checking this.** Meta lists `region` as a **Type 1** breakdown, defined as *"the following breakdowns will no longer be available for off-Meta action metrics"*, alongside `dma` and the two `hourly_stats_*` breakdowns. Since `conversions`/`conversion_values` are largely off-Meta pixel metrics, `insights-by-region` has the same class of problem: expect empty or partial values.

I documented region rather than removing the fields there, for two reasons:

1. `insights-by-region` **already** ships `actions`/`action_values` in its defaults today, and the Type 1 restriction hits those identically. Removing conversions while keeping actions would be inconsistent, and would not fix the existing exposure.
2. Unlike the asset breakdown, there is no documented list that excludes these fields outright, so this degrades data rather than rejecting the request.

Say the word if you would rather I drop them from by-region too. That is a defensible reading, and it would need the same treatment for `actions`/`action_values` to be coherent.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
prokopetsv: **Fair, and added in e5a7ebad1.**

I extended the existing row for `Please reduce the amount of data you're asking for` in `TROUBLESHOOTING.md` rather than adding a new one, since it is the same failure mode:

> Reduce the date range, fields, or breakdowns. Insights Data Marts select `conversions` and `conversion_values` by default. Both cost Meta a lot to compute, so clear them first if you do not report on them. Then lower **API Page Limit** in **Advanced settings** and rerun.

That names the two fields explicitly, so someone hitting the limit knows which defaults to clear first instead of guessing across the whole field list.

You are right that this is a deliberate trade-off. Conversion data is the common reason people connect Meta at all, so paying for it by default seems correct, as long as the escape hatch is documented instead of something users have to work out themselves.

prokopetsv:
Thanks for the review. It caught a real defect, and one of the items changed what this PR ships. Everything is addressed in e5a7ebad1.

## The blocking item

`insights-by-link-url-asset` no longer gets the two fields. Meta's docs state that Dynamic Creative asset breakdowns support only `impressions`, `clicks`, `spend`, `reach`, `actions`, and `action_values`.

To be explicit, because you asked for a live check: **I did not run the endpoint against a real ad account.** No Meta credentials are available to me here. Rather than report a verification I did not perform, I took the fallback you offered and excluded the fields. Details in the thread.

## Your description suggestion

Taken. This was the most valuable non-blocking item, since the placeholders were about to become user-facing metadata.

Both fields now carry Meta's official definitions, with a clause that distinguishes them from `actions`:

| Field | Description |
| --- | --- |
| `conversions` | The total number of conversions attributed to your ads. Counts conversion events only, unlike actions, which also counts engagement and clicks. |
| `conversion_values` | The total value of all conversions attributed to your ads. Covers the same conversion events as the conversions field. |

Worth noting that Meta documents `conversion_values` and `action_values` with **identical** text ("The total value of all conversions attributed to your ads"), which is not much help to an assistant trying to tell them apart. The trailing clause on each is mine, derived from the documented scope difference, not invented behaviour.

Applied to all ten `ad-account-insights-fields*.js` files and to the matching rows in `ENDPOINTS_AND_FIELDS.md`, which carried the same placeholders.

I left the three inactive files alone: `ad-group-insights-fields.js`, `ad-campaign-insignts-fields.js`, and `ad-campaign-group-insights-fields.js`. Their endpoints are commented out in `FieldsSchema.js`, so the "these are user-facing now" rationale does not apply. Easy to include if you would rather avoid the drift.

## One item you did not flag

Meta lists `region` as a **Type 1** breakdown, which it defines as no longer available for off-Meta action metrics. `insights-by-region` therefore has the same data-quality problem you raised for `product_id`.

I documented it instead of removing the fields, because by-region already ships `actions`/`action_values`, which the restriction hits identically. Reasoning is in the `product_id` thread, and I am happy to change course.

## Everything in this push

- Removed `conversions`/`conversion_values` from `insights-by-link-url-asset` defaults
- Restored the alphabetical `defaultFields` order for `insights-by-country` from #1247
- Replaced the placeholder descriptions in ten field files and in `ENDPOINTS_AND_FIELDS.md`
- Added a "Conversion Metrics and Breakdowns" section documenting the link URL asset, product ID, and region limits
- Named both fields in `TROUBLESHOOTING.md` as the first to clear when Meta asks for less data
- Updated the changeset to match

Checks: `eslint` clean on `@owox/connectors`, 284 tests pass across 21 files, markdownlint clean. I also scripted a check that every entry in every `defaultFields` array resolves against its endpoint's field schema, since `_prepareFields()` drops unknown fields silently. All ten insights endpoints resolve.

## Two open questions for you

1. **`actions`/`action_values` on `insights-by-campaign`** — I did not add them, to avoid quietly reopening the #1265 decision inside a conversions PR. Want them added here, or in a follow-up that makes the ad set and campaign defaults converge deliberately?
2. **`insights-by-region`** — document the Type 1 limit as I have done, or drop the conversion fields there as well?

Neither blocks merge from my side, but both are yours to decide.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
laskevych: **Answering your region question directly: documenting it is fine, I would not hold the PR for it. But the symmetry argument does not quite hold, and the doc row currently overstates it.**

I confirmed `region` in Meta's list: Type 1 is *"region, dma, hourly_stats_aggregated_by_audience_time_zone, hourly_stats_aggregated_by_advertiser_time_zone"*, and Type 1 means *"The Insights API will not return unsupported off-Meta metrics."* Good catch — that was not in my review.

Where I read it differently is argument 1. Type 1 drops the **off-Meta** action metrics, not the whole field:

- `actions`/`action_values` are mixed metrics. Their on-Meta rows — engagement, link clicks, video views, on-site events — still come back under a region breakdown, so those columns stay useful there.
- `conversions`/`conversion_values` count conversion events, which are predominantly pixel/app/offline, i.e. off-Meta. Under region they are not just "partial" — they are systematically undercounted, with at most the on-site conversion events surviving.

So keeping `actions` on by-region and dropping `conversions` would not have been inconsistent. That said, your argument 2 is the one that decides it for me: unlike the asset breakdown there is no documented exclusion list, the request is not rejected, and the behaviour is account-dependent. Shipping them documented is a reasonable call, and I would rather not remove a field on an inference about which events a given advertiser tracks.

One follow-up on the doc row, since it is the only place a user learns this:

> Meta no longer splits off-Meta conversions by region. Expect empty or partial values. The same limit applies to `actions` and `action_values`.

*"The same limit applies"* reads as if those columns break too, which would push people away from fields that still work there. Something closer to:

> Meta does not return off-Meta conversions with a region breakdown, so expect these columns to be empty or heavily undercounted. `actions` and `action_values` lose their off-Meta rows for the same reason, but keep their on-Meta ones.

Non-blocking — take it or leave it.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/ad-account-insights-fields.js
```diff
@@ -112,11 +112,11 @@ var adAccountInsightsFields = {
     'type': DATA_TYPES.NUMBER
   },
   'conversion_values': {
```
laskevych, line 114: 💡 **Suggestion (non-blocking):** the new `conversions` description does exactly what it needs to — it names `actions` and says how the two differ. `conversion_values` does not get the same treatment: its first sentence is verbatim the description `action_values` already carries in these files.

```js
'action_values':     'The total value of all conversions attributed to your ads.'
'conversion_values': 'The total value of all conversions attributed to your ads. Covers the same conversion events as the conversions field.'
```

The disambiguation is in the second sentence, and field-picker tooltips are where it is most likely to be cut off — so the two ARRAY value columns can still read as identical at the exact moment someone is choosing between them. Mirroring the `conversions` wording would fix it, e.g. *"The total value of the conversions attributed to your ads. Covers conversion events only, unlike action_values, which also values engagement and clicks."*

Same text appears in all ten insights field files and in the `ENDPOINTS_AND_FIELDS.md` table (line 327), so it would be one find-and-replace.

laskevych (approved):
Approving. `e5a7ebad1` resolves the one item I had flagged as blocking, and the rest of the round is handled well.

Re-verified on the head commit:

- **Link URL Asset** — both fields are out of `defaultFields` and still present in `ad-account-insights-fields-by-link-url-asset.js`, so they stay selectable by hand. That was the right call to make without live credentials, and saying so plainly instead of claiming a verification you did not run is the more useful answer.
- **by-country** — order is genuinely alphabetical again (`clicks`, `conversion_values`, `conversions`, `impressions`, …). I also checked Meta's breakdown lists: `country` appears under neither Type 1 nor Type 2, so your reasoning for keeping the fields there holds.
- **by-campaign** — accepted as explained. Keeping the #1265 decision about campaign default shape out of a two-field PR is the right scope call; the remaining `actions`/`action_values` asymmetry is fine as a separate discussion.
- **Docs** — the new "Conversion Metrics and Breakdowns" section and the extended `TROUBLESHOOTING.md` row both land where a user would actually look. `ENDPOINTS_AND_FIELDS.md` is hand-maintained, so nothing regenerates over them.
- CI is green across all checks on `e5a7ebad1`.

I replied in the product ID thread about `region` — short version: documenting it is fine and does not block, but the doc row slightly overstates the effect on `actions`/`action_values`.

One small non-blocking note inline on the new descriptions. Neither it nor the region wording needs to gate the merge.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/FieldsSchema.js
prokopetsv: **You are right, and I withdraw argument 1. Reworded in 632b8e138.**

Your reading of Type 1 is the correct one. I collapsed "the field is affected" into "the field is equally affected", and those are not the same thing. `actions`/`action_values` are mixed metrics whose on-Meta rows — engagement, link clicks, video views, on-site events — survive a region breakdown untouched. `conversions`/`conversion_values` are predominantly pixel, app, and offline events, so under region they lose most of their substance rather than a slice of it.

So my symmetry argument was wrong, and keeping `actions` while dropping `conversions` would have been defensible after all. Good thing it was not the argument that decided it.

The doc row now reads:

> Meta does not return off-Meta conversions with a region breakdown. Expect these columns to be empty or heavily undercounted. `actions` and `action_values` lose their off-Meta rows for the same reason, but keep their on-Meta ones.

That is your wording, with the first sentence split in two to stay under the repository's 20-word sentence limit for docs. It keeps the part that matters: these columns are not merely "partial", and the action columns still work here.

Thanks for pushing back on this rather than letting the doc row ship as written. Steering people away from fields that still work would have been the worse outcome of the two.

packages/connectors/src/Sources/FacebookMarketing/MarketingAPIReference/ad-account-insights-fields.js
prokopetsv: **Valid, and fixed in 632b8e138 — though I went a step further than your suggested text, and want to flag why.**

You are right that `conversion_values` was the weak one. Its first sentence was verbatim `action_values`, so the two ARRAY value columns were indistinguishable exactly where someone is choosing between them.

Your proposed wording keeps the disambiguation in the **second** sentence:

> The total value of the conversions attributed to your ads. Covers conversion events only, unlike `action_values`, which also values engagement and clicks.

By your own argument that is still exposed to truncation. Sentence one differs from `action_values` only by "the" instead of "all", which is not enough to tell them apart in a clipped tooltip. So I pulled the distinction into the first sentence:

```js
'action_values':     'The total value of all conversions attributed to your ads.'
'conversion_values': 'The total value of the conversions attributed to your ads, counting conversion events only. Unlike action_values, it excludes engagement and clicks.'
```

Now the clause that separates them survives a cut after the first sentence, and the full text still names `action_values` explicitly.

One small wording departure: I avoided *"which also values engagement and clicks"*. `action_values` covers those action types in scope, but they rarely carry a monetary value, so "values engagement and clicks" risks implying a number that will almost always be absent. "Excludes engagement and clicks", stated from the `conversion_values` side, says the same thing without that implication.

Applied to all ten insights field files and to line 327 of `ENDPOINTS_AND_FIELDS.md`. I left `conversions` alone, since you confirmed it already does its job.

PR #1612, opened 2026-09-15
feat(destinations): add AI assistant promo card to Destinations tab

artemfenoga:
## Summary
Analysts publishing a Data Mart can now discover the option to connect **Claude** or **ChatGPT** right from the Destinations tab — even after they've already set up Google Sheets, Looker Studio, or any other destination. Previously this offer only appeared before any destination existed, so it went unnoticed by everyone who already had a working setup.

The new card sits among your existing destinations, looks and behaves like them (you can collapse it if you're not interested), and offers the same one-click path to Claude/ChatGPT plus the MCP setup guide used elsewhere in the product — so you can start asking questions about your data in plain language, with answers pulled straight from your published Data Mart instead of guesses.


apps/web/src/features/data-marts/reports/list/components/AiAssistantDestinationCard/AiAssistantDestinationCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * backed by a DataDestination — it points analysts to the MCP connection
+ * (Claude/ChatGPT) as another way to consume this Data Mart's published data.
+ * Collapsible like a real card, so a user not interested can fold it once; the
+ * choice persists per-browser in localStorage under a fixed name shared across
+ * all Data Marts.
+ */
+export function AiAssistantDestinationCard() {
+  return (
+    <div className='flex flex-col gap-0.5' data-testid='aiAssistantDestCard'>
+      <CollapsibleCard name={AI_ASSISTANT_CARD_STORAGE_NAME} collapsible defaultCollapsed={false}>
```
laskevych: 🟡 **Important (non-blocking):** the collapsed state is persisted under one fixed key (`collapsed-card-ai-assistant-destination-card`), so a single collapse hides the card on **every** Data Mart, permanently. A user who folds it once on Data Mart A will never see the promo body on Data Marts B, C… — including after the card's copy or offer changes later. Real `DestinationCard`s key their state on `destination.id`, so users have no reason to expect this one to be global.

The JSDoc says this is deliberate — if that's the product decision, fine as is; otherwise keying it per Data Mart (`ai-assistant-destination-card-${dataMartId}`) would match the surrounding cards.

apps/web/src/features/data-marts/reports/list/components/AiAssistantDestinationCard/AiAssistantDestinationCard.tsx
```diff
@@ -0,0 +1,43 @@
+import { Bot } from 'lucide-react';
+import {
+  CollapsibleCard,
+  CollapsibleCardContent,
+  CollapsibleCardHeader,
+  CollapsibleCardHeaderTitle,
+} from '../../../../../../shared/components/CollapsibleCard';
+import { ConnectAiAssistantPromoActions } from '../../../../../../pages/data-marts/reports/ConnectAiAssistantPromoActions';
```
laskevych: 💡 **Suggestion (non-blocking):** this is a `features/` module importing from `pages/`, and `ConnectAiAssistantPromoActions` now has three call sites — `DataMartReportsPage`, `EmptyDataMartDestinationsState`, and this card — two of them in `features/`. It's purely presentational with no side effects, so `shared/components/` would fit it better: same direction as `PromoBlock` / `InviteTeammatesCard`, and it drops the six-level `../../../../../../pages/...` path from two feature modules.

apps/web/src/features/data-marts/reports/list/components/AiAssistantDestinationCard/AiAssistantDestinationCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+          </CollapsibleCardHeaderTitle>
+        </CollapsibleCardHeader>
+
+        <CollapsibleCardContent>
+          <div className='flex flex-col items-center gap-3 pb-8 text-center'>
+            <p className='text-muted-foreground max-w-xl text-sm'>
+              Ask in plain language and get answers pulled straight from your Data Marts, not
+              guesses. Connect via Claude or ChatGPT — whichever your team already uses.
+            </p>
+            <ConnectAiAssistantPromoActions align='center' />
```
laskevych: 💡 **Suggestion (non-blocking):** every other promo on this tab tags its docs link (`?utm_source=owox_data_marts&utm_medium=dm_page_destinations_tab&utm_campaign=…`), including the MCP tab in the empty state, but `MCP_SETUP_GUIDE_URL` inside the shared actions component is untagged. Clicks from this new placement are therefore indistinguishable from the Reports page and empty-state promos, which makes it hard to tell whether this card actually drives MCP setups — the exact thing the PR is meant to improve. An optional `utmCampaign` prop on the shared component, or pairing the card with an `InviteTeammatesCard` the way the sibling promos do, would cover it.

apps/web/src/features/data-marts/reports/list/components/AiAssistantDestinationCard/AiAssistantDestinationCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <div className='flex flex-col gap-0.5' data-testid='aiAssistantDestCard'>
+      <CollapsibleCard name={AI_ASSISTANT_CARD_STORAGE_NAME} collapsible defaultCollapsed={false}>
+        <CollapsibleCardHeader>
+          <CollapsibleCardHeaderTitle icon={Bot} subtitle='Claude, ChatGPT'>
+            AI Assistants
+          </CollapsibleCardHeaderTitle>
+        </CollapsibleCardHeader>
+
+        <CollapsibleCardContent>
+          <div className='flex flex-col items-center gap-3 pb-8 text-center'>
```
laskevych: 🧹 **Nit (non-blocking):** `pb-8` hardcodes the bottom spacing that every neighbouring destination card gets from `<CollapsibleCardFooter />` (`py-4`). The card ends up slightly taller than its neighbours, and a future change to the shared footer spacing will skip this one. If the extra room is a deliberate choice for the centered promo layout, ignore this.

apps/web/src/pages/data-marts/reports/ConnectAiAssistantPromoActions.tsx
```diff
@@ -1,22 +1,41 @@
...
+ * Actions for the "Get answers in Claude or ChatGPT" promo block: outline
+ * buttons (icon + label) linking to the OWOX Data Marts connector in Claude
+ * and ChatGPT, wrapping onto a second line on narrow viewports, plus a link
+ * to the MCP setup guide. Every link opens in a new tab. Purely
+ * presentational, no side effects.
  */
-export function ConnectAiAssistantPromoActions() {
+export function ConnectAiAssistantPromoActions({
+  align = 'start',
+}: ConnectAiAssistantPromoActionsProps = {}) {
```
laskevych: 🧹 **Nit (non-blocking):** the `= {}` default on the props parameter is unreachable — React always passes a props object, so the inner `align = 'start'` already covers `<ConnectAiAssistantPromoActions />`.

laskevych (approved):
Approving — no blocking issues found.

Checked: the visibility gate (`dataDestinations.length > 0 && isPublished`) matches the neighbouring Sheets upsell and the empty-state promo, and the card cannot be shown together with the empty-state MCP tab (mutually exclusive render branches); `subtitle` is a real `CollapsibleCardHeaderTitle` prop and renders only while collapsed; the fixed storage name cannot collide with real destination cards (those key on the destination UUID); the new barrel export introduces no import cycle; the `align='start'` default keeps the Reports page and empty-state promo rendering unchanged; the changeset follows the existing convention.

Left five non-blocking comments inline (one worth a product confirmation, the rest suggestions/nits).

apps/web/src/features/data-marts/reports/list/components/AiAssistantDestinationCard/AiAssistantDestinationCard.tsx
artemfenoga: Good catch, fixed in 236dafb7a's follow-up — thanks for flagging it before it shipped. The collapse-state key now includes the Data Mart's id (ai-assistant-destination-card-${dataMartId}), so folding the card on one Data Mart no longer collapses it everywhere — it now behaves the same as real DestinationCards, which key on destination.id. Updated the JSDoc too, since "shared across all Data Marts" was never actually a confirmed product decision — just an implementation shortcut on my end.

apps/web/src/features/data-marts/reports/list/components/AiAssistantDestinationCard/AiAssistantDestinationCard.tsx
artemfenoga: Good suggestion — moved ConnectAiAssistantPromoActions to shared/components/ConnectAiAssistantPromoActions (same pattern as PromoBlock/InviteTeammatesCard) and updated all three call sites. Agreed it shouldn't have lived under pages/ once features/ started depending on it.

apps/web/src/features/data-marts/reports/list/components/AiAssistantDestinationCard/AiAssistantDestinationCard.tsx
artemfenoga: Good catch — added optional utmMedium/utmCampaign props to the shared component and tagged all three call sites consistently with the surrounding docs links: dm_page_destinations_tab/ai_assistant_card for this new card, dm_page_destinations_tab/empty_state for the empty-state MCP tab (matching its sibling InviteTeammatesCard), and project_reports_page/reports_list for the Reports page. Claude/ChatGPT buttons stay untagged since those are external platforms whose analytics we don't see.

apps/web/src/features/data-marts/reports/list/components/AiAssistantDestinationCard/AiAssistantDestinationCard.tsx
artemfenoga: Swapped the hardcoded pb-8 for an empty <CollapsibleCardFooter />, same as DestinationCard, so the bottom spacing now comes from the shared component and stays in sync if it changes.

apps/web/src/pages/data-marts/reports/ConnectAiAssistantPromoActions.tsx
artemfenoga: Removed the unreachable = {} — all props are optional so <ConnectAiAssistantPromoActions /> still type-checks fine without it.

PR #1613, opened 2026-09-16
test(backend): ask for the row count aggregated reports no longer append

zapolsky:
## The failure

`Integration Tests (Real DB)` → **Row-level Calculated Field** has failed every
scheduled run since **2026-08-26**, identically: `55 failed, 125 passed, 180 total`.
Not flaky — the same eleven tests on each of the five warehouses, with the same
counts across runs days apart.

Every failure is `Number(row['Row Count'])` → `NaN`: the column is simply absent
from the response, while the assertion on the line above it (`amount | SUM`)
passes.

## The cause

#1536 removed the `Row Count` column from aggregated reports entirely — *"Team
decision: Unique Count covers the counting need"*. This suite landed eight days
later in #1566, written on a branch that predates that removal, and reads
`row['Row Count']` in eleven tests while requesting it in none.

Nothing caught it at merge time because `test-integration.yml` runs on
`schedule` and `workflow_dispatch` only, never on `pull_request`.

## The fix

The row count is a fixture check the tests still need — it is what pins the
grain and gives `COUNT_DISTINCT` something to be told apart from. So each
affected report asks for it explicitly:

```diff
-        columnConfig: [SESSION_KEY, 'amount'],
-        aggregationConfig: [{ column: 'amount', function: 'SUM' }],
+        columnConfig: [SESSION_KEY, CUSTOMER_ID, 'amount'],
+        aggregationConfig: [ROW_COUNT_RULE, { column: 'amount', function: 'SUM' }],
```

`customer_id` is the column that carries it on all five fixture tables: it is
the join key, so every table has one; it is non-null on every seeded row, so
`COUNT` over it is the group's row count exactly; and it is a STRING, which is
what makes `COUNT` admissible at all — `field-aggregation-governance.ts` allows
a number only `SUM/AVG/MIN/MAX`, so `amount | COUNT` would be refused at save
time and never reach a warehouse.

**Neither grain moves**, and that is a property of the code rather than of the
numbers:

- `collectReportDimensions` keeps only columns with *no* aggregation rule, so
  `customer_id` lands in neither the outer `GROUP BY` nor any sleeve grain;
- `collectSleeveMetrics` keeps only blended columns, so a main-native aggregate
  is never sleeve-routed — the test asserting its blended report carries **no
  sleeve at all** stays valid, as does each one asserting a sleeve is present.

The output name is `customer_id | COUNT`, never a bare `customer_id`, so the two
tests that assert an exact key set stay exact.

Also removes a second leftover of the same removal: `rowCount: false` in
`readTotals`. It no longer exists on `PrepareReportDataOptions` and was this
file's only type error — invisible to CI, which excludes `test/` from
`tsconfig.build.json`, and to ts-jest, which runs with `diagnostics: false`.

## Second commit: dispatching one suite

Proving the fix meant re-running one warehouse suite, and there was no way to
ask for less than all nine — the matrix was hardcoded and `workflow_dispatch`
took no inputs, and a single leg of a running matrix cannot be cancelled. Each
of the other eight seeds and drops real tables on a paid warehouse for nothing.

A dispatch may now name one suite by its jest pattern or its full display name:

```
gh workflow run test-integration.yml --ref <branch> -f suite=row-level-calculated-field-real
```

The suite list moves into the selector job as JSON so it stays a single source
of truth, and it is byte-identical to the matrix it replaces (checked by parsing
both versions and diffing the normalised structures). `inputs.suite` is empty on
a `schedule`, so the nightly run still expands to all nine.

A typo would otherwise yield an empty matrix and an opaque workflow-level error,
so the selector rejects it by name and prints the patterns it knows.
`notify-on-failure` now also needs the selector, so a scheduled run that fails
there is still reported.

## Verification

- `tsc -p tsconfig.json --noEmit` — clean for this file (the repo's other
  pre-existing spec errors are untouched)
- ESLint and Prettier clean on both commits
- The selector script was exercised locally for all four cases: empty (9
  suites), by pattern (1), by full name (1), and a typo (exit 1 naming the known
  patterns)
- The suite itself needs live Athena / BigQuery / Redshift / Databricks /
  Snowflake. Dispatched on this branch — see the run linked in the comments.

No changeset: the release policy excludes test-only and CI-only changes.

## Follow-ups, deliberately not in this PR

- `rowCount: false` also survives in `sql-clause-renderer.spec.ts` and
  `report-data-headers.utils.spec.ts` — harmless, CI green on both.
- **The failure notification is louder than it looks.** `notify-on-failure`
  worked throughout: it comments on the open `integration-failure` issue instead
  of opening a new one, so #1424 has been collecting comments since 2026-07-16
  and is now at 70. After the first one the signal is effectively silent. Worth
  deciding separately whether a three-week-old red deserves something noisier.
- Whether this suite should gate pull requests at all. It does not today, which
  is why this particular break survived a month.


zapolsky:
Verified on live warehouses.

[Run 35076909046](https://github.com/OWOX/owox-data-marts/actions/runs/35076909046), dispatched on this branch with `-f suite=row-level-calculated-field-real`:

```
Test Suites: 1 passed, 1 total
Tests:       180 passed, 180 total
Time:        1082.395 s
```

`180 total` is the same total the failing runs reported (`55 failed, 125 passed`), so no storage was silently skipped for missing credentials — all five `describeIfCredentials` blocks ran.

Two notes on the run itself:

- The dispatch expanded to **one** job rather than nine, so the selector job and the `fromJSON` matrix work on real GitHub, not only against a local simulation of the script.
- The first attempt failed on `npm ci` with `ECONNRESET` before any test executed — a transient runner network fault, unrelated to the change. `gh run rerun --failed` carries the dispatch inputs through, so the re-run still built the single-suite matrix.

.github/workflows/test-integration.yml
```diff
@@ -13,35 +19,70 @@ concurrency:
   cancel-in-progress: true
```
laskevych, line 19: 🟡 **Important** (non-blocking): the concurrency group keys on workflow + ref only, so a narrowed dispatch on `main` now cancels an in-flight nightly — and the loss is silent.

The nightly runs on `refs/heads/main`. A `-f suite=… --ref main` dispatch, which is exactly the workflow this PR is adding, lands in the same group and `cancel-in-progress: true` kills the running nightly mid-flight, replacing nine suites with one. The other eight warehouses get no coverage that night, and because a cancellation does not satisfy `failure()`, `notify-on-failure` never fires and nothing records it.

Before this change the collision was harmless: a dispatch was always the full nine, so the cancelled nightly was replaced by an equivalent run. Narrowing the dispatch is what makes it lossy.

Keying the group on the input as well keeps schedule-vs-schedule and same-suite-vs-same-suite cancellation while letting a targeted re-run coexist with a full one — `github.event.inputs.suite` is empty on a schedule, so the nightly's group is unchanged:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event.inputs.suite }}
  cancel-in-progress: true
```

.github/workflows/test-integration.yml
```diff
@@ -2,6 +2,12 @@ name: Integration Tests (Real DB)
 
 on:
   workflow_dispatch:
+    inputs:
+      suite:
+        description: "One suite's jest pattern, e.g. row-level-calculated-field-real. Empty runs them all."
```
laskevych, line 7: 💡 **Suggestion** (non-blocking): the new input is discoverable only by opening this file.

`docs/contributing/testing.md` (the Integration CI Workflow section) still describes the triggers as `workflow_dispatch` (manual) + nightly cron, so a contributor looking for how to run one suite will not find it. Worth a line there — and while in that file, its stated cron is stale: it reads 3 AM UTC where the workflow runs `0 5,13 * * *`.

Separately, the selector matches `.pattern == $s or .name == $s`, but this description names only the jest pattern and the error message below prints only patterns. Someone who copies a job's display name out of the Actions UI has no way to tell from either text that it will work.

.github/workflows/test-integration.yml
```diff
@@ -13,35 +19,70 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    outputs:
+      suites: ${{ steps.select.outputs.suites }}
+    steps:
+      - name: Select suites
+        id: select
+        env:
+          SUITE: ${{ inputs.suite }}
+        run: |
+          # google-sheets.integration.ts is intentionally absent (not CI-ready);
```
laskevych, line 37: 🧹 **Nit** (optional): this comment used to sit directly above the entry it qualifies; it is now eight entries away from `Google Sheets (column preservation)`.

Someone adding a `google-sheets.integration` entry at the bottom of the list reads the surrounding lines, sees no constraint, and re-adds the suite that was deliberately left out. Moving it back inside the JSON, immediately above the column-preservation entry, restores the locality at no cost.

laskevych (approved):
Approving — nothing blocking.

**What I verified beyond the description**

The two claims the fix rests on hold in the code, not only in the numbers: `collectReportDimensions` keeps only columns with *no* aggregation rule, and `collectSleeveMetrics` filters on `outputAliasToRoot.has(r.column)` — a blended column. `customer_id` is main-native and aggregated, so it lands in neither the outer `GROUP BY` nor any sleeve grain, and the sleeve-free shape the one test asserts stays sleeve-free.

I traced every one of the ten `ROW_COUNT` reads, shared helpers included (`assertAggGroups`, `assertFilterGroups`, `joinedGroupedReport`, `blendedFilterReport`) — each corresponding report now carries `ROW_COUNT_RULE`, with no reader left behind. The two exact-key-set assertions stay exact because the output name is `customer_id | COUNT`. `customer_id` is non-null on all four main fixture tables, and the join-back happens at report grain through the sleeve, so the count cannot be inflated by fan-out regardless of the joined table's cardinality.

`rowCount` is indeed gone from `PrepareReportDataOptions`, so the `readTotals` line was a real type error rather than a tidy-up; `tsconfig.build.json` excludes both `test` and `**/*spec.ts`, which explains why neither it nor the two remaining occurrences ever reached CI.

On the workflow: the dispatch input reaches the script through `env:` rather than shell interpolation and is matched against a fixed list, so no arbitrary string can reach the `npm run test:integration -- <pattern>` line. Adding `select-suites` to `notify-on-failure`'s `needs` is load-bearing — a selector failure on a scheduled run is now reported rather than swallowed.

Three non-blocking notes inline.

PR #1614, opened 2026-09-16
feat(data-marts): avoid duplicate rows in reports with no chosen aggregation (#6948)

zapolsky:
Fibery: [#6948 Avoid duplicates in Reports as a default behaviour](https://owox.fibery.io/Sprint_Work/Work_Item/Avoid-duplicates-in-Reports-as-a-default-behaviour-6948)

## What this changes

A report with an explicit column selection but no aggregation, date bucket, or Unique Count set anywhere could deliver duplicate rows — most often from a join fanning one source row out across many. A stored report that OWOX itself delivers now collapses automatically:

- a selection made only of dimensions is returned `DISTINCT`;
- a selection carrying a metric gets the first aggregation its Data Mart governance allows, in a fixed priority order (`SUM` → `AVG` → `MIN` → `MAX` for numbers, `MIN` → `MAX` for dates and times);
- a row-level calculated formula is rewritten to its group-level equivalent when the rewrite is provably value-preserving, and left as a grouping key when it is not.

## Where it applies

The opt-in is confined to `RunReportService` and `GetReportGeneratedSqlService`. Every ad-hoc path — HTTP Data, the MCP tools, `apps/ctl`, the Looker Studio cache fill, copy-as-Data-Mart, and the save-time dry run — keeps returning the raw projection. Microsoft Excel and Looker Studio pull their own data and have no server-side run, so neither the SQL preview nor the picker's ghost predicts a collapse for them.

That guarantee is negative, so it is asserted as one: `auto-collapse-opt-in.spec.ts` fails if `applyAutoCollapse` is ever called from a third service.

## Where it deliberately does nothing

The resolver aborts as a whole rather than half-collapsing a report. It refuses when a sort names a column outside the projection, when a selected column cannot be a `GROUP BY` key, when a filter already targets an aggregate-level calculated field, when a metric has no aggregation its governance allows (a text or boolean metric, notably), when the report selects a column from a joined Data Mart, and when a formula cannot be lifted. Calculated levels are re-derived from the formula rather than read off the stored field, so the resolver and the query builders cannot disagree about what already aggregates.

## Impact, measured on production

| | reports |
| --- | --- |
| Total reports | 896 |
| In scope (explicit selection, no controls, push destination) | 104 |
| Of those, relabel a column (`sessions` → `sessions \| SUM`) | 64 |
| Of those, run on an active schedule | 7 |
| Refused for a sort outside the projection | 0 |

The rename is the only externally visible change, it happens even where the report had no duplicates to fix, and it breaks a consumer only if something downstream reads the column **by name** (a `QUERY()` with a header label, a pivot, a named range). The 7 scheduled ones are the only set that can change without a person present. No field in production carries an explicit `aggregationRole`, so all governance today is type-derived.

## Review response

A multi-agent review raised 21 findings. Both blockers are fixed, with regression tests that reproduce them:

1. **A lifted formula moved its own filter.** The guard that refuses a report filtered on an aggregate-level calculated field runs before the lift, and the lift is what makes the field aggregate-level — so `WHERE clicks/impressions > 0.5` silently became `HAVING SUM(clicks)/SUM(impressions) > 0.5` and kept a different set of rows. A formula the report also filters or sorts on is no longer lifted.
2. **An unresolvable column became a grouping key.** A joined column has no descriptor in the report's own schema, so a dimension cannot be told from a metric — and grouping by a joined metric drops its duplicate rows and moves its total exactly as `DISTINCT` would. Such a report is now left alone.

Two of the review's proposals were not adopted, and the reasoning is in the commits:

- Making a metric with no usable auto-aggregation a grouping key would reproduce the same defect: rows (A,10), (A,10), (A,5) grouped by `(country, sessions)` total 15, not 25. The abort was correct; the **documentation** was what promised `MIN` → `MAX` for text that governance never allows, and that is what was fixed.
- The web mirror refusing a liftable calculated metric is the documented safe direction — a missing prediction over one that never materialises. Closing it needs the server's verdict, which is out of scope here.

Everything else in the review is addressed: NULL semantics in the lift (a formula without division is now summed whole, so the collapsed total matches the values the uncollapsed report displayed), `DISTINCT` and the refusal reason are now recorded on the run, the entity prototype survives `applyAutoCollapse`, per-dialect `DISTINCT` is covered for all six builders, and the duplicated type-category sets have a parity spec.

## Verification

- `apps/backend`: 8197 unit tests in 462 suites, 59 output-controls e2e, `nest build` clean
- `apps/web`: 1809 tests, `tsc -b` clean
- Mirrored branch in `google-sheets-extension`: 657 tests, `tsc --noEmit` clean
- Changeset: `.changeset/6948-avoid-duplicates-in-reports.md`

## Companion PR

The report picker's preview lives in the separate spreadsheets repository and ships alongside this: OWOX/owox-data-marts-spreadsheets#61.


apps/backend/src/data-marts/services/auto-collapse.resolver.ts
```diff
@@ -0,0 +1,282 @@
...
+    const governance = resolveFieldGovernance(descriptor.type, {
+      aggregationRole: descriptor.field.aggregationRole as AggregationRole | undefined,
+      allowedAggregations: descriptor.field.allowedAggregations as
+        | ReportAggregateFunction[]
+        | undefined,
+    });
+    // A calculated dimension collapses by being grouped; lifting it would discard the grouping.
+    if (governance.role !== 'metric') continue;
+
+    if (isCalculatedField(descriptor.field)) {
```
laskevych, line 122: 🔴 Blocking: The lift guards only the lifted field itself, not the fields that reference it. `withLiftedFormulas` rewrites the lifted formula to `SUM(...)` on the cloned schema, and the composer re-derives every level from that clone, so any other row-level calculated field whose formula reads the lifted column silently becomes aggregate-level.

Two concrete cases (calc-to-calc references are allowed by the validator, only cycles are rejected):

- `margin = {{revenue}} - {{cost}}` (liftable metric) and `margin_label = CONCAT('m:', CAST({{margin}} AS STRING))` (STRING, dimension). Report projects `[date, margin, margin_label]`. `margin_label` `continue`s here as a grouping key, but on the clone `calculatedFieldLevelOf(margin_label)` walks the dependency and returns aggregate, so `isCalculatedGroupingKey` drops it from `GROUP BY` and it renders as `CONCAT('m:', CAST(SUM(revenue - cost) AS STRING))`. The report collapses to one row per date.
- Report projects `[date, ctr]` (`ctr` lifted) and filters on an unprojected `pct = {{ctr}} * 100 > 5`. `referencedByFilterOrSort('ctr')` is false, `ctr` is lifted, `routeFilterClauses` re-derives `pct` from the clone as aggregate-level, and the `WHERE` becomes a `HAVING` on group totals. This is exactly the case the comment above says must be refused, one hop removed.

`aggregatesByItself` / `filtersAnAggregateCalculatedField` run against the original schema, so they cannot see this. Suggested fix: before lifting a column, refuse the collapse (`calculated-not-liftable`) when any other calculated field in the schema has a live reference to it (transitively), or re-run the two checks against the cloned schema and abort if any answer changed. Neither resolver spec covers a dependent of the lifted field (the "DIFFERENT column" test uses a native column).

apps/backend/src/data-marts/calculated-fields/formula-lifting.ts
```diff
@@ -0,0 +1,482 @@
...
+    tokens,
+    refs,
+    index => resolved[index]?.truncatesUnderDivision === true
+  );
+  if (refusal) return { formula: null, reason: refusal };
+  if (resolved.includes(undefined)) {
+    return { formula: null, reason: 'unaggregatable-reference' };
+  }
+
+  if (!dividesAnywhere(tokens)) {
```
laskevych: 🔴 Blocking: `dividesAnywhere` switches to per-reference wrapping for any `/`, including division by a numeric literal, so a linear formula loses the whole-wrap NULL semantics the docblock above (and the setup guide) promise.

`({{revenue}} - {{cost}}) / 2`, rows `(100, NULL)` and `(200, 50)`: the uncollapsed report shows `NULL` and `75`, so the analyst's total is 75 and `SUM((revenue - cost) / 2)` is 75. The lift emits `(SUM(revenue) - SUM(cost)) / 2` = `(300 - 50) / 2` = **125**. The per-reference rationale ("ratio of totals") only holds when the divisor holds a reference; a literal divisor is the same linear scaling as `* 0.5`, and the spec already treats `{{amount}} * 1.2` and `{{amount}} / 100` differently for no arithmetic reason. The ACCEPTED sweep has no NULL rows and every literal-divisor entry has a single reference in the dividend, so the suite cannot see it. Affects BigQuery, Snowflake and Databricks (Athena/Redshift already refuse integer dividends).

Suggested fix: decide whole vs per-reference from the parsed tree, per-reference only when the ratio-root divisor (the `/` right operand or `SAFE_DIVIDE`'s `denominatorArgument`) holds a reference. That also removes `DIVISION_FUNCTIONS` as a second source of truth for "is this a ratio" (a future `LIFTABLE_FUNCTIONS` entry with `denominatorArgument` but missing from that set would be whole-wrapped as `SUM(fn(a, b))`, and the sweep's `enumerate()` hardcodes `SAFE_DIVIDE`/`NULLIF`, so it would not catch it). Please add a NULL-row case with two references and a literal divisor.

apps/backend/src/data-marts/use-cases/run-report.service.ts
```diff
@@ -284,13 +290,13 @@ export class RunReportService {
         sqlOverride = blendingDecision.blendedSql;
         sqlOverrideParams = blendingDecision.params;
         calculatedFields = blendingDecision.calculatedFields;
-      } else if (hasOutputControls(report)) {
+      } else if (hasOutputControls(effectiveReport)) {
```
laskevych, line 302: 🟡 Important (non-blocking, but worth a rollout note): a columns-only report on a SQL-defined Data Mart previously read the definition as an inline subquery and performed no DDL. Auto-collapse makes `hasOutputControls(effectiveReport)` true, so such a report now goes through `compose()` → `resolveTableName` → `CREATE OR REPLACE VIEW` with the Data Mart's storage credential, on every run and on every Generated SQL preview (no `TableReferenceMemo` is passed, so the view is recreated each time). The same applies to `GetReportGeneratedSqlService`, where a viewer clicking Preview SQL now performs DDL; `get-report-output-schema.service.ts` explicitly avoids `compose()` for that reason.

This widens an existing side effect (filter/sort/limit reports already do this) to the 104 in-scope reports. The concrete regression is a read-only warehouse credential that worked yesterday and now fails both the scheduled run and the preview with a permission error, on all six storages. Suggest calling it out in the PR/changeset and considering a follow-up for a no-DDL subquery path for SQL-defined marts.

docs/getting-started/setup-guide/report-aggregations.md
```diff
@@ -87,6 +87,54 @@ You can apply several functions to one column — for example `SUM` and `AVG` of
...
+A report with an explicit column selection but **no** aggregation, date bucket, or Unique Count set anywhere can still return duplicate rows — most commonly when a join fans a source row out across several joined rows, or a [calculated field](./calculated-fields.md) formula is row-level. OWOX now closes that gap on its own, without you adding anything:
+
+- If every selected column is a **dimension**, OWOX returns the projection **distinct** — the same rows, de-duplicated, with no `GROUP BY` and no relabeled columns.
+- If the selection includes at least one **metric** column, OWOX groups by the remaining dimensions and applies the **first aggregation the column's allowed-aggregations set permits**, in this fixed priority order:
+
+  | Field type             | Priority order                |
+  | ---------------------- | ----------------------------- |
+  | Numeric                | `SUM` → `AVG` → `MIN` → `MAX` |
+  | Date / Time            | `MIN` → `MAX`                 |
+  | String, Boolean, Other | never auto-aggregated         |
```
laskevych: 🟡 Important: the table and prose say String metrics are "never auto-aggregated" (and the changeset: "A text or boolean metric is left alone"), but `PRIORITY_BY_CATEGORY.string` is `['MIN', 'MAX']` on both backend and web, and `SUPPORTED_BY_CATEGORY.string` includes `MIN`/`MAX`, so a STRING metric whose allowed set includes `MIN` is delivered as `country | MIN`. Either drop `string` from the priority table (both copies) so the docs hold, or fix the guide and the changeset. As written the release note tells users text metrics are untouched.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
```diff
@@ -0,0 +1,282 @@
...
+      continue;
+    }
+
+    const fn = pickAutoAggregation(descriptor.type, governance.allowedAggregations);
+    if (!fn) return { kind: 'none', reason: 'no-allowed-aggregation' };
+    aggregations.push({ column: name, function: fn });
+  }
+
+  // Nothing to aggregate, so collapsing is exactly DISTINCT and cannot change a value.
+  if (aggregations.length === 0 && liftedFormulas.length === 0) return { kind: 'distinct' };
```
laskevych: 💡 Suggestion: a `distinct` plan plus a sort on a projected row-level calculated **dimension** of FLOAT/NUMERIC/DECIMAL type renders `ORDER BY CAST(<expr> AS <type>)` (never the alias, see `buildCalculatedSortExpressions`), which is not in the `SELECT DISTINCT` list and is rejected by BigQuery, Athena/Trino and Redshift. Narrow, since only the API can put a dimension role on a numeric calculated field, but the report ran fine uncollapsed. Consider refusing the collapse when a sort names a calculated field whose declared type imposes a comparison cast, and adding a sort case to `distinct-projection.dialects.spec.ts`.

apps/backend/src/data-marts/use-cases/run-report.service.ts
```diff
@@ -310,6 +316,36 @@ export class RunReportService {
...
+      // rows were delivered — so it is recorded too, rather than leaving Run History to imply the
+      // raw projection was returned.
+      if (autoCollapsePlan.kind === 'distinct' && dataMartRun?.reportDefinition) {
+        (dataMartRun.reportDefinition.outputConfig ??= {}).autoAppliedDistinct = true;
+      }
+      // The six refusal reasons exist to answer "why does this report still return duplicates?",
+      // and a log line is the only place that question gets asked after the fact.
+      if (autoCollapsePlan.kind === 'none') {
+        reportRunLogger?.log({
+          message: 'Automatic duplicate collapse skipped',
```
laskevych: 💡 Suggestion: `columnConfig: null` is the create default, so roughly 790 of the 896 production reports will write `Automatic duplicate collapse skipped / no-explicit-projection` into persisted run logs on every run, and `StructuredLogsView` renders the two-key object as a JSON blob. The reasons that actually answer "why do I still get duplicates" drown in the trivially true one. Suggest logging only when `columnConfig` has entries (skip `no-explicit-projection`, arguably `analyst-aggregated` too) and emitting a single message string.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
```diff
@@ -0,0 +1,282 @@
...
+      : field;
+  });
+
+  // Keeps the prototype: `DataMart` exposes its owner ids as accessors, which a spread would drop.
+  const dataMart = Object.assign(
+    Object.create(Object.getPrototypeOf(report.dataMart) as object) as DataMart,
+    report.dataMart,
+    { schema: { ...schema, fields } as DataMartSchema }
+  );
+  return { ...report, dataMart };
```
laskevych: 💡 Suggestion: `withLiftedFormulas` returns `{ ...report, dataMart }`, which drops the `Report` prototype that `patched()` was written to preserve, so on the lift path `applyAutoCollapse` hands back a prototype-less object typed as `T`. Latent today (no consumer of `effectiveReport` calls `isEmailBasedDestination()` / `ownerIds`), and the "keeps the report prototype on every branch" spec passes only because it builds a `ReportLikeReadPlan`. One-line fix: `return patched(report, { dataMart })`, and a spec that uses a real `Report` instance.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -1727,6 +1724,20 @@ export function ReportColumnPicker({
     effectiveOutputConfig.dateTruncConfig,
   ]);
 
+  // Read against the raw `value`/`outputConfig` props, not the `effective*` ones: a null `value`
+  // means no explicit projection, which the resolver must read as "nothing to predict yet".
+  const autoAggregations = useMemo(
```
laskevych, line 1757: 💡 Suggestion: this memo depends on the `outputConfig` prop, which both forms build as an inline object literal every render, and `autoAggregationByColumn` returns a fresh `Map` even for `none`/`distinct` plans. The Map is passed to every memo'd `NativeFieldRow`/`BlendedFieldRow`, so every parent render (e.g. each keystroke in the email message textarea, which is watched at the form root) re-renders every field row, including unselected ones that previously kept stable props. Suggest passing a per-row scalar `autoAggregation={autoAggregations.get(field.name)}` like the existing `aggregation` prop, keying the memo on the config arrays, and returning the shared empty map for non-aggregate plans.

Related: the `autoAggregationByColumn` plumbing through `BlendedGroupItem`/`BlendedFieldRow` can never hit, since the resolver returns `unresolvable-column` for any joined column in `columnConfig`; either drop it or mark it as reserved for the joined-fields follow-up.

apps/backend/src/data-marts/dto/schemas/field-type-category.web-parity.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+);
+
+/** `const NAME = new Set([...])`, read as the literal list it is. */
+function webSet(source: string, name: string): string[] {
+  const match = new RegExp(`const ${name}\\s*=\\s*new Set\\(\\[([^\\]]*)\\]`).exec(source);
+  if (!match) throw new Error(`${name} not found in the web operators file`);
+  return [...match[1].matchAll(/'([^']*)'/g)].map(entry => entry[1]).sort();
+}
+
+describe('field type categories match the web picker', () => {
```
laskevych, line 68: 🧪 Test coverage: the new parity spec covers the five type sets but not `PRIORITY_BY_CATEGORY`, which is hand-mirrored between `field-aggregation-governance.ts` and the web `aggregation-governance.ts`. Drift there shows a ghost function the server does not apply, the exact failure this spec's header describes. The same read-as-text approach works for the priority object.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
```diff
@@ -0,0 +1,282 @@
...
+    (report.filterConfig ?? []).some(rule => rule.column === column) ||
+    (report.sortConfig ?? []).some(rule => rule.column === column)
+  );
+}
+
+/**
+ * Filtering on an aggregate-level calculated field already forces a GROUP BY with a HAVING, so the
+ * report is collapsed before we look at it.
+ */
+function filtersAnAggregateCalculatedField(
```
laskevych: 💡 Suggestion: `filtersAnAggregateCalculatedField` re-derives the "this filter lands in HAVING" verdict that `routeFilterClauses` in `filter-clause-routing.ts` already owns (it also treats a rule with `function` as HAVING, which here relies on the validator invariant). `routeFilterClauses(report.filterConfig, schemaFields).some(r => r.clause === 'having')` keeps one owner for that decision.

laskevych (changes requested):
Thanks for the thorough PR and the detailed write-up. The resolver's abort-as-a-whole design and the negative opt-in spec are the right shape. Two confirmed cases move a number, though, and since "provably value-preserving" is the premise of the lift, I'm requesting changes for those. Everything else below is non-blocking.

**Blocking** (inline):
- 🔴 A lifted formula changes the level of every calculated field that references it (`auto-collapse.resolver.ts`).
- 🔴 Division by a literal forces per-reference wrapping and moves NULL totals (`formula-lifting.ts`).

**Important, non-blocking** (two are on files the PR does not touch, so they live here):

🟡 Important: `GetReportOutputSchemaService` (`GET /reports/:id/output-schema` and the MCP `get_report_output_schema` tool) still resolves headers from the stored `aggregationConfig`, so for an auto-collapsed report it describes `sessions` with `aggregateFunction: null` while the run delivers `sessions | SUM`. Its doc comment ("the same `resolveReportDataHeaders` path a run uses") and the tool description ("the names to put above the values the report delivers") are now inaccurate for every collapsed report, and `auto-collapse-opt-in.spec.ts` locks that in. Since the Generated SQL preview did opt in, I'd expect the describe endpoint to opt in with the same pull-destination skip and be added to the allowlist. If that is deliberately out of scope, please correct the comment and the tool description.

🟡 Important (rollout note, not a code defect): the Google Sheets column planner (`column-plan-builder.ts`) diffs by header name only, so the auto-applied `sessions` → `sessions | SUM` rename becomes a `deleteDimension` of the old column plus an append at the right edge of the imported range. Column positions shift and cross-sheet references by letter (`=Sheet1!C2`) turn into `#REF!`. The PR body only mentions consumers that read the header **by name**. Same mechanics as a manual aggregation today, but now triggered on the 64 production reports (7 scheduled) with nobody present. Worth adding to the blast-radius statement, and possibly a follow-up for rename-aware planning when the base name matches an auto-applied aggregation.

📝 Docs:
- `docs/contributing/testing.md` line 8 was corrected, but the workflow-reference table at lines 832-833 still says `Manual dispatch` for both E2E workflows, and `e2e-browser.yml` runs on `pull_request` too, so line 9 is also stale.
- The "When automatic aggregation does not apply" list does not mention the `calculated-not-liftable` refusals for a formula that is itself filtered or sorted on, nested in a RECORD, or referencing another calculated / joined / hidden field, while the changeset promises the guide "lists every such case".
- The comment at `run-report.service.ts:246-248` says ad-hoc callers "can ask for their own aggregation", which is not true for `GET reports/:id.ndjson` (`HttpReportDataQuerySchema` is `.strict()` with only `limit`). The exclusion is fine, the stated reason is not.

⚙️ CI: the ESLint workflow is red on both runs of this branch. Web: `ReportColumnPicker.test.tsx:4960` `as HTMLElement` violates `@typescript-eslint/non-nullable-type-assertion-style` (use `!`). Backend: the latest run died with a JS heap OOM; the previous run on this branch and `main` are green, so that one looks like a flake and a rerun should do.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
zapolsky: Confirmed, and fixed in e0540570f. I reproduced both of your cases before touching anything — `margin_label` did drop out of the `GROUP BY`, and the unprojected `pct` filter did move to `HAVING`.

The fix refuses the collapse (`calculated-not-liftable`) when any *other* calculated field in the schema holds a live reference to the column about to be lifted, checked against the original schema before the clone is built.

Direct readers only, and that is not an approximation: the lift rewrites this column's formula on the clone, every level downstream is re-derived from that clone, so a direct reader is already rewritten — one is all it takes to refuse. A chain `c -> b -> a` is caught at `b` before `c` is ever reached.

Both of your scenarios are now specs in `auto-collapse.resolver.spec.ts` (the `margin_label` grouping-key case and the unprojected `pct` filter), alongside the existing native-column test that did not cover this.

apps/backend/src/data-marts/calculated-fields/formula-lifting.ts
zapolsky: Confirmed, and fixed in e0540570f. `(revenue - cost) / 2` over `(100, NULL)` and `(200, 50)` gave 125 against the analyst's 75, exactly as described.

`classifyLiftableShape` now decides whole vs per-reference from the parsed tree, as you suggested: per-reference only when the ratio-root divisor holds a reference — the `/` right operand, or the `denominatorArgument` of the `LIFTABLE_FUNCTIONS` entry. `dividesAnywhere` and `DIVISION_FUNCTIONS` are gone, so `LIFTABLE_FUNCTIONS` is the only place that says what a ratio is; a future entry with a `denominatorArgument` is now correct by construction rather than by remembering a second list.

The NULL-row case is in `formula-lifting.spec.ts` with two references and a literal divisor.

One detail worth recording, because an existing test caught it and I would not have thought of it: the whole-wrap emits `SUM(<stored>\n)`. The newline is load-bearing — a stored formula ending in a `--` comment would otherwise swallow the closing paren.

apps/backend/src/data-marts/use-cases/run-report.service.ts
zapolsky: Agreed, and the changeset now says so, in its own bolded paragraph rather than a footnote:

> **If your Data Mart is defined by SQL and its storage credential is read-only, check it before upgrading.** A report that only picked columns previously ran its definition as an inline subquery and wrote nothing to the warehouse. Once it collapses it counts as having output controls, which takes the same path a filter or a sort already takes: OWOX materialises the definition as a view (`CREATE OR REPLACE VIEW`) using that Data Mart's credential, on each run and on each Generated SQL preview. A credential that may read but not create will now fail where it worked before.

The preview half is the part I would not have called out on my own — a viewer clicking Preview SQL performing DDL is a worse surprise than a scheduled run failing, because nothing about the action suggests a write.

The no-DDL subquery path belongs in its own ticket rather than this PR: it changes behaviour for every filter/sort report that already takes this path today, not just the ones auto-collapse newly adds.

docs/getting-started/setup-guide/report-aggregations.md
zapolsky: You are right, and this one caught a mistake of mine in the opposite direction — an earlier pass had already "fixed" this text and made it wrong the other way.

Fixed in e0540570f, in the guide and the changeset both. The wording now distinguishes the two cases instead of collapsing them: a text metric is left alone **under its default governance** — `DEFAULTS_BY_CATEGORY.string` turns on `COUNT`/`COUNT_DISTINCT`/`STRING_AGG`/`ANY_VALUE`, none of which returns the column you selected — and explicitly allowing `MIN` or `MAX` opts it back in, delivering `country | MIN`.

I kept `string: ['MIN', 'MAX']` in the priority table on both sides rather than dropping it: dropping it would make that explicit-allow path unreachable, which is a product change, not a docs fix.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
zapolsky: Taken, in a81a1c139. I had parked this one as too narrow to be worth the branch; when I went back to check, it reproduces exactly as written.

The resolver now refuses when a sort names a calculated field whose declared type imposes a comparison cast (`sort-outside-projection`), and only on the pure-`DISTINCT` branch — on the aggregate branch the same field becomes a grouping key, so the expression *is* in the projection and the sort resolves.

The category answers "does this take a cast" without the resolver learning a storage: the cast tables are per dialect, but every one of them holds numeric declarations only.

I put the coverage in `auto-collapse.resolver.spec.ts` rather than `distinct-projection.dialects.spec.ts` — two specs, a FLOAT calculated dimension refused and a STRING one still collapsing. The dialects spec drives the builders for the projection shape; this verdict is the resolver's, and asserting it there keeps the reason code visible in the failure.

apps/backend/src/data-marts/use-cases/run-report.service.ts
zapolsky: Fixed in e0540570f. The log is now gated on a set of the five reasons that answer "why do I still get duplicates" — `non-groupable-column`, `unresolvable-column`, `no-allowed-aggregation`, `calculated-not-liftable`, `sort-outside-projection` — and emits one message string rather than a two-key object.

`no-explicit-projection` and `analyst-aggregated` are both silent. Your ~790 figure is the reason: a log line that is true of most reports on most runs is not information, and it would have buried the five that are.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
zapolsky: Fixed in e0540570f — `return patched(report, { dataMart })`, as suggested.

The spec you asked for is there too, and it needed to be a separate one: the existing "keeps the report prototype on every branch" test returns on the `distinct` branch before `withLiftedFormulas` is ever reached, which is why it passed over this. The new one drives a report through the lift path with a real class exposing `ownerIds` as an accessor and `isEmailBasedDestination` as a method, and asserts both survive.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
zapolsky: Taken, in a81a1c139 — the second of the two I had parked, and the second that turned out to bite.

Rows now take a per-row scalar `autoAggregation={autoAggregations.get(field.name)}`, mirroring the existing `aggregation` prop, and `autoAggregationByColumn` returns a shared empty map for `none`/`distinct` plans. The memo also reads `schema.nativeFields` raw rather than the flattened list, which is what its extension twin walks.

I kept the `outputConfig` dependency rather than keying on the config arrays: with the scalar prop, a recompute no longer reaches the rows, and the raw `outputConfig` is what the resolver has to read anyway.

On the related note: I kept the blended plumbing and documented why instead (3eb58f5cf) — it is empty for a joined group today because the resolver refuses any report projecting a joined column at all, and it is wired so the follow-up that lifts that refusal has nothing left to connect.

apps/backend/src/data-marts/dto/schemas/field-type-category.web-parity.spec.ts
zapolsky: Done in e0540570f — the spec now reads `PRIORITY_BY_CATEGORY` from both sources as text and compares it order-sensitively, which is the part that matters: a set comparison would pass while the server applied `AVG` where the picker promised `SUM`.

Worth noting for whoever edits this next: my first regex silently matched `SUPPORTED_BY_CATEGORY` instead, and passed. The spec now scopes to the `PRIORITY_BY_CATEGORY` block before extracting.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
zapolsky: Done in 77559b95d.

Worth saying that this is not only a de-duplication: the delegation changes behaviour. A filter rule carrying its own aggregate `function` now puts the report on the aggregated branch, which the field-level check could not see — it asked about field levels, and that rule's level lives on the rule. Today only a validator invariant keeps such a rule out; the router already knew, and now one answer serves both.

Pinned with a spec for exactly that case, so the parity cannot be quietly removed again.

docs/getting-started/setup-guide/report-aggregations.md
```diff
@@ -87,6 +87,55 @@ You can apply several functions to one column — for example `SUM` and `AVG` of
...
+
+Where the rewrite puts its `SUM` depends on whether the formula divides, and empty values are what decides it. A formula that does not divide is summed as a whole — `{{revenue}} - {{cost}}` becomes `SUM({{revenue}} - {{cost}})` — so the collapsed report totals exactly the values the uncollapsed one displayed, including rows where one side was empty and the whole cell was therefore empty. A formula that does divide has each of its references summed instead, because summing a column of row-by-row ratios is not the group's ratio. That form deliberately reads a numerator whose denominator is empty: `SUM(revenue) / SUM(cost)` is the ratio of the two totals, which is what a collapsed ratio means, even though the row it came from showed no value.
```
laskevych: 📝 Docs (non-blocking): this paragraph still says the placement "depends on whether the formula divides", which is the rule the fix replaced. `{{amount}} / 100` divides and is now summed whole. The deciding fact is whether the formula divides **by a reference**; a literal divisor is linear scaling. Same section, the "does not apply" list below is also missing the two refusals added in this round: a lifted column that another calculated field reads, and a sort on a numeric calculated dimension when the report would collapse to `DISTINCT`.

apps/backend/src/data-marts/calculated-fields/formula-lifting.ts
```diff
@@ -0,0 +1,486 @@
...
+ * SUM is always the wrapper, never the field's own aggregation: the rewrite recomputes the formula
+ * from group totals, and only a sum is a group total. WHERE it wraps depends on the shape, and the
+ * difference is NULL:
+ *
+ * - No division — the formula is a linear combination, so the whole text is wrapped once:
+ *   `{{revenue}} - {{cost}}` becomes `SUM({{revenue}} - {{cost}})`. Over rows this is the sum of
+ *   exactly the values the uncollapsed report displayed. Wrapping each reference instead would
+ *   read `SUM(revenue) - SUM(cost)`, which differs the moment one side is NULL: rows (100, NULL)
+ *   and (200, 50) display NULL and 150 and total 150, while the per-reference form answers 250.
+ * - A division — each reference is wrapped instead, because `SUM(a/b)` is the average of row
```
laskevych: 📝 Docs (non-blocking): the docblock still splits on "No division" vs "A division", while the code now splits on `dividesByAReference`. Worth rewording to "no reference in the divisor" vs "a reference in the ratio-root divisor", so the comment and `classifyLiftableShape` say the same thing.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
```diff
@@ -0,0 +1,348 @@
...
+
+/**
+ * Whether another calculated field reads this column.
+ *
+ * DIRECT readers are enough, and that is not an approximation: lifting rewrites this column's
+ * formula on a clone, every level downstream is re-derived from that clone, and so a direct reader
+ * is already rewritten — one is all it takes to refuse. A chain `c -> b -> a` is caught at `b`
+ * before `c` is ever reached.
+ */
+function readByAnotherCalculatedField(
```
laskevych, line 213: 💡 Suggestion (follow-up, non-blocking): this refuses whenever any calculated field in the schema reads the column, even one the report never projects, filters or sorts on. That is the safe direction and the direct-reader argument is sound, but it also switches the lift off for the common `ctr` + `ctr_pct = {{ctr}} * 100` pair in a report that only selects `ctr`. If that turns out to matter in practice, the check can be narrowed to readers the report actually reaches (transitively), since an unreferenced reader never enters the SQL.

laskevych (approved):
Re-reviewed at 3eb58f5cf. Both blockers are fixed the right way and pinned with specs, CI is green, approving.

- **Lift vs dependents:** `readByAnotherCalculatedField` refuses before the clone is built; I checked that `calculatedFieldsOf` includes hidden fields, so a hidden intermediary in a chain is caught too. Both scenarios are now specs.
- **Literal divisor:** the whole vs per-reference decision now comes from the parsed tree (`dividesByAReference`), and since a reference in a divisor is only accepted at the ratio root, checking the root node is complete. `DIVISION_FUNCTIONS` is gone, NULL-row spec added.
- Also taken: DISTINCT + cast sort refusal, `patched()` on the lift path with a real-class spec, gated single-string skip log, per-row scalar prop, order-sensitive `PRIORITY_BY_CATEGORY` parity spec, `routeFilterClauses` as the single owner of the HAVING verdict, string-metric wording, and the read-only-credential warning in the changeset.

Still open from the previous round, all non-blocking and fine as follow-ups:
- 🟡 `GetReportOutputSchemaService` / MCP `get_report_output_schema` still describe the uncollapsed headers (`sessions`, `aggregateFunction: null`) for a report whose run delivers `sessions | SUM`. At minimum the doc comment and the tool description should stop promising run parity.
- 🟡 Rollout note: the Sheets planner turns the auto rename into delete + append at the right edge, so column-letter references break too, not only by-name consumers.
- 📝 `docs/contributing/testing.md` lines 832-833 still say `Manual dispatch`; the comment above `applyAutoCollapse(report)` in `run-report.service.ts` still says ad-hoc callers "can ask for their own aggregation", which does not hold for `GET reports/:id.ndjson`.

docs/getting-started/setup-guide/report-aggregations.md
zapolsky: Fixed in 8d3e20a6e. Both halves were right.

The placement paragraph now reads "divides **by another column**", and names the literal-divisor case explicitly: `({{revenue}} - {{cost}}) / 2` divides but scales exactly like `* 0.5`, so it is summed whole.

The "does not apply" list gained the two refusals it was missing:

- a lifted column that another calculated field reads — stated as the consequence rather than the mechanism ("every formula built on it is then read at the group level too; a dimension made from it stops being a GROUP BY key, and a filter on it moves to the grouped query");
- a sort on a numeric calculated dimension when the report would collapse to `DISTINCT` — kept separate from the existing "sort names a column the report does not display" bullet, since the failure is different: the column IS displayed, but the sort renders as a cast expression that the DISTINCT list does not contain.

Also added one clause to the changeset, which had the same gap — it described what the lift cannot prove, but not that a formula another field reads is left alone regardless of what it could prove.

apps/backend/src/data-marts/calculated-fields/formula-lifting.ts
zapolsky: Fixed in 8d3e20a6e. The two bullets now read "No reference in the divisor" and "A reference in the ratio-root divisor", matching `classifyLiftableShape`, and the first one carries `({{revenue}} - {{cost}}) / 2` as its example so the literal-divisor case is impossible to miss on a skim.

Worth saying plainly: this comment and the one on the setup guide are both my staleness, not an oversight in the original text. I changed the rule and left every sentence that stated the old one — in three places. The docblock is the one that would have misled longest, since it sits directly above the function it describes.

apps/backend/src/data-marts/services/auto-collapse.resolver.ts
zapolsky: Agreed it is narrowable, and I checked the premise rather than taking it on trust: `buildCalculatedFieldPlans` is called with the projection (`decision.columnFilter`) and with filter/having columns, and filters `calculatedFieldsOf(...)` down to those names — so a calculated field the report names nowhere genuinely never becomes a plan. The `ctr` + `ctr_pct` pair does cost a collapse today for no benefit.

One trap for whoever picks this up, because the obvious implementation is wrong: the check cannot be "is the reader named by the report". `calculatedDependencyPlans` recurses (it calls itself on each dependency's formula), so an UNNAMED intermediate still reaches the SQL. Schema `x` (lifted), `r = {{x}} * 2`, `a_pct = {{r}} + 1`, report projects `[x, a_pct]`: `r` is named nowhere, but it is carried into `a_pct`'s plan and substituted, so rewriting `x` changes `a_pct`'s level through it. The narrowing has to run over the transitive closure of what the named fields read, not over the named set.

I am leaving it out of this PR deliberately. The current check errs toward keeping duplicates, which costs a collapse and never a number, and I have no measurement of how often calc-to-calc references actually occur — that is what your "if it turns out to matter" hinges on, and I would rather answer it with a count than a guess. I can run a read-only query against production for the frequency of calculated fields referencing other calculated fields, the same way the 896 → 104 → 64 → 7 numbers in this PR were obtained. Happy to do that and open the follow-up if the number justifies it.

zapolsky:
Rebased onto `main` (force-pushed, same 14 commits). Flagging one conflict because the resolution was a behaviour decision, not a mechanical pick.

**`flattenNativeFields` — this branch moved it, #1606 changed it.**

This branch extracts it from `ReportColumnPicker.tsx` into `shared/utils/flatten-native-fields.ts` so the web auto-collapse resolver and the picker walk the schema the same way. Meanwhile #1606 ("handle array fields in output controls") added a guard to the *local* copy: do not recurse into an ARRAY field's children, since its elements are not reportable columns of their own.

Taking either side wholesale loses something. Taking mine silently reverts #1606; taking main's leaves the extraction with a duplicate definition. So I ported the guard into the extracted function:

```ts
if (!isArrayFieldType(reportType) && field.fields && Array.isArray(field.fields)) {
  result.push(...flattenNativeFields(field.fields, fullName));
}
```

Also in `ReportColumnPicker.tsx`: `isArrayFieldType` stays imported — it is used in five other places in that file — and `effectiveComparisonType` is dropped, because the extracted function was its only remaining user there.

Worth a second look precisely because no test would have caught the wrong choice: this branch's suite passes either way, and #1606's coverage lives on the path I moved.

Verified after the rebase: backend 9783 tests in 622 suites, web 2545 in 269, `nest build` and `tsc -b` clean.

The ESLint job should go green on its own now — #1618 raised the backend lint heap limit and is already on `main`.

PR #1615, opened 2026-09-16
fix(google-sheets): name the first sheet of an auto-created document after the report

laskevych:
## Summary

A Google Sheet created for a report — by `add_report` without `spreadsheet_id`, or with the **Create document** button in the report form — named only the Drive file; its single sheet kept Google's default `Sheet1`. A second report added as a tab of the same document (`spreadsheet_id`) was named after its report, so the document read `Sheet1 / <second report>` and the first report was the only one whose tab did not carry its name. The `add_report` tool description already promised that `name` is "also the new sheet's title".

The create now names the sheet too:

- **Sheets API path** (`createSpreadsheet`): the sheet title is set in the create request itself (`sheets[0].properties.title`), so the tab is never observable as `Sheet1`.
- **Drive API paths** (`createSpreadsheetViaDrive`, `createSpreadsheetInFolder`): Drive cannot name the sheet on create, so `CreateGoogleSheetDocumentService` renames the default sheet right after the create with the new `renameSheet`. This is best-effort, like sharing: the document is already created and usable, so a rename failure (a transient Sheets fault, quota exhaustion) is logged and the sheet keeps Google's default name instead of failing the creation or being reported as a folder-access problem.
- The sheet title is derived with the same `toSheetTitle` rule used for added tabs and Reconnect (trimmed, capped at 100 code points); the Drive file name keeps the full title.

Documents created before this change are not renamed. `update_report` still changes only the report's name in OWOX, as before.

## Changes

- `google-sheets-api.adapter.ts`: `createSpreadsheet` takes a `sheetTitle`; new `renameSheet`.
- `create-google-sheet-document.service.ts`: passes `toSheetTitle(title)` to the Sheets API create, renames best-effort after a Drive create; logs the sheet title and gid.
- `add-report.tool.ts` description and `docs/getting-started/setup-guide/mcp.md`: state that both the file and its sheet are named after the report.
- Changeset `6960-google-sheet-first-sheet-named-after-report.md`.

## Tests

- Adapter spec: `createSpreadsheet` sends the sheet title in the create body; `renameSheet` issues `updateSheetProperties` with a title-only field mask.
- Service spec: the Sheets API path names the sheet in the create and never renames; every Drive path renames the resolved gid to the capped title; a failed rename keeps the created document, still shares it, and returns the normal response.
- Verified locally: the adapter, service, and `mcp-reports.facade.impl` specs pass; `nest build` and eslint on the changed files are clean.


prokopetsv (commented):
Reviewed the branch and ran the adapter, service, and MCP reports facade specs locally: 3 suites, 167 tests pass.

No blockers. The Sheets API path names the tab in the create request, the Drive paths rename after the create and outside the folder-error translation, and the gid comes from the create response. Three 🟡 Non-Blocker notes inline.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
```diff
@@ -195,7 +195,7 @@ type AddReportInput = z.infer<typeof addReportInputSchema>;
...
...sts). For push destinations, the new report runs immediately by default: this starts one billed Report Run and delivers data, then initial_run returns the run_id to poll with get_report_run_status. Set run_immediately=false only when the user explicitly wants to create configuration without delivering data, such as before adding a schedule. Every destination type accepts the same optional output controls as query_data_mart — filters, slices, aggregations, date_buckets, sort — applied on each run: when the user asks to export numbers they saw in a query, copy those parameters verbatim from that query so the report matches what they saw. Google Sheets: a new Google Sheet is created automatically (an external Google Drive side effect) with its file and sheet named after `name`, and linked to the report; pass spreadsheet_id to add the report as a sheet of an existing spreadsheet instead, so several related exports share one document. Looker Studio is pull-based, never runs through this tool, and accepts no name; omit run_immediately or set it false. Email, Slack, Microsoft Teams, Google Chat: requires message; the default initial run sends it to the configured recipients or channels.';
```
prokopetsv, line 198: 🟡 **Non-Blocker: a same-named second report in the shared document now fails.**

`AddGoogleSheetToSpreadsheetService` rejects a duplicate tab title by design (`add-google-sheet-to-spreadsheet.service.ts:76-82`). Before this PR the first tab was `Sheet1`, so adding a second report with the same name as the first as a tab of the same file worked. Now it returns "already has a sheet named …". That is the correct outcome, but it is a new failure the assistant can hit in the flow this description promotes: create "Revenue", then "Revenue" with a different filter and `spreadsheet_id`.

Suggest one clause here telling the assistant that each report added to a shared document needs a distinct name, plus a sentence in the changeset. The error already carries the remedy, so this is guidance only.

.changeset/6960-google-sheet-first-sheet-named-after-report.md
```diff
@@ -0,0 +1,7 @@
+---
+'owox': minor
+---
+
+# Google Sheets: the first sheet of an auto-created document is named after the report
+
+When a Google Sheet is created for a report — by `add_report` through the assistant without a `spreadsheet_id`, or with the "Create document" button in the report form — its single sheet (tab) is now named after the report, like the Drive file already was. Previously that sheet kept Google's default "Sheet1", so a document holding several related exports read "Sheet1 / <second report>": the first report's tab was the only one not carrying its name. Sheets added to an existing spreadsheet and sheets restored by "Reconnect" were already named after their report; documents created before this change are not renamed.
```
prokopetsv: 🟡 **Non-Blocker: the Create document button does not always name the tab after the report.**

`GoogleSheetsTargetSection.tsx:63-67` sends the data mart title when the report still has the default name, so in that case the new tab is named after the data mart. That is pre-existing behavior for the file name and reasonable, but this text says the sheet is named after the report for the button path. A clarifying phrase would keep support from chasing it as a bug.

docs/getting-started/setup-guide/mcp.md
```diff
@@ -482,7 +482,7 @@ Creates a report that exports a data mart to an existing destination (see `list_
...
+- **Google Sheets**: a new Google Sheet is created automatically and linked to the report, then the initial run writes data into it by default; both the file and its sheet are named after the report. Unlike every other tool, this path reaches outside OWOX: it creates a file in Google Drive and attempts to share it with you. When you ask for several related exports, the assistant passes the first report's `spreadsheet_id` to the next `add_report` calls, so each report becomes a separate sheet of the same document instead of a separate file; that sheet is named after its report too. See [Google Sheets destination](../../destinations/supported-destinations/google-sheets.md) for how to set the destination up.
```
prokopetsv: 🟡 **Non-Blocker: docs style.** The first sentence is now about 35 words in passive voice. It was already over the house limit before this PR, but the change extends it. Splitting at the semicolon brings it within the 20-word active-voice rule, for example: "The assistant creates a new Google Sheet and links it to the report. The initial run then writes data into it by default. The file and its sheet both take the report's name."

prokopetsv (approved):
Approved.

Verified locally on the branch: the adapter, service, and MCP reports facade specs pass (3 suites, 167 tests).

No blockers. The Sheets API path names the tab in the create request, so it is never observable as Sheet1. The Drive paths rename after the create, and the rename sits outside the folder-error translation, so a rename failure cannot be misreported as a folder-sharing problem. The gid comes from the create response rather than an assumed 0, and error normalization uses castError as required.

Three non-blocking notes are inline: a same-named second report added as a tab now fails with the duplicate-title error (guidance for the assistant would help), the Create document button names the tab after the data mart when the report still has its default name, and the docs bullet exceeds the 20-word sentence limit. None of these need to hold the merge.

apps/backend/src/ee/mcp/tools/add-report.tool.ts
laskevych: Agreed, applied in 178829c8a. I put the clause on the `spreadsheet_id` parameter rather than the top-level description, since that is where the sheet-naming rule already lives: it now says every report added to the same document needs a distinct name, including the first report's, whose name is also the document's first sheet. The changeset carries the same sentence. One note on the example: "Revenue" then "Revenue" with a different filter on the same data mart is caught earlier by `similar_report_exists` (same fields) and routed to `update_report`; the duplicate-title error is reachable only for a same-named report on a different data mart.

.changeset/6960-google-sheet-first-sheet-named-after-report.md
laskevych: Agreed, applied in 178829c8a. The changeset now says the sheet is named after the document title: the report name for `add_report`, and for the Create document button the report title or, while the report has no name yet, the data mart title.

docs/getting-started/setup-guide/mcp.md
laskevych: Applied in 178829c8a, split into three sentences. I kept the passive subject ("a new Google Sheet is created automatically") rather than "the assistant creates": the tool creates the file, and the neighbouring bullets in this list use the same voice.

PR #1620, opened 2026-09-17
fix(idp): defer PKCE state minting until real sign-in intent

amarchenk0:
## Summary
- Unauthenticated visits to `/auth/sign-in` and `/auth/sign-up` no longer eagerly bounce through Platform to mint a PKCE state; they render locally until the user actually clicks a sign-in action.
- Clicking Google/Microsoft/the email form threads a `pendingAction` through the existing Platform round trip and resumes that exact action once state comes back.
- The email form additionally tries a hidden same-site iframe first, so an already-typed password is never lost or stored anywhere; it falls back to the existing full-page bounce if that doesn't complete within 15s.
- Fixes the redirect loop where a user who didn't act within the state's TTL got bounced to a fresh `/auth/sign-in?state=...` repeatedly.

## Why
Previously, `state` (2-minute TTL, owned by IB) was minted on every unauthenticated page load, before the user showed any intent to sign in. If they didn't complete sign-in before it expired (e.g. left the tab open), the app cleared cookies and redirected back to sign-in with a new state - which could loop.

## Companion change
Requires https://gitlab.owox.com/bi/ui/frontend/-/merge_requests/new?merge_request%5Bsource_branch%5D=feat%2Fframe-ancestors-app-owox-signin (Platform CSP: allow `app.owox.com`/`next.app.owox.com` to frame just `/ui/p/signin` and `/ui/p/signup`) for the hidden-iframe email path to work. Without it, the email form still works correctly via the existing full-page-bounce fallback (functionally safe either way, just an extra retype of the password).

## Test plan
- [x] `npm run typecheck` / `npm run lint` / `npm test` (333/333) in `packages/idp-owox-better-auth`
- [ ] Manual browser QA of the hidden-iframe timing/fallback (not automatable in Jest)

⚠️ No changeset included - this is a user-visible bug fix and normally requires one; the owning Fibery Work Item ID wasn't available at implementation time. Should be added before merge.


packages/idp-owox-better-auth/src/resources/templates/pages/sign-in.ejs
```diff
@@ -125,6 +125,91 @@
...
+        }
+      }
+
+      frame.addEventListener('load', onLoad);
+      const params = new URLSearchParams(window.location.search);
+      params.set('pendingAction', pendingAction);
+      frame.src = `${window.location.pathname}?${params.toString()}`;
+      document.body.appendChild(frame);
+    });
+  }
```
laskevych, line 211: 🔴 **Blocker:** The iframe's final hop is ODM's own `/auth/sign-in?state=…`, which `PageController.signInPage` sends through `sendSecureHtml`. That helper always sets `Content-Security-Policy: frame-ancestors 'none'` (`packages/internal-helpers/src/security/security-headers.ts`), and `'none'` forbids same-origin framing too. So regardless of the Platform CSP companion change, the browser replaces the framed document with an opaque-origin error page, `frame.contentWindow.location.href` throws, `onLoad` takes the "still on Platform" `return`, and `finish(true)` is unreachable.

Net effect for every email sign-in from a state-less page (now the default entry): 15 s with no feedback → full-page bounce → typed password lost, i.e. strictly worse than not having the iframe. Most likely also an extra Platform bounce: browsers store `Set-Cookie` from the blocked response, so the fallback navigation comes back with a new `state` that mismatches the iframe's cookie and goes through the `hasMismatch()` re-bounce in `signInMiddleware` (three states minted for one sign-in).

Options:
- (a) drop the iframe for now and call `ensureStateThenRun('email', …)` directly — same end result minus the stall, and `hasAuthState` can stay `const`;
- (b) give the iframe a dedicated minimal landing response that is frameable by `'self'` (e.g. keyed on `Sec-Fetch-Dest: iframe` or an explicit param; no layout/GTM; `postMessage` to the parent) instead of rendering the full sign-in page. That also avoids loading the whole page and GTM a second time in a hidden frame.

The header values in `security-headers.ts` are documented as pinned, so relaxing them globally isn't the way.

packages/idp-owox-better-auth/src/controllers/page-controller.ts
```diff
@@ -26,7 +41,15 @@ export class PageController {
   }
 
   async signInPage(req: ExpressRequest, res: ExpressResponse): Promise<void> {
+    const hasState = Boolean(extractState(req));
```
laskevych, line 44: 🟡 **Important:** `extractState()` returns the cookie value first, and `idp-owox-state` is a session cookie with no `maxAge` that is only cleared on a successful callback, a mismatch, or a completion error. Before this PR a state-less load always bounced to Platform and a leftover cookie was healed by the mismatch path, so the state was fresh at render time. Now a plain `/auth/sign-in` with a leftover cookie renders `hasAuthState = true`, the click skips minting, and the whole provider/email sign-in runs against an expired (or already consumed) state → completion fails, all auth cookies are wiped, and the user lands on a blank sign-in page with no message. That is the failure this PR is meant to remove, now on the first attempt.

Ways to end up with a leftover cookie: abandoning at the provider chooser; the `/auth/callback` catch (redirects to sign-in without clearing it); password reset → `/auth/password/success` → `/auth/sign-in`; and every browser that opened the sign-in page before this deploys.

Suggestion: treat the state as ready only when this request carries it (`req.query.state`, no mismatch) — the iframe path flips `hasAuthState` client-side anyway, so nothing else depends on the cookie — and clear the state cookie before `redirectToPlatform` in the `pendingAction` branch (the MCP branch already does this with `clearAuthFlowStateCookie`), otherwise the leftover cookie costs an extra mismatch bounce. A `maxAge` aligned with the state TTL in `StateManager.persist` would be a cheap belt-and-braces addition. Same applies to `signUpPage` below.

packages/idp-owox-better-auth/src/resources/templates/pages/sign-in.ejs
```diff
@@ -241,9 +326,34 @@
...
+    // Try the hidden-iframe path first so a password already typed here is
+    // never lost; only fall back to the full-page bounce (which does lose
+    // it, requiring the user to resubmit) if that path doesn't pan out.
+    ensureStateViaHiddenFrame('email')
+      .then(() => {
+        hasAuthState = true;
+        submitEmailSignIn(event);
+      })
+      .catch(() => ensureStateThenRun('email', () => submitEmailSignIn(event)));
+  });
```
laskevych, line 350: 🟡 **Important:** Nothing is locked while the round trip is pending — `submitBtn.disabled` and the status text are only set inside `submitEmailSignIn`, which runs after the state arrives. Each extra Enter/click creates another iframe + timer + Platform round trip; concurrent frames then fight over the single state cookie (the second frame returns with a different `state` → mismatch → `clearAuthFlowCookies` while the first frame's POST may be in flight), and each resolved promise calls `submitEmailSignIn` again (duplicate credential POST). Also, the form is `novalidate` and the empty-field check lives in `submitEmailSignIn`, so submitting an empty form now triggers the whole round trip before "Enter your email and password." can appear.

Suggestion: validate first, then disable the button, show a status, and keep a single in-flight promise that repeat submits reuse.

Two small related things: inside `.catch`, `hasAuthState` is always false, so the `run` callback passed to `ensureStateThenRun` is dead code; and after the full-page fallback `pendingAction=email` does nothing on the return leg — the user lands on an empty form with no hint, so an `info` message asking to enter the password again would help.

laskevych (changes requested):
Deferring the state mint until there is real sign-in intent is the right direction, and the social path (`pendingAction` → round trip → auto-resume, single-use) reads well. Requesting changes because of the blocker on the hidden-iframe email path; the two other inline notes are important but not blocking on their own.

packages/idp-owox-better-auth/src/controllers/page-controller.ts
```diff
@@ -26,7 +41,15 @@ export class PageController {
...
+    const autoSubmitProvider = resolveAutoSubmitProvider(
+      hasState,
+      extractAuthFlowParams(req).pendingAction
```
zapolsky, line 47: **[Security — verified]** The auto-submit is driven by a **query-controlled** `pendingAction`, so a crafted link starts an OAuth flow in the victim's browser with no user gesture.

**Evidence chain:**
1. `page-controller.ts:45-48` feeds `resolveAutoSubmitProvider` from `extractAuthFlowParams(req).pendingAction`.
2. `request-utils.ts:336-338,349` — `extractAuthFlowParams` resolves `pendingAction: pendingAction || cookieParams.pendingAction`, i.e. **the query wins over the cookie**.
3. `page-controller.ts:44` — `hasState = Boolean(extractState(req))`, and `extractState` (`request-utils.ts:155-160`) returns `stateFromCookie || queryState`, so `?state=<anything>` alone satisfies the gate; `owox-better-auth-idp.ts:481` `hasMismatch()` needs *both* sides non-empty (`request-utils.ts:166-170`), so a victim with no state cookie sails through and `:491` persists the attacker's string unchecked.
4. `page-controller.ts:25` returns `'google'` → `sign-in.ejs:129` renders `const autoSubmitProvider = "google";`
5. `sign-in.ejs:354-356` (and `sign-up.ejs:233-235`) run `submitSocial(autoSubmitProvider)` at **top level, on load** — `socialAuthInProcess` is `false` at that point (`partials/social-auth.ejs:2,22`), so nothing gates it.
6. `partials/social-auth.ejs:47-68` POSTs `/auth/better-auth/sign-in/social` and navigates to the provider consent URL.

Cookies are `sameSite: 'lax'` (`cookie-policy.ts:16`) so they ride a top-level cross-site GET, and there is no Origin/Referer/`Sec-Fetch-*`/CSRF check on `/auth/sign-in`.

**Also new:** the query-borne value survives `clearPendingAction` (which only rewrites the cookie, `page-controller.ts:52`). When the social POST returns neither a redirect nor JSON with `url`, `partials/social-auth.ejs:72-73` calls `window.location.reload()`, which preserves the query string → auto-submit fires again → **unbounded reload/POST loop**. The cookie-borne path is immune precisely because `clearPendingAction` makes it single-use.

**Scope, stated honestly.** This is *not* token theft (`callbackURL` is bounded by `trustedOrigins`, no wildcards — `idp-better-auth-config.ts:83-98`) and *not* sign-in-as-attacker (the provider account is the victim's own). It is forced, zero-click OAuth initiation plus the reload loop. The `stateManager.persist` of an unvalidated query state is **pre-existing** on `main` and is not attributed to this PR — what is new is that a URL can now make the page act on it. Verified new: `git show main:…/pages/sign-in.ejs` has `submitSocial` only in the two click handlers; there was no load-time call, and `main`'s `page-controller.ts` has no `hasState`/`autoSubmitProvider` at all.

**Fix:** resume from the cookie only — the legitimate return leg never carries `pendingAction` in the query anyway (`platform-redirect-builder.ts:41-64` doesn't forward it, and it's excluded from `extraParams` by its `AUTH_PARAMS` membership at `request-utils.ts:30`). Add a `readPendingActionFromCookie` mirroring `readPendingActionFromQuery` and use it here. Worth also gating on `req.query.state` rather than `extractState(req)`, so the resume only fires on an actual return leg — that closes the stale-cookie half too. Every test in this PR feeds `pendingAction` through the cookie, so they all stay green.

---

**Same root cause, second reachability path (narrower, but it argues for the same fix).** Both inputs to `resolveAutoSubmitProvider` are cookie-derived, so a bare `GET /auth/sign-in` with *no query string at all* can also auto-submit: `hasState` comes from `extractState` (cookie first) and `pendingAction` from the cookie fallback. `signOutMiddleware` (`owox-better-auth-idp.ts:658-673`) clears `refreshToken` and the Better Auth cookies but neither auth-flow cookie, and neither cookie has a `maxAge` (`cookie-policy.ts:13-21`). Getting both live at once is genuinely narrow — it needs a `pendingAction` round trip abandoned at Platform *while* a state cookie was already present (e.g. the hidden-iframe window at `sign-in.ejs:149`, a second tab, or a refresh-token failure that doesn't clear the cookie at `:584-595`) — so I am not claiming the common sign-out path hits it. But the consequence is the same gesture-free navigation to Google, and gating the resume on `req.query.state` closes both variants at once. Adding `clearAuthFlowCookies(res, req)` to `signOutMiddleware` is a cheap independent hardening.

Worth noting the new tests can't catch either variant: `page-controller.test.ts` only ever supplies state via the **query** (`{ state: 'fresh-state' }`), and the case that sounds like it covers this — *"never auto-submits when there is no state yet, even if a stale pendingAction cookie exists"* — passes only because there is no state cookie either.

zapolsky (commented):
🔴 **Blocked issue** — see the inline comment. This should be addressed before merge, but this review does not formally block; a human owns the merge decision. Posted by automated OWOX review.

packages/idp-owox-better-auth/src/resources/templates/pages/sign-in.ejs
zapolsky: Confirming your blocker independently, and adding the **server-side** half of it, which changes the cost from "no benefit" to "net regression".

`frame-ancestors` is a *response*-time check with no request-time hook, and `SECURITY_HEADERS` (`internal-helpers/src/security/security-headers.ts:17-23`) sets no `frame-src`/`child-src`. So the frame's request is unconditionally sent and the whole ODM → Platform → ODM chain **completes server-side** before anything is refused. The intermediate hops are 302s from `redirectToPlatform`, which never touch `sendSecureHtml`, so only the final document is blocked — after `stateManager.persist` (`owox-better-auth-idp.ts:491`) has already put `Set-Cookie: idp-owox-state=S1` on that same response.

Then `sign-in.ejs:139-147` makes it worse: the fallback copies `window.location.search` and *unconditionally* re-adds `pendingAction`, and `handleNoState` never consults the state cookie — so a perfectly usable S1 is discarded for a fresh S2, and the two collide in `hasMismatch()` (`request-utils.ts:166-170`) → `warn` + `clearAuthFlowCookies` + a third bounce for S3. Not a permanent loop (the clear breaks it on trip 3), and the `warn`'s `queryState` field is unchanged from `main`, so nothing secret is newly logged — but a rare anomaly signal becomes per-attempt background noise.

Your `onLoad` reasoning holds for a second reason too: if the frame ends on `about:blank` instead of throwing, it parses fine and `searchParams.has('state')` is just false — and if no `load` fires at all, `onLoad` never runs. All three paths land on the same 15s timeout.

Pre/post delta for one email sign-in:

| | states minted | Platform round trips | dead wait | password |
|---|---|---|---|---|
| `main` | 1 (on page load) | 1 | 0s | kept |
| this PR | 3 | 3 | ~15s | lost |
| this PR, if the blocked response's `Set-Cookie` is dropped | 2 | 2 | ~15s | lost |

So it isn't merely "no better than not having the iframe" — the email path is measurably worse than `main`. Social is genuinely improved (0 states on load, 1 on click, auto-resumed), so the cost is concentrated entirely on email. Your option (a) looks right to me.

One caveat on my own claim: whether a browser commits `Set-Cookie` from a `frame-ancestors`-blocked response is spec-derived (cookie storage is hooked in HTTP-network-fetch, before the CSP navigation-response check), **not** runtime-verified — I ran nothing. If it turns out false, drop the S3/`warn` row above; the 2-state / 15s / password-lost regression stands either way. DevTools settles it in 30 seconds: submit the email form and watch for `idp-owox-state` appearing while the console shows the `frame-ancestors` refusal.

packages/idp-owox-better-auth/src/controllers/page-controller.ts
zapolsky: Agreeing, and strengthening it: the leftover cookie isn't only reachable by abandonment — **the product's own error paths produce it deterministically.**

`/auth/callback`'s token-exchange `catch` redirects to `/auth/sign-in` with no cookie clearing at all (`owox-better-auth-idp.ts:449-468`), and so does the `!code` branch (`:375-378`). Compare `:380-384` (`!state`) and `:415`, which *do* call `clearAuthFlowCookies`. So the single most likely way to land on sign-in — a state that died mid-flow — is exactly the way that leaves `idp-owox-state` behind. `signOutMiddleware` is a third producer: `:667-668` clears `refreshToken` and the Better Auth cookies but neither auth-flow cookie.

Pre-PR that orphan was healed for free: the page could only be reached with a query state, so `hasMismatch()` fired and re-minted. This PR removes that healing and then trusts the orphan, so the first click after any state failure is now guaranteed to fail — a full visible Google round trip that dumps the user back on a pristine sign-in page (and with no `?error=`, since none of the four `pkce-flow-orchestrator.ts` redirects at `:105/:116/:146/:171` set one).

There's also a cheaper fix than re-plumbing the read, and it composes with your suggestion: `StateManager.persist` (`request-utils.ts:172-175`) calls `setCookie` with no options, and `buildCookieOptions` only sets `maxAge` when `maxAgeMs` is passed (`cookie-policy.ts:13-21`) — so the cookie is a session cookie backing a 1-minute server-side record (`auth-flow-middleware.ts:32`, `ms('1m')`). Passing `{ maxAgeMs: ms('2m') }` there makes `hasState` self-correcting at the source, for every reader, in one line.

Worth noting one more re-entry route your note doesn't cover: `signInMiddleware:491` persists **any** `?state=` with no liveness check, so pressing Back onto a `/auth/sign-in?state=S1` history entry re-plants the dead state after `clearAllAuthCookies` had just removed it. A `maxAge` covers most of that too.

packages/idp-owox-better-auth/src/resources/templates/pages/sign-in.ejs
zapolsky: Confirming, and flagging that the same gap exists on a path your note doesn't cover: **the social buttons**.

`partials/social-auth.ejs:21-27` — `submitSocial` sets `socialAuthInProcess = true` and `setSocialButtonsDisabled(true)` immediately, which is what used to give a Google/Microsoft click its instant busy state. After this PR, `sign-in.ejs:329-334` (and `sign-up.ejs:223-228`) route the click through `ensureStateThenRun`, and on the `!hasAuthState` path (`:139-147`) that never calls `submitSocial` at all — it just assigns `window.location.href`. So on the now-default state-less page, clicking "Continue with Google" leaves the button un-dimmed and still clickable across two full HTTP round trips.

`setSocialButtonsDisabled` is already in scope (same document, the partial is included at `sign-in.ejs:126`), so this is a one-line addition inside `ensureStateThenRun` before the navigation.

zapolsky:

The server-side half of this PR is well built — the query-only `readPendingActionFromQuery`, the single-use `clearPendingAction` (including its deliberate work-around of `persistAuthFlowParams`'s skip-empty shortcut), keeping the MCP `/oauth/authorize` continuation above the new gate, and correctly *not* auto-submitting for `'email'` are all right. `pendingAction` provably cannot leak to Platform (`platform-redirect-builder.ts:41-64` is an explicit per-field allowlist, and `AUTH_PARAMS` membership keeps it out of `extraParams`), and the cookie schema is rolling-deploy safe in both directions (non-strict zod + `TEXT` columns). One blocker is inline; @laskevych's three findings are confirmed and answered in-thread. Everything below is non-blocking.

Primary notes, highest criticality first:

- `packages/idp-owox-better-auth/src/owox-better-auth-idp.ts:521-532` — **the flow where the 2-minute TTL hurts most is exempt from the fix.** The MCP / ChatGPT `/oauth/authorize` continuation is matched *before* the new `!pendingAction` gate, so for connector auth the state is still minted before intent and its TTL still burns while the user reads a consent screen and picks an account. `apps/backend/.../oauth-authorization.controller.ts:236-249` sets `redirect`/`redirect-to`/`app-redirect-to` all to `/oauth/authorize…`, which is exactly what `hasOAuthAuthorizeContinuation` matches. Keeping that fast path is probably deliberate — but then the PR description's claim that unauthenticated visits "no longer eagerly bounce" should be narrowed to say so, because this is the flow with the known production history.
- `owox-better-auth-idp.ts:644-648` — **the deferral silently doesn't apply to invited / project-scoped users on sign-up.** `hasEstablishedIntent` mixes the query-only `readPendingActionFromQuery` with `authFlowParams.projectId` / `.appRedirectTo`, which fall back to the `idp-owox-params` cookie (`request-utils.ts:342,347`) — a session cookie with no `maxAge`. Nothing in the repo ever builds `/auth/sign-up?projectId=…`, and the link at `sign-in.ejs:122` is a bare `<a href="/auth/sign-up">`, so on that route those two gates are in practice satisfied **only** from the cookie. Dominant journey: invite link → `/auth/sign-in?projectId=X` (the render persists `{projectId:X}`) → "Create your account" → `/auth/sign-up` → eager bounce, TTL ticking while they fill the form. To be clear this is an **incomplete fix, not a regression** — pre-PR that route bounced unconditionally. Your own docblock at `request-utils.ts:353-358` states the rule this violates; reading both from `req.query` fixes it.
- `services/auth/pkce-flow-orchestrator.ts:105,116,146,171` — **every "your state died" path lands the user on a pristine form with no explanation.** All four build `new URL('/auth/sign-in', …)` with no `?error=`, and `page-controller.ts:53` only renders a banner when `req.query.error` is set. Pre-PR the user at least saw a visible bounce through Platform; now the page renders instantly and looks identical to the one they just left, so a failed sign-in is indistinguishable from "my click did nothing". This is the user-facing half of the bug the PR sets out to fix, and the rendering already exists (`sign-in.ejs:8-12`) — appending `?error=Your sign-in session expired. Please try again.` at those four sites would close it. (Nice side effect already achieved: `password-flow-controller.ts:130-132`'s `?error=Session expired…` is now actually visible instead of being thrown away by the bounce.)
- `page-controller.ts:44-48` + `owox-better-auth-idp.ts:658-673` — **both inputs to `resolveAutoSubmitProvider` are cookie-derived, and sign-out clears neither auth-flow cookie.** Covered in the inline blocker as the second reachability path; calling it out here because the `signOutMiddleware` half is a one-line hardening that is worth doing on its own merits (`clearAuthFlowCookies(res, req)`), independent of how you fix the resume gate.
- `sign-in.ejs:329-334` / `sign-up.ejs:223-228` — **the social buttons lost their busy state on the now-default path.** `submitSocial` used to set `socialAuthInProcess` + `setSocialButtonsDisabled(true)` immediately (`partials/social-auth.ejs:21-27`); on the `!hasAuthState` path `ensureStateThenRun` never calls it, so a first-time visitor clicking "Continue with Google" gets an un-dimmed, still-clickable button across two full round trips. `setSocialButtonsDisabled` is already in scope — one line inside `ensureStateThenRun`. (Same class as @laskevych's note on the email form, different code path.)
- `page-controller.ts:20-26` — **`resolveAutoSubmitProvider` never checks the provider is enabled**, although `this.providers` is right there at `:33`. On a deployment with Google disabled, `?pendingAction=google` burns a full Platform round trip and then auto-POSTs for a button that was never rendered; the failure surfaces only in `console.error` (`social-auth.ejs:74-75`).
- **Question (unproven — needs a Platform-side answer):** does this remove **zero-click re-authentication** for users with a live Platform session but a lapsed ODM refresh token? `apps/web` bounces to `/auth/sign-in` with *no* query context at all on any refresh failure (`AuthGuard.tsx:43-46`, `AuthContext.tsx:268,318` → `auth-api.service.ts:82-100`, which stashes the target in `RedirectStorageService` rather than the URL). Pre-PR that hit the unconditional `redirectToPlatform`; now it renders the form. Whether anything is actually lost depends on whether Platform auto-issues a `code` to a browser carrying a live session — the existence of `/auth/callback?code=&state=` handling suggests it can, but that behavior lives in the Identity service and I could not settle it from this repo. If it does, this is a silent regression on the most common auth event in the product and should be stated in the PR description and changeset; if it doesn't, the deferral costs nothing here. Worth one check before merge. (Note the fix is *not* simply to mirror `signUpMiddleware`'s gate — that's the cookie-backed one flagged two notes up.)

<details>

- `owox-better-auth-idp.ts:483` / `:639` — `clearAuthFlowCookies` is immediately undone for the params half: `redirectToPlatform` re-reads `extractAuthFlowParams(req)` from the **request** (`:895`) and appends a second `Set-Cookie` (`:923` → `request-utils.ts:371`), and the later header wins. So after a state mismatch the browser keeps `idp-owox-params` (now including `pendingAction`) and loses only `idp-owox-state`. This is **pre-existing**, and it's arguably load-bearing — a genuine clear would silently drop the user's click. Reporting it as a naming/expectation trap: anyone reading `clearAuthFlowCookies` as "we dropped all auth-flow context here" is misled. Consider splitting it or passing explicit params to `redirectToPlatform`.
- `page-controller.ts:50-51` — the "single-use" comment isn't true for a query-borne `pendingAction`: `clearPendingAction` only touches the cookie while `extractAuthFlowParams` prefers the query, so such a URL replays on every reload. Either soften the comment or `history.replaceState` the param away after the resume.
- `owox-better-auth-idp.ts:491` — `stateManager.persist` writes **any** `?state=` with no liveness or store check, so Back onto a `/auth/sign-in?state=S1` history entry re-plants a state that `clearAllAuthCookies` had just removed. Pre-existing, but the new render path makes it user-visible; a `maxAge` on the state cookie covers most of it.
- `request-utils.ts:172-175` — `StateManager.persist` passes no `maxAgeMs`, so a cookie backing a ~1-minute record (`auth-flow-middleware.ts:32`) is a session cookie. `setCookie` already supports the option; this is the cheapest fix for the whole stale-state class (details in the reply to @laskevych).
- `sign-in.ejs:335-350` — no single-flight guard on `ensureStateViaHiddenFrame`: each extra Enter/click appends another iframe and drives another Platform round trip, and they all contend for one `idp-owox-state` slot, so the parent can resolve against a state that a later frame then overwrites.
- `partials/social-auth.ejs:2,22-27,76-81` — `socialAuthInProcess` stays latched when navigating away by design; now that a page can auto-submit without a gesture, a Back navigation that restores from bfcache would return to a page with every social button permanently disabled. Chrome is likely protected (`Cache-Control: no-store` from `protocol-middleware.ts:160` blocks bfcache); Safari/Firefox may not be. Unproven without a browser, but a `pageshow`/`event.persisted` reset is one line and makes it moot.
- `partials/social-auth.ejs:41` — the "Last used provider" write happens *before* the fetch, unlike the email path which writes only on success (`sign-in.ejs:291,298`). With auto-submit this now runs with no user action, so a failed attempt leaves the badge pointing at the provider that just failed, steering the user back into the same broken path.
- `request-utils.ts:336-338,359-361` — both new readers hand-roll the `typeof === 'string'` check instead of using the repo's own array-safe `readQueryString` (`:221-225`, as `auth-error-controller.ts:70` does). `?pendingAction=google&pendingAction=google` parses to an array → treated as absent → the click is silently swallowed.
- `request-utils.ts:42` + `owox-better-auth-idp.ts:645` — `'email'` counts as established intent on **sign-up**, where there is no email sign-in action at all (the form is a magic link that needs no PKCE state). `/auth/sign-up?pendingAction=email` performs exactly the eager bounce this PR removes. Latent rather than live, since nothing generates that URL.
- Tests — the suites are honest about the server-side gates, but the highest-risk new logic is asserted only by substring matching, and two assertions are near-tautological (`template-service.test.ts:31-32` checks for literals that never existed, and would still pass if the social buttons were routed through the iframe helper). Worth pinning instead: the `Content-Security-Policy` value `signInPage` emits (`page-controller.test.ts:11` currently stubs `setHeader` with no assertion — the one place the iframe conflict could have been caught); a fixture with an `idp-owox-state` cookie; the pendingAction-only cookie shape, which takes the `clearCookie` branch that `lastParamsCookieValue` ignores entirely; and `tokenFacade`, whose absence makes the `handleExistingRefreshToken` fall-through untestable.
- No changeset. Required here per `docs/contributing/repository/release-strategy.md` ("a bug fix whose effect is visible to users"; precedent in this package: `.changeset/6894-user-bound-project-redirect.md`). You already flagged this — noting only that the filename must start with the owning Fibery public ID, which shouldn't be guessed or substituted with the PR number.
- The client-side state machine is duplicated across both templates (`ensureStateThenRun` at `sign-in.ejs:139-147` and `sign-up.ejs:105-113`, plus the auto-submit block) although `partials/social-auth.ejs` is the package's established shared-script pattern and says so in its own docblock. The tests now pin the divergence down to `let` vs `const`, so unifying them later also means touching tests.
- Cross-repo rollout isn't documented in-repo, only in a code comment (`sign-in.ejs:156-163`), while `docs/review-rules.md` asks for mixed-version compatibility to be documented and verified. `packages/idp-owox-better-auth/README.md:121` still describes the pre-PR sign-in contract.
- minor: `AutoSubmitProvider` (`page-controller.ts:18-26`) re-declares a subset of `PENDING_ACTION_VALUES` as literals with a `string` parameter, so adding a provider silently fails to auto-submit with no compile error · `JSON.stringify` output is interpolated raw into `<script>` without escaping `<` (safe today, but `renderSignIn` takes `Record<string, unknown>`, so types don't protect it) while `hasAuthState` uses HTML-escaping `<%= %>` to emit a JS literal — two conventions for one job · `ensureStateThenRun` drops `window.location.hash` · three unrelated timeouts now gate one flow (`STATE_FRAME_TIMEOUT_MS = 15000`, `ms('1m')`, and Platform's 2 minutes) with none derived from another.

</details>

<sub>Automated OWOX review · worktree multi-agent passes (5 finders + 2 second-wave, every blocker adversarially verified) · nothing executed locally; CI parity read from `gh pr checks`: 16/16 green.</sub>

PR #1621, opened 2026-09-17
feat(mcp): tell the agent to use a metric the Data Mart already computes (#6964)

zapolsky:
## Why

A Calculated Field whose formula aggregates is published with an empty `allowedAggregations`, so that an agent cannot ask for an aggregation the validator would refuse with a 400. Read in the payload, though, that empty list is the only thing the field says about itself: it sits beside the very columns it is computed from, which advertise `SUM`, under a tool description that tells the agent to build queries from `allowedAggregations`. Read that way it is a dead end, so the agent pulls the inputs and divides them in its own reply.

Everything that would have told it otherwise — what `calculated.level: "metric"` means and what to do with such a field — lives in the tool's `outputSchema`, which clients weight far below the rows they are handed.

This matters beyond tidiness. Arithmetic the agent performs is performed over the rows the `limit` let through, and the number it prints carries no truncation marker. Selected as a field, the metric is computed in the warehouse over every matching row, with whatever weighting or `NULLIF` guard the formula holds — and the definition stays the one the analyst wrote.

## What changed

- **`get_data_mart_details_by_id` carries the affordance in the data.** Every aggregate-level Calculated Field now ships a `usage` note saying the value is already computed at the grain the query asks for, and that it must not be recomputed from other fields or named in `aggregations`. Aggregate level only — a row-level formula is a dimension a query is supposed to aggregate, and a note there would be both wrong and the noise that stops this one landing.
- **Both tool descriptions say what the empty set means.** `query_data_mart` gains an explicit preference rule; `get_data_mart_details_by_id` states that an empty `allowedAggregations` on a `metric`-level field means *already computed*, not *unusable*.
- **Docs**: the two tools in `mcp.md`, plus a new "Using the Field from an AI Assistant" section in `calculated-fields.md`.

No schema change, no migration, no change to any Data Mart.

## Verification

**Unit** — 473 tests across 40 suites in `src/ee/mcp`, green. The three new ones were written first and watched fail: the `usage` note in the payload, and each description's new text pinned verbatim so that editing either is a visible change in the spec rather than a quiet change in a model's context.

**Live** — a fixture Data Mart built for this check: a SQL mart that synthesises its own rows (nothing to seed, nothing to go stale), with `date / source / medium / campaign / revenue / adCost` and a `metric`-level `ROAS = SUM(revenue) / NULLIF(SUM(adCost), 0)`. 61 sources over 7 days, so the default row limit truncates. Driven through a real MCP client in four fresh conversations, asking as a business user would and never naming the field:

| | result |
| --- | --- |
| `ROAS` named in `fields` | **7 / 7** calls |
| `SUM` aggregations on its inputs alongside | **7 / 7** |
| raw money columns as `GROUP BY` keys | **0 / 7** |

Every value returned matched the warehouse ground truth, per source and in the totals.

Three behaviours were confirmed live for the first time in the same run: a filter on a `metric`-level Calculated Field lands in `HAVING`; sorting by one renders the expression rather than the alias; and when the row limit truncates, the agent raises the limit instead of dividing the partial rows.

## Notes for the reviewer

The `usage` string and both description fragments are asserted verbatim in the specs — they are prompt surface, and a stale sentence in them fails no other test while silently degrading every agent that reads it.


laskevych (approved):
No blocking findings.

Checked: the `usage` note is decided by the same `isAggregateLevel` rule as the empty `allowedAggregations`, so the two cannot disagree; it never lands on joined, nested or array fields (the facade omits joined calculated fields); the payload still validates against `outputSchema`; `totals` keep returning the metric recomputed over all matching rows, so the existing "use totals directly" rule stays correct; the changeset is present and named per the release policy. CI is green.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -86,6 +86,7 @@ When building the query:
 - Request only the fields relevant to the user's question — never request all fields.
 - Use limit to control how many rows come back (1–1000, default 20). There is no offset/pagination: the tool returns a bounded subset.
 - aggregations: SUM, COUNT, COUNT_DISTINCT, AVG, MIN, MAX, and percentiles P25/P50/P75/P95 — which of them a given field allows depends on the field's type and the data mart's per-field settings (see the matrix below). Group-by is implied by the non-aggregated fields you select.
+- Prefer a metric the data mart already defines: when get_data_mart_details_by_id shows a field whose "calculated" level is "metric", that field IS the quantity, computed at the grain your query asks for — list it in "fields" and read the value. Do not rebuild it by aggregating its inputs and dividing them yourself; an empty allowedAggregations means "already computed", never "unusable".
```
laskevych: 🟡 Important (non-blocking): the closing clause — `an empty allowedAggregations means "already computed", never "unusable"` — reads as a general definition of `[]`, but this server publishes an empty set for other field classes where "already computed" is false: array fields, a native field whose aggregations the analyst cleared (or that holds only non-MCP functions like `ANY_VALUE`/`STRING_AGG`), and a joined field with an empty `postJoinAggregations` override. For a cleared numeric `cost`, an agent taking the clause literally selects `["campaign", "cost"]` with no aggregation, `cost` becomes a GROUP BY key, and raw per-row values get presented as OWOX-computed totals.

The sibling sentence in `get_data_mart_details_by_id` is scoped correctly, so one word fixes it here — `its empty allowedAggregations…` / `on such a field, an empty…` (the spec pins this substring, so it needs the same edit). Worth a glance at `makeAllowedAggregationsField` too: "an empty array means the field cannot be aggregated" sits right on the property agents misread and could carry the reason in a few words.

laskevych: 💡 Suggestion: "list it in "fields" and read the value" doesn't say what happens to the inputs if the agent lists them beside the metric. `fields: ["campaign", "revenue", "adCost", "roas"]` with no aggregations is a valid query: the metric forces the aggregated path, every projected column without an aggregation becomes a grouping key, and ROAS comes back at near-row grain, truncated at the default limit, with no error (auto-collapse is not enabled on the MCP path).

Your live run shows 0/7 such calls, so this is cheap insurance rather than a fix — a short clause like "if you also list its inputs, aggregate them — otherwise they become group-by keys" would close it.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
```diff
@@ -321,6 +337,9 @@ export class GetDataMartDetailsTool implements McpToolDefinition<GetDataMartDeta
               aggregationRole: out.aggregationRole as AggregationRole | undefined,
               allowedAggregations: out.allowedAggregations as ReportAggregateFunction[] | undefined,
             });
+      if (isAggregateLevelCalculatedField(out)) {
+        out.usage = AGGREGATE_LEVEL_USAGE;
+      }
```
laskevych, line 342: 🔭 Follow-up (not for this PR): the note is unconditional, so it is also stamped on a metric that can't be computed for this caller — a formula whose input column got DISCONNECTED (`isConnected()` is true for every calculated field and `prepareSchema` strips `status`), or one that reads a joined Data Mart the caller has no reporting access to. The payload then says "list it in fields and read the value" and the query answers 400 `calculated_field_broken` / `permission_denied`.

The publication predates this PR and the error messages recover the agent, so the cost is one paid round trip — but now that agents are steered towards these fields, it may be worth omitting `usage` (or the field) when its references are broken.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.spec.ts
```diff
@@ -325,6 +325,43 @@ describe('GetDataMartDetailsTool', () => {
...
+  it('tells the agent, in the payload, that an aggregate-level field is already computed', async () => {
+    const facade = {
+      getDataMartDetails: jest.fn().mockResolvedValue({
+        id: 'dm_1',
+        name: 'Ads',
+        description: '',
+        fields: [
+          { name: 'revenue', type: 'FLOAT' },
+          { name: 'roas', type: 'FLOAT', calculated: { level: 'metric' } },
+          { name: 'ctr', type: 'FLOAT', calculated: { level: 'column' } },
```
laskevych, line 343: 🧪 Test: the level-less case (`calculated: {}`) is the one `isAggregateLevel` is spelled `!== 'column'` for, and it is asserted (L503–514) for `allowedAggregations` only. Adding `{ name: 'legacy_ratio', type: 'FLOAT', calculated: {} }` here with the same `usage` expectation would pin it — today narrowing this `if` to `level === 'metric'` passes every spec.

.changeset/6964-mcp-prefers-ready-calculated-metrics.md
```diff
@@ -0,0 +1,13 @@
+---
+'owox': minor
+---
+
+# AI assistants now use the metrics your Data Mart already defines
+
+An assistant connected through the MCP server now selects a Calculated Field whose formula aggregates — a ROAS written as `SUM(revenue) / NULLIF(SUM(adCost), 0)`, a CTR, a conversion rate — instead of pulling the columns behind it and doing the division in its reply.
+
+That matters for two reasons. OWOX computes the field in your warehouse over every row matching the question, so the answer stays correct even when the row limit truncates what the assistant sees; arithmetic performed on truncated rows looked like a normal number and gave no sign it was incomplete. And the definition stays the one you wrote — any weighting, exclusion or `NULLIF` guard inside the formula is applied, rather than quietly lost in a plain division.
```
laskevych: 🧹 Nit (release text): "gave no sign it was incomplete" is a bit stronger than what shipped before — the response already carried `truncated` / `truncation.reasons` and an `_instruction` telling the assistant to warn the user and to label its own arithmetic. What lacked a marker was the number the assistant computed itself. Since this lands in the public changelog, something like "…and nothing on the number itself showed it was incomplete" keeps the point without implying earlier answers were silently wrong. Same phrasing in `calculated-fields.md` ("carries no sign of either fact").

docs/getting-started/setup-guide/calculated-fields.md
```diff
@@ -307,6 +307,16 @@ A calculated bucket carries **no time zone**, on any of the five warehouses. The
...
 
+## Using the Field from an AI Assistant
+
+An assistant connected through the [MCP server](./mcp.md) sees a calculated field sitting beside the very columns it is computed from, and it has to be told which of them to ask for. Left to itself it pulls `revenue` and `adCost`, adds them up and divides them in its own reply — arithmetic the assistant performed, over only the rows the row limit let through, and the number it prints carries no sign of either fact.
+
+So a **metric** is published with a `usage` note saying the value is already computed at whatever grain the question asks for, and the assistant is told to select it by name rather than rebuild it. Its `allowedAggregations` is empty for the reason it is empty everywhere else — the formula already aggregates — and that empty list means _already computed_, never _unavailable_.
+
+A **dimension** carries no such note, because aggregating one is exactly what a query is allowed to do.
+
+Nothing about the Data Mart changes here. A field defined as `SUM(revenue) / NULLIF(SUM(adCost), 0)` has always been recomputed correctly for each group the question asks for; what changed is that the assistant now asks for it.
```
laskevych: 📝 Docs: on this page `revenue` is already a calculated field (`revenue = SUM(amount)`), and `SUM(revenue)` is the canonical *refused* formula in two tables above ("Wrapping an aggregated calculated field in an aggregation"). Using `SUM(revenue) / NULLIF(SUM(adCost), 0)` as the valid example here clashes with that for someone reading top to bottom — reusing the page's own `roas = revenue / NULLIF(cost, 0)` would keep it consistent. `mcp.md` and the changeset are standalone, so the example is fine there.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.spec.ts
```diff
@@ -57,6 +57,22 @@ describe('QueryDataMartTool', () => {
     expect(tool.description).toContain('Arrays are column-only');
   });
 
+  // The gap this closes is in query BUILDING, not in reading the response: the description already
+  // says never to recompute a value the response contains, and says nothing about preferring a
+  // metric the schema already defines. So the agent pulls the inputs, aggregates them, and divides
+  // — arithmetic it performs over the rows that survived the limit, with no truncation marker on
+  // the number it prints. Prod, 03.09–17.09: a ready metric was named in 32 of 292 calls.
```
laskevych: 🧹 Nit: "Prod, 03.09–17.09: … 32 of 292 calls" (and "32 of 292 prod calls" in `data-mart-details.tool.spec.ts`) puts production usage numbers into a public repo, with a year-less date range that can't be checked after merge. The qualitative reason is enough for the comment; the numbers fit better in the PR description.

laskevych (commented):
Non-blocking notes to go with the approval above.

💡 Suggestion: the sibling query-building rules ("how many" → COUNT, joined Unique Count, "use server-provided totals directly") live both in the `query_data_mart` description and in the server instructions (`mcp-system-instructions.ts`), and the PRs that introduced them touched both. The new "prefer a metric the data mart already defines" rule is only in the tool descriptions, while the instructions still only say "make it clear whether OWOX returned/calculated it or whether you calculated it yourself". A one-line rule there would cover clients that weight server instructions above per-tool descriptions; `mcp-instructions.service.spec.ts` uses `toContain`, so nothing breaks.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Fixed — `on such a field an empty allowedAggregations means "already computed", never "unusable"`.

The general reading was wrong for at least three classes, and I checked rather than took it on faith: `effectiveMcpAggregations` returns `[]` for an array type outright, and filters everything else through `MCP_AGG_SET`, so a field left with only `ANY_VALUE`/`STRING_AGG` lands empty too. The spec now pins the scoped substring and asserts the unscoped one is gone.

Took the second half as well. `makeAllowedAggregationsField` no longer says only "an empty array means the field cannot be aggregated" — it names the three causes and points at `calculated`/`usage` to tell them apart.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.spec.ts
zapolsky: Fixed, and it was load-bearing exactly as described. Verified by sabotage in both directions: narrowing the emit to `level === 'metric'` passed 473/473 before, and with `{ name: 'legacy_ratio', type: 'FLOAT', calculated: {} }` added to that fixture the same sabotage now fails this spec.

.changeset/6964-mcp-prefers-ready-calculated-metrics.md
zapolsky: Fixed in both places. Checked the response shape first — `truncated`, `truncation.reasons` and the `_instruction` telling the assistant to warn the user are all there, so the sentence did overstate what shipped before.

Changeset now reads "…and nothing on the number itself showed it was incomplete"; the docs line reads "…and nothing on the number itself shows either".

docs/getting-started/setup-guide/calculated-fields.md
zapolsky: Fixed. Verified on the page: `revenue = SUM(amount)` at L148, and `SUM(revenue)` appears as the refused example twice, L171 and L243 — so the section was restating a formula the page refuses four sections earlier. It now points at that page's own `roas` via an anchor instead of carrying its own example. `mcp.md` and the changeset left as they are, per your note.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Not taking this one for this PR.

The reasoning is right: it is a valid query, group-by is implied by the projected columns, and nothing on the MCP path collapses it. But the evidence around it is worth recording, because it looks contradictory.

An earlier run DID produce exactly that shape, and I nearly added the clause on the strength of it. Those calls came from a conversation that had been through a licence-block retry loop: it had sent the right shape first — the metric plus `SUM` on its inputs — got `query_failed` naming "aggregations", and dropped them eleven seconds later. Every later call in that conversation kept the damaged pattern. Four fresh conversations afterwards gave 7/7 with the aggregations present.

So the shape is reachable, but so far only after the context has been poisoned by something else. Weighed against another sentence in a description that is already long, the decision was to leave it and revisit if it turns up in production.

apps/backend/src/ee/mcp/tools/data-mart-details.tool.ts
zapolsky: Agreed, and out of scope here as you say. Recorded as a follow-up: omit `usage` (or the field) when its references are broken, or when it reads a joined Data Mart the caller cannot report on.

One thing that did change with this PR is the likelihood rather than the mechanism — the wasted round trip was mostly hypothetical while agents ignored these fields, and 7/7 selection makes it a real path.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.spec.ts
zapolsky: Removed from both spec comments; the qualitative reason stays, and the numbers live in the PR description instead.

One I cannot clean up the same way: the first commit message carries the same figures and is already pushed. Removing it means rewriting a commit under an open review, which is not something I will do unprompted — say the word if you would rather take that trade.

PR #1622, opened 2026-09-17
feat(destinations): allow creating Microsoft Excel destinations in the web app

romandubovyi:
## What

While the Excel add-in was under Microsoft Marketplace verification, **Microsoft Excel** was hidden from the destination type list in the web app. The add-in is now verified and published, so users can create Excel destinations themselves again.

- `canCreateDestinationInApp` guard and its filter in `DestinationTypeField` removed; Excel is listed right after Google Sheets.
- The add-in keeps resolving or creating an Excel destination automatically when the caller has none they can use (`ResolveExcelDestinationService` unchanged).
- Form hint and docs explain that a title is all that is needed, and that the add-in always uses the oldest Excel destination the user can use.
- `manage-destinations.md` lists Excel among the types and no longer says it cannot be created from **+ New Destination**.
- MCP `add_destination` still withholds Excel; its comments now give an MCP-only reason instead of pointing at the removed web guard.

## Also fixed

`DataDestinationForm` now clears the copy-credentials source when the destination type changes. Previously, picking a source under Google Sheets and switching to another type sent `sourceDestinationId` for a type that rejects it (HTTP 400 on a form with no credential fields).

## Follow-up

Picking between several Excel destinations inside the add-in lands in a separate OWOX/owox-data-marts-spreadsheets PR.

## Checks

- `npm test -w @owox/web`: 264 files, 2489 tests
- eslint, prettier, `tsc -b` for web; eslint, prettier for the two backend files
- markdownlint for both docs pages
- Changeset: `.changeset/6966-excel-destination-manual-creation.md`, `owox` minor


apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/FormDescriptions/descriptions/ExcelDescription.tsx
```diff
@@ -7,8 +7,10 @@ export default function ExcelDescription() {
       <AccordionTrigger>How do I connect to Microsoft Excel?</AccordionTrigger>
       <AccordionContent>
         <p className='mb-2'>
-          There is nothing to set up here. The OWOX add-in for Excel creates this destination for
-          you the first time you build a report from a workbook.
+          There is nothing to set up here beyond a title. You can also leave this to the OWOX add-in
+          for Excel: if you have no Excel destination you can use, it creates one the first time you
+          build a report from a workbook. The add-in always uses the oldest Excel destination you
```
laskevych: 🟡 **Important:** reword "The add-in always uses the oldest Excel destination you can use, so a second one is not picked up by it" — the companion PR `OWOX/owox-data-marts-spreadsheets#62` makes it false. Same sentence at `docs/destinations/supported-destinations/microsoft-excel.md:5`.

**Why:** #62 adds `if (excelDestinations.length > 1) { setShowDestinationPicker(true); return; }` to `CreateReportForm`. With two or more Excel destinations the add-in stops calling `POST /data-destinations/excel` and asks the user to choose, so a second one *is* picked up — #62 states that contract in its own `data-destination-service.ts` doc comment. Both copies here are user-facing (docs.owox.com and this accordion), and this is the text a user reads at the exact moment they create the second destination. Because the add-in reaches users through the Microsoft Marketplace on Microsoft's update schedule, the two statements cannot be made to flip at the same moment.

**Fix:** state the rule without claiming permanence, e.g. "If your project has more than one Excel destination, which one the add-in uses depends on your add-in version — older builds always take the oldest one you can use." Or drop the sentence here entirely, keep one copy on the docs page this accordion already links to, and make updating that page part of #62's checklist.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/DataDestinationForm.tsx
```diff
@@ -162,6 +162,15 @@ export function DataDestinationForm({
   // Get the current destination type
   const destinationType = form.watch('type');
 
+  // The copy-source picker lives inside the Google Sheets fields only, so once the type changes
+  // its clear button is gone while the selection would still ride along into the request as
+  // sourceDestinationId — which every other type rejects.
+  useEffect(() => {
+    if (selectedSource) handleSourceClear();
```
laskevych, line 174: 🟡 **Important:** `handleSourceClear()` only half-clears here — its `form.resetField('credentials.credentialId')` (line 159) is a no-op unless that field was registered, so on service-account-only setups the placeholder survives the type change.

**Why:** `credentials.credentialId` is registered only by the `FormField` at `GoogleSheetsFields.tsx:221`, guarded by `isOAuthAvailable && authMethod === 'oauth'`, while the copy-source button renders unconditionally at `:188`. react-hook-form 7.87's `resetField` bails on `get(_fields, name)`. So with destination OAuth unconfigured, or with the user on the Service Account JSON tab:

1. pick a copy source → `setValue('credentials.credentialId', COPY_SOURCE_CREDENTIAL_PLACEHOLDER)`;
2. switch the type to Excel → this effect nulls `selectedSource` but leaves the placeholder behind;
3. switch back to Google Sheets and save → Zod passes (the placeholder is a valid UUID for `googleCredentialsWithOAuthSchema`), `!dirtyFields.credentials` deletes `payload.credentials` because the placeholder was set with `shouldDirty: false`, `source` is now null so the sheet takes the mapper branch, and `GoogleSheetsMapper.mapToCreateRequest` dereferences `formData.credentials.serviceAccount` on `undefined`. Save silently does nothing.

The same latent bug is already reachable through the manual clear button, so this PR does not introduce it — but this effect is what makes it fire without the user asking for it.

**Fix:** clear the value rather than the registration — `form.setValue('credentials.credentialId', initialData?.credentials?.credentialId ?? undefined, { shouldDirty: false })` — or guard `GoogleSheetsMapper.mapToCreateRequest` against a missing `credentials`.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/DataDestinationForm.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  it('drops the picked copy source once the destination type changes', async () => {
+    // The clear button lives inside the Google Sheets fields, so after the switch nothing
+    // else could undo the selection — and every other type rejects sourceDestinationId.
+    const onSubmit = renderForm();
+
+    fireEvent.click(screen.getByRole('button', { name: 'Pick source' }));
+    fireEvent.click(screen.getByRole('button', { name: 'Switch to Excel' }));
+    fireEvent.submit(screen.getByRole('button', { name: /save/i }));
```
laskevych: 🧪 **Test gap:** this test does not pin the behaviour it names — it submits a disabled button, and its payload assertion holds with the effect deleted.

**Why:** three independent problems.

1. The `DestinationTypeField` stub calls `form.setValue('type', EXCEL)` without `shouldDirty`, unlike the real Controller's `field.onChange`. So `isDirty` stays false and every term of the Save button's `disabled` expression is false — this line only works because `fireEvent.submit` dispatches a bubbling event that reaches the `<form>` regardless of the button's state.
2. `expect(data).not.toHaveProperty('credentials')` is satisfied by the `!dirtyFields.credentials` arm in `handleSubmit` whether or not the effect ran, because `handleSourceSelect` sets the placeholder with `shouldDirty: false`.
3. The `GoogleSheetsFields` stub never registers `credentials.credentialId`, so `resetField` no-ops inside the test exactly as it does in production on the service-account path.

Delete the whole `useEffect` and only `expect(source).toBeNull()` fails.

**Fix:** drive the real `DestinationTypeField` (as `DestinationTypeField.test.tsx` already does) instead of stubbing it, click Save rather than dispatching `submit`, and assert the request rather than the callback argument — `source` only becomes `sourceDestinationId` one level up in `DataDestinationConfigSheet.tsx`, which is itself uncovered because `DataDestinationConfigSheet.test.tsx` mocks `DataDestinationForm: () => null`.

.changeset/6966-excel-destination-manual-creation.md
```diff
@@ -0,0 +1,14 @@
+---
+'owox': minor
+---
+
+# Microsoft Excel destinations can now be created from the web app
+
+**Microsoft Excel** is back in the destination type list. While the OWOX add-in
+for Excel was under Microsoft Marketplace verification, the only way to get an
+Excel destination was to let the add-in create one on your first report run.
+Now you can also add one yourself from the Destinations page: give it a title
```
laskevych, line 10: 🟡 **Important:** the changeset covers only one of the PR's two user-visible outcomes, and omits the caveat both doc surfaces carry.

**Why:** `docs/contributing/repository/release-strategy.md` requires a changeset for "a bug fix whose effect is visible to users". The commit message names the second outcome itself — "switching away from Google Sheets no longer sends sourceDestinationId for a type that rejects it" — and its effect was a user-facing HTTP 400 from `create-data-destination.service.ts`. Separately, the release note says users can now add an Excel destination themselves but never mentions what both `microsoft-excel.md` and `ExcelDescription.tsx` do say: that today's add-in will not use it unless it is the oldest. The same policy asks for "what changed and its practical effect".

**Fix:** add two sentences to this file — one for the copy-source reset, one naming the oldest-destination caveat. A second changeset would need its own Work Item ID, so only split if the copy-source fix is owned by a different one.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/DestinationTypeField.test.tsx
```diff
@@ -79,26 +63,16 @@ describe('DestinationTypeField', () => {
...
-    // Filtering it out unconditionally would leave the type field of an existing Excel
-    // destination blank.
-    render(<ExcelTestForm />);
-
-    openTypeList();
-
-    expect(
-      await within(document.body).findByRole('option', { name: /Microsoft Excel/ })
-    ).toBeInTheDocument();
+  it('lists Microsoft Excel right after Google Sheets', () => {
```
laskevych: 🧪 **Test gap:** this test never renders `DestinationTypeField` — it restates `DataDestinationTypeModel` back to itself, in a component spec, twenty lines below a second copy of the same name list.

**Why:** the two deleted tests exercised the component's option list; this one asserts `getAllTypes().map(d => d.displayName)`. The substantive change in this PR is removing `.filter(({ type }) => type === field.value || canCreateDestinationInApp(type))` from the component, and nothing now covers the rendered list's filtering — if a filter comes back, this test stays green. It also hand-keeps a second catalogue: adding a destination type means editing both this array and the loop at line 47. `data-destination-type.enum.test.ts` states the principle already — "Derived rather than written out: a second hand-kept list is what let Excel be added to one place and forgotten in the other."

**Fix:** assert the order on the rendered options inside the existing render test (`within(document.body).getAllByRole('option')`), which leaves one catalogue and covers the list users actually see. If a pure model assertion is wanted, it belongs in a `data-destination-type.model.test.ts` beside the model — the pattern `data-storage-type.model.test.ts` already follows.

apps/backend/src/data-marts/facades/mcp-destination-type.ts
```diff
@@ -23,13 +23,12 @@ export type McpDestinationType = (typeof MCP_DESTINATION_TYPES)[number];
  *
  * The vocabulary itself has to stay exhaustive over the enum, because every type is *named* on
  * the way out — `list_destinations` and `get_data_mart_reports` report whatever a project
- * already has, Excel included. Being creatable is a separate question, and Excel answers it the
- * same way it does in the web app's type list: the add-in resolves one on first use and it
- * holds nothing to fill in, so offering it would invite a second destination indistinguishable
- * from the automatic one — and possibly not the one the add-in goes on to use.
+ * already has, Excel included. Being creatable is a separate question: this tool has no
```
laskevych: 💡 **Suggestion:** the rewrite drops the only substantive reason and leaves one that restates the exclusion — "this tool has no creation flow for Excel" is the decision, not a cause for it.

**Why:** the deleted text gave a real reason (a second destination "possibly not even the one the add-in goes on to use"). What remains is "no creation flow" plus "it holds nothing an agent could fill in" — and this PR is the proof that the second clause argues the other way: the web form now creates one from a title alone, through an `ExcelMapper` that already existed. `add_destination` already has exactly that shape for `looker_studio` (direct create, no credentials, no OAuth). So the file now says the web app accepts something it also says an agent could not fill in.

**Fix:** pick one. Either wire Excel like `looker_studio` and drop `MCP_NON_CREATABLE_DESTINATION_TYPES` with its `.exclude()` and the spec assertion, or restore the real reason and say plainly why the web app is exempt from it — for example, a person can see the existing rows in the list first, an agent creating one blind cannot.

laskevych (approved):
## Review: approve

Verified the feature end to end against the backend and the companion add-in PR `OWOX/owox-data-marts-spreadsheets#62`. No blocking findings — the inline notes and the four below are all non-blocking.

What I checked and found correct:

- The REST path already accepts a credential-less Excel create (`requiresCredentials(EXCEL)` is false and `create-data-destination.service.spec.ts` covers all three branches), and `ExcelMapper.mapToCreateRequest` returns `{ title, type }` only — no stale credentials can ride along after a type switch.
- `canCreateDestinationInApp` is removed cleanly: no remaining reference in `apps/`, `packages/` or `docs/`, and the barrels re-export with `export *`.
- The "oldest" claim matches the server — `listByProjectIdAndType` has an explicit `order: { createdAt: 'ASC' }`.
- The type select is `disabled={!!isEditMode}`, so dropping the `type === field.value` escape hatch cannot blank the field on an existing destination.
- `apps/docs/src/content/docs` is generated from `docs/` by the sync script, so it correctly stays untouched here.

Four notes on files this PR does not touch, so they could not be left inline:

---

**💡 Suggestion:** the root cause of the `sourceDestinationId` 400 is `DataDestinationConfigSheet.tsx:65`, not the form.

**Why:** when `source` is truthy that branch skips `mapper.mapToCreateRequest` entirely and posts `{ title, type, sourceDestinationId }` whatever `type` says. It also silently drops type-specific fields — a Google Sheets → Email switch with a stale source would discard the typed addresses and return `Cannot copy credentials from GOOGLE_SHEETS to EMAIL destination`. The new effect makes that unreachable today, but only by ordering: anything that sets `selectedSource` without a following type change (a second copy-source picker, a prefilled `initialData`) re-opens it.

**Fix:** gate on the same fact the picker is gated on — `if (source && formFields.type === DataDestinationType.GOOGLE_SHEETS)` — or always start from `mapper.mapToCreateRequest(formFields)` and merge `sourceDestinationId` on top, so copying is an addition rather than a replacement. The backend guard is correct and should stay.

---

**💡 Suggestion:** the Excel tab of `EmptyDataMartDestinationsState.tsx:144` still offers only the Marketplace link.

**Why:** the Google Sheets tab's primary action is `onClick: onOpenCreateDestination`; Excel's is `href: 'https://marketplace.microsoft.com/en-us/product/WA200011946'`. That was right while `canCreateDestinationInApp(EXCEL)` was false. Now a user on this empty state clicks "Connect Microsoft Excel", lands on Microsoft Marketplace, and never learns that **+ New Destination** would also have worked — the discoverability gap this PR set out to close.

**Fix:** either mirror the Google Sheets shape (primary = `onOpenCreateDestination`, secondary = the Marketplace link), or keep the Marketplace link primary and add the create action as the secondary. Installing the add-in is still required either way.

---

**💡 Suggestion:** say in `ExcelDescription.tsx` that the title is what the add-in will show.

**Why:** for Excel the title is the only field, and after `OWOX/owox-data-marts-spreadsheets#62` it is also the only label in the add-in's picker — an Excel destination's `credentials` is always `{}`, so the picker falls back to a creator·date subtitle. Auto-created rows are all `Microsoft Excel`; a hand-created one defaults to `New Destination`, which does not even identify itself as Excel. The copy says "There is nothing to set up here beyond a title" but never says where that title is read.

**Fix:** one sentence — "Name it after the workbook or team that will use it; this is how you tell your Excel destinations apart." A type-aware default title would be a larger change (`defaultValues` is evaluated once) and is probably not worth it.

---

**📘 Docs nit:** `docs/destinations/manage-destinations.md:5` still lists Excel third.

**Why:** the intro reads "**Google Sheets**, **Data Studio**, **Microsoft Excel**, …" while the numbered list at line 27 now reads Google Sheets → Microsoft Excel → Data Studio, matching the reorder in `data-destination-type.model.ts`. Same page, two orders.

**Fix:** swap **Data Studio** and **Microsoft Excel** on line 5.

---

**CI note (not a finding):** the two red checks are unrelated to this branch — `UserProvisioningSettings.test.tsx` is a known timing flake (the test carries its own comment about it) and the `datamart-calculated-field-formula` browser spec fails on `main` as well. The web suite is 269 files passing, 1 failing. Worth a re-run rather than a fix here.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/FormDescriptions/descriptions/ExcelDescription.tsx
romandubovyi: Reworded in 0021d664a. The add-in release ships first, so both the form hint and microsoft-excel.md now describe the picker: "If your project has more than one Excel destination, the add-in asks which one to use when you create a report." The oldest-destination sentence is gone from both.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/DataDestinationForm.tsx
romandubovyi: Fixed in 0021d664a: handleSourceClear now sets credentials.credentialId back to its initial value with setValue(..., { shouldDirty: false }) instead of resetField, so the placeholder is cleared on the service-account tab as well. The type-change effect and the manual clear button share the same path.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/DataDestinationForm.test.tsx
romandubovyi: Tightened in 0021d664a: the type stub now sets the value with shouldDirty: true, the tests click Save instead of dispatching submit, and the credentials assertion that held regardless of the effect is dropped. With the effect deleted, the source assertion fails (checked by mutation). Driving the real DestinationTypeField and asserting at the DataDestinationConfigSheet level is left for a follow-up.

.changeset/6966-excel-destination-manual-creation.md
romandubovyi: Added a paragraph for the copy-source fix in 0021d664a. The oldest-destination caveat is not added: per the thread on ExcelDescription.tsx, that rule is replaced by the add-in picker, so the release note no longer needs it.

apps/web/src/features/data-destination/edit/components/DataDestinationEditForm/DestinationTypeField.test.tsx
romandubovyi: Moved in 0021d664a: the order is now asserted on the rendered options inside the existing render test (getAllByRole('option')), and the separate model-only test with its second name list is removed.

apps/backend/src/data-marts/facades/mcp-destination-type.ts
romandubovyi: Restored the reason in 0021d664a: a person creating an Excel destination in the web app sees the project's existing ones in the list first; an agent creates blind and would add a second, indistinguishable one. Wiring Excel into add_destination is left as a separate change.

romandubovyi:
Thanks for the review. Addressed the four notes in 0021d664a:

- DataDestinationConfigSheet takes the copy-source branch only when the type is Google Sheets, so a stale source can no longer replace the request for another type.
- The Excel tab of EmptyDataMartDestinationsState keeps the Marketplace link as primary and adds "Create destination" as the secondary action.
- ExcelDescription now says the title is how Excel destinations are told apart in the add-in.
- manage-destinations.md intro lists Excel second, matching the list below.

CI: the new commit re-runs both red checks.

PR #1624, opened 2026-09-17
fix(data-marts): save the automatic aggregation as a chosen one (#6971)

zapolsky:
Fibery: [#6971 Automatic aggregation looks like a chosen one, and is saved as one](https://owox.fibery.io/Sprint_Work/Work_Item/Automatic-aggregation-looks-like-a-chosen-one-and-is-saved-as-one-6971)

Extension side: OWOX/owox-data-marts-spreadsheets#63. Follow-up to #1614 (#6948), from review feedback on it.

## Why

#6948 derived the automatic aggregation on every run and deliberately never stored it. The picker drew it as a dimmed ghost: visible, but not a rule — it could not be edited, and the report config stayed empty. That left a third state next to "set" and "not set", and the review asked for it to go away: what the product chooses should be a rule like any other, indistinguishable from one the analyst made.

## What this changes

Opening a report that set no aggregation now fills the rule in for real — ticked on the column, counted on the **Aggregations** button, listed in the panel, changed or removed exactly like a rule the analyst picked. A note names the columns OWOX chose for, so they stay tellable apart, and the rule is saved with the report the next time it is saved.

The fill-in is self-terminating: `resolveAutoCollapse` answers `analyst-aggregated` the moment `aggregationConfig` is non-empty. Removing the rule would otherwise re-trigger it, so a ref guards the effect — remove it once and it stays removed. The write is emitted as `{ isRepair: true }`, so filling the rule in does not by itself dirty the form; a real edit on top of it does.

## The Excel divergence

Review also caught a genuine gap between the agreed #6948 spec and what shipped. An Excel report reads its rows over `/external/http-data/reports/:id.ndjson`, and that endpoint returned every underlying row — so the same report answered one way in Google Sheets and another in Excel.

The endpoint now collapses when the fetch **is** the delivery: the destination pulls its own data *and* that delivery collapses. That is Excel alone.

- A caller pulling somebody's Google Sheets report over HTTP Data stays a third party and keeps every row — `collapsesOnDelivery` alone would have collapsed it, which is why the gate is a composition of the two predicates rather than one of them.
- Looker Studio pulls too, but its connector reads a report the way any ad-hoc caller does, so its rows stay raw.

The Generated SQL preview moves to the same rule, and the report editor predicts for an Excel report in both the web app and the add-in. The changeset and the Report Aggregations page no longer claim Excel is left alone.

## Tests

- `stream-http-data.service.spec.ts` — one `it.each` over the three destinations, so the distinction is asserted rather than resting on a flag. Excel collapses; Google Sheets and Looker Studio do not.
- `get-report-generated-sql.service.spec.ts` — the Looker Studio / Excel pair is split: Looker Studio stays uncollapsed, Excel collapses.
- `auto-collapse-opt-in.spec.ts` — the negative guarantee widens from two allowed callers to three, with the reason recorded in the spec itself.

Backend 9789/9789, `nest build` clean, web `tsc -b` clean.


.changeset/6948-avoid-duplicates-in-reports.md
```diff
@@ -19,8 +19,8 @@ What a report gets depends on what it selected:
...
+Ad-hoc reads are unchanged: **HTTP Data**, the MCP `query_data_mart` tool, `apps/ctl`, the Looker Studio cache-fill query, "copy as Data Mart", and a report's save-time dry run all keep returning exactly what was asked for, duplicates included — including pulling somebody's Google Sheets report over HTTP Data, where the caller is a third party rather than the report's reader. The two pull destinations part ways: a **Microsoft Excel** report collapses like every other report, because the add-in fetches the very rows it writes into the workbook, while a **Looker Studio** report stays uncollapsed, because its connector reads the report the way any other ad-hoc caller does. See [Report Aggregations](../../docs/getting-started/setup-guide/report-aggregations.md) for the full list of cases where a report is deliberately left uncollapsed.
```
laskevych, line 26: 💡 **Suggestion:** worth a line about how an auto-aggregated column is named, now that Excel is in scope.

**Why**
An auto-applied aggregation relabels its output column (`aggregatedColumnLabel`): `sessions` is delivered as `sessions | SUM`. Until now that only happened where a report was already collapsing, so it never reached an Excel workbook; this PR brings Excel in, and the add-in will write the suffixed header into the sheet.

I first read this as a compatibility event worth a warning, but the Excel destination is about a month old, the store-install path landed only a couple of PRs ago, and the in-product link to the add-in is newer still — so there is no meaningful population of workbooks with formulas bound to the old header. This is "how it works" for a young integration rather than "what changed under you", and it does not need to be framed as a break.

**Fix**
Optional, and a docs matter rather than a release-note one: the relabeling paragraph in `report-aggregations.md` currently names only "a downstream Google Sheets formula or Looker Studio binding" as things that stop resolving. Adding Excel there keeps the page complete for the destination this PR opts in.

~~Original comment: blocking, on the grounds that the rename needed a release note under the "compatibility… requires user or operator awareness" rule. Withdrawn — that rule needs an installed base to protect, and this one does not have one yet.~~

apps/web/src/features/data-marts/reports/edit/components/ReportEditForm/ReportEditForm.tsx
```diff
@@ -415,7 +416,7 @@ export const ReportEditForm = forwardRef<HTMLFormElement, ReportEditFormProps>(
                             dataMartId={dataMart.id}
                             dataMartTitle={dataMart.title}
                             storageType={dataMart.storage.type}
-                            collapsesOnDelivery={!isPullDestination}
+                            collapsesOnDelivery={collapsesOnDelivery(destinationType)}
```
laskevych, line 424: 🟡 **Important:** in the web app the filled-in rule has no way to reach the server on its own.

**Why**
`ReportFormActions` disables the primary button in EDIT mode when nothing is dirty: `disabledPrimary = isSubmitting || (mode === EDIT && !(isDirty || triggersDirty || ownersDirty))`. The materialisation is deliberately emitted as `{ isRepair: true }`, which moves the baseline instead of dirtying — so an analyst who merely opens a report sees the rule ticked, counted and listed, with **Save** greyed out. The rule reaches the server only if they happen to change something else.

The extension side of this task solved exactly that: `hasPendingRepair` now includes a pending automatic aggregation, so the next save carries it. The web form has no equivalent, so the two hosts of one feature persist differently — a report filled in from the add-in comes back with a stored rule, the same report filled in from the web app does not. It is also why the attribution note survives one host's save and not the other's.

**Fix**
Either give the web form the same pending-repair flush the extension has, or say in the description that in the web app this is a display-only fill-in until an unrelated edit is made — the changeset's "the choice is saved with the report the next time you save it" currently reads as though pressing Save is enough.

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -350,6 +355,18 @@ export class StreamHttpDataService {
           limitConfig: limit ?? report.limitConfig ?? null,
         };
 
+        // This endpoint is the report's delivery only when the destination pulls: the Excel
+        // add-in fetches the very rows the analyst will read, so they have to match what the
+        // same report returns in Google Sheets. A push destination reads raw here — an HTTP
+        // Data caller asking for a Sheets report is a third party, and its contract stays every
+        // underlying row — and Looker Studio pulls but never collapses.
+        const destinationType = report.dataDestination?.type;
```
laskevych: 🟡 **Important:** an Excel run that collapses records nothing about having collapsed.

**Why**
`RunReportService` keeps the plan and writes it into the run — `autoAppliedAggregations`, `autoAppliedLiftedColumns`, `autoAppliedDistinct`, plus a log line for the skip reasons worth logging — and the run-history UI reads those fields (`data-mart-run-report-definition.dto.ts`).

Here `applyAutoCollapse(basePlan).report` discards `.plan`, and `initialMetadata` is built before `buildPlan` from `report.aggregationConfig ?? undefined`, which is empty — an empty config being the precondition for collapsing at all. So an Excel run shows "no aggregation" in its own history while the query it ran grouped. `executionSqlQuery` is captured, so the SQL contradicts the structured record beside it.

`report-aggregations.md` states "Run history for the report also records which columns were auto-aggregated". That stops being true for Excel with this PR, and the doc change here moves Excel into the collapsing set without touching that sentence.

**Fix**
Keep the plan from `applyAutoCollapse` and fold `autoAppliedAggregations` / `autoAppliedDistinct` into the metadata on the `excelReportRun` path, the way `RunReportService` does.

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
```diff
@@ -350,6 +355,18 @@ export class StreamHttpDataService {
           limitConfig: limit ?? report.limitConfig ?? null,
         };
 
+        // This endpoint is the report's delivery only when the destination pulls: the Excel
```
laskevych: 🟡 **Important:** this file already has the predicate `fetchIsDelivery` re-derives.

**Why**
`asExcelReportRun`, thirty lines up, answers exactly this question and documents it in those words: "The report this read IS the run of, or undefined when the read is just a read… Every other destination — including Looker Studio, which is pull-based too — has a run of its own elsewhere, and reading the same report over HTTP is just a read." Its result is already computed in `streamReport` and is in `buildPlan`'s closure.

So the file now decides "this fetch is the delivery" twice, by two routes, and they can drift — even though the outcomes they drive are joined at the hip: whether the read stamps `lastRunAt`, records a run and bills it, *and* whether it collapses. They already disagree on nullability: `asExcelReportRun` reads `report.dataDestination.type` while this reads `report.dataDestination?.type`, so if a soft-deleted destination can really yield `undefined`, the first throws before this guard is ever reached.

**Fix**
`const readPlan = excelReportRun ? applyAutoCollapse(basePlan).report : basePlan;` — one predicate, and "a read billed as a run is a read that collapses" becomes structural rather than coincidental. If the two-predicate composition is wanted for its explanatory value, express it inside `asExcelReportRun` so there is still one answer.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/RowAggregationIcon.tsx
```diff
@@ -106,7 +108,9 @@ export function RowAggregationIcon({
       fieldType={fieldType}
       displayLabel={displayLabel}
       dataMartName={dataMartName}
-      autoFunctionLabel={showsAuto ? REPORT_AGGREGATE_FUNCTION_LABELS[autoFunction] : undefined}
+      autoFunctionLabel={
```
laskevych, line 101: 🟡 **Important:** the editor popover now contradicts itself.

**Why**
`AggregationEditorPopover` renders `autoFunctionLabel` as "{label} is applied automatically **unless you choose one**". After this change `autoFunction` is only ever set for a column whose rule is live in the config — `autoAggregations` is `autoApplied` intersected with `effectiveOutputConfig.aggregationConfig` — so that sentence now shows with the Sum checkbox already ticked. It was written for the prediction this replaces, and the test that pinned it (`names the automatic choice in the editor it opens, without preselecting it`) was deleted rather than rewritten, so nothing catches it. The same line ships in the extension PR.

Two smaller things follow in this file:

- `showsAuto = !isActive && autoFunction !== undefined` is now unreachable for the same reason, so the `title` attribute and the `opacity-60` branch are dead.
- The `autoFunction` prop doc still says "Drawn dimmed and always visible… Ignored once `activeFunctions` is non-empty".

**Fix**
Reword for what it now means — in the shape of "Sum was applied automatically; change or remove it", matching the panel note's new "Change or remove it below" — keep a test on the new wording, and drop `showsAuto` with its doc. Same edit in both repos.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -1766,6 +1771,53 @@ export function ReportColumnPicker({
...
+          column,
+          function: fn,
+        })),
+      },
+      { isRepair: true, changed: ['aggregationConfig'] }
+    );
+  }, [predictedAggregations, effectiveOutputConfig, onOutputConfigChange]);
+
+  // An automatic choice stops being ours to announce the moment the analyst changes or drops it.
+  const autoAggregations = useMemo(() => {
```
laskevych, line 1814: 🟡 **Important:** removing the filled-in rule makes the editor stop telling the truth about delivery.

**Why**
The ref guard stops the effect re-firing, and this memo filters `autoApplied` by what is still in the config — so once the analyst deletes the rule every surface goes quiet: no tick, no badge count, no panel note, no dot. `predictedAggregations` becomes non-empty again, and the comment above it is right that nothing renders it.

The backend has not changed its mind. `resolveAutoCollapse` sees an empty `aggregationConfig` and collapses on delivery exactly as before, so the rows are still grouped and `sessions` is still delivered as `sessions | SUM`. Before this PR the ghost said so; now nothing does. And on the next open the effect fires again and the rule reappears, so "remove it once and it stays removed" holds only for the current session.

This is the one place where materialising the rule *loses* something the prediction carried.

**Fix**
When the config has been emptied and `predictedAggregations` is non-empty again, keep showing what delivery will do — the existing note is the natural surface. Alternatively state in the description that a report on a collapsing destination has no "no aggregation" state to return to, so the editor need not offer one.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -1752,9 +1752,14 @@ export function ReportColumnPicker({
...
   ]);
 
-  // Read against the raw `value`/`outputConfig` props, not the `effective*` ones: a null `value`
-  // means no explicit projection, which the resolver must read as "nothing to predict yet".
-  const autoAggregations = useMemo(
+  // What the server WOULD apply if this report ran as it stands. Read against the raw
+  // `value`/`outputConfig` props, not the `effective*` ones: a null `value` means no explicit
+  // projection, which the resolver must read as "nothing to predict yet".
+  //
+  // This goes empty the moment anything is materialised below, because the resolver refuses a
```
laskevych, line 1760: 🟡 **Important:** `auto-collapse.ts` is now a writer, and its header still reasons as if it only drew a ghost.

**Why**
That file's contract is written for a prediction: "a missing ghost surprises nobody while a ghost that never materialises does", and the deliberate `calculated-not-liftable` divergence is justified as "the safe direction". Sound while the output only dimmed a sigma. From this PR the same map is written into `aggregationConfig` and saved with the report, so the two resolvers differ in consequence, not just in appearance:

- Where the client is **narrower** than the server (the lift case), nothing is written and the server still collapses at run time — fine, and worth stating.
- Where the client is or becomes **wider**, it writes a rule the server would never have chosen, and that rule is now stored. The two are not structurally identical: the server checks `aggregatesByItself` and `filtersIntoHaving` against the full schema and walks `collectSchemaFieldPathDescriptors`; the client checks `!isRowLevelCalculatedField` over `flattenNativeFields`.

**Fix**
Update that header for the new stakes and say which direction of divergence is still acceptable and why. No behaviour change needed for the cases known today — the point is that the bar for the mirror's fidelity just moved.

laskevych (commented):
Reviewed the backend, web and docs changes together with the extension PR, since the two halves of this behaviour only make sense read as a pair.

The shape of the change is right: a prediction rendered as its own third state was the thing to remove, and keying the HTTP Data path on "is this fetch the delivery" rather than "does this destination pull" is the correct distinction.

**Edit:** this review originally flagged the Excel column rename (`sessions` → `sessions | SUM`) as blocking for want of a release note. Withdrawn — the Excel destination is about a month old and the store-install path newer still, so there is no installed base of workbooks the rename can break, and the changeset rule it leaned on exists to protect one. That comment is now a docs suggestion, and nothing here blocks: approving separately.

One item does still need fixing, but it is in the other repo: `OWOX/owox-data-marts-spreadsheets#63` never clears `autoAggregations`, so the card latches dirty and a later rename writes a deleted rule back to the server. That code is shared with the Google Sheets host, so it is not an Excel-only concern. Commented there.

**Worth a look, none of it blocking**

- An Excel run that collapses records no `autoAppliedAggregations`, so its run history contradicts the SQL stored beside it.
- In the web app the filled-in rule cannot be saved on its own (Save is gated on `isDirty`), while the extension persists it via `hasPendingRepair` — the same feature persists differently per host. Applies to Google Sheets and Email reports too.
- The aggregation editor says "applied automatically unless you choose one" next to an already-ticked checkbox.

Everything else is docs, naming and test shape.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -1766,6 +1771,53 @@ export function ReportColumnPicker({
     [collapsesOnDelivery, schema, value, outputConfig]
   );
 
+  // What we actually wrote into the config on the analyst's behalf, kept for this editing session
+  // so the panel note and the button dot can still say so. The config itself cannot: a rule we
+  // added and a rule the analyst typed are the same three fields.
+  const [autoApplied, setAutoApplied] =
```
laskevych, line 1779: 💡 **Suggestion:** the attribution note does not survive the save — worth saying so.

**Why**
`autoApplied` is seeded only by the effect, and the effect only fires while `predictedAggregations` is non-empty, which requires `aggregationConfig` to be empty. Once the rule is stored, the next open predicts nothing, this stays empty, and the note, the dot and the "Automatic aggregations applied for fields: …" announcement are all gone. The rule then reads as one the analyst wrote, everywhere.

That may be the intended end state — the whole point is that it becomes an ordinary rule — but the changeset promises more: "A note names the columns OWOX chose for, **so you can tell them apart from your own**". After the first save you cannot, and the same holds cross-host once the add-in persists it.

**Fix**
Either trim that clause to what holds (the note explains the fill-in at the moment it happens), or persist the attribution if telling them apart is meant to last.

apps/web/src/features/data-destination/shared/enums/data-destination-type.enum.ts
```diff
@@ -36,6 +36,18 @@ export function pullBasedRunHint(type: DataDestinationType): string | null {
...
 
+/**
+ * Whether OWOX collapses this destination's delivered rows when the analyst set no aggregation.
+ *
+ * Mirrors the backend predicate of the same name, and is NOT the inverse of pull-based: an Excel
+ * report collapses exactly like a Google Sheets one, because the add-in is its reader rather than
+ * a third party. Looker Studio is the one exception — its connector reads a report the way any
+ * ad-hoc caller does, so those rows stay raw.
+ */
+export function collapsesOnDelivery(type: DataDestinationType): boolean {
```
laskevych, line 47: 💡 **Suggestion:** the prop doc that mirrors this predicate still describes the behaviour it replaces.

**Why**
`ReportColumnPicker.tsx:151` still reads: "False for a pull-based destination (Looker Studio, Excel): its consumer reads the uncollapsed projection, so the ghost below would predict something that never happens." Both halves are now wrong — Excel collapses, and there is no ghost. That prop is the contract every form reads before deciding what to pass, and `ReportEditForm` now passes `collapsesOnDelivery(destinationType)` precisely because the sentence no longer holds.

This function and its backend twin both got careful new doc comments; the prop callers actually read was left behind.

**Fix**
Rewrite the prop doc against the new predicate. The default and the "a form that forgets loses a prediction" reasoning still stand — only the Excel clause and the word "ghost" need replacing.

docs/getting-started/setup-guide/report-aggregations.md
```diff
@@ -109,7 +109,7 @@ An auto-applied aggregation relabels its column exactly like a manually chosen o
...
+> This applies wherever the report is **delivered** — a scheduled or manual Google Sheets, email, Slack, Teams, or Google Chat run, an MCP "run report" call against a saved report, and the **Microsoft Excel** add-in's own fetch, which is how an Excel report reaches its reader. It does **not** apply to ad-hoc queries: **HTTP Data**, the MCP `query_data_mart` tool, `apps/ctl`, the Looker Studio cache-fill query, "copy as Data Mart", and the report's save-time dry run all keep returning exactly what you asked for, duplicates included — including pulling a Google Sheets report over HTTP Data, where the caller is a third party rather than the report's reader. **Looker Studio** is the one destination left out entirely, because its connector reads the report over exactly those ad-hoc paths; see the exclusion list below.
```
laskevych, line 114: 💡 **Suggestion:** the paragraph above this note still describes the editor as only predicting.

**Why**
The two paragraphs about *where* the collapse applies were updated, but the one above still reads as it did before: "The report editor marks every column OWOX aggregated on your behalf with a small dot on the **Aggregations** button… it is a one-time nudge to look, not a setting." A reader now sees a ticked checkbox, a badge counting 1 and a rule in the panel — none of which this page mentions — and "not a setting" says the opposite of what now happens.

This page is where the changeset sends people for the full behaviour, so it is where an analyst goes to find out whether the rule in front of them is theirs.

**Fix**
Add the fill-in to this section: the rule is written into the report on open, is editable and removable like any other, is saved with the report on the next save, and carries a note naming the columns OWOX chose for while it is fresh.

apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts
```diff
@@ -51,6 +51,22 @@ export function isPullBasedDataDestinationType(type: DataDestinationType): boole
...
+ * NOT the inverse of pull-based. Excel pulls, yet an Excel report still has to return what the
+ * same report returns in Google Sheets — the add-in is its reader, not a third party. Looker
+ * Studio is the one exception: the connector reads a report the way any ad-hoc caller does, so
+ * those rows stay raw.
+ *
+ * Answers what delivery does, never who is asking. A caller pulling a Google Sheets report over
+ * the HTTP data endpoint is a third party, which is why that read pairs this with
+ * `isPullBasedDataDestinationType`.
+ */
+export function collapsesOnDelivery(type: DataDestinationType): boolean {
```
laskevych, line 66: 💡 **Suggestion:** consider an allowlist here, as its neighbour is.

**Why**
`isPullBasedDataDestinationType` names the two types it covers; this is `type !== LOOKER_STUDIO`. Side by side in one file they fail in opposite directions when a destination is added: the first leaves a new type out until someone thinks about it, the second opts it in silently. Opting in is the side that changes delivered column names and row counts, so the silent direction is the costly one.

The doc comment above is excellent about *why* Looker Studio is the exception — which is exactly the reasoning a new type added by someone who never opens this file would lose.

**Fix**
If a `switch` over `DataDestinationType` with no default is viable, it forces the decision at compile time for every type added later, and the existing comment becomes the reason attached to the Looker Studio arm. Otherwise this is fine as it stands — flagging the asymmetry, not insisting on it.

apps/backend/src/data-marts/services/auto-collapse-opt-in.spec.ts
```diff
@@ -16,8 +20,9 @@ import { join } from 'node:path';
 const SRC_ROOT = join(__dirname, '..', '..');
 
 const ALLOWED_CALLERS = [
-  'data-marts/use-cases/run-report.service.ts',
   'data-marts/use-cases/get-report-generated-sql.service.ts',
+  'data-marts/use-cases/run-report.service.ts',
+  'data-marts/use-cases/stream-http-data.service.ts',
```
laskevych, line 25: 💡 **Suggestion:** the guard is now weaker than its own new comment admits.

**Why**
The allowlist is per-file, and `stream-http-data.service.ts` hosts both paths this feature distinguishes: `streamReport`, which may collapse, and `streamDataMart`, the ad-hoc query, which must never. Allowlisting the file means a future `applyAutoCollapse(` added to the ad-hoc `buildPlan` — the exact mistake this spec exists to catch, and the one the new comment calls "the subtle one" — passes unnoticed.

The new `it.each` covers the three destinations on the report path, but nothing behaviourally pins the ad-hoc Data Mart stream as uncollapsed.

**Fix**
Add one behavioural case beside that `it.each`: a `streamDataMart` read whose plan would collapse, asserting `aggregationConfig` stays empty. That restores the guarantee the file-level allowlist can no longer make on its own.

apps/backend/src/data-marts/use-cases/stream-http-data.service.spec.ts
```diff
@@ -1324,6 +1324,46 @@ describe('StreamHttpDataService', () => {
       expect(usesSuffixedJoinedFieldNames(readPlan)).toBe(false);
     });
 
+    // One endpoint, two readers with opposite contracts. The rows must differ, and only by the
+    // report's destination — not by who asked.
+    it.each([
+      [DataDestinationType.EXCEL, true, 'the add-in fetch IS the delivery'],
+      [DataDestinationType.GOOGLE_SHEETS, undefined, 'an HTTP Data caller is a third party'],
+      [DataDestinationType.LOOKER_STUDIO, undefined, 'the connector reads raw rows'],
```
laskevych: 💅 **Nit:** the `it.each` title asserts the opposite of the case for two of the three rows.

**Why**
`'collapses for %s: %s — %s'` prints `collapses for GOOGLE_SHEETS: undefined — an HTTP Data caller is a third party`, and the same for Looker Studio. In a CI log or a `--verbose` run, two of the three names read as claiming a collapse the case exists to disprove — on the one distinction this PR is about. The assertion is also doing a small dance to compare against `undefined`: `expect(plan.aggregationConfig?.length ? true : undefined).toBe(expected)`.

**Fix**
Carry the expected verb in the row instead of the boolean — `['EXCEL', 'collapses', …]` / `['GOOGLE_SHEETS', 'keeps every row', …]` — and assert on `aggregationConfig` length directly. The titles then read as three statements rather than one statement contradicted twice.

laskevych (approved):
Approving.

No blocking findings on this PR. The earlier review's one blocker — a release note for the Excel column rename — is withdrawn: the destination is too young to have an installed base that the rename could surprise, so the compatibility rule it rested on does not apply. It now stands as a docs suggestion on `report-aggregations.md`.

The comments left on the earlier review are all follow-up material. If any of them are worth doing in this round, my order would be the editor copy (one line, visible to users on every destination), then the web-side save gap, then the Excel run-history record.

The one thing I would not merge without is in the extension PR rather than this one — `OWOX/owox-data-marts-spreadsheets#63` needs `autoAggregations` cleared alongside `uniqueCountRepair`, since that hook is shared with the Google Sheets host.

.changeset/6948-avoid-duplicates-in-reports.md
zapolsky: Added, in the relabeling paragraph itself: it now names "a Google Sheets formula, a Looker Studio binding, or a formula in the Excel workbook the add-in refreshes" rather than the first two only. Agreed on the release-note question — the rule protects an installed base, and this destination does not have one yet.

apps/web/src/features/data-marts/reports/edit/components/ReportEditForm/ReportEditForm.tsx
zapolsky: Fixed rather than documented — the two hosts persisting differently was the part that could not stand.

`ReportFormActions` takes `hasPendingRepair`, folded into the same `hasSomethingToSave` that drives both the button and the Save & Run label. `usePendingAutoAggregation` holds the flag in `ReportEditForm` and `EmailReportEditForm`, set when a repair reports `aggregationConfig` among its changed keys and cleared once the report has stored it. Deliberately absent from `onDirtyChange`: the analyst has no unsaved edits, so the leave guard stays quiet.

Two tests on the actions, and the changeset now says pressing Save is enough.

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
zapolsky: Confirmed and fixed. `baseMetadata` does get rebuilt from the collapsed `readPlan`, so the HTTP metadata was right — but `reportDefinition` is built from the stored report, and that is what run history reads, so the structured record said no aggregation.

The plan now travels with the plan: `ExecuteStreamPlan`'s report arm carries `autoApplied`, `StreamPlanContext` hands it to both recorders (a read that fails after the plan was built still says what it was going to deliver), and `recordHttpDataRun` folds it into the snapshot — creating the `outputConfig` container when the stored report has none, which is the normal case here.

Mapping it is now one function, `autoAppliedOutputConfig` in the resolver, and `RunReportService` uses it too — the fields are easy to fill in from only one path, which is how they came to be filled in from only one. Two tests: EXCEL records `autoAppliedAggregations`, GOOGLE_SHEETS records nothing.

apps/backend/src/data-marts/use-cases/stream-http-data.service.ts
zapolsky: Taken — it is now `const collapsed = excelReportRun ? applyAutoCollapse(basePlan) : undefined;`, and the two destination predicates are no longer imported here at all.

I did not move the composition into `asExcelReportRun`, though. That function also decides whether the read stamps `lastRunAt`, records a ReportRun and bills an Excel unit; expressed as `isPullBased && collapsesOnDelivery` it would silently adopt the next pull-based destination into all of that. `=== EXCEL` is the safe spelling for a function named after Excel, so the reasoning lives in the comment at the call site instead.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/RowAggregationIcon.tsx
zapolsky: All three, in both repos. The note reads "Sum was applied automatically — change or remove it", matching the panel's "Change or remove it below"; `showsAuto`, the `title` and the `opacity-60` branch are gone; and both prop docs are rewritten. Pinned by a test in `RowAggregationIcon.test.tsx` on each side — the web one asserts the wording with the Sum box already checked, and the extension's ghost test is rewritten to assert the icon is now indistinguishable from any set rule.

Unrelated find while writing that test, reported rather than fixed: opening this editor from a field row also toggles the column's checkbox. The row is a `<label>`, and the label's activation behaviour fires the Radix checkbox regardless of `stopPropagation` or `preventDefault` on the button — `FieldInfoTooltip` beside it carries a `stopParentToggle` for what looks like the same reason, so the sigma and filter icons may simply be missing it. In the picker's controlled tests the deselection prunes the aggregation immediately. It predates this PR and I could not reproduce it in a real browser here, so I left it alone — worth its own ticket if it reproduces.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
zapolsky: Right, and this was the one that deserved behaviour rather than prose. The panel now carries a second note, fed by `predictedAggregations` — non-empty in exactly one visible state, which is the one you describe: "This report sets no aggregation, so delivery will apply: sessions — Sum. Add one of your own to decide it yourself." Same in the extension.

So the editor never goes quiet about a report that collapses: before the deletion it says who chose, after it says what delivery will do. Test asserts the note after removing the rule.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
zapolsky: Rewritten on both sides, around the asymmetry rather than the old "safe direction" line: narrower than the server writes nothing and the server still collapses at run time; wider writes a rule the server would never have chosen into a report that then stores it. It also now names why they are not structurally identical — governance against the full schema plus `collectSchemaFieldPathDescriptors` on one side, `!isRowLevelCalculatedField` over `flattenNativeFields` on the other — and says a new case belongs on the narrow side or on both, never on the client alone. No behaviour change, as you suggested.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
zapolsky: Trimmed to what holds. The changeset now says the note names the columns "while the choice is fresh; once the report is saved the rule is simply yours, like any other" — which is the intended end state, and the clause promising you could always tell them apart is gone.

apps/web/src/features/data-destination/shared/enums/data-destination-type.enum.ts
zapolsky: Rewritten against the new predicate, and the stakes moved with it: the prop now decides whether a rule is WRITTEN, not whether a ghost is drawn. The default argument changed shape too — a form that forgets shows the report as it is stored, which is at worst incomplete, while the opposite default would write a rule into a report nothing ever collapses. Extension prop doc updated the same way.

docs/getting-started/setup-guide/report-aggregations.md
zapolsky: Added. That section now leads with the fill-in — written on open, editable and removable like any other rule, saved on the next save, with the note while it is fresh and the delivery note after a removal — and the dot paragraph follows as what it is. The run-history sentence now says "including an Excel report refreshed from the add-in", which this PR made true.

apps/backend/src/data-marts/data-destination-types/enums/data-destination-type.enum.ts
zapolsky: Done — a `switch` with no default, so a new destination fails to compile until someone answers here. Mirrored in the web enum for the same reason. Your framing is why: the silent direction is the one that changes delivered column names and row counts.

apps/backend/src/data-marts/services/auto-collapse-opt-in.spec.ts
zapolsky: Added beside the report cases: an ad-hoc `stream` read of a projection that would collapse — one dimension, one metric, nothing aggregated — asserting the plan reaches the blending decision with an empty `aggregationConfig` and no `distinct`. The allowlist keeps its file-level promise; this covers the half of the file the allowlist can no longer speak for.

apps/backend/src/data-marts/use-cases/stream-http-data.service.spec.ts
zapolsky: Renamed so each row states its own outcome: `EXCEL collapses — the add-in fetch IS the delivery`, `GOOGLE_SHEETS keeps every row — an HTTP Data caller is a third party`, `LOOKER_STUDIO keeps every row — the connector reads raw rows`. The assertion compares `aggregationConfig?.length ?? 0` against a rule count, so the `undefined` dance is gone too. The shared setup moved into a `collapsibleReportRead` helper, which the two new run-record tests reuse.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/RowAggregationIcon.tsx
zapolsky: Correction to what I wrote above: I checked it in Chrome, and there is no product bug.

A minimal page — `<label><button role="checkbox"><button>Σ</button></label>` — does not toggle the checkbox when the inner button is clicked, for a native `input` or a Radix-style `button[role=checkbox]` alike. The HTML spec exempts interactive content descendants from a label's activation behaviour; jsdom does not implement that exemption, which is where my earlier reading came from. No ticket needed.

The tests were the real casualty, and in a way that mattered: both of the ones about removing the rule reached that state by clicking the row's sigma, which in jsdom deselected the column and pruned the rule as a side effect. The delivery-note test would have passed even if unticking the function did nothing at all. They now remove the rule through the Aggregations panel, with the artefact written down beside the helper so the next person does not spend the same hour on it.

PR #1626, opened 2026-09-18
fix(ci): wait for npm to actually serve a published version

zapolsky:
## Problem

The snapshot job has been failing on `Build and push "next" container image` with
`ETARGET No matching version found for @owox/backend@<version>`, minutes after
`changeset publish` reported that same version published successfully.

Five of eight recent pushes to `main` failed this way, plus three manual re-runs.

## Why the existing gate could not catch it

The "Check if npm package with this version exists" steps were meant to prevent
exactly this, but:

1. **They could not fail.** Each per-package loop fell through silently once all
   30 attempts were spent. Only the final `owox` loop ended in `exit 1`. The log
   read `Waiting for @owox/backend ... (30/30)` and the step still went green.
2. **They checked the wrong document.** `npm view` reads the full packument;
   `npm install` reads the abbreviated one. Those are separate CDN objects with
   separate expiry, so a hit did not prove the installer could resolve the version.
3. **The window was too small.** Measured lag between a successful publish and
   `@owox/backend` becoming resolvable: ~8 min, ~8 min, ~16 min. The window was
   5 min. Siblings of the same snapshot were ready in under a second.

## Change

One script (`tools/wait-for-npm-version.mjs`) replaces twenty near-identical
shell loops. It resolves through the abbreviated packument, fails the job when a
package never appears, and gets a 25 minute budget.

`publish.yml` goes from 321 lines to 151. The package list and its order are
unchanged: ten packages, `owox` last.

## Verification

Run locally against the live registry:

| case | result |
|---|---|
| all 10 packages at a real published version | exit 0, ~1s each |
| version that does not exist | exit 1, single-line error, stops at the first package |
| no arguments | exit 1 + usage |
| `npm run format:root:check` | clean |
| `npm run lint:root` | clean |

The workflow itself only triggers on push to `main`, so this cannot be exercised
on the PR branch.

## This is a painkiller, not a cure

The lag tracks packument size. `@owox/backend` carries over 1300 snapshot
versions in a 7.4 MB document, and every push to `main` adds another. The other
ten packages are unaffected because their documents are small.

25 minutes buys a week or two. The real fix is shrinking the document — not
snapshot-publishing every package on every merge, and pruning old `-next-`
versions — which needs its own discussion.

No changeset: `docs/contributing/repository/release-strategy.md` excludes
"CI, local developer tooling, or agent-instruction changes only".


.github/workflows/publish.yml
```diff
@@ -214,99 +130,15 @@ jobs:
...
-            fi
-            echo "Waiting for owox@${{ steps.extract_next_version.outputs.version }} to appear on npm ($i/30)..."
-            sleep 10
-          done
-
-          exit 1
+          node tools/wait-for-npm-version.mjs "${{ steps.extract_next_version.outputs.version }}" \
+            @owox/web @owox/backend @owox/internal-helpers @owox/idp-protocol \
+            @owox/idp-better-auth @owox/idp-owox-better-auth @owox/api-client \
+            @owox/plugin-sdk @owox/ctl owox
```
laskevych, line 141: 🟡 Important (non-blocking): Add `@owox/connectors` to both package lists (here and in the "latest" step above).

**Why:** It is public, sits in the same changesets `fixed` group, and `@owox/backend` depends on it, so `npm install -g owox@<version>` has to resolve it too. Its packument is in the same size class as the slow one: 1366 versions and 1.3 MB abbreviated, against 1355 versions for `@owox/backend`. If it lags, every gated package passes and the image build still fails with `ETARGET`.

**Fix:** Put `@owox/connectors` before `owox` in both steps. I ran the script with it added against `0.34.0` and the current `next` snapshot — it resolves, so this cannot turn the job red.

tools/wait-for-npm-version.mjs
```diff
@@ -0,0 +1,105 @@
...
+ *   CDN object with its own expiry, so a hit there does not prove the installer
+ *   will see the version;
+ * - a package that never appears fails the process. Anything less lets the
+ *   build run against a version npm cannot resolve.
+ *
+ * Usage: node tools/wait-for-npm-version.mjs <version> <package...>
+ */
+
+const REGISTRY = process.env.NPM_REGISTRY ?? 'https://registry.npmjs.org';
+const TIMEOUT_SECONDS = Number(process.env.WAIT_TIMEOUT_SECONDS ?? 1500);
```
laskevych, line 27: 💡 Suggestion: Reject a non-numeric `WAIT_TIMEOUT_SECONDS` / `WAIT_INTERVAL_SECONDS`.

**Why:** `Number('abc')` is `NaN`, and `Date.now() >= NaN` is never true, so the loop never times out. Reproduced with `WAIT_TIMEOUT_SECONDS=abc` — it polls forever. A typo would turn the gate back into a step that cannot fail, until the 6 h job limit.

**Fix:** Check `Number.isFinite(...) && > 0` after parsing and exit 1 with a clear message.

tools/wait-for-npm-version.mjs
```diff
@@ -0,0 +1,105 @@
...
+ * Resolves the version through the document `npm install` reads, so a positive
+ * answer here means the installer can see it too.
+ *
+ * @param {string} name npm package name, scoped or not
+ * @param {string} wanted exact version to look for
+ * @returns {Promise<boolean>} whether the registry currently serves that version
+ */
+async function isPublished(name, wanted) {
+  try {
+    const response = await fetch(`${REGISTRY}/${name.replace('/', '%2F')}`, {
```
laskevych, line 51: 💡 Suggestion: Give each registry request its own timeout.

**Why:** Without a signal, a stalled connection holds one attempt for Node's default 300 s, and the deadline is only checked between attempts. A few stalls push the real wait well past the budget.

**Fix:** `signal: AbortSignal.timeout(30_000)` in the `fetch` options — the existing `catch` already treats it as "not there yet".

tools/wait-for-npm-version.mjs
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+/**
+ * @param {string} name npm package name to wait for
+ * @param {string} wanted exact version to wait for
+ * @returns {Promise<void>} resolves once published, rejects on timeout
+ */
+async function waitFor(name, wanted) {
+  const deadline = Date.now() + TIMEOUT_SECONDS * 1000;
```
laskevych, line 71: 💡 Suggestion: Make the 25 minutes a budget for the whole step, or say it is per package.

**Why:** `deadline` is recomputed for every package, so the worst case is 10 × 1500 s, while the workflow comment and the PR text read as one 25 minute window. The job has no `timeout-minutes` to cap it.

**Fix:** Compute the deadline once before the loop and pass it into `waitFor`. If per-package is intended, adjust the comment and add a job-level `timeout-minutes`.

.github/workflows/publish.yml
```diff
@@ -214,99 +130,15 @@ jobs:
           cat $GITHUB_OUTPUT
 
       - name: Check if npm package with this "next" version exists
```
laskevych, line 132: 💡 Suggestion (follow-up is fine): Consider a `concurrency` group for this workflow.

**Why:** With waits of up to 25 minutes, overlapping runs for consecutive merges become the normal case. The run that finishes last wins the floating `:next` image tag, so it can end up on an older snapshot than npm's `next` dist-tag.

**Fix:** `concurrency: { group: snapshot, cancel-in-progress: false }`.

laskevych (approved):
✅ Approving. I ran the script against the live registry: all ten packages resolve for both the current `latest` and the latest `next` snapshot, a missing version fails with exit 1, and the package list and order are unchanged. The comments are non-blocking. The one I'd take before merge is adding `@owox/connectors` to the lists, since it is the remaining way to hit the same `ETARGET`.

zapolsky:
Superseded by Fibery 6977 — the image will be built from locally packed tarballs instead of waiting for the npm snapshot, so the wait gate this PR enlarges is being removed rather than tuned.

Keeping the measurements here for the record: lag between a successful `changeset publish` and `@owox/backend` becoming resolvable was ~8 min, ~8 min, ~16 min and ~9m20s across four runs, against a 5 min window, while the other ten packages of the same snapshot were installable within seconds.

PR #1627, opened 2026-09-18
feat(ci): build the container image from packed tarballs, stop publishing snapshots to npm

zapolsky:
Closes [6977](https://owox.fibery.io/Sprint_Work/Work_Item/6977).

## Problem

The snapshot job kept failing on `Build and push "next" container image` with
`ETARGET No matching version found for @owox/backend@<version>`, minutes after
`changeset publish` reported that same version published successfully.

Measured lag between a successful publish and `@owox/backend` becoming
resolvable, across four runs: **~8 min, ~8 min, ~16 min, ~9m20s** — against a
gate that waited 5. The other ten packages of the same snapshot were installable
within seconds. The delay tracks the size of the registry metadata, and
`@owox/backend` had reached 1355 versions in a 7.4 MB document because every
merge to `main` added another snapshot.

Deleting old versions is not available: npm's
[unpublish policy](https://docs.npmjs.com/policies/unpublish) requires that no
package in the public registry depends on the version, and every
`owox@<snapshot>` pins `@owox/backend@<same snapshot>` exactly, so the condition
fails for all 1314 of them. Deprecation leaves them in the metadata.

## Change

- **The image installs tarballs packed in the same job**, so the registry is off
  the critical path entirely.
- **Snapshots are no longer published to npm.** They are still versioned — the
  image needs a version to carry and tag — and `ghcr.io` keeps emitting the same
  `0.x.0-next-<timestamp>` tags the `k8s` overlays pin.
- **Releases are unchanged.** `latest` still goes to npm and still gets an image.
  Its build uses the same packed tarballs, so the snapshot path exercises that
  code on every merge rather than first on a release day.

`publish.yml` drops from 321 to ~150 lines: both npm wait gates are gone, and
with them a bug where each per-package loop fell through silently once its
attempts were spent, so the log read `Waiting for @owox/backend ... (30/30)`
while the step went green.

## Image parity

Installed as **one project rather than eight global packages**. A global install
gives every argument its own tree, which stops npm deduplicating between them:
that cost ~240 MB and ~950 duplicated installs when it was tried, with
`@owox/backend` carrying a 929 MB nested `node_modules` of its own. `overrides`
pointing each internal dependency at its tarball keeps a single hoisted root.

Verified against an image built the old way from npm at the same version:

| | from npm (old) | from tarballs (new) |
|---|---|---|
| size | 1.31 GB | 1.30 GB |
| packages installed | 1033 | 1040 |
| `package.json` files in image | 1498 | 1499 |

## Review fixes (second commit)

Four agents reviewed the first commit. Two blockers, both introduced by it:

- **`WORKDIR /opt/owox` moved the container's cwd** from `/`. `.env`,
  `owox serve -e <relative>` and a relative `SQLITE_DB_PATH` all resolve against
  cwd, so an operator mounting `/.env` would have booted on defaults — silently,
  and with `IDP_PROVIDER` falling back to `none`. The install stays under
  `/opt/owox`; the process starts at `/` again. Managed staging/prod were never
  exposed (both overlays configure everything as env vars via ConfigMap).
- **The `latest` image was gated only on its tag being absent from GHCR.** That
  was self-correcting while the content came from the registry; now the content
  is the checkout, so an ordinary merge — or one transient
  `docker manifest inspect` failure — could publish unreleased code as `:latest`.
  It now also requires that this run actually published, with
  `workflow_dispatch` to rebuild a genuinely missing release image. A
  `concurrency` group stops two runs racing to push the same tag; this workflow
  was the only one of twelve without one.

Hardening, all of which turns a silent wrong result into a loud failure:

- the pack script refuses an output directory outside the repository **before it
  does anything else** — it calls `rm -rf` with `force`, and `.` resolves to the
  whole tree;
- it walks `peerDependencies` and rejects a publishable dependency that escapes
  the pack set (the one way `ETARGET` could return);
- it asserts each workspace's `dist` exists rather than trusting a `prepack` that
  five of the eight do not have — `npm pack` on a missing `dist` warns and exits
  0, which would ship an image with no UI;
- it rejects a version that disagrees with the tag, and the Dockerfile re-checks
  the installed version against `$version` and runs `owox --version`, so a
  dangling symlink or a stale `docker-packages/` cannot be pushed;
- publishability reads `private !== true` — the test npm and changesets apply —
  instead of a changeset `fixed` group, which means version linkage and which
  the sibling version tool reads differently (`forEach` over all groups vs
  `fixed[0]`). Both give the same eleven packages today.

## Verification

Run locally, end to end:

- built an image at `0.35.0-next-20260918093216`, a version that returns **404**
  from npm — the exact scenario that was failing. `docker run ... --version`
  prints it back, and `--help` lists the real commands;
- confirmed the cwd fix: `docker run --entrypoint sh ... -c pwd` → `/`;
- confirmed the version guard fires: building with a deliberately wrong
  `--build-arg version` fails with
  `tarballs ship owox@0.34.0 but this image is tagged 9.9.9-next-fake`;
- confirmed the path guard rejects `.`, `..`, `/tmp` and `../../`;
- the `dist` assertion caught a genuinely missing `@owox/backend/dist` during
  testing, which is exactly the case it exists for;
- Prettier, ESLint and Markdownlint clean.

`publish.yml` only triggers on push to `main`, so none of this can be exercised
on the PR branch — it is first exercised at merge.

## Notes for review

- **Tag shape is load-bearing and unchanged.** Both
  `apps/owox-data-marts/overlays/{staging,production}` in `OWOX/k8s` pin an exact
  `0.x.0-next-<timestamp>` tag, and `update-owox-data-marts-version.yml` resolves
  from GHCR. `step-preview-npm-versions.js` there only logs npm versions via
  `core.info` and does not gate, so it will simply stop showing snapshots —
  worth deleting or relabelling in `k8s` so it does not mislead during an
  incident.
- **`.dockerignore` is new and was overdue.** `main` has none at all, so every
  build to date shipped the whole repo as build context — `.git`, `node_modules`,
  any local `.env`.
- **Not fixed here, deliberately:** published `@owox/web` declares
  `"@owox/ui": "../packages/ui"`, which npm resolves to a symlink pointing at an
  empty directory. Harmless today (`@owox/web` ships a built `dist/index.html`
  and nothing imports it at runtime) and identical to the old install path, but
  it is a landmine. `@owox/ui` belongs in `devDependencies`.

## Follow-ups

- `npm dist-tag rm owox next` (and the other ten packages). Until then
  `owox@next` keeps silently installing the frozen 17 Sep build instead of
  failing loudly.
- Carrying the root `overrides` (21 security pins) into `image-package.json`.
  Pre-existing — they never reached the image — but newly one line. Wants a
  decision, because it widens the divergence from `npm install -g owox`; the
  real fix is the shrinkwrap TODO in `publish.yml`, which this change unblocks.
- Changeset heading style: the pre-commit hook lints `.changeset/*.md` (it passes
  explicit paths, and the config only applies its exclusions when there are none),
  pushing everyone to `# Heading` — which renders as a literal `#` mid-line in
  `CHANGELOG.md`. Repo-wide, seven of eight pending changesets are affected.


.github/workflows/publish.yml
```diff
@@ -2,6 +2,18 @@ name: Snapshot
 on:
   push:
     branches: [main]
+  workflow_dispatch:
```
laskevych, line 5: 🟡 Important (non-blocking, but cheap to take before merge): Restrict the job to `main`.

**Why:** Until now this workflow could only run on push to `main`. With `workflow_dispatch` it can be started from any branch in the Run workflow dropdown. That run does `changeset publish` to npm (irreversible) and pushes a `:<x>-next-<timestamp>` + `:next` image built from unreviewed code. The `k8s` version-update workflow auto-selects the newest GHCR tag when no version is given, so that image becomes the default deploy target.

**Fix:** `if: github.ref == 'refs/heads/main'` on the `publish` job.

.github/workflows/publish.yml
```diff
@@ -64,8 +76,16 @@ jobs:
       #   run: node tools/generate-shrinkwrap.mjs
 
       - name: Publish to "latest"
+        id: publish_latest
         run: |
-          npx changeset publish
+          set -o pipefail
+          npx changeset publish | tee /tmp/changeset-publish.log
+          if grep -q "New tag:" /tmp/changeset-publish.log; then
```
laskevych, line 88: 🟡 Important (non-blocking): Make a skipped release image loud.

**Why:** The `latest` image now requires `published == 'true'` from this very run. If `changeset publish` succeeds and the multi-arch build then flakes, "Re-run failed jobs" finds nothing to publish, skips both `latest` steps and goes green. The release is on npm, `:latest` never appears, and the log does not say so.

**Fix:** Add a step for `exists == 'false'` && not published && not `rebuild_latest_image` that emits `::warning::` (or fails) and names the dispatch input to use.

.github/workflows/publish.yml
```diff
@@ -2,6 +2,18 @@ name: Snapshot
 on:
   push:
     branches: [main]
+  workflow_dispatch:
+    inputs:
+      rebuild_latest_image:
```
laskevych, line 7: 💡 Suggestion: Say in the input description that the rebuild packs the dispatched ref's HEAD.

**Why:** The repo has no git tags, so the release commit cannot be targeted. If `main` has moved on since the release, `package.json` still carries the release version, the Dockerfile version check passes, and `:<version>` + `:latest` ship unreleased code. That is the case the `published` gate was added to prevent.

**Fix:** Note it in the description ("run before anything else merges, or from a branch cut at the release commit"). A follow-up could push the tags `changeset publish` already creates locally.

.changeset/6977-pack-docker-image-without-npm-snapshot.md
```diff
@@ -0,0 +1,13 @@
+---
+'owox': minor
+---
+
+# Snapshot builds are now distributed as container images only
+
+`owox@next` is no longer published to npm. To run a pre-release build, pull
```
laskevych, line 7: ❓ Question: Is losing the snapshot channel for `@owox/plugin-sdk`, `@owox/ctl` and `@owox/api-client` intended?

**Why:** The changeset and the docs only mention `owox@next`, but snapshots stop for every package. These three are not in the image at all, so the container gives no replacement. Already-published snapshot ranges (plugins pin `@owox/plugin-sdk@^0.x.0-next-…`) keep resolving. New SDK or ctl changes will only reach plugin authors with a release.

**Fix:** If intended, add one sentence to the changeset and `release-strategy.md`. If not, keep snapshot-publishing just the small packages — their packuments are not the slow ones.

.github/workflows/publish.yml
```diff
@@ -2,6 +2,18 @@ name: Snapshot
...
+        description: 'Rebuild the "latest" image from this ref even though nothing was published'
+        type: boolean
+        default: false
+
+# Two runs packing the same version from different trees would race to push the
+# same tag. Never cancel: a half-finished publish is worse than a queued one.
+concurrency:
+  group: publish-${{ github.ref }}
+  cancel-in-progress: false
```
laskevych, line 18: 💡 Suggestion (no change needed now): Be aware that the queue keeps only one pending run.

**Why:** With `cancel-in-progress: false`, GitHub still cancels a pending run when a newer one queues. If the "Version Packages" merge is the one that gets dropped, the next run does the release publish from a later commit. The npm release and `:latest` then include a change that is not in that release's changelog.

**Fix:** None required. Worth a line in the comment above `concurrency`, plus a habit of not merging right behind the Version PR.

tools/pack-owox-image-packages.mjs
```diff
@@ -0,0 +1,267 @@
...
+// workspaces are built by the step before this one rather than by their own
+// prepack. `npm pack` on a missing dist warns and exits 0, which would ship an
+// image with no UI.
+for (const name of toPack) {
+  const { dir, manifest } = workspaces.get(name);
+  const needsDist = (manifest.files ?? []).some(entry =>
+    entry.replace(/^\.\//, '').startsWith('dist')
+  );
+
+  if (needsDist && !fs.existsSync(path.join(dir, 'dist'))) {
```
laskevych: 💡 Suggestion: Assert that `dist/` is non-empty, not just present.

**Why:** An empty `dist/` left by a failed or cleaned build passes the guard. For workspaces without `prepack` (`@owox/web`), `npm pack` then ships a tarball with no output. `owox --version` still succeeds, so the image has no UI and nothing fails.

**Fix:** `fs.readdirSync(distDir).length > 0`, or check the manifest's `main` / entry file.

laskevych (approved):
✅ Approving. Taking the registry off the critical path is the right depth of fix, and the guards (path, version, dist, peer closure) turn the silent failure modes into loud ones. I checked that `changeset publish` does print `New tag:` for every published package and exits 1 on a partial failure, so the `published` detection holds. I also checked that nothing in the repo or the `k8s` overlays depends on the old global install path or on npm snapshots for the image. All comments are non-blocking. The one I'd take before merge is the `main`-only guard on the job, since `workflow_dispatch` is the first time this workflow can run from another branch.

.github/workflows/publish.yml
zapolsky: Taken in `fc969d9` — `if: github.ref == 'refs/heads/main'` on the job.

You are right that I opened this myself: the workflow was `push: branches: [main]` only until I added `workflow_dispatch` for the rebuild escape hatch, and I did not think through that the dropdown offers every branch. Publishing to npm is irreversible and the k8s updater auto-selects the newest GHCR tag, so a dispatch from a feature branch was a live hole, not a theoretical one.

.github/workflows/publish.yml
zapolsky: Taken in `fc969d9`. A step now emits `::warning::` when the tag is absent, this run published nothing, and `rebuild_latest_image` was not set — naming the dispatch input to use.

Worth saying plainly: this is the same failure class the PR exists to remove. The old gate fell through silently after 30 attempts; my `published == 'true'` gate turned a different case silent in the same way. Thanks for catching that I had reintroduced it one step over.

.github/workflows/publish.yml
zapolsky: Taken as wording in `fc969d9`:

> Rebuild the "latest" image from this ref HEAD (there are no release tags, so run before anything else merges, or from a branch cut at the release commit)

I looked for a real guard first and did not find one worth its cost. Comparing the packed tarball's shasum against `npm view owox@<version> dist.shasum` would be exact, but it leans on `npm pack` being byte-reproducible; if it is not, a legitimate rebuild fails and the escape hatch is useless exactly when it is needed. Pushing the tags `changeset publish` already creates locally is the better fix and belongs in its own change.

.github/workflows/publish.yml
zapolsky: Added to the comment above `concurrency` in `fc969d9`:

> GitHub still keeps only one run pending, so merging right behind the Version Packages PR can drop that run and publish the release from a later commit.

No code change, agreed. Noting for anyone reading later that this interacts with the warning added for the sibling comment: if the dropped run was the release one, the next run publishes nothing, so the `latest` image is skipped — and now at least says so.

tools/pack-owox-image-packages.mjs
zapolsky: Taken in `fc969d9` — `fs.readdirSync(distDir).length === 0` now fails alongside the missing-directory case.

Confirmed the gap was real rather than theoretical: with `apps/web/dist` replaced by an empty directory the old guard passed, and the new one fails with `@owox/web has no built dist/`. That is the worst shape of this bug, since `owox --version` in the image still succeeds and the only symptom is a container with no UI.

.changeset/6977-pack-docker-image-without-npm-snapshot.md
zapolsky: Fair question, and you are right on the facts — those three are not what made this slow:

| package | packument |
|---|---|
| `@owox/plugin-sdk` | 402 KB |
| `@owox/api-client` | 955 KB |
| `@owox/ctl` | 1.1 MB |
| `@owox/backend` | **7.4 MB** |

So keeping snapshots for them would cost nothing measurable, and technically it is about ten lines.

I did not change the behaviour, for one reason only: the task's DoD says "don't publish next npm release" without carve-outs, and the meeting that decided this discussed `owox` specifically. Narrowing it unilaterally felt like the wrong call to make inside a review thread.

What I did do is stop the changeset and `release-strategy.md` understating the scope — both now say it applies to every released package and that these three are left with no snapshot channel at all, since they are not in the image either:

> This applies to every package the repository releases, not just `owox`. In particular `@owox/plugin-sdk`, `@owox/api-client` and `@owox/ctl` no longer get snapshots either, and they are not part of the container image, so changes to them now reach consumers only in a release.

What I could not establish: whether anyone actually consumes those `next` tags. Weekly download counts (2.2k–2.8k) do not separate tag from version, and I have no visibility into plugin authors' pins outside this org. If you know of a consumer — the extension repo, or a plugin pinning `@owox/plugin-sdk@^0.x.0-next-…` — say so and I will restore snapshots for just those three in this PR. Otherwise I would rather raise it at the daily than widen the change on my own.

kalnik-a-a (approved):

PR #1577, opened 2026-08-28
feat(connectors): custom no-code connectors — declarative engine, web builder, MCP

zapolsky:
A connector to any REST API is described by a JSON manifest instead of code. One engine
executes that manifest, a web builder edits it with a live test against the real API, and
versions are published and rolled back without releasing the application.

The complete manifest grammar — every key and every accepted value — is added by this PR as
[`docs/connectors/manifest-reference.md`](https://github.com/OWOX/owox-data-marts/blob/feat/connector-builder-declarative/docs/connectors/manifest-reference.md),
with an assistant-facing mirror in `manifest-reference.llms.txt` that a test pins byte for
byte against the MCP tool's own constant.

---

## What changes

The single product change: adding a data source no longer means shipping a release.

| Before | Now |
| --- | --- |
| The connector catalogue is fixed in package code. | A connector is data in the project's database. |
| A new source means a developer task, a PR and a release. | A new source is created in the UI in one sitting. |
| A customer with a niche API waits for the roadmap. | The customer or a partner builds it themselves. |
| A fix to a connector ships with the next release. | Versions publish and roll back instantly, per project. |

A custom connector is a peer of a bundled one — same data mart binding, same schedule, same
storage. The only difference is where its definition came from.

## Three editors, one manifest

Not three features: three editors of the same document, and switching between them loses
nothing.

- **Visual builder** — forms, three panes, live test.
- **Code mode** — the raw JSON, in the same tab.
- **MCP assistant** — 11 of the 29 MCP tools cover connector discovery, authoring and runs.

An assistant can go from "I need data from X" to a published connector without ever touching a
secret: it reads the grammar, writes the manifest, runs a test against the live API and hands
the user a link to open it.

## Inside the builder

The left rail carries `General`, `Parameters`, `Authentication` and the node list. The centre
is the editor for the selected section. The bottom dock shows the test run: raw rows from the
API, and the sample after processing.

Every save creates a draft with the next version number. Publishing parses and validates the
manifest and makes that version active. Any past version can be activated again — that is the
rollback.

Parameters pane of the connector builder: a table of configuration parameters with type, required, secret, attributes, label, default and description columns, and an Advanced Parameters card below it

`Parameters` are the fields of the setup form that whoever connects this connector to a data
mart fills in. `SECRET` encrypts and masks the value; `PINNED` and `ADVANCED` control where the
field lands in that form. The two parameters the engine injects into every manifest by itself
get their own card.

[...]

PR #1628, opened 2026-09-18
feat(data-marts): cap manual backfill at 31 days per run

prokopetsv:
# Problem

A manual backfill accepted any `StartDate`/`EndDate` range. Nothing in the form, the API, or the connector core capped it, so a user could request months of history in one run. That run fetched the whole range in a single connector process, took hours, hit the 60-minute stuck-trigger timeout, and could not be resumed because manual backfill never writes `LastRequestedDate`. The only feedback was a run that eventually failed.

# Solution

Enforce one limit, 31 inclusive days, at every layer, and tell the user before they start. 31 rather than 30 so that any full calendar month fits in a single run.

- The connectors package owns the constant (`MAX_MANUAL_BACKFILL_DAYS`) and `AbstractConnector._getManualBackfillDateRange` rejects a longer range as a last line of defense. Every source reaches that method through the shared `getStartDateAndDaysToFetch`, so no connector bypasses it.
- The backend validates the range in `ConnectorRunService.run` before any run row is written, so the API returns a clear 4xx instead of creating a run that fails later. A new helper parses the dates with a zod schema (including impossible days such as February 30), defaults a missing `EndDate` to today, clamps a future `EndDate`, and applies the limit. It reads the limit from the connectors package lazily, so specs that stub `Core` for unrelated reasons still load, while a stale connectors build fails loudly on the first real backfill instead of silently using a different value.
- The web form explains the limit, shows how many days the chosen period covers, validates end-after-start and the 31-day maximum, and caps both date inputs at today. The web keeps its own copy of the constant because it does not depend on the connectors package.

Splitting longer periods into sequential runs was considered and deliberately left out; it is tracked as separate work. The `AbstractConnector.js` diff is the 4-line guard only.

# Changes

- Added `MAX_MANUAL_BACKFILL_DAYS = 31` to the connectors `CommonConstants` and enforced it in `AbstractConnector._getManualBackfillDateRange`, with a test
- Added `manual-backfill-range.ts` in the backend with `getMaxManualBackfillDays`, `parseManualBackfillRange`, and `prepareManualBackfillPayload`, plus a spec covering format, order, future dates, clamping, and the limit
- Validated and normalized the backfill payload in `ConnectorRunService.run` before creating the run and trigger; injected `SystemTimeService` for "today"
- Documented the limit in the `RunDataMartRequestApiDto` payload description
- Added a web `manual-backfill` constant module with the limit and date helpers, and a test
- Updated `ConnectorRunForm` with a limit notice, a live day count, end-after-start and 31-day validation, and `max={today}` on date inputs; removed the explicit input `id` so the label now points at the input; added a form test
- Documented the limit in the connector data mart setup guide
- Added changeset `6953-manual-backfill-31-day-limit.md`

Manual Run sheet with Backfill selected, empty Start Date and End Date fields, and a notice that one run covers at most 31 days
Two Manual Run sheets: a 31-day period with the Run button enabled, and a 49-day period with a limit warning and Run disabled


apps/backend/src/data-marts/utils/manual-backfill-range.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Validates a MANUAL_BACKFILL payload before a run is created and normalizes its dates
+ * (EndDate filled in and clamped). Non-backfill payloads pass through untouched.
+ */
+export function prepareManualBackfillPayload(
+  payload: Record<string, unknown> | undefined,
+  today: Date
+): Record<string, unknown> | undefined {
+  if (!payload || payload.runType !== MANUAL_BACKFILL_RUN_TYPE) return payload;
```
laskevych, line 108: 🔴 **Blocking:** Return the payload unchanged when `data` carries no `StartDate`, instead of rejecting the run.

**Why:** `prepareManualBackfillPayload` requires `StartDate` for *every* `MANUAL_BACKFILL` payload, but not every connector has backfill date fields. `GoogleSheetsSource` declares no `CONFIG_ATTRIBUTES.MANUAL_BACKFILL` parameter, and `GoogleSheetsConnector` overrides `_processRunConfig` ("Full refresh mode activated") and never calls `getStartDateAndDaysToFetch` — a backfill there is a full refresh. Today the web renders that section with no date inputs, `formState.isValid` is `true`, and it posts `{ runType: 'MANUAL_BACKFILL' }` with `data` undefined. After this change that request fails with `400 StartDate is required in YYYY-MM-DD format`; before it started a run. The same shape is a documented public contract: `docs/api/api-client.md` states "connectors without backfill fields can omit `data`", and `packages/api-client/src/data-mart-runs.test.ts` asserts "accepts a manual backfill without connector-specific fields". No e2e caught it because every `MANUAL_BACKFILL` e2e case targets a non-existent Data Mart and stops at 404.

**Fix:** Skip validation when `data` has no `StartDate` key and pass the payload through; keep validating as soon as `StartDate` is present. `AbstractConnector._getManualBackfillDateRange` already raises "StartDate is required for manual backfill" for the connectors that genuinely need it, so nothing is lost. Worth a unit case for a specification with no date fields, and — if a dateless backfill is meant to stay reachable — hiding the "Backfill (custom period)" option when the specification exposes no `MANUAL_BACKFILL` fields, since that section currently shows only the limit notice.

packages/connectors/src/Constants/CommonConstants.js
```diff
@@ -18,6 +18,11 @@ var RUN_CONFIG_TYPE = {
   MANUAL_BACKFILL: 'MANUAL_BACKFILL',
 };
 
+// Upper bound on the inclusive date range a single MANUAL_BACKFILL run may cover.
+// 31 so that any full calendar month fits in one run. The backend splits longer
```
laskevych, line 22: 🟡 **Important:** Drop the "backend splits longer ranges into sequential runs" sentence.

**Why:** The backend rejects longer ranges; splitting is explicitly out of scope for this PR. The comment promises a guarantee that does not exist, in the one file a connector author is most likely to read.

**Fix:** Say the backend rejects ranges above the limit before a run is created, and that the connector re-checks as a last line of defense.

packages/connectors/test/Core/AbstractConnector.test.js
```diff
@@ -0,0 +1,37 @@
...
+  proto._getManualBackfillDateRange.call({
+    config: {
+      StartDate: { value: new Date(start) },
+      EndDate: { value: new Date(end) },
+      logMessage: () => {},
+    },
+  });
+
+describe('_getManualBackfillDateRange', () => {
+  it('accepts a full 31-day calendar month as one run', () => {
```
laskevych, line 25: 🟡 **Important:** This test never runs in CI.

**Why:** The `test-owox.yml` matrix covers `@owox/backend`, `@owox/web`, `owox`, both idp packages, `@owox/api-client` and `@owox/ctl` — there is no `@owox/connectors` entry, so `npm run test -w @owox/connectors` is executed by no workflow. The only automated coverage of the connector-side guard (the stated last line of defense) is dead weight today.

**Fix:** Add `- name: 'Overall test @owox/connectors' / command: 'npm run test -w @owox/connectors'` to the matrix, or note here that the suite is intentionally local-only.

packages/connectors/src/Core/AbstractConnector.js
```diff
@@ -7,343 +7,348 @@
...
-      const daysToFetch = Math.max(0, Math.floor((today.getTime() - startDate.getTime()) / (1000 * 60 * 60 * 24)) + 1);
-
-      return [startDate, daysToFetch];
+    // Calculate days between start and end date
+    const daysToFetch = Math.max(
+      0,
+      Math.floor((endDate.getTime() - startDate.getTime()) / (1000 * 60 * 60 * 24)) + 1
+    );
+
+    if (daysToFetch > MAX_MANUAL_BACKFILL_DAYS) {
```
laskevych, line 251: 🟡 **Important:** Coordinate this guard with #1577.

**Why:** `feat/connector-builder-declarative` adds a second `_getManualBackfillDateRange` in this same file for manifest-driven connectors. It has no day-limit check, so after both merge, declarative connectors would resolve their range through a path that never sees `MAX_MANUAL_BACKFILL_DAYS` — and both branches touch this file, so the conflict needs a manual resolution anyway.

**Fix:** Whichever merges second, port the same check into the declarative engine's range method (and its test).

apps/backend/src/data-marts/utils/manual-backfill-range.ts
```diff
@@ -0,0 +1,114 @@
...
+function formatUtcDay(ms: number): string {
+  return new Date(ms).toISOString().slice(0, 10);
+}
+
+function toUtcDay(value: string): number {
+  return Date.parse(`${value}T00:00:00.000Z`);
+}
+
+/** A calendar day that exists: the regex catches the shape, the refine catches 2026-02-30. */
+const isoDaySchema = z
```
laskevych, line 33: 💡 **Suggestion:** Use `z.string().date()` instead of the regex plus round-trip refine.

**Why:** The pinned `zod ^3.25.64` already validates both the `YYYY-MM-DD` shape and real calendar days (it rejects `2026-02-30`), which is exactly what `ISO_DAY_PATTERN` and the `formatUtcDay(ms) === value` refine reproduce.

**Fix:** `const isoDaySchema = z.string().date();` — the existing spec cases (`01/09/2026`, `2026-02-30`, `20260901`) should keep passing unchanged.

apps/backend/src/data-marts/dto/presentation/run-data-mart-request-api.dto.ts
```diff
@@ -94,7 +94,8 @@ export class RunDataMartRequestApiDto {
       data: { StartDate: '2026-07-01', EndDate: '2026-07-31' },
     },
     description: `Payload for the manual run. Omit it or select INCREMENTAL for an incremental run.
-    MANUAL_BACKFILL can include connector-specific fields in data.`,
+    MANUAL_BACKFILL can include connector-specific fields in data. The StartDate/EndDate range may
+    cover at most 31 days; longer ranges are rejected.`,
```
laskevych, line 98: 💡 **Suggestion:** Document the 31-day cap in the api-client docs too, together with the accepted date format and how a missing or future `EndDate` is counted.

**Why:**
1. This OpenAPI description is updated, but integrators read "Manage Data Mart runs" in `docs/api/api-client.md` — the example there is `2026-07-01`…`2026-07-31` with no mention that a longer range is now rejected.
2. The limit applies after a missing or future `EndDate` becomes today, so "longer ranges are rejected" is not quite what happens: `2026-09-01`…`2026-12-31` is accepted and runs only up to today.
3. Neither place says which date format is accepted (see the comment on line 22 of `manual-backfill-range.ts`).

**Fix:** One sentence here and next to the `MANUAL_BACKFILL` example in `docs/api/api-client.md`: which date format is accepted, that a missing or future `EndDate` counts as today, and that the resulting range may cover at most 31 days.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
```diff
@@ -45,6 +92,14 @@ export function ConnectorRunForm({ configuration, onClose, onSubmit }: Connector
 
   const { dataMart } = useDataMartContext();
 
+  const runType = form.watch('runType');
+  const startDate = form.watch('data.StartDate');
+  const endDate = form.watch('data.EndDate');
+  const today = new Date().toISOString().slice(0, 10);
+  // The backend defaults a missing EndDate to today, so the preview does the same.
+  const effectiveEndDate = endDate === undefined || endDate === '' ? today : endDate;
```
laskevych, line 100: 💡 **Suggestion:** Do not preview an end date the form still requires.

**Why:** Every backfill field is registered with `required: true`, so an empty End Date can never reach the backend — but `effectiveEndDate` falls back to today, so with only a Start Date of `2026-01-01` the notice reports "This period covers 261 days, which exceeds the 31-day limit" for a period the user never chose, while Run is disabled for a different reason. The comment "The backend defaults a missing EndDate to today" describes a path the required field prevents.

**Fix:** Keep the "Pick a start and end date…" text until both dates are set, and count days only from the values the user actually entered.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
```diff
@@ -174,6 +232,9 @@ export function ConnectorRunForm({ configuration, onClose, onSubmit }: Connector
                     )}
                   />
                 ))}
+              <Alert data-testid='backfill-limit-notice'>
```
laskevych, line 235: 💡 **Suggestion:** Render the over-limit state with the `destructive` Alert variant.

**Why:** `@owox/ui` `Alert` has `default` and `destructive`; this notice always uses `default`, so the blocking message looks exactly like the informational one and only the wording changes.

**Fix:** `variant={backfillDays > MAX_MANUAL_BACKFILL_DAYS ? 'destructive' : 'default'}`.

apps/backend/src/data-marts/services/connector/connector-run.service.ts
```diff
@@ -40,15 +43,23 @@ export class ConnectorRunService {
       );
     }
 
-    const dataMartRun = await this.createDataMartRun(dataMart, createdById, runType, payload);
+    // Validates the backfill period (format, order, per-run day limit) before any row is
+    // written, so the caller gets a clear error instead of a failed run.
+    const preparedPayload = prepareManualBackfillPayload(payload, this.systemTimeService.now());
```
laskevych, line 48: ℹ️ **Note (rollout):** `executeInterruptedRuns` below replays `run.additionalParams` without this validation. A pre-deploy backfill longer than 31 days that the release interrupts is picked up again by the `RETRY_INTERRUPTED_CONNECTOR_RUNS` task (every 15 minutes) and hits the connector guard "Manual backfill is limited to 31 days per run".
- Most connectors end that run as FAILED — the right outcome, but it shows up as a failed run right after the release.
- TikTok Ads wraps `getStartDateAndDaysToFetch()` in a try/catch that only logs, so there the run ends as SUCCESS and is billed with no rows loaded.

Worth one line in the changeset so operators recognise it; rethrowing in TikTok's catch would turn the second case into a normal failure.

laskevych (changes requested):
One blocking defect, the rest are non-blocking.

I also checked the 573-line `AbstractConnector.js` diff separately: a whitespace- and quote-normalized comparison of `main` vs this branch shows the only semantic change is the new guard (everything else is trailing commas and the closing `};`).

Two comments are anchored on the nearest changed line because the file they are really about is not part of this diff: the CI matrix (`.github/workflows/test-owox.yml`) and `docs/api/api-client.md`.

apps/backend/src/data-marts/utils/manual-backfill-range.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (!Number.isInteger(limit) || limit < 1) {
+    throw new Error(
+      'MAX_MANUAL_BACKFILL_DAYS is missing from @owox/connectors; rebuild the package'
+    );
+  }
+  return limit;
+}
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+const ISO_DAY_PATTERN = /^\d{4}-\d{2}-\d{2}$/;
```
laskevych, line 22: 🟡 **Important:** Accept the date part of an ISO-8601 timestamp here, or document that `StartDate`/`EndDate` must now be exactly `YYYY-MM-DD`.

**Why:**
1. On `main` any value `new Date()` can parse worked: the DTO only checks that `data` is an object, and `AbstractConnector._processRunConfig` converts it with `new Date(value)`. `StartDate: '2026-09-01T00:00:00.000Z'` ran a normal backfill.
2. This schema rejects it with `400 StartDate is required in YYYY-MM-DD format`. An api-client caller that passes `Date` objects hits exactly this: `data` is typed `Record<string, unknown>`, and `JSON.stringify` turns a `Date` into an ISO timestamp.
3. The changeset, the OpenAPI description and `docs/api/api-client.md` don't mention the stricter format, and the docstring on `parseManualBackfillRange` ("the same way AbstractConnector does") no longer holds.

**Fix:** Keep the calendar day of a timestamp and still reject impossible days:

```ts
const isoDaySchema = z
  .string()
  .regex(/^\d{4}-\d{2}-\d{2}(?:T.*)?$/)
  .transform(value => value.slice(0, 10))
  .pipe(z.string().date());
```

If the strict format is intended, state it in the DTO description, `docs/api/api-client.md` and the changeset instead.

laskevych (commented):
Second pass: a few more non-blocking comments.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
```diff
@@ -24,13 +25,59 @@ import type { ConnectorRunFormData } from '../../../shared/model/types/connector
...
+        'Start date cannot be in the future',
+    };
+  }
+  if (fieldName === 'EndDate') {
+    return {
+      period: (value, formValues) => {
+        const startDate = formValues.data.StartDate;
+        if (toUtcDayMs(value) === undefined || toUtcDayMs(startDate) === undefined) return true;
+        const days = countBackfillDays(startDate, value);
+        if (days === 0) return 'End date must be on or after the start date';
```
laskevych, line 70: 🟡 **Important:** Tell the user why Run is disabled when End Date is before Start Date or Start Date is in the future.

**Why:**
1. The form uses the default `onSubmit` mode, so `validate` messages are written only by `handleSubmit`. These rules also make `isValid` false, which disables Run — the only submit button — and Enter can't submit through a disabled default button. "End date must be on or after the start date" and "Start date cannot be in the future" therefore never render.
2. For Start `2026-07-10` / End `2026-07-01`, `countBackfillDays` returns `0`, so the notice asks to "Pick a start and end date…" although both are picked.
3. A typed future Start Date (`max` only limits the picker, and the form is `noValidate`) shows "This backfill covers 6 days." next to a disabled Run.
4. On `main` the same reversed input started a run that failed with an explicit "EndDate … cannot be earlier than StartDate …", so users now get less explanation than before.
5. The tests reach these messages only through `fireEvent.submit(form)`, which bypasses the disabled button, and none asserts that Run is disabled.

**Fix:** Derive one `{ days, reason }` from the watched dates and use it for both `validate` and the notice, so a reversed or future period gets its own notice text. Alternatively, switch to `mode: 'onChange'` and add `deps: ['data.EndDate']` to the Start Date registration. `min={startDate}` on End Date also prevents the reversed pick.

**Test:** Assert `expect(screen.getByRole('button', { name: 'Run' })).toBeDisabled()` together with the notice text instead of submitting the form directly.

packages/connectors/src/Core/AbstractConnector.js
```diff
@@ -7,343 +7,348 @@
...
   //---- constructor -------------------------------------------------
-    constructor(config, source, storage = null, runConfig = null) {
-
-      if( typeof config.setParametersValues !== "function" ) { 
-        throw new Error(`Unable to create a Connector. The first parameter must inherit from the AbstractConfig class`);
-        
-      } else if( typeof source.fetchData !== "function" ) {
-        throw new Error(`Unable to create a Connector. The second parameter must inherit from the AbstractSource class`);
+  constructor(config, source, storage = null, runConfig = null) {
+    if (typeof config.setParametersValues !== 'function') {
```
laskevych, line 11: 💡 **Suggestion:** Revert the formatting-only changes in this file and keep just the guard.

**Why:**
1. The pre-commit hook skips this file: `packages/connectors/.prettierignore` lists `src/`, and lint-staged hands files under `packages/connectors/` only to the package config, which runs Prettier with `--ignore-path ./.prettierignore`. The formatting came from an editor or a root Prettier run, so the note in the description doesn't hold.
2. The package keeps `src/` unformatted: 181 of its 185 files are not Prettier-formatted, and no CI check covers them, so reverting breaks nothing.
3. ~280 reformatted lines take over `git blame` for the core connector class and bury the 6-line behavior change in the diff.

**Fix:** `git checkout origin/main -- packages/connectors/src/Core/AbstractConnector.js`, re-add the 6-line guard, and drop the sentence about the hook from the description.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
laskevych: 💡 **Nit:** Pass `role='status'` to this `Alert`.

**Why:** `@owox/ui` `Alert` sets `role='alert'`, which is assertive and atomic. This notice changes on every date input, so screen readers interrupt the field to re-read the whole sentence each time. `FormulaEditor` uses `role='status'` for text that updates while the user types, for the same reason.

**Fix:** `<Alert role='status' data-testid='backfill-limit-notice'>`. `Alert` spreads props after its default `role`, so this overrides it.

apps/backend/src/data-marts/utils/manual-backfill-range.ts
prokopetsv: Correct, and a real regression. Verified: `GoogleSheetsSource` declares no `MANUAL_BACKFILL` parameter, and `GoogleSheetsConnector._processRunConfig` logs "Full refresh mode activated" without ever calling `getStartDateAndDaysToFetch`. `docs/api/api-client.md:310` and `data-mart-runs.test.ts:222` both document the dateless shape.

Fixed as you suggested: `prepareManualBackfillPayload` now returns the payload untouched when `data` is absent or carries no `StartDate`, and validates as soon as `StartDate` is present. `AbstractConnector` still raises "StartDate is required for manual backfill" for the connectors that need one. Added two unit cases: no `data` at all, and `data` without `StartDate`.

I did not hide the Backfill option for dateless connectors. That changes behaviour beyond this PR, and the full-refresh path is intentionally reachable.

packages/connectors/src/Constants/CommonConstants.js
prokopetsv: Fixed. The comment now says the backend rejects a longer range before a run is created, and that this is re-checked here as a last line of defense.

packages/connectors/test/Core/AbstractConnector.test.js
prokopetsv: Confirmed, the matrix had no `@owox/connectors` entry. Added the job to `test-owox.yml`, so the connector-side guard is now actually covered in CI.

packages/connectors/src/Core/AbstractConnector.js
prokopetsv: Acknowledged, no change in this PR. #1577 adds a second `_getManualBackfillDateRange` for manifest-driven connectors with no day check, and both branches touch this file, so the conflict needs a manual resolution anyway. Whichever merges second ports the guard and its test into the declarative engine. I will flag the same on #1577 so it is not lost.

apps/backend/src/data-marts/utils/manual-backfill-range.ts
prokopetsv: Done. Verified against the pinned zod 3.25.76: it accepts `2026-07-01` and rejects `2026-02-30` and `01/09/2026`, so the regex plus round-trip refine was redundant. Combined with your timestamp comment below, the schema is now a regex that tolerates a timestamp suffix, a transform to the date part, then `.pipe(z.string().date())`. The existing spec cases pass unchanged.

apps/backend/src/data-marts/dto/presentation/run-data-mart-request-api.dto.ts
prokopetsv: Done, in both places. The OpenAPI description and `docs/api/api-client.md` now state the accepted format, that a missing or future `EndDate` counts as today, and that the resulting period may cover at most 31 days.

Your point 2 was right and I had it wrong: the limit applies after clamping, so `2026-09-01`…`2026-12-31` is accepted and runs only up to today. The wording reflects that now rather than saying "longer ranges are rejected".

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
prokopetsv: Fixed. `effectiveEndDate` is gone, and the notice reports only what the user actually picked, so the misleading "261 days" case can no longer occur. The comment about the backend defaulting a missing `EndDate` went with it, since the required field makes that path unreachable from the form.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
prokopetsv: Done. The notice uses `destructive` for every blocked state and `default` only when the period is valid.

apps/backend/src/data-marts/services/connector/connector-run.service.ts
prokopetsv: Good catch, confirmed at `connector-run.service.ts:117` — `executeInterruptedRuns` replays `run.additionalParams` without this validation. Added a line to the changeset so operators recognise the failed run right after the release.

I left the TikTok Ads try/catch alone. Turning a swallowed error into a real failure is a behaviour change in a connector this PR does not otherwise touch, and it deserves its own PR with its own test.

apps/backend/src/data-marts/utils/manual-backfill-range.ts
prokopetsv: Correct, and a second regression: `JSON.stringify` turns a `Date` into an ISO timestamp, and that shape ran fine on `main` through `new Date(value)`. Applied your snippet, so the date part is kept and impossible days are still rejected. Added a unit case asserting `2026-09-01T00:00:00.000Z` normalises to `2026-09-01`.

Both accepted formats are now documented in the OpenAPI description and `docs/api/api-client.md`.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
prokopetsv: Correct on all five points. Reworked it around a single `readBackfillPeriod` that returns a discriminated state, read by both the notice and the field validators.

A reversed period now says "The end date must be on or after the start date." and a future start says "The start date cannot be in the future.", instead of "Pick a start and end date…". The tests no longer submit the form directly: each asserts the notice text and the Run button state together inside `waitFor`, since `isValid` settles asynchronously. That async settling was real — the enabled-button assertion failed until I moved it inside the `waitFor`.

packages/connectors/src/Core/AbstractConnector.js
prokopetsv: You are right and my PR description was wrong. `packages/connectors/.prettierignore` lists `src/`, and `packages/connectors/.lintstagedrc.json` runs Prettier with `--ignore-path ./.prettierignore`, so the hook never touches this file. The reformat came from a root Prettier run of mine.

Restored from `origin/main` and re-added only the guard: the diff is 4 lines now. Confirmed after committing that the pre-commit hook left it untouched, exactly as you said. Removing the claim from the PR description too.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
prokopetsv: Done, with a test asserting the attribute. `Alert` spreads props after its default `role`, so the override works as you described.

prokopetsv:
Thanks, this was a thorough pass. All 13 comments were valid; 12 are fixed in 57d1c57, the 13th (#1577 coordination) is a cross-PR note.

Two were real regressions I had missed, both now covered by unit cases:
- a `MANUAL_BACKFILL` with no `StartDate` got a 400, breaking dateless connectors like Google Sheets and the documented api-client contract;
- an ISO-8601 timestamp, which is what `JSON.stringify` produces for a `Date`, was rejected although it worked on `main`.

You were also right that the `AbstractConnector.js` reformat did not come from the pre-commit hook. The package ignores `src/` in its Prettier config, so that was my own root Prettier run. The file is restored from `main` with only the 4-line guard on top, and I have corrected the PR description, which claimed otherwise.

Verification after the fixes: connectors 293 passed, web connectors 66 passed, backend helper/run-service/OpenAPI specs 38 passed, `nest build` clean, lint and markdownlint clean. The connectors suite now runs in CI, so that last number is no longer local-only.

The branch is behind `main` and needs a rebase before merge.

prokopetsv:
Cross-reference for the open coordination point raised in review: posted the details on #1577 (OWOX/owox-data-marts#1577 (comment)).

Summary: that branch rewrites `AbstractConnector.js` (349 → 1294 lines) and rebuilds `_getManualBackfillDateRange` around the declarative context with no day-limit check. It returns `{ startDate, endDate }` instead of the `[startDate, daysToFetch]` tuple this guard inspects, so the four lines cannot be copied across and the check needs rewriting there.

Six files overlap between the two PRs, so a manual resolution is needed either way. Whichever lands second owns porting the guard and its test.

prokopetsv:
Scope note, raised while reviewing what the cap actually guarantees: **it bounds the manual backfill form, not the amount of data a single run fetches.** An incremental run can exceed 31 days today, and still can after this PR.

**Why.** `getStartDateAndDaysToFetch` branches into `_getManualBackfillDateRange`, where the new guard lives, and `_getIncrementalDateRange`, which has no limit. On the backend, `prepareManualBackfillPayload` returns early unless `runType === MANUAL_BACKFILL`, so incremental payloads are never inspected either.

**Two ways past the limit, neither of them adversarial:**

1. **Reimport lookback window.** `_getIncrementalStartDate` returns `LastRequestedDate - ReimportLookbackWindow` days. Setting that field to 60 fetches ~61 days in one run. It is a plain `requiredType: "number"` with `default: 2` and no maximum, so 365 is equally accepted.
2. **The first incremental run, with default settings.** With no `LastRequestedDate` in state, the start date is the 1st of last month. Run today (21 Sep) that is 52 days — already over the cap, on a completely ordinary run.

**Why I am not changing it here.** Applying the limit to the incremental path would change behaviour for every existing connector and break every first run under default configuration. That is a much larger change than this PR, and it needs a product decision about what happens to a first run that legitimately needs ~2 months of history.

**What I think is worth confirming before merge:** which of these two was the goal.

- *Bound the backfill form*, so a user cannot request a year of history in one click — this PR does that, and the changeset and docs describe it accurately ("a manual backfill run covers at most 31 days").
- *Bound run duration*, so no single run can time out — this PR does not do that, and the longest runs in practice (first incremental runs, wide lookback windows) are untouched.

If it is the first, no action needed; the wording is already scoped correctly, and this comment is just a record of the boundary. If it is the second, the incremental path needs its own issue, and the 60-minute stuck-trigger timeout cited in the PR description is still reachable without going anywhere near a manual backfill.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
```diff
@@ -174,6 +254,13 @@ export function ConnectorRunForm({ configuration, onClose, onSubmit }: Connector
                     )}
                   />
                 ))}
+              <Alert
+                role='status'
+                variant={backfillPeriod.status === 'ok' ? 'default' : 'destructive'}
```
laskevych: 💡 **Suggestion:** Keep the untouched `incomplete` state on the `default` variant, and skip the notice when the specification has no `StartDate` field.

**Why:**
1. `variant` is `destructive` for everything except `ok`, so the hint "A backfill run can cover at most 31 days. Pick a start and end date…" renders as a red error the moment Backfill is selected, before the user has done anything wrong.
2. The `Alert` sits outside the field filter, so a connector with no backfill date fields (Google Sheets) shows the same red "Pick a start and end date" with no date inputs, while Run is enabled and performs a full refresh.

**Fix:** `variant={status === 'ok' || status === 'incomplete' ? 'default' : 'destructive'}`, and render the `Alert` only when the filtered specification contains `StartDate`.

laskevych (approved):
All 13 comments from the previous rounds are addressed, and CI is green, including the new `@owox/connectors` job. I re-checked the dateless pass-through, the ISO-timestamp path, and the backend/connector day count across a UTC midnight (the normalized `EndDate` is persisted, so the two cannot disagree). One non-blocking suggestion below.

apps/web/src/features/connectors/edit/components/ConnectorRunSheet/ConnectorRunForm.tsx
prokopetsv: Both valid, fixed in 2b6c643.

1. **Untouched state is no longer red.** `variant` is now `default` for `ok` and `incomplete`, `destructive` only for the three states that actually block. Selecting Backfill shows the hint in neutral styling, as you said it should.

2. **The notice is gone for dateless connectors.** I hoisted the filtered backfill fields into `backfillFields` so the field list and the notice read the same source, and gate the `Alert` on `backfillFields.some(field => field.name === 'StartDate')`. A Google Sheets style connector now renders the Run configuration section with no period notice, and Run stays enabled for the full refresh.

Tests: a spec with no date fields asserts the notice is absent and Run is enabled, and the untouched-state test now asserts the Alert class does not contain `destructive` (with the over-limit test asserting it does). Web connectors suite is 67 passing.

PR #1629, opened 2026-09-21
fix(data-marts): a hidden column is not a broken schema (#6925)

zapolsky:
Fibery: [#6925 Allow hiding calculated fields](https://owox.fibery.io/Sprint_Work/Work_Item/Allow-hiding-calculated-fields-6925)

Extension side: OWOX/owox-data-marts-spreadsheets#66

## What the task turned out to be

The report said hiding was impossible. It is not — I reproduced the whole thing against a local stand, on the same `ROAS = SUM(revenue)/NULLIF(SUM(adCost),0)` shape from the bug report:

- hiding a calculated field and saving the schema works, and always did: nine requests, all `200`/`201`, no report SQL touched;
- the error in the screenshot comes from **Preview SQL** on a report that had already selected the field, not from saving the schema — the toast outlives the tab switch, which is why it looks like the schema refused to save;
- a hidden **plain** column behaves identically — same `400`, same wording — so there was no calculated-field-specific defect to fix;
- a hidden column a formula references keeps computing, as intended.

What is actually wrong is the diagnosis. Hidden and disconnected reach every surface the same way — each list built for reporting drops both — so the picker files a hidden column under **Disconnected columns** and the API answers *"They are missing from the current Data Mart output schema… contact your analyst to restore the schema."* Nothing is missing, and the analyst being contacted is usually the person who hid it.

## The change

`blendableSchema` now publishes `hiddenFieldNames`. It is the one place that reads the raw schema and does the hiding, so the answer costs nothing there and cannot drift from the lists it pruned. It covers every switch that hides a column: a Data Mart's own field (calculated ones included), a joined Data Mart's field, and `Hidden` in Join Settings. Joined names are built by the same flattening and the same unified-name builder as the fields beside them, so a name a report stored before the field was hidden matches byte for byte; the two namespaces cannot collide, since an own column is a dotted path and a joined one is prefixed.

With that list, both halves separate the two states:

- **the error** names Hidden and Disconnected columns apart, each with the fix that applies — show it again vs restore the schema. A report carrying one of each is told about both and keeps the schema advice, because one of those columns really is gone.
- **the picker** gives hidden columns their own amber block, out of the red one. Unchecking works the same. A pre-join slice stays with the disconnected: a join orphaned it, not a hidden column of this Data Mart.

`unknownColumns` in the error details still carries both kinds — it is what consumers key off, and narrowing it would change what they see for no reason. An older cached blendable schema names nothing as hidden, and everything reads exactly as it did before.

Two consumers that were reading the old diagnosis move with it: MCP's `field_not_found` used to send an agent to re-fetch a schema that will never list a hidden column — an endless loop — and the badge over the output-controls button said "Disconnected output controls" two centimetres above an amber "Hidden columns" block.

## Tests

Both throw sites were unpinned before this. The blended one asserted `errorDetails` only partially and passed identically before and after the change; the validator's schema mocks carried no hidden list at all, so that branch never ran. Each now has a hidden case and a mixed case, asserting the message and `hiddenColumns`. Plus: the walker's complement invariant (hidden ∪ visible = every connected field), the error helper's three shapes, four picker cases including the no-`hiddenFieldNames` fallback, the two MCP messages, and `hiddenFieldNames` in the OpenAPI spec beside `mainUniqueCountKeyFields`.

Backend 9802/9802, `nest build` clean, web 2549/2549.

## One decision left to the team

The `owox-changelog-entry` skill asks for a **bold** entry title and explicitly not a Markdown heading. This repository's markdownlint rejects that (MD036/MD041), and eight of the nine other pending entries use a heading. This entry keeps the heading; the conflict is reported rather than worked around.


apps/backend/src/data-marts/services/blended-report-data.service.ts
```diff
@@ -925,7 +926,13 @@ export class BlendedReportDataService {
     });
     if (unknownColumns.length === 0) return;
 
-    throwDisconnectedReportColumnsError(dataMart.id, unknownColumns);
+    // Which of them the analyst hid rather than lost. Read from the same RAW schema this check
+    // already holds, so the answer costs nothing and cannot drift from what `nativeNames` pruned.
+    throwDisconnectedReportColumnsError(
+      dataMart.id,
+      unknownColumns,
+      new Set(collectHiddenForReportingPaths(schemaFields))
```
laskevych: 🔴 **Blocking:** Pass `blendableSchema.hiddenFieldNames` into `assertNoOrphanedColumnReferences` instead of re-deriving the hidden set from the main Data Mart's own schema.

**Why:**
1. `unknownColumns` here also holds joined columns (`!blendedField || blendedField.isHidden`), but `collectHiddenForReportingPaths(schemaFields)` only knows the main Data Mart's own paths.
2. A report that selects `orders__internal_id` after `internal_id` is hidden on the `orders` Data Mart, or with **Hide from reports** in the join's Report Fields, still fails run, Generated SQL, Looker Studio and MCP with "Disconnected columns … contact your analyst to restore the schema" and no `hiddenColumns`.
3. The picker shows the same column in the amber **Hidden columns** block, so the editor and the run error disagree.
4. MCP never reaches the new hidden branch for these fields, so the agent still gets the "re-fetch the schema" advice.
5. The changeset promises this case: "a joined Data Mart's field, and **Hidden** in Join Settings".

**Fix:** `blendableSchema` is already computed in `resolveBlendingDecision`, and its `hiddenFieldNames` covers own paths too. Pass it as a parameter and build the set from it. The two new run-path specs then need `hiddenFieldNames` in their `makeBlendableSchema` mock.

**Test:**
- Run path: add one spec per joined switch (hidden on the joined Data Mart, hidden per join), asserting the message and `hiddenColumns`. The existing "treats blended fields hidden in the joined data marts setup as no longer available" spec checks only `unknownColumns`, so it passes with either wording.
- Producer: `blendable-schema.service.spec.ts` asserts `hiddenFieldNames` for own fields only. The joined loop and the `dto.isHidden` push are untested, including a nested (hashed) name.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
```diff
@@ -1979,6 +1994,64 @@ export function ReportColumnPicker({
...
+    [visibleUnresolvedColumns, visibleUnresolvedFilterOnlyColumns]
+  );
+
+  const hiddenUnresolvedRows = useMemo(
+    () => unresolvedRows.filter(({ name }) => hiddenColumnNames.has(name)),
+    [unresolvedRows, hiddenColumnNames]
+  );
+
+  // A pre-join slice belongs here whatever happened to it: it is addressed through a join, not
+  // through the main schema, so hiding a column of this Data Mart is never what orphaned it.
```
laskevych: 🟡 **Important (non-blocking):** Split `visibleUnresolvedSlices` by `hiddenColumnNames` too, so a slice on a hidden joined field lands in the amber block.

**Why:**
1. A slice is unresolved here when its joined field is gone or hidden: `knownSliceKeys` skips `isHidden` fields, and fields hidden on the joined Data Mart never reach `blendedFields`.
2. The slice's `rule.column` is the unified name, and `hiddenFieldNames` already lists it.
3. Save and run already call it hidden: "Hidden columns: "orders__status" … ask your analyst to show them in reports again".
4. The picker shows the same rule in red with "…contact your analyst to restore the schema". If the report also selects `orders__status`, the name appears in both blocks.
5. The comment's reason holds only for the main Data Mart's own columns, which a slice cannot address. It also sits above a memo that never touches slices.

**Fix:** render the slices whose `column` is in `hiddenColumnNames` in the amber block and keep the rest red. A removed join or a deleted field is never in `hiddenFieldNames`, so those stay red. Skip joined calculated fields: the backend refuses them with their own error.

**Test:** the fixture in "flags pre-join slices on hidden joined fields as disconnected with the slice removable" has no `hiddenFieldNames`. Add a variant with `hiddenFieldNames: ['b__hidden_field']` that expects the amber block. The new tests cover only a flat own name (`ROAS`).

The extension needs the same change: OWOX/owox-data-marts-spreadsheets#66.

docs/getting-started/setup-guide/output-controls.md
```diff
@@ -192,6 +192,22 @@ You have two options:
...
+
+Hiding a field from reporting is not the same event, and the picker says so. A report that already selected the field lists it under an amber **Hidden columns** label with an 👁 icon:
+
+> _They are still in the Data Mart, but hidden from reporting. Uncheck them and remove any filter, sort, aggregation or date bucket rule that references them, or ask your analyst to show them in reports again._
+
+Nothing is missing here and there is no schema to restore — someone decided the column should not be reported on. Any of the three switches leads here, and the fix is to reverse the one that was used:
+
+- **Hide from reports** in a field's ⋯ menu on the Data Mart's **Data Setup → Output Schema** — the Data Mart's own columns, calculated fields included.
+- The same action [on a joined Data Mart's field](./joinable-data-marts.md#per-field-overrides), which hides it for every report built on that Data Mart.
+- **Hidden** in Join Settings, which hides a joined field for this Data Mart's reports only.
```
laskevych: 🟡 **Important (non-blocking):** Fix the names and the link of the two joined switches, here and in the changeset.

**Why:**
1. Bullet 2 links to `#per-field-overrides`. That section describes the per-join hide in the join's **Report Fields** tab (bullet 3's switch), not the joined Data Mart's own Output Schema.
2. **Join Settings** has no **Hidden** control; it holds Joined Data Mart, SQL Alias and Join Fields. The per-join switch is **Hide from reports** in the ⋯ menu of the **Report Fields** tab.
3. The changeset (line 16) repeats "**Hidden** in Join Settings", so the release note sends users to the wrong tab.

**Fix:**
```md
- The same action on the joined Data Mart's own **Output Schema**, which hides the field from its reports and from every Data Mart that joins it.
- **Hide from reports** in the ⋯ menu of the join's **Report Fields** tab ([per-field overrides](./joinable-data-marts.md#per-field-overrides)), which hides a joined field for this Data Mart's reports only.
```

**Also:**
- Line 197: the icon is an eye-off, not 👁. The open eye is the "Show in reports" action, and the changeset alt text already says "eye-off".
- "Validation error on save" still says every stale rule gets a **Disconnected columns** error. A rule on a hidden column now gets **Hidden columns**.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
```diff
@@ -455,10 +455,23 @@ If truncated is true, not all matching rows were returned: narrow the query (few
...
+        return toStructuredToolError(
+          'field_not_found',
+          `Field(s) ${hidden.join(', ')} exist in this data mart but are hidden from reporting, so they cannot be queried and will not appear in get_data_mart_details_by_id. Drop them from "fields" and retry; only the data mart's analyst can show them again. Do not re-fetch the schema looking for them.`
+        );
+      }
       const cols = (err.errorDetails['unknownColumns'] as string[]).join(', ');
       return toStructuredToolError(
         'field_not_found',
-        `Unknown field(s) in this data mart: ${cols}. Call get_data_mart_details_by_id to get this data mart's exact field names (including joined/blended fields) and copy them verbatim into "fields"; never guess or invent field names.`
+        `Unknown field(s) in this data mart: ${cols}. Call get_data_mart_details_by_id to get this data mart's exact field names (including joined/blended fields) and copy them verbatim into "fields"; never guess or invent field names.${hidden.length > 0 ? ` Of these, ${hidden.join(', ')} exist but are hidden from reporting: drop them and do not look for them in the schema.` : ''}`
```
laskevych: 💡 **Suggestion:** Make the MCP message name every place a hidden field can come from, and give it its own `error_code`.

**Why:**
1. When the validator raises the error, `hiddenColumns` names come from `filters`, `slices`, `sort`, `aggregations` or `date_buckets`, not from `fields`. For `fields: ['date']` plus a filter on a hidden `cost`, "Drop them from "fields" and retry" changes nothing, so the retry fails the same way.
2. The branch still returns `field_not_found`, while system instruction 4 says "after field_not_found, call it again with detail_level=with_joined_fields". That is the opposite of "Do not re-fetch the schema".
3. The mixed branch lists the hidden names under "Unknown field(s) … copy them verbatim" before carving them out. `unknown` is computed, but `cols` re-reads the raw `unknownColumns`.

**Fix:**
- Hidden-only: use a dedicated code, e.g. `field_hidden_from_reporting` (like `field_name_reserved` / `calculated_field_not_supported` in this file), and say "remove them from "fields" and from any filter, slice, sort, aggregation or date bucket that names them".
- Mixed: `cols = unknown.join(', ')`, plus a separate "Field(s) … exist but are hidden from reporting" sentence.

**Test:** add a case where the hidden name is only in `filters` or `sort`.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/OutputSettingsButton.tsx
```diff
@@ -35,8 +35,12 @@ export function OutputSettingsButton({
           {typeof count === 'number' && count > 0 && (
             <Badge
               variant={hasDisconnectedControls ? 'destructive' : 'default'}
+              // "Unresolved", not "disconnected": the rule may name a column that is merely
+              // hidden, which the picker's own block right below now says in so many words.
```
laskevych, line 39: 💡 **Suggestion:** Use the hidden wording for rules on a hidden column inside the Output controls panel too. A follow-up is fine.

**Why:**
1. `FilterRow`, `SortRow`, `AggregationRow` and `DateTruncRow` still mark every rule on a hidden column with "Column not found in schema" and "…Remove this rule or restore the column". The dropdown only gets a yes/no orphan state.
2. The amber block above says the same column is "still in the Data Mart", so the two contradict each other on one screen.
3. A sort on an unselected hidden column never appears in either block. The panel row is the only place that names it, and save answers "Hidden columns: …".
4. So this comment holds only for selected and filtered columns.

**Fix:** pass `hiddenColumnNames` to `OutputSettingsDropdown` and use the hidden advice as the orphan title and aria-label for those names.

apps/backend/src/data-marts/dto/domain/blendable-schema.dto.ts
```diff
@@ -236,4 +236,14 @@ export class BlendableSchemaDto {
       'and that list has already had those stripped. A field with no issue is simply absent here.',
   })
   calculatedFieldIssues: CalculatedFieldIssueDto[];
+
+  @ApiProperty({
+    type: [String],
+    description:
+      'Column paths the analyst hid from reporting, as full dotted paths (a hidden RECORD brings ' +
```
laskevych: 💡 **Nit:** Describe the joined names in the `hiddenFieldNames` API description.

**Why:**
1. It says "as full dotted paths", but joined entries use the unified blended name (`orders__internal_id`, with a hash suffix when nested).
2. It says the names are "Absent from `nativeFields`", but per-join hidden names stay in `blendedFields` with `isHidden: true`.
3. This text ships in the OpenAPI spec, and the web and extension types copy it.

**Fix:** "Own columns as dotted paths (`metrics.cost`); joined columns by their unified blended name, the same form as `blendedFields[].name`."

laskevych (changes requested):
Cross-reviewed together with OWOX/owox-data-marts-spreadsheets#66. The own-field path works as described; one blocking gap for joined hidden columns at run time (inline), the rest is non-blocking. On the changeset title: keeping the heading is fine — it matches 10 of the 12 pending entries on `main`.

apps/backend/src/data-marts/services/blended-report-data.service.ts
zapolsky: Right on every point, and the one that mattered most: the set came from the main Data Mart while the names being checked came from both namespaces, so the run kept saying "restore the schema" about a column the picker two clicks away already called hidden.

`blendableSchema.hiddenFieldNames` is now passed in, which covers own paths and joined names at once.

Pinned per switch, because they fail differently and I wanted the difference on the record: a field hidden on the joined Data Mart never becomes a `BlendedFieldDto` at all, one hidden per join arrives carrying `isHidden` — same dead end, two routes. Plus a case for a joined name that is neither (removed join, deleted field), which must stay red.

Producer side too, which had nothing for joined: both switches, and the hashed nested form, since that is the name a report actually stores.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/ReportColumnPicker.tsx
zapolsky: Done, and the reasoning corrected with it — my comment claimed a slice can only be orphaned by the main schema, which is exactly backwards: it is addressed through a join, so the joined side is the only thing that CAN orphan it, and "hidden there" is one of those ways.

Slices now split by `hiddenFieldNames` like the rows do. A joined calculated field stays red: the backend refuses it by name with its own error, whatever anyone hid.

Test added on the existing fixture plus `hiddenFieldNames: ['b__hidden_field']`, expecting the amber block and no red one. The extension has the same change in OWOX/owox-data-marts-spreadsheets#67 — #66 had already landed by then.

docs/getting-started/setup-guide/output-controls.md
zapolsky: All three fixed. The anchor was the giveaway: I linked `#per-field-overrides` while describing the joined Data Mart's own Output Schema — two different switches, and the link pointed at the one the sentence was not about.

The three are now named as they actually appear: the Data Mart's own `Hide from reports` in Data Setup → Output Schema, the same action on the joined Data Mart's own Output Schema, and `Hide from reports` in the ⋯ menu of the join's Report Fields tab, which is the one `#per-field-overrides` documents. The changeset's "**Hidden** in Join Settings" is gone the same way — that control does not exist.

Also: 👁 → eye-off, and "Validation error on save" no longer promises a Disconnected columns error for every stale rule; it names both errors and the mixed case.

apps/backend/src/ee/mcp/tools/query-data-mart.tool.ts
zapolsky: Taken, including the part I had wrong: the branch still returned `field_not_found`, which the system instructions answer with "call it again with detail_level=with_joined_fields" — the same loop I was trying to break, by another route.

It is `field_hidden_from_reporting` now, beside `field_name_reserved` and `calculated_field_not_supported`. The message names every clause the field has to leave, since the validator raises this for a filter, slice, sort, aggregation or date bucket and not only for `fields` — your `fields: ['date']` plus a filter on a hidden `cost` is the case that proved it, and it now has a test.

Mixed branch: `cols` was re-reading the raw `unknownColumns`, so the hidden names were listed under "copy them verbatim" and carved out one sentence later. It reads `unknown` now, with the hidden ones named separately, and the test asserts the list holds only the real unknown.

apps/backend/src/data-marts/dto/domain/blendable-schema.dto.ts
zapolsky: Fixed — both halves were wrong for joined names. It now says: own columns as dotted paths (`metrics.cost`, a hidden RECORD bringing its subtree), joined columns by their unified blended name, the same form as `blendedFields[].name`, hash suffix and all; and it says explicitly that a per-join hidden field stays in `blendedFields` with `isHidden: true` rather than being absent. The web type's copy is updated to match.

apps/web/src/features/data-marts/edit/components/ReportColumnPicker/OutputSettingsButton.tsx
zapolsky: Agreed, and taking you up on the follow-up. Not in this PR: the panel rows read their orphan state from a yes/no flag that predates the split, so giving them the hidden wording means threading `hiddenColumnNames` through `OutputSettingsDropdown` into four row components — a change with its own blast radius, next to a picker change that is already sizeable.

Your point 3 is the part I want recorded, because it is not cosmetic: a sort on an UNSELECTED hidden column appears in neither block, so the panel row is the only surface that names it at all — and it currently says "Column not found in schema" about a column save calls hidden. That is the one a follow-up has to fix, not just the wording on rules already visible above.

laskevych (approved):
All review points are addressed in e83bd039d: the run path reads `hiddenFieldNames` (both joined switches and the still-disconnected case pinned, plus producer specs), slices split in the picker, MCP has its own `field_hidden_from_reporting` code, and the docs, changeset and DTO text are fixed. The Output controls panel wording stays as the agreed follow-up.

PR #1630, opened 2026-09-21
fix(idp): return invalid_grant instead of 500 for expired MCP refresh tokens

amarchenk0:
## Summary
- IB reports an expired/invalid MCP refresh (or authorization-code) grant as a plain 401. ODM wrapped that in `AuthenticationException`, which isn't a Nest `HttpException`, so it fell through the catch-all filter as an unmapped `500` instead of a proper OAuth `invalid_grant` (`400`) — MCP clients couldn't recognize the grant was dead and restart authorization.
- `OAuthTokenController.token()` now catches that specific failure (`isAuthenticationError`, matching by `.name`/`.status` — no runtime import from the ESM-only `@owox/idp-owox-better-auth` package, same convention as the existing `isIdpNotFoundError`) and returns `400 { error: 'invalid_grant', error_description }`.
- Every other failure (timeout, C2C outage, IB 5xx) is untouched and still surfaces as 5xx via the existing fallback.

## Test plan
- [x] `oauth-token.controller.spec.ts`: new cases for an `AuthenticationException`-named error and a plain `status: 401` error → both map to `400 invalid_grant`; existing success and generic-failure-passthrough tests still pass.
- [x] `is-authentication-error.spec.ts`: new unit tests for the matcher.
- [x] `npm run build -w @owox/backend` — clean.
- [ ] Production verification (per ticket DoD): confirm expired refresh-token attempts return `400 invalid_grant` with no corresponding ODM 500s.

Fixes https://owox.fibery.io/Sprint_Work/Work_Item/Return-invalid_grant-for-expired-MCP-OAuth-refresh-tokens-instead-of-HTTP-500-6958

---
_Published without the AGENTS.md-required `GENERATED_FOOTER` trailer: the variable was unset in this interactive session; omission was explicitly requested by the repository owner (Andrii) after being told about the rule._


zapolsky:
⚡ **Light pass — not a full review.** Single-context static read of the changed files; no multi-agent verification. Run `/review-pr` for the deep pass.

✅ No blocking issues found in this pass.

The chain was traced end to end: `handleAxiosError` (`IdentityOwoxClient.ts:762`) sets `name = 'AuthenticationException'` via `BaseException` (`core/exceptions.ts:21,36`), so the `.name` match in `is-authentication-error.ts` holds; `IdpFailedException` is not a Nest `HttpException`, and `GlobalExceptionFilter` (`global-exception.filter.ts:131-135`) maps non-HTTP errors to 500 — the "C2C failures stay 5xx" claim checks out.

## 🔵 Notes (non-blocking)

- `packages/idp-owox-better-auth/src/client/dto/oauthDto.ts:36` — `McpOAuthGrantErrorResponseSchema` pins two exact IB sentences (`'Authentication Error'`, `'Authentication Error, Invalid token'`). Any upstream rewording silently drops the 401 back into `IdpFailedException` → the same 500 this PR removes, with no test or alert catching the regression. A prefix match (`startsWith('Authentication Error')`) or a stable error code would be less brittle. The fail direction is the safe one (old behaviour), so this is a maintenance risk, not a bug.
- **Question:** the split relies on the IB C2C interceptor never emitting a `{ message: 'Authentication Error…' }` body on its 401. That was verified on the ODM side only — the IB service is not in this checkout, so the chain is open there. Was it confirmed against the interceptor's actual response?
- `apps/backend/src/idp/oauth/controllers/oauth-token.controller.ts:34` — the mapping catches *every* `AuthenticationException` reaching the controller, not just the upstream 401 from IB. `owox-token-facade.ts:57` (`Code verifier has expired`) and `services/core/user-context-service.ts:31-62` (`Invalid token payload`, `User not found in Better Auth DB`, `User email is not verified`) now also answer `400 invalid_grant`. That is defensible under RFC 6749, but it is wider than the PR description suggests — worth stating so the behaviour is not a surprise later.
- `apps/backend/src/idp/utils/is-authentication-error.ts` — the sibling `isIdpNotFoundError` also accepts `status === 404` as a fallback for "transport repackaging that drops the original name"; this matcher deliberately does not. The reason is documented, so this is only a flag that the two helpers now differ in robustness.

CI: 20 green · 0 red · 0 pending

<sub>Automated OWOX review · light mode · no subagents.</sub>



🤖 Generated by AI

zapolsky:

The core contract was traced end to end on both sides (ODM + IB `master`):
- IB grant failures on `/internal-api/idp/oauth/token` (bad/expired/reused code, PKCE mismatch, expired/forged/foreign refresh token) are `AuthenticationException`/`InvalidTokenException` → `IpdApiExceptionHandler` → `401 {message:'Authentication Error[, Invalid token]', code:null, details:null}`. The non-strict `z.object` strips `code`/`details`, so these match and become `400 invalid_grant`.
- The C2C interceptor (`InternalAuthorizationInterceptor`) only calls `setStatus(401)` and returns `false`, with no body and no `/error` dispatch. That fails the schema → `IdpFailedException` → `GlobalExceptionFilter` 500, as the PR claims.
- The exchange path does not re-wrap errors: controller → `IdpProviderOAuthAdapter` → `OwoxBetterAuthIdp.exchangeMcpOAuthToken` → `IdentityOwoxClient`. The only `AuthenticationException` on it comes from `handleAxiosError`'s 401 branch. C2C header/ID-token failures, zod errors and the "no provider" error all have different names.
- Replacing `{ req }` with `{ grantType, clientId, resource }` in the exception context stops the refresh token, code and code verifier from being carried in `context`. Good change.

- `packages/idp-owox-better-auth/src/client/dto/oauthDto.ts:38` — **latent: a Google JWKS outage in IB produces the exact `invalid_grant` body.** IB `RsaKeyService.verifyJwt` fetches the JWKS live on every refresh-token check, with no cache (`GcpIamServiceImpl.getServiceAccountJwks` → `IamCredentialsApiService`). Its blanket `catch (Exception e)` turns a fetch failure into `InvalidTokenException("Failed to fetch JWKS")`, i.e. `401 {message:'Authentication Error, Invalid token'}`, which is byte-identical to a genuinely expired token. Today it cannot reach MCP clients only because of timing: IB's `@Retryable(maxAttempts=3, multiplier=2)` sleeps at least 1 s + 2 s before answering, while ODM aborts at `IDP_OWOX_TIMEOUT=3s` (k8s `components/cloud/kustomization.yaml:62`, same code default). Raising that timeout, or shortening IB's retry, would turn a Google blip into forced re-authorization for every connected MCP client, which contradicts the changeset line "Identity Provider outages remain server errors". Suggest: an IB follow-up so JWKS fetch failures map to a 5xx (and/or cache the JWKS), plus a one-line note next to this schema recording the timeout coupling.
- **Revoked grants still return 500 on refresh (scope gap vs. the PR goal).** IB `McpOAuthService.refreshMcpToken` re-resolves claims via `CustomTokenClaimsService`, which throws `AuthorizationException(INACTIVE_USER | REMOVED_PROJECT)`. That is a **403** `{message:'Authentication Error, INACTIVE_USER'}`. ODM maps it to `ForbiddenException`, which is not an `HttpException`, so it becomes 500 (`global-exception.filter.ts`). A connector for a deactivated user or a removed project keeps getting 500 and never restarts authorization, which is the same symptom this PR fixes for expiry (RFC 6749 §5.2 counts "revoked" as `invalid_grant`). The changeset does not claim this case, so the text is accurate. Suggest mapping those two 403 bodies in `handleMcpOAuthTokenExchangeError`, or tracking it as a named follow-up.
- `packages/idp-owox-better-auth/src/client/IdentityOwoxClient.ts:802` — **pre-existing, UNPROVEN at runtime: `cause: error` attaches the raw `AxiosError`.** Its `config.data` holds `refreshToken`/`code`/`codeVerifier`, and `config.headers.Authorization` holds the C2C bearer. On the 500 path, `GlobalExceptionFilter` → `CustomLoggerService.error` spreads the exception into metadata → pino serializes it, and there is no redact config in `internal-helpers`. This PR makes it better by dropping `{ req }` from `context`. The new unknown-401 branch still attaches the raw error, like every `handleAxiosError` branch. Worth a sanitized cause (status + response body only) here or in a follow-up. To settle it, check a staging log entry for `cause.config.data`.

<details>

- `apps/backend/src/idp/utils/is-authentication-error.ts:14` — the controller now relies on the string name of a class private to `@owox/idp-owox-better-auth`, from behind `OAuthIdpPort`. It follows the `isIdpNotFoundError` precedent, and the fixed changeset group rules out deploy skew. But a rename would still pass CI (the spec fakes the name) and silently bring the 500 back. Longer term, an `idp-protocol` error (e.g. `OAuthInvalidGrantError` documented on `exchangeMcpOAuthToken`) with an `instanceof` check would remove the implicit contract.
- `packages/idp-owox-better-auth/src/client/dto/oauthDto.ts:39` — matching on free-text reasons is brittle. The failure direction is safe (it falls back to 500), and IB's `PublicApiException` already has an unused `code` field, so a stable code such as `MCP_INVALID_GRANT` would be the durable contract.
- `apps/backend/src/idp/oauth/controllers/oauth-token.controller.ts:35` — the `invalid_grant` path leaves no reason in ODM logs: no `cause`, and the filter logs only "Handled HTTP 400". One sanitized `logger.log` with `grantType`, `clientId` and IB's `message` (never the token) would make a burst of rejections diagnosable.
- `packages/idp-owox-better-auth/src/client/IdentityOwoxClient.ts:801` — the contract-drift 401 branch drops `responseData`, which the default branch keeps. The unexpected body is exactly what an operator needs to see here.
- `apps/backend/src/idp/oauth/oauth-request.validator.ts` — pre-existing: other `/oauth/token` failures (`unknown client_id`, `unsupported grant_type`, …) are still Nest `BadRequestException` bodies with no RFC `error` code. The endpoint now returns two error shapes. An `oauthError(code, description)` helper would let a follow-up align them.
- minor: test fixtures are slightly narrower than the wire: the C2C case uses `data: undefined` where axios yields `''`, and the grant case omits `code:null, details:null`. The `narrows the type` test in `is-authentication-error.spec.ts` asserts nothing if the guard returns false (no `expect.assertions(1)`). It is redundant with the preceding test.

</details>

**On the earlier ⚡ light-pass notes:**
- The open question about the C2C 401 body is now answered from IB source: the interceptor emits an empty 401, and it cannot match the schema.
- **Correction:** the light-pass note claiming that `owox-token-facade.ts` ("Code verifier has expired") and `user-context-service.ts` `AuthenticationException`s "now also answer `400 invalid_grant`" is wrong. Those throw sites are not in the `exchangeToken` call graph (controller → adapter → provider → `IdentityOwoxClient`); only `handleAxiosError`'s 401 branch produces that name there. Please disregard it.

CI: 20 green · 0 red · 0 pending (read from `gh pr checks`; nothing executed locally).

<sub>Automated OWOX review · worktree multi-agent passes · all blockers adversarially verified (1 candidate refuted: the JWKS path is latent under the current timeouts).</sub>



🤖 Generated by AI

zapolsky (approved):

PR #1631, opened 2026-09-21
feat(idp): add RFC 9207 issuer to project-specific MCP OAuth callbacks

amarchenk0:
## Summary
- Project-specific MCP OAuth (`https://{projectId}.mcp.owox.com`) intentionally uses a different origin for its issuer than for `authorization_endpoint` (`app.owox.com`), to keep dynamic client registration bound to the right project resource. OAuth clients that enforce RFC 9207 issuer-bound callbacks (observed: Codex) refuse this split-origin metadata outright and never start the browser flow — discovery stops right after Protected Resource + AS Metadata.
- `OAuthMetadataController` now advertises `authorization_response_iss_parameter_supported: true`, and `OAuthAuthorizationController`'s success redirect now carries `iss`, both scoped to project-specific issuers only. The shared MCP host is untouched (its issuer and `authorization_endpoint` already share one origin, so it has nothing to prove).
- **Out of scope, follow-up needed:** `/oauth/authorize` has no error-redirect mechanism at all today — every failure is a direct JSON response on ODM's own origin, never a redirect to the client's `redirect_uri`. RFC 9207 compliance for error responses specifically would require building that mechanism from scratch (and doing it safely — only after `redirect_uri` is validated, to avoid an open-redirect). The bug this PR fixes happens during metadata discovery, before `/oauth/authorize` is ever reached, so the success-path fix alone unblocks it; the error-redirect gap is a separate, deliberately unscoped follow-up.

## Test plan
- [x] `oauth-metadata.controller.spec.ts`: project-host metadata now asserts the new flag; shared-host metadata asserts it's absent.
- [x] `oauth-authorization.controller.spec.ts`: project-flow redirect now asserts `iss`; both existing shared-flow redirect tests are unchanged and act as a regression guard (no `iss` for shared).
- [x] `npm run build -w @owox/backend` — clean.
- [ ] Production verification (per ticket DoD): Codex completes authentication for a project-specific MCP URL end to end.

Fixes https://owox.fibery.io/Sprint_Work/Work_Item/Support-RFC-9207-issuer-bound-OAuth-callbacks-for-project-specific-MCP-URLs-6961

---
_Published without the AGENTS.md-required `GENERATED_FOOTER` trailer: the variable was unset in this interactive session; omission was explicitly requested by the repository owner (Andrii) after being told about the rule._


zapolsky:
✅ No blocking issues found in this pass.

**Checked, chains close.** The security-critical part is the new error redirect, and it is correctly gated: `resolveAuthorizationErrorContext` (`oauth-request.validator.ts:70`) refuses to build a context unless the `client_id` resolves in the registry **and** `client.redirectUris.includes(redirectUri)` is an exact match (`:80`), returning `null` otherwise so `authorize` rethrows and no redirect happens (`oauth-authorization.controller.ts:63-66`) — `oauth-request.validator.spec.ts:215` and `oauth-authorization.controller.spec.ts:372` pin both halves. The post-validation path reuses `validated.request.redirectUri`, which `validateAuthorizationRequest` already checked against the same registry list (`oauth-request.validator.ts:52`), so neither branch is an open redirect. Attacker-controlled `state` goes through `URLSearchParams.set`, so it cannot break out of the URL.

The `iss` value is consistent with the advertised `issuer`: both come from the same `McpResourceContext.publicBaseUrl` — the success path at `oauth-authorization.controller.ts:142`, the error path at `:160`, and in both cases the context comes from the registered `client.resource`, since `resolveClientResource` (`oauth-request.validator.ts:175`) ignores the request-supplied `resource` for this purpose. A strict RFC 9207 client therefore sees the same string it discovered. `ViewOnlyModeError extends AuthorizationError` (`packages/idp-protocol/src/types/errors.ts:50`), so a view-only session maps to `access_denied`, not `server_error`. The metadata refactor is behaviour-preserving: inlining `resolveIssuer` yields the same `issuer` for both the AS-metadata and the openid-configuration paths, and the new flag is gated on `kind === 'project'` exactly as the redirect is.

## 🔵 Notes (non-blocking)
- **The PR description contradicts the head commit.** The body states the error-redirect mechanism is "Out of scope, follow-up needed … would require building that mechanism from scratch", but `d2a83c4 fix(idp): add issuer-bound OAuth error callbacks` builds exactly that, and it is the larger and more security-sensitive half of the diff (~110 of the 136 added lines in `oauth-authorization.controller.ts`). A reviewer who trusts the description will not look at it. Worth rewriting the body before merge.
- `.changeset/6961-mcp-oauth-project-issuer.md` — the entry only promises that Codex can now sign in against project MCP URLs, but the error path changes behaviour for **every** OAuth client, shared host included: a failure after `redirect_uri` is trusted no longer renders a JSON error on ODM's own origin, it redirects the browser back to the client with `error=…` (`oauth-authorization.controller.ts:147`). Only the `iss` parameter itself is project-gated (`:160`). That is an operator-visible change worth its own sentence in the changeset.
- `oauth-authorization.controller.ts:192` — `mapAuthorizationError` folds everything that is neither `AuthorizationError` nor `BadRequestException` into `server_error`, so an IdP outage and a genuine internal fault reach the client identically. Fine for the client contract; just note that `error_description` is a fixed string, so the server log (`:169` / `:175`, carrying `clientId` and `issuer`, stack only on `server_error`) is the only place the real cause survives.

CI: 20 green · 0 red · 0 pending.

<sub>Automated OWOX review · light mode · no subagents.</sub>



🤖 Generated by AI

zapolsky:
✅ No blocking issues found. Looks ready to approve (a maintainer should give the formal approval).

**Checked, and the chains close:**
- **No open redirect.** An error redirect only happens once `client_id` resolves in the registry and `redirect_uri` exactly matches a registered URI (`oauth-request.validator.ts:79-80`). Otherwise the original error is rethrown (`oauth-authorization.controller.ts:63-66`). Dynamic registration only accepts loopback `http` or allowlisted `https` origins (`oauth-dynamic-client.service.ts:104-133`).
- **`iss` is byte-identical to the discovered `issuer`.** On the metadata side, `tryResolveRequest` → `resolveResource` produces `${url.protocol}//${url.host}` (`mcp-resource-resolver.service.ts:55,60-69`). On the callback side, the client's `resource` is bound at `/oauth/register` on the same host and then goes through the same `resolveResource` (`oauth-request.validator.ts:86,188`). Both go through WHATWG URL normalisation.
- **Every project-issuer authorization response carries `iss`:** the success redirect (`:141-143`), post-validation errors (`:77` → `:159-161`), and pre-validation errors with a trusted context (`:67`). The project flow never reaches the project-selection page.
- **View-only and API-key sessions map to `access_denied`,** because `ViewOnlyModeError extends AuthorizationError` (`packages/idp-protocol/src/types/errors.ts:50`). The `headersSent` guard at `:74` is sound: the sign-in redirect and the selection page return `null` instead of throwing.
- **The metadata refactor preserves behaviour.** The flag is gated on `kind === 'project'`, exactly like the redirect `iss`.

- `apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts:192` — **Routine permission denials surface as `server_error` and are logged at ERROR level with a stack trace.** `mapAuthorizationError` only recognises `AuthorizationError` → `access_denied` and `BadRequestException` → `invalid_request`. The ordinary "you can't use this project" paths throw other types, so they all fall through to `server_error` and hit `logger.error` (`:168-173`):
  - Nest `UnauthorizedException` from `oauth-project-member.resolver.ts:9,12`, reached on the shared flow via `:258`.
  - Nest `UnauthorizedException` from `oauth-project-selection.service.ts:57,62` ("User must be an active project member" / "unsupported project role"). Both flows reach it; the project host goes through `resolveProjectHostMember` → `:88`.
  - The IdP's own `ForbiddenException('Identity inactive or blocked')` on an IB 403 (`packages/idp-owox-better-auth/src/client/IdentityOwoxClient.ts:769-774`), reached via `createAuthorizationCode`.

  Result: a non-member opening a project MCP URL in Codex sees "the authorization server encountered an unexpected condition", and on-call gets error-level noise for a normal denial. The changeset's "Unexpected internal failures use `server_error`" is not accurate for these cases. The reverse also happens: "project … is not available for user" (`oauth-project-selection.service.ts:40,43,85`) is a `BadRequestException`, so it becomes `invalid_request` rather than `access_denied`. Fix: have the resolver and selection service throw `AuthorizationError` for membership denials, or at least map `UnauthorizedException` / `ForbiddenException` to `access_denied` with a warn-level log. Decide explicitly how to map the IB 403. No spec currently exercises any of these.
- `apps/backend/src/idp/oauth/controllers/oauth-authorization.controller.ts:175` — **The reason for `invalid_request` / `access_denied` is no longer recorded anywhere.** `error_description` is a fixed string, and the warn log carries only `{clientId, error, issuer}`. Before this PR, these errors went through `GlobalExceptionFilter` / `IdpExceptionFilter`, and the specific message ("unsupported scope: x", "selected project conflicts with MCP resource", "code_challenge_method must be S256", the view-only text) was in the JSON body and in the logs. Now roughly ten distinct causes are indistinguishable when supporting an MCP client integration. Fix: add `reason: castError(cause).message` to the warn metadata. These messages are authored strings with no tokens; the only interpolation is the client-supplied scope. Also use a distinct log message for the pre-validation path (`:67`), which is currently logged as "rejected after request validation".
- **The PR description still contradicts the code** (carried over from the previous pass, still open). The body says `/oauth/authorize` "has no error-redirect mechanism" and calls it a "deliberately unscoped follow-up", and the test plan lists only success-path tests. But `d2a83c4` builds exactly that mechanism, and it is the larger and more security-sensitive half of the diff. The changeset now describes it correctly; please update the body before merge.

<details>
<summary>Additional observations (below the ≥80 bar; non-blocking, listed for completeness)</summary>

- `oauth-authorization.controller.ts:199` — All `BadRequestException`s become `invalid_request`. RFC 6749 §4.1.2.1 defines `unsupported_response_type` (validator `:42`) and `invalid_scope` (`:154,158`), and RFC 8707 defines `invalid_target` for a resource mismatch (`:192,196`). Clients that branch on the code will misclassify these. The fix is typed validator errors that carry the OAuth code.
- `oauth-request.validator.ts:70` — `resolveAuthorizationErrorContext` re-implements the "trusted `redirect_uri`" check that `validateAuthorizationRequest` does (`:49-54`), and fetches the client a second time on failure. Two parallel definitions of a security decision can drift apart. Consider validating `client_id` + `redirect_uri` first (the order RFC 6749 prescribes) and throwing a typed error that carries the trusted context. That error-mapping, redirect and logging logic could then live in a small injectable next to `OAuthProjectSelectionService` instead of growing the ~390-line controller.
- **Rollout:** during a rolling deploy, a client can read metadata from a new pod (flag advertised) and have its callback served by an old pod (no `iss`). A strict RFC 9207 client rejects that. It is short-lived and only on project hosts, and not a regression for Codex (it fails today), but worth knowing if Claude or ChatGPT enforce RFC 9207.
- `oauth-authorization.controller.ts:63` — If `resolveAuthorizationErrorContext` itself throws (e.g. a DB error from `clientRegistry.get`, validator `:79`), that exception replaces the original validation error.
- `oauth-metadata.controller.ts:54-57` (and the matching comment at `:138-140` in the authorization controller and in the spec) — "the shared MCP host's issuer and `authorization_endpoint` already share one origin" holds for the discovery path clients actually follow: shared protected-resource metadata points `authorization_servers` at `OWOX_AUTH_PUBLIC_BASE_URL` (`ee/mcp/controllers/mcp-metadata.controller.ts:38-45`). It is not true of AS metadata fetched straight from `mcp.owox.com`, which the PR's own "split production config" spec shows as `issuer: https://mcp.owox.com` + `authorization_endpoint: https://app.owox.com/...`. The real reason `iss` can't be added for shared clients is that both hosts register `kind: 'shared'` clients, so `/authorize` can't know which issuer the client saw. Worth stating that in the comment.
- minor: the log field `issuer` (`:166`) is `MCP_PUBLIC_BASE_URL` for shared clients, not the issuer they discovered (rename to `resourceBaseUrl`) · `state` is echoed trimmed via `optionalString` / `requiredString`, whereas RFC 6749 wants the exact value (pre-existing on the success path, now mirrored) · `getSelectedProjectId` duplicates the new `optionalString`.

</details>

**Since the previous (light) pass:** the changeset note about the operator-visible error-redirect change is **addressed** in `e3e97a9`.

CI: 20 green · 0 red · 0 pending.

<sub>Automated OWOX review · worktree multi-agent passes (reviewer, logic-hunter, architect, tester, fresh-eyes) · no blocker candidates, so no skeptic gate was needed.</sub>



🤖 Generated by AI

zapolsky (approved):

PR #1632, opened 2026-09-21
feat(data-marts): resume interrupted backfills from the last loaded day

prokopetsv:
# Problem

A manual backfill that stopped before finishing restarted from its `StartDate` on the next attempt, re-importing every day it had already loaded. Users saw the run fail with `Connector process finished without terminal success status`, most often when a deploy or pod restart killed the connector process mid-run. Incremental runs already resumed, because every time-series connector emits `updateLastRequstedDate` after each fully imported day and the backend stores it as the incremental cursor. Backfills deliberately suppressed that message behind an `if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL)` guard in each connector, so a backfill had no record of progress at all. The interrupted-run sweep then replayed the original payload unchanged, and a long backfill could burn its whole API budget repeating work.

# Solution

Connectors now emit the per-day message for both run types, and the backend decides where it belongs. The routing key is the run type the backend itself built and shipped to the child process in `OW_RUN_CONFIG`, so the two sides cannot disagree about which run produced a date.

- **Connectors**: removed the incremental-only guard at 11 checkpoint sites. The invariant that a date is only reported once it is complete for every node and account already held there, and is documented in each file.
- **Backend routing** (`connector-executor.service.ts`): an incremental run still writes the `ConnectorState` cursor exactly as before; a backfill records `additionalParams.backfillProgress[configId]` on the `DataMartRun` row instead. A backfill therefore never moves the incremental cursor, which would otherwise rewind it to the backfill's period and make the next scheduled run skip everything imported since.
- **Durability**: progress writes are chained and awaited in each configuration's `finally`, so the run's terminal status is never written ahead of the checkpoint a retry must start from. They use a targeted column `update` rather than `save`, because the terminal status write and cancellation touch the same row concurrently. A failed write is logged and swallowed, since a lost checkpoint only costs an idempotent re-import.
- **Resume** (`manual-backfill-range.ts`): `resumeManualBackfillPayload` shifts `StartDate` to the day after the last loaded one, clamped to `EndDate`. The resumed range is always a subset of an already-validated period, so it can never exceed the 31-day cap. Stale progress from an earlier period, and malformed dates, are ignored rather than allowed to widen the range.
- **Excluded connectors**: Shopify and TikTok Ads keep the old guard. Shopify walks the full date range per node with the node loop outside it, so a date finished for `orders` says nothing about `customers`. TikTok Ads swallows every fetch and storage error and reports failures only after the whole range, so a date that imported nothing would still be recorded. In both cases a resumed backfill would have skipped days that were never stored, so they continue to reload the whole period.

Re-running a day is safe because every storage flushes its buffer with a `MERGE` on unique keys before `saveData` returns, so the checkpoint is only emitted once the write is durable. No migration, no API change, and no new environment variable: the progress lives in an existing JSON column that the run response already masks for connector runs.

# Changes

- Removed the incremental-only guard around `updateLastRequstedDate` in 11 connectors, so manual backfills report each completed day.
- Kept the guard in Shopify and TikTok Ads, whose loop structure and error handling make a backfill checkpoint unsafe, and documented why at each call site.
- Added `unwrapRunPayload`, `isManualBackfillPayload`, `parseBackfillDay`, `readBackfillProgress` and `resumeManualBackfillPayload` to `manual-backfill-range.ts`.
- Routed `REQUESTED_DATE` messages by run type in `ConnectorExecutorService`, persisting backfill progress to `DataMartRun.additionalParams.backfillProgress` through a serialised, awaited promise chain.
- Shifted a resumed backfill's `StartDate` to the last loaded day plus one, clamped to `EndDate`, and recorded the reason as a run log so Run History explains the shorter period.
- Moved the run-payload unwrap out of `ConnectorSourceConfigService.buildRunConfig` into a shared helper so both callers handle the replayed and first-attempt shapes identically.
- Added `perDayCheckpoint` suites for Shopify and TikTok Ads pinning that neither records backfill progress.
- Updated the six existing `perDayCheckpoint` suites to assert that a backfill now reports each completed day, plus a partial-progress case for GoogleAds.
- Added backend coverage for progress persistence, cursor isolation, checkpoint ordering before the terminal status write, multi-configuration progress, resume clamping and write-failure tolerance.
- Documented the behaviour and its two exceptions in the connector setup guide.
- Added changeset `6888-resume-manual-backfill.md`.


prokopetsv (commented):
Review withdrawn.

.changeset/6888-resume-manual-backfill.md
```diff
@@ -0,0 +1,19 @@
+---
+'owox': minor
+---
+
+# Resume a manual backfill from the last fully loaded date
+
+A manual backfill that stops before it finishes now continues from the day after the last one it fully
```
laskevych: 🟡 Important (non-blocking): Narrow the wording to runs that are interrupted and retried automatically. The same applies to the new paragraph in `docs/getting-started/setup-guide/connector-data-mart.md`.

**Why:** Progress is stored on the `DataMartRun` row and only the interrupted-run sweep replays that row. A backfill that ends `FAILED` (or is cancelled) and is started again by the user creates a new run with `additionalParams = { payload }`, so it reloads the whole period. "A manual backfill that stops before it finishes now continues…" and "If a backfill stops early, the next attempt resumes…" promise more than that.

**Fix:** Say the resume applies when a deploy or restart interrupts the run and the automatic retry picks it up; a backfill started again by hand loads the full period.

apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
```diff
@@ -1235,4 +1235,250 @@ describe('ConnectorExecutorService', () => {
...
+      );
+
+      const updates = progressUpdates(dataMartRunRepository);
+      expect(updates.at(-1)?.[1].additionalParams).toEqual({
+        payload: BACKFILL,
+        backfillProgress: { 'cfg-1': '2026-08-20', 'cfg-2': '2026-08-05' },
+      });
+    });
+
+    it('records the last checkpoint before the run reaches a terminal status', async () => {
```
laskevych: 💡 Suggestion: Make this test fail when `await progressWrite` is removed.

**Why:** The `update` mock resolves immediately, and the chained `.then` calls it in a microtask long before `updateRunStatus`. `invocationCallOrder` stays the same with or without the `await` in `finally`, so the durability guarantee this PR adds is not actually pinned.

**Fix:** Return a deferred promise from the progress `update`, assert that no terminal-status `update` has been issued while it is pending, then resolve it and assert the terminal write follows.

packages/connectors/src/Sources/GoogleAds/Connector.js
```diff
@@ -89,10 +89,8 @@ var GoogleAdsConnector = class GoogleAdsConnector extends AbstractConnector {
         }
       }
 
-      // Every customer and node stored this date, so the cursor can move past it.
-      if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
-        this.config.updateLastRequstedDate(currentDate);
-      }
+      // Every customer and node stored this date, so the checkpoint can move past it.
```
laskevych, line 92: 💡 Suggestion: Update "Saving incremental progress" in `packages/connectors/CONTRIBUTING.md` with the new contract.

**Why:** The section still describes `updateLastRequstedDate` as incremental-only. It is now also the backfill resume checkpoint, and it is only safe when the date loop encloses every node and account and no per-day error is swallowed. Those are exactly the reasons Shopify and TikTok Ads are excluded. A new connector that misses them gets silent gaps after a resumed backfill.

**Fix:** Add those two preconditions and point to Shopify / TikTok Ads as the "keep the guard" examples.

packages/connectors/src/Sources/FacebookMarketing/Connector.js
```diff
@@ -248,13 +248,11 @@ var FacebookMarketingConnector = class FacebookMarketingConnector extends Abstra
...
 
-        // Runs before the cursor moves: a day nobody could import must be requested again
+        // Runs before the checkpoint moves: a day nobody could import must be requested again
         this._throwIfAllAccountsSkipped(accountsIds, skipped, `for ${DateUtils.formatDate(startDate)}`);
 
-        // Only update LastRequestedDate for incremental runs
-        if (this.runConfig.type === RUN_CONFIG_TYPE.INCREMENTAL) {
-          this.config.updateLastRequstedDate(startDate);
-        }
+        // Every account stored this date, so the checkpoint can move past it.
```
laskevych: 🧹 Nit: Reword "Every account stored this date".

**Why:** Accounts the token cannot access are skipped on purpose, and the loop throws only when all of them are skipped. The checkpoint can therefore move while some accounts have no data for that day. The comment states an invariant the code does not hold, and the resume safety argument rests on it.

**Fix:** For example: "At least one account stored this date and the rest were skipped deliberately, so the checkpoint can move past it."

apps/backend/src/data-marts/services/connector/connector-executor.service.ts
```diff
@@ -487,7 +540,26 @@ export class ConnectorExecutorService {
           storage: await this.storageConfigService.buildStorageConfig(dataMart),
         });
 
-        const runConfig = this.sourceConfigService.buildRunConfig(payload, configState);
+        if (resume.resumedFrom) {
+          // Recorded as a run log, not only through the application logger: run history is
+          // where someone looks to understand why a retry requested fewer days.
+          const resumeMessage = `Resuming manual backfill from ${resume.resumedFrom}; days through ${resume.lastLoadedDate} were loaded by a previous attempt`;
```
laskevych: 🧹 Nit: Special-case the message for the clamped resume.

**Why:** When a previous attempt already reached `EndDate` (configuration 1 finished, configuration 2 was interrupted), Run History shows "Resuming manual backfill from 2026-08-20; days through 2026-08-20 were loaded by a previous attempt", which reads like a bug.

**Fix:** When `resumedFrom === lastLoadedDate`, log something like "The whole period was loaded by a previous attempt; re-checking the last day".

laskevych (approved):
Approving: routing by run type, checkpoint durability and the Shopify / TikTok Ads exclusions all hold up. Five non-blocking notes inline.

.changeset/6888-resume-manual-backfill.md
prokopetsv: Agreed, and this was the most important of the five. Progress lives on the `DataMartRun` row, and only `executeInterruptedRuns` replays that row, so a `FAILED` or cancelled backfill started again by hand creates a fresh run with `additionalParams = { payload }` and reloads everything. The old wording promised resume for that case too.

Both texts now name the automatic retry explicitly and state the other case rather than leaving it implied:

- changeset: "When a deploy or a restart interrupts a manual backfill, the automatic retry now continues from the day after the last one it fully loaded" plus "Starting a backfill yourself always loads the whole period you choose, including after one that failed. Only the automatic retry of an interrupted run resumes."
- setup guide: same split, within the 20-word sentence rule.

Fixed in fe64a92e5.

apps/backend/src/data-marts/services/connector/connector-executor.service.spec.ts
prokopetsv: You are right, and I confirmed it the direct way before fixing: I deleted `await progressWrite` from the configuration's `finally` and the test still passed. It pinned nothing. There are several awaits between the checkpoint and `updateRunStatus`, so the immediately-resolved `.then` always drained first regardless.

Rewrote it as you suggested. The progress `update` now returns a promise the test holds open, and the assertions are:

1. after letting every unrelated microtask settle, no terminal-status `update` has been issued;
2. release the checkpoint, await the run, and a terminal write has now happened.

Verified it fails with the `await` removed and passes with it, so the guarantee is genuinely pinned. Renamed to `waits for an in-flight checkpoint before writing the terminal status`.

Fixed in fe64a92e5.

packages/connectors/src/Sources/GoogleAds/Connector.js
prokopetsv: Valid, and this is the one that protects future connectors rather than this PR. The section described the call as incremental-only and mentioned accounts but not nodes.

Rewrote it to state the new contract and, more importantly, the two conditions that make an unconditional call safe:

- the date loop encloses every node and every account;
- nothing between the fetch and the call swallows an error, so a date that imported nothing stops the run instead of reaching the checkpoint.

It then says to keep the `RUN_CONFIG_TYPE.INCREMENTAL` guard when either fails, and names `Sources/Shopify` as the first failure and `Sources/TikTokAds` as the second, noting both are pinned by a `perDayCheckpoint` test.

Fixed in fe64a92e5.

packages/connectors/src/Sources/FacebookMarketing/Connector.js
prokopetsv: Correct, the comment claimed an invariant the code does not hold. `_throwIfAllAccountsSkipped` only throws when every account was skipped, so the checkpoint can move while some accounts stored nothing for that day.

Worth separating the two things, because the safety argument survives even though the comment was wrong: `_skipOrRethrow` re-throws anything without `isWarning`, so the only skipped accounts are ones the token cannot reach, and those stay unreachable however often they are retried. Re-fetching them on resume would recover nothing. So the checkpoint moving is still right, the sentence justifying it was not.

Reworded close to your suggestion:

```js
// At least one account stored this date and any others were skipped deliberately,
// as accounts the token cannot reach stay unreachable however often they are retried.
// A date no account could import throws above, so the checkpoint can move past it.
```

Fixed in fe64a92e5.

apps/backend/src/data-marts/services/connector/connector-executor.service.ts
prokopetsv: Agreed, that message reads like a defect. It is reachable exactly as you describe: a previous attempt that reached `EndDate` leaves `resumedFrom` clamped onto `lastLoadedDate`, so both halves of the sentence name the same day.

Split the message on `resume.resumedFrom === resume.lastLoadedDate`:

- clamped: "A previous attempt loaded this manual backfill through 2026-08-20, the end of the period; re-checking that day"
- normal: unchanged.

Kept the re-check itself rather than skipping the configuration, for the reason already in `resumeManualBackfillPayload`: skipping would need a second path reporting success without the connector confirming its own field and catalog work, and one already-stored day is idempotent.

Fixed in fe64a92e5.

zapolsky:

The mechanism is sound, and the three load-bearing claims were attacked directly and survived: every storage merges on unique keys inside `saveData` before it returns (BigQuery `:511-531`, Snowflake `:499-517`, Databricks `:536-567`, Athena `:564-611`, Redshift `:823-890`), so the clamped re-run of one day really is idempotent; the run type is decided once from the body actually shipped to the child process; and the eleven de-guarded connectors were checked line by line, not taken from the summary. The **cross-package version-skew** hazard — a newer `@owox/connectors` emitting backfill dates to an older backend that would route them into the incremental cursor — was the strongest candidate blocker and is **dead**: `tools/update-package-versions-for-publishing.mjs:161-173` rewrites every internal `"*"` to an exact version at publish time (`.github/workflows/publish.yml:66`), and `.changeset/config.json` keeps these packages in one `fixed` group, so `@owox/backend@X` pins `@owox/connectors@X` exactly.

laskevych's five notes are all addressed at this head and are not repeated below. One of them is worth closing explicitly: the rewritten checkpoint-ordering test **does** now pin the guarantee. The `update` mock only blocks when the write carries `additionalParams`, and the terminal write never does; the whole path from the configuration `finally` to `updateRunStatus` is microtask-only (no timer, no real I/O, no fake timers in the backend jest config), so `setImmediate` fires strictly after it drains. Deleting `await progressWrite` would make `expect(terminalWrites()).toHaveLength(0)` fail.

- `packages/connectors/src/Sources/BankOfCanada/Connector.js:78` — **three of the eleven de-guarded connectors gain nothing from this change, and the docs promise them something they structurally cannot do.** BankOfCanada (`:51-79`), OpenHolidays (`:52-79`) and LinkedInPages (`:39-62`) fetch the *entire* period in one call and checkpoint once, after it. An interruption before the range completes records nothing; the only value they can ever store is `EndDate` (LinkedInPages: `actualEndDate = StartDate + daysToFetch - 1`, `:145-146`), which `manual-backfill-range.ts:210` always clamps to the one-day re-check branch. So their retry can never resume mid-period — yet `.changeset/6888-resume-manual-backfill.md` and `docs/getting-started/setup-guide/connector-data-mart.md:143-147` say the automatic retry "continues from the day after the last one it fully loaded" and name only Shopify and TikTok Ads as exceptions. That is a second, independent overpromise in the sentence already narrowed once. Since de-guarding buys these three nothing while adding the trap in the next note, the cheapest fix is to keep their guard and list them with Shopify/TikTok — or, if they stay de-guarded, add them to the exception wording and change their comment to say the checkpoint is always the end of the range.
- `apps/backend/src/data-marts/utils/manual-backfill-range.ts:204` — **`resumeManualBackfillPayload` guards only the lower bound**, so a checkpoint *later* than `EndDate` passes and `:210`'s `Math.min(lastLoaded + DAY_MS, end)` collapses the retry to a single day that then reports SUCCESS — the exact failure the function's own doc block (`:181-190`) says it prevents. Not reachable today (all 13 connectors emit inside `[StartDate, EndDate]`), but `LinkedInPages/Connector.js:150-151` already computes a sibling `endDate = actualEndDate + 1 day` for the LinkedIn API right next to the value that gets checkpointed. Fix: `if (lastLoaded < start || lastLoaded > end) return { body };`.
- `packages/connectors/src/Sources/LinkedInPages/Connector.js:57` — **the "single time-series node" invariant that four connectors now depend on is enforced only by prose, and the warning sits in the wrong file.** `CONTRIBUTING.md:77-84` states two permissible shapes and names the disqualifying one by example ("`Sources/Shopify` breaks the first rule, because its date loop sits inside the node loop") — yet `OpenExchangeRates` (`startImportProcess:28-33` loops nodes, day loop at `:65` inside `processTimeSeriesNode`) is structurally identical to that counter-example and is de-guarded in the same PR, while BankOfCanada/OpenHolidays/LinkedInPages have no date loop at all. All four are safe only because each declares exactly one time-series node — a corollary that appears nowhere in the docs (`grep -rin "single time-series"` over `docs/`, `CONTRIBUTING.md`, `README.md` finds nothing) and is checked by no test: nothing in the backend reads `isTimeSeries`, and the storages merge on unique keys, so absent rows are invisible. **LinkedInPages is one schema edit away**: add a second `isTimeSeries: true` node and an interrupted backfill checkpoints `EndDate` after the first node, the retry gives the second node only the final day, every intermediate day is lost, and the run reports success while Run History prints "A previous attempt loaded this manual backfill through {EndDate}" — which is false for that node. The edit that breaks it happens in `LinkedInPagesAPIReference/LinkedInPagesFieldsSchema.js`, which carries no warning at all. Three fixes, cheapest first: put the warning (or a pointer) next to the `isTimeSeries` flags in the schema file; add the third permissible shape to `CONTRIBUTING.md:77`; add a `perDayCheckpoint` test asserting the schema declares exactly one time-series node. Note the count: of the 11 de-guarded connectors, **5 have no `perDayCheckpoint` test at all** — LinkedInPages, BankOfCanada, OpenExchangeRates, OpenHolidays and FacebookMarketing.
- `apps/backend/src/data-marts/services/connector/connector-executor.service.ts:442` — **the checkpoint is "last reported", not "furthest reached".** `backfillProgress = { ...backfillProgress, [configId]: completedDay }` overwrites unconditionally, while the spec that covers it is named `keeps every completed day, so the run reflects the furthest one reached` (`connector-executor.service.spec.ts:1305`). Monotonicity holds today only by accident of every caller. Fix: keep the later of the two (lexicographic compare is safe for `YYYY-MM-DD`), or rename the test to what the code does.
- `apps/backend/src/data-marts/services/connector/connector-source-config.service.ts:65` — **an unrecognised or absent `runType` silently becomes `INCREMENTAL`**, and that is the one hop that would turn a lost `additionalParams.payload` into precisely the incident this PR exists to prevent: `isBackfill` false → the retry's dates go into the real `ConnectorState` cursor. The chain is broken at this head (`connector-run.service.ts:176` always sets `{ payload }` for a backfill, and the entity is re-loaded at claim time), so this is about the failure *direction*, not a live bug. Compare `manual-backfill-range.ts:11-19`, where the same kind of contract assumption is deliberately made to fail loudly. Fix: throw or log at error level on an unrecognised `runType`, and keep `payload` explicitly in `persistBackfillProgress` rather than relying on the spread.
- `packages/connectors/src/Sources/LinkedInAds/Connector.js:86` — **a resumed backfill silently drops the analytics-truncation warning for every day it skips.** `reportTruncatedAnalytics()` runs after the date loop and reads `this.source.truncatedAnalyticsDays`, which lives in the child process and dies with it; an interrupted attempt never emits it, and the resumed attempt never re-walks those days. Before this PR the whole range was re-walked on every retry, so the warning was always regenerated. The rows are identical either way — this loses a signal, not data. FacebookMarketing has the same end-of-run shape but is unaffected, because its per-skip warning is already emitted inline at `:105`. Fix: emit the truncation warning per day inside the loop, or document LinkedInAds as a third caveat in `CONTRIBUTING.md:70-86`.

## 📝 On comment density (raised separately by the repo owner)

Measured rather than eyeballed — comment lines ÷ (comment + code), whole files before vs after this PR:

| | before | after | house baseline |
|---|---|---|---|
| all 15 `Sources/*/Connector.js` | 34% | **35%** | 34% |
| `connector-executor.service.ts` | 10% | **12%** | `data-marts/services/**` = 12% |
| `manual-backfill-range.ts` | 19% | **31%** | `data-marts/utils/*.ts` = 19% |

No connector file moved by more than +3 points, and the executor landed exactly on its folder's baseline. The 79%-comment figure the diff suggests for connectors is an artifact: those hunks mostly *delete* the guard and leave an explanation behind. **The density is not out of line with this repo — with one exception:** `manual-backfill-range.ts` was precisely on the `utils/` norm before this PR and is now ~27 comment lines above it.

What does read as noise is **repetition**, not volume. Of the 41 comment lines added under `packages/connectors/`, 16 are reworded existing lines (`cursor` → `checkpoint`) and 25 are new; of those 25, **9** are the Shopify/TikTok Ads exclusion blocks, **4** are the FacebookMarketing rewrite, and the remaining **12 are one three-line paragraph copy-pasted into four files** — while this same PR gives that rule a canonical home at `CONTRIBUTING.md:70-86`.

- **Keep:** `Shopify/Connector.js:84-87` and `TikTokAds/Connector.js:208-212` — the best comments here; each states a verifiable structural reason *and* a restore condition, and each is pinned by a test. Also `connector-executor.service.ts:327-328` (prevents the `run.runType` "simplification", which would always be false — the column holds `manual | scheduled`, never `MANUAL_BACKFILL`), `:427-431`, `:927-935`, and `manual-backfill-range.ts:127-134` (`unwrapRunPayload`'s two-shaped input is genuinely invisible from the call site).
- **Consolidate:** the four identical "single time-series node" blocks → one sentence in `CONTRIBUTING.md` plus a one-line pointer per file (and a test — see the third note above, which is the same problem seen from the correctness side). The six-times-repeated "A backfill checkpoints the same way an incremental run does…" preamble in the updated `perDayCheckpoint` suites is fully covered by `CONTRIBUTING.md:73-75`.
- **Trim:** `manual-backfill-range.ts:185-187` enumerates in English the guard on `:195-199`; `:146-148`'s second sentence repeats the `isoDaySchema` comment 110 lines above; `connector-executor.service.ts:672-674` restates `:368-369`; the clamp rationale appears in four places — keep the canonical one at `manual-backfill-range.ts:206-209` (it names the rejected alternative) and point at it from the others.

The `cursor` → `checkpoint` rename (32 diff lines, 2% of the diff) is worth keeping: after this PR one call feeds two different sinks, so "the incremental cursor may only move once a date is complete" would be factually wrong, not merely stale.

<details>

- `connector-executor.service.ts:919-925` — the `persistBackfillProgress` JSDoc says the row "is concurrently written by the terminal status update and by a cancellation, and saving a stale entity would revert them". True for *other* columns, which is why `update`-not-`save` is the right call — but it reads as if `additionalParams` itself were contended. It is not: every writer of that column (`connector-run.service.ts:176`, `data-mart-run.service.ts:618`/`:669`, `ai-insights/data-mart-insight-template.facade.ts:43`) misses connector runs. Reword to "safe because nothing else writes `additionalParams` for a connector run", and name them — otherwise a future `additionalParams.cancelledBy` gets silently clobbered by a checkpoint while the comment reassures the reader it cannot.
- `connector-executor.service.ts:933` — the checkpoint `update` uses `{ id: run.id }` with no `status: In(NON_TERMINAL_DATA_MART_RUN_STATUSES)` guard, unlike both neighbouring writes in the same file (`:129-131`, `:1002-1003`) which carry it and document why. Inert today (cancelled runs are never swept) but inconsistent.
- `packages/connectors/src/Core/AbstractConnector.js:166-169` — the 12th checkpoint site still carries the old guard and the pre-PR comment, with no note. No shipped connector is affected (all 15 override `startImportProcess`), but it is the default a new connector inherits, and `CONTRIBUTING.md` now tells contributors to call unconditionally. Either convert it or give it the one-line "deliberately guarded because…" treatment Shopify and TikTok Ads got.
- `packages/connectors/src/Core/AbstractConfig.js` — the `updateLastRequstedDate` JSDoc still describes the method as incremental-only ("prevents duplicate data requests"). That is the contract surface a third-party host of the published package reads.
- `connector-types/connector-message/schemas/types/message-requested-date.schema.ts:7` — `date: z.string()` now feeds two sinks with different validity contracts: the backfill path normalizes and drops unusable values with a warning, while the incremental path still writes the raw string into `ConnectorState`. Tightening the schema to the day-normalizing shape would make both agree (trade-off: a malformed date becomes `UNKNOWN` at the parser instead of reaching either sink).
- `connector-executor.service.spec.ts` — `createRun()` has no `additionalParams`, so the two headline backfill tests assert `{ additionalParams: { backfillProgress: {...} } }`, a row shape production never produces (a backfill row always carries `{ payload }`). The spread *is* pinned, but only by `:1391`; dropping it at `:937` would keep the headline tests green.
- `connector-executor.service.spec.ts:1244` — `emitCompletedDay` sends `date: '2026-08-11'`, while the wire always carries a full ISO timestamp (`NodeJsConfig.js:122-131` `JSON.stringify`s a `Date`). Narrowing `isoDaySchema` to `^\d{4}-\d{2}-\d{2}$` would silently drop every backfill checkpoint in production and leave the whole executor suite green.
- `connector-run.service.ts:117` — the sweep seam that forwards `payload: run.additionalParams` is the one line the whole resume feature rests on, and `connector-run.service.spec.ts:184` asserts only `expect(createTrigger).toHaveBeenCalled()`; the executor specs hand-construct the replayed shape themselves.
- `connector-executor.service.ts:326` vs `connector-source-config.service.ts:60` — `unwrapRunPayload` runs twice on the same body, so the "can never disagree" comment is stronger than the code; the reassurance actually comes from the DTO rejecting any key other than `runType`/`data` (`run-data-mart-request-api.dto.ts:24`). Unreachable today; consider dropping the second unwrap.
- `connector-executor.service.ts:444-451` — one full `additionalParams` rewrite per imported day per configuration (a 31-day × 3-config backfill is 93 rewrites of a column that also carries the run payload). Coalescing when the day is unchanged would be cheap; matters most on SQLite deployments, which serialise on one connection.
- `CONTRIBUTING.md:70-86` — the new rule "nothing between the fetch and the call swallows an error" is stated absolutely, but FacebookMarketing deliberately skips permanently-unreachable accounts and LinkedInAds checkpoints days whose analytics hit LinkedIn's element cap. Both are defensible (re-running recovers neither), but they are exceptions the rule does not name.
- `manual-backfill-range.ts` — now carries two lifecycle concerns under one backfill-specific name: request-time validation (called from `connector-run.service.ts:48`) and run-time resume/progress (called only from the executor), plus `unwrapRunPayload`, which is not backfill logic at all and is used for every run type. At 218 lines it is 5× the next-largest file in `utils/`. Splitting out `utils/run-payload.ts` would let the rest keep the name it was given.
- `packages/connectors/src/Sources/GoogleAds/Connector.js:78` — six connectors advance the day loop with local-time `setDate(getDate()+k)` (GoogleAds `:78-79`, CriteoAds `:68-69`, MicrosoftAds `:90-91`, RedditAds `:64-65`, OpenExchangeRates `:66-67`, FacebookMarketing `:257-258`), unlike LinkedInAds `:68-69` and XAds `:80`, which use `setUTCDate` and say why. This cannot open a hole — the fetched and checkpointed days come from the same `Date` via `toISOString()`, so the drift can only duplicate a UTC day, and a skip would need a spring→autumn round trip, well beyond the 31-day cap. It does mean a backfill spanning a spring-forward transition on a non-UTC host ends one day short of `EndDate` while re-fetching one day. Entirely pre-existing and unreachable on the shipped runtime (`node:22-slim` defaults to UTC and no `TZ` is set anywhere — `connector-process-spawner.service.ts:48-54` plain-inherits the env); reachable only for self-hosted `owox serve` on a non-UTC host. Worth knowing, since this PR's resume path is what would repair it.
- minor: `manual-backfill-range.ts:166-169` rebuilds the accumulator with a spread per entry (O(n²), irrelevant at this n, but `progress[configId] = day` is simpler); `connector-executor.service.ts:945` introduces `castError` while the other ~8 error sites in the same file use the inline ternary, so the file now does it two ways; `connector-executor.service.spec.ts` declares `emitSuccessMessageFor` at module scope *after* the closing `});` of the top-level `describe` and it duplicates the `emitSuccessMessage` that `createService()` already returns; `runConnectorConfigurations` now takes the whole `DataMartRun` entity where every other private helper in the class takes `runId: string`, and it only ever reads `id` and `additionalParams`.

</details>

**ci-parity (read from `gh pr checks`; no test, build or type-check was run locally):** 21 green, 0 red, 0 pending — lint, Prettier, Markdownlint, every workspace unit suite, Audit, Build, all four API E2E shards and Browser E2E. Changeset present and correctly named (`.changeset/6888-resume-manual-backfill.md`, Fibery id + `'owox': minor`). No migration, no facade, no controller endpoint touched, so the migration, facade-test and `@Auth` gates are not engaged.

<sub>Automated OWOX review · worktree checkout · 5 analysis agents + 3 adversarial skeptics · no test, build or type-check run locally (the DST verdict additionally used standalone `Date`-semantics probes over all 418 IANA zones, touching no repo code). Candidate blockers: 2 through the adversarial gate → 0 confirmed. A DST/UTC resume-skew theory was refuted on the code (and by an exhaustive sweep of all 418 IANA zones); the single-time-series-node gap was confirmed but is latent-only, so it is a note rather than a blocker. A third candidate — cross-package version skew — was disproved by two agents independently before it reached the gate.</sub>

